PR #469, opened 2013-03-21
oauth work

timdewhirst:
This fork now contains a complete implementation of the oauth work; I've added some code to kick off oauth authentication however some further changes need to be made to make the workflows async. which needs some input from you.


DeepDiver1975:
@dragotin @danimo I guess we can happily kill this pr - after two years I do not expect this code to work anyhow. Agreed?

danimo:
@DeepDiver1975 Yes, let's kill it with a weeping eye and let's hope that the server comes up with oauth support that the client can benefit from soon.

PR #1799, opened 2014-05-22
Multiple instances for single user.

ghost:
Usable by --instance <name> parameter.


dragotin:
With 2.0.0 we will have multi account support, so this becomes obsolete.

Thanks anyway!

PR #1874, opened 2014-06-12
1.6 trashbin

moscicki:
This is a prototype of a local trashbin for the client to make 100% that incidental file deletions are not leading to data loss. Any file deleted by the client ends up in <pre> _syncfolder_/.trash/_timestamp_/....</pre>

This has the following advantages:
- it is easy to cleanup the trashbin manually (either entire trashbin or based on timestamps)
- the directory structure is preserved - it is easy to find and restore directories if needed 
- the timestamp has "yyMMdd-hhmmss" format so it is  assumed that there is at most one sync run started per second (per sync folder)

Basic tests passed on Linux. NOT TESTED on Windows/Mac. I would see this feature as an option in the config file.


cabillman:
This feature is huge for us. I just had a user who lost a large number of files and this feature would have saved them. I'm considering manually patching and rebuilding our clients to include this functionality until something gets merged upstream. 

What is the quickest path to getting something like this accepted into the client? If this patch were cleaned up with a flag in the config/gui would that be enough? 

My specific case was nasty. The users files were not on the server long enough to be pulled into our nightly backups.

moscicki:
@cabillman: please discuss with @dragotin and @MTRichards

danimo:
@cabillman If you are willing to build such a feature, why not contribute it upstream in the first place. After all, ownCloud is an Open Source project, and collaboration is the very nature of open source development.

moscicki:
@cabillman: could you describe your incident? What happened and which client version?

cabillman:
@danimo I certainly would open source any change. I only mentioned patching our own clients because I would love to have this ASAP even if it isn't acceptable to the community at large in its current form.  This  proposed patch has been sitting here for almost a year already.

@mosciki I'm digging through the logs still and will open up bug reports. Client was version 1.8.0 and the files are not on the server even at the file system level.

From my perspective the specific bug isn't important. There will always be another bug. I keep extensive nightly backups including deleted files. But in this particular case the files were uploaded and deleted in the same day so they never appeared in a backup.  The files were deleted by the client and thus not in the recycling bin.

moscicki:
@cabillman: what you say about bugs is true and that was also my intention when I created this feature. On the other hand, I would hope to be as confident with the sync client robustness as I with my local linux filesystem. It's a long way but I hope the current development goes in this direction and that at some point the core codebase will be stable and mature enough to really reach another level of quality. 

For 1.8.0 client: there was a bug which was deleting files without reason (apperantly fixed in 1.8.1). Check with the developers. I am not clear if this should only affect the files that were previously uploaded to the server or all files (the second case is of course a distaster).

enoch85:
I haven't read all posts, but wouldn't the .trash conflict with the .trash that exits on Ubuntu Desktop systems? I have it on 15.10 anyway.

ckamm:
@moscicki This has been open for a long while. Are you still interested in pursuing this beyond the prototype? Otherwise I'd close it.

enoch85:
I'm surprised that nobody choose to continue to work on this and closes it instead. It's half done and seems to be a great feature. Just sayin.

DeepDiver1975:
> I'm surprised that nobody choose to continue to work on this and closes it instead. It's half done and seems to be a great feature. Just sayin.

open source is about doing it ... so do it 😉

enoch85:
@DeepDiver1975 If I could I would. :)

guruz:
https://wiki.qt.io/New_Features_in_Qt_5.15

>  QFile::moveToTrash provides a cross platform API for moving files to the trash/recycling bin on Windows, macOS, and Linux desktop (implementing the freedesktop.org specification).

FYI @michaelstingl

michaelstingl:
Hm, sounds interesting. But what does it mean with local files that are virtual/dehydrated? Needs another look… Open a new issue?

/cc @TheOneRing

TheOneRing:
Just won't work with virtual files

PR #2174, opened 2014-09-05
mtime of local folders is not set, so we should not check on it either.

tim-vandecasteele:
As mentioned in https://github.com/owncloud/core/issues/7009#issuecomment-36616528 the modification time of a local folder is never touched. But in csync_update there is a check if the local mtime matches with the one from the database.

This commit doesn't trigger any action when the local modification time of a folder and the database differ. In the old scenario a CSYNC_INSTRUCTION_EVAL was triggered, resulting in a lot of additional useless propfinds for the folders.


danimo:
@ogoffart ping?

ogoffart:
I am closing this merge request because i don't think this fixes anything.
There are cases however in which we may rely on directory to have a different instruction. (altough this may not be the case anymore)

PR #2195, opened 2014-09-12
1.7 - Adding support for hires-retina state icons.

ladrua:
Adding support for hires(retina) display icons. Tested on 1.6.3, and only on the state icons.
Still need the actual *@2x.png files to be made.


ogoffart:
1.8 now has support for hi-dpi.
I will close this merge request.
Feel free to open another bug or merge requests if there is more items to fix.

PR #2308, opened 2014-10-12
improve owncloud --version

jnweiger:
Currently it prints 1.7.0 although we are in beta state before 1.7.0 -- should it print 1.7.0~beta3 instead?
Suggested by this PR.

Note that I also suggest prefix MIRALL_VERSION_SUFFIX  in VERSION.cmake with a debian style '~' character. Maybe OwnCloudCPack.cmake should also do that?
Also made MIRALL_VERSION_STRING show the GIT_SHA1 when no numeric BUILD number is there.
This needs a move in CMakeList.txt -- hopefully without other side effects.

Improving on https://github.com/owncloud/mirall/issues/2211 (somewhat optional, but hey)


danimo:
@dragotin do we want to merge this?

ogoffart:
Commit d2848b9c26f36ac3dad8cb47f1fc87fd7dc795bd already fix it partially.
It does not include the git sha1 into the version.
But this pull request is outdated now and it cannot be merged anymore.

Issue #2215, opened 2014-09-16
[OS X] Yosemite "Dark Mode" menubar icon stays black on dark background


StrangeNoises:
OSX Yosemite has a "dark mode" for the user interface, in which the menubar background goes from light grey to dark grey and the text and most menubar app icons are white.

Owncloud with monochrome icons selected doesn't adapt to this. The icon stays very dark grey, rendering it almost invisible on the dark background.

(Obviously no background-colour awareness either when colour icons selected, but at least visibility isn't an issue then.)

Desired behaviour: Awareness of the configured foreground and background colours in the menubar in Yosemite, and changing the colour or inverting the palette of the monochrome icon to match. FWIW the newest Viscosity (1.5.1) does invert the palette as you'd expect when using its new default icons, so it's not just Apple's own components that know how and when to do it. :)

I tried restarting it, in case it only picks up the colours on startup, but it didn't help. (In fact, upgraded belatedly from 1.6.2 to 1.6.3 - that didn't help either.)

Screenshot shows the problem. You're in good company of course, if you check out the icon of the app next to the Owncloud one. ;-)

![screen shot 2014-09-16 at 23 23 22](https://cloud.githubusercontent.com/assets/1295001/4296051/fcf4ad4c-3df0-11e4-9951-6aa6454fc582.png)

I assure you the owncloud icon is there, in the middle, a very slightly lighter grey than its background.

soDEVch:
I'm using dark mode on OS X 10.10 and I'm always having trouble finding the icon in the menubar. Would really be nice to have the white icon. As mentioned above owncloud isn't the only app still having this issue. How soon can we expect this to be fixed?

leipert:
Corresponding qt bug report here: 
https://bugreports.qt.io/browse/QTBUG-42109

AlexanderWillner:
Seems that the corresponding change in QT has been merged in July 2015: https://codereview.qt-project.org/#/c/115120/

chesterdkat:
Just checked with Version 2.0.2rc1 (build 2790) and this bug is still present... 

THANK YOU!

guruz:
@AlexanderWillner thanks for digging this patch out.. we will need to modify our ownCloud-Qt for OS X

@dragotin @danimo

StrangeNoises:
confused now because I thought this had been fixed. Is for me anyway in 2.0.1 release. Is it a regression in 2.0.2 builds?

[...]

PR #2393, opened 2014-10-24
Switch monochrome systray icon color depending on systray brightness

svkurowski:
I am working on issue #2215 in this branch.

As of version 10.10 Yosemite, Mac OS X offers the user the option to use a dark version of the dock and menu bar. The systray icon was black even on the dark menu bar black-on-dark, thus rendering it nearly impossible to spot.

This patch checks the NSUserDefaults to check whether the user has chosen the black dock and if so, will use the white monochrome versions for the systray icons.
White icon on dark menu bar: white-on-dark

Right now, the systray icon will not change "live", that is it will stay in the old color when the user changes the dock/menu bar brightness during execution of the app until either the icon image changes (because of state change) or the app is restarted.

The patch needs to be tested on Unix and Windows systems, unfortunately I don't have a development machine setup for those operating systems at the moment. The systray icon behaviour should be unchanged for those operating systems and for earlier versions of OS X.

### DIFF
--- a/src/libsync/theme.cpp
+++ b/src/libsync/theme.cpp
@@ -15,6 +15,7 @@
 #include "theme.h"
 #include "version.h"
 #include "config.h"
+#include "utility.h"
 
 #include <QtCore>
 #ifndef TOKEN_AUTH_ONLY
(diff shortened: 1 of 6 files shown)


svkurowski:
Sorry for the delay. I cherry-picked the commit into the 1.7 branch in 01a72121049a103727b47cb0bde2db2d4d311b88

danimo:
Why, there will be no 1.7 release, and 1.8 is pending.

svkurowski:
I was not aware of that. It doesn't hurt though I guess. If you want me to, however, I can revert it.

PR #2597, opened 2014-12-08
Propagator on Windows: Allow other applications to read/delete a file

guruz:
...while we upload it.


ckamm:
I looked at qfsfileengien_win.cpp in Qt 5.4: files are already opened with FILE_SHARE_READ and FILE_SHARE_WRITE! That means the main change here would be:
* FILE_SHARE_DELETE (maybe this is what's needed?)
* FILE_FLAG_BACKUP_SEMANTICS (I'm not sure whether this does anything if we don't have the backup/restore privileges)

ckamm:
Cleaned up and merged as b3c0279

Issue #2607, opened 2014-12-10
Shibboleth-403-error not shown via sync-client


michaelstingl:
Users login via Shibboleth, but it is only allowed for users that have the right attribute. Users that don't have the attribute will get a 403-error page.

The problem is that, when an user tries to login via the sync-client, he will never see the 403-error. So the user is completely unaware "what is wrong".

Can the sync-client "understand" when there is no access granted after a successfull shibboleth login.

@dragotin @danimo 

00002241

cdamken:
@MorrisJobke: @danimo made some modifications that helps for now, but the solution will be probably in 1.7.2 or 1.8. it was not possible for 1.7.1

danimo:
> Improve user experience (currently the wizard at this point assumes that HTTPS is broken/misconfigured and asks the user if he wants to try HTTP only. There are some things we can do to make that guess more correct.

This one just landed in master.

craigpg:
Here's what you see in the setup wizard after authenticating when you don't have proper permissions:

![image](https://cloud.githubusercontent.com/assets/160667/5651116/bc1795f2-9672-11e4-8ceb-b1f0c1d03595.png)


Since this is not a client-specific error condition, the user will run into the permissions issue when trying to access their ownCloud via the browser.  At least in this particular case, when going through the browser they get a much more useful error page:

![image](https://cloud.githubusercontent.com/assets/160667/5651160/17ff8a1e-9673-11e4-985a-fe7e0a022011.png)

We should enhance the "Error downloading..." message in the wizard to suggest the user try to access the service directly in the browser and provide them the link to click on.  For example, "To ensure you have proper access, please try accessing the service directly in your browser: https://.../files"

danimo:
@michaelstingl Does what Craig proposes sound ok to you?

danimo:
@michaelstingl ping?

cdamken:
@danimo That solution sounds good to me :+1:

michaelstingl:
@danimo is @craigpg 's proposition "To ensure you have proper access, please try accessing the service directly in your browser: https://.../files" in master yet?

danimo:
@michaelstingl No, I will tackle this in the next days.

danimo:
@michaelstingl This is what it looks like now: ![image](https://cloud.githubusercontent.com/assets/1315934/5935630/2d2aa138-a6dd-11e4-9c1f-25fdfe001a8b.png)

danimo:
[...]

PR #2626, opened 2014-12-17
Do not show the HTTP-reconnect popup when its not warranted

danimo:
Fixes usability issues in the workaround of #2607


danimo:
Manually included this commit (which was against 1.7) into master, and amended it.

PR #2634, opened 2014-12-21
Support NetBSD like FreeBSD or Linux.

ryo-on:
Please support NetBSD Operation System.
It has no builtin inotify, use libinotify like FreeBSD.
statvfs part should be like OS X or FreeBSD.
Other parts should be like Linux.

This is licensed under MIT license.


danimo:
I'll fix that in master.

danimo:
Please check c86d8e5acf074cfa977d680de5b17117f4909cba.

ryo-on:
It works fine.
Thank you.

PR #2638, opened 2014-12-23
fix link and add slash as bad character

MorrisJobke:
Found this by refering to the documentation for https://github.com/owncloud/core/issues/13005

cc @RealRancor


danimo:
Thanks!

@JimLoose Please request reopening of bugs in the respective report. You can also comment on closed bugs.

Issue #1444, opened 2014-02-18
Allow Sharing from the Desktop


MTRichards:
As a user, I want to be able to click on a file, and select share with ownCloud, and then be able to share a file with someone as a link, or as a file shared directly with them in ownCloud using the share API.

This is partly a deeper integration into ownCloud and the desktop OS to enable the menu, and partly an implementation of the sharing API on the desktop.  Likely there are two steps:
1) share with link (share, unshare)
2) share with user of the system (share, unshare, set CRUDS permissions)

dragotin:
@rullzer would you mind to create a submit request against owncloud/client master branch? That would be nice.

dragotin:
I merged the branch of @rullzer today (thanks!) and adopted the Nautilus plugin accordingly, so that sharing from Linux Nautilus is possible now.

Remaining tasks:
* [x] SocketAPI share commando
* [ ] Windows Explorer plugin adopted
* [ ] Mac Finder plugin adopted
* [x] Nautilus plugin adopted
* [ ] KDE Dolphin plugin adopted (not release critical)

rullzer:
Public url sharing should be working pretty OK in that merge.

However sharing with users and groups (and modifying access) is still under development. I'll hopefully submit a patch for that soon. Then somebody can play with making it look decent.

rullzer:
I just created another pull request https://github.com/owncloud/client/pull/2700
This allows user/group sharing. Still not complete. But maybe somebody can have a look at the UI already :)

jancborchardt:
@rullzer awesome! Some feedback regarding the sharing dialog:

* [ ] It needs to be clearer what will be shared. On the top of the dialog there should ideally be a thumbnail and the filename (or folder name, or multiple file names).
* [ ] Merge the user/group field into one and label it »User or group …«. The label above should only be »Share with«
* [ ] We need type-ahead autocompletion for the user/group. Otherwise you need to know the exact username. A server-API is needed for that (cc @DeepDiver1975) – and we shouldn’t offer user/group sharing until we have it since we can’t expect people to know usernames of people.

alefattorini:
@jancborchardt can I help? How can I test this dev version on my pc?

jancborchardt:
@rullzer @dragotin @guruz basically what I’m saying is that maybe it’s wiser to start with only the »Share by link« feature in the dialog, and postpone the sharing to other users. Until we have the proper API to return users from the server.

Bridgesltd:
[...]

PR #2700, opened 2015-01-14
Sharedialog: User/Group sharing 

rullzer:
Some patches to allow sharing with users and groups.
Also some patches make a start on properly checking return values.
Some commits are a bit dirty and clearly needs fixing. But I figured maybe somebody can have a look already.

The UI needs some work. Which could probably lead to more code reduction.

More commit coming soon I hope :)


rullzer:
As discussed in #1444 I modified the code so we only share links for now. Together with more fixes this now starts to look like something we could actually use ;)

dragotin:
Thanks, but can you please apply your merge request based on master? The share dialog branch is quite outdated.

rullzer:
New pull request created against the master branch wich merges a lot cleaner. So closing this one.

PR #2730, opened 2015-01-21
Sharedialog: 404 is also an acceptable response when retriving shares

rullzer:
When we try to get the shares for a file/dir that has no shares yet 404 is an
acceptable response from the server.


LukasReschke:
I'd also check for fatal 500 errors

LukasReschke:
(or is that check performed elsewhere?)

rullzer:
It is a bit reversed. We show a warning box if the code is not 100 or 404. So all others are caught. But maybe we should have a part where check for fatal errors.

This patch is mainly ment to get rid of the dialogbox that pops up if you try to share a previously unshared file/dir

Issue #2719, opened 2015-01-19
[windows] When creating explorer favorite use more specific windows functions


remixtj:
I'm experiencing problems with the creation of the favourite on Windows 7, using folder redirection. I made some research for the problem and found out that client tries to create a link on the wrong place. My "Links" directory is on a cifs share: \\server\homes\username\Links. This wrong behavior is caused by the function setupFavLink_private in [utility_win.cpp](https://github.com/owncloud/client/blob/master/src/libsync/utility_win.cpp)

At the moment in function _setupFavLink_private_ at  [line 24](https://github.com/owncloud/client/blob/master/src/libsync/utility_win.cpp#L24) the function

```
    SHGetSpecialFolderPath(0, path, CSIDL_PROFILE, FALSE);
```

is used.
[MSDN documentation](http://msdn.microsoft.com/en-us/library/windows/desktop/bb762204%28v=vs.85%29.aspx) says that [ShGetFolderPath](http://msdn.microsoft.com/en-us/library/windows/desktop/bb762181%28v=vs.85%29.aspx) should be used. But says also that ShGetFolderPath is deprecated. A note states:

> Note  As of Windows Vista, this function is merely a wrapper for SHGetKnownFolderPath. The CSIDL value is translated to its associated KNOWNFOLDERID and then SHGetKnownFolderPath is called. 
> New applications should use the known folder system rather than the older CSIDL system, which is supported only for backward compatibility.

Please note that SHGetKnownFolder is a function available only on Vista and later. But, you can also note that also favourites on explorer are available since that windows release.

I made a test and tried to print the path value that is returned by the function call:

```
    SHGetKnownFolderPath(FOLDERID_Links, 0, NULL, path);
```

and i get (correctly) the value of _\\server\homes\username\Links_.
I printed also the output of 

```
    SHGetSpecialFolderPath(0, path, CSIDL_PROFILE, FALSE);
```

and i got _C:\Users\username\_, that is correctly my profile home, but the path where the function will create the link is _C:\Users\username\Links_, that is wrong.

remixtj:
@danimo what do you think about this new patch? better?

jturcotte:
Reopening since we had to revert this in https://github.com/owncloud/client/pull/3562

hodyroff:
Is this current in master? What exactly is the issue from a UX perspective?

remixtj:
Hi,
[...]

PR #2738, opened 2015-01-22
Proposed fix for #2719

remixtj:
Tested on Windows 7, works well. Needs testing on other Windows version, specially earlier than Windows Vista because SHGetKnownFolderPath is available since that version

Should solve issue #2719


remixtj:
@danimo this should be safer. In this case the symbol is dynamically loaded and if not existing will use the legacy one (that exists for sure)

remixtj:
I'm preparing a new fix using the suggestions you gave me. Closing this PR and then will open a new one with the fix.

PR #2749, opened 2015-01-26
Update Toolchain-mingw32-openSUSE.cmake

vtsingaras:
Fix cross-compiling for windows. change qt4 to qt5 since windows and OSX require Qt5 now.


danimo:
Thanks

PR #2778, opened 2015-02-04
Allow building natively on Windows with MinGW

jturcotte:


guruz:
CC @kwillems

csync/CMakeLists.txt
```diff
@@ -28,10 +28,13 @@ if (NOT WIN32)
     find_package(Iconv)
 endif (NOT WIN32)
 
+find_package(SQLite3 3.8.0 REQUIRED)
```
guruz: Heads up to @dragotin: This might break builds for older Linuxes?

jturcotte: This line was copied from /CMakeLists.txt.

Adding the find_package is the way I found to get the defines done in cmake/modules/FindSQLite3.cmake to apply in that file too. Let me know if there's a better way, I'm a cmake noob.

PR #2793, opened 2015-02-07
use touch from path

glensc:
on some systems touch(1) is in /bin, not /usr/bin.
instead hardcoding another path, take it from $PATH, like other commands used here (rm, echo)


danimo:
Thanks!

Issue #2726, opened 2015-01-20
Show file Thumbnail in Share dialog


dragotin:
The share dialog would look even more cool if it had a thumbnail of the file to share. It can be retrieved from the server using this api:
https://github.com/owncloud/core/blob/84bb4cc2e9013eb3db8d7114a0e6eb132a723adf/apps/files/appinfo/routes.php#L17-L20
by doing a GET on 
index.php/apps/files/api/v1/thumbnail/{x}/{y}/{filename}

rullzer:
Ah that was the API I was looking for! I'll try to pick this up shortly.

rullzer:
Initial commit is available. More checking needs to be done but it is working pretty well for now. But no rush since it only makes sense to create a pull request when owncloud 8 is out.

DeepDiver1975:
@dragotin please get in touch with me to find proper solutions before using some crazy api existing somewhere . we are currently navigating into an api nightmare  because a hell of different apis are used - this will soon remain unmaintainable for us. THX

DeepDiver1975:
> The share dialog would look even more cool if it had a thumbnail of the file to share. It can be retrieved from the server using this api:

The file is existing on the client as well - right?
Is there any possibility to generate the thumbnail local on the client?

Getting the thumbnail from the server will put additional load to the server.

THX

DeepDiver1975:
In case you really want to grab it from the server we don't need no capabilites api.

In case thumbnails are disabled the api route simply will not exist and a request to it will return 404

dragotin:
There is no easy way of creating thumbnails on the client at least for Linux, also no *easy* one on Win. And for the sake of consistency, it would be better to get it from server.

However, no rush here, that is very low priority IMO.

rullzer:
@DeepDiver1975 yes that is true and it is how it is implented currently (well actually I check for 200). I'll create a PR for the client later tonight (it is pretty straight forward). We could just leave it like that for now and use the capabilities later when (if) that gets sorted out.

DeepDiver1975:
> There is no easy way of creating thumbnails on the client at least for Linux, also no easy one on Win. And for the sake of consistency, it would be better to get it from server.

okay - understood - THX.

DeepDiver1975:
[...]

PR #2805, opened 2015-02-09
Sharedialog: retrive thumbnail from server for shares

rullzer:
Since owncloud 8 is out this is the PR for thumbnails as described in #2726


rullzer:
Fixed in #3426

Issue #2803, opened 2015-02-09
[1.8.0] Client uses old share URL when sharing files


LukasReschke:
Pre-ownCloud 8 URL: `http://localhost/core/public.php?service=files&t=34grKMpDGWxAU3h`
ownCloud 8 URL: `http://localhost/core/index.php/s/34grKMpDGWxAU3h`

Anyways, the old URL will also work on 8 as it has a redirect as fallback. URL looks just a little bit terribly long in the share dialog.

rullzer:
PR #2813 makes the client use the new url if the server version is >= 8.

@LukasReschke it seems the OCS API still returns the old URL when creating a share..

MegaV0lt:
Is it possible and/or useful ti integrate a kind of "shortener? Ther exist many like j.mp 
is it even possible to integrate a own one? (oc.dyndns.org/hE5hX) or similar?

rullzer:
This should be fixed now.
The server now returns the new URL (which we use if available) and else we generate it ourself the new fancy url if the server is new enough.

dragotin:
reopening for test. @Dianafg76

Dianafg76:
I tested de issue and the server returns the new URL. I close the issue
Desktop v1.8.2 (build 2359)

PR #2813, opened 2015-02-10
Sharedialog newurl

rullzer:
Use the new owncloud 8 sharing url if the server is owncloud version 8 as stated in #2803.

### DIFF
--- a/src/gui/sharedialog.cpp
+++ b/src/gui/sharedialog.cpp
@@ -256,7 +256,13 @@ void ShareDialog::slotSharesFetched(const QString &reply)
                 _ui->checkBox_expire->setChecked(true);
             }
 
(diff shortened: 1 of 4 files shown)


dragotin:
@rullzer pls check d37dd040d0a121f5ad14228e0440f2582bfd519a - that is basically this PR but @ogoffart has committed the account changes before. That makes this one obsolete. Thanks!

rullzer:
@dragotin looks good. However now we use the url given back by the API. Which is still the old one.

The second part of https://github.com/rullzer/client/commit/634077cbf1f200adb68b11a30014d3aec2e59cbf is to prevent that. Also since the capabilities API is not yet in place we do not know if an expiration date is enforced. If we just re-fetc the shares that is at least partly solved.

rullzer:
I'm busy on a new PR. That will include the enhancement here. So I'll close this ticket for now.

PR #2859, opened 2015-02-18
Use a themable WebDAV path all over.

dragotin:
This is needed for enterprise#481

@danimo @ogoffart please review.


ogoffart:
I would like to avoid too much coupling between libsync and the theme.
In libsync, please use the path from the Account class rather than the one from the Theme.

src/libsync/owncloudpropagator.cpp
```diff
@@ -341,7 +343,7 @@ void OwncloudPropagator::start(const SyncFileItemVector& items)
 bool OwncloudPropagator::isInSharedDirectory(const QString& file)
 {
     bool re = false;
-    if( _remoteDir.contains("remote.php/webdav/Shared") ) {
+    if( _remoteDir.contains( Theme::instance()->webDavPath() + QLatin1String("Shared") ) ) {
```
ogoffart: Please use the webdav path from the account instead of from the theme here.

dragotin:
@ogoffart: see my last patch

ogoffart:
:+1:

dragotin:
superseeded by https://github.com/owncloud/client/pull/3802

Issue #2869, opened 2015-02-20
Missing copyright header for some files in 1.8.0beta1


hefee:
there are some files without a copyright header:
* src/gui/addcertificatedialog.cpp
* src/gui/addcertificatedialog.h
* src/gui/sharedialog.cpp
* src/gui/wizard/owncloudconnectionmethoddialog.cpp
* src/gui/wizard/owncloudconnectionmethoddialog.h

dragotin:
@rullzer @nocteau @jdhoek please add your emails and names if you want.

jdhoek:
@dragotin 
Sure. Just to verify: I consider my contribution to be utterly trivial, I take it you agree (so no Contributor Agreement needed right?).

dragotin:
@jdhoek well, yes, yet, it might be better to send one in if you don't mind. Puts us all on the save side.

jdhoek:
Not at all. Will do.

The files I've contributed were amended by others afterwards, should I submit a PR with the (now missing) default header as used in other files with just my name on it?

rullzer:
done :)

PR #2890, opened 2015-02-25
Add copyright header

jdhoek:
Added to two new cpp/h files introduced in an earlier commit.

As per #2869.


danimo:
Obsolete. Already fixed.

PR #2943, opened 2015-03-11
[WIP] Initial commit to use the capabilities api for better UX

rullzer:
* Make use of available information to give a better UX
  * Ask for password if required
  * If password/expire date are enforced make sure it is like that

Currently I use https://github.com/owncloud/core/pull/13964 as a basis. But due to the new class it should not be to
hard to change until that PR is stable and merged.

-----

Once https://github.com/owncloud/core/pull/13964 is merged (be it with some changes) and the API is stable this can be merged. For now this PR is just to keep track of the work.

### DIFF
--- a/src/gui/sharedialog.cpp
+++ b/src/gui/sharedialog.cpp
@@ -25,6 +25,7 @@
 #include <QBuffer>
 #include <QFileIconProvider>
 #include <QClipboard>
+#include <QInputDialog>
 
 namespace {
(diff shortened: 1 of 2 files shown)


ogoffart:
We already have the capabilities in Account::capabilities.  Or is that not the same?

rullzer:
Yes it is the same... this was just an abstraction.
Let me see tomorrow if I reuse this or create a new PR...

PR #2966, opened 2015-03-16
Only show share dialog if we are connectedt

rullzer:
Thanks to the lousy wifi connection at the airport I stumbled upon this.

We only want to show the share dialog once we are connected. Easy fix.


rullzer:
@ckamm thanks for the hint. Did not look into it that deep yet. Fixed now.

ckamm:
Manually fixed and merged into 1.8.

Issue #2889, opened 2015-02-25
Feedback on settings in sharing dialogue


plages:
I noticed that the desktop sharing takes over the default expiration time from the server. Which makes perfect sense to me. Does anyone agree therew should be feedback about this the first time i create a share? When i click 'Share with ownCloud'the second time on a share, the expiration date and the 'Set expiration data'checkbox are automatically populated. Maybe upon initial creation of the share someting like 'Default expiration time of %serversetting% applies' should appear in the dialogue.
Or else the dialogue should display the same behaviour as when clicked for the second, third etc. time.
Also when expiration and password are mandatory on the server the checkboxes should be enabled by default in the sharing dialogue.

rullzer:
I am working on the mandatory password and expiration date. However this requires changes in the server as well. But it is coming.

dragotin:
Hmm,  I reworked the dialog for the final again, so some of the mentioned issues should have gone already.

plages:
The default expiration time is set on a share when created but this is not reflected in the interface until the user clicks "share with ownCloud" the second time. But maybe this is due to the fact that i'm still on 7.0.4 on the server side.

rullzer:
A fix for not displaying the expiration date is in PR #2989. This forces retrieval of the shares on share creation. Since we get much more info that way.

plages:
Great that'll solve the issue. Is this also possible for the password checkbox in case the server has the password set to mandatory?

rullzer:
That should already happen.

Dianafg76:
Cheked for the password checkbox in case the server has the password set to mandatory

PR #2989, opened 2015-03-19
On creation of a share always retrieve the share

rullzer:
This makes sure that if a default expiration date is set this is reflected in
the dialog. 

AS described in #2889


ckamm:
@rullzer @dragotin This should probably go into the 1.8 branch. Shall I test and merge it there?

rullzer:
@ckamm also in the 1.8 branch. Probably good to have it in master as well.

ckamm:
@rullzer We sometimes merge 1.8 into master, so everything that shall be in both should go into 1.8. I'll take a look now.

ckamm:
Merged into 1.8. It'd be nice if the share creation call already contained these details...

PR #2997, opened 2015-03-22
fix python3 syntax errors

hefee:
The print statement needs brackets.


ogoffart:
And does it still work with python2?

danimo:
Yes, you can add any amount of braces in python.

hefee:
@ogoffart "And does it still work with python2?"
If it compiles under python2 is complety uninteressting. The File header says 
  #!/usr/bin/python3

=> it is a python3 script.

dragotin:
hmm, do we really have to force it to python 3? There might still be systems out that default to python 2 and we do not pull python3 for them, right?

PR #3052, opened 2015-03-30
Sharing UI - complete overhaul

stonerl:
Here you can see what it looks like now
screen shot 2015-03-30 at 18 26 41
screen shot 2015-03-30 at 18 26 48
screen shot 2015-03-30 at 18 29 43
screen shot 2015-03-30 at 18 34 24

### DIFF
--- a/src/gui/sharedialog.cpp
+++ b/src/gui/sharedialog.cpp
@@ -77,7 +77,7 @@ ShareDialog::ShareDialog(AccountPtr account, const QString &sharePath, const QSt
     _ui->pushButton_setPassword->hide();
 
     _ui->calendar->setDate(QDate::currentDate().addDays(1));
-    _ui->calendar->setEnabled(false);
+    //_ui->calendar->setEnabled(false);
 	_ui->calendar->hide();
 
     QFileInfo f_info(_localPath);
@@ -292,6 +292,10 @@ void ShareDialog::slotSharesFetched(const QString &reply)
 				_ui->checkBox_expire->setEnabled(true);
             } else {
                 _ui->checkBox_password->setChecked(false);
+				_ui->checkBox_password->setEnabled(true);
+				_ui->checkBox_password->setText(tr("Set p&assword"));
+				_ui->lineEdit_password->setPlaceholderText(tr("Choose a password for the public link"));
+				_ui->pushButton_setPassword->setText(tr("Set p&assword"));
                 _ui->lineEdit_password->hide();
                 _ui->pushButton_setPassword->hide();
             }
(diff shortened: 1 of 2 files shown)


danimo:
Nice. Ideally, the password field and the expiration date shouldn't expand all the way to the right.

stonerl:
I redesigned to sharing UI to achieve less clutter. Here are some screenshots.

1. General View
screen shot 2015-03-31 at 11 02 14
A window can't get wider than you see in the picture. I limited the link-label to 505, even if a link would look like this: https://www.myfantasticdomainname.com/owncloud/index.php/s/K0xqBAp0iGkwxjk

2. Setting a password
screen shot 2015-03-31 at 11 02 35
3. Enforcing passwords
screen shot 2015-03-31 at 11 42 41
The "Set password" check box cannot be unset and the link-label is used to inform the user. The "Set expiration date" check box is also disabled and only the password can be typed.

4. Passwords can be changed but not remove
screen shot 2015-03-31 at 11 37 02
This is, because if a password is enforced by the server, in the current implementation the users can remove the password and gets an API-Error-Message at the bottom at the window (404).

5. Setting the expiration date
screen shot 2015-03-31 at 11 03 08
6. Example for an API-Error
screen shot 2015-03-31 at 11 24 32

danimo:
This actually clutters more than before. 100% width buttons are never a good idea. Why not go with the proposal I made yesterday? We were pretty close.

stonerl:
Hm, the idea I head in mind was that all buttons  will always have the same size.  When I limit the date and password field it would look like this.
screen shot 2015-03-31 at 12 21 24
All the buttons and fields have different sizes and there is so much free space between them, imho this looked much more cluttered. That is why I came up with this idea. I could set a max size for the buttons and fields of, for example 300 so they won't max out. Currently the max size would be 517 since I limited the link field and a window won't get wider then 541. This would then look like this:
screen shot 2015-03-31 at 12 31 24

stonerl:
@danimo maybe I went a little bit to far. So I took your yesterdays proposal and made only some slight changes. I hope this will get your blessing :)

screen shot 2015-03-31 at 15 37 45
screen shot 2015-03-31 at 15 37 55
screen shot 2015-03-31 at 15 38 14
screen shot 2015-03-31 at 15 38 22
screen shot 2015-03-31 at 15 38 50
screen shot 2015-03-31 at 15 39 53

danimo:
Better, but please don't grid-align "Set password" with "Copy link to clipboard". Rather move the "set password" button next to the line edit. They belong together and the spacing looks weird.

stonerl:
OK here it is:
screen shot 2015-03-31 at 20 07 36

stonerl:
I made some changes and like to hear your opinion. First of all I took the freedom to rename the check-box "Set password" to "Password protect" to match the wording with the share-dialogue from the Web-UI.

The second change I made concerns the link. Currently only the "openExternalLinks" property is set. I added the "LinksAccessibleByMouse" property. Now a left-click opens the link in the browser and a right-click gives you a context-menu that lets you copy the link.

Therefore I changed the "Copy password to clipboard" button to "Email Link". Clicking this button while create a new email in your default mail client and adds the link to the email-body.

You can see all the changes in the picture:

screen shot 2015-04-05 at 21 40 59

What do you think?

dragotin:
We have gone through more reworks of this dialog with @jancborchardt 

This pr is hanging for quite some time, so I hope you're fine with the current results. 

Thanks for jumping in anyway, but I am closing this now.

PR #3071, opened 2015-04-07
Fix spelling of German "Synchronisation"

mabl:
See http://www.duden.de/suchen/dudenonline/Synchronisation


danimo:
Thank you for your contribution. However, client translations are managed through Transifex and will be overwritten from there. I'd hence ask you to contribute the fixed through https://www.transifex.com/projects/p/owncloud/translate/#de_DE/client/4633187. Thanks.

PR #3072, opened 2015-04-07
added adler32 implementation; use OC-Checksum header instead of OC-MD5

moscicki:
for testing env variable OWNCLOUD_CHECKSUM_TYPE may be set to Adler32, otherwise default is MD5

This aligns better with the protocol specified in: https://github.com/cernbox/smashbox/blob/master/protocol/checksum.md

### DIFF
--- a/src/mirall/filesystem.cpp
+++ b/src/mirall/filesystem.cpp
@@ -22,6 +22,9 @@
 #include <qtconcurrentrun.h>
 
 
+#include <zlib.h>
+
+
 #if QT_VERSION < QT_VERSION_CHECK(5, 0, 0)
 #include <qabstractfileengine.h>
 #endif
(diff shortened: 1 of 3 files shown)


stonerl:
Has any of you read this paper?

http://www.zlib.net/maxino06_fletcher-adler.pdf

dragotin:
@stonerl I haven't yet, but what is the point? Why do you think we should have read it?

stonerl:
@dragotin Just let me cite the conclusion:

"The error detection properties of the Fletcher and Adler
checksums are very similar. However, based on our studies
of undetected error probabilities, the Fletcher checksum is
often a better choice. It has a lower computational cost than
the Adler checksum and, contrary to popular belief, is also
more effective in most situations."

dragotin:
@stonerl thanks! Do you happen to know if there is a freely available implementation of the fletcher sums?

moscicki:
The point is: our server uses adler32. So it is good to research on better options but we need adler32 anyway to work with our server. The proposed protocol extensions for checksumming are flexible to allow implementation of any type of checksum one wants.

moscicki:
And by the way, we checksum 130PB of data with Adler32 so I would not worry too much about it.

stonerl:
@moscicki well you have a point there :-)

@dragotin 
http://srecord.sourceforge.net/srecord/fletcher32_8h_source.html
https://github.com/bachp/alljoyn_core/blob/master/src/Fletcher32.h

PR #3086, opened 2015-04-09
recall file prototype

moscicki:
This is a very simple prototype to illustrate what we need to achieve.

Here is how it works:
 - admin triggers a recall by putting a file .sys.admin#recall# in chosen directory on the server which contains a list of files to be recalled; the list may contain relative paths to files in subdirectories
 - recalled files are uploaded with a modified name (similar to the conflict file naming) and with a special request header OC-Tag

Tested on our server which currently ignores the presence of OC-Tag and simply puts the files in the same directory on the server (within the user's space). Desired server-side behaviour is to place all files with OC-Tag in some dedicated staging area not visible directly by the user.

Note that this mechanism will also allow to recall files from shared folders from other users.

It would be best, of course, if none of these files would be visible on the client (or at least hidden). Also, security implications need to be analysed.


dragotin:
I rebased the checksum_verify branch to the latest 1.7.1 tag (v1.7.1-qt54) and pushed that.

dragotin:
About the recall feature:
- why do you want to upload the file with a different name?
- what happens if the recall file is put into a directory on the server that is shared with multiple users? Wouldn't that result in having all the clients that sync the shared dir reuploading?

moscicki:
I see it as a two-step process: 
 1) admin recalls the files
 2) he decides that the file on the server should be replaced with one of the recalled copies (after some analysis) 

Ideally I would like to recall this file into some other location (that's why OC-Tag header so the server may decide where to put it) because then you avoid to re-download the recall copies into other clients.

About sharing: yes, you want to reach to the maximum number of available local copies of a file. I don't think it is a problem unless you are sharing with 1000s of users.

dragotin:
@ogoffart could you port the recall patch to the 1.8 branch?

ogoffart:
@moscicki   Do you have any use of the CSYNC_RECALL_FILE_USERNAME environement variable?

ogoffart:
I ported the recall patch to 1.8 in the 'recall' branch.  (I did not test it)

ogoffart:
The code will be integrated in 1.8:  See https://github.com/owncloud/client/pull/3246

moscicki:
Hi,
Sorry for the delay (I was offline). I do not think I really use env variable CSYNC_RECALL_FILE_USERNAME for anything. I saw that in your new branch this was removed. Looks OK.

Issue #2672, opened 2015-01-07
A simple typo


enoch85:
https://github.com/owncloud/mirall/pull/1717/files#diff-f2d6e0a4d2b25efe708acce987058200R467

![1 years](https://cloud.githubusercontent.com/assets/4511254/5328522/b5e52644-7d82-11e4-8e98-384e60ed9459.png)

danimo:
Yes, that needs to use the plural translation scheme: http://doc.qt.digia.com/qq/qq19-plurals.html

PR #3097, opened 2015-04-10
make the remaining time string translatable and support plural forms

blizzz:
To fix https://github.com/owncloud/client/issues/2672 just need translations

@danimo is that how to solve it?

(It builds and shows the time string still properly in English)


danimo:
No, you need to use %n, see http://doc.qt.digia.com/qq/qq19-plurals.html.

blizzz:
Oh, missed this detail :-/ ok, that seems to make it slightly more complicated. Will have a look, ty,

blizzz:
@danimo next attempt. Is this better?

ogoffart:
Unfortunately this won't work.  the string extractors need that the string is directly in the `tr()` call.  This indirection breaks the translation, i'll try to come with a fix

blizzz:
@ogoffart yupp, i forgot to mention it here. I will think of something else.

PR #3126, opened 2015-04-16
Use identical User Agents for csync and the Qt parts

danimo:
Also, use the mirall version for ocsync.
    
Currently, the csync engine part and the Qt part have different UAs, and this makes debugging (i.e. reading access logs) difficult. On top, we haven't increased the ocsync version number in ages. So
as a consequence, I think it would be the best to have ocsync and the rest share the same version number, and make them identify with the same user agent.
    
To ease debugging for our side, we'll still append "(csyncoC)" for calls made by csync.


danimo:
@LukasReschke @owncloud/desktop-developers

LukasReschke:
Yes please. This is really a debugging nightmare, experienced this in a huge EE setup. Correlating this without tools, especially if you don't get the full logs, is impossible :+1:

LukasReschke:
(sure we can tell them that this is *maybe* an outdated client but hey, we can't tell you for real :see_no_evil:)

PR #3134, opened 2015-04-20
Themed user agent string based on Theme::appName()

scolebrook:
Simple change to the user agent string to replace mirall with whatever Theme:appName() returns. Without a theme this will return ownCloud thanks to owncloudtheme.cpp. With a theme it will return the value of APPLICATION_SHORTNAME in OEM.cmake.


danimo:
Obsoleted by ec83295b99d67e28454ca303b922cd4514448793, which will be in 1.8.2

PR #3139, opened 2015-04-21
add note to restart windows after update

carlaschroder:
Re https://github.com/owncloud/enterprise/issues/600, context menu not visible without Windows restart


pierrejochem:
A logoff is enough! No complete reboot required!

PR #3183, opened 2015-05-04
Windows Explorer: fix issues when syncing whole drive.

christian-burger:
When the local sync target is just a drive letter (e.g. "X:\"), neither
the display of the sync status via file icon overlay, nor the creation of a
share link works. In the latter case no pop-up comes up and no server
request is done.

[QDir::cleanPath() usually removes trailing slashes, but not if the path to
be cleaned is just "X:\" and the OS it is running on is Windows](http://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/io/qdir.cpp?id=9ae7e33f28f972ec566d62ab110e173ebe933520#n2167). In that case the trailing slash is kept.

Addendum: I tried to keep number of changes as small as possible. But the change to `Folder::path()` was necessary, so that the local sync target `X:\` would not end up as `X:\/`. [Which would result in the construction of a wrong remote path for a file](https://github.com/owncloud/client/blob/ae17f58b80208df0e6318dbabbfabd7ae842332e/src/gui/socketapi.cpp#L436) when trying to retrieve the share link for the file.

The patch was developed and tested on Windows with the version behind tag `v1.8.0`. I just rebased it to `master`. I encountered no conflicts during rebase, but did no tests since compilation of `master` aborts with `client\src\gui\accountsettings.cpp:625: Fehler: undefined reference to 'OCC::ProgressInfo::Progress::completed() const'` and 24 other "undefined reference" errors, all related to `OCC::ProgressInfo`.

### DIFF
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -180,10 +180,20 @@ QString Folder::alias() const
 QString Folder::path() const
 {
     QString p(_path);
-    if( ! p.endsWith(QLatin1Char('/')) ) {
-        p.append(QLatin1Char('/'));
+    if( ! p.endsWith(QDir::separator()) ) {
+        p.append(QDir::separator());
     }
     return p;
+}
+
+QString Folder::cleanPath(QString path)
+{
+    QString cleanedPath = QDir::cleanPath(path);
+
+    if(cleanedPath.length() == 3 && cleanedPath.endsWith(":/"))
+        return cleanedPath.left(2);
+
+    return cleanedPath;
 }
 
 bool Folder::isBusy() const
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -186,12 +186,12 @@ QString Folder::path() const
     return p;
 }
 
-QString Folder::cleanPath(QString path)
+QString Folder::cleanPath()
 {
-    QString cleanedPath = QDir::cleanPath(path);
+    QString cleanedPath = QDir::cleanPath(_path);
 
     if(cleanedPath.length() == 3 && cleanedPath.endsWith(":/"))
-        return cleanedPath.left(2);
+        cleanedPath.remove(2,1);
 
     return cleanedPath;
 }
--- a/src/gui/socketapi.cpp
+++ b/src/gui/socketapi.cpp
@@ -402,7 +402,7 @@ void SocketApi::command_RETRIEVE_FILE_STATUS(const QString& argument, QLocalSock
         DEBUG << "folder offline or not watched:" << argument;
         statusString = QLatin1String("NOP");
     } else {
-        const QString file = QDir::cleanPath(argument).mid(Folder::cleanPath(syncFolder->path()).length()+1);
+        const QString file = QDir::cleanPath(argument).mid(syncFolder->cleanPath().length()+1);
         SyncFileStatus fileStatus = this->fileStatus(syncFolder, file, _excludes);
 
         statusString = fileStatus.toSocketAPIString();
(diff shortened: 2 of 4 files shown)


src/gui/socketapi.cpp
```diff
@@ -402,7 +402,7 @@ void SocketApi::command_RETRIEVE_FILE_STATUS(const QString& argument, QLocalSock
         DEBUG << "folder offline or not watched:" << argument;
         statusString = QLatin1String("NOP");
     } else {
-        const QString file = QDir::cleanPath(argument).mid(QDir::cleanPath(syncFolder->path()).length()+1);
+        const QString file = QDir::cleanPath(argument).mid(Folder::cleanPath(syncFolder->path()).length()+1);
```
jturcotte: Could you make it non-static, syncFolder->cleanPath() instead of Folder::cleanPath(folder->path())?

christian-burger:
On Monday 04 May 2015 08:09:46 Jocelyn Turcotte wrote:
> Could you make it non-static, syncFolder->cleanPath() instead of
> Folder::cleanPath(folder->path())?

Stupid me, of course I can do that. And while I was at it, I saved a tiny bit of memory in `Folder::cleanPath()` by editing the QString `cleanedPath` with `.remove()` instead of creating another QString instance with `.left()`.

On top of that, I thought about using preprocessor macros like Q_OS_WIN to enable the code only when needed. But, then again: Trying to keep the patch simple.

I would prefer to squash the two commits into one, before you merge, to not clutter the commit history. If that's okay with you?


On the risk of getting a bit off-topic here, but I wanted to implement a test for my modification and I was not able to get the tests running. On Windows as well as Linux the first of many similar error messages was:

```
CMake Error at owncloud_add_test.cmake:14 (qt_wrap_cpp):
  qt_wrap_cpp called with incorrect number of arguments
Call Stack (most recent call first):
  CMakeLists.txt:7 (owncloud_add_test)
```

The last commit from last year changing that line owncloud/client@45d1567057618736e018a15e27b4930c29f09829 does not mention anything on why the number of arguments was reduced and I could find any documentation on `qt_wrap_cpp` with only one argument with a quick search.

It seems to work again if I revert that change, but then another error pops up. It looks to me as if you abandoned this unit test system for now? But then again, people recently committed to the `/test` directory. So the test suite seems to work for those people; just that I cannot find the instructions on how to make it work.

jturcotte:
Thanks for the updates. It's possible that tests aren't building out of the box on Windows, but thanks for trying anyhow :-)

jturcotte:
Hi, so I just realised that you guys need to sign the CLA before we can accept your patches.

The info is there: https://owncloud.org/contribute/agreement/
If you can't sign it we'll unfortunately have to revert the merge request.

christian-burger:
On Thursday 07 May 2015 01:47:58 Jocelyn Turcotte wrote:
> Hi, so I just realised that you guys need to sign the CLA before we can
> accept your patches.

Well, as long as I don't have to sell my soul, I will take care of it some 
time next week. If it's the default dual licensing stuff, then I don't think 
that there is any problem. I will update you as soon as it's done, so you can 
confirm it on your end.

christian-burger:
Took me a bit longer than expected, but I signed, scanned and submitted it via e-mail, today.

jturcotte:
Cool, thank you!

src/gui/folder.cpp
```diff
@@ -180,12 +180,22 @@ QString Folder::alias() const
 QString Folder::path() const
 {
     QString p(_path);
-    if( ! p.endsWith(QLatin1Char('/')) ) {
-        p.append(QLatin1Char('/'));
+    if( ! p.endsWith(QDir::separator()) ) {
+        p.append(QDir::separator());
     }
```
dragotin: I think we can not accept this change. It is important that we always use '/' as path delimiter. Only if the path is displayed to the user, we convert it to the platform specific delimiters.

Can you rewrite this patch without this change please?

PR #3225, opened 2015-05-12
Use predefined function macro from C99.

flocke:
Fixes compilation with gcc 5.1 which uses --std=gnu11 as default.


danimo:
We already have fixed that in the 1.8 branch. I've now done an upmerge to master via the 1.9 branch.

Warning: At this point Do not work on master unless you want to contribute new features.

flocke:
Oh, thanks ... didn't notice the existing patch.

danimo:
Thanks anyway!

PR #3232, opened 2015-05-13
[Unix] Install architecture-independent files to DATADIR.

flocke:
Change the installation target for the translations and the .desktop file to the architecture-independent DATADIR.

### DIFF
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -204,7 +204,7 @@ if(NOT BUILD_OWNCLOUD_OSX_BUNDLE)
         endforeach( _file )
     endif(NOT WIN32)
 
-    install(FILES ${client_I18N} DESTINATION share/${APPLICATION_EXECUTABLE}/i18n)
+    install(FILES ${client_I18N} DESTINATION ${DATADIR}/${APPLICATION_EXECUTABLE}/i18n)
 
     # we may not add MACOSX_BUNDLE here, if not building one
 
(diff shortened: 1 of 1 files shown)


danimo:
Looks good. But out of curiousity: What else can `$DATADIR` be?

/cc @jnweiger

jnweiger:
@danimo in RPM context %{_datadir} is /usr/share. It is defined like this
<pre>
# ---- configure macros.
#       Macro(s) slavishly copied from autoconf's config.status.
#
%_prefix                /usr
%_exec_prefix           %{_prefix}
%_bindir                %{_exec_prefix}/bin
%_sbindir               %{_exec_prefix}/sbin
%_libexecdir            %{_exec_prefix}/libexec
%_datadir               %{_prefix}/share
</pre>


In DEB context there is probably a similar variable with a similar value. I don't use that at all in the packaging scripts. I assume that this $DATADIR here is defined via autoconfig or some pkgconfig. 

@flocke  It was a relative path starting with share/... now $DATADIR probably expands to an absolute path starting with /usr/share/... -- correct?
Is $DATADIR correctly prefixed in case of a build root environment?

Anyway, I somehow miss what we are trying to fix here.

flocke:
@jnweiger DATADIR gets set to DATA_INSTALL_DIR from GNUInstallDirs.cmake, as far as this value isn't specified during the call to cmake it is still the same relative path (share/).

My only goal with this was to give a possibility to change the location of the architecture-independent files during make. This is needed for the distribution I use (Exherbo Linux), which just introduced cross-compilation support to the package manager. All architecture-dependent files get installed to /usr/<arch> (e.g. /usr/x86_64-pc-linux-gnu) while all architecture-independent files are installed to /usr.
I just needed a way to specify the location for the architecture-independent files manually.

Again: as long as no -DDATA_INSTALL_DIR=... parameter is passed to cmake this should not change a thing.

jnweiger:
@flocke thanks for the explanation. That is a good improvement!

The cmake call looks like this:

    cmake .. -DWITH_DOC=TRUE '-DCMAKE_C_FLAGS:STRING=-fmessage-length=0 -grecord-gcc-switches -fstack-protector -O2 -Wall -D_FORTIFY_SOURCE=2 -funwind-tables -fasynchronous-unwind-tables -g' '-DCMAKE_CXX_FLAGS:STRING=-fmessage-length=0 -grecord-gcc-switches -fstack-protector -O2 -Wall -D_FORTIFY_SOURCE=2 -funwind-tables -fasynchronous-unwind-tables -g' -DCMAKE_SKIP_RPATH=OFF -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_INSTALL_PREFIX=/usr -DLIB_SUFFIX=64 -DSYSCONF_INSTALL_DIR=/etc


Thumbs up :+1:

flocke:
Any news on merging this patch?

PR #3246, opened 2015-05-19
Recall

ogoffart:


dragotin:
Nitpicking: 
- Issues with paranthesis in kuba's code: We prefer to have always paranthesis and the opening one on the same line.
- functions parseRecallFile() and makeRecallFileName() and maybe also downloadFinished() are nice candidates for a unit test.
- Adding more static functions to the propagate*.cpp files makes them more and more big can spaghetti codish. Does it make sense to move that to a little service class?

@ogoffart what do you think?

ogoffart:
> Adding more static functions to the propagate*.cpp files makes them more and more big can spaghetti codish. Does it make sense to move little service class?

I think it is not big enough to be moved to its own file.  Do you think it deserve its own file?

dragotin:
No, fine with me.

PR #3247, opened 2015-05-19
Checksum 1.8

dragotin:
Add checksums to HTTP transmissions.


src/libsync/propagateupload.cpp
```diff
@@ -190,22 +194,53 @@ bool PollJob::finished()
...
+    // change during the checksum calculation
+    _item._modtime = FileSystem::getModTime(filePath);
+
+    _stopWatch.start();
+
+    // do whatever is needed to add a checksum to the http upload request.
+    // in any case, the validator will emit signal startUpload to let the flow
+    // continue in slotStartUpload here.
+    _validator = new TransmissionChecksumValidator(filePath);
+    connect(_validator, SIGNAL(validated()), this, SLOT(slotStartUpload()));
```
jturcotte: _validator is neither initialized to null, nor set to null after being deleted. This could end up causing a crash as the code evolve.

You also don't seem to need to keep the pointer, you could set the PropagateUploadFileQNAM as it's QObject parent (just to for avoid leaking in edge cases) and connect its deleteLater() slot to validated() as well.

src/libsync/transmissionchecksumvalidator.cpp
```diff
@@ -0,0 +1,149 @@
...
+    // no validation can happen. Just continue.
+    if( checksumHeader.isEmpty() ) {
+        emit validated();
+        return;
+    }
+
+    int indx = checksumHeader.indexOf(':');
+    if( indx < 0 ) {
+        qDebug() << "Checksum header malformed:" << checksumHeader;
+        emit validationFailed(tr("The checksum header is malformed.")); // show must go on - even not validated.
```
jturcotte: It seems like this will cause an infinite sync trigger, since validationFailed is interpreted as "the file was corrupted on its way by the network". Is this what we want? Or are we just assuming that it will never happen?

src/libsync/propagatorjobs.h
```diff
@@ -21,6 +21,22 @@
...
+ * They are here for being shared between Upload- and Download Job
+ */
+
+// the header itself
+static const char checkSumHeaderC[] = "OC-Checksum";
+// ...and it's values
+static const char checkSumMD5C[] = "MD5";
+static const char checkSumSHA1C[] = "SHA1";
+static const char checkSumAdlerC[] = "Adler32";
+static const char checkSumAdlerUpperC[] = "ADLER32";
```
danimo: Is there a reason to not treat the names case-insensitively and get rid of the duplication here?

cmake/modules/QtVersionAbstraction.cmake
```diff
@@ -111,7 +112,7 @@ endif( Qt5Core_FOUND )
 if(NOT Qt5Core_FOUND)
     message(STATUS "Could not find Qt5, searching for Qt4 instead...")
 
-    set(NEEDED_QT4_COMPONENTS "QtCore" "QtXml" "QtNetwork" "QtGui" "QtWebkit")
+    set(NEEDED_QT4_COMPONENTS "QtCore" "QtXml" "QtNetwork" "QtGui" "QtWebkit" "QtConcurrent")
```
ogoffart: I don't think this change is neede. QtConcurrent was part of QtCore in Qt4

src/libsync/filesystem.cpp
```diff
@@ -18,6 +19,13 @@
 #include <QFileInfo>
 #include <QCoreApplication>
 #include <QDebug>
+#include <QCryptographicHash>
+#include <QFuture>
+#include <qtconcurrentrun.h>
```
ogoffart: I think you are not using any of those 2 headers anymore in this file

src/libsync/filesystem.cpp
```diff
@@ -394,4 +402,58 @@ QString FileSystem::fileSystemForPath(const QString & path)
...
+{
+    QByteArray arr;
+
+    QCryptographicHash crypto( QCryptographicHash::Md5 );
+
+    QFile file(filename);
+    if (file.open(QIODevice::ReadOnly)) {
+        QByteArray data;
+        while (!file.atEnd()) {
+            data = file.read(1024*1024*10);
```
ogoffart: Is is much more efficiant to call the  read overload that takes a char* and put data in a buffer.  Because this re-allocates a new buffer all the time.

src/libsync/filesystem.cpp
```diff
@@ -394,4 +402,58 @@ QString FileSystem::fileSystemForPath(const QString & path)
...
+            adler = adler32(adler, (const Bytef*) data.data(), data.size());
+        }
+    }
+
+    return QString::number( adler, 16 ).toUtf8();
```
ogoffart: There is also QByteArray::number()

src/libsync/owncloudpropagator.cpp
```diff
@@ -259,6 +259,21 @@ void OwncloudPropagator::start(const SyncFileItemVector& items)
...
+    const QString checksumType = cfg.transmissionChecksum().toUpper();
+
+    /* if the checksum type is empty, it is not send. No error */
+    if( !checksumType.isEmpty() ) {
+        if( checksumType == checkSumAdlerUpperC ||
+                checksumType == checkSumMD5C    ||
+                checksumType == checkSumSHA1C ) {
+            qDebug() << "Client sends and expects transmission checksum type" << checksumType;
+        } else {
+            qDebug() << "WARN: Unknown transmission checksum type from config" << checksumType;
```
ogoffart: You can use qWarning for warnings.

src/libsync/propagatedownload.cpp
```diff
@@ -484,7 +486,22 @@ void PropagateDownloadFileQNAM::slotGetFinished()
         return;
     }
 
-    downloadFinished();
```
ogoffart: You could call downloadFinished in any case if the header did not contains the checksum

src/libsync/propagatedownload.cpp
```diff
@@ -484,7 +486,22 @@ void PropagateDownloadFileQNAM::slotGetFinished()
...
+    // do whatever is needed to add a checksum to the http upload request.
```
ogoffart: "http uload request" ?   I think the comment is wrong here

src/libsync/propagatedownload.h
```diff
@@ -101,19 +103,24 @@ private slots:
...
+
 private slots:
     void slotGetFinished();
     void abort() Q_DECL_OVERRIDE;
     void downloadFinished();
     void slotDownloadProgress(qint64,qint64);
+    void slotChecksumFail( const QString& errMsg );
+
+private:
+    // Utility::StopWatch _stopWatch;
```
ogoffart: you could remove this comment

src/libsync/propagatedownload.cpp
```diff
@@ -484,7 +486,22 @@ void PropagateDownloadFileQNAM::slotGetFinished()
...
+    connect(_validator, SIGNAL(validated()), this, SLOT(downloadFinished()));
+    connect(_validator, SIGNAL(validationFailed(QString)), this, SLOT(slotChecksumFail(QString)));
+    _validator->downloadValidation(job->reply()->rawHeader(checkSumHeaderC));
+
+}
+
+void PropagateDownloadFileQNAM::slotChecksumFail( const QString& errMsg )
+{
+    _validator->deleteLater();
+    _tmpFile.remove();
```
ogoffart: What if this fails?   I have seen cases in which mirall still has the _tmpFile open and it would not be removed.  Was the file properly closed?

src/libsync/propagateupload.cpp
```diff
@@ -21,6 +22,9 @@
 #include "utility.h"
 #include "filesystem.h"
 #include "propagatorjobs.h"
+#include "transmissionchecksumvalidator.h"
+#include "configfile.h"
```
ogoffart: Oh no :-( a dependency in the sync engine from the config file.

Maybe there should be a syncengine->setChecsumAlgo instead.

src/libsync/propagateupload.cpp
```diff
@@ -190,22 +194,53 @@ bool PollJob::finished()
...
+
     const QString fullFilePath(_propagator->getFilePath(_item._file));
 
     if (!FileSystem::fileExists(fullFilePath)) {
         done(SyncFileItem::SoftError, tr("File Removed"));
         return;
     }
+    _stopWatch.addLapTime(QLatin1String("Checksum"));
+
+    // Update the mtime and size, it might have changed since discovery.
```
guruz: dup comment with some lines down

src/libsync/propagatedownload.cpp
```diff
@@ -484,7 +486,22 @@ void PropagateDownloadFileQNAM::slotGetFinished()
         return;
     }
 
-    downloadFinished();
+    // do whatever is needed to add a checksum to the http upload request.
+    // in any case, the validator will emit signal startUpload to let the flow
+    // continue in slotStartUpload here.
+    _validator = new TransmissionChecksumValidator( _tmpFile.fileName() );
```
ogoffart: You should either use a QScopedPoointer for it, or set a parent. Otherwise it might leak if the sync engine is destroyed during the sync.

src/libsync/propagateupload.cpp
```diff
@@ -190,22 +194,53 @@ bool PollJob::finished()
...
+    time_t prevModtime = _item._modtime; // the value was set in PropagateUploadFileQNAM::start()
+    // but a potential checksum calculation could have taken some time during which the file could
+    // have been changed again, so better check again here.
 
     // Update the mtime and size, it might have changed since discovery.
     _item._modtime = FileSystem::getModTime(fullFilePath);
+    if( prevModtime != _item._modtime ) {
+        _propagator->_anotherSyncNeeded = true;
+        done(SyncFileItem::SoftError, tr("Local file changed while calculating the checksum."));
```
ogoffart: "calculating the checksum" is an implementation details.  For an error which is likely to happen i think we should not let the user that.  Just maybe "File changed during sync" would be a better error message.

src/libsync/propagatorjobs.h
```diff
@@ -21,6 +21,22 @@
...
+
+/**
+ * Declaration of the other propation jobs
```
ogoffart: Like this, this comments seems to apply to the PropagateLocalRemove only

src/libsync/transmissionchecksumvalidator.cpp
```diff
@@ -0,0 +1,149 @@
...
+}
+
+void TransmissionChecksumValidator::slotUploadChecksumCalculated( )
+{
+    QByteArray checksum = _watcher.future().result();
+
+    if( !checksum.isEmpty() ) {
+        _item->_checksum.append(checksum);
+    } else {
+        _item->_checksum.clear();
```
ogoffart: Should that not be an error?   If the file could not be open to read the checksum or anything?  (because this is also likely case of causing a regression)

src/libsync/transmissionchecksumvalidator.cpp
```diff
@@ -0,0 +1,149 @@
...
+}
+
+void TransmissionChecksumValidator::slotDownloadChecksumCalculated()
+{
+    const QByteArray hash = _watcher.future().result();
+
+    if( hash != _expectedHash ) {
+        emit validationFailed(tr("The downloaded file does not match the checksum, it will be resumed."));
+    } else {
+        qDebug() << "Checksum checked and matching: " << _expectedHash;
```
ogoffart: isn't this a bit of noise?

cmake/modules/QtVersionAbstraction.cmake
dragotin: Removed in 34593cccb6f8385a5075433ae6dd185076c89511

src/libsync/filesystem.cpp
dragotin: done in 64ce0cd7a212fcda54a80e32c3d76cfd3b8db548

src/libsync/filesystem.cpp
dragotin: also 64ce0cd7a212fcda54a80e32c3d76cfd3b8db548

src/libsync/filesystem.cpp
danimo: forgot to push?

src/libsync/propagatedownload.cpp
dragotin: downloadFinished() is bound to the validator-signal validated(). That is called in either case, if the checksum is not there at all, or if it is correct. Both are correct states.

src/libsync/propagatedownload.cpp
dragotin: This is handled as child of the jobs now, see 5b5a636cc172d3406a06be2da2533dd5df9edc73

src/libsync/propagatedownload.cpp
dragotin: Yes, the _tmpFile is closed a couple of lines above, unconditionally. Should be good.

src/libsync/propagateupload.cpp
dragotin: Actually configfile.h can be removed here, but it is (still) used in the transmissionchecksumvalidator. Which is ... not much better I suppose?

src/libsync/propagatorjobs.h
dragotin: definesive programming ;-)

PR #3281, opened 2015-05-26
Confirm feature (sync engine part only)

ogoffart:
The idea is that the UI would set SyncEngine->setNewSharedFolderSizeLimit and listen to the newSharedFolder signal.  It would then add the new folder in the blacklist, and the unconfirmed list.  Then from the UI we can move the unconfirmed item to the white list when the selective sync dialog is closed.

### DIFF
--- a/src/cmd/cmd.cpp
+++ b/src/cmd/cmd.cpp
@@ -443,8 +443,9 @@ restart_sync:
 
     Cmd cmd;
     SyncJournalDb db(options.source_dir);
-    if (!selectiveSyncList.empty())
+    if (!selectiveSyncList.empty()) {
         selectiveSyncFixup(&db, selectiveSyncList);
+    }
 
     SyncEngine engine(account, _csync_ctx, options.source_dir, QUrl(options.target_url).path(), folder, &db);
     QObject::connect(&engine, SIGNAL(finished()), &app, SLOT(quit()));
--- a/src/cmd/cmd.cpp
+++ b/src/cmd/cmd.cpp
@@ -251,38 +251,14 @@ void selectiveSyncFixup(OCC::SyncJournalDb *journal, const QStringList &newList)
         return;
     }
 
-    SqlQuery select("SELECT path FROM last_selective_sync", db);
-    QSet<QString> oldBlackListSet;
-    if (select.exec()) {
-        while(select.next()) {
-            oldBlackListSet.insert(select.stringValue(0));
-        }
-    }
-
-
+    auto oldBlackListSet = journal->getSelectiveSyncList(SyncJournalDb::SelectiveSyncBlackList).toSet();
     auto blackListSet = newList.toSet();
     auto changes = (oldBlackListSet - blackListSet) + (blackListSet - oldBlackListSet);
     foreach(const auto &it, changes) {
         journal->avoidReadFromDbOnNextSync(it);
     }
 
-    SqlQuery drop("DROP TABLE last_selective_sync", db);
-    drop.exec();
-
-    if (!newList.isEmpty()) {
-        SqlQuery createQuery(db);
-        createQuery.prepare("CREATE TABLE IF NOT EXISTS last_selective_sync(path VARCHAR(4096));");
-        createQuery.exec();
-
-        SqlQuery insertQuery(db);
-        insertQuery.prepare("INSERT INTO last_selective_sync VALUES (?1);");
-
-        foreach(const auto &s, newList) {
-            insertQuery.reset();
-            insertQuery.bindValue(1, s);
-            insertQuery.exec();
-        }
-    }
+    journal->setSelectiveSyncList(SyncJournalDb::SelectiveSyncBlackList, newList);
 }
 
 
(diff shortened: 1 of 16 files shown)


ogoffart:
CC/ @ckamm @dragotin

src/cmd/cmd.cpp
```diff
@@ -467,14 +443,13 @@ int main(int argc, char **argv) {
 
     Cmd cmd;
     SyncJournalDb db(options.source_dir);
-    selectiveSyncFixup(&db, selectiveSyncList);
+    if (!selectiveSyncList.empty())
+        selectiveSyncFixup(&db, selectiveSyncList);
```
ckamm: Doesn't this change the behavior from 'no list means an empty list' to 'no list means: keep the previous list'?

src/libsync/discoveryphase.cpp
```diff
@@ -22,6 +22,29 @@
 
 namespace OCC {
 
+
+/* Given a sorted list of paths ending with '/', return weather or not the given path is within one of the paths of the list*/
```
ckamm: minor: whether

src/libsync/discoveryphase.cpp
```diff
@@ -29,36 +52,59 @@ bool DiscoveryJob::isInSelectiveSyncBlackList(const QString& path) const
...
+    qint64 result = -1;
+
+    {
+        QMutexLocker locker(&_vioMutex);
+        emit doGetSizeSignal(path, &result);
+        _vioWaitCondition.wait(&_vioMutex);
+    }
+
+    auto limit = _newSharedFolderSizeLimit;
+    if (true || result > limit) {
```
ckamm: remove 'true'?

src/libsync/discoveryphase.cpp
```diff
@@ -29,36 +52,59 @@ bool DiscoveryJob::isInSelectiveSyncBlackList(const QString& path) const
...
+        // we tell the UI there is a new folder
+        emit newSharedFolder(path);
         return true;
+    } else {
+        // it is not too big, but it in the white list and do not block
```
ckamm: it's *not* in the white list, I think

src/libsync/syncjournaldb.cpp
```diff
@@ -1245,6 +1259,55 @@ void SyncJournalDb::setPollInfo(const SyncJournalDb::PollInfo& info)
...
+    }
+
+    //first, delete all entries of this type
+    SqlQuery delQuery("DELETE FROM selectivesync WHERE type == ?1", _db);
+    delQuery.bindValue(1, int(type));
+    if( !delQuery.exec() ) {
+        qWarning() << "SQL error when deleting selective sync list" << list << delQuery.error();
+    }
+
+    SqlQuery insQuery("INSERT INTO selectivesync VALUES (?1, ?2)" , _db);
```
ckamm: minor: prepare statement beforehand, like the others?

src/libsync/syncjournaldb.h
```diff
@@ -91,6 +91,24 @@ class OWNCLOUDSYNC_EXPORT SyncJournalDb : public QObject
...
+         * For the sync engine, those folders are considered as if they were not there, so the local
+         * folders will be deleted */
+        SelectiveSyncBlackList = 1,
+        /** When a shared flder has a size bigger than a configured size, it is by default not sync'ed
+         * Unless it is in the white list, in which case the folder is sync'ed and all its children.
+         * If a folder is both on the black and the white list, the black list wins */
+        SelectiveSyncWhiteList = 2,
+        /** List of big sync folder that have not been confirmed by the user yet and that the UI
+         * should notify about */
+        SelectiveSyncUndecidedList = 3
```
ckamm: Okay, this will be filled by the slot that connects the the newSharedFolder signal, then?

src/libsync/discoveryphase.cpp
```diff
@@ -29,36 +52,59 @@ bool DiscoveryJob::isInSelectiveSyncBlackList(const QString& path) const
...
-    --it;
-    Q_ASSERT(it->endsWith(QLatin1Char('/'))); // Folder::setSelectiveSyncBlackList makes sure of that
-    if (pathSlash.startsWith(*it)) {
+
+    // Go in the main thread to do a PROPFIND to know the size of this directory
+    qint64 result = -1;
+
+    {
+        QMutexLocker locker(&_vioMutex);
+        emit doGetSizeSignal(path, &result);
```
ckamm: Should we guard against going here for each discovery? In particular, should we do this if the folder already is in the third 'undecided' list?

src/libsync/discoveryphase.cpp
ogoffart: oops...  leftover from my testing :-)

src/cmd/cmd.cpp
ogoffart: Yes. This change the behaviour and it is intentional.  
I think the behaviour is better like that where the option allow to change the blacklist rather than using it.

src/libsync/discoveryphase.cpp
ogoffart: good point. My idea was that the ui would anyway add it to the blacklist. but we could consider the undecided list as the blacklist.

src/libsync/discoveryphase.cpp
ogoffart: i means s/but/put/    we put it in the whitelist so we are not going to query the size of the child directories. (if the parent is small enough, then all the childs are)

src/libsync/syncjournaldb.cpp
ogoffart: not all statements are prepared. in general only the statements that happens often during the sync are.  But maybe it deserve to be prepared.

src/libsync/syncjournaldb.h
ogoffart: yes

Issue #3283, opened 2015-05-26
[Security][SSL] Self signed certificate seems to be always accepted.


jklmnn:
### Expected behaviour  
When a self signed certificate is used and the warning is shown, clicking Cancel should not build up a connection.

### Actual behaviour  
When the warning is shown, a connection is build up, even if you have clicked Cancel.  

### Steps to reproduce  
1. Run ownCloud desktop client.  
2. Intercept traffic and set up mitm attack.  
3. Wait for certificate warning and click Cancel.
4. See traffic in mitm.

### Server configuration
Operating system: Debian 8.0

Web server: Apache 2.4

Database: Sqlite 3.8

PHP version: 5.6

ownCloud version: 8.0.3

Storage backend: SQLite

### Client configuration
Client version: 1.8.1+dfsg-1

Operating system: Debian Stretch

OS language: German

Installation path of client: /usr/bin/owncloud

LukasReschke:
@jklmnn Thanks for reporting this issue back to us. Generally speaking please always contact security@owncloud.com in case of a potential security bug. For now we will handle this one public though.

cc @danimo Can you take a look? THX.

danimo:
@jklmnn I've just tried your steps to reproduce (although with our package repo, and on Ubuntu 14.04), but I have not been able to get any data transferred to mitmproxy so far. @LukasReschke will give it another try.

danimo:
@jklmnn Are you using bandwidth limiting? Does this link against Qt4 or Qt5?

jklmnn:
I'm not using a bandwith limit, and it's linking against Qt5.

jklmnn:
I examined it further and it does not appear if the connection was mitmed before owncloud was started. when you start the client and cancel the certificate it shows an ssl error (as expected). But when you already have a connection that is getting hijacked then the new certificate will be used, either if you accepted it or not.

LukasReschke:
Couldn't reproduce as well. What I did:

1. Install Debian 8.0 without any GUI
2. Login
3. Install Gnome using `tasksel install gnome-desktop --new-install`
4. Reboot
5. `sudo su`
6. `apt-get install owncloud-client`
7. Start the ownCloud client (`owncloud --logwindow`) and connect against a HTTPS protected host
8. Try to sync files: Works
9. Install Ettercap on another machine: `apt-get install ettercap-graphical
[...]

PR #3301, opened 2015-06-02
Abort the request and reset the QNAM if user does not ACK a new cert.

dragotin:
This is supposed to fix bug #3283

### DIFF
--- a/src/libsync/account.cpp
+++ b/src/libsync/account.cpp
@@ -503,6 +503,11 @@ void Account::slotHandleErrors(QNetworkReply *reply , QList<QSslError> errors)
         reply->ignoreSslErrors();
     } else {
         _treatSslErrorsAsFailure = true;
+        // if during normal operation, a new certificate was MITM'ed, and the user does not
+        // ACK it, the running request must be aborted and the QNAM must be reset, to not
+        // treat the new cert as granted. See bug #3283
+        reply->abort();
+        resetNetworkAccessManager();
         return;
     }
 }


dragotin:
@guruz asked me to verify again if the QNAM reset is really needed. It is. If it is not there, at the next sync the client proceeds without popping up the window again. That happens probably because of the if routine here: https://github.com/owncloud/client/blob/master/src/libsync/account.cpp#L497 which just returns in case of having ```_treatSslErrorsAsFailure``` set to true.

I fail to understand why the cert is accepted in that case and QNAM continues to handle the requests.

Because of that,  my suggestion is to keep the QNAM reset for 1.8.2 which fixes the initial problem. Later on, we should refactor the AbstractNetworkJob to care for the _treatSslErrorsAsFailure properly.

LukasReschke:
Patch confirmed :+1:

@jklmnn You are welcome to test this patch as well. We have built a nightly available from https://software.opensuse.org/download/package?project=isv:ownCloud:community:nightly&package=owncloud-client that includes the patch. If you do this please change back to the regular ownCloud client packages afterwards as we don't recommend using that nightly repository :smile:

jklmnn:
@LukasReschke thanks for that info, I'll test it as soon as I will have time.

ogoffart:
It would be good to understand why this works. 
Or rather why it breaks now,  we should not call ignoreSslErrors if the dialog is not accepted.  Do we do that? Then that's the problem.  We don't?  Then why are the request continuing?

dragotin:
@ogoffart no, we do not call ignoreSslErrors if the dialog is not accepted.

ckamm:
@dragotin I also confirm that the patch fixes the issue.

dragotin:
Ok. I'll merge to 1.8 branch and create an issue that this need further investigation.

PR #3303, opened 2015-06-03
NSIS installer module updates

danimo:


danimo:
@dragotin @jturcotte This is the reference installer. built from above branch. Please test http://download.owncloud.com/desktop/daily/ownCloud-1.8.2.5138-nightly20150603-setup.exe.

jturcotte:
Upgrading with an "Administrator" account type on Win8.1 still works as expected. The UAC popup pops and the installation works as expected, though I couldn't reproduce the original bug with that setup.

PR #3316, opened 2015-06-08
Add travis file for coverity_scan branch

LukasReschke:
All things pushed to the "coverity_scan" branch will now get build by Travis and the resulting binary is pushed to Coverity and analyzed automatically.

Since Coverity has a fair usage policy which allows us up to 8 builds per week with a maximum of 2 builds per day we cannot run this automatedly on every push. Which also doesn't make that much sense.

Currently this can be used by reating a "coverity_scan" branch and force-pushing the new code to analyze. If all went well the Travis build should pass and Coverity should display the new information (give it some time to process the build though). In Coverity the SHA of the processed build will be used as version number as well as descriptive text.
(we can automate this though in the future as well if we want that too)

This will fail at the moment on master since Qt4 builds are failing as per https://ci.owncloud.org/job/client-build-matrix-linux/ - probably something that should get fixed. 1.8 branch worked fine though.

@DeepDiver1975 @danimo

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,4 +34,4 @@ install:
 - curl --form token=$token --form email=lukas@statuscode.ch --form file=@$PWD/client.tgz --form version="$checkout" --form description="$checkout" https://scan.coverity.com/builds?project=owncloud%2Fmirall
 
 # Hack to stop processing
-script: time
+script: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,37 @@
+language: cpp
+branches:
+  only:
+  - coverity_scan 
+
+env:
+  global:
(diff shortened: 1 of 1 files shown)


LukasReschke:
Example build: https://travis-ci.org/LukasReschke/client/builds/65917396

.travis.yml
```diff
@@ -0,0 +1,37 @@
...
+- cd ../
+- mkdir client-build
+- cd client-build
+- cmake -DCMAKE_BUILD_TYPE="Debug" $TRAVIS_BUILD_DIR
+- cov-build --dir cov-int make
+- tar czvf client.tgz cov-int 
+- curl --form token=$token --form email=lukas@statuscode.ch --form file=@$PWD/client.tgz --form version="$checkout" --form description="$checkout" https://scan.coverity.com/builds?project=owncloud%2Fmirall
+
+# Hack to stop processing
+script: time
```
LukasReschke: This is a huge hack since it will otherwise call `make` and so on. Shouldn't matter though since if a step fails before it will fail before that and if it reaches this line all is fine…

LukasReschke:
Qt4 builds again. Let's get this in and test it :smile:

LukasReschke:
Thanks @ogoffart!

PR #3344, opened 2015-06-15
Regular update check

dragotin:
This pull requests enables that the update checker to check regularly for available updates, reading the frequency value from the config file.

@danimo please.

### DIFF
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -178,7 +178,8 @@ Application::Application(int &argc, char **argv) :
     QTimer::singleShot( 0, this, SLOT( slotCheckConnection() ));
 
     // Update checks
-    _updaterTimer.setInterval(1000*120*60); // check every two hours
+    _updaterTimer.setInterval(cfg.updateCheckInterval()); // check every two hours
+
     connect(&_updaterTimer, SIGNAL(timeout()), this, SLOT(slotStartUpdateDetector()));
     QTimer::singleShot( 3000, this, SLOT( slotStartUpdateDetector() ));
 
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -34,6 +34,8 @@
 #include "sharedialog.h"
 
 #include "updater/updater.h"
+#include "updater/ocupdater.h"
+
 #include "creds/abstractcredentials.h"
 
 #include "config.h"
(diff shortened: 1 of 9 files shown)


moscicki:
I know it adds little value to the thread but here it is: :+1:

src/libsync/configfile.cpp
```diff
@@ -381,6 +382,24 @@ quint64 ConfigFile::forceSyncInterval(const QString& connection) const
...
 }
 
+bool ConfigFile::updateCheckInterval( const QString& connection ) const
+{
+    QString con( connection );
+    if( connection.isEmpty() ) con = defaultConnection();
+    QSettings settings(configFile(), QSettings::IniFormat);
+    settings.beginGroup( con );
+
+    quint64 defaultInterval = 1000*120*60;
```
ogoffart: how about 10 hours?  the original bug say once a day.

(It will otherwise create a lot more traffic on the update hour. Like 5 times as much.  Is it not better to spread the updates along the day)

src/gui/application.cpp
```diff
@@ -175,11 +177,11 @@ Application::Application(int &argc, char **argv) :
     // Also check immediatly
     QTimer::singleShot( 0, this, SLOT( slotCheckConnection() ));
 
-    if( cfg.skipUpdateCheck() ) {
```
ogoffart: You are no longer checking for skipUpdateCheck ?

dragotin: the check is now in Application::slotStartUpdateDetector(). Even if we skip the update stack *now*, we might not in 10 hours, thats why we need the timer even if skip now.

src/libsync/configfile.cpp
dragotin: yes, 10 hours is better.

guruz:
Should there be a random() factor added of some 0-300 seconds or so to avoid spikes (e.g. all office workers turning on their machine at the same time? or some power outage situation?)

dragotin:
well... Even if all start at seven in the morning, I think there will *still* be enough randomness, we're talking about one http request...

danimo:
@dragotin ... which triggers a PHP script. and even small scripts are very consuming.

ogoffart:
:+1: 
:

src/libsync/configfile.cpp
```diff
@@ -381,6 +382,24 @@ quint64 ConfigFile::forceSyncInterval(const QString& connection) const
...
+    if( connection.isEmpty() ) con = defaultConnection();
+    QSettings settings(configFile(), QSettings::IniFormat);
+    settings.beginGroup( con );
+
+    quint64 defaultInterval = 1000*60*60*10; // ten hours
+    quint64 interval = settings.value( QLatin1String(updateCheckIntervalC), defaultInterval ).toULongLong();
+
+    quint64 minInterval = 1000*60*5;
+    if( interval < minInterval) {
+        qDebug() << "Update check interval less than fife minutes, setting " << minInterval;
```
danimo: five

src/gui/updater/ocupdater.cpp
```diff
@@ -65,8 +65,16 @@ bool OCUpdater::performUpdate()
...
-    checkForUpdate();
+    int dlState = downloadState();
+
+     if( dlState == Unknown ||
+             dlState == UpToDate         ||
+             /* dlState == DownloadComplete ||  <- are we checking if a previous download was successful already? */
+             dlState == DownloadFailed   ||
+             dlState == DownloadTimedOut ) {
+         // how about  UpdateOnlyAvailableThroughSystem?
+         checkForUpdate();
```
danimo: I'd prefer having this as a fully-enumerated switch statement so newly added members cause a warning/error because you're of course right with adding this safe-guard. This can be nasty.

src/gui/updater/updater.cpp
```diff
@@ -96,6 +96,7 @@ Updater *Updater::create()
 #else
     return new PassiveUpdateNotifier(QUrl(updateBaseUrl));
 #endif
+
```
danimo: Remove?

dragotin:
I tested the updater on all platforms, works well.

danimo:
:+1:

Issue #2951, opened 2015-03-12
Schedule Quota job only when Settings dialog is displayed or Tray menu is clicked?


guruz:
I think @ogoffart had suggested this some days ago.

CC @PVince81 who had been wondering about request amounts.

strugee:
Clarifying question - you mean that the client would only run a sync when Settings is open or the tray menu is clicked?

PVince81:
No, that wouldn't make sense. The sync job would always run as before.

It's just that additionally to that, there is another job which only purpose is to update the quota information. Instead of reading the quota from the server often, it would only read it when the dialog is open, which makes sense if the quota value isn't used for anything else than displaying it.

ogoffart:
https://github.com/owncloud/client/pull/3378

PR #3378, opened 2015-06-26
QuotaInfo: only request the quota when the UI is visible

ogoffart:
CC @dragotin


guruz:
Commits should have link to #2951 (even though it is not exactly the same issue)

dragotin:
:+1:

PR #3380, opened 2015-06-26
Structure developer documentation

danimo:
- rename target "doc-dev"
- group into modules
- move to doc/dev


danimo:
@owncloud/desktop-developers @carlaschroder

danimo:
I've assumed that we want Qt style comments, even tough we used both before. Any preferences?

carlaschroder:
This is over my head, so I have nothing helpful to offer except this represents a lot of work! Good job. Now let's see that good code commenting commence :) Which really does help-- for example, I read the oC source code looking for clues to how things are supposed to behave.

danimo:
@carlaschroder That's exactly the purpose. Thanks! :)

ogoffart:
by Qt style, you mean '/*!'  vs. '/**'  ?  
Personally i prefer the doxygen style '/**' as this is what is used everywhere outside Qt.   But '/*!' is fine too if doxygen supports it.

danimo:
The `/*!` notation is consistent with the stuff in 3rdparty and doxygen doesn't care...

danimo:
csync also uses doxygen style, so what the hell :)

Issue #3241, opened 2015-05-16
"Hard link not supported" with NTFS on Linux


jamfx:
Hi Everybody,
i am on an Xubuntu 15.04 with my OwnCloud 8.0.3 and the client 1.8.1.
MANY (and I really mean a lot) of my Files won't be sycned, becaue of the "Error" - "Hard links are not supported"

These Files are not copied anywhere. They are on an internal NTFS-Formatted Harddrive. 
It looks similar to this issue: https://github.com/owncloud/client/issues/2557
But not under the exact same circumstances. So please have a closer look to the OC-Client or did I do something wrong? Just to make it clear again: I don't have made any copies or hard links to this files by myself. What can I do against this? Like this my OC is unusable for me.
OK I found out (ls -la), that many of my files do have a second hard link. But I can not find out where and how to delete this... 

Thanks JamFX

ogoffart:
Relates to #1143

jamfx:
@ogoffart Are you sure it relates to this bug? Doesn't look quite similar. Although I used the following workaround:
a) stop OC Sync
b) Copy all OC-files to a "backup-Folder" like "OC-Backup" (and I copied really the whole OC because of so many files complaining)
c) Copy all Files back from "backup-Folder" to the original OC-Folder overwriting existing files
d) Start OC Sync again

Now the files got synced and only 1 conflict-File showed up! Lucky me?

ogoffart:
I think NTFS are making hard links internaly, and owncloud will refuse to sync some of the files because of that.

We might need to change our policy regarding links

grawlinson:
Hello, just chiming in to say that I've started to experience this issue.

I dual-boot Ubuntu & Windows 8 with a shared NTFS partition, which contains my ownCloud directory.

If I try and overwrite the affected files with copies, ownCloud will still not sync them.

justlilpete:
Just to say I also am experiencing this issue. Dual booting Ubuntu 15.04 with Windows 8.1, ownCloud directory on a shared NTFS partition. A number of files, seemingly random spread of types (avi, jpg, exe, txt, zip) in a number of folders (and not all files in the folder) fail with this error.

MTRichards:
From Nov 2014:
> "No, we neither plan to sync hard nor softlinks"
https://github.com/owncloud/client/issues/530#issuecomment-64792249

[...]

PR #3388, opened 2015-07-01
csync: Do not ignore hard links anymore

ogoffart:
There is no reason to ignore them. Downloading a file that is hardlinked
will break the link.

Will solve syncing NTFS directories #3241


dragotin:
:+1:

PR #3390, opened 2015-07-01
Enable long file names in windows (longer than 250 bytes)

dragotin:
This patch converts pathes to UNC format if the path length exceeds 250 bytes.

This would fix bug #60

### DIFF
--- a/csync/src/csync_util.c
+++ b/csync/src/csync_util.c
@@ -103,7 +103,7 @@ void csync_memstat_check(void) {
   CSYNC_LOG(CSYNC_LOG_PRIORITY_INFO, "Memory: %dK total size, %dK resident, %dK shared",
                  m.size * 4, m.resident * 4, m.shared * 4);
 }
-
+#if 0
 void csync_win32_set_file_hidden( const char *file, bool h ) {
 #ifdef _WIN32
   const mbchar_t *fileName;
@@ -127,6 +127,7 @@ void csync_win32_set_file_hidden( const char *file, bool h ) {
     (void) file;
 #endif
 }
+#endif
 
 bool (*csync_file_locked_or_open_ext) (const char*) = 0; // filled in by library user
 void set_csync_file_locked_or_open_ext(bool (*f) (const char*));
--- a/csync/src/csync_util.c
+++ b/csync/src/csync_util.c
@@ -103,31 +103,6 @@ void csync_memstat_check(void) {
   CSYNC_LOG(CSYNC_LOG_PRIORITY_INFO, "Memory: %dK total size, %dK resident, %dK shared",
                  m.size * 4, m.resident * 4, m.shared * 4);
 }
-#if 0
-void csync_win32_set_file_hidden( const char *file, bool h ) {
-#ifdef _WIN32
-  const mbchar_t *fileName;
-  DWORD dwAttrs;
-  if( !file ) return;
-
-  fileName = c_utf8_to_locale( file );
-  dwAttrs = GetFileAttributesW(fileName);
-
-  if (dwAttrs != INVALID_FILE_ATTRIBUTES) {
-      if (h && !(dwAttrs & FILE_ATTRIBUTE_HIDDEN)) {
-          SetFileAttributesW(fileName, dwAttrs | FILE_ATTRIBUTE_HIDDEN );
-      } else if (!h && (dwAttrs & FILE_ATTRIBUTE_HIDDEN)) {
-          SetFileAttributesW(fileName, dwAttrs & ~FILE_ATTRIBUTE_HIDDEN );
-      }
-  }
-
-  c_free_locale_string(fileName);
-#else
-    (void) h;
-    (void) file;
-#endif
-}
-#endif
 
 bool (*csync_file_locked_or_open_ext) (const char*) = 0; // filled in by library user
 void set_csync_file_locked_or_open_ext(bool (*f) (const char*));
--- a/csync/src/csync_util.h
+++ b/csync/src/csync_util.h
@@ -30,7 +30,7 @@ const char *csync_instruction_str(enum csync_instructions_e instr);
 
 void csync_memstat_check(void);
 
-void csync_win32_set_file_hidden( const char *file, bool hidden );
+// void csync_win32_set_file_hidden( const char *file, bool hidden );
 
 bool csync_file_locked_or_open( const char *dir, const char *fname);
 #endif /* _CSYNC_UTIL_H */
--- a/csync/src/csync_util.h
+++ b/csync/src/csync_util.h
@@ -30,7 +30,5 @@ const char *csync_instruction_str(enum csync_instructions_e instr);
 
 void csync_memstat_check(void);
 
-// void csync_win32_set_file_hidden( const char *file, bool hidden );
-
 bool csync_file_locked_or_open( const char *dir, const char *fname);
 #endif /* _CSYNC_UTIL_H */
--- a/csync/src/std/c_path.c
+++ b/csync/src/std/c_path.c
@@ -389,3 +389,49 @@ int c_parse_uri(const char *uri,
   return -1;
 }
 
+
+/*
+ * This function takes a path and converts it to a UNC representation of the
+ * string. That means that it prepends a \\?\ and convertes all slashes to
+ * backslashes.
+ *
+ * Note the following:
+ *  - The string must be absolute.
+ *  - it needs to contain a drive character to be a valid UNC
+ *  - A conversion is only done if the path len is larger than 245. Otherwise
+ *    the windows API functions work with the normal "unixoid" representation too.
+ *
+ * Since the function reallocs memory that it can not free itself, the number of
+ * newly allocated bytes are returned in parameter mem_reserved. The calling
+ * function will call free on the result pointer if mem_reserved is > 0.
+ *
+ */
+ const char *c_path_to_UNC(const char *str)
+ {
+     int len = 0;
+     char *longStr = NULL;
+     int i = 4; // index where to start changing "/"=>"\"
+
+     len = strlen(str);
+     longStr = c_malloc(len+5);
+     *longStr = '\0';
+
+     // prepend \\?\ and convert '/' => '\' to support long names
+     if( str[0] == '/' ) {
+         strncpy( longStr, "\\\\?", 4);
+         i=3;
+     } else {
+         strncpy( longStr, "\\\\?\\", 5); // prepend string by this four magic chars.
+     }
+     strncat( longStr, str, len );
+
+     /* replace all occurences of / with the windows native \ */
+     while(longStr[i] != '\0') {
+         if(longStr[i] == '/') {
+             longStr[i] = '\\';
+         }
+         i++;
+     }
+     return longStr;
+ }
+
--- a/csync/src/std/c_path.c
+++ b/csync/src/std/c_path.c
@@ -32,6 +32,7 @@
 #include "c_private.h"
 #include "c_alloc.h"
 #include "c_path.h"
+#include "c_string.h"
 
 /*
  * dirname - parse directory component.
@@ -435,3 +436,18 @@ int c_parse_uri(const char *uri,
      return longStr;
  }
 
+ mbchar_t* c_utf8_path_to_locale(const char *str)
+ {
+     if( str == NULL ) {
+         return NULL;
+     } else {
+ #ifdef _WIN32
+         const char *unc_str = c_path_to_UNC(str);
+         mbchar_t *dst = c_utf8_string_to_locale(unc_str);
+         SAFE_FREE(unc_str);
+         return dst;
+ #else
+         return c_utf8_string_to_locale(str);
+ #endif
+     }
+ }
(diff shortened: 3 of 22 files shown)


csync/src/csync_util.c
```diff
@@ -103,7 +103,7 @@ void csync_memstat_check(void) {
   CSYNC_LOG(CSYNC_LOG_PRIORITY_INFO, "Memory: %dK total size, %dK resident, %dK shared",
                  m.size * 4, m.resident * 4, m.shared * 4);
 }
-
+#if 0
```
ogoffart: please then just remove the dead code

csync/src/csync_util.h
```diff
@@ -30,7 +30,7 @@ const char *csync_instruction_str(enum csync_instructions_e instr);
 
 void csync_memstat_check(void);
 
-void csync_win32_set_file_hidden( const char *file, bool hidden );
+// void csync_win32_set_file_hidden( const char *file, bool hidden );
```
ogoffart: remove

csync/src/std/c_string.c
```diff
@@ -274,13 +274,72 @@ char* c_utf8_from_locale(const mbchar_t *wstr)
...
+ *  - it needs to contain a drive character to be a valid UNC
+ *  - A conversion is only done if the path len is larger than 245. Otherwise
+ *    the windows API functions work with the normal "unixoid" representation too.
+ *
+ * Since the function reallocs memory that it can not free itself, the number of
+ * newly allocated bytes are returned in parameter mem_reserved. The calling
+ * function will call free on the result pointer if mem_reserved is > 0.
+ *
+ */
+ const char *makeLongWinPath(const char *str, int *mem_reserved)
```
ogoffart: I think this function should not be in c_string,  but more like somewhere in vio.  It is not a string related function  (not something you would see in QString or QByteArray)  but a function related to path, so more like something you would find in QDir or QFile


Anyway, the real problem of the function is the allocation behaviour.  it sometimes allocate and sometimes not.   and that is i think a problem.

csync/src/std/c_string.c
```diff
@@ -274,13 +274,72 @@ char* c_utf8_from_locale(const mbchar_t *wstr)
...
+            i++;
+        }
+        return longStr;
+    } else {
+        return str;
+    }
+}
+
 /* Convert a an UTF8 string to multibyte */
 mbchar_t* c_utf8_to_locale(const char *str)
```
ogoffart: this is a function that is documented to just change the charset of a string, not chaging the content of it

csync/src/std/c_string.c
```diff
@@ -274,13 +274,72 @@ char* c_utf8_from_locale(const mbchar_t *wstr)
...
+    int len = 0;
+    char *longStr = NULL;
+
+    if( mem_reserved ) {
+        *mem_reserved = 0;
+    }
+
+    len = strlen(str);
+    // prepend \\?\ and convert '/' => '\' to support long names
+    if( len > 245 ) {  // Only do realloc for long pathes. Shorter pathes are fine.
```
jturcotte: Is there any reason not to always use UNC paths?

csync/tests/encoding_tests/check_encoding.c
```diff
@@ -142,13 +143,46 @@ static void check_to_multibyte(void **state)
...
+
+    (void) state; /* unused */
+
+    assert_string_equal(new_short, path);
+    assert_int_equal(mem_reserved, 0);
+
+    const char *longPath = "D://alonglonglonglong/blonglonglonglong/clonglonglonglong/dlonglonglonglong/"
+            "elonglonglonglong/flonglonglonglong/glonglonglonglong/hlonglonglonglong/ilonglonglonglong/"
+            "jlonglonglonglong/klonglonglonglong/llonglonglonglong/mlonglonglonglong/nlonglonglonglong/"
+            "olonglonglonglong/file.txt";
```
ogoffart: maybe add some unicode character in the test?

csync/src/vio/csync_vio_local.c
```diff
@@ -44,23 +44,60 @@
...
+
+  if( dirname ) {
+      handle->hFind = FindFirstFile(dirname, &(handle->ffd));
+  }
+
+  if (!dirname || handle->hFind == INVALID_HANDLE_VALUE) {
+      SAFE_FREE(handle);
+      return NULL;
+  }
+  // FIXME: Handle the case of no files in a dir.
```
jturcotte: It would be nice if the comment would say both why is this important and why it's acceptable that it's not yet fixed.

csync/src/std/c_string.c
ogoffart: Also the csync coding style is with underscores

csync/src/vio/csync_vio_local.c
```diff
@@ -44,23 +44,60 @@
...
 
   handle = c_malloc(sizeof(dhandle_t));
 
+#ifdef _WIN32
+  // the file wildcard has to be attached
+  int len_name = strlen(name);
+  if( len_name ) {
+      char *h = NULL;
+
+      // alloc an enough large buffer to take the name + '/*' + the closing zero.
```
jturcotte: I really have no idea, but is there a way that we could use Qt instead of implementing this ourselves?

csync/src/std/c_string.c
ogoffart: and it should start with the proper namespace.  So something like csyc_vio_make_long_path

src/libsync/filesystem.cpp
```diff
@@ -235,8 +262,12 @@ bool FileSystem::uncheckedRenameReplace(const QString& originFileName,
 
 #else //Q_OS_WIN
     BOOL ok;
-    ok = MoveFileEx((wchar_t*)originFileName.utf16(),
-                    (wchar_t*)destinationFileName.utf16(),
+    QString orig = longWinPath(originFileName);
+    QString dest = longWinPath(destinationFileName);
+
+    qDebug() << "** MOVE: " << orig;
```
jturcotte: This qDebug seems to lack a bit of context, could you either make it useful for production or remove it?

src/libsync/filesystem.cpp
```diff
@@ -45,6 +46,16 @@ extern "C" {
 
 namespace OCC {
 
+QString FileSystem::longWinPath( const QString& inpath )
+{
+    QString path(inpath);
+
+    path.replace('/', '\\');
+    path.prepend(QLatin1String("\\\\?\\"));
```
jturcotte: So this is the same as makeLongWinPath but in 3 lines of code? Could you possibly not have two implementations somehow?

csync/src/std/c_string.c
guruz: yes, please always use them so that there is only 1 codepath

ckamm: I'd prefer a 
if (len <= 245)
    return str;
here to reduce indent below.

csync/src/std/c_string.c
```diff
@@ -274,13 +274,72 @@ char* c_utf8_from_locale(const mbchar_t *wstr)
...
+        // reserve mem for a new string with the prefix
+        if( mem_reserved ) {
+            *mem_reserved = len + 5;
+        }
+        longStr = c_malloc(len+5);
+        *longStr = '\0';
+
+        if( str[0] == '/' ) {
+            strcpy( longStr, "\\\\?");
+            i=3;
```
ckamm: minor: I'd just get rid of `i` and start from 0 in the while loop below. It doesn't seem worth it.

src/libsync/filesystem.cpp
ckamm: The csync version of this function has a special case for paths that start with / or \. Should this one have it too?

csync/src/vio/csync_vio_local.c
ckamm: How important is it to fix that?

csync/src/vio/csync_vio_local.c
```diff
@@ -44,23 +44,60 @@
...
+      h = c_malloc(len_name+3);
+      strcpy( h, name);
+      strcat(h, "/*");
```
ckamm: consider using strncpy, strncat etc...

csync/src/csync_util.c
```diff
@@ -103,7 +103,7 @@ void csync_memstat_check(void) {
...
 void csync_win32_set_file_hidden( const char *file, bool h ) {
```
ckamm: delete for real

csync/src/vio/csync_vio_local.c
guruz: You mean the whole opendir/readdir stuff?
We anyway need information that is not acessible via Qt so it's better we use the Windows functions (there's also room for optimizations with regards to _stat vs _readdir that should be done at a later point)..

csync/tests/vio_tests/check_vio_ext.c
```diff
@@ -0,0 +1,335 @@
...
+    create_dirs( t1 );
+
+    traverse_dir(state, CSYNC_TEST_DIR);
+
+    assert_string_equal( sv->result,
+                         "<DIR> C:/tmp/csync_test/alibaba"
+                         "<DIR> C:/tmp/csync_test/alibaba/und"
+                         "<DIR> C:/tmp/csync_test/alibaba/und/die"
+                         "<DIR> C:/tmp/csync_test/alibaba/und/die/vierzig"
+                         "<DIR> C:/tmp/csync_test/alibaba/und/die/vierzig/räuber" );
```
guruz: Could you also test special characters with more width? Like Chinese or Cyrillic Characters?

csync/src/vio/csync_vio_local.c
dragotin: It's not an issue, I added an extra test case to the unit test.

dragotin:
I have reworked this according to your feedback. Thanks.

Please take a look again. Next week this will have to be rebased to master.

csync/src/std/c_path.c
```diff
@@ -389,3 +390,64 @@ int c_parse_uri(const char *uri,
...
+ * string. That means that it prepends a \\?\ and convertes all slashes to
+ * backslashes.
+ *
+ * Note the following:
+ *  - The string must be absolute.
+ *  - it needs to contain a drive character to be a valid UNC
+ *  - A conversion is only done if the path len is larger than 245. Otherwise
+ *    the windows API functions work with the normal "unixoid" representation too.
+ *
+ * Since the function reallocs memory that it can not free itself, the number of
```
ogoffart: The documentation seems to be out of sync

ogoffart: it should be:  the pointer returned by this function need to be free-ed with c_free

csync/src/vio/csync_vio_local.c
ogoffart: then the comment can be removed?

csync/src/vio/csync_vio_local.c
```diff
@@ -94,6 +137,41 @@ csync_vio_file_stat_t *csync_vio_local_readdir(csync_vio_handle_t *dhandle) {
...
+  } else {
+      if( FindNextFile(handle->hFind, &(handle->ffd)) == 0 ) {
+          // might be error, check!
+          int dwError = GetLastError();
+          if (dwError != ERROR_NO_MORE_FILES) {
+              errno = EACCES; // FIXME: Is this a good errno?
+              goto err;
+          } else {
+              // Normal case that no more is in the dir.
+              return NULL;
```
ogoffart: leaking file_stat?

csync/src/vio/csync_vio_local.c
```diff
@@ -44,23 +44,57 @@
...
+#ifdef _WIN32
+  // the file wildcard has to be attached
+  int len_name = strlen(name);
+  if( len_name ) {
+      char *h = NULL;
+
+      // alloc an enough large buffer to take the name + '/*' + the closing zero.
+      h = c_malloc(len_name+3);
+      strncpy( h, name, 1+len_name);
+      strncat(h, "/*",3);
```
ogoffart: the 3 should actually be 2,  but it does not matter since it is a raw litteral.
I personaly would disagree with @ckamm here and say that strcpy and strcat were just fine.

csync/src/std/c_path.h
```diff
@@ -107,6 +108,32 @@ typedef struct
...
+/**
+ * @brief c_utf8_path_to_locale converts a unixoid path to the locale aware format
+ *
+ * On windows, it converts to UNC and multibyte.
+ * On Mac, it converts to the correct utf8 using iconv.
+ * On Linux, it returns utf8
+ *
+ * @param str The path to convert
+ *
+ * @return a pointer to the converted string. Caller has to free it.
```
ogoffart: maybe mention it has to free it with c_free_local_string

ogoffart:
looks good. I could only find details.

dragotin:
The code was merged to master, not, as this request here, into 1.8. However, all the feedback was picked up. Closing this.

dragotin:
Test description added to https://github.com/owncloud/client/wiki/Testing-Scenarios-2.0#long-pathnames-in-windows-httpsgithubcomowncloudclientissues60

Issue #3002, opened 2015-03-23
Hide Share link checkbox in share dialog until share information is returned


labkode:
Imagine we have shared this file previously, then we want to see the share link for example.
So when you click on Share with ownCloud, this menu appears
![screen shot 2015-03-23 at 09 45 51](https://cloud.githubusercontent.com/assets/5175912/6776658/d8bddf4a-d141-11e4-997c-ad4b8fa209c3.png)
and seconds later the info appears
![screen shot 2015-03-23 at 09 45 54](https://cloud.githubusercontent.com/assets/5175912/6776685/36152658-d142-11e4-84ec-f300d8f8b28a.png).
The problem is that if you have slow internet connexion, the second screenshot does not appear immediately  and this can trigger the unwanted behaviour of the user clicking on "Share link" and creating two shares by link to the same file.
The easy solution is to show to the user in first place a "Please wait ..." until the call to ocs/v1.php/apps/files_sharing/api/shares?path=/home/esa.png&format=json is finished.

rullzer:
Ah yes you are right!
I had this problem at the airport not to long ago but forgo to make a note.

I'll have a look at it later today. Fix should not be that hard.

guruz:
@rullzer Any news from your side? :)

rullzer:
Finally my Qt setup works again... PR in #3425

rullzer:
Should be fixed in #3425

PR #3425, opened 2015-07-08
Wait until we retrieved shares before showing checkbox

rullzer:
If we have a slow connection we do now want to show the share checkbox until we
have retrieved the shares (so we know the state). Else we might try to enable
sharing on an already shared file.

Fix for #3002


ogoffart:
:+1:

rullzer:
@ckamm you are right :)

In my first version I implemented it with the checkbox hidden... but later figured this was a bit nicer.

dragotin:
:+1:

PR #3426, opened 2015-07-09
Sharedialog show thumbnail of file (if available)

rullzer:
Basically another go at #2805

The last time this code was removed since the endpoint might still change. (CC: @dragotin) however, to my knowledge the mobile clients also use this endpoint. So after a quick chat with @DeepDiver1975 he said to just do it :)

### DIFF
--- a/src/gui/sharedialog.cpp
+++ b/src/gui/sharedialog.cpp
@@ -275,6 +275,12 @@ void ShareDialog::getShares()
     job->addPassStatusCode(404); // don't report error if share doesn't exist yet
     connect(job, SIGNAL(jobFinished(QVariantMap)), this, SLOT(slotSharesFetched(QVariantMap)));
     job->start();
+
+    if (QFileInfo(_localPath).isFile()) {
+        ThumbnailJob *job2 = new ThumbnailJob(_sharePath, _account, this);
+        connect(job2, SIGNAL(jobFinished(int, QByteArray)), SLOT(slotThumbnailFetched(int, QByteArray)));
+        job2->start();
+    }
 }
 
 void ShareDialog::slotSharesFetched(const QVariantMap &reply)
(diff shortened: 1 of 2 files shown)


src/gui/sharedialog.cpp
```diff
@@ -275,6 +275,13 @@ void ShareDialog::getShares()
     job->addPassStatusCode(404); // don't report error if share doesn't exist yet
     connect(job, SIGNAL(jobFinished(QVariantMap)), this, SLOT(slotSharesFetched(QVariantMap)));
     job->start();
+
+    if (QFileInfo(_localPath).isFile()) {
+        QUrl url2 = Account::concatUrlPath(_account->url(), QString("index.php/apps/files/api/v1/thumbnail/150/150/%1").arg(_sharePath));
```
ogoffart: using `Qlatin1String("index.php/apps/files/api/v1/thumbnail/150/150/") + _sharedPath`
Is both more readable (at least for me) and efficient that  `QString("...").arg(...)`

ogoffart:
:+1:

src/gui/sharedialog.cpp
rullzer: Fixed :)

ckamm:
:+1:

Minor: Consider moving the thumbnail URL construction into `ThumbnailJob` and passing the `_sharePath`to it.

rullzer:
@ckamm fixed. And it looks nicer indeed :)

dragotin:
Shouldn't we have all jobs in the networkjobs.cpp ?

rullzer:
@dragotin fine by me. But the OcsShareJob was already in the sharedialog.cpp so it looked like a logical place.

ogoffart:
> Shouldn't we have all jobs in the networkjobs.cpp ?

I don't think so.  I thin it make sens to put the job where it belongs logicaly.

dragotin:
:+1:

Issue #2086, opened 2014-08-21
Don't sync hidden files for Windows and OS X.


guruz:
For a customer for OS X, we're currently implementing that (```UF_HIDDEN``` files for temp files created by MS Word) hidden files (not the ```.``` dot hidden files) .

Find out if we want to do that for Windows too.

CC @MTRichards @karlitschek @dragotin

cdamken:
:+1:

guruz:
There is a pull request here for the Windows part: https://github.com/owncloud/client/pull/3428

CC @dragotin

moscicki:
@MTRichards: it would be really great to understand where we are heading with hidden files (and similar sync issues such as *.tmp files going back and forth the ignored file patterns etc). as @guruz says, on Linux this is an entirely different use-case for hidden files and it would be good to make a clever distinction.

It would be great to be able to summarise the sync behaviour on a sheet of A4 paper, so anyone could understand what to expect from the system. My worry is that as the system evolves and change subtly the semantics we are potentially breaking user's expectations (either explicit or implied by their usage experience).

dragotin:
@moscicki I agree, we should have this one-pager describing the sync rules.

Concerning the hidden files, we always had the strategy to *not* sync them. However, there was a mistake in the implementation that made them sync under some circumstances.

I work on this problem currently. And I think the best solution is to give the user a choice, with a checkbox in the same dialog where we allow the user to edit the ignored files pattern. Default should be to *not* sync hidden files, but if people want, they can switch on.

@moscicki @MTRichards does that match your thoughts?

moscicki:
Is the reason not to sync certain hidden files (especially on Windows) because of data integrity protection (e.g. not to corrupt some temporary hidden cache of applications due to possible conflicts) or is it purely not to sync uselessly files which we think user are not interested to sync anyway? In the first case we cannot really leave it to the user (but I would suggest to have a special handling of this: a specific list of hidden file patterns which match specific applications -- which unfortunately would need to be kept updated somehow), in the second case it is an option indeed.

In the second case my only worry is that users assume that they sync a directory and everything there is "safe" because it is synced to cloud storage. In reality it may be not true because some stuff is not synced.

dragotin:
some information about the extra files word is creating: https://support.microsoft.com/en-us/kb/211632

I am not sure however if these files are really hidden.

moscicki:
[...]

PR #3428, opened 2015-07-09
Hidden file handling

dragotin:
This fixes our handling of hidden files. Now they are ignored completely, not only in the watcher.

Additionally they are listed in the Activity log.

Do we need an option to enable/disable this behaviour?

@ogoffart pls review.


csync/src/csync_update.c
```diff
@@ -163,29 +163,29 @@ static int _csync_detect_update(CSYNC *ctx, const char *file,
...
+      }
+      if (excluded == CSYNC_FILE_SILENTLY_EXCLUDED) {
+          return 1;
+      }
 
-    if (ctx->current_fs) {
-        ctx->current_fs->has_ignored_files = true;
-    }
+      if (ctx->current_fs) {
+          ctx->current_fs->has_ignored_files = true;
```
ogoffart: this should also be done in the hidden case

csync/src/csync.h
```diff
@@ -103,7 +103,8 @@ enum csync_status_codes_e {
     CSYNC_STATUS_INDIVIDUAL_IGNORE_LIST,
     CSYNC_STATUS_INDIVIDUAL_IS_INVALID_CHARS,
     CSYNC_STATUS_INDIVIDUAL_EXCLUDE_LONG_FILENAME,
-    CYSNC_STATUS_FILE_LOCKED_OR_OPEN
+    CYSNC_STATUS_FILE_LOCKED_OR_OPEN,
+    CSYNC_STATUS_INDIVIDUAL_EXCLUDE_HIDDEN
```
guruz: Can we have a code comment what this means?

guruz:
The commit(s) need to mention the issue number https://github.com/owncloud/client/issues/2086

csync/src/vio/csync_vio_local.c
```diff
@@ -384,6 +389,10 @@ int csync_vio_local_stat(const char *uri, csync_vio_file_stat_t *buf) {
       buf->flags |= CSYNC_VIO_FILE_FLAGS_HIDDEN;
   }
 #endif
+  if( strlen(buf->name) > 1 && buf->name[0] == '.' ) {
```
guruz: Is there ever a chance that this gets called with "." as current directory or ".." or so? i guess not?

ogoffart: the strlen(bif->name) test in not usefull.

src/libsync/syncengine.cpp
```diff
@@ -349,7 +349,8 @@ int SyncEngine::treewalkFile( TREE_WALK_FILE *file, bool remote )
         item->_remotePerm = QByteArray(file->remotePerm);
     }
     item->_should_update_etag = item->_should_update_etag || file->should_update_etag;
-
+    item->_hasIgnoredFiles    = file->has_ignored_files > 0 ? true : false;
```
ogoffart: Why ">0" ?
also why  "? true : false"

src/libsync/syncengine.cpp
```diff
@@ -349,7 +349,8 @@ int SyncEngine::treewalkFile( TREE_WALK_FILE *file, bool remote )
...
+    qDebug() << "HAS IGNORED FILES for " << item->_file << item->_hasIgnoredFiles << key;
```
ogoffart: is that not log polution?

ogoffart:
We should only store in the database that has_ignored_file  for the server. (not for the locale)

And did we not said we wanted more comments?

csync/src/vio/csync_vio_local.c
dragotin: Why not? If strlen == 1 then its not a hidden file but the current dir.

src/libsync/syncengine.cpp
dragotin: file->has_ignored_files is an int coming from the db. item->hasIgnoredFiles however is a bool. This conversion is makes it clear how this value is computed.

ogoffart: "file->has_ignored_files > 0 "  is a bool.

And there is also the rules that an int converts to a bool automaticall

csync/src/csync_update.c
```diff
@@ -163,29 +163,24 @@ static int _csync_detect_update(CSYNC *ctx, const char *file,
...
-     https://github.com/owncloud/mirall/issues/2086 */
-  if (fs->flags & CSYNC_VIO_FILE_FLAGS_HIDDEN) {
-      CSYNC_LOG(CSYNC_LOG_PRIORITY_TRACE, "file excluded because it is a hidden file: %s", path);
-      return 0;
-  }
-
   /* Check if file is excluded */
   excluded = csync_excluded(ctx, path,type);
 
   if (excluded != CSYNC_NOT_EXCLUDED) {
```
ckamm: minor: Since this double negation is hard to read and it now has an else branch, maybe negate the condition here?

src/gui/folder.cpp
```diff
@@ -191,6 +191,17 @@ QString Folder::path() const
     return p;
 }
 
+bool Folder::ignoreHiddenFiles()
+{
+    bool re(_definition.ignoreHiddenFiles);
+    return re;
```
ckamm: Why not `return _definition.ignoreHiddenFiles`?

dragotin: This way it can be checked in the debugger if needed. And I expect the compiler to optimize that as needed.

csync/src/csync_update.c
dragotin: fixed with 2cc0d37b3639597c58dc2f2713a13e5bb7012fe1

Issue #2815, opened 2015-02-10
When you enable the option of " Set default expiration date", you can share by link without knowing it


lmaestro:
### Expected behavior
You can share by link using set default expiration date from the admin side.
When you click on share by link the expiration date is enable by default.
When you enforced expiration date, you can't select more days than it is shown by default but you can select less.

### Actual behavior
You can share by link without knowing that there are an expiration date 
And when you enable the option "Set expiry date", an error is shown:
![screen shot 2015-02-10 at 3 24 11 pm](https://cloud.githubusercontent.com/assets/4689430/6128743/e558dc42-b138-11e4-8ef0-e49d04d49d19.png)

### Steps to reproduce
1.  Enable the option  "Set default expiration date"  from the admin side
2. Try to share by link form the desktop 

### Client configuration
Version 1.8.0-nightly20150209(build2001)

Operating system: OS X 10.10.1

dragotin:
We can not fix this as long as the proper API is not in place, see https://github.com/owncloud/core/pull/13964
Moving to future milestone.

guruz:
Looks like the server side API has been merged..

FYI @rullzer

rullzer:
Yes it is. I'll make sure to work on the capabilities stuff tomorrow... should improve the UX!

Dianafg76:
I try this Issue and is working fine, so I close the issue
Desktop v ownCloud-2.0.0.2555-nightly20150722.pkg

PR #3439, opened 2015-07-10
Use server capabilities

rullzer:
Replaces #2943
Should fix: #2815

By making use of the capabilities provided by the server we can provide a better UX.

1. Do not allow a user to remove expire date/password if the admin enforces it
2. Require a password from the user if this is enforced.

### DIFF
--- a/src/libsync/account.h
+++ b/src/libsync/account.h
@@ -26,6 +26,7 @@
 #include <QSharedPointer>
 #include "utility.h"
 #include <memory>
+#include "capabilities.h"
 
 class QSettings;
 class QNetworkReply;
@@ -139,7 +140,7 @@ public:
     void setCertificate(const QByteArray certficate = QByteArray(), const QString privateKey = QString());
 
     void setCapabilities(const QVariantMap &caps);
-    QVariantMap capabilities();
+    const Capabilities * capabilities() const;
     void setServerVersion(const QString &version);
     QString serverVersion();
 
@@ -170,7 +171,7 @@ private:
     QUrl _url;
     QList<QSslCertificate> _approvedCerts;
     QSslConfiguration _sslConfiguration;
-    QVariantMap _capabilities;
+    Capabilities _capabilities;
     QString _serverVersion;
     QScopedPointer<AbstractSslErrorHandler> _sslErrorHandler;
     QuotaInfo *_quotaInfo;
--- a/src/libsync/capabilities.h
+++ b/src/libsync/capabilities.h
@@ -1,0 +1,45 @@
+/*
+ * Copyright (C) by Roeland Jago Douma <roeland@famdouma.nl>
+ *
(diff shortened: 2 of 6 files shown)


dragotin:
Two questions/suggestions:
- I think we should have a Capabilities object that with proper methods, so that it reads:
```if( account()->capabilities()->pubLinkForcePassword() ) { ...``` instead of the map stuff we have now.
- Please do not use the QInputDialog. The dialog has a password field already, which should be made visible rather.

What do you think?

**Comment by rullzer** [comment]
> - I think we should have a Capabilities object that with proper methods, so that it reads:
```if( account()->capabilities()->pubLinkForcePassword() ) { ...``` instead of the map stuff we have now.

I agree. Now the questions is where and how to do this. Do we do this on a as needed basis... so I create a shareCapabilities class where all that is properly parsed. Or do we do it in 1 place where we parse everything?

> - Please do not use the QInputDialog. The dialog has a password field already, which should be made visible rather.

Yes, this should be changed indeed.

**Comment by rullzer** [comment]
@dragotin both your comments are address in the latest two commits...
Code should be better readable now and no annoying extra inputdialog anymore.

File: src/libsync/account.h [hunk 1]
```diff
@@ -170,7 +171,7 @@ protected Q_SLOTS:
     QUrl _url;
     QList<QSslCertificate> _approvedCerts;
     QSslConfiguration _sslConfiguration;
-    QVariantMap _capabilities;
+    Capabilities *_capabilities;
```
ogoffart: This should not be a pointer.  Just make it a normal object like the QVariantMap was.

src/libsync/account.h
```diff
@@ -139,7 +140,7 @@ class OWNCLOUDSYNC_EXPORT Account : public QObject {
     void setCertificate(const QByteArray certficate = QByteArray(), const QString privateKey = QString());
 
     void setCapabilities(const QVariantMap &caps);
-    QVariantMap capabilities();
+    Capabilities * capabilities();
```
ogoffart: return a const reference

ogoffart: and make const  (it should have been const before)

rullzer:
@ogoffart thanks and fixed!

src/libsync/capabilities.h
```diff
@@ -0,0 +1,50 @@
...
+ * server
+ * @ingroup libsync
+ */
+class Capabilities {
+
+public:
+    Capabilities(const Capabilities& capabilities);
+    Capabilities(const QVariantMap capabilities);
+
+    Capabilities operator=(Capabilities rhs)
```
ogoffart: This looks wrong, you are not assinging anything.

The default generated copy constructor and operator are fine.  Implementing them breaks the move sementic.  (see "Rule of zero")

rullzer: Aaah you are right...

PR #3467, opened 2015-07-16
Win32: Remove confusing option to remove AppData

### LINKED ISSUE #3461: [NSIS Installer] In case of update, do NOT show the choice to remove the data folder.
If somebody gets an auto update, it's pretty obvious that she wants to continue to use ownCloud, and thus does not want to remove the data directory. Is there a way to not show the following installer page if the installer was started from the updater:

rem_datafolder

danimo:
- This only works for the current user (while being a HKLM installer)
- The naming is confusing.

Fixes #3461


danimo:
Done: https://github.com/owncloud/client/wiki/Testing-Scenarios-2.0

PR #3468, opened 2015-07-16
Win32: Do not silent-update if the application is already up-to-date

### LINKED ISSUE #3368: [Windows] Update checker needs to check if the version it tries to silent-update to is still relevant to the current running version
We had this when switching to an old version 1.8.0 which downloaded a 1.8.1
We then manually installed 1.8.3 on top and the auto updater silently forced the update to 1.8.1
Meaning we got an old version 1.8.1 even though we had manually installed 1.8.3 some minutes before.

As discussed with @dragotin and @danimo today..
Found out when QAing around with @Dianafg76

danimo:
Fixes #3368


danimo:
@guruz please review

guruz:
Looks sensible :+1:

PR #3472, opened 2015-07-17
On Linux restart if new version is found on disk

### LINKED ISSUE #3253: background linux packaging update does not notify end user in client
I have 1.8.0 running on opensuse 13.1 -- yast already installed 1.8.1 as online update in the background.

The 1.8.0 client keeps running and its General tab says 'no updates available'

I expect to see a notification saying e.g: 
  An update has been installed. Please restart your owncloud client to complete the update.

Workaround: logout or reboot frequently.

dragotin:
This will fix #3253 

@ogoffart, @ckamm please.


test/testutility.h
```diff
@@ -103,6 +103,17 @@ private slots:
 
 
     }
+
+    void testVersionOfInstalledBinary()
+    {
+	if( isLinux() ) {
+                QString ver = versionOfInstalledBinary("/home/kf/owncloud.com/buildmirall/bin/owncloud");
```
ogoffart: this cannot work. 

Also it is only testing ver is not empty,  not that it is a version.

guruz:
Do we really want this automatic?
It's a bit impolite

(i think @ogoffart wrote this too but i don't see his comment anymore)

CC @danimo

test/testutility.h
dragotin: both fixed with subsequent commits.

ogoffart:
I still think it is not wise to restart the app.

Anyway, regarding the patch,  is see no reason to put this in utility, since it is only used on one place from the GUI.
Also it only checks the first line,  but maybe the first line will be something like `XError on Display XYZ` for some reason. and this may lead to infinite restart.

We also do not put all the parametter.  Maybe the user put some -style or -geometry parameter in the .dekstop file, and when restarting this would not be applied.

We would also restart even if the user is doing something in the UI like adding a new folder.

danimo:
I merged this for beta1, we can still refine/revert it if we do not like it.

Issue #3148, opened 2015-04-23
Desktop Sync Confirmation when New Folders Added (Confirm feature)


MTRichards:
As a user, I want the desktop client to ask me if I want to sync a folder that was newly shared with me to the device I am using so that I do not fill my computer up with all of the files from a network drive or sharepoint instance when it is shared with me. 
  
Acceptance criteria: 
-          When a new folder is shared with a user – external storage, a new internal storage folder, whatever – and the size of the folder exceeds the setting, a user is asked if they want to sync this folder to this desktop device 
-          The setting for the size is in the ownCloud settings tab in the desktop client, defaults to 100 MB, and can be changed by the end user to the size of their choice (and the default is brandable) 
-          When a single folder is shared and exceeds the threshold, a popup is created by the ownCloud desktop client and asks if the newly shared folder should be synced to this device – a yes no is required 
-          If a user has more than one new folder, a popup for each new share is created (as this should only happen infrequently) 
-          The popup identifies the name of the folder and the size 
-          The popup identifies the ownCloud desktop client as the originator of the message. 
-          The popup includes a message “A Folder larger than [setting] has been shared with you. Do you want to sync this folder to this device? (yes/no) 
-          The popup also includes a message “This can be changed at any time in the selective sync menu.” And has a link to the selective sync menu. 
-          The popup can be turned off in the settings menu, and an option to automatically sync or not sync all shares above this threshold can be set 
-          ownBrander would also be able to configure the “automatically do not sync new shares larger than setting” to set default behavior

moscicki:
Yes, that's definitely what is desired (and currently missing). On the pop-up: I am not sure if this the best way of communicating with a user. I would personally hate this (especially if that would then block the client application from other activity).  I would guess it would make for a far better user experience to have a notification (and also maybe an special icon in the tray)  and then a list of shared folders waiting for acceptance in a separate tab of the settings menu of the client.  Or having that integrated in a smart way with the existing "Choose What To Sync" dialog (which would then need to handle the logic of excluding by default newly created folders). The latter might be too confusing though with existing functionality of Choose What To Sync.

MTRichards:
@jancborchardt thoughts?

moscicki:
A case should also be considered of missing information on the folder size of external mounted storage (I believe it is not always possible to provide this, or the information may be inaccurate).

jancborchardt:
[...]

PR #3479, opened 2015-07-20
Confirm notifications.

ogoffart:
Issue #3148
Implement the tray notification as specified.
Unfortunately, the tray notification cannot be clicked


ogoffart:
@dragotin 

I guess a question does not make sens in the tray notification bubble since there is no buttons.
Should the acceptence criteria be changed?

dragotin:
As discussed yesterday, we do not go this way. Thanks anyway...

Issue #3483, opened 2015-07-21
[Multiaccount] In MAC, You can use the "Share with ownCloud " if you select the folder ROOT and introduce de Password and Set, appears error 403


Dianafg76:
### Steps to reproduce

  1  Open your ownCloud desktop client
  2  Go to Open Folder "ownCloud"
  3  If you select the folder Root (onwCloud)
  4  Right click on the Folder Root (onwCloud)
  5  Select "Share with ownCloud"
  6  Introduce de Password
  7  Press Enter

### Expected behaviour
You should not have the option "Share with ownCloud"

### Actual behaviour
Appears an error: OCS API error code: 403, related to #3463 

![screen shot 2015-07-21 at 15 22 00](https://cloud.githubusercontent.com/assets/12032058/8801623/51df3724-2fbc-11e5-898c-8fde39d59065.png)

### Server configuration
Desktop v ownCloud-2.0.0.2551-nightly20150721.pkg
Server v {"installed":true,"maintenance":false,"version":"8.1.0.8","versionstring":"8.1 RC2","edition":"Enterprise"}

rullzer:
Well it is not possible to share the root. So this should be fixed when #3483 is fixed.

rullzer:
Quick fix in #3495

rullzer:
Should be fixed by #3495

PR #3495, opened 2015-07-24
Don't allow sharing of the root folder

### LINKED ISSUE #3482: [Multiaccount] In MAC, You can use the "Share with ownCloud " if you select the folder ROOT
### Steps to reproduce
1. Open your ownCloud desktop client
2. Go to Open Folder "ownCloud"
3. If you select the folder Root (onwCloud)
4. Right click

### Expected behaviour
You should not have the option "Share with ownCloud", realed con #3463 

### Actual behaviour
You have the option "Share with ownCloud"
screen shot 2015-07-21 at 15 05 25

### Server configuration
Desktop v ownCloud-2.0.0.2551-nightly20150721.pkg
Server v {"installed":true,"maintenance":false,"version":"8.1.0.8","versionstring":"8.1 RC2","edition":"Enterprise"}

rullzer:
Sharing the root folder is not possible so don't allow the user to do this. Fixes #3482 and #3483 

Note that this is just a quick fix.
For better UX we should have a way to tell the plugins that the registered path is the root folder. That way they can simply grey out the menu item on the root folder.

### DIFF
--- a/src/gui/socketapi.cpp
+++ b/src/gui/socketapi.cpp
@@ -420,6 +420,13 @@ void SocketApi::command_SHARE(const QString& localFile, QIODevice* socket)
         const QString folderForPath = shareFolder->path();
         const QString remotePath = shareFolder->remotePath() + localFile.right(localFile.count()-folderForPath.count()+1);
 
+        // Can't share root folder
+        if (QDir::cleanPath(remotePath) == "/") {
+           const QString message = QLatin1String("SHARE:CANNOTSHAREROOT:")+QDir::toNativeSeparators(localFile);
+            sendMessage(socket, message);
+            return;
+        }
+
         SyncJournalFileRecord rec = dbFileRecord_capi(shareFolder, localFile);
 
         bool allowReshare = true; // lets assume the good


guruz:
Let's say I have a sync folder syncing

remote.php/webdav

Or two sync folders

remote.php/webdav/f1
remote.php/webdav/f2

Would your patch deny sharing the root of both scenarios?
Even though I think scenario 2 is fine?

rullzer:
@guruz nope this still works. It only checks if you try to share the root of the server.

dragotin:
While this patch is good (:+1:) I think it should also have a component in the explorer/finder plugins that does not offer the share link if it is the root folder.

rullzer:
@dragotin I totally agree. But I do not use any of the plugins currently (running xmonad) so not trivially for me to fix in this PR.

I propose to get this in (to at least fix the error) and to create separate issues for the plugins to fix stuff.

dragotin:
@rullzer yes, very good.

rullzer:
Issues created: #3503 #3504 #3505 and #3506

dragotin:
thanks @rullzer

Siedlerchr:
I just ran into this problem and I want to say that this behavior is not (longer) compatible with the server side.

 I have set up multiple folders for sync on the client side, and on the client side I can't share the "root" of each folder.

Example: Shared Folders.
 \XXXX\myFolder
  \YYYY
Via the client only "myFolder" is sharable. Neither XXXX nor YYYY are sharable via the client.
But on the server side, I can share both XXXX and YYYY, which map on the server side to 
<OCRoot>/XXXX  and <OCRoot>/YYYY 

Client version 2.1.0 (build 5683
Server: 8.2.1 (stable)

PR #3513, opened 2015-07-29
Fix nautilus plugin

### LINKED ISSUE #2775: [OS X/Windows] When the overlay icon is not green, you can't share by link from desktop
### Expected behavior

You can't share by link a file or folder which not shown a green overlay icon  
### Actual behavior

We don't have into account he state of the file, if it shows red/yellow/ or even blue overlay icon, you  have the option to share it by link but it keeps loading and it never works 

captura de pantalla 2015-02-04 a las 8 50 04 am
### Steps to reproduce
1. Click on share by link in a file/folder with red/yellow/ or blue overlay icon 
### Client configuration

Version 1.8.0-nightly20150204(build 1971)

Operating system: OS X 10.10.1

dragotin:
Fixes #2775 for the nautilus plugin, plus another bugfix for files with colons in it.


danimo:
:+1:

Issue #2340, opened 2014-10-20
[OS X] Use 10.10 Finder Sync icon indication


guruz:
If running on 10.10:

https://developer.apple.com/library/mac/releasenotes/MacOSX/WhatsNewInOSX/Articles/MacOSX10_10.html

> FinderSync (FinderSync.framework). The FinderSync framework enables Finder extensions to:
Express interest in specific folder hierarchies
Provide "badges" to indicate the status of items inside those hierarchies
Provide dynamic menu items in Finder contextual menus, when the selected items (or the window target) are in those hierarchies
Provide a Toolbar Item that displays a menu with dynamic items (even if the selection is unrelated)

This would replace our overlay icon code.

https://developer.apple.com/library/mac/documentation/FinderSync/Reference/FinderSync_Framework/

nemphys:
Yes, this is the way to go; I suppose DropBox is using this, since the overlays were missing when I first upgraded to 10.10 and after a few weeks (ie. DB update), they came back, but different than ever before.

iliyasbasir:
Screen Shot
![screen shot 2015-03-31 at 16 34 32](https://cloud.githubusercontent.com/assets/10195796/7420697/9a4ac358-ef4c-11e4-8ff2-1cbded32c437.png)
![screen shot 2015-03-31 at 16 34 18](https://cloud.githubusercontent.com/assets/10195796/7420699/9ce19b6e-ef4c-11e4-98a0-fc33eddcf65a.png)

jturcotte:
I tried to get it to work by hacking things together but I'm currently stuck in a corner: the FinderSync requires sandboxing to be enabled for the app extension bundle, and that means in return that the code can only communicate through unix sockets with other sandboxed applications (of the same "application group")

https://developer.apple.com/library/mac/documentation/Security/Conceptual/AppSandboxDesignGuide/AppSandboxInDepth/AppSandboxInDepth.html#//apple_ref/doc/uid/TP40011183-CH3-SW24

I tried to mark the current socket path with temporary sandbox exceptions, or to tell the client to open the server inside the app group container path and OS X obviously doesn't let me open the socket for listening.

This means that we won't be able to use those fancy new icons unless we also make the desktop client application sandboxed. I'm assuming that this won't be trivial.

guruz:
@danimo @dragotin ^^ FYI

jturcotte:
[...]

PR #3531, opened 2015-07-31
Remove need for UPDATE_VIEW to refetch the status #2340

jturcotte:


danimo:
@ogoffart please comment on the items

guruz:
@ogoffart Probably best to move into the beta to get most testing?

@jturcotte It would be good to mention the improved/fixed issue numbers in the commits (if there is issues..)

This would help the Eagle-issue?

jturcotte:
@guruz This is currently only the case for OS X 10.10, others are still doing the refetching but possibly unnecessarily. We can change that for 2.1, but the eagle-issue should already be fixed with the status cache cleaning patch that went into 1.8.x

jturcotte:
Anybody interested in touching that soft "Merge pull request" button?

Issue #3542, opened 2015-08-03
New empty folder downloads when confirmation limit is set to zero MB


phil-davis:
### Expected behaviour
Technically the confirmation limit is for folders larger than the limit.
There are a few users here with job-support type workflows where they would create a new folder for each job. The staff doing the jobs can set their confirmation limit to 0 MB. Then no new folders would be automatically downloaded. When the staff member wants to look at or work on the files in a job (folder) they can choose to sync it.

### Actual behaviour
In the case of a new empty folder, the empty folder is always downloaded. There is no way to set the confirmation limit so that empty folders are not downloaded.

### Steps to reproduce
1. Create and empty folder
2. Share it with someone who has their confirmation limit set to zero.
3. They get the new empty folder.

### Server configuration
ownCloud version: 8.0

### Client configuration
Client version: 2.0.0-nightly 201508101 (build 5326)

Operating system: Windows10

OS language: English (US)

Installation path of client: default

phil-davis:
This can also be an issue if the workflow is that the central user creates new "job" folders that are empty for a few minutes and shares them. Those get auto-synced to ownCloud clients. Then a few minutes later they put a bunch of files in the folder. Since it has already been synced (when it was empty) the new files just added now all download.
IMHO it would be more convenient if the user could set the confirmation limit to 0 MB and that applied to new folders of size >=0 MB.
In a sense, an empty folder is already larger than 0MB - there must be some folder meta-data that takes some space, the non-empty folder name guarantees that the total size of (folder+file in the folder) is strictly > 0MB anyway.

guruz:
Maybe a special handling of ```-1``` = "Don't download any folder without confirmation" or so would be interesting. Or just changing the meaning of 0.

phil-davis:
Maybe just changing the limit comparison from "if >" to "if >=" will make this work. (I haven't looked in the code to see exactly what is done for the comparison, so pardon my guess if it is wrong).
I don't think anyone would complain if they set their limit to "larger than 20MB" but then got notified about a new folder that is exactly 20MB. In any case, users are not going to be sure if the limit is 20,000,000 bytes (decimal MB) or calculated internally as MiB anyway. So they are not really specifying a totally exact "to the byte" limit here.

MTRichards:
Related https://github.com/owncloud/client/issues/3543

MTRichards:
[...]

PR #3549, opened 2015-08-04
Notify of big folders that are greater than or equal to the limit

phil-davis:
This should allow for the case when a user has set the limit to 0 and a new empty folder appears on the server. The folder will have size 0 (no files in it). Doing the >= test here will mean that the user will be prompted about the new folder, which I think is the behaviour they would expect.
The side-effect of this change is that if the user has a limit of, for example, 10,000,000 bytes (10 MB) and a new folder comes along with exactly 10,000,000 bytes of content then they will now be prompted about it. Before the change such a new folder would have been auto-synced without prompting the user. I do not think this is a big deal - I cannot believe that users will be counting exact bytes for this limit, they are just setting a rough number of MB at the UI.
Should fix https://github.com/owncloud/client/issues/3542

### DIFF
--- a/src/libsync/discoveryphase.cpp
+++ b/src/libsync/discoveryphase.cpp
@@ -84,7 +84,7 @@ bool DiscoveryJob::checkSelectiveSyncNewFolder(const QString& path)
     }
 
     auto limit = _newBigFolderSizeLimit;
-    if (result > limit) {
+    if (result >= limit) {
         // we tell the UI there is a new folder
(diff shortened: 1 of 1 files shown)


danimo:
@ogoffart: Should we do this, or make 0 a special case?

ogoffart:
Looks good to me, we already handle -1 as no limit.

Thanks for the patch.

MTRichards:
Thanks @phil-davis

Issue #3382, opened 2015-06-28
Slow connections: Less Parallelism


rienesl:
### Expected behaviour

Sync multiple big files
### Actual behaviour

One file is aborted with "Connection closed", two others with "Operation canceled"
### Steps to reproduce
1. Have minimum 3 big files (>=100MB) which are not synchronized
2. Have a slow connection (<=56k)
3. Try to sync that
### Server configuration

Operating system:
Linux (Xubuntu)

Web server: 
apache 2.2.22-1

Database:
mysql 5.5.43-0

PHP version:
5.4.42-1

ownCloud version:
8.0.4

Storage backend:
DAS
### Client configuration

Client version:
1.8.1, 1.8.2, 1.8.3

Operating system:
Linux (Xubuntu)

OS language:
English

Installation path of client:
default
### Additional Comments

I guess, everything would run (slow but) fine, if I could deactivate the simultaneous downloads. If you download more than one big file at the time, one of the files will get a timeout because of the slow connection. I have similar problems when downloading multiple files with a browser. There are two ways of solving that problem:
1) Download just one file
2) using a download manager with resume-function
### Edit:

Maybe it has something in common with #3318? Looks similar, but neither this is a Mac system, nor a limit is set

guruz:
Do you mean uploading from desktop or downloading to desktop?
What is your Qt version? (see settings dialog)
Does it work if you ```export OWNCLOUD_MAX_PARALLEL=1``` in the shell before running ```owncloud``` itself?

guruz:
( @ogoffart @dragotin If the ```export``` workaround works: We could tone down parallelism if we detect a slow connection..?)

rienesl:
@guruz:
About the question of up- or downloading: I realized this problem while downloading, I havn't tried it with uploading (seen from desktop client).
About the Qt version and the parameter to set: Currently I'm at work, but I'll check this evening.

rienesl:
Hi, Qt is version 4.8.6. I have started owncloud with the parameter mentioned above, sync started. I'll leave the computer on and go to sleep now. Tomorrow morning, we'll see the results :)

rienesl:
First results seem to be great. This night it successfully downloaded a 103MB file, when I left home it was at 66% of a 146MB file. I'll report this evening, when I'm back home, but I think this parameter does what I need.
[...]

Issue #3095, opened 2015-04-10
[1.8.0] displayed Transfer speed Higher than limit


cdamken:
### Expected behavior
Limit upload speed to 50 KB/s
Upload to 50KB/s

### Actual behavior
Limit upload speed to 50 KB/s
Upload to 70KB/s
![speed](https://cloud.githubusercontent.com/assets/7264495/7089065/5eb2a936-df99-11e4-972b-ee48d9f47cde.png)

### How to reproduce it:
Sync-client 1.8.0 
1.- Add a big File to the upload queue (I added a 2 GB file)
2.- Change the transfer sped to 50 KB/s
3.- Pause 
4.- Resume
5.- Wait until the speed stabilizes.

Tested with different speeds.
when 5KB/s
upload 8KB/s

@MorrisJobke 

00002794

dragotin:
@ogoffart mind to check this? Maybe it's just fuzzy calculation in the progress widget.

ogoffart:
What platform ?  (which version of Qt)

CC @guruz

guruz:
Carlos, have you tried a system tool that displays actual speed? I dont know if windows task manager can do it

cdamken:
@ogoffart  Windows  with 1.8.0. How can I find out the Qt version?

@guruz I will test wit another tool, it's hard to see directly in the Windows network notification, but anyway the widget tells another speed.

caratsch:
@guruz the speed looks right on the system (Resource Monitor shows the right speed). The sync client shows now 260 KiB/s, but only 10 KiB/s are allowed and currently transferred.

cdamken:
@guruz Then the only problem is the speed shown in the widget

ogoffart:
We probably should remove that speed there, it is totaly not accurate.

guruz:
...let's make it a rolling average calculation instead of momentary speed display.

ckamm:
@guruz I've code for that ready to be merged, pending review. It changes it to an exponential moving average. See #2328. Maybe I'll just merge these to master for testing...

guruz:
@ckamm Sounds good.. :)

cdamken:
Is something fixed in 1.8.1? Should I test it?

ckamm:
@cdamken Not that I'm aware. There are some patches in master (for 2.0).

guruz:
@cdamken Can you please try ownCloud-2.0.0.5092-nightly20150520-setup.exe ?

[...]

PR #3563, opened 2015-08-06
Propagator: do not use parallelism when we have bandwidth limit

ogoffart:
When user wants to limit the bandwidth, he does not care about speed
anymore. And parallelism on slow network might cause problems.

For issue #3382
Will also help for #3095


phil-davis:
That looks like "a good thing". It should help us in remote places where bandwidth is low and high packet loss... 1 file at a time will give the best chance for an up/down load to actually complete.

guruz:
(cherry picked instead of merged)

PR #3606, opened 2015-08-13
Autoupdate Doc bullet point rendering

phil-davis:
A space at the start of the bullet point about "Locate and open the following file" seems to be causing that bullet point to not line up nicely on the rendered page.


phil-davis:
Closed in favour of https://github.com/owncloud/client/pull/3624

PR #3611, opened 2015-08-13
Fix linking with libinotify on OpenBSD.

yason:
Builds fine on NetBSD, builds and runs on FreeBSD.

### DIFF
--- a/cmake/modules/FindINotify.cmake
+++ b/cmake/modules/FindINotify.cmake
@@ -4,20 +4,27 @@
 
 # This module defines
 #  INOTIFY_INCLUDE_DIR, where to find inotify.h, etc.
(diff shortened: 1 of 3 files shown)


guruz:
@ogoffart Can you check if this breaks the linux build?

guruz:
@yason Thanks a lot, can you fill https://owncloud.org/contribute/agreement/ ?

yason:
I have already sent this form to Frank. BTW, this is not referenced to an issue I've opened some time ago. I'm still looking for solution.

ogoffart:
:+1:  compiled on linux and looks good to me

ogoffart:
This is now merged in 2.0

PR #3661, opened 2015-08-20
Allow setting of public upload on link shares

### LINKED ISSUE #3551: sharing with write-access not possible from client
In the windows client it is not possible to share a directory with the allowance to edit it. The checkbox for "edit/write access" is missing.

Web-Version: Owncloud 8.1
Client 1.8.4 (build 5267)

rullzer:
Fixes #3551

Allow users to set write access on link shares if this is enabled by the server.
Will require the 8.2 (current master) on the server to expose the capability.

### DIFF
--- a/src/gui/sharedialog.cpp
+++ b/src/gui/sharedialog.cpp
@@ -29,10 +29,20 @@
 #include <QBuffer>
 #include <QFileIconProvider>
 #include <QClipboard>
+#include <QFileInfo>
 
 namespace {
     int SHARETYPE_PUBLIC = 3;
+
+//    int PERMISSION_READ = 1;
+    int PERMISSION_UPDATE = 2;
+    int PERMISSION_CREATE = 4;
+//    int PERMISSION_DELETE = 8;
+//    int PERMISSION_SHARE = 16;
+//    int PERMISSION_ALL = 31;
 }
+
+
 
 namespace OCC {
 
@@ -50,6 +60,10 @@ ShareDialog::ShareDialog(AccountPtr account, const QString &sharePath, const QSt
     setObjectName("SharingDialog"); // required as group for saveGeometry call
 
     _ui->setupUi(this);
+
+    //Is this a file or folder?
+    _isFile = QFileInfo(localPath).isFile();
+
     _ui->pushButton_copy->setIcon(QIcon::fromTheme("edit-copy"));
     _ui->pushButton_copy->setEnabled(false);
     connect(_ui->pushButton_copy, SIGNAL(clicked(bool)), SLOT(slotPushButtonCopyLinkPressed()));
@@ -64,8 +78,10 @@ ShareDialog::ShareDialog(AccountPtr account, const QString &sharePath, const QSt
     _pi_link     = new QProgressIndicator();
     _pi_password = new QProgressIndicator();
     _pi_date     = new QProgressIndicator();
+    _pi_editing  = new QProgressIndicator();
     _ui->horizontalLayout_shareLink->addWidget(_pi_link);
     _ui->horizontalLayout_password->addWidget(_pi_password);
+    _ui->horizontalLayout_editing->addWidget(_pi_editing);
     // _ui->horizontalLayout_expire->addWidget(_pi_date);
 
     connect(_ui->checkBox_shareLink, SIGNAL(clicked()), this, SLOT(slotCheckBoxShareLinkClicked()));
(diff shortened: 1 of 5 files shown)


src/gui/sharedialog.cpp
```diff
@@ -558,6 +572,46 @@ void ShareDialog::slotPushButtonCopyLinkPressed()
...
+    _ui->checkBox_editing->setEnabled(false);
+    _pi_editing->startAnimation();
+
+    QUrl url;
+    QList<QPair<QString, QString> > requestParams;
+    QByteArray verb("PUT");
+
+    QString value = QString::fromLatin1(publicUpload ? "true" : "false");
+
+    url = Account::concatUrlPath(_account->url(), QString("ocs/v1.php/apps/files_sharing/api/v1/shares/%1").arg(_public_share_id));
```
dragotin: Minor: I think it is preferable to use QLatin1String("") rather than QString::fromLatin1()
Also consider using const, ie. const QByteArray verb("PUT");

src/gui/sharedialog.cpp
```diff
@@ -558,6 +572,46 @@ void ShareDialog::slotPushButtonCopyLinkPressed()
...
+{
+    qDebug() << Q_FUNC_INFO << reply;
+
+    QString message;
+    int code = getJsonReturnCode(reply, message);
+    if (code != 100) {
+        displayError(code);
+    }
+
+    _ui->checkBox_editing->setEnabled(true);
```
dragotin: If the code was != 100 (error case) is than still good to enable the editing checkbox here?

src/gui/sharedialog.cpp
rullzer: mmm chaning to QLating1String fails because it is a QList<QPair<QString, QString> >
Maybe we can fix that in a later PR... Becuase it needs changes at more places. Probably it could become a QList<QPair<QLatin1String, QString> >

You are right about the const stuff.

src/gui/sharedialog.cpp
rullzer: Good point

rullzer:
@dragotin fixed comments and rebased onto master.

rullzer:
Rebased :)

src/gui/sharedialog.cpp
```diff
@@ -355,6 +362,13 @@ void ShareDialog::slotSharesFetched(const QVariantMap &reply)
                 _ui->checkBox_expire->setChecked(false);
             }
 
+            if (data.value("permissions").isValid()) {
+                int permissions = data.value("permissions").toInt();
+                if (permissions == 7) {
+                    _ui->checkBox_editing->setChecked(true);
```
ckamm: Should this be set to false if permissions != 7?

src/gui/sharedialog.cpp
```diff
@@ -559,6 +573,43 @@ void ShareDialog::slotPushButtonCopyLinkPressed()
     clipboard->setText(_shareUrl);
 }
 
+void ShareDialog::slotCheckBoxEditingClicked()
+{
+    ShareDialog::setPublicUpload(_ui->checkBox_editing->checkState() == Qt::Checked);
```
ckamm: why the `ShareDialog::`?

src/libsync/capabilities.h
```diff
@@ -36,6 +36,7 @@ class OWNCLOUDSYNC_EXPORT Capabilities {
     bool publicLinkEnforceExpireDate() const;
     int  publicLinkExpireDateDays() const;
     QStringList supportedChecksumTypes() const;
+    bool publicLinkAllowUpload() const;
```
ckamm: Minor: Put it with the other sharing capabilities.

ckamm:
Looks good! :)

src/gui/sharedialog.cpp
dragotin: Why '7' at all? Can't we have a kind of speaking define or so?

dragotin:
Yes, merge please :+1:

src/gui/sharedialog.cpp
rullzer: Fixed

src/gui/sharedialog.cpp
rullzer: Good question... fixed...

src/libsync/capabilities.h
rullzer: Done :)

PR #3663, opened 2015-08-20
don't show share menu item for top level folders

### LINKED ISSUE #3662: nautilus plugin shows "Share with Owncloud" menu item for top leven folders
Even top level folders cannot be shared the nautilus plugin shows the "Share with Owncloud" menu item for them.
When trying to use it, it is possible to enter a password and expiry date but after doing so the Server returns a 403 error

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25990011-nautilus-plugin-shows-share-with-owncloud-menu-item-for-top-leven-folders?utm_campaign=plugin&utm_content=tracker%2F216457&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F216457&utm_medium=issues&utm_source=github).
</bountysource-plugin>

individual-it:
Top level folders cannot be shared. If you try you receive an 403 error.
So we should not show the "Share with Owncloud" menu item for them

fixes #3662

### DIFF
--- a/shell_integration/nautilus/syncstate.py
+++ b/shell_integration/nautilus/syncstate.py
@@ -157,13 +157,18 @@ class MenuExtension(GObject.GObject, Nautilus.MenuProvider):
         # internal or external file?!
         syncedFile = False
         for reg_path in socketConnect.registered_paths:
+            topLevelFolder=False
             filename = get_local_path(file.get_uri())
             #check if its a folder (ends with an /), if yes add a "/" otherwise it will not find the entry in the table
             if os.path.isdir(filename+"/"):
                 filename=filename+"/"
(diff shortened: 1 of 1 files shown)


phil-davis:
This is the PR against master for this fix. Is there any reason that either this PR or the one against 2.0 https://github.com/owncloud/client/pull/3691 has not been merged?
And then whichever becomes the "real" one will end up getting merged or cherry picked or whatever into its opposite branch.

ogoffart:
Was merged in 2.0 branch

PR #3712, opened 2015-08-30
Support for sandstorm.io server URLs

mnutt:
This pull request is intended to support owncloud/client connecting to a server running under http://sandstorm.io.

Sandstorm apps can generate web keys to give API access to a third party or client app. Web keys are of the form:

http://api.local.sandstorm.io:6080#WHLMLAAPFg1uHZOAcIxXsPKbYAUjM6GFALDzl5H9f_w

From that URL, one can connect to api.local.sandstorm.io with basic auth with the URL fragment as the password and anything as the username. Sandstorm tokens act as both authentication and routing.

I chose to go the route of detecting sandstorm URLs and prefilling http user/pass rather than coming up with a whole different SandstormCreds class/wizard page (a la Shibboleth) because it was a much smaller change, but that's another way this could be implemented.

### DIFF
--- a/src/gui/owncloudsetupwizard.cpp
+++ b/src/gui/owncloudsetupwizard.cpp
@@ -34,6 +34,7 @@
 #include "creds/credentialsfactory.h"
 #include "creds/abstractcredentials.h"
 #include "creds/dummycredentials.h"
+#include "creds/httpcredentialsgui.h"
 
 namespace OCC {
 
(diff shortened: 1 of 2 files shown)


src/gui/owncloudsetupwizard.cpp
```diff
@@ -124,12 +125,21 @@ void OwncloudSetupWizard::slotDetermineAuthType(const QString &urlString)
...
     }
     AccountPtr account = _ocWizard->account();
+
+    // Detect a Sandstorm web key and derive the credentials from it.
+    if (url.hasFragment() && url.path() == "") {
+      account->setCredentials(new HttpCredentialsGui("sandstorm", url.fragment(), NULL, NULL));
+      account->setCredentialSetting("user", "sandstorm");
+      account->setCredentialSetting("password", url.fragment());
+    } else {
+      // Set fake credentials beforfe we check what credential it actually is.
```
phil-davis: s/beforfe/before
Might as well fix the typo that was in the original comment text :)

danimo:
I don't see why we would add support for proprietary services that we don't even use and that we have no tests for. Sorry.

mnutt:
I probably should have explained Sandstorm a bit better. It's an open source platform for hosting open source web apps. It serves some of the same purpose as Bitnami, but Sandstorm provides its own authentication system, which is what this PR addresses.

When you refer to the tests, are you talking about testing the code in the PR or testing Owncloud within Sandstorm? I'm happy to write tests for the code in the PR.

ocdtrekkie:
Sandstorm is Apache2 licensed, @danimo : https://github.com/sandstorm-io/sandstorm/blob/master/LICENSE

jancborchardt:
Reopening just so this doesn’t get lost. Sandstorm is not proprietary. :)

@ocdtrekkie can you maybe test this pull request?

ocdtrekkie:
Yeah, I can try to find some time to build this and test it.

LukasReschke:
On the risk of getting some negative internet karma here, I technically do agree with @danimo.

The official ownCloud client cannot add workarounds for each custom integration. If we start this way we're getting into a massive maintenance hell.
Even with this change here we simply **cannot** provide proper QA and any change may obviously also introduce potential new bugs.

:-1: from my side. Sorry.

dragotin:
I agree with @danimo and @LukasReschke. No special implementations for special services. Especially because there is no easy way to disable it.

Maybe there could be a clever way to plug these kind of stuff in and out using a config file, or a plugin or so?

mnutt:
@dragontin, @LukasReschke: your reasoning makes sense to me, I don't want to add extra support burden for something outside of the standard OwnCloud setup.

As a workaround, would you consider a pull request that allowed Server URLs to be entered as https://username:password@example.com and then pre-populated the second screen with username and password? It wouldn't be Sandstorm-specific, but would allow someone to copy-and-paste a single string rather than three separate ones. OwnCloud currently already supports that URL structure for the initial request, the only change would be prepopulating the username and password.

guruz:
@mnutt I'd say yes (if you mean the URL edit in the wizard page 1)

LukasReschke:
Seconded. That would be nice I think.

mnutt:
Great, thanks! The follow-on PR is here: https://github.com/owncloud/client/pull/4100

Issue #3737, opened 2015-09-02
Enable Full Sharing (user/group) from Desktop


MTRichards:
Desktop Full Sharing
As an ownCloud user, I want to be able to use the ownCloud internal sharing mechanism from Windows explorer so that I can quickly and easily share files with other ownCloud users without using share links.

Acceptance Criteria
- User should right click on the file and see the current share with menu, and then be able to select a new option “share within ownCloud”
- Share within ownCloud opens up a new window on the desktop client
- In this case, the user can enter a group name or a user name (or part of either) and click a button to search for the user or group. Note: this is different than the current web auto search interface. We want to be certain we are not flooding the server API with false starts, so the search now button is important
- While the API is searching, the desktop window shows a little “searching” icon that shows the system is searching and in progress (@janborchart for specific look and feel)
- There is a cancel button for this search. However, while one search is running a second cannot be initiated to avoid flooding the server with too many requests
- When the search returns group or usernames (in the same order as on the server, no change in behavior on the search results) and a user can select from the drop down list
- At most 40 results are returned. If more results are returned, the dropdown shows the 40 but notifies the user that the user and group search returned too many results, and that the search needs to be narrowed.
- When a user or group is selected, the standard CRUDS capabilities are lined up.
- This assumes that the complete desktop share with link option is already implemented, including setting a password, setting a password that meets policy,setting expiration.
- This also assumes that the “notify user via email” option currently on the server is also implemented on the desktop
- Should link to the server user and group search issue for 8.2 - dependency
- Should link to mobile sync design for consistency

@dragotin @danimo

MTRichards:
@dragotin @danimo @guruz @ogoffart @karlitschek @jancborchardt

rullzer:
Now that the Sharee API is in core I can revive some old code. Cool to finally implement this on the desktop.

Just to be clear. We want to have this in a different window?
Not 1 window wit 2 tabs or whatever?

karlitschek:
this will be a huge step forward! :+1:  :-)

phil-davis:
+1 also
And I will just mention that the group/user sharing dialogs in the client will need to respect the admin server-side settings of sharing restrictions, like:
- Allow apps to use the Share API
- Allow resharing
- Restrict users to only share with users in their groups
- Exclude groups from sharing

That will avoid users on the client trying to do something and then the server spitting back a "not allowed" message.

rullzer:
@phil-davis yes good points!

> - Allow apps to use the Share API

[...]

PR #3773, opened 2015-09-07
Split sharedialog code

rullzer:
In preparation of the full desktop sharing (https://github.com/owncloud/client/issues/3737) the sharedialog code had to be somewhat unentangled.

There now is more code in libsync and the gui is cleaned up a bit.

* OCSJob class added. To allow a nice abstraction also for the new sharee endpoint code (which is required)
* OcsShareJob class is now in libsync and has more generic functions like createShare, setPassword etc. So the GUI part can just say do this.
* ShareDialog is cleaned up

My tests show that everything still works as expected but please run your own tests to make sure I did not break anything!

CC: @dragotin @guruz @ogoffart and others that want to test!

### DIFF
--- a/src/gui/ocsjob.h
+++ b/src/gui/ocsjob.h
@@ -48,18 +48,13 @@ protected:
     void setVerb(const QByteArray& verb);
 
     /**
-     * The url of the OCS endpoint
+     * Add a new parameter to the request.
+     * Depending on the verb this is GET or POST parameter
      *
-     * @param url
+     * @param name The name of the parameter
+     * @param value The value of the parameter
      */
-    void setUrl(const QUrl& url);
-
-    /**
-     * Set the get parameters to the url
-     *
-     * @param getParams list of pairs to add to the url
-     */
-    void setGetParams(const QList<QPair<QString, QString> >& getParams);
+    void addParam(const QString& name, const QString &value);
 
     /**
      * Set the post parameters
@@ -110,8 +105,7 @@ private slots:
 
 private:
     QByteArray _verb;
-    QUrl _url;
-    QList<QPair<QString, QString> > _postParams;
+    QList<QPair<QString, QString> > _params;
     QVector<int> _passStatusCodes;
 };
 
--- a/src/gui/ocsjob.h
+++ b/src/gui/ocsjob.h
@@ -27,13 +27,18 @@ namespace OCC {
 /**
  * @brief The OcsShareJob class
  * @ingroup gui
+ *
+ * Base class for jobs that talk to the OCS endpoints on the server.
+ * All the comminication logic is handled in this class.
+ *
+ * All OCS jobs (e.g. sharing) should extend this class.
  */
-class OCSJob : public AbstractNetworkJob {
+class OcsJob : public AbstractNetworkJob {
     Q_OBJECT
 
 protected:
 
-    explicit OCSJob(AccountPtr account, QObject* parent = 0);
+    explicit OcsJob(AccountPtr account, QObject* parent = 0);
 
     /**
      * Set the verb for the job
--- a/src/gui/ocsjob.h
+++ b/src/gui/ocsjob.h
@@ -1,0 +1,115 @@
+/*
+ * Copyright (C) by Roeland Jago Douma <roeland@famdouma.nl>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; version 2 of the License.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+ * for more details.
+ */
+
+#ifndef OCSJOB_H
+#define OCSJOB_H
+
+#include "accountfwd.h"
+#include "abstractnetworkjob.h"
+
+#include <QVector>
+#include <QList>
+#include <QPair>
+#include <QUrl>
+
+namespace OCC {
+
+/**
+ * @brief The OcsShareJob class
+ * @ingroup gui
+ */
+class OCSJob : public AbstractNetworkJob {
+    Q_OBJECT
+
+protected:
+
+    explicit OCSJob(AccountPtr account, QObject* parent = 0);
+
+    /**
+     * Set the verb for the job
+     *
+     * @param verb currently supported PUT POST DELETE
+     */
+    void setVerb(const QByteArray& verb);
+
+    /**
+     * The url of the OCS endpoint
+     *
+     * @param url
+     */
+    void setUrl(const QUrl& url);
+
+    /**
+     * Set the get parameters to the url
+     *
+     * @param getParams list of pairs to add to the url
+     */
+    void setGetParams(const QList<QPair<QString, QString> >& getParams);
+
+    /**
+     * Set the post parameters
+     *
+     * @param postParams list of pairs to add (urlEncoded) to the body of the
+     * request
+     */
+    void setPostParams(const QList<QPair<QString, QString> >& postParams);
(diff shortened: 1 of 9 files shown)


src/libsync/ocssharejob.h
```diff
@@ -0,0 +1,90 @@
...
+     *
+     * @param date The expire date, if this date is invalid the expire date
+     * will be removed
+     */
+    void setExpireDate(const QDate& date);
+
+    /**
+     * Set the password of a share
+     *
+     * @param password The apssword of the share, if the password is empty the
```
phil-davis: s/apssword/password

ogoffart:
Ideally, my long term goal would be that libsync would be for the sync engine.
Adding these jobs there goes against my goal.

rullzer:
@ogoffart ah mm, then we could also move the jobs to the gui for now... 
but also they are not really gui... yet is not ideal...

I'll move them to the gui for now.

guruz:
Not sure I agree with @ogoffart .. sharing will be more prominent in future.. hm

dragotin:
Not agreeing with @ogoffart is hardly an option.

rullzer:
So finally my qt5 setup is running :)
All the files are now back in gui. But we might want to move to an libocs or libshare in the not to distant future. But I need somebody with some cmake-fu for that.

Please tests and review :)

src/gui/ocssharejob.h
```diff
@@ -0,0 +1,90 @@
...
+        LINK = 3
+    };
+
+    /**
+     * Constructor for new shares or listing of shares
+     */
+    explicit OcsShareJob(AccountPtr account, QObject *parent = 0);
+
+    /**
+     * Constructors for exisiting shares of which we know the shareId
```
phil-davis: s/exisitng/existing/

src/gui/ocssharejob.h
```diff
@@ -0,0 +1,90 @@
...
+     */
+    explicit OcsShareJob(int shareId, AccountPtr account, QObject *parent = 0);
+
+    /**
+     * Get all the shares
+     *
+     * @param path Path to requests shares for (default all shares)
```
phil-davis: s/requests/request/

src/gui/ocsjob.h
```diff
@@ -0,0 +1,115 @@
...
+private:
+    QByteArray _verb;
+    QUrl _url;
+    QList<QPair<QString, QString> > _postParams;
+    QVector<int> _passStatusCodes;
+};
+
+}
+
+#endif // OCSSHAREJOB_H
```
phil-davis: OCSJOB_H

rullzer:
@phil-davis thanks. Fixed

guruz:
I'm not working on oC these days, maybe @danimo @dragotin can have a look.

src/gui/ocssharejob.cpp
```diff
@@ -0,0 +1,106 @@
...
+}
+
+void OcsShareJob::getShares(const QString &path)
+{
+    setVerb("GET");
+    
+    QList<QPair<QString, QString> > getParams;
+    getParams.append(qMakePair(QString::fromLatin1("path"), path));
+    setGetParams(getParams);
+
```
dragotin: I think an API that provides a method `addParam( QPair())` instead of the `setGetParams(QList())` would be nicer because it would save to set up the List and all the pairing here. Could we also use just an `addParam()`  rather than addGetParam() and `addPostParam()`?

dragotin:
Apart from my comment above (which is kind of minor but maybe still nice to have) I like this very much  :+1:

As usual my question: Do we see any chance to (unit-) test our job implementations, especially with a focus on error handling that we do? Any idea?

src/gui/ocssharejob.cpp
rullzer: Yes I like that. let me take care of that.

ckamm:
@rullzer Feel free to defer my comment about splitting this into several classes and result parsing to later. I think this is already way nicer right now! :)

rullzer:
Fixes are on their way. Basically what I noticed is that yet another layer of abstraction would be nice. So that we have share objects. 

Then the gui interacts with a share object (sets password etc). The object in the background then makes sure that the proper jobs are fired etc. That should make the gui code even cleaner and allow us to have really proper separation between the logic and the gui. But also that is future work.

rullzer:
Ok pushed. Lots of commits (but squashing will come later).
Code is a lot cleaner now. But we still have room for improvement.

@ckamm I addressed most of your comments I think. Thanks.

@ckamm @dragotin Could you have another look before I start squashing commits and merge it.

src/gui/ocsjob.h
```diff
@@ -0,0 +1,122 @@
...
+#include <QUrl>
+
+namespace OCC {
+
+/**
+ * @brief The OcsShareJob class
+ * @ingroup gui
+ *
+ * Base class for jobs that talk to the OCS endpoints on the server.
+ * All the comminication logic is handled in this class.
```
phil-davis: s/comminication/communication/
I will just add low-value comments by finding the comment typos, and along the way maybe some day be able to learn how this beast all fits together and then contribute something actually useful :)

src/gui/ocsjob.h
```diff
@@ -0,0 +1,122 @@
...
+     * List of expected statuscodes for this request
+     * A warning will be printed to the debug log if a different status code is
+     * encountered
+     *
+     * @param code Accepted status code
+     */
+    void addPassStatusCode(int code);
+
+    /**
+     * The base path for an OcsJob is always the same. But it could be that case that
```
phil-davis: "that case that" => "the case that"

src/gui/ocsjob.h
```diff
@@ -0,0 +1,122 @@
...
+     * certain operations need to append something to the URL.
+     *
+     * This functions appends the common id. so <PATH>/<ID>
```
phil-davis: s/functions/function/

src/gui/ocssharejob.h
```diff
@@ -0,0 +1,107 @@
...
+    /**
+     * Set the password of a share
+     *
+     * @param password The password of the share, if the password is empty the
+     * share will be removed
+     */
+    void setPassword(int shareId, const QString& password);
+
+    /**
+     * Void set the a share to be public upload
```
phil-davis: "the a" - pick just one of these.

src/gui/thumbnailjob.h
```diff
@@ -0,0 +1,51 @@
...
+#include "accountfwd.h"
+
+namespace OCC {
+
+/**
+ * @brief Job to fetch a thumbnail for a file
+ * @ingroup gui
+ *
+ * Job that allows fetching a preview (of 150x150 for now) of a given file.
+ * Once the job finished the jobFinished signal will be emitted.
```
phil-davis: "job has finished"

ckamm:
@rullzer The fixes so far look good. Unfortunately I won't have time for another full review next week.

rullzer:
Thanks @ckamm @phil-davis

src/gui/ocssharejob.h
```diff
@@ -0,0 +1,107 @@
...
+    /**
+     * Possible permissions
+     */
+    enum class PERMISSION : int {
+        READ = 1,
+        UPDATE = 2,
+        CREATE = 4,
+        DELETE = 8,
+        SHARE = 16,
+        ALL = 31
```
guruz: Why not READ | UPDATE | ... ?

src/gui/ocssharejob.h
```diff
@@ -0,0 +1,107 @@
...
+ * For creation, deletion and modification of shares.
+ */
+class OcsShareJob : public OcsJob {
+    Q_OBJECT
+public:
+
+    /**
+     * Support sharetypes
+     */
+    enum class SHARETYPE : int {
```
ogoffart: is enum class  supported on the minimum compiler we support?

Issue #3753, opened 2015-09-04
2.0.1 german translations incomplete


jnweiger:
2.0.1 running with all locale elements set to de_DE.utf8
We currently six text lengthy texts in english on an otherwise german page:

![s3-sync-screenshot-201-blackened](https://cloud.githubusercontent.com/assets/1108546/9684695/6dd7c556-5319-11e5-95ea-2c7ba2ca62ed.png)

jnweiger:
The line below the progress bar alternates between "Herunterladen .... (0 B von 807 kB)"
and "Syncing '....', '...' a SSASTS 262 kB/s"

The SSASTS part looks weird.

tflidd:
Instead of SSASTS, I have to boxes:
![bildschirmfoto vom 2015-09-06 10 40 31](https://cloud.githubusercontent.com/assets/8111200/9703446/ca082566-5484-11e5-803d-fc0dd52f023b.png)
Version 2.0.1 (git: 58ea249) on Ubuntu 14.04 (from SuSE repo)

MegaV0lt:
Same her in Win client 2.01
http://i.imgur.com/CKrkQLJ.png

danimo:
I have now added all missing translations. Btw: We should really check why the German translation team has run away. Looking at the contextless translation in transifex I get an idea as to why...

phil-davis:
@individual-it - your German is quite good :)

individual-it:
I've build the client from the current master on Linux.
The newly build client does not pick up the German translation when started from "client-build/bin/. Anything fancy I have to do to make it change the language?
But I can copy the "client_de.qm" to "/usr/share/owncloud/i18n/client_de.qm" and start the 2.0.1 client

The translation is better but there should be an "arrow down" (\u2193) or "arrow up" (\u2191) where in windows it says "download" or "upload"
https://github.com/owncloud/client/blob/master/src/gui/folderstatusmodel.cpp#L794

That still does not work for me.

danimo:
@individual-it That's mostly because the translation is only synced every 24 hrs

danimo:
> The translation is better but there should be an "arrow down" (\u2193) or "arrow up" (\u2191) where in windows it says "download" or "upload"

Not sure what you mean but this is correct. Windows seems to not provide these unicode characters safely, so we use upload and download as a workaround instead.

individual-it:
[...]

PR #3777, opened 2015-09-08
fix unicode issue #3753

individual-it:
fixes the unicode issue #3753

I don't know much CPP and Qt but after some google research I found this http://wiki.qt.io/Strings_and_encodings_in_Qt
it does mention trUtf8() that has done the trick for me on Linux with Gnome 3.16.
Haven't tested it on other systems


guruz:
thanks!

PR #3789, opened 2015-09-09
/doc/autoupdate.rst: "32-/64-bit" => "32-/64-bit-Windows"

hh-lohmann:
Should be stated clearly that there are no alternative 32- / 64-bit-ownCloud-Clients, but just a 32-bit-ownCloud-Client that is treated differently in 32- / 64-bit-Windows


danimo:
:+1:

PR #3797, opened 2015-09-10
Minor build etc typos

phil-davis:
Just some random stuff I noticed while looking at how this beast is put
together.


danimo:
:+1:

PR #3799, opened 2015-09-10
Minor shell integration typos

phil-davis:
all in comments or other places that have no effect on the user.


danimo:
@jturcotte pls have a look

PR #3801, opened 2015-09-10
Minor typos in client test section

phil-davis:
Note: I am not sure that nostoredpasswd even exists in the current
version. Maybe the stuff in testplan.txt is old and not relevant anyway?


danimo:
:+1:

PR #3808, opened 2015-09-11
libsync comment and message typos

phil-davis:
mostly unimportant typos in comments
note 1 place where units where said to be microseconds and I fixed the
comment to say milliseconds.
A couple of places where text of log messages is corrected.


phil-davis:
This is old. Closing in favour of #3919

Issue #3809, opened 2015-09-11
ocupdater.h meaning of comment


phil-davis:
https://github.com/owncloud/client/blob/master/src/gui/updater/ocupdater.h#L163
"The implementation does how show popups"

What is that intended to mean?
"The implementation does now show popups" == "The implementation shows popups"
or
"The implementation does not show popups"
or
???

guruz:
Try ```git blame src/gui/updater/ocupdater.h``` :-)

phil-davis:
Good thought - in a previous version it used to say:
"Does not do popups"

guruz:
Thanks, closing this in favor of your PR

PR #3818, opened 2015-09-12
OCUpdater description text

phil-davis:
Make it say something understandable. Looking at previous versions with git blame, the word "not" fits here.
Issue #3809


guruz:
Thanks! next time please try a subject like "Updater: Fix comment" so it is clear that this is about code and in which section..

PR #3882, opened 2015-09-24
csync header installation

sebasje:
This branch contains a few patches to allow building libsync against an installed csync.

I'm basically installing the csync headers that are used, and hides the _private.h from the includes.

These headers are included from libsync headers, so they need to be installed if anyone wants to use libsync from outside of this repo.

### DIFF
--- a/csync/src/CMakeLists.txt
+++ b/csync/src/CMakeLists.txt
@@ -89,9 +89,7 @@ configure_file(csync_version.h.in ${CMAKE_CURRENT_BINARY_DIR}/csync_version.h)
 set(csync_HDRS
   ${CMAKE_CURRENT_BINARY_DIR}/csync_version.h
   csync.h
-  vio/csync_vio.h
-  vio/csync_vio_method.h
-  vio/csync_vio_module.h
+  csync_private.h
 )
 
 # Statically include sqlite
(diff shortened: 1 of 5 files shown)


ogoffart:
It does not make sens to install private header.

ocsync is a full part of libsync and there should not be a need to install it separatly.

sebasje:
Yep, I've done some more work on the splitting, and it turns out using the private header in the public one is a really bad idea since it brings in a whole slew of other private stuff.

> ocsync is a full part of libsync and there should not be a need to install it separatly.

I've been wondering this, do you mean ocsync should be statically built into libsync?

As you suggested in an earlier email, I've started moving some bits out of src/gui into its own library, similar to libsync. That means that the syncing bits are in ocsync (lower level csync stuff) and in libsync (Qt API on top of that). Yet at a higher level, the non-QWidget stuff that's useful for other client (my plasma headless one, for example) go into libowncloud client, a new lib. I think I've got that fracture line between the client itself and libowncloudclient almost good now (ssl error dialog is I think the only todo item left).

I haven't figured out what exactly can move up from libowncloudsync to libowncloudclient.

My work in progress is in git@github.com:sebasje/mirall.git in the sebas/libsplit branch. It starts to look pretty good, and I can use it to build my plasmaclient outside of the owncloud-client source. Certainly needs a bunch of cleanups, but looks promising. In that branch, I've also cleaned up the csync usage to "almost not bleed into" libclient. cmake Find* modules are in my source tree, which I've split out, find it here: git://anongit.kde.org/scratch/sebas/plasma-owncloud

Anyway, this specific pull request can be discarded.

ogoffart:
> do you mean ocsync should be statically built into libsync?

Technically, yes.

But i think this cannot be done for licenses reason. csync is LGPL, but everything else can be licensed by owncloud under any license, So in order to respect csync's LGPL it has to stay a dynamic library  (at least that's how i understand the problem)

> I've started moving some bits out of src/gui into its own library, similar to libsync.  [...]

That's great news.

> (ssl error dialog is I think the only todo item left).

Ah yeah! (Would be nice if Qt could take care of that kind of stuff actually)  (maybe some room for a KDE framework?)

PR #3884, opened 2015-09-25
AccountSettings: Do not allow to expand the folder list when disconne…

### LINKED ISSUE #3860: [2.0.2] Error shown after logging out and then click on "Choose What to Sync"
### Steps to reproduce
1. Install v 2.0.2 
2. Right click on oC icon
3. Select one account (e.g user24)
4. Click on Log out
5. Right click on oC icon
6. Click on Settings...
7. Select the account user24
8. Click on Choose What to Sync

### Actual behaviour
1. Clicking choose which Sync, you can see the screen to enter the password.
2. The folders open , but you can see an error ( error loading the list of folders on the server )
screen shot 2015-09-17 at 17 23 53

### Server configuration
Desktop v ownCloud-2.0.2.2721-nightly20150917.pkg this happens on the MAC client and on the Windows too
Server v {"installed":true,"maintenance":false,"version":"8.1.3.0","versionstring":"8.1.3","edition":"Enterprise"}

dragotin:
…cted.

This is supposed to fix #3860 

If the account is not connected, the folder item can not longer be expanded. If it is expanded, it is actually de-expanded.

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -125,12 +125,19 @@ AccountSettings::AccountSettings(AccountState *accountState, QWidget *parent) :
     // Expand already on single click
     ui->_folderList->setExpandsOnDoubleClick(false);
     QObject::connect(ui->_folderList, SIGNAL(clicked(const QModelIndex &)),
-        ui->_folderList, SLOT(expand(const QModelIndex &)));
+                     this, SLOT(slotFolderListClicked(const QModelIndex&)));
 }
 
 void AccountSettings::doExpand()
 {
     ui->_folderList->expandToDepth(0);
+}
+
+void AccountSettings::slotFolderListClicked( const QModelIndex& indx )
+{
+    if( _accountState && _accountState->state() == AccountState::Connected ) {
+        ui->_folderList->expand(indx);
+    }
 }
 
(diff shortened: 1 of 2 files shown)


src/gui/accountsettings.cpp
```diff
@@ -133,6 +133,13 @@ void AccountSettings::doExpand()
     ui->_folderList->expandToDepth(0);
 }
 
+void AccountSettings::slotFolderListClicked( const QModelIndex& indx )
+{
+    if( _accountState && _accountState->state() == AccountState::Connected ) {
```
ogoffart: We probably only want that for root

ogoffart:
Looks good, but we probably only want to open root items of folder on single click.

src/gui/accountsettings.cpp
```diff
@@ -471,6 +480,18 @@ void AccountSettings::slotAccountStateChanged(int state)
         // ownCloud is not yet configured.
         showConnectionLabel( tr("No %1 connection configured.").arg(Theme::instance()->appNameGUI()) );
     }
+
+    /* Allow to expand the item if the account is connected. */
+    ui->_folderList->setItemsExpandable( state == AccountState::Connected );
```
guruz: Can one still deselect folders while being paused?

FYI @ogoffart @dragotin

PR #3899, opened 2015-09-28
fix multiarch library path for Debian/Ubuntu

sebasje:
Simplification taken from upstream cmake to make the ocsync library behave correctly on Debian and Ubuntu multiarch systems.


dragotin:
Hmm, can you elaborate why its ok to remove the other stuff? I am a bit hesitant to do that...

sebasje:
Sure. The multiarch prefix is used in any case, it doesn't depend on it being under /usr or not, any library should  be under the multiarch-correct path, so they can be co-installed with a cross-build library.

The library path that's currently being used is simply wrong, because nowadays all new-ish debian and ubuntu systems are multiarch. For other distros (non-multiarch ones) the codepath isn't being used, so has no effect.

dragotin:
Thanks, sounds reasonable.

PR #3910, opened 2015-10-02
Adding swedish translation

grasberg:
Just adding a few more translation to sv (swedish)


danimo:
Sorry, but we can only accept translations via transifex, as all contributions to github will be overwritten. Can you sign up there?

danimo:
https://transifex.com/projects/p/owncloud/

Issue #3879, opened 2015-09-24
[2.0.2] Deselect to the option "Allow apps to use the Share API" on the Server, and Share one folder on the Desktop you can see "OCS API error code:403"


Dianafg76:
### Steps to reproduce
1. Install v 2.0.2
2. Go to Server Client
3. Sign in as Admin user
4. Go to Admin
5. On Sharing option, deselect to the option "Allow apps to use the Share API"
![screen shot 2015-09-24 at 14 01 32](https://cloud.githubusercontent.com/assets/12032058/10072818/d54080ec-62c4-11e5-96f4-06c292e10c0e.png)

6. Go to the Desktop Client
7. Configurate one account
8.  Open folder (....../....)
9. Create new folder (e.g call Example1)
10. Right click on the folder Example1
11. Click on "Share with oC"
![screen shot 2015-09-24 at 14 05 54](https://cloud.githubusercontent.com/assets/12032058/10072918/6cbc9b86-62c5-11e5-9ad3-423ce757a94a.png)

12. If you introduce the Password and press enter

### Actual behaviour
You can see the error: OCS API error code: 403
![screen shot 2015-09-24 at 14 07 54](https://cloud.githubusercontent.com/assets/12032058/10072966/addcf12e-62c5-11e5-86e4-7123f578c6de.png)

### Server configuration
Desktop v ownCloud-2.0.2.2740-nightly20150924.pkg
Server v {"installed":true,"maintenance":false,"version":"8.1.3.0","versionstring":"8.1.3","edition":"Enterprise"}

dragotin:
@rullzer do you have time to look into that?

rullzer:
Yes will do.

rullzer:
Ok got it we need some proper server support with this. Let me fix that first.

dragotin:
Awesome, thanks :-)

rullzer:
Server side PR in #19362.
Now lets see what we need to do in the client part :)

rullzer:
Start in #3923

rullzer:
Ok the client side info in now in master (#3923). Now we need all the plugins to be properly adjusted.

ckamm:
@rullzer I've tested this with 2.1 and what I get now is a message: "The file can not be shared because it was shared without sharing permission." That seems a bit confusing? :)

Dianafg76:
[...]

PR #3923, opened 2015-10-06
Provide information about share availablity for files

rullzer:
For #3879

The socketAPI `SHARE_STATUS` allows to retrieve the share status for the current file. So can it be shared or not. And if it can how (UserGroup, Link).

We will need additional fixes in the shell_intergration stuff.

CC: @dragotin


guruz:
I'm not working on oC these days, maybe @danimo @dragotin can have a look.

ckamm:
@rullzer I suggested some adjustments to the replies to SHARE_STATUS inline.

PR #3931, opened 2015-10-08
Update "..." menu documentation for 2.0

phil-davis:
This menu in 2.0.* now has 4 options.
The small screenshot needs updating also. @carlaschroder it is probably best if you take a new screen shot from whatever client version you have been using for screen shots. That way they will all have a consistent look.

### DIFF
--- a/doc/navigating.rst
+++ b/doc/navigating.rst
@@ -92,18 +92,27 @@ have the following features:
   removed synchronization on an account (see **Remove Sync** below).
 
 The little button with three dots that sits to the right of the sync status bar 
-offers three additional options:
+offers four additional options:
 
 * Open Folder
-* Pause Sync
+* Choose What to Sync
+* Pause Sync / Resume Sync
 * Remove Sync
 
-**Pause Sync** pauses sync operations without making any changes to your account.
+**Open Folder** opens a file explorer window displaying the client-side folder
+that is being synced.
 
-**Remove Sync** suspends synchronization without removing the account, and it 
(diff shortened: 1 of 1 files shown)


phil-davis:
Note: An updated screenshot is needed of this "..." menu, hopefully can be supplied by @carlaschroder

carlaschroder:
@phil-davis I'm looking at the 2.0.1 (build 5446) release, and there are only three options, as in the original text. What version are you looking at?

phil-davis:
I am on a 2.0.2 version. I guess the 4 menu options are in 2.0.2-beta1? I think it came in between 2.0.1 and 2.0.2.
If that is correct, then this documentation change needs to be "released" along with 2.0.2 - I am not sure how if that is possible for UI features that actually change with point releases of the software.

carlaschroder:
I'm going to revert this PR & save your changes for when 2.0.2 is released. It's available from the Testing repo, but we maintain only stable release manuals. Thanks and wait until we catch up :)

phil-davis:
No problem - I had seen 2.0.2-beta1 come out a while ago and assumed that a 2.0.2 release was imminent, so proposed this doc update. It seems that 2.0.2 has stopped at beta1 for the moment.

carlaschroder:
yay @phil-davis 2.0.2 is released! I shall restore this update.

carlaschroder:
https://github.com/owncloud/client/pull/3982

phil-davis:
:+1:

PR #3937, opened 2015-10-10
Increased fileID size to 36 bytes

labkode:
Right now the fileID is created using the SQL DB as a central auto-incremental fileID generator.
Some storages can provide their own FileID or use de-centralised fileID generators like UUIDs or checksums.

The sync client is limited to 21 byte size fileID, so when using UUIDs in string format (36 bytes), the FileID is not respected and the client cannot track remote moves.

See https://github.com/owncloud/client/blob/524aa507e6ac0331a6cd909fa8bed297ae0f3bfb/csync/src/vio/csync_vio_file_stat.c#L74

### DIFF
--- a/csync/src/csync.h
+++ b/csync/src/csync.h
@@ -144,7 +144,7 @@ enum csync_ftw_type_e {
 };
 
 
-#define FILE_ID_BUF_SIZE 21
+#define FILE_ID_BUF_SIZE 36
 
 // currently specified at https://github.com/owncloud/core/issues/8322 are 9 to 10
 #define REMOTE_PERM_BUF_SIZE 15


dragotin:
I am very reluctant to this change. We need to keep the memory footprint small, please try to limit the fileid to the given 21 byte.

labkode:
Got it!

guruz:
@dragotin how can it track the moves then?

dragotin:
@labkode is there any possibility for you to deal with 21 bytes? I know it sounds picky if I discuss about 15 bytes extra, but that is  for every entry. Maybe we expose it as a compile option?

labkode:
@dragotin the increase of 15 bytes is really a low increase in memory (with 1M files it is just an increase of ~15MB). 

When the new server storage interface comes out, there will storage implementations that won't rely on an incremental number for fileIDs and they will find the same problem. 

As a workaround I managed to reduce it to 21 bytes using Inodes instead of UUIDs, but this is not a great solution. 

I tried to use base64 and asci85 to do the binary-to-text encoding of the UUID but it is still greater than 21 bytes.

It would be great if this could be a compile option and give some information for developers about this limit without having to check the sync client error logs. 

Maybe in the Sync Protocol Documentation ?

guruz:
Thanks, cherry-picked into 2.0 instead of merge into master :)
4d1d6dff3b535222a0aea963d68245fb7e170ac8

In the longer run we really need to look into not having the csync data structures that are not converted around so often..

dragotin:
ok. @labkode guruz and olivier convinced me finally :-) We will squeeze out the bytes again at other places.

guruz:
4df828d0b030ee7ca4a62d7c18e9e9d526ee6f76

Issue #3735, opened 2015-09-02
File Transport Checksums Between Server and Desktop


MTRichards:
As an ownCloud admin, I want ownCloud to ensure that files transported from the desktop client to the server are not corrupted in transit so that we can catch bugs and avoid corrupting customer data

Acceptance Criteria:
- Desktop sync client takes a checksum of files on the desktop when they are discovered and stores the checksum in the database
- When the file is transmitted to the ownCloud server, the checksum is also transmitted
- The server checksums the file when it arrives at the server, compares it to the checksum transmitted by the client previously
- If the checksums match, a success return code is sent
- If the chechsums fail to match, the file has been corrupted and the desktop skips that file. The file will be retransmitted on the next upload.
- If the file is corrupted in transit, the file transfer log on the desktop reflects that this file was corrupted in transit, and that it will be retried on the next sync run 
Note: this links to server issue XXX

ckamm:
What needs to be done in the client?
* Determine whether the server accepts checksums, possibly via capabilities. Currently checksumming needs to be explicitly enabled in the config file.
* Store checksums in the database. (for performance optimization only, I guess?)
* Check that checksum errors are handled nicely.

ckamm:
We also need to make sure getting an empty checksum header from the server during downloads isn't considered a validation failure - currently that seems to be the case. I wonder why b05ca526a46c8a306ccfb4aca0d5cf94881ba14f did that -- @dragotin?

dragotin:
@ckamm the patch you're pointing to checks the expected checksum **type** as in Adler32 or SHA or so. If that is empty, we do not have checksumming and assume validated() - which means all is ok.

ckamm:
@dragotin Yes, but why would the checksum type configured in the config file matter when one validates a checksum the server sends? 

And shouldn't an empty checksum header still yield `validated`? It sounds like only files that were uploaded by new clients will have checksums when using the owncloud server.

ckamm:
@moscicki I'm touching the checksumming code and wonder: Do you use `Adler32` or `ADLER32`? What does the server accept and what does it send? I'd prefer to just use one spelling and there's a small chance you're requiring one for uploads and the other for downloads...

moscicki:
We treat the request header name as case-insensitive. So all these are accepted: Adler32, adler32, ADLER32

Response given by the server: Adler32

ckamm:
@moscicki Great, then the client will standardize on `Adler32`.

[...]

PR #3951, opened 2015-10-15
Checksums work for #3735

ckamm:
Store the checksums in the db during up and downloads.

### DIFF
--- a/src/libsync/capabilities.cpp
+++ b/src/libsync/capabilities.cpp
@@ -65,14 +65,14 @@ bool Capabilities::shareResharing() const
     return _capabilities["files_sharing"].toMap()["resharing"].toBool();
 }
 
-QList<QByteArray> Capabilities::supportedChecksumTypesRaw() const
+QList<QByteArray> Capabilities::supportedChecksumTypesAdvertised() const
 {
     return QList<QByteArray>();
 }
 
 QList<QByteArray> Capabilities::supportedChecksumTypes() const
 {
-    auto list = supportedChecksumTypesRaw();
+    auto list = supportedChecksumTypesAdvertised();
     QByteArray cfgType = ConfigFile().transmissionChecksum().toLatin1();
     if (!cfgType.isEmpty()) {
         list.prepend(cfgType);
--- a/src/libsync/capabilities.cpp
+++ b/src/libsync/capabilities.cpp
@@ -13,6 +13,8 @@
 
 #include "capabilities.h"
 
+#include "configfile.h"
+
 #include <QVariantMap>
 
 namespace OCC {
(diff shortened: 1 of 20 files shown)


src/libsync/capabilities.cpp
```diff
@@ -63,9 +65,28 @@ bool Capabilities::shareResharing() const
...
+QByteArray Capabilities::preferredChecksumType() const
+{
+    auto list = supportedChecksumTypes();
+    if (list.isEmpty()) {
+        return QByteArray();
+    }
+    return list.first();
 }
 
 }
```
dragotin: I assume the function `supportedChecksumTypesRaw()` is supposed to fetch the supported checksums from the server capabilities, but that is not yet implemented, right? Why not name it like that? 'raw' is pretty generic imo.

Also, if we have a list of supported ones from the server, the prepend can lead to duplicate entries, right?

src/libsync/propagateupload.cpp
```diff
@@ -209,30 +209,45 @@ void PropagateUploadFileQNAM::start()
...
-    if (validator->checksumType().isEmpty()) {
-        QStringList checksumTypes = _propagator->account()->capabilities().supportedChecksumTypes();
-        if (!checksumTypes.isEmpty()) {
-            // TODO: We might want to prefer some types over others instead
-            // of choosing the first.
-            validator->setChecksumType(checksumTypes.first());
+    auto supportedChecksumTypes = _propagator->account()->capabilities().supportedChecksumTypes();
+
+    // If we already have a checksum header and the checksum type is supported
+    // by the server, we keep that - otherwise recompute.
```
dragotin: minor: maybe add a comment where we get could possibly have the checksum header from

src/libsync/propagateupload.cpp
```diff
@@ -209,30 +209,45 @@ void PropagateUploadFileQNAM::start()
...
+    if (!_item->_checksumHeader.isEmpty()) {
+        QByteArray checksumType;
+        QByteArray checksum;
+        if (parseChecksumHeader(_item->_checksumHeader, &checksumType, &checksum)
+                && supportedChecksumTypes.contains(checksumType)) {
```
dragotin: I would love to see that split into two lines to make it more clear that the `contains()` relies on having `checksumType` set in the call to `parseChecksumHeader()`.

dragotin:
Apart from my  minor comments: :+1:

src/libsync/capabilities.cpp
ckamm: Yes. This is mainly intended for testing and will definitely need to be revised as the server API comes into existence and as we figure out how exactly we want to use the checksums.

dragotin:
:+1:

Issue #3677, opened 2015-08-24
[2.0 nightly] remotePollInterval no effect on connectivity-polls?


tflidd:
OS: Win 7 home premium 64 bit
client: 2.0.0-nightly20150824 (build 5421)
only /music-folder is synced.
OC 8.1.1 (SuSE repo on debian 8.1.1, without SSL, without encryption)
my testpilotcloud.cfg (was changed when the client was not running):
````
[General]
optionalDesktopNotifications=true

[ActivityListHeader]
geometry=@ByteArray(\0\0\0\xff\0\0\0\0\0\0\0\x1\0\0\0\x1\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\x2\xbc\0\0\0\x5\x1\0\0\x1\0\0\0\0\0\0\0\0\0\0\0\0\x64\xff\xff\xff\xff\0\0\0\x81\0\0\0\0\0\0\0\x5\0\0\0\x64\0\0\0\x1\0\0\0\0\0\0\0\xb4\0\0\0\x1\0\0\0\0\0\0\0\x64\0\0\0\x1\0\0\0\0\0\0\0\xdc\0\0\0\x1\0\0\0\0\0\0\0\x64\0\0\0\x1\0\0\0\0\0\0\x3\xe8)

[Settings]
geometry=@ByteArray(\x1\xd9\xd0\xcb\0\x2\0\0\0\0\x4s\0\0\0^\0\0\a7\0\0\x3\x5\0\0\x4{\0\0\0|\0\0\a/\0\0\x2\xfd\0\0\0\0\0\0\0\0\a\x80)

[testpilotcloud]
[...]

PR #3964, opened 2015-10-19
AccountState: Avoid ConnectionCheck if ETag job was just done.

dragotin:
This patch lets a successful etag job check mark a timestamp.
If next time a connection check is requested, it is checked if
the last ETag happened within the last 30 seconds and if so the
connection check can be checked.

This way we avoid half of the PROPFINDs if all goes well.

@ckamm @ogoffart @guruz please check.

### DIFF
--- a/src/gui/accountstate.h
+++ b/src/gui/accountstate.h
@@ -95,6 +95,14 @@ public:
      */
     QString shortDisplayNameForSettings(int width = 0) const;
 
+    /** Mark the timestamp when the last successful ETag check happened for
+     *  this account.
+     *  The checkConnectivity() method uses the timestamp to save a call to
+     *  the server to validate the connection if the last successful etag job
+     *  is not so lang away.
+     */
+    void tagLastSuccessfullETagRequest();
+
 private:
     void setState(State state);
 
@@ -114,6 +122,7 @@ private:
     QStringList _connectionErrors;
     bool _waitingForNewCredentials;
     CredentialFetchMode _credentialsFetchMode;
+    QElapsedTimer _timeSinceLastETagCheck;
     QPointer<ConnectionValidator> _connectionValidator;
 };
 
--- a/src/gui/accountstate.cpp
+++ b/src/gui/accountstate.cpp
@@ -16,6 +16,7 @@
 #include "account.h"
 #include "creds/abstractcredentials.h"
 #include "logger.h"
+#include "configfile.h"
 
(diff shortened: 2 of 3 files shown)


guruz:
(haven't tested this theory)

With multiple sync folders/accounts the ETag checks will be done sequentially (at least that's how it was before multi-account)

Having a check with <= 30 sec will therefore mostly be false since those checks were sequential and could be longer than 30 sec ago

guruz:
Please mention the related issue ID in the commit message

https://github.com/owncloud/client/issues/3677

guruz:
Ah, and the 30 sec should be instead related to remotePollInterval ? (Plus the comment above I mentioned with regards to multiple accounts/folders)

dragotin:
@guruz yes, exactly my thought, it should be the same as remotePollInterval. Will follow up.

ogoffart:
I think the right fix is not to do the connection check chen the account is online.   The etag check will make the connection drop.

src/gui/accountstate.h
```diff
@@ -95,6 +95,14 @@ class AccountState : public QObject {
      */
     QString shortDisplayNameForSettings(int width = 0) const;
 
+    /** Mark the timestamp when the last successful ETag check happened for
+     *  this account.
+     *  The checkConnectivity() method uses the timestamp to save a call to
+     *  the server to validate the connection if the last successful etag job
+     *  is not so lang away.
```
phil-davis: "was not too far in the past." ?

guruz:
btw, we also must not break
```
void AccountState::checkConnectivity(CredentialFetchMode credentialsFetchMode)
...
#ifdef Q_OS_WIN
        // There seems to be a bug in Qt on Windows where QNAM sometimes stops
        // working correctly after the computer woke up from sleep. See #2895 #2899
        // and #2973.
        // As an attempted workaround, reset the QNAM regularly if the account is
        // disconnected.
        account()->resetNetworkAccessManager();
```

guruz:
@dragotin You had asked me for an explanation on https://github.com/owncloud/client/pull/3964#issuecomment-149203019

What I meant is that ```FolderMan::slotEtagPollTimerTimeout()``` handles the ```ETagJob``` for all accounts and it only schedules one job at the same time. So if you have two accounts then your 30 sec check would not work.

I'll have a look at this and possibly fix this here.

guruz:
Cherry picked into 2.1

PR #3975, opened 2015-10-20
Pro-forma PR for @sebasje's libsplit branch against master

danimo:


src/libclient/CMakeLists.txt
```diff
@@ -0,0 +1,175 @@
+project(libowncloudclient)
```
danimo: lib must not use the name owncloud (due to theming), let's rather use libui or something more descriptive (analogous to `libsync`).

src/libclient/CMakeLists.txt
```diff
@@ -0,0 +1,175 @@
...
+#
+if ( APPLE )
+    list(APPEND OS_SPECIFIC_LINK_LIBRARIES
+         /System/Library/Frameworks/CoreServices.framework
+         /System/Library/Frameworks/Foundation.framework
+         /System/Library/Frameworks/AppKit.framework
+    )
+endif()
+
+set(owncloudclientlib_NAME "owncloudclient")
```
danimo: see above.

danimo:
What is your intention with breaking up those libraries? Being able to have a private fork (i.e. by using the theming option) is fine, but at least I would still be reluctant to commit to a stable API.

src/libclient/CMakeLists.txt
```diff
@@ -0,0 +1,175 @@
...
+#     if(NEON_WITH_LFS)
+#         add_definitions(-DNE_LFS)
+#     endif()
+# endif()
+#
+# if(ZLIB_FOUND)
+#     list(APPEND libowncloudclient_LINK_TARGETS ${ZLIB_LIBRARIES})
+#     include_directories(${ZLIB_INCLUDE_DIRS})
+# endif(ZLIB_FOUND)
+#
```
dragotin: Why are these checks commented?

dragotin:
I think the directory name "libclient" is not good, not specific enough. Maybe something like libsyncenvironment (yes, I know, too long...) ?

Also, we must not forget to verifiy the RPATH setting since we install that library to a application directory in `/usr/lib*`, right? 

And finally: We include `csync_private.h` too much. It was intended to be... well private :-) We should reconsider to rename it and/or reinvestigate which parts of it really need to be public.

src/libclient/CMakeLists.txt
sebasje: libsyncui? libui is too generic, IMO.

src/libclient/CMakeLists.txt
sebasje: They're not needed in this library, I might as well remove them there.

sebasje:
From an earlier email exchange with Klaas and Olivier:

My goal is to use as much as possible of the existing 
owncloud client's code and put a Plasma UI (and perhaps some deeper 
integration bits) on top of it.

My approach is to create a headless synching client that runs in the 
background (autostarted in Plasma) and have a Plasmoid and a Systemsettings 
module communicate with it. This allows us to put different UIs for different 
devices and make owncloud synching behave more like a service provided by the 
system rather than an application in the traditional sense.

src/libclient/CMakeLists.txt
ogoffart: Yes, they can be removed.

ogoffart:
the include csync_private can hopefully disapear in the future.  Ideally, the SyncEngine would create the csync_context on demand.  
The optimisation to keep the csync_context between sync is not worth it anymore since we do not reload the owncloud plugin all the time and such.
But for now i think it's fine to include it

src/libclient/CMakeLists.txt
ogoffart: what's wront with the `libowncloudclient` name?  It is an internal name only so branding should not care.  Or am i missing something.

dragotin: I agree with @sebasje, `libsyncui` is cool. We really want to avoid the word "owncloud" here.

ogoffart: how about  `libsyncclient`?
(maybe we can rename the other lib in `libsyncengine`

ogoffart:
@sebasje  will you have time to address the issue that @danimo raised?

I think we should merge this sooner than later if we want it.

sebasje:
Yes, I'll try to make some time shortly.

It would be good if someone could test my changes on WIndows and Mac OS, I don't have tese OSes available.

ogoffart:
ping?

sebasje:
I haven't forgotten about it, it just slipped down my list a bit with some other projects going on. I'll try to update my branch shortly and will let you have another go at it.

Thanks for the nudge. :)

danimo:
This is obsolete now

PR #3982, opened 2015-10-21
update 3-dot menu

carlaschroder:
My windows laptop is packed in a box somewheres, so anyone who wants a prettier screenshot can make one. Please review @phil-davis @danimo @dragotin


danimo:
tnx!

PR #4017, opened 2015-10-28
Exclude .nfs temporary files by default

eltrai:
Whenever a file is deleted while being still opened in another host in a NFS, it will create temporary .nfsNNNNN files, so that the file still remains accessible for as long as the process needs it.
There is no logic in syncing those files in owncloud, so they should be excluded by default.


guruz:
Thanks, because of a merge conflict I manually added this 3ee8beb8a34f7ffbc6bfc8afe171747a6069094a

PR #4033, opened 2015-10-29
Add share manager and the share objects

rullzer:
Step closer to https://github.com/owncloud/client/issues/3737

Pretty big PR (I'm sorry... I'll promise to do better).

Basically this introduces the abstraction I want. And which should help significantly in making the user/group sharing easier to implement.

There is now a ShareManager, this is the entry point to do anything with shares. Be it creating, or fetching.

Then there is the Share class which represents a generic share and there is the LinkShare class. Since link shares are just like regular shares but slightly different :P. Share objects allow for interaction with the object. E.g. setting the password etc. 

The sharedialog code is converted to use the new code and still works for me so far. And is a lot simpler.

TODO:
- [ ] Triplle check share dialog functionality
- [x] Check share dialog functionality against older servers
- [ ] More comments
- [ ] Maybe some cleanup?
- [x] Connect error reporting from manager/share

@guruz @dragotin @ckamm please have a look. More commits will follow. But since it kind of works now

### DIFF
--- a/src/gui/ocsjob.cpp
+++ b/src/gui/ocsjob.cpp
@@ -42,9 +42,9 @@ void OcsJob::addPassStatusCode(int code)
     _passStatusCodes.append(code);
 }
 
-void OcsJob::appendPath(int id)
+void OcsJob::appendPath(const QString &id)
 {
-    setPath(path() + QString("/%1").arg(id));
+    setPath(path() + "/" + id);
 }
 
 void OcsJob::start()
--- a/src/gui/ocsjob.cpp
+++ b/src/gui/ocsjob.cpp
@@ -105,9 +105,10 @@ bool OcsJob::finished()
                  << Account::concatUrlPath(account()->url(), path())
                  << _params
                  << "has unexpected status code:" << statusCode << replyData;
+        emit ocsError(statusCode, message);
+    } else {
+        emit jobFinished(json);
     }
-
-    emit jobFinished(json);
     deleteLater();
     return true;
 }
--- a/src/gui/ocsjob.cpp
+++ b/src/gui/ocsjob.cpp
@@ -108,6 +108,7 @@ bool OcsJob::finished()
     }
 
     emit jobFinished(json);
+    deleteLater();
     return true;
 }
 
--- a/src/gui/ocsjob.cpp
+++ b/src/gui/ocsjob.cpp
@@ -44,7 +44,7 @@ void OcsJob::addPassStatusCode(int code)
 
 void OcsJob::appendPath(const QString &id)
 {
-    setPath(path() + "/" + id);
+    setPath(path() + QLatin1Char('/') + id);
 }
 
 void OcsJob::start()
--- a/src/gui/ocssharejob.h
+++ b/src/gui/ocssharejob.h
@@ -15,6 +15,7 @@
 #define OCSSHAREJOB_H
 
 #include "ocsjob.h"
+#include "share.h"
 #include <QVector>
 #include <QList>
 #include <QPair>
@@ -31,25 +32,6 @@ namespace OCC {
 class OcsShareJob : public OcsJob {
     Q_OBJECT
 public:
-
-    /**
-     * Support sharetypes
-     */
-    enum ShareType : int {
-        Link = 3
-    };
-
-    /**
-     * Possible permissions
-     */
-    enum Permission : int {
-        Read = 1,
-        Update = 2,
-        Create = 4,
-        Delete = 8,
-        Share = 16,
-        All = 31
-    };
 
     /**
      * Constructor for new shares or listing of shares
@@ -99,7 +81,7 @@ public:
      * @param password Optionally a password for the share
      * @param date Optionally an expire date for the share
      */
-    void createShare(const QString& path, ShareType shareType, const QString& password = "", const QDate& date = QDate());
+    void createShare(const QString& path, Share::ShareType shareType, const QString& password = "", const QDate& date = QDate());
 
 signals:
     /**
--- a/src/gui/ocssharejob.h
+++ b/src/gui/ocssharejob.h
@@ -35,14 +35,14 @@ public:
     /**
      * Support sharetypes
      */
-    enum class ShareType : int {
+    enum ShareType : int {
         Link = 3
     };
 
     /**
      * Possible permissions
      */
-    enum class Permission : int {
+    enum Permission : int {
         Read = 1,
         Update = 2,
         Create = 4,
@@ -66,7 +66,7 @@ public:
     /**
      * Delete the current Share
      */
-    void deleteShare(int shareId);
+    void deleteShare(const QString &shareId);
 
     /**
      * Set the expiration date of a share
@@ -74,7 +74,7 @@ public:
      * @param date The expire date, if this date is invalid the expire date
      * will be removed
      */
-    void setExpireDate(int shareId, const QDate& date);
+    void setExpireDate(const QString &shareId, const QDate& date);
 
     /**
      * Set the password of a share
@@ -82,14 +82,14 @@ public:
      * @param password The password of the share, if the password is empty the
      * share will be removed
      */
-    void setPassword(int shareId, const QString& password);
+    void setPassword(const QString &shareId, const QString& password);
 
     /**
      * Void set the share to be public upload
      * 
      * @param publicUpload Set or remove public upload
      */
-    void setPublicUpload(int shareId, bool publicUpload);
+    void setPublicUpload(const QString &shareId, bool publicUpload);
 
     /**
      * Create a new share
@@ -100,6 +100,24 @@ public:
      * @param date Optionally an expire date for the share
      */
     void createShare(const QString& path, ShareType shareType, const QString& password = "", const QDate& date = QDate());
+
+signals:
+    /**
+     * Result of the OCS request
+     * The value parameter is only set if this was a put request.
+     * e.g. if we set the password to 'foo' the QVariant will hold a QString with 'foo'.
+     * This is needed so we can update the share objects properly
+     *
+     * @param reply The reply
+     * @param value To what did we set a varialble (if we set any).
+     */
+    void shareJobFinished(QVariantMap reply, QVariant value);
+
+private slots:
+    void jobDone(QVariantMap reply);
+
+private:
+    QVariant _value;
 };
 
 }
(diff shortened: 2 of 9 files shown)


src/gui/ocssharejob.h
```diff
@@ -100,6 +100,24 @@ class OcsShareJob : public OcsJob {
...
+
+signals:
+    /**
+     * Result of the OCS request
+     * The value parameter is only set if this was a put request.
+     * e.g. if we set the password to 'foo' the QVariant will hold a QString with 'foo'.
+     * This is needed so we can update the share objects properly
+     *
+     * @param reply The reply
+     * @param value To what did we set a varialble (if we set any).
```
phil-davis: variable

src/gui/share.cpp
```diff
@@ -0,0 +1,268 @@
...
+}
+
+void ShareManager::slotLinkShareCreated(const QVariantMap &reply)
+{
+    QString message;
+    int code = OcsShareJob::getJsonReturnCode(reply, message);
+
+    /*
+     * Before we had decent sharing capabilities on the server a 403 "generally"
+     * ment that a share was password protected
```
phil-davis: meant

src/gui/share.h
```diff
@@ -0,0 +1,205 @@
...
+     * On success the permissionsSet signal is emitted
+     * In case of a server error the serverError signal is emitted.
+     */
+    void setPermissions(int permissions);
+
+    /**
+     * Deletes a share
+     *
+     * On success the shareDeleted signal is emitted
+     * In case of a server rror the serverError signal is emitted.
```
phil-davis: error

src/gui/share.cpp
```diff
@@ -0,0 +1,268 @@
...
+}
+
+LinkShare *ShareManager::parseLinkShare(const QVariantMap &data) {
+    QUrl url;
+    const QString versionString = _account->serverVersion();
+
+    // From ownCloud server 8.2 the url field is always set for public shares
+    if (data.contains("url")) {
+        url = QUrl(data.value("url").toString());
+    } else if (versionString.contains('.') && versionString.split('.')[0].toInt() >= 8) {
```
ogoffart: you can use _account->serverVersionInt()

rullzer: awesome. will fix :)

src/gui/share.h
```diff
@@ -0,0 +1,205 @@
...
+
+namespace OCC {
+
+class Share : public QObject {
+    Q_OBJECT
+
+public:
+
+    /*
+     * Constructor for link shares
```
ckamm: I don't think this is for link shares

src/gui/share.h
```diff
@@ -0,0 +1,205 @@
...
+     */
+    explicit Share(AccountPtr account,
+                   const QString& id,
+                   const QString& path,
+                   int shareType,
+                   int permissions,
```
ckamm: It'd be nice if these had the ShareType and Permission types. Consider using Q_DECLARE_FLAGS on the enums (at least Permissions).

src/gui/share.h
```diff
@@ -0,0 +1,205 @@
...
+                   QObject *parent = NULL);
+
+    /*
+     * Get the id
+     */
+    const QString getId();
```
ckamm: const? also the other getters

src/gui/share.cpp
```diff
@@ -0,0 +1,268 @@
...
+
+void Share::slotDeleted(const QVariantMap &reply)
+{
+    QString message;
+    int code = OcsShareJob::getJsonReturnCode(reply, message);
+    if (code != 100) {
+        //emit error!
+    }
+
+    emit shareDeleted();
```
ckamm: I don't remember: Does the OcsShareJob delete itself with deleteLater once done? If not, please check whether these jobs are cleaned up when done.

src/gui/share.cpp
```diff
@@ -0,0 +1,268 @@
...
+        //emit error!
+    }
+
+    auto tmpShares = reply.value("ocs").toMap().value("data").toList();
+    const QString versionString = _account->serverVersion();
+    qDebug() << Q_FUNC_INFO << versionString << "Fetched" << tmpShares.count() << "shares";
+
+    QList<QSharedPointer<Share>> shares;
+
+    Q_FOREACH(auto share, tmpShares) {
```
ckamm: Why not 'foreach'?

src/gui/share.cpp
```diff
@@ -0,0 +1,268 @@
...
+
+    return new LinkShare(_account,
+                         data.value("id").toString(),
+                         data.value("path").toString(),
+                         data.value("share_type").toInt(),
+                         data.value("permissions").toInt(),
+                         data.value("share_with").isValid(),
+                         url,
+                         expireDate,
+                         this);
```
ckamm: Hmm. You put pointers to these objects into QSharedPointer *and* parent them to this. Who owns these objects?

src/gui/sharedialog.cpp
```diff
@@ -168,6 +169,15 @@ ShareDialog::ShareDialog(AccountPtr account, const QString &sharePath, const QSt
             _ui->checkBox_editing->setEnabled(false);
         }
     }
+
+    /*
+     * Create the share manager and connect it properly
+     */
+    _manager = QSharedPointer<ShareManager>(new ShareManager(_account, this));
```
ckamm: Why's this a shared pointer?

src/gui/sharedialog.cpp
```diff
@@ -168,6 +169,15 @@ ShareDialog::ShareDialog(AccountPtr account, const QString &sharePath, const QSt
...
+
+    connect(_manager.data(), SIGNAL(sharesFetched(QList<QSharedPointer<Share>>)), this, SLOT(slotSharesFetched(QList<QSharedPointer<Share>>)));
+    connect(_manager.data(), SIGNAL(linkShareCreated(const QSharedPointer<LinkShare>)), this, SLOT(slotCreateShareFetched(const QSharedPointer<LinkShare>)));
+    connect(_manager.data(), SIGNAL(linkShareRequiresPassword()), this, SLOT(slotCreateShareRequiresPassword()));
```
ckamm: btw: inside an object, you can omit the third argument if you want to connect to `this`

src/gui/share.cpp
ogoffart: also,  `const auto &`  to avoid a copy

src/gui/share.cpp
```diff
@@ -0,0 +1,268 @@
...
+        }
+
+        shares.append(QSharedPointer<Share>(newShare));    
+    }
+
+    qDebug() << Q_FUNC_INFO << "Sending " << shares.count() << "shares";
+    emit sharesFetched(shares);
+}
+
+LinkShare *ShareManager::parseLinkShare(const QVariantMap &data) {
```
ogoffart: You should return a QSharedPointer here to make the owncership explicit.

src/gui/share.cpp
ogoffart: right, there should not be a parent

src/gui/share.h
```diff
@@ -0,0 +1,205 @@
...
+     *
+     * @param path The path to get the shares for relative to the users folder on the server
+     *
+     * On success the sharesFetched signal is emitted
+     * In case of a server error the serverError signal is emitted
+     */
+    void fetchShares(const QString& path);
+
+signals:
+    void linkShareCreated(const QSharedPointer<LinkShare> share);
```
ogoffart: missing &

src/gui/share.h
```diff
@@ -0,0 +1,205 @@
...
+    void linkShareRequiresPassword();
+    void sharesFetched(QList<QSharedPointer<Share>>);
```
ogoffart: should be a const &

Also, QVector is more efficient than QList (but that should not matter much)

src/gui/sharedialog.cpp
ogoffart: right this goes out of scope immediatly so this looks wrong.
Also, QSharedPointer::create would avoid one alocation.

src/gui/share.h
rullzer: Yes my mistake... was late yesterday ;)

src/gui/share.cpp
rullzer: nope we missed that last time. Added.

src/gui/sharedialog.cpp
rullzer: yeah no need for a sharedpointer indeed.

rullzer: @ogoffart QSharedPointer::create is not available on Qt-4 right?

rullzer:
Addressed your comments :) thanks. Already looking better.

src/gui/ocsjob.cpp
```diff
@@ -42,9 +42,9 @@ void OcsJob::addPassStatusCode(int code)
     _passStatusCodes.append(code);
 }
 
-void OcsJob::appendPath(int id)
+void OcsJob::appendPath(const QString &id)
 {
-    setPath(path() + QString("/%1").arg(id));
+    setPath(path() + "/" + id);
```
dragotin: I think you should use `QLatin1Char('/')` here rather than just `"/"`

src/gui/share.cpp
```diff
@@ -0,0 +1,270 @@
+#include "share.h"
```
dragotin: Please do not forget a proper license header...

dragotin:
Nice PR. 
One last thing: Do you have some kind of error handling with the jobs, in case they fail? I did not see that.
Also, please add proper headers to the files.

src/gui/share.cpp
rullzer: fixed

rullzer:
@dragotin no error handling still needs to be done. I plan to do that very simple for now (just like we had it before, show the status code in the footer). But now that we are properly abstracted away we can actually work on proper error handling and reporting in future PR's.

rullzer:
Ok now there is error reporting. Much in the same way it was before. So no improvements there yet. But it seems to be working again.

src/gui/share.h
rullzer: Totally agree. They are back now!

rullzer:
Now using QFlags to have better type checking in the code. They are less strict than enum classes but I think they fit better in the Qt coding way here.

rullzer:
Please let me know if there are further comments. Else I'll merge somewhere tomorrow.

PR #4038, opened 2015-10-29
Set the library so version to the current major version.

dragotin:
This names the libraries we ship properly to version 2.


VERSION.cmake
```diff
@@ -1,7 +1,7 @@
 set( MIRALL_VERSION_MAJOR 2 )
 set( MIRALL_VERSION_MINOR 1 )
 set( MIRALL_VERSION_PATCH 0 )
-set( MIRALL_SOVERSION 0 )
+set( MIRALL_SOVERSION 2 )
```
guruz: Why not depend this on MIRALL_VERSION_MAJOR ?

ogoffart:
Does it make sens? We don't maintain binary compatibility, so we should update the major so version at every release. Or keep 0

guruz:
OK, you want to keep 0 because then we don't have side effects like filling the hd when using "make install"?
Makes sense..

ogoffart:
I want to keep 0 because the .so version is not the library version, it's the interface version.  Since we don't maintain binary compatibility, we should keep it to 0   (best would be not to have any version actually)

PR #4052, opened 2015-10-31
Make registry key a code block in autoupdate docs.

RealRancor:
Currently only looks like this:

``HKEY_LOCAL_MACHINESoftwarePoliciesownCloudownCloud``

https://doc.owncloud.org/desktop/2.0/autoupdate.html#preventing-automatic-updates-in-windows-environments


RealRancor:
Backport in https://github.com/owncloud/client/issues/4073

PR #4055, opened 2015-11-01
User group sharing

rullzer:
For https://github.com/owncloud/client/issues/3737

With the new sharing and OCS code in place this was a lot easier to write. Not really finished yet. But sharing with user/groups works!

There is still a lot of stuff that needs to be connected in the dialog.
The code needs comments. And more cleanup. 

But overall it is looking promising :)

TODO:
- [ ] Comments
- [x] Only allow this on >8.2 servers (Sharee API)
- [x] Connect lots of stuff in the dialog
- [x] Cleanup sharee stuff
- [ ] Better UI

For the last point I think it is best if someone with more estethic taste has a look once this is merged ;)

Now time for bed.

@MTRichards FYI
CC: @dragotin @ckamm @ogoffart @guruz


src/gui/ocssharejob.h
```diff
@@ -74,14 +74,34 @@ class OcsShareJob : public OcsJob {
     void setPublicUpload(const QString &shareId, bool publicUpload);
 
     /**
+     * Set the permissions
+     *
+     * @pararm permissions
```
phil-davis: s/pararm/param/

src/gui/sharee.cpp
```diff
@@ -0,0 +1,161 @@
...
+    job->getSharees(_search, _type, 1, 50);
+}
+
+void ShareeModel::shareesFetched(const QVariantMap &reply)
+{
+    qDebug() << Q_FUNC_INFO << reply;
+
+    auto data = reply.value("ocs").toMap().value("data").toMap();
+
+    QVector<QSharedPointer<Sharee>> newSharees;
```
guruz: FYI, I think some compilers don't like ```>>``` and you need to write it as ```QVector<QSharedPointer<Sharee> >``` ?
@ogoffart

src/gui/sharee.cpp
```diff
@@ -0,0 +1,161 @@
...
+ShareeModel::ShareeModel(AccountPtr account,
+                         const QString search,
+                         const QString type,
+                         QObject *parent)
+: QAbstractTableModel(parent),
+  _account(account),
+  _search(search),
+  _type(type)
+{
+    OcsShareeJob *job = new OcsShareeJob(_account, this);
```
guruz: error signal?

src/gui/shareusergroupdialog.cpp
```diff
@@ -0,0 +1,226 @@
...
+    cfg.saveGeometry(this);
+    QDialog::done(r);
+}
+
+ShareUserGroupDialog::~ShareUserGroupDialog()
+{
+    delete _ui;
+}
+
+void ShareUserGroupDialog::on_shareeLineEdit_textEdited(const QString &text)
```
guruz: btw i once had a situation where ctrl+V did not trigger this. something to check out..

src/gui/shareusergroupdialog.cpp
```diff
@@ -0,0 +1,226 @@
...
+    _ui->sharePushButton->setEnabled(true);
+}
+
+void ShareUserGroupDialog::on_sharePushButton_clicked()
+{
+    const QModelIndex index = _ui->shareeView->currentIndex();
+
+    auto model = _ui->shareeView->model();
+
+    const QModelIndex shareWithIndex = model->index(index.row(), 2);
```
guruz: Hm, if you want you could replace the integer indices with a constant somewhere?

guruz:
Looks good!

I added some comments..

So ```command_SHARE_USER_GROUP``` it is specified that it will be a different menu entry in the file managers?

src/gui/sharee.cpp
ogoffart: `>>`  is fine, it's C++11 but all compiler we use already support this syntax.

src/gui/sharee.cpp
```diff
@@ -0,0 +1,161 @@
...
+    connect(job, SIGNAL(shareeJobFinished(QVariantMap)), SLOT(shareesFetched(QVariantMap)));
+    job->getSharees(_search, _type, 1, 50);
+}
+
+void ShareeModel::shareesFetched(const QVariantMap &reply)
+{
+    qDebug() << Q_FUNC_INFO << reply;
```
ogoffart: Q_FUNC_INFO is redudent

src/gui/ocssharejob.h
rullzer: fixed

src/gui/shareusergroupdialog.cpp
rullzer: O sure, but I think we need something else than the tableview for the final design. I just needed something I could properly check.

rullzer:
> So command_SHARE_USER_GROUP it is specified that it will be a different menu entry in the file managers?

Mmmm, so it is done that way now. We could maybe merge them later. For now having separte stuff make developing and debugging so much easier.

rullzer:
Ok it is getting more and more ready.

Now we just have 1 share dialog. That uses both widgets. sharing seems to work for me pretty well now.
Still some stuff left but I think it would be good to merge somewhere tomorrow so testing can begin.

TODO
- [ ] Spinner when searching
- [ ] Disable lineedit and button when searching/sharing

I'll take a look at those issues this evening/tomorrow.

Then we need some UI finetuning I think. But that is of later concern (and I'm not that good at it :P)

rullzer:
Ok let me merge this now. We can have a look at the remaining issues later. But then testing can begin!

Issue #4025, opened 2015-10-29
owncloudcmd: Custom WebDAV path


NvanAdrichem:
At my work, where in this case I am just a user, not an administrator, we use OwnCloud with Shibboleth authentication. In order to work with regular WebDAV cliens we use the special "remote.php/nonshib-webdav/" folder with alternative password generation as explained in subsection "WebDAV Support" in https://doc.owncloud.org/server/8.0/admin_manual/enterprise_user_management/user_auth_shibboleth.html . This works quite well.

However, if I want to use `owncloudcmd` from CLI on my server to synchronize a folder, it tries to access "remote.php/webdav/" which initially accepts my nonshib password and starts to synchronize (note that I cannot use my Shibboleth credentials as it requires a web page pop-up to our Single-Sign-On environment). However, it will fail after trying to sync the first file in a subdirectory.

This problem is solved in the source code by changing the following:
```
> diff --git a/src/libsync/theme.cpp b/src/libsync/theme.cpp
> index 709d4c3..404f1e2 100644
> --- a/src/libsync/theme.cpp
> +++ b/src/libsync/theme.cpp
> @@ -409,7 +409,7 @@ bool Theme::wizardSelectiveSyncDefaultNothing() const
 
> QString Theme::webDavPath() const
> {
> -    return QLatin1String("remote.php/webdav/");
> +    return QLatin1String("remote.php/nonshib-webdav/");
> }>
> 
> } // end namespace client
```

However, I expect this to break regular installations that do not have Shibboleth authentication. I tried fixing it by introducing an option `nonShibWebDAV` or `-ns` from CLI defaulting to False to choose URLs, but I got stuck on getting it through the class pointer pattern at Account::create() as my knowledge of Qt at this point is not up-to-par.

Is using an additional option the right way to go and, if yes, how to get the new parameter through the class pointer pattern? If not, do you have any other suggestions that could be implemented?

guruz:
Maybe you want to work together with @DeepDiver1975 on https://github.com/owncloud/client/issues/4007

DeepDiver1975:
@NvanAdrichem what you are trying to achieve can already be done by compiling the client with a custom theme - but maybe having it as pure config option would be a good idea.

@guruz and yes - to some extend this is related to #4007 because we need for configure the dav url

NvanAdrichem:
@DeepDiver1975 that would be an option, but I like to get updates through the repository, which would be a bit harder then.

[...]

Issue #4007, opened 2015-10-26
OC9.0 will come with an additional WebDAV endpoint ....


DeepDiver1975:
..... which would be great if the client could make use of.

``` sh
deepdiver@alien:~/Development/ownCloud/master [master]$ cadaver http://localhost:8080/remote.php/dav
Authentication required for ownCloud on server `localhost':
Username: admin
Password: 
dav:/remote.php/dav/> ls
Listing collection `/remote.php/dav/': succeeded.
Coll:   files                                  0  Okt 26 17:12
Coll:   principals                             0  Okt 26 17:12
dav:/remote.php/dav/> cd files/admin
dav:/remote.php/dav/files/admin/> ls
Listing collection `/remote.php/dav/files/admin/': succeeded.
Coll:   a b c                                  0  Okt 21 15:04
Coll:   litmus                                 0  Okt 23 12:34
Coll:   test                                   0  Okt 21 15:04
        IMG_2003.JPG                      218414  Okt  8 12:38
        IMG_2004.JPG                      238147  Okt  8 12:38
        IMG_2020.JPG                      109440  Okt  8 12:38
        IMG_2021.JPG                       92282  Okt  8 12:38
        IMG_2033.JPG                      139402  Okt  8 12:38
        IMG_2034.JPG                      261785  Okt  8 12:38
        IMG_2035.JPG                      279182  Okt  8 12:38
        IMG_2110.JPG                      107861  Okt  8 12:38
        IMG_2148.JPG                      249447  Okt  8 12:38
        IMG_2165.JPG                      208738  Okt  8 12:38
        New text file.txt                      0  Okt 12 14:04
        import.ics                        187203  Okt 12 14:03
        test-import.ics                      237  Okt 13 14:33
        welcome.txt                          163  Okt  5 17:18
dav:/remote.php/dav/files/admin/> 
```

DeepDiver1975:
... more resources will come

guruz:
Return it as part of the capabilities call then we try to favor it over the default endpoint?

DeepDiver1975:
> Return it as part of the capabilities call then we try to favor it over the default endpoint?

In the first place I need this as a config option to some degree to start testing the endpoint.
especially on the cmd client for smashbox'ing.

The decision to enable it could be done on server version at the end from my understanding

ogoffart:
It can be configured in the theme already  (in Theme::webDavPath).  Is that enough for you?
[...]

PR #4060, opened 2015-11-02
Added Non Shibboleth WebDAV authentication and Dav Path customization to `owncloudcmd`

NvanAdrichem:
According to issue #4025 and @DeepDiver1975 his remark on customized dav path URLs in https://github.com/owncloud/client/issues/4007#issuecomment-152114570 I have created the following changes. I realize it currently doesn't merge, but the current Master branch gave compilation errors, hence I could not merge it by hand as I don't want to send in something untested.


src/cmd/cmd.cpp
```diff
@@ -155,6 +157,8 @@ void help()
     std::cout << "  --password, -p [pass]  Use [pass] as password" << std::endl;
     std::cout << "  -n                     Use netrc (5) for login" << std::endl;
     std::cout << "  --non-interactive      Do not block execution with interaction" << std::endl;
+    std::cout << "  --nonshib, -ns         Use Non Shibboleth WebDAV authentication" << std::endl;
```
DeepDiver1975: As soon as the dav path is customizable there is no need for the nonshib switch from my pov

NvanAdrichem: Maybe not strictly necessary, but adding the nonshib switch (which is also theme customizable) is a lot easier and understandable from a user perspective than remembering and adding '--davpath remote.php/nonshib-webdav'. I don't need the davpath option, I added it while I was at it to do you a favor, I would regret missing out on the nonshib switch.

On Nov 2, 2015 23:10, Thomas Müller <notifications@github.com> wrote:

In src/cmd/cmd.cpp<https://github.com/owncloud/client/pull/4060#discussion_r43690443>:

> @@ -155,6 +157,8 @@ void help()
>      std::cout << "  --password, -p [pass]  Use [pass] as password" << std::endl;
>      std::cout << "  -n                     Use netrc (5) for login" << std::endl;
>      std::cout << "  --non-interactive      Do not block execution with interaction" << std::endl;
> +    std::cout << "  --nonshib, -ns         Use Non Shibboleth WebDAV authentication" << std::endl;


As soon as the dav path is customizable there is no need for the nonshib switch from my pov

DeepDiver1975: > Maybe not strictly necessary, but adding the nonshib switch (which is also theme customizable) is a lot easier and understandable from a user perspective than remembering and adding '--davpath remote.php/nonshib-webdav'.

Since we are not talking about end-user experience I do not consider this too critical.

DeepDiver1975:
Anyway - we need an opinion from the client team @dragotin @danimo @guruz @ogoffart THX

src/libsync/theme.cpp
```diff
@@ -412,5 +412,10 @@ QString Theme::webDavPath() const
     return QLatin1String("remote.php/webdav/");
 }
 
+QString Theme::webDavPathNonShib() const
+{
+    return QLatin1String("remote.php/nonshib-webdav/");
```
guruz: This is a custom endpoint you patched into the server?
Or is that also what @DeepDiver1975 is using?

NvanAdrichem: This is the default alternative path for webdav clients that do not implement Shibboleth authentication in case the server has Shibboleth Integration, hence the "nonshib". See https://doc.owncloud.org/server/9.0/admin_manual/enterprise_user_management/user_auth_shibboleth.html#webdav-support

src/cmd/cmd.cpp
dragotin: It is a misconception that `owncloudcmd` does not reach end users. It does.

dragotin: @NvanAdrichem please do not use the "-ns" thing. Don't ask why, just don't ;-) thanks.

DeepDiver1975: end-user in the sense of lesser technical users. From a command line client user one can expect to type a url path properly as command line parameter

dragotin:
Please resolve the merge problems against master.

Also, please update the documentation in `doc/owncloudcmd.1.rst` within this pr. 

Thanks!

src/cmd/cmd.cpp
NvanAdrichem: I consider myself an end-user... I only use the server installation of my employer, who again lease it as a service from SURFdrive, for which I am as far from an administrator, maintainer or developer as I could ever be. Being unable to make use of the owncloudcmd client through simple authentication options on non-customized, but possibly themed, servers at least bothers my end-user experience, especially since the nonshib davpath is not custom. I consider it relevant, else I would not have gone through the trouble of implementing it and filing a pull request.

NvanAdrichem:
Now mergeable. I removed the -ns and -dp abbreviations (I guess I was too much concerned with our recent non-results in football ;-) ).

I updated doc/owncloudcmd.1.rst and doc/owncloudcmd.rst, where one appears to be a man file and the other a section from the User Documentation. However, I noticed they are not completely in sync, and miss at least the --max-retries option that was recently added in https://github.com/owncloud/client/commit/16c078963b375fac4cb820ff9f3703c80711d3a2.

ogoffart:
I don't understand the nonshib dav path.
remote.php/nonshib-webdav  does not work for me on my 8.1 instance which does not have webdav.

Therefore, I personaly don't think it needs to be in the Theme.

DeepDiver1975:
> remote.php/nonshib-webdav does not work for me on my 8.1 instance which does not have webdav.

you need to enable the user_shibboleth app

guruz:
@dragotin Can we merge this? It seems to fulfill what both @DeepDiver1975 and @NvanAdrichem need.

guruz:
@NvanAdrichem You did not sign the contributor agreement yet can that be? At least you don't show up here in github as contributor

https://owncloud.org/contribute/agreement/

We'd need this asap to have it in 2.1 :)

NvanAdrichem:
@guruz, @karlitschek has go mail!

karlitschek:
Thanks. confirmed! :-)

dragotin:
This is merged into 2.1.

PR #4073, opened 2015-11-04
Backport #4052 to 2.0

RealRancor:
Ref: #4052

### DIFF
--- a/doc/autoupdate.rst
+++ b/doc/autoupdate.rst
@@ -89,7 +89,7 @@ To prevent automatic updates and disallow manual overrides:
 
 1. Edit this Registry key:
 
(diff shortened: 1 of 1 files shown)


guruz:
As far as I heard there won't be a new 2.0.x, sorry.

RealRancor:
But docs are still generated for 2.0 or are the generation of those stopped? Just wondering because https://github.com/owncloud/client/pull/4058 was also merged into 2.0.

danimo:
@guruz the documentation for 2.0 is updated regardless, so this is fine.

guruz:
ok, you guys are of course right.

Issue #1745, opened 2014-05-08
In »Activity«, keep errors at the top


jancborchardt:
Currently when you have errors, it’s very hard to find them in the activity feed because they could be buried down below.

Instead, errors should always be very first in the list. There should ideally be 3 sections (similar to the mobile clients):

* Errors
* In progress
* History (or different word)

In that order, because errors are the most important and probably need user action.

@dragotin @danimo @MTRichards

dragotin:
I don't think it makes sense to clutter the sync protocol with "errors", or rather say "information about not synced files", on top because many users are using the list to verify their sync activity.

The most information that lines up as "information about not synced files" is about why files where ignored from the sync. Ignorings are often intentional and annoy people if these are displayed above their sync progress. OTOH we can not remove the list of ignored files and other problems completely as people sometimes need it to verify their settings. The list should be available, but not prominent.

I think the implementation in #4083 nicely builds that. Closing this.

jancborchardt:
> many users are using the list to verify their sync activity.

An _assumption_ of how people use the current design (which needs to be improved) has little to do with how we need to change it.

What you said above seems to assume that most ownCloud users like to »verify their settings«. However a normal person just wants to sync files, and if there are issues like illegal characters they should be informed of that in one simple activity list.

PR #4083, opened 2015-11-06
New Activity View

dragotin:
I'd like to create this pull request already to discuss the changes done here as early as possible.

This branch changes the way we display the sync activity in the former client releases and adds three new views to the Activity-tab in the client.

1. **Server Activiity**
This displays the server activity list. Currently it is pulled from the server and displayed if the server is reachable. 
2. **Sync Protocol**
Very similar to the former list of synced files, but without the ignored and error files in between. It's a plain positive list now.
3. **Not Synced**
A list of files that were not successfully synced because they were ignored (by user intention or because of other reasons) Later on we can also handle the blacklisting here more user friendly.

Some Screenshots:

server_activity
 
sync_proto

not_synced

Please give me your feedback on this changes and if you think they help to improve user experience. Nothing is set in stone yet, please do not stumble above details. @MTRichards @moscicki @jancborchardt @DeepDiver1975 @karlitschek 

Thanks

### DIFF
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -21,6 +21,7 @@ set(client_UI
     ignorelisteditor.ui
     networksettings.ui
     protocolwidget.ui
+    synclogdialog.ui
     settingsdialog.ui
     sharedialog.ui
     sslerrordialog.ui
@@ -68,6 +69,7 @@ set(client_SRCS
     authenticationdialog.cpp
     proxyauthhandler.cpp
     proxyauthdialog.cpp
+    synclogdialog.cpp
     creds/credentialsfactory.cpp
     creds/httpcredentialsgui.cpp
     creds/shibbolethcredentials.cpp
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -21,6 +21,7 @@ set(client_UI
     ignorelisteditor.ui
     networksettings.ui
     protocolwidget.ui
+    activitywidget.ui
     synclogdialog.ui
     settingsdialog.ui
     sharedialog.ui
@@ -55,6 +56,7 @@ set(client_SRCS
     owncloudgui.cpp
     owncloudsetupwizard.cpp
     protocolwidget.cpp
+    activitywidget.cpp
     selectivesyncdialog.cpp
     settingsdialog.cpp
     sharedialog.cpp
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -57,6 +57,7 @@ set(client_SRCS
     owncloudsetupwizard.cpp
     protocolwidget.cpp
     activitywidget.cpp
+    activityitemdelegate.cpp
     selectivesyncdialog.cpp
     settingsdialog.cpp
     sharedialog.cpp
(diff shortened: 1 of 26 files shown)


jancborchardt:
Isn’t the goal still to make it similar to the web-app activity view? As in, similarly simple to see what happened, with nice profile images of other people who shared stuff with you, file thumbnails, ability to share files from there / open them locally etc.? Keep in mind those are just plans and we should of course do this step-by-step, but this 3-tab view is more complicated than the current view.

One problem it solves is that errors drown in the log – but the proposal for that from me since long ago was to just keep them at the top. See: In »Activity«, keep errors at the top #1745 (from May last year without reply from anyone of you).

dragotin:
About "similar to the web-app": Yes, we want to make more useful. However, the current available API does not allow more than what I have implemented so far. The needed data is not available.

A new API is in discussion here: https://github.com/owncloud/activity/issues/405

About the "three tab solution": It is better for the user to have the three tabs. The main focus is on the default tab which is the server activity, as @karlitschek wanted. In Berlin we agreed to combine the server api and the client sync protocol, however, that is currently not possible due to missing API capabilities. So we want to keep the sync protocol as people are using it to verify their syncs.

I will answer on why I think it makes sense to have a separate tab with "bugs" in #1745

jancborchardt:
> It is better for the user to have the three tabs. The main focus is on the default tab which is the server activity

Why is it better to have three?
The most important thing here is to know when there have been issues, and being able to solve them. That’s why the errors (files not synced, because of illegal chars or whatever) should show up top in the list.

I also answered in https://github.com/owncloud/client/issues/1745#issuecomment-155054748
I’m not happy with this solution and think that splitting it up into 3 separate views overcomplicates this view even more than it already is. cc @karlitschek @DeepDiver1975 @owcnloud/designers

dragotin:
I tried to explain this. If there are real issues, they are displayed more prominent at the account settings anyway. So this is for people who want to dig deeper anyway.

We are talking here about files that were not synced. That is only in rare cases an **error** condition. Mostly it's intentional by ignore patterns. Note that we are away from literally syncing everything, and do not want to go there with our selective sync and ignore patterns.

So what we are talking here is mainly a list of ignored files. That must be available somewhere.

*> Why is it better to have three?*
Because we need two anyway (Server Activity and Sync Protocol), and because of the arguments I gave in #1745 we do not want to clutter the sync protocol any more.

@jancborchardt before talking down the practical solution I was coming up here with (also based on a very difficult API) please take the time to really dig into the nifty details and come up with proposals for **better** solutions which take the given circumstances into consideration.

guruz:
I think it is fine for the 2.1
Better to have that out of the door earlier than later.
We can then wait for feedback and iterate in 2.2

guruz:
Could "Not synced" be renamed to "Issues" and you also list created conflict files there? This was requested at several other points here on github.

jancborchardt:
> I think it is fine for the 2.1

I’m not sure, as I said. It’s more complicated.

> please take the time to really dig into the nifty details and come up with proposals for better solutions which take the given circumstances into consideration.

We did that at the conf. It seems like all of that was forgotten? See also the email I just sent for a recounting of what we talked about. I can mock this up, if needed, but I guess coming to Nürnberg and working face-to-face on a whiteboard seems to be the correct way to settle this without another misunderstanding.

jancborchardt:
Maybe cc @ogoffart whom I specifically talked to about the redesign of this.

MegaV0lt:
Maybe add symbols in the Aktion row for Up and Download?

PR #4100, opened 2015-11-09
Populate account setup credentials from server URL

mnutt:
If a user enters a server URL in the form of https://user:pass@example.com/, this will pre-populate the following credentials page with those values. This streamlines the setup process in the case where the URL already has the credentials.

To test:

1. Start the client and set up a new account. Use `https://user:pass@example.com` for the server URL. Click Next.
2. See `user` pre-populated for the username. See four dots pre-populated for the password.
3. Click Next and check that `user` and `pass` were used for authentication.

### DIFF
--- a/src/gui/wizard/owncloudhttpcredspage.cpp
+++ b/src/gui/wizard/owncloudhttpcredspage.cpp
@@ -78,6 +78,18 @@ void OwncloudHttpCredsPage::initializePage()
         if (!user.isEmpty()) {
             _ui.leUsername->setText(user);
         }
+    } else {
+        QUrl url = ocWizard->account()->url();
+
+        const QString user = url.userName();
+        const QString password = url.password();
+
+        if(!user.isEmpty()) {
+            _ui.leUsername->setText(user);
+        }
+        if(!password.isEmpty()) {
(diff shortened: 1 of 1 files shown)


guruz:
@mnutt Thanks! The only issue still is that you seem not have to signed the contributor agreement?
https://owncloud.org/contribute/agreement/

mnutt:
Hi @guruz, I signed it a while back, I just hadn't followed up with Frank on joining the org. That is done now.

guruz:
:+1:

MTRichards:
We are morphing the URL already on the http: and https: side, and we also have https://github.com/owncloud/android/pull/1119 that changes the behavior of this URL shortening. While that is for Android, we will be adding this to iOS likely in the next sprint, which starts Dec 3. 

We will need to rebase this on that work as soon as we have that work completed. Ok to @mnutt to stay tuned until that is done and then rebase to take that into account (may not need much), and review and get it in for the following release?

mnutt:
@MTRichards that sounds great, I'll keep an eye out for it. Thanks.

guruz:
@MTRichards It is in 2.1 already, but we can make changes to it as needed.

PR #4113, opened 2015-11-11
CMake version abstractions: Remove some dependencies.

dragotin:
That lets us get rid of the hard dependency on Qt tools which are
not easily available everywhere.


dragotin:
@ogoffart, @danimo do you agree we want this change?

guruz:
:+1:

dragotin:
Merged into branch 2.1

Issue #3235, opened 2015-05-13
Store checksum in journal and verifiy upload for eml-files and others


dragotin:
Once #2542 is implement, the checksum should be stored in local sync journal. We could use it to verify if an upload is needed, or if just the files metadata (mtime specifically) has changed. This would solve trouble we have with eml files on windows, and this could be a nice first usage case.

Problems: If we detect that only the mtime has changed, we could propagate the new mtime to the server, using PROPPATCH (needs to be verified). But that will probably not change the ETag of the file and also not the ETags of the parent directories. If it would, a client seeing a changed ETag would not understand that only the mtime has changed, and not the content of the file. 

Note to self: If the server could do file checksums as well, it should also maintain directory checksums. The checksums only would change if the file **contents** or the **contents** of files within a directory changes. The ETag would just be an indication that **something** changed, either content or metadata. The client would need to consider the checksum to find that out.

guruz:
It would be good in a way to have it support multiple checksum algos at the same time.
For example Adler32 is only good for error or change detection but not for deduplication (e.g. for #230)

MTRichards:
Moving to 2.1 to keep 2.0 reasonable size.

cdamken:
@MorrisJobke 

00002839

dragotin:
@cdamken please elaborate. Pasting numbers and adding people is not enough.

MorrisJobke:
> @cdamken please elaborate. Pasting numbers and adding people is not enough.

I think this is just for reference. I was mentioned, because it became a blue ticket and the number is the backlink to their system ;)

dragotin:
@MorrisJobke yes, I know, but I would like to know the content of the blue-ticket as well.

cdamken:
@dragotin  The content of the ticket: *.EML files can't be deleted, even they are deleted in one client will reappear again and again, additionally generates a lot of network traffic.

guruz:
@dragotin @ogoffart @ckamm @moscicki  I was in general wondering if we should have a ```csync_journal``` table
```phash, mtime, checksumtype, checksum``` that is only lazily filled.
Else we risk taking a lot of time to checksum everything..

ckamm:
Storing the checksums in the DB will be part of #3735, this ticket can then be about using them to avoid unnecessary uploads.

ckamm:
This has a couple of steps
* Calculate checksums on upload even if the server doesn't support them, store them in the local db
[...]

PR #4127, opened 2015-11-12
Use content checksums to avoid duplicate uploads #3235

ckamm:
This is the current work in progress for addressing #3235.

Critical bits:
* Computing content checksums when uploading eml files and storing them in the db should already work.
* The change to the treewalk function in syncengine.cpp needs careful review.
* One of the unittests is still failing. I think it's because of my changes to csync_statedb, but my update to test_journal.db didn't seem to help.
* The main thing needed for finishing this is in csync_update.c: We need to compute the current checksum and compare it to the old one.
* It would be cool if we didn't need to recompute the content checksum in propagateupload if it was computed in csync_update

I don't like that there's no parallelism in checksum computation. That was better when the checksums were computed in the propagation phase (see second commit). This would also be very easy to finish if the checksum comparison code lived in propagateupload.


csync/src/csync_statedb.c
```diff
@@ -226,6 +226,8 @@ int csync_statedb_close(CSYNC *ctx) {
   return rc;
 }
 
+#define METADATA_COLUMNS "phash, pathlen, path, inode, uid, gid, mode, modtime, type, md5, fileid, remotePerm, filesize, ignoredChildrenRemote, contentChecksum, contentChecksumTypeId"
```
ogoffart: we could save some CPU cycle by not reading stuff lie uid, gid, mode  (long terme we should remove them totaly)

csync/src/csync_update.c
```diff
@@ -270,25 +270,40 @@ static int _csync_detect_update(CSYNC *ctx, const char *file,
...
-                                                  || (tmp->size != 0 && fs->size != tmp->size)
+        if (ctx->current == REMOTE_REPLICA && !c_streq(fs->etag, tmp->etag)) {
+            st->instruction = CSYNC_INSTRUCTION_EVAL;
+            goto out;
+        }
+        if (ctx->current == LOCAL_REPLICA &&
+                (!_csync_mtime_equal(fs->mtime, tmp->modtime)
+                 // zero size in statedb can happen during migration
+                 || (tmp->size != 0 && fs->size != tmp->size)
 #if 0
```
ogoffart: i guess this can be removed too.  we don't need to compare the inode anymore if we compare the checksum

csync/src/csync_update.c
```diff
@@ -270,25 +270,40 @@ static int _csync_detect_update(CSYNC *ctx, const char *file,
...
-             * sync them.
-             * $> rm a.txt && mv b.txt a.txt
-             * makes b.txt appearing as a.txt yet a sync is not performed because
-             * both have the same modtime as mv does not change that.
-             */
+                 )) {
+
+            if (fs->size == tmp->size && tmp->checksumTypeId) {
+                // ### figure out the right checksum type string by db lookup
+                // ### run the checksum computation
```
ogoffart: quite a big TODO item :-)

ckamm:
This should now target the 2.1 branch. I'll have to close this and open another pull request to change that.

PR #4133, opened 2015-11-13
[Sharedialog] Fix UI stuff

rullzer:
As described in https://github.com/owncloud/client/issues/3737#issuecomment-156036279


rullzer:
Let me merge this so we can move forward with other UI stuff.

Issue #4134, opened 2015-11-13
Account Action Menu misses Log out 


dragotin:
In the Account settings window there is the button on the right providing the action menu with stuff like "open folder", "pause sync", "remove sync" and such.

In that menu, "Log out from Account" is missing as an action. It is available in the tray menu, yet nowadays as people start to work without tray menu, it should be there too.

phil-davis:
Those actions are all per-folder-sync-connection actions.
"Log out of Account" is an action that effects all folder sync connections - like the "Remove Account" button. So should "Log out of Account" appear somehow near "Remove Account", or should that be made into another "..." menu of account-actions? Or?

dragotin:
Thanks @phil-davis you're absolutely right. The menu is the wrong place.

dragotin:
Yes, maybe the "Remove Account" button should also be a ".." menu with Account actions like 
' Log off
* Remove Account
* Add a new Account

Dianafg76:
I tested this issue and is working Ok
Desktop v Desktop v ownCloud-2.1.0.2896-nightly20151120.pkg
Server v {"installed":true,"maintenance":false,"version":"8.2.1.2","versionstring":"8.2.1 RC3","edition":"Enterprise"}
Closed

PR #4139, opened 2015-11-13
AccountSettings: Add a toolbox button for the account specific actions.

dragotin:
Also move the 'Add Account' button from the General Tab, where it
is not properly found, to the new account toolbox.

This results in a look like that:
toolboxbutton
Also note the Add-Account functionality, that moved from the General tab to here. 

@jancborchardt how is that? @phil-davis as we talked about in the other issue #4134

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -135,19 +135,19 @@ AccountSettings::AccountSettings(AccountState *accountState, QWidget *parent) :
 void AccountSettings::createAccountToolbox()
 {
     QMenu *menu = new QMenu();
-    _toggleSignInOutAction = new QAction(tr("Sign Out"), this);
-    connect(_toggleSignInOutAction, SIGNAL(triggered(bool)), SLOT(slotToggleSignInState()));
-    menu->addAction(_toggleSignInOutAction);
-
-    QAction *action = new QAction(tr("Remove Account"), this);
-    menu->addAction(action);
-    connect( action, SIGNAL(triggered(bool)), SLOT(slotDeleteAccount()));
-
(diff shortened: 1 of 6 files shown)


ogoffart:
:+1:  Looks great

danimo:
:+1: Liking it

phil-davis:
:+1: Looks good.
This is one place to put the "Add account" option. I know there has been discussion about this before, and new users had trouble finding in its old place, in the General tab. This seems better, and when there is no account defined at all the wizard will start.
The other option is to have a "+" Add Account button up in the top row, along with the existing account icons and the Activity, General and Network tabs.
Hmmm - maybe what is here is the way to go.

pierrejochem:
:+1:

jancborchardt:
Great stuff @dragotin! :)

Only details:
* [ ] »Add new account« should be sorted in first (as destructive actions always should go to the very bottom of a list)
* [ ] »account« lowercase and no ellipsis at the end
* [ ] to make that dropdown more discoverable, can we add a triangle to the right? If there’s no default image, we can use the [triangle](https://github.com/owncloud/core/blob/master/core/img/actions/triangle-s.png) from core (SVG with same name also in there)

guruz:
This would have been better in the tray menu..

guruz:
If I see it at this place, I'll think this is about adding (creating?) an account on that server. In the tray menu you can put it in a more generic place.

jancborchardt:
@guruz the tray menu is for actions which are performed regularly. Adding an account is not, so it does not belong in there.

ogoffart:
I think the first entry should probably be "Log In"  especially when we are in disconnected mode (because the password was changed for example, or that the shiboeth session expired)

PR #4159, opened 2015-11-18
Fix #3490

### LINKED ISSUE #3490: FileFirewall: Handle Reply Codes Gracefully
If the client connects to a server with a file firewall enabled, it should handle the reply codes sent by the firewall gracefully, ie. display proper error messages and not block the entire sync. 

@MTRichards is it that? 

@nickvergessen can you provide us with a list of the possible reply codes?

dragotin:
Handling firewall issues correctly.


ogoffart:
:+1: +

PR #4169, opened 2015-11-20
[Sharing] Make sure the file is actually synced to the server

rullzer:
Before we attemt any sharing make sure the file is synced to the server.

We should make sure actually that non of the file browser plugins tries to share a file if it does not have to proper status.

### DIFF
--- a/src/gui/socketapi.cpp
+++ b/src/gui/socketapi.cpp
@@ -378,12 +378,23 @@ void SocketApi::command_SHARE(const QString& localFile, QIODevice* socket)
         // if the folder isn't connected, don't open the share dialog
         sendMessage(socket, message);
     } else {
+        const QString file = QDir::cleanPath(localFile).mid(shareFolder->cleanPath().length()+1);
+        SyncFileStatus fileStatus = this->fileStatus(shareFolder, file);
+
+        // Verify the file is on the server (to our knowledge of course)
+        if (fileStatus.tag() != SyncFileStatus::STATUS_SYNC &&
+            fileStatus.tag() != SyncFileStatus::STATUS_UPDATED) {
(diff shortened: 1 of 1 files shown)


rullzer:
Now also the SHARE_STATUS command will reply with `SHARE_STATUS:NOTSYNCED:<FILE>` if the file/folder is not properly synced yet.

BuenGenio:
Actually this change is causing our design team a lot of grief. They used to rely on the ability to share links while the upload is still in progress, as they will sometimes be working late and would send these links in order to meet deadlines, or close the project without forgetting to send those emails, late at night as they wouldn't be read until morning, allowing uploading to complete and the designers to go home. 

Can we make this an option, rather than enforce this change, as it's making defending my case for switching our 100-strong operation from Dropbox to ownCloud that much more difficult (the other one being that there is no delta-sync, and it's taking our 6 designers a month to download 1.5TB worth of data (times 6!) - clogging up everyone else's bandwidth...

Thanks

rullzer:
@BuenGenio well then it just worked by chance. The thing is we do not have any quarantee from the server side that a file is uploaded and available before the sync is done. This has mainly to do with the server internals. If you are uploading a big file which is chunked it is even less safe to do so.

We need to think how to properly do this. But getting the old behaviour back is not right since it is in fact a bug that it was allowed.

BuenGenio:
Well, would something as simple as inserting the link<->path for the file being uploaded into oc_shares as soon as the upload starts work?

Both are known at the very beginning, so should be trivial. Or is there anything else missing I'm not taking into account?

rullzer:
On an abstract level yes. But the reality is a bit more complex. Basically we do not intesert paths into the oc_share table, we insert file ids. And as far as I know they are not guaranteed to be stable until the whole file is uploaded.

PR #4180, opened 2015-11-23
Use content checksums to avoid duplicate uploads #3235

ckamm:
See also #4127 

Please don't merge this, I want to squash commits. They're just preserved to make reviewing easier


src/libsync/networkjobs.cpp
```diff
@@ -565,6 +565,82 @@ bool PropfindJob::finished()
...
+
+}
+
+void ProppatchJob::start()
+{
+    if (_properties.isEmpty()) {
+        qWarning() << "Proppatch with no properties!";
+    }
+    QNetworkRequest req;
+    req.setRawHeader("Depth", "0");
```
ogoffart: I did not know there is was a Depth in PROPPATCH

src/libsync/propagateupload.cpp
```diff
@@ -216,24 +216,38 @@ void PropagateUploadFileQNAM::start()
...
-        if (supportedChecksumTypes.contains(_item->_transmissionChecksumType)) {
-            // TODO: We could validate the old checksum and thereby determine whether
-            // an upload is necessary or not.
-            slotStartUpload(_item->_transmissionChecksumType, _item->_transmissionChecksum);
-            return;
-        }
+    // Compute the content checksum.
+    auto computeChecksum = new ComputeChecksum(this);
+    // We currently only do content checksums for the particular .eml case
+    // This should be done more generally in the future!
```
ogoffart: i think we could actually do it for every files... no?

ckamm:
Thanks for the review @ogoffart, it's merged.

PR #4199, opened 2015-11-25
[Sharing] Permissions now in dropdown menu

rullzer:
As discusses with jan.

* Detailed permissions displayed in qtoolboxmenu
* Made share rows slightly smaller

Bug fix:

* Do not show delete permissions for file shares

### DIFF
--- a/src/gui/shareusergroupwidget.cpp
+++ b/src/gui/shareusergroupwidget.cpp
@@ -225,6 +225,9 @@ ShareWidget::ShareWidget(QSharedPointer<Share> share,
     _ui->permissionToolButton->setMenu(menu);
     _ui->permissionToolButton->setPopupMode(QToolButton::InstantPopup);
 
+    QIcon icon(QLatin1String(":/client/resources/more.png"));
+    _ui->permissionToolButton->setIcon(icon);
+
     // Set the permissions checkboxes
     displayPermissions();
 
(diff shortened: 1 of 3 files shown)


rullzer:
@jancborchardt this is how it looks now:

new

src/gui/shareusergroupwidget.cpp
```diff
@@ -196,27 +198,39 @@ void ShareUserGroupWidget::slotCompleterActivated(const QModelIndex & index)
...
-  _showDetailedPermissions(false)
+  _isFile(isFile)
 {
     _ui->setupUi(this);
 
     _ui->sharedWith->setText(share->getShareWith()->format());
+ 
+    // Create detailed permissions menu
+    QMenu *menu = new QMenu(this);
+    _permissionCreate = new QAction(tr("create"), this);
```
ckamm: isn't create also only relevant for !isFile?

ckamm:
Rest looks good!

dragotin:
:+1:

jancborchardt:
@rullzer awesome! Just one thing – similar to the delete icon, use the `more` icon for that button: https://github.com/owncloud/core/blob/master/core/img/actions/more.png
(not just the dots and neither the triangle)

src/gui/shareusergroupwidget.cpp
rullzer: create is also overwrite I think... (I just followed the webinterface).

rullzer:
@jancborchardt I now use the more icon. Howver I have no idea how to remove the tiny arrow...

rullzer:
I'll merge this and somebody with more styling fu shall fix the tiny triangle...

PR #4212, opened 2015-11-27
show displaynames of users in share dialog instead of userid

blizzz:
Problem found by danimo and investigated by me: searching for LDAP users was buggy. They only appeared, when the search started with the userid.  

Server works correctly.

Turns out, displayname variable was not used. 

If we need this for 2.1 i can rebase and open a PR against that branch.

@danimo @rullzer


rullzer:
Ai... my bad. :+1:

blizzz:
→ will open RP against 2.1

blizzz:
→ https://github.com/owncloud/client/pull/4213

PR #4213, opened 2015-11-27
show displaynames of users in share dialog instead of userid

blizzz:
Now against 2.1 :)

Problem found by danimo and investigated by me: searching for LDAP users was buggy. They only appeared, when the search started with the userid.

Server works correctly.

Turns out, displayname variable was not used.

@danimo @rullzer


rullzer:
:+1:

PR #4217, opened 2015-11-30
[Sharing] Files should not have create permissions

### LINKED ISSUE #4216: Create permission is shown when sharing a file with a user
The create permission appears in the dropdown sharing a file with a user.

It should only be shown whe we share a folder; sharing files should only let 3 permission:
- Can share
- Can edit
- Can change

screen shot 2015-11-30 at 10 34 27

rullzer:
Fixes #4216


dragotin:
:+1:

Issue #4218, opened 2015-11-30
You can't share by link with normal users, and with SHIB users, with Server v8.0.9


Dianafg76:
### Steps to reproduce
1. Install the Desktop client v 2.1.0 with the server v 8.0.9
2. Go to the Desktop client
3. Create a new folder (e.g. aaa)
4. Right click on the folder aaa
5. Click Share with oC
 

### Expected behaviour
You should be able to share by link
### Actual behaviour
You can't share by link with normal users, and with SHIB users
![screen shot 2015-11-30 at 16 35 39](https://cloud.githubusercontent.com/assets/12032058/11475965/76e60b0c-9780-11e5-8e73-f3f6b70ae4b0.png)

### Server configuration
Desktop client v ownCloud-2.1.0.2922-nightly20151130.pkg
Server v {"installed":true,"maintenance":false,"version":"8.0.9.2","versionstring":"8.0.9","edition":"Enterprise"}

Dianafg76:
### Desktop + shib


| Desktop v | Server v  | Result |
|:------------- |:---------------:| -------------:|
| 2.0.2   | 8.0.9 |        OK |
| 2.1.0     | 8.0.9        |           NOK, https://github.com/owncloud/client/issues/4218 |
| 2.1.0 | 8.1.4        |            OK|
| 2.1.0 | 8.2.1        |            OK|

Dianafg76:
I tested this issue, and now I can share by link with Desktop v 2.1.0 and Server v 8.0.9
Desktop v ownCloud-2.1.0.2924-nightly20151201.pkg
Server v {"installed":true,"maintenance":false,"version":"8.0.9.2","versionstring":"8.0.9","edition":"Enterprise"}

Closed

PR #4219, opened 2015-11-30
[Sharing] Default capabilitie for link sharing

rullzer:
Required to fix sharing for pre 8.1 servers

I hope this Fixes https://github.com/owncloud/client/issues/4218

CC: @Dianafg76 @dragotin


dragotin:
Tested and works fine. Note that with servers < 8.2 a share link is created right away once its clicked on the menu item. Works as designed.

PR #4224, opened 2015-12-01
updates for 2.1 release

carlaschroder:
This is the first batch of updates for 2.1, some updated screenshots and descriptions.


carlaschroder:
@danimo @dragotin

danimo:
:+1:

dragotin:
@carlaschroder do you realized that the "Add Account" button, that was in the General tab before, was moved to this toolbox-button on the account page. This has the implication that, if the wizard is closed, people need to left-click on the tray icon to get the configuration dialog started again. No way to do it from the settings dialog currently.

Not sure if that is a good solution, but not blocking 2.1 currrently I'd say.

carlaschroder:
thx @dragotin, I'll add a section highlighting how to add new accounts.

Issue #4227, opened 2015-12-01
[Sharing link] OCS API error code: 403, cannot remove a password, when you unselect the option Public sharing requires a password


Dianafg76:
### Steps to reproduce
1. Go to the Server web with Admin
2. Go to the option Admin/admin
3. On the option Sharing, select the option "Enforce password protection"
4. Go to the Desktop v with v 2.1.0, with user1
5. Click on the oC icon
6. Click on "Open folder ....."
7. Create a new folder (e.g aaa)
8. Right click on the folder aaa
9. Select the option Share with ownCloud
10. Select the option Share link
11. Introduce the password
12. Click on Set Password
13. If you unselect the option Public sharing requires a password

### Actual behaviour
You can see this error:  OCS API error code: 403, cannot remove password
![screen shot 2015-12-01 at 12 09 43](https://cloud.githubusercontent.com/assets/12032058/11499186/83c6496e-9824-11e5-88f7-ba6ad0550004.png)

### Server configuration
Desktop v ownCloud-2.1.0.2924-nightly20151201.pkg 
Server v {"installed":true,"maintenance":false,"version":"8.2.1.4","versionstring":"8.2.1","edition":"Enterprise"}

Dianafg76:
Log
```
12-01 12:15:28:732 0x7fe80342e710 OCC::AbstractNetworkJob::start: !!! OCC::OcsShareJob created for "https://docker.oc.solidgear.es" + "ocs/v1.php/apps/files_sharing/api/v1/shares/47"
12-01 12:15:28:902 0x7fe80342e710 OCC::OcsJob::finished: Reply to "PUT"  QUrl( "https://docker.oc.solidgear.es:53871/ocs/v1.php/apps/files_sharing/api/v1/shares/47" )  (QPair("password","")) has unexpected status code: 403 "{"ocs":{"meta":{"status":"failure","statuscode":403,"message":"Cannot remove password"},"data":[]}}"
12-01 12:15:39:345 0x7fe80342e710 OCC::ConnectionValidator::checkAuthentication: # Check whether authenticated propfind works.
12-01 12:15:39:345 0x7fe80342e710 OCC::AbstractNetworkJob::setTimeout: void OCC::AbstractNetworkJob::setTimeout(qint64) 27000
12-01 12:15:39:346 0x7fe80342e710 OCC::AbstractNetworkJob::start: !!! OCC::PropfindJob created for "https://docker.oc.solidgear.es" + "/"
```

[...]

PR #4228, opened 2015-12-01
[Sharing] Do not allow users to uset the password if it is enforced

rullzer:
Fixes https://github.com/owncloud/client/issues/4227

### DIFF
--- a/src/gui/sharelinkwidget.cpp
+++ b/src/gui/sharelinkwidget.cpp
@@ -236,9 +236,8 @@ void ShareLinkWidget::slotSharesFetched(const QList<QSharedPointer<Share>> &shar
             _ui->widget_shareLink->show();
             _ui->checkBox_shareLink->setChecked(true);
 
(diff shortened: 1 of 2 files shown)


src/gui/sharelinkwidget.cpp
```diff
@@ -234,7 +236,9 @@ void ShareLinkWidget::slotSharesFetched(const QList<QSharedPointer<Share>> &shar
             _ui->widget_shareLink->show();
             _ui->checkBox_shareLink->setChecked(true);
 
-            _ui->checkBox_password->setEnabled(true);
+            if (!_passwordRequired) {
+                _ui->checkBox_password->setEnabled(true);
+            }
```
dragotin: how about doing _ui->checkBox_password->setEnabled( !_passwordRequireed) ? That way it's shorter and you always have a defined state.

dragotin:
One minor comment, otherwise :+1:

PR #4245, opened 2015-12-03
Fixing bug in opening Activity log from tray icon menu 'Recent Changes/Details...'

masoudcs:
Fixing bug in opening Activity log from tray icon menu 'Recent Changes/Details...'.
such that Settings dialog shows Activity tab, but highlighted tab could be different.
the problem fixed in settingsdialog.cpp.
settingsdialogmac.cpp checked, doesn't have the problem.

owncloud-bug-screenshot

### DIFF
--- a/src/gui/settingsdialog.cpp
+++ b/src/gui/settingsdialog.cpp
@@ -183,7 +183,7 @@ void SettingsDialog::showFirstPage()
 void SettingsDialog::showActivityPage()
 {
     if (_activityAction) {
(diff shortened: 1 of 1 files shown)


guruz:
Thanks @masoudcs 

@karlitschek do we need https://owncloud.org/contribute/agreement/ for such oneliners?

karlitschek:
@guruz no. FSF best practice is that it is not needed if less then 7 lines.

PR #4310, opened 2015-12-22
Avatars in share dialog

rullzer:
Requires https://github.com/owncloud/core/pull/21311

Add avatars to the share dialog just like on the server. No avatars for federated shares yet but it is a good first step.


dragotin:
API is not available on server version 9.0, moving to backlog.

guruz:
:+1: 

@rullzer API not available is OK, it should not produce a visible error..

rullzer:
I prefer not to merge this just yet. Since the endpoint stuff might change (significantly). For now lets keep this PR here.

dragotin:
ok

guruz:
@rullzer @ckamm The Avatar code by @dragotin had been merged in master some time ago.

The share UI code has changed a lot.

Does anyone of you want to re-base this on current master so we can get it in?

ckamm:
@guruz Yep, I could imagine adapting this for the share dialog some time, possibly for 2.4.0.

ckamm:
I've rebased it on top of master and plan to fix and merge it. @rullzer fyi!

SamuAlfageme:
@ckamm out of curiosity: what should we do when `https://<server>/remote.php/dav/avatars/demo/128.png` returns 404? same as account tab i.e. use the `@` icon; server strategy: generate an avatar with user initials in the client side? ask the core for such avatar?

ckamm:
@SamuAlfageme What @rullzer's code does currently is to generate a fallback avatar locally based on the first letter of the name.

ckamm:
@ogoffart I think this is ready for review

src/gui/shareusergroupwidget.cpp
```diff
@@ -416,6 +419,71 @@ ShareUserLine::ShareUserLine(QSharedPointer<Share> share,
...
+     *
+     * This will be shown until the avatar image data arrives.
+     */
+    QString seed = _share->getShareWith()->shareWith();
+    if (_share->getShareWith()->type() != Sharee::User) {
+        seed += QString(" %1").arg(_share->getShareWith()->type());
+    }
+
+    const QByteArray hash = QCryptographicHash::hash(seed.toUtf8(), QCryptographicHash::Md5);
+    int hue = ((double)hash.mid(0, 3).toHex().toInt(0, 16) / (double)0xffffff) * 255;
```
ogoffart: nitpick: C cast

I guess this could be rewriten as int(hash[0]),  so we don't do two allocations and some non-trivial computation.

SamuAlfageme:
@ckamm very cool! 😍 

Maybe we could consider delimiting each entry (share) using a list with alternate colors (like the one in "Sync Protocol/Not Synced") to get rid of the rectangle that encloses the checkboxes now. I remember a mockup from some time ago:

<p align="center">
  
</p>

## Screenshot for reference: 

<p align="center">
  
</p>

cc/ @michaelstingl @pmaier1

ckamm:
@SamuAlfageme Good idea, that'd make it nicer looking -> new issue please!

@ogoffart I've addressed the nitpick. Just using the first byte would probably break the server equivalence of the code. (I'm not 100% it's worth keeping)

ckamm:
Actually, I tested and the colors of the fallback avatars already don't match with the server colors.

ckamm:
This looks good now and actually includes @SamuAlfageme's suggestion from #6176.

PR #4338, opened 2016-01-06
add note re mtime syncing

carlaschroder:
Please review @danimo @ogoffart @dragotin 

Reference: https://github.com/owncloud/core/issues/7009


ogoffart:
:+1:

Issue #1408, opened 2014-02-10
Option to disable prompt to delete all files from a folder


wrestlingstud:
I would like to request that a client configuration switch be enabled in the client's config file that would disable the user confirmation dialog box about deleting files from the local pc that have already been deleted from the server

guruz:
@dragotin @danimo I remember you two discussing about a similar issue last week.. FYI.

guruz:
Merged.. thanks @Bottswana

PR #4389, opened 2016-01-24
Add option to disable the delete all files prompt

Bottswana:
Hello Folks,
I added this feature for my own purposes and thought I'd create a PR in case you wanted to merge it, given you have this as a request pending: https://github.com/owncloud/client/issues/1408

This PR adds an option to the generic settings page, default enabled, to control the display of the all files deletion prompt.

I'm not sure if this patch is considered trivial or not, so if you need me to sign the contributor agreement I can do that.

### DIFF
--- a/src/gui/generalsettings.cpp
+++ b/src/gui/generalsettings.cpp
@@ -62,7 +62,6 @@ GeneralSettings::GeneralSettings(QWidget *parent) :
 
     // misc
     connect(_ui->monoIconsCheckBox, SIGNAL(toggled(bool)), SLOT(saveMiscSettings()));
-    connect(_ui->promptDeleteAllFiles, SIGNAL(toggled(bool)), SLOT(saveMiscSettings()));
     connect(_ui->crashreporterCheckBox, SIGNAL(toggled(bool)), SLOT(saveMiscSettings()));
     connect(_ui->newFolderLimitCheckBox, SIGNAL(toggled(bool)), SLOT(saveMiscSettings()));
     connect(_ui->newFolderLimitSpinBox, SIGNAL(valueChanged(int)), SLOT(saveMiscSettings()));
@@ -101,7 +100,6 @@ void GeneralSettings::loadMiscSettings()
 {
     ConfigFile cfgFile;
     _ui->monoIconsCheckBox->setChecked(cfgFile.monoIcons());
-    _ui->promptDeleteAllFiles->setChecked(cfgFile.promptDeleteFiles());
     _ui->desktopNotificationsCheckBox->setChecked(cfgFile.optionalDesktopNotifications());
     _ui->crashreporterCheckBox->setChecked(cfgFile.crashReporter());
     auto newFolderLimit = cfgFile.newBigFolderSizeLimit();
@@ -128,8 +126,6 @@ void GeneralSettings::saveMiscSettings()
     ConfigFile cfgFile;
     bool isChecked = _ui->monoIconsCheckBox->isChecked();
     cfgFile.setMonoIcons(isChecked);
-    bool promptIsChecked = _ui->promptDeleteAllFiles->isChecked();
-    cfgFile.setPromptDeleteFiles(promptIsChecked);
     Theme::instance()->setSystrayUseMonoIcons(isChecked);
     cfgFile.setCrashReporter(_ui->crashreporterCheckBox->isChecked());
 
--- a/src/gui/generalsettings.cpp
+++ b/src/gui/generalsettings.cpp
@@ -62,6 +62,7 @@ GeneralSettings::GeneralSettings(QWidget *parent) :
 
     // misc
     connect(_ui->monoIconsCheckBox, SIGNAL(toggled(bool)), SLOT(saveMiscSettings()));
+    connect(_ui->promptDeleteAllFiles, SIGNAL(toggled(bool)), SLOT(saveMiscSettings()));
     connect(_ui->crashreporterCheckBox, SIGNAL(toggled(bool)), SLOT(saveMiscSettings()));
     connect(_ui->newFolderLimitCheckBox, SIGNAL(toggled(bool)), SLOT(saveMiscSettings()));
     connect(_ui->newFolderLimitSpinBox, SIGNAL(valueChanged(int)), SLOT(saveMiscSettings()));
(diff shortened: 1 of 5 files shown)


guruz:
Hi, in General: Yes, we'd Need you to sign the contributor Agreement :)

If the patch will be taken or not is not decided, in General adding Options in UI is always a bit discouraged.

FYI @jancborchardt

Bottswana:
That's no issue, I'll wait until that is decided before signing the agreement though to save workload.
If there's an alternative suggestion other than providing an option in the GUI, I'm happy to implement, but based on the linked issue, the removal of the dialog seemed 50/50, so it made sense to provide the choice to the end user.

Thanks!

jancborchardt:
What’s the use-case of this setting?

Bottswana:
So our use case for this in this scenario is we create users with no files in and the client running on a machine, and then share folders or files to the user when they need access to them. In this scenario, when a sync occurs that would remove all files (Which we do at the end of each day), nothing is removed and the prompt appears requiring user intervention.

In our use case, we need to have full control of what the client is doing, and not prompt the user if the action should take place. 

For my home use of ownCloud, I like this prompt and would not like it removed, but in the use case I stated above, its presence breaks the entire model we are creating.

I'm happy with just merging this in for our clients when we use it in this case, but you had an open issue for it, so I presume other people would want this option too.

Thanks

dragotin:
I would not like to have that exposed in the GUI, but it could be an option that can be set in the config file by the admin, and also later be set through server based options.

Bottswana:
I can easily remove the option from the GUI and leave it as config file option only, but that does seem a little cryptic for something which is only affecting the GUI itself in my opinion.

jancborchardt:
Yeah, totally agree with @dragotin here. This setting would deactivate a pretty useful and crucial feature. Also, it seems useful only for very few use-cases compared to the other settings, so adding it in the GUI isn’t really justified.

Bottswana:
Are we happy then to drop the GUI option and leave it as a config level
option for use cases such as mine?

guruz:
@Bottswana Yes please change the PR :)
And submit the contributor agreement.
Thanks for contributing!

Bottswana:
Thanks,
PR Updated and agreement signed and emailed.

karlitschek:
confirmed that I got the agreement :-)

ogoffart:
Thanks for the pull request:

Regarding the code: please do not add a dependency between the sync engine and the config file.  
Please put the logic in Folder::slotAboutToRemoveAllFiles.

Regarding the UI, maybe a "don't show again" checkbox would be acceptable

Bottswana:
No worries, Will do this after I get back from cisco live.
In regards to the UI, if you wish me to implement that box, thats no worries. But if it is still preferred to be config only, I dont want to waste any time.
-J

ogoffart:
i don't know what @jancborchardt  will think about this checkbox.  Personally i believe it's a good thing to have.  But this can be done in another pull request regardless

wmarkus85:
HI my config look like:
[General]
optionalDesktopNotifications=true
crashReporter=false
monoIcons=false
newBigFolderSizeLimit=500
useNewBigFolderSizeLimit=false
promptDeleteFiles=false

[Accounts]

Client Version  2.2.0-nightly20160215 (build 5851)

but the pop up still appears. -> what's wrong?
BR Markus

Bottswana:
This isn't merged yet, are you using the copy from my fork?

wmarkus85:
which copy?

jancborchardt:
@ogoffart as said above I agree with @dragotin 

> I would not like to have that exposed in the GUI, but it could be an option that can be set in the config file by the admin, and also later be set through server based options.

ogoffart:
@jancborchardt the original patch added an option in the settings UI.  Here I was talking about a "[x] do not show again"  checkbox on the message box itself.

jancborchardt:
Doesn’t fully make sense in this case because it’s a valid warning every time it appears. And it appears every time for a reason. As people are prone to click »ok« on these warnings and check everything that looks like a »don’t show again«, it wouldn’t be proper to offer that here.

guruz:
I merged it prematuerely... thanks @ogoffart for fixing the things mentioned in https://github.com/owncloud/client/pull/4389#issuecomment-181418098 .. :+1:

RealRancor:
Please don't forget to add this option to the documentation of the available config files at:

https://doc.owncloud.org/desktop/2.2/advancedusage.html#configuration-file

guruz:
done..

JarbyDev:
Hello everytime a file got deleted in web UI, i got this warning in the desktop, how i can  disabled this, Thanks

nc

RealRancor:
@JarbyDev As you're running the nextcloud client (and probably also nextcloud itself) most likely the sames an in https://github.com/owncloud/client/issues/5265#issuecomment-255511121 applies here.

JarbyDev:
I have 2 instances one owncloud and one nextcloud, (i get the same pop up in both instances) also i tried to go to the configuration file but i dont see the option promptDeleteAllFiles.

This is my current owncloud,cfg

[General]
optionalDesktopNotifications=true

[Accounts]
0\Folders\1\localPath=C:/Users/jarbydev/JMH Drive/
version=2
0\Folders\1\targetPath=/
0\url=https://owncloud.example.com
0\Folders\1\paused=false
0\serverVersion=
0\Folders\1\ignoreHiddenFiles=true
0\http_certificatePasswd=
0\http_certificatePath=
0\http_user=jarbydev
0\authType=http
0\user=jagudelo

RealRancor:
@JarbyDev You're really in the wrong place here (in this Pull Request). If you think there is an issue here create a new bugreport...

RealRancor:
To sum-up:

- This is a bugtracker. Questions like this should always go to https://central.owncloud.org/
- A Pull Request shouldn't be misused to ask questions about a feature. If there is a issue with a new introduced feature like this a new issue should be created
- If you have issues with the nextcloud client against a ownCloud server and a Nextcloud server issues with that client needs to go to https://github.com/nextcloud/client_theming
- If you think the issue is caused by the server a new issue should go to the specific server issue tracker (oC or NC)

Bottswana:
Just add the line to your config file under general and restart the client.
Its a 'hidden' config option.

promptDeleteAllFiles=false

PR #4396, opened 2016-01-27
Fix enum csync

dragotin:
This fixes a  compatibility issue with the definitions of the type members within the two used structs.

### DIFF
--- a/csync/src/csync_update.c
+++ b/csync/src/csync_update.c
@@ -106,6 +106,49 @@ static bool _last_db_return_error(CSYNC* ctx) {
     return ctx->statedb.lastReturnValue != SQLITE_OK && ctx->statedb.lastReturnValue != SQLITE_DONE && ctx->statedb.lastReturnValue != SQLITE_ROW;
 }
 
+/*
+ * This static method is needed because the type members of the two structs use
+ * different enum values. A direct comparion is not neccessarily correct.
+ *
+ * tmp is csync_file_stat_t
+ * fs  is csync_vio_file_stat_t with this vio type:
+ *  enum csync_vio_file_type_e {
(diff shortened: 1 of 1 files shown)


csync/src/csync_update.c
```diff
@@ -106,6 +106,49 @@ static bool _last_db_return_error(CSYNC* ctx) {
...
+    if( !(tmp && fs)) return false;
+
+    if( tmp->type == CSYNC_FTW_TYPE_SKIP ) return true;
+
+    if( tmp->type == CSYNC_FTW_TYPE_DIR && fs->type != CSYNC_VIO_FILE_TYPE_DIRECTORY )
+        return true;
+    if( tmp->type == CSYNC_FTW_TYPE_FILE && fs->type != CSYNC_VIO_FILE_TYPE_REGULAR )
+        return true;
+    if( tmp->type == CSYNC_FTW_TYPE_SLINK && fs->type != CSYNC_VIO_FILE_TYPE_SYMBOLIC_LINK )
+        return true;
```
ckamm: After this, the only possibe cases left are DIR/DIRECTORY, FILE/REGULAR, SLINK/SYMBOLIC_LINK (exhaustive check on tmp->type), the checks below can be removed.

ckamm:
Thanks for fixing this!

csync/src/csync_update.c
ogoffart: yes, the next 3 if will  always be fale

dragotin:
Pushed the fix to the above comments, @ckamm, @ogoffart please review again.

ckamm:
@dragotin :+1:

ogoffart:
:+1: 1

PR #4404, opened 2016-01-28
Size of sync folders

ouafnico:
Hello

On the sync client, when having multiple synchronization folders, can we have the size of each?

Thanks


ogoffart:
Hi, please use the the forums or other support mediums for questions: https://owncloud.org/support/

PR #4437, opened 2016-02-05
Don't install tests and cmocka lib

hefee:
Neither tests nor the libcmocka needs to be installed globally.

ctest still works.


ogoffart:
Merged in master   (for 2.2)

Thanks for the patch, it indeed made no sens to install these.

Issue #4484, opened 2016-02-18
Once the server is down, there is not any error message when you try to share with ownCloud


Dianafg76:
### Expected behaviour
An error should appear letting the user know that there is not connection 

### Actual behaviour
After clicking on "share with ownCloud" nothing happens. Meaning nothing that the sharing view is not open neither any error message for the user to know what's going on.

LOG
```
02-18 09:14:25:215 0x7fb959d1b070 OCC::SocketApi::sendMessage: SocketApi:  Sending message:  "SHARE:NOTCONNECTED:/Users/Diana/Https_federatedsharing/shareadmin"
02-18 09:14:40:298 0x7fb959d1b070 OCC::ConnectionValidator::checkServerAndAuth: Checking server and authentication
02-18 09:14:40:299 0x7fb959d1b070 OCC::ConnectionValidator::checkServerAndAuth: Trying to look up system proxy
02-18 09:14:40:299 0x7fb95c425690 OCC::SystemProxyRunnable::run: virtual void OCC::SystemProxyRunnable::run() Starting system proxy lookup
02-18 09:14:40:303 0x7fb959d1b070 OCC::ConnectionValidator::systemProxyLookupDone: No system proxy set by OS
02-18 09:14:40:304 0x7fb959d1b070 OCC::AbstractNetworkJob::start: !!! OCC::CheckServerJob created for "https://xxxxxx” + "status.php" "OCC::ConnectionValidator"
02-18 09:14:40:665 0x7fb959d1b070 OCC::AbstractNetworkJob::slotFinished: void OCC::AbstractNetworkJob::slotFinished() 1 "Connection refused" QVariant(Invalid)
02-18 09:14:40:665 0x7fb959d1b070 OCC::ConnectionValidator::slotNoStatusFound: void OCC::ConnectionValidator::slotNoStatusFound(QNetworkReply *) 1 "Connection refused"
02-18 09:14:40:666 0x7fb959d1b070 OCC::AccountSettings::showConnectionLabel: "No connection to ownCloud at <a href="https://docker.oc.solidgear.es:55155">https://xxxxxxxxx</a>.
Connection refused"
02-18 09:14:40:668 0x7fb959d1b070 OCC::FolderMan::slotAccountStateChanged: Account "admin@xxxxxxxxxxxxx” disconnected, terminating or descheduling sync folders
02-18 09:14:40:668 0x7fb959d1b070 OCC::CheckServerJob::finished: error: status.php replied  0 ""
```

[...]

PR #4493, opened 2016-02-22
Disable sharing dialog if account state is not connected

### LINKED ISSUE #4185: [Share with users, groups] error OCS API error code: 0, when the server connection is down
### Steps to reproduce
1. Access to the Desktop client
2. Select to share a file
3. The server connection is down
4. Select to look for the user
5. Share the folder/file with user win

### Expected behaviour
the error should not appear

### Actual behaviour
You can this error: OCS API error code: 0,
screen shot 2015-11-24 at 11 33 07

### Server configuration
Desktop v ownCloud-2.1.0.5649-nightly20151124-setup.exe 
Server v {"installed":true,"maintenance":false,"version":"8.2.1.4","versionstring":"8.2.1","edition":"Enterprise"}

rullzer:
Fixes #4185 
And probably also #4484

@Dianafg76 please verify this

### DIFF
--- a/src/gui/owncloudgui.cpp
+++ b/src/gui/owncloudgui.cpp
@@ -777,7 +777,7 @@ void ownCloudGui::slotShowShareDialog(const QString &sharePath, const QString &l
         w = _shareDialogs[localPath];
     } else {
         qDebug() << Q_FUNC_INFO << "Opening share dialog" << sharePath << localPath;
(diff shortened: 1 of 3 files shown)


guruz:
:+1:

rullzer:
Rebased due to the theming conflict.

PR #4496, opened 2016-02-22
Add theming options control sharing operations

### LINKED ISSUE #4325: [Share options] I want to decide whether share with users is shown or not
AC:
- Include a variable that defines whether the share with user option is shown or not.
- In case it´s set not to be shown, on the UI it won't be shown the "share with users or groups" field

Bonus track:
- For parity with mobile apps it would be great to have another variable to define whether the share with link option is shown, being both of them independent. 

Reference:
https://github.com/owncloud/ownbrander/issues/483

rullzer:
Fixes #4325 

To test I just modified the settings in theme.cpp.
Is there anyway to test this?

CC: @Dianafg76 @rperezb @MTRichards

### DIFF
--- a/src/gui/sharedialog.cpp
+++ b/src/gui/sharedialog.cpp
@@ -92,10 +92,11 @@ ShareDialog::ShareDialog(AccountPtr account, const QString &sharePath, const QSt
         return;
     }
 
+    auto theme = Theme::instance();
     bool autoShare = true;
 
     // We only do user/group sharing from 8.2.0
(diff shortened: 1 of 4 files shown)


rperezb:
Thanks @rullzer 
@Dianafg76 this is important to be tested, I open a new ticket

ogoffart:
:+1:

guruz:
@rullzer We're a small team, one :+1: is enough I'd say.. merge on ;)

PR #4501, opened 2016-02-23
Do not send reshare permissions when creating a federated share

### LINKED ISSUE #4481: [Federated Sharing] Different permissions on the desktop that on the server
### Steps to reproduce
1. Go to the Desktop client with user1(server1)
2. Create a new folder (e.g Hello)
3. Share with owncloud
4. Write on share users/groups user2(server2) (e.g user2@..........)
5. The folder is shared with all permissions
screen shot 2016-02-17 at 11 51 15
6. Go to the Server web with user1(server1)
7. Click on the folder Hello
8. Share with owncloud

### Expected behaviour
different permissions on the desktop that on the server

It should have the same permissions 

### Actual behaviour
Only have share with one permission (CAN EDIT)
screen shot 2016-02-17 at 11 47 52

### Server configuration
Desktop v 2.1.1 build 3107
Server v {"installed":true,"maintenance":false,"version":"9.0.0.11","versionstring":"9.0 beta 1","edition":""}

rullzer:
See https://github.com/owncloud/core/issues/22122#issuecomment-185637344

Fixes #4481 

This patch makes sure we have the same behaviour on the client as on the server with regards to federated shares.

CC: @Dianafg76

### DIFF
--- a/src/gui/shareusergroupwidget.cpp
+++ b/src/gui/shareusergroupwidget.cpp
@@ -232,8 +232,21 @@ void ShareUserGroupWidget::slotCompleterActivated(const QModelIndex & index)
         return;
     }
 
-    _manager->createShare(_sharePath, Share::ShareType(sharee->type()),
-                          sharee->shareWith(), Share::PermissionDefault);
+    /*
+     * Don't send the reshare permissions for federataed shares
+     * https://github.com/owncloud/core/issues/22122#issuecomment-185637344
+     */
+    if (sharee->type() == Sharee::Federated) {
+        int permissions = Share::PermissionRead | Share::PermissionUpdate;
+        if (!_isFile) {
+            permissions |= Share::PermissionCreate | Share::PermissionDelete;
+        }
+        _manager->createShare(_sharePath, Share::ShareType(sharee->type()),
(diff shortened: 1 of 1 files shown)


src/gui/shareusergroupwidget.cpp
```diff
@@ -231,8 +231,21 @@ void ShareUserGroupWidget::slotCompleterActivated(const QModelIndex & index)
         return;
     }
 
-    _manager->createShare(_sharePath, Share::ShareType(sharee->type()),
-                          sharee->shareWith(), Share::PermissionDefault);
+    /*
+     * Don't send the reshare permissions for federataed shares
+     * https://github.com/owncloud/core/issues/22122#issuecomment-185637344
+     */
+    if (sharee->type() == Sharee::Federated) {
```
guruz: Those permission stuff is a big magic. Are you sure about it?

rullzer: Yes, sure. It is crappy right now. We are working in the server to improve this. But for now this is the servers behaviour so we should somehom mirror this.

guruz:
:+1:

PR #4509, opened 2016-02-24
improve client build instructions

carlaschroder:
Please review @danimo @dragotin

### DIFF
--- a/doc/building.rst
+++ b/doc/building.rst
@@ -30,11 +30,14 @@ you the exact sources from which the binary packages are built. These are
 hosted on the `ownCloud repository from OBS`_. Go to the `Index of 
 repositories`_ to see all the Linux client repos.
 
-1. At the `bottom of the page for each distribution <ownCloud repository from 
-   OBS>`_ is a "Grab binary packages directly" section. These contain source 
-   RPMs for CentOS, RHEL, Fedora, SLES, and openSUSE. except Debian and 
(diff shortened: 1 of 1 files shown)


jnweiger:
added some inline comments to
https://github.com/owncloud/client/commit/c13637b105acf740ccc3dae01505cc24a8acaffe

carlaschroder:
> From the generic-build-instructions we don't do the git commands, we only need:
> mkdir build-client; cd build-client; cmake -DCMAKE_BUILD_TYPE="Debug" ..; make

Where does the source code come from?

jnweiger:
Ah, good point, that needs explanation.
 
I added ``apt-get source owncloud-client;´´ to get the source. (need a zypper si, or yumdownloader --source for the others)

because the linux section says 
> You may wish to use source packages for your Linux distribution, as these give you the exact sources from which the binary packages are built. These are hosted on the ownCloud repository from OBS. Go to the Index of repositories to see all the Linux client repos.

The generic build instructions don't take their sources from source packages, they take them directly from github.

Both are valid approaches, depending on the current task:
* For feature development often the latest github source is best.
* For bug-hunting often the exact source from packages is needed.

carlaschroder:
Allrightythen, this version should make it all clear. Generic builds and Linux src packages are two different operations, so that should be all detailed correctly now. Thanks much for your help, @jnweiger.

Issue #4487, opened 2016-02-18
Checksums not stored in Journal


dragotin:
Currently the client does not store any checksums in the journal, even though the colums are prepared. It should keep the checksum that comes with a download, and also if it uploads a file.

@ckamm is that intended?

ckamm:
@dragotin It's because we haven't resolved all questions around how we want to do checksumming. See also some discussion in #4375. Storing the transmission checksum from the server when we don't compute any other checksum ourselves seems obviously good though.

ogoffart:
We store the content checksum as MD5 for .eml files only.

ogoffart:
With  https://github.com/owncloud/client/pull/4512  we store the checksum in the database if it is configured in the config file.  I don't know how usefull it is anyway since we don't use it for anything else that .eml files
@dragotin  what do you think?

ckamm:
As of #4512 transmission checksums are now stored in the db as content checksums. Enabling content checksums is left for #4375

ckamm:
I tested this by using mitmproxy with the filter `~s OC-Checksum MD5:...` with the checksum matching the content of a file I created on the server. Afterwards I checked the local database and saw the transfer checksum in there.

PR #4512, opened 2016-02-25
Checksums: keep the transfer checksum in the database as the content …

ogoffart:
…checksum

Currently, we only use this for .eml files. But we can just store this checksum
in the database if it was computed anyway.

Issue #4487


dragotin:
Yes please :+1:

ckamm:
Looks good to me!

PR #4513, opened 2016-02-25
Move the csync_context creation in SyncEngine

jturcotte:


src/libsync/excludedfiles.h
```diff
@@ -60,13 +63,10 @@ public slots:
...
      */
-    void reloadExcludes();
+    bool reloadExcludes();
 
 private:
-    ExcludedFiles();
-    ~ExcludedFiles();
-
-    c_strlist_t* _excludes;
+    c_strlist_t** _excludesPtr;
```
ogoffart: could there be a comment explaining what this pointer is to and what is the ownership and life time of it

src/libsync/syncengine.cpp
```diff
@@ -15,6 +15,7 @@
 
 #include "syncengine.h"
 #include "account.h"
+#include "logger.h"
```
ogoffart: oh no!  this one stuff should be kept in the gui.

src/libsync/syncengine.cpp
```diff
@@ -56,10 +57,17 @@ bool SyncEngine::_syncRunning = false;
 
 qint64 SyncEngine::minimumFileAgeForUpload = 2000;
 
-SyncEngine::SyncEngine(AccountPtr account, CSYNC *ctx, const QString& localPath,
-                       const QString& remoteURL, const QString& remotePath, OCC::SyncJournalDb* journal)
+static void csyncLogCatcher(int /*verbosity*/,
```
ogoffart: maybe keep this stuff in the logger itself.

src/libsync/syncengine.cpp
```diff
@@ -79,12 +87,31 @@ SyncEngine::SyncEngine(AccountPtr account, CSYNC *ctx, const QString& localPath,
...
+    // We need to reconstruct the url because the path needs to be fully decoded, as csync will re-encode the path:
+    //  Remember that csync will just append the filename to the path and pass it to the vio plugin.
+    //  csync_owncloud will then re-encode everything.
+#if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
+    QString url_string = _remoteUrl.scheme() + QLatin1String("://") + _remoteUrl.authority(QUrl::EncodeDelimiters) + _remoteUrl.path(QUrl::FullyDecoded);
+#else
+    // Qt4 was broken anyway as it did not encode the '#' as it should have done  (it was actually a problem when parsing the path from QUrl::setPath
+    QString url_string = _remoteUrl.toString();
+#endif
+    url_string = Utility::toCSyncScheme(url_string);
```
ogoffart: do we still need that actually?

src/libsync/syncengine.h
```diff
@@ -72,7 +73,9 @@ class OWNCLOUDSYNC_EXPORT SyncEngine : public QObject
      * -1 means infinite
      */
     void setNewBigFolderSizeLimit(qint64 limit) { _newBigFolderSizeLimit = limit; }
+    void setIgnoreHiddenFiles(bool ignore) { _csync_ctx->ignore_hidden_files = ignore; }
 
+    ExcludedFiles &excludedFiles() { return *_excludedFiles; }
```
ogoffart: a bit unusual to return a reference.

ogoffart:
yay, much code removed :-)

src/libsync/syncengine.h
jturcotte: Would you prefer a pointer? The reference makes the non-nullability clear though.

src/libsync/syncengine.cpp
jturcotte: Why, isn't Logger is in src/libsync?

src/libsync/syncengine.cpp
jturcotte: You tell me :P

ogoffart: I think not

ogoffart: since you removed the test for owncloud://  (which was required before to load the proper module, but now we can just use http)

src/libsync/syncengine.cpp
ogoffart: It is amongst the things that should mobe to gui

src/libsync/syncengine.cpp
danimo: If you remove the legacy support for `owncloud(s)://`, please add that to the ChangeLog and adjust the documentation accordingly.

src/libsync/syncengine.h
ogoffart: I'm fine with that.  I was just surprised to see excludedFiles().addExcludeFilePath()
It's just unusual in this codebase.

dragotin: I think its good, saves a if(..:). @ogoffart is it more usual in other codebases? Maybe we should consider that more often.

src/libsync/syncengine.cpp
```diff
@@ -889,7 +916,7 @@ void SyncEngine::slotDiscoveryJobFinished(int discoveryResult)
     _journal->commit("post treewalk");
 
     _propagator = QSharedPointer<OwncloudPropagator>(
-        new OwncloudPropagator (_account, _localPath, _remoteUrl, _remotePath, _journal));
+        new OwncloudPropagator (_account, _localPath, _remoteUrl.path(), _remotePath, _journal));
```
dragotin: I think changing the remoteUrl to QUrl is a dangerous thing. We need to make sure that _remoteUrl.path() still contains the same as the previous string (trailing slashes and stuff...) What benefit does this change bring?

jturcotte: It is the same, I just moved the .path() from Folder and OwnCloudCmd in here so that I could move the weird scheme logic above in here instead of outside. What we gain is that remoteUrl is an actual url (else it should be named remotePath, but we already have a remotePath, so anyway).

guruz:
:+1:

src/libsync/excludedfiles.h
jturcotte: Done

src/libsync/syncengine.cpp
jturcotte: Since I could just move that code and it's not technically on my way to get tests running, I'd rather leave it like that for now if that's OK.

ogoffart:
:+1:

PR #4514, opened 2016-02-26
factor out window raising method

sebasje:
The interesting change is moving a static function that only raises the dialog into its own class. This improves separation between syncing client and window system. The reason why I'm separating this one out is because it's going to be needed from my clientlib shared bits, such as the credentialstore, but would drag in the whole gui application, which would be defeating the purpose entirely.

This pull request also initializes an otherwise empty pointer, in a separate commit.

### DIFF
--- a/src/gui/windowraiser.h
+++ b/src/gui/windowraiser.h
@@ -1,0 +1,40 @@
+/*
+ * Copyright (C) by Klaas Freitag <freitag@owncloud.com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; version 2 of the License.
+ *
+ * This program is distributed in the hope that it will be useful, but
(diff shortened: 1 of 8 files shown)


sebasje:
The bot is wrong, I've signed and sent in the CLA.

dragotin:
like it :+1:

src/gui/windowraiser.h
```diff
@@ -0,0 +1,40 @@
...
+class QWidget;
+
+namespace OCC {
+
+
+/**
+ * @brief The WindowRaiser class
+ * @ingroup gui
+ */
+class WindowRaiser
```
ogoffart: This does not need to be a class.

We have lots of plain functions in the OCC namespace (for example: utility.h, filesystem.h)  so i think it's fine to have the raise dialog in a funciton without a class.

ogoffart:
:+1:    only that it dooes not really need to be in a class but could be its own function

guruz:
@sebasje Is this still something you'd want to do?

I also think it could just be a function with OCC:: namespace.

PR #4515, opened 2016-02-26
libsplit - second round: now featuring libsyncclient

sebasje:
I've cleaned up my "libsplit" branch which we discussed here: https://github.com/owncloud/client/pull/3975 

So currently, this branch contains:
- the separate library named now "libsyncclient"
- header changes which make csync_private.h build-time only required
- RPATH doesn't use the APPLICATION_EXECUTABLE anymore (this means that we don't use the name "owncloud" in third party applications loading this library

Two independent, but needed patches which makes this possible have been PR'ed #4514 separately.


sebasje:
The bot is wrong, I've signed and sent in the CLA.

danimo:
@karlitschek can you fix this, please?

karlitschek:
@sebasje Can you check again for an github email to confirm? Thanks. Let me know if this doesn't work!
@danimo Confirmed!

sebasje:
@karlitschek other than the bot-reply to this PR (and the other one), I haven't received any github confirmation. What should I look for exactly?

karlitschek:
@sebasje Strange. I have 'invited' you to the core-developer and desktop-developer groups. You have to accept this invitation somehow otherwise you are not a member. I can't force this. Are you sure there is no email? Or maybe a way to accept this invitation in the github notification or settings section somewhere?

sebasje:
@karlitschek After looking through my github settings, I can't find it. It should be listed under "organizations", once subscribed to this group, no?
When did you invite me to this group, just today, or is it already longer ago? (This would help narrowing down my search.)

Could you try inviting me again?

karlitschek:
I did it when you send the agreement and again today.

sebasje:
Ok, the email gets lost apparently. I dug up the one from december and that has worked out now. I'm not member of the ownCloud organization and the 'core developer', 'developer' and desktop-developer' teams.

Sorted. :-)

src/cmd/cmd.cpp
```diff
@@ -42,6 +43,8 @@
 #include <termios.h>
 #endif
 
+#include <csync_private.h>
```
ogoffart: I guess it does not need to be included twice.

src/gui/folder.cpp
```diff
@@ -32,8 +32,10 @@
 #include "theme.h"
 #include "filesystem.h"
 #include "excludedfiles.h"
+#include <csync_private.h>
 
 #include "creds/abstractcredentials.h"
+#include <csync_private.h>
```
ogoffart: Same

src/gui/folderwatcher.h
```diff
@@ -14,8 +14,6 @@
 #ifndef MIRALL_FOLDERWATCHER_H
 #define MIRALL_FOLDERWATCHER_H
 
-#include "config.h"
```
ogoffart: nice :-)

src/libsync/discoveryphase.cpp
```diff
@@ -13,13 +13,15 @@
...
 #include "discoveryphase.h"
-#include <csync_private.h>
 #include <qdebug.h>
+#include <csync_private.h>
 
 #include <QUrl>
 #include "account.h"
 #include <QFileInfo>
 
+#include <csync_private.h>
```
ogoffart: no need to have it twice

ogoffart: Duplicate

src/cmd/cmd.cpp
ogoffart: Duplicate

ogoffart:
Looks good to me. appart from the duplicate csync_private entry. But it's minor.

sebasje:
Ping ... can this be merged so I can prepare the next set of patches? (This one in would already be a huge improvement, since it means I can build against upstream owncloudclient, not my hacked clone.)

ogoffart:
looks good to me. 

 @danimo: any comments before i merge?

src/libsync/CMakeLists.txt
```diff
@@ -103,6 +111,10 @@ IF (NOT APPLE)
         FILES ${creds_HEADERS}
 	DESTINATION ${INCLUDE_INSTALL_DIR}/${synclib_NAME}/creds
     )
+    INSTALL(
+        FILES ${CMAKE_CURRENT_BINARY_DIR}/owncloudlib.h
+        DESTINATION ${INCLUDE_INSTALL_DIR}/${synclib_NAME}
+    )
```
dragotin: Why is the owncloudlib.h only installed when not on Apple? That will break the FindOwncloudSync.cmake on apple, or not?

cmake/modules/FindLibSync.cmake
```diff
@@ -0,0 +1,37 @@
+# (c) 2014 Copyright ownCloud, Inc.
+# Redistribution and use is allowed according to the terms of the BSD license.
+# For details see the accompanying COPYING* file.
+
+# - Try to find QtKeychain
+# Once done this will define
+#  OWNCLOUDSYNC_FOUND - System has QtKeychain
+#  OWNCLOUDSYNC_INCLUDE_DIRS - The QtKeychain include directories
+#  OWNCLOUDSYNC_LIBRARIES - The libraries needed to use QtKeychain
+#  OWNCLOUDSYNC_DEFINITIONS - Compiler switches required for using LibXml2
```
dragotin: please do not name the variables starting OWNCLOUDSYNC_* but LIBSYNC_*

We need to be brandable where we do not wanna see the name ownCloud anywhere. Not so hot here, admitted, but still ;-)

csync/src/CMakeLists.txt
```diff
@@ -104,34 +102,22 @@ set_target_properties(
...
-    )
-else()
-  INSTALL(
+
+INSTALL(
   TARGETS
     ${CSYNC_LIBRARY}
   LIBRARY DESTINATION
-    ${LIB_INSTALL_DIR}/${APPLICATION_EXECUTABLE}
+    ${LIB_INSTALL_DIR}
```
dragotin: This is only valid if the csync_library is now a public lib. Does it have to be? This has effects on our packaging as well.

src/cmd/CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ if(NOT BUILD_LIBRARIES_ONLY)
 	set_target_properties(${cmd_NAME} PROPERTIES
 	        RUNTIME_OUTPUT_DIRECTORY  ${BIN_OUTPUT_DIRECTORY} )
         set_target_properties(${cmd_NAME} PROPERTIES
-		INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/${LIB_INSTALL_DIR}/${APPLICATION_EXECUTABLE}" )
+		INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/${LIB_INSTALL_DIR}" )
```
dragotin: I am not sure if we have to install an rpath at all if a standard lib path is used. I think rpath is only neccessary if we ship the lib in a private dir.

src/crashreporter/CMakeLists.txt
```diff
@@ -36,7 +36,7 @@ if(NOT BUILD_LIBRARIES_ONLY)
 
     set_target_properties(${CRASHREPORTER_EXECUTABLE} PROPERTIES AUTOMOC ON)
     set_target_properties(${CRASHREPORTER_EXECUTABLE} PROPERTIES RUNTIME_OUTPUT_DIRECTORY  ${BIN_OUTPUT_DIRECTORY} )
-    set_target_properties(${CRASHREPORTER_EXECUTABLE} PROPERTIES INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/${LIB_INSTALL_DIR}/${APPLICATION_EXECUTABLE}" )
+    set_target_properties(${CRASHREPORTER_EXECUTABLE} PROPERTIES INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/${LIB_INSTALL_DIR}" )
```
dragotin: Again rpath: not sure if it is needed if the lib is in a standard library path.

ckamm:
@sebasje There seems to have been no progress on this for a long while. I'd close it unless you're up for the significant task of resurrecting and updating it.

Issue #3829, opened 2015-09-14
Add option to disable/pause a ownCloud account


schiesbn:
Just an idea but maybe it would be a nice small feature:
If you sync multiple accounts with the clients it can happen that you don't always want to sync all accounts. Therefore it would be nice if the user could disable/enable different ownCloud accounts.

guruz:
You should be able to disable an individual sync folder with the ```...``` button.

Would it help you to have this option also in the tray menu for the whole account instead of the sync folder?

FYI @jancborchardt

schiesbn:
@guruz ah, nice... well hidden feature. :wink:  Putting it in the tray menu sounds like a good idea to me.

Maybe also the name should be different? "pause sync" sounds to me like something you do only for a really short period of time and which gets reseted if you restart the client. I would rather call it "disable account" and "enable account", or something similar. @jancborchardt what do you think?

jancborchardt:
The usecases for pausing/disabling vary, and so does the time. Other apps also call it something like Pause so I would stick with that. :) We have that since some time and haven’t had any complaints so far about it not being understandable.

schiesbn:
I only looked for it now that we have multi-account support. Because I don't want that my devel account syncs all the time but only when I need it. I wouldn't have thought that "pause sync" is what I'm looking for. I looked for a option to enable/disable the account. Anyway, I can live with the name but it would be nice if would be easier to find, like suggested by @guruz

michaelstingl:
I would need a pause button in the tray menu to pause ALL syncing to save my 3G data plan.

00004038

michaelstingl:
Does this interfere with https://github.com/owncloud/client/issues/1971 ?

r2evans:
My vote/suggestion:

I prefer the label of "Pause" to "disable", personally, and since the pause state persists across app restarts (so, if paused when the machine shuts down or reboots, it will be paused when it starts up again), I'm generally happy with its current behavior.

I do like the idea of adding an option to the taskbar, something like this:

![ocpause](https://cloud.githubusercontent.com/assets/5815808/11321219/4547e6e0-906b-11e5-9f47-4708bd22b0a0.png)

MTRichards:
@dragotin @danimo something to consider for 2.2.

danimo:
@MTRichards I'd rather fix the server...

MTRichards:
We already have the pause button, all I want to consider is the request for putting it in the context menu. It is a blue ticket.

[...]

PR #4527, opened 2016-03-01
Account pausing #3829

ckamm:
It's a patch for #3829 plus some improvements to how accounts are written to and read from settings.


guruz:
looks OK :+1: 

A lot of combinatorism... *sigh*.

Issue #4081, opened 2015-11-05
Bandwidth setting "limit automatically" ineffective with small files


cloo:
When uploading lots of small files with bandwidth setting "limit automatically", uploading becomes unreasonably slow.

To describe the difference, with this setting off, about 10 similar files are uplaoded per second (10-50 kb in size), and with this setting on - it takes one to two seconds per file.

Maybe available bandwidth is measured after each individual file and happens way too often - or something like that.

guruz:
@cloo Besides bandwidth measuring/limiting, the parallelism is going down from 3 to 1 parallel request. For small files, the network latency and ownCloud server per-request slowness will decide on your speed.

Which server version do you use?

cloo:
I'm using server version 8.2.

guruz:
Please note that vanilla 8.2 has some performance regressions, especially if you don't use the Redis locking. The update will fix this.

https://doc.owncloud.org/server/8.1/admin_manual/configuration_files/files_locking_transactional.html
https://github.com/owncloud/core/issues/19888#issuecomment-153696788 (and others)

guruz:
@cloo btw, have you tried meanwhile with a newer server and the redis locking?

hodyroff:
No answer in regards of newer version of the server and redis. Closing for now.

Issue #331, opened 2013-02-12
Upload of many Files really slow


myzinsky:
### Expected behaviour

Upload of a lot of MP3 Files (4GB) should be fast.
### Actual behaviour

Upload Only 6 Files (ca 40mb) in 4 hours!
### Server configuration

Operating system: Debian Stable

Web server: Apache2

Database: Mysql

PHP version: PHP Version 5.3.3-7+squeeze14

ownCloud version:  4.5.6 
### Client configuration

Client version: 1.2

Operating system: Debian Unstable

OS language: German

Installation path of client: /usr/bin/owncloud
### Logs

ownclouds://xxx/Blue Moon
02-12 14:18:20:533 oc_module: owncloud_stat ownclouds://xxx/Blue Moon called
02-12 14:18:20:678 oc_module: Simple propfind result code 207.
02-12 14:18:20:678 oc_module: Server Date from HTTP header value: Tue, 12 Feb 2013 13:18:20 GMT
02-12 14:18:20:679 oc_module: Ok: Time delta remained (almost) the same: 0.
02-12 14:18:20:679 oc_module: => Errno after fetch resource list for ownclouds://xxx/Blue Moon: 0
02-12 14:18:20:679 oc_module: Working on file Blue Moon
02-12 14:18:20:679 oc_module:   :> Subtracting 0 from modtime 1360675100
02-12 14:18:20:679 oc_module: STAT result from propfind: Blue Moon, mtime: 1360675100
02-12 14:18:20:679 oc_module: Directory of file to open exists.
02-12 14:18:20:679 oc_module: PUT request on //xxx/Blue%20Moon/Robben%20Ford%20-%20My%20Everything.mp3!
02-12 14:18:20:679 oc_module: Sendfile handling request type PUT.
02-12 14:18:20:679 oc_module: Put file size: 6858752, variable sizeof: 8

This is an example, it waits a really loooong time :( (Paths shortened with xxx)
#### Web server error log

<b>Deprecated</b>:  Directive 'register_long_arrays' is deprecated in PHP 5.3 and greater in <b>Unknown</b> on line <b>0</b><br />

There is a lot of this stuff, but i'm not shure if this is realted to owncloud
#### ownCloud log (data/owncloud.log)

{"app":"media","message":"error reading title tag in '\/var\/www\/owncloud\/xxx.mp3'","level":2,"time":1360575380}### Expected behaviour
Upload of a lot of MP3 Files (4GB) should be fast.

[...]

PR #4529, opened 2016-03-01
Propagator: Pump in more requests if we think current ones are quick

guruz:
Helps with small file sync #331
When I benchmarked this, it went up to 6 parallelism and
was about 1/3 faster than the previous fixed 3 parallelism.
Doing more than 6 is dangerous because QNAM limits to 6 TCP
connections and also the server might become a bottleneck.

Should also help for #4081

### DIFF
--- a/src/libsync/owncloudpropagator.cpp
+++ b/src/libsync/owncloudpropagator.cpp
@@ -87,6 +87,16 @@ int OwncloudPropagator::maximumActiveJob()
     return max;
 }
 
+int OwncloudPropagator::hardMaximumActiveJob()
+{
+    int max = maximumActiveJob();
+    return max*2;
+    // FIXME: Wondering if we should hard-limit to 1 if maximumActiveJob() is 1
+    // to support our old use case of limiting concurrency (when "automatic" bandwidth
+    // limiting is set. But this causes https://github.com/owncloud/client/issues/4081
+}
+
+
 /** Updates, creates or removes a blacklist entry for the given item.
  *
  * Returns whether the file is in the blacklist now.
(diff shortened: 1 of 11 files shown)


src/libsync/owncloudpropagator.cpp
```diff
@@ -518,11 +528,34 @@ QString OwncloudPropagator::getFilePath(const QString& tmp_file_name) const
...
+    // need to check how to avoid this.
+    // Down-scaling on slow networks? https://github.com/owncloud/client/issues/3382
+    // Making sure we do up/down at same time? https://github.com/owncloud/client/issues/1633
+
+    if (this->_activeJobList.count() < maximumActiveJob()) {
         if (_rootJob->scheduleNextJob()) {
             QTimer::singleShot(0, this, SLOT(scheduleNextJob()));
         }
     }
+    else if (this->_activeJobList.count() < hardMaximumActiveJob()){
```
ckamm: minor: this-> is unnecessary here and above

src/libsync/propagatedownload.cpp
```diff
@@ -731,6 +732,11 @@ void PropagateDownloadFileQNAM::downloadFinished()
     if(_item->_file == QLatin1String(".sys.admin#recall#") || _item->_file.endsWith("/.sys.admin#recall#")) {
         handleRecallFile(fn);
     }
+
+    qint64 duration = _stopwatch.elapsed();
+    if (isLikelyFinishedQuickly() && duration > 5*1000) {
+        qDebug() << "WARNING: Unexpectedly slow connection";
```
ckamm: maybe add information about what kind of job it was and how long it took?

ckamm:
Looks good, only minor comments!

guruz:
@ckamm Thanks, merged with your recommended changes

rienesl:
Hm, there are still timeout problems when using parallelism with big files on slow connections, so it would be really nice, if you could let decide the user, how much parallelism he want's (dropdownbox?)!

dragotin:
I think we need some specific QA for this, @guruz what do you think? Could you add some hints how to test and compare that to the state before to the test instructions in the wiki?

davidpablomitchell:
hello team, where and how i and change parallelism or any manner to use all my bandwidth???? thank you

guruz:
@davidpablomitchell hi! it should already use all your bandwidth. Did you accidently activate the bandwidth limit in the settings?
What are your symptoms?
Which OS? What do you upload/download?

davidpablomitchell:
when i have lot of files i see the bandwidth is not all o have

davidpablomitchell:
Hello sorry i download a bug file 1.5TB from server to client and my connection don't pass 30mb how can i use my line of 300mb with other type of software i can use 300mb downloading.
thank you

PR #4553, opened 2016-03-08
Add spinner to show we are creating the share

### LINKED ISSUE #3737: Enable Full Sharing (user/group) from Desktop

rullzer:
Fixes #3737

The rest of the sharing stuff I consider new enhancements and/or bugfixes.

CC: @guruz @ckamm @dragotin

### DIFF
--- a/src/gui/shareusergroupwidget.cpp
+++ b/src/gui/shareusergroupwidget.cpp
@@ -233,6 +233,15 @@ void ShareUserGroupWidget::slotCompleterActivated(const QModelIndex & index)
     }
 
     /*
+     * Add spinner to the bottom of the widget list
+     */
+    auto viewPort = _ui->scrollArea->widget();
+    auto layout = viewPort->layout();
+    auto indicator = new QProgressIndicator(viewPort);
+    indicator->startAnimation();
+    layout->addWidget(indicator);
+
+    /*
      * Don't send the reshare permissions for federataed shares
      * https://github.com/owncloud/core/issues/22122#issuecomment-185637344
      */


dragotin:
Where is the spinner stopped again?

rullzer:
Nowhere... we reload the shares and create a new widget there..  so the old one is removed.

dragotin:
ok :+1:

Issue #4439, opened 2016-02-05
Windows: Changing time zone changes timestamp in the sync journal


dragotin:
@msrex had an alarming finding on his windows installation: For some reason the client tried to re-upload a lot of data.

I was debugging and found out that the sync journal contains a last-modified timestamp that is exactly fife hours away from the real UTC timestamp. The time zone of the windows machine is set to ETC, which exactly is fife hours away from UTC. 

The header that is written to the sync journal is taken from the HTTP last-modified header during download job. I tunneled that through mitmproxy and it shows that the Last-modified header is correctly on UTC. 

The problem disappeared when we rebooted the windows machine. However, if the timezone is changed on windows, the timestamps are changed again. That is also visible in the Activity Tab which also uses parsed timestamps, in that case from JSON format.

The suspicion is that the `QVariant::toDateTime` function is malfunctioning somehow.

I suggest to keep this as a gold ticket for 2.1.1 until we know if that is a regression and how this is triggered. 

@ogoffart for help please.

@cmonteroluque and @MTRichards FYI.

dragotin:
maybe related: https://github.com/notepad-plus-plus/notepad-plus-plus/pull/792

Maybe we also need to add the _tzset() when stat'ing?

ogoffart:
> The header that is written to the sync journal is taken from the HTTP last-modified header during download job

And then we set the modtime to this file, and then we check again what is the modtime.  (in PropagateDownloadFileQNAM::downloadFinished)

```
    FileSystem::setModTime(_tmpFile.fileName(), _item->_modtime);
    // We need to fetch the time again because some file systems such as FAT have worse than a second
    // Accuracy, and we really need the time from the file system. (#3103)
    _item->_modtime = FileSystem::getModTime(_tmpFile.fileName());

```

What was the version and the build of the client? 
Do we have any logs?
How does the file system time compare to the server time and to the database time? Where did it go wrong?
Was the timezone changed while the program was running?

The `Activity Tab`  issue  might be totally unrelated

ogoffart:
Note that there was a bug in FileSystem::getModTime on windows in 2.1.0 which was fixed shortly after the release in (fixed in commit cf5b1e401c0e0c8284bcfd8c12721a09175ed1b4 )

dragotin:
[...]

Issue #3733, opened 2015-09-02
Sending User Asynchronous Notifications from the Admin


MTRichards:
Notifications
As an admin, I want to be able to send notifications from the server to any owncloud client (desktop, client, mobile) so that I can send messages, and receive confirmation, from users of the ownCloud system

Note: this has several parts: desktop, server, web and mobile

Acceptance Criteria:
- A new app on the server (system notifications) can be enabled by the administrator
- When this app is enabled, the clients ping the server on every sync run to see if a new system wide notification is waiting
- If a new notification is waiting, the client downloads the message XML, parses the error, and displays it in a window for the desktop client
- Notifications can be of 2 types initially: those requiring acceptance and those that are simply push notifications
- A simple push notification simply displays the message set by the administrator
- A request for yes / no response is not necessary in this version of ownCloud client, but will be needed at a later date
- If the web client, it is posted in a new popup; if mobile, a new message tag is added to a new system messages view.

guruz:
FYI @nickvergessen

nickvergessen:
The server side will be done with https://github.com/owncloud/notifications and https://github.com/nickv-oc/announcementcenter
So nothing special needs to be done on the client. Just display and have the actions of the normal notifications.

ogoffart:
Is there any documentation of the API?
A running server with that installed so we can test would be nice.

nickvergessen:
This is currently on hold, until the activity API is settled, ref https://github.com/owncloud/activity/issues/405 . Notifications will mostlikly use a similar mechanism.

nickvergessen:
The docs for the OCS API endpoint for 9.0 are now done and settled:
https://github.com/owncloud/notifications/blob/master/docs/ocs-endpoint-v1.md

Master branch of the notifications app already gives that output.
As described above it can be used together with the announcement center app.
Just enable both in the apps, make an announcement as an admin and then
read from the OCS endpoint as a different user.

dragotin:
Affected by owncloud/core/22786

dragotin:
pull request #4567 was merged.

@carlaschroder we need documentation about that, just as a reminder. I will provide input in the wiki.

SergioBertolinSG:
Works in OS X (10.11). Version 2.2.0-nightly20160408 (build 3250)
Works in Windows 10. 2.2.0-nightly20160408 (build 5980)

Not sure if checking this on mobile has been done before, has it @jesmrec ?

SergioBertolinSG:
[...]

PR #4567, opened 2016-03-11
Show Server Notifications on the Client 

dragotin:
This patch grew big already, before it grows even bigger, here is the pull request.

It implements the notifications from the server as requested in #3733 

Still missing:
* [ ] Proper icon
* [x] A regular check on new notifications
* [x] Make the notification widget disappear after a while when a button was clicked.

@danimo @ogoffart @ckamm @guruz


src/gui/activitywidget.cpp
```diff
@@ -400,6 +214,154 @@ void ActivityWidget::slotOpenFile(QModelIndex indx)
...
+    }
+
+    _ui->_notifyLabel->setHidden( _widgetForNotifId.isEmpty() );
+    _ui->_notifyScroll->setHidden( _widgetForNotifId.isEmpty() );
+
+    if( newGuiLogs > 0 ) {
+        // restart the gui log timer now that we show a notification
+        _guiLogTimer.restart();
+
+        // Assemble a tray notification
```
danimo: I don't think you can build strings this way. Even with translator comments This will break in all kind of funny ways e.g. with RTL languages. Also, with too many notifiers, the message cannot be parsed fast enough, so we should probably limit the amount of displayed notification sources (?) as well. You at have to account for three cases, with the n=2 one being optional:

* `n =1` (You received %n notifications from %1).arg(notifier)
* `n = 2` (You received %n notifications from %1 and %2).arg(notifiers).
* `n > 2` (You received %n notifications from %1, %2 and others).arg(notifiers.at(0), notifiers.at(1). (ideally, we would specify a quantity for "others", but we cannot have more than one plural form in one tr()...

src/gui/activitylistmodel.cpp
```diff
@@ -0,0 +1,237 @@
...
+                     this, SLOT(slotActivitiesReceived(QVariantMap, int)));
+    job->setProperty("AccountStatePtr", QVariant::fromValue<AccountState*>(s));
+
+    QList< QPair<QString,QString> > params;
+    params.append(qMakePair(QString::fromLatin1("page"),     QString::fromLatin1("0")));
+    params.append(qMakePair(QString::fromLatin1("pagesize"), QString::fromLatin1("100")));
+    job->addQueryParams(params);
+
+    _currentlyFetching.insert(s);
+    qDebug() << "Start fetching activities for " << s->account()->displayName();
```
danimo: Remove?

src/gui/activitylistmodel.cpp
```diff
@@ -0,0 +1,237 @@
...
+    job->start();
+}
+
+void ActivityListModel::slotActivitiesReceived(const QVariantMap& json, int statusCode)
+{
+    auto activities = json.value("ocs").toMap().value("data").toList();
+    // qDebug() << "*** activities" << activities;
```
danimo: Remove?

src/gui/activitylistmodel.cpp
```diff
@@ -0,0 +1,237 @@
...
+        if( newItem ) {
+            startFetchJob(asp.data());
+        }
+    }
+}
+
+void ActivityListModel::slotRefreshActivity(AccountState *ast)
+{
+    if(ast && _activityLists.contains(ast)) {
+        qDebug() << "**** Refreshing Activity list for" << ast->account()->displayName();
```
danimo: Remove?

src/gui/activitywidget.cpp
```diff
@@ -300,6 +100,18 @@ ActivityWidget::~ActivityWidget()
...
+    // start a server notification handler if no notification requests
+    // are running
+    if( _notificationRequests == 0 ) {
+        ServerNotificationHandler *snh = new ServerNotificationHandler;
+        connect(snh, SIGNAL(newNotificationList(ActivityList)), this,
+                SLOT(slotBuildNotificationDisplay(ActivityList)));
+
+        snh->slotFetchNotifications(ptr);
+    } else {
+        qDebug() << "========> notification request counter not zero.";
```
danimo: Remove?

src/gui/activitywidget.cpp
```diff
@@ -400,6 +214,154 @@ void ActivityWidget::slotOpenFile(QModelIndex indx)
...
+        widget->setAccountName( activity._accName );
+        widget->setActivity( activity );
+
+        // handle gui logs. In order to NOT annoy the user with every fetching of the
+        // notifications the notification id is stored in a Set. Only if an id
+        // is not in the set, it qualifies for guiLog.
+        // Important: The _guiLoggedNotifications set must be wiped regularly which
+        // will repeat the gui log.
+
+        // after one hour, clear the gui log notification store
```
danimo: Why is this a good choice? How can we be sure that the ids will not reappear after one hour?

Also, pls move the magic values out of the code flow.

src/gui/servernotificationhandler.cpp
```diff
@@ -0,0 +1,103 @@
...
+    /* start the notification fetch job as well */
+    if( !ptr) {
+        return;
+    }
+
+    // check if the account has notifications enabled. If the capabilities are
+    // not yet valid, its assumed that notifications are available.
+    if( ptr->account() && ptr->account()->capabilities().isValid() ) {
+        if( ! ptr->account()->capabilities().notificationsAvailable() ) {
+            qDebug() << "Account" << ptr->account()->displayName() << "does not have notifications enabled.";
```
danimo: Please remove or use qDebug() << Q_FUNC_INFO consistent with the rest of the general code base. We should probably clean this up.

src/gui/activitywidget.cpp
```diff
@@ -400,6 +214,154 @@ void ActivityWidget::slotOpenFile(QModelIndex indx)
...
+        msg += addM;
+
+        emit guiLog(Theme::instance()->appNameGUI() + QLatin1String(" ") + tr("Notifications - Action Required"),
+                    msg);
+    }
+}
+
+void ActivityWidget::slotSendNotificationRequest(const QString& accountName, const QString& link, const QString& verb)
+{
+    qDebug() << "Server Notification Request " << verb << link << "on account" << accountName;
```
danimo: Remove?

src/gui/activitywidget.cpp
```diff
@@ -400,6 +214,154 @@ void ActivityWidget::slotOpenFile(QModelIndex indx)
...
+            connect( job, SIGNAL( jobFinished(QString, int)),
+                     this, SLOT(slotNotifyServerFinished(QString, int)) );
+            job->start();
+
+            // count the number of running notification requests. If this member var
+            // is larger than zero, no new fetching of notifications is started
+            _notificationRequests++;
+        }
+    } else {
+        qDebug() << "Notification Links: Invalid verb:" << verb;
```
danimo: Remove or `Q_FUNC_INFO`

src/gui/activitywidget.cpp
```diff
@@ -400,6 +214,154 @@ void ActivityWidget::slotOpenFile(QModelIndex indx)
...
+        return;
+    }
+
+    int resultCode =0;
+    if( reply ) {
+        resultCode = reply->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
+    }
+
+    endNotificationRequest(job->widget(), resultCode);
+    qDebug() << "Server notify job failed.";
```
danimo: Remove or `Q_FUNC_INFO`

src/gui/activitywidget.cpp
```diff
@@ -400,6 +214,154 @@ void ActivityWidget::slotOpenFile(QModelIndex indx)
...
+void ActivityWidget::slotNotifyServerFinished( const QString& reply, int replyCode )
+{
+    NotificationConfirmJob *job = qobject_cast<NotificationConfirmJob*>(sender());
+    if( !job ) {
+        return;
+    }
+
+    endNotificationRequest(job->widget(), replyCode);
+    // FIXME: remove the  widget after a couple of seconds
+    qDebug() << "Server Notification reply code"<< replyCode << reply;
```
danimo: Remove or `Q_FUNC_INFO`

src/gui/notificationwidget.cpp
```diff
@@ -0,0 +1,107 @@
...
+            if( _buttons.at(i) == buttonWidget ) {
+                index = i;
+            }
+            _buttons.at(i)->setEnabled(false);
+        }
+
+        // if the button was found, the link must be called
+        if( index > -1 && index < _myActivity._links.count() ) {
+            ActivityLink triggeredLink = _myActivity._links.at(index);
+            qDebug() << "Notification Link: "<< triggeredLink._verb << triggeredLink._link;
```
danimo: Remove or `Q_FUNC_INFO`

src/gui/servernotificationhandler.cpp
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+
+    // if the previous notification job has finished, start next.
+    _notificationJob = new JsonApiJob( ptr->account(), QLatin1String("ocs/v2.php/apps/notifications/api/v1/notifications"), this );
+    QObject::connect(_notificationJob.data(), SIGNAL(jsonReceived(QVariantMap, int)),
+                     this, SLOT(slotNotificationsReceived(QVariantMap, int)));
+    _notificationJob->setProperty("AccountStatePtr", QVariant::fromValue<AccountState*>(ptr));
+
+    qDebug() << "Start fetching notifications for " << ptr->account()->displayName();
```
danimo: Remove or `Q_FUNC_INFO`

src/gui/servernotificationhandler.cpp
```diff
@@ -0,0 +1,103 @@
...
+    _notificationJob->start();
+}
+
+void ServerNotificationHandler::slotNotificationsReceived(const QVariantMap& json, int statusCode)
+{
+    if( statusCode != 200 ) {
+        qDebug() << "Failed for Notifications";
```
danimo: Remove or `Q_FUNC_INFO`

src/gui/servernotificationhandler.cpp
```diff
@@ -0,0 +1,103 @@
...
+        return;
+    }
+
+    auto notifies = json.value("ocs").toMap().value("data").toList();
+
+    AccountState* ai = qvariant_cast<AccountState*>(sender()->property("AccountStatePtr"));
+
+    // qDebug() << "Notifications for " << ai->account()->displayName() << notifies;
```
danimo: Remove.

src/gui/notificationwidget.cpp
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+}
+
+void NotificationWidget::slotNotificationRequestFinished(int statusCode)
+{
+    int i = 0;
+    // the ocs API returns stat code 100 if it succeeded.
+    if( statusCode != 100 ) {
+        qDebug() << "Notification Request to Server failed, leave button visible.";
```
danimo: Remove or `Q_FUNC_INFO`

src/gui/notificationwidget.cpp
```diff
@@ -0,0 +1,107 @@
...
+            emit sendNotificationRequest( _accountName, triggeredLink._link, triggeredLink._verb );
+        }
+    }
+}
+
+void NotificationWidget::slotNotificationRequestFinished(int statusCode)
+{
+    int i = 0;
+    // the ocs API returns stat code 100 if it succeeded.
+    if( statusCode != 100 ) {
```
danimo: rather use constants for this?

src/gui/activitydata.h
```diff
@@ -0,0 +1,106 @@
...
+ * @brief The ActivityLink class describes actions of an activity
+ *
+ * These are part of notifications which are mapped into activities.
+ */
+
+class ActivityLink
+{
+public:
+    QHash <QString, QVariant> toVariantHash() {
+        QHash<QString, QVariant> hash;
```
ckamm: use QVariantHash (introduced in Qt 4.5)?

src/gui/activitydata.h
```diff
@@ -0,0 +1,106 @@
...
+};
+
+/* ==================================================================== */
+/**
+ * @brief The ActivityList
+ * @ingroup gui
+ *
+ * A QList based list of Activities
+ */
+class ActivityList:public QList<Activity>
```
ckamm: If inheriting here isn't mandatory for some reason, I'd prefer to have the list as a member.

src/gui/activitylistmodel.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+    if (!index.isValid())
+        return QVariant();
+
+    a = _finalList.at(index.row());
+    AccountStatePtr ast = AccountManager::instance()->account(a._accName);
+    QStringList list;
+
+    if (role == Qt::EditRole)
+        return QVariant();
```
ckamm: Remove? (or put next to the `default` case in the switch stmt)

src/gui/activitylistmodel.cpp
```diff
@@ -0,0 +1,237 @@
...
+{
+    auto activities = json.value("ocs").toMap().value("data").toList();
+    // qDebug() << "*** activities" << activities;
+
+    ActivityList list;
+    AccountState* ast = qvariant_cast<AccountState*>(sender()->property("AccountStatePtr"));
+    _currentlyFetching.remove(ast);
+    list.setAccountName( ast->account()->displayName());
+
+    foreach( auto activ, activities ) {
```
ckamm: const auto &, to avoid unnecessary copies of activ

src/gui/activitylistmodel.cpp
```diff
@@ -0,0 +1,237 @@
...
+    QList<AccountStatePtr> accounts = AccountManager::instance()->accounts();
+
+    foreach (AccountStatePtr asp, accounts) {
+        bool newItem = false;
+
+        if( !_activityLists.contains(asp.data()) && asp->isConnected() ) {
+            _activityLists[asp.data()] = ActivityList();
+            newItem = true;
+        }
+        if( newItem ) {
```
ckamm: `newItem` seems unnecessary.

src/gui/activitylistmodel.cpp
```diff
@@ -0,0 +1,237 @@
...
+{
+    ActivityList resultList;
+
+    foreach( ActivityList list, _activityLists.values() ) {
+        resultList.append(list);
+    }
+
+    std::sort( resultList.begin(), resultList.end() );
+
+    beginInsertRows(QModelIndex(), 0, resultList.count()-1);
```
ckamm: What if there used to be more rows before? Does one need to explicitly clear the other rows?

dragotin:
Considered your feedback @danimo and @ckamm  - thanks a lot!

dragotin:
@danimo fixed. Note that the other place below always deals with item number > 2, so no real plural handling needed, its always plural.

ckamm:
@dragotin I only reviewed about half of the changes yesterday, I'll go through the rest today.

src/gui/servernotificationhandler.cpp
```diff
@@ -0,0 +1,100 @@
...
+        a._accName  = ai->account()->displayName();
+        a._id       = json.value("notification_id").toLongLong();
+        a._subject  = json.value("subject").toString();
+        a._message  = json.value("message").toString();
+        QString s   = json.value("link").toString();
+        if( !s.isEmpty() ) {
+            a._link     = QUrl(s);
+        }
+        a._dateTime = json.value("datetime").toDateTime();
+        a._dateTime.setTimeSpec(Qt::UTC);
```
ckamm: I'm not convinced that this setTimeSpec is a correct. Doesn't the ISO date string already denote the timezone (usually UTC)? And then this would either be a noop or, if the current timezone is different, change the dateTime to an incorrect value?

It does depend a lot on what the server communicates exactly.

src/gui/notificationwidget.h
```diff
@@ -0,0 +1,55 @@
...
+
+namespace OCC {
+
+class NotificationWidget : public QWidget
+{
+    Q_OBJECT
+public:
+    explicit NotificationWidget(QWidget *parent = 0);
+
+    void setAccountName( const QString& name );
```
ckamm: This function is unnecessary, since setActivity() also sets the account name based on the activity.

src/gui/notificationconfirmjob.h
```diff
@@ -0,0 +1,80 @@
...
+     * @param verb currently supported GET PUT POST DELETE
+     */
+    void setLinkAndVerb(const QUrl& link, const QString &verb);
+
+    /**
+     * Start the OCS request
+     */
+    void start() Q_DECL_OVERRIDE;
+
+    void setWidget( NotificationWidget *widget );
```
ckamm: Maybe add a brief comment on what this widget is about? (it's just stored for convenience) It's a bit surprising for a network job to directly have a widget. :)

src/gui/activitywidget.cpp
```diff
@@ -400,6 +214,154 @@ void ActivityWidget::slotOpenFile(QModelIndex indx)
...
+            job->setWidget(theSender);
+            connect( job, SIGNAL( networkError(QNetworkReply*)),
+                                  this, SLOT(slotNotifyNetworkError(QNetworkReply*)));
+            connect( job, SIGNAL( jobFinished(QString, int)),
+                     this, SLOT(slotNotifyServerFinished(QString, int)) );
+            job->start();
+
+            // count the number of running notification requests. If this member var
+            // is larger than zero, no new fetching of notifications is started
+            _notificationRequests++;
```
ckamm: What if this happens:
1) Old notifications were fetched
2) Fetching current notifications starts
3) NotificationConfirmJob created
4) New notifications come in, replacing the in-progress notifications

We probably need a guard somewhere that ignores incoming notifications while a confirmation job is pending?

src/gui/activitywidget.cpp
```diff
@@ -400,6 +214,154 @@ void ActivityWidget::slotOpenFile(QModelIndex indx)
...
+{
+    QHash<QString, int> accNotified;
+
+    foreach( auto activity, list ) {
+        NotificationWidget *widget = 0;
+
+        if( _widgetForNotifId.contains(activity._id) ) {
+            widget = _widgetForNotifId[activity._id];
+        } else {
+            widget = new NotificationWidget(this);
```
ckamm: When are these widgets deleted?

src/gui/activitywidget.cpp
dragotin: The idea here is that after one hour, the tray notification pops up again for those notifications that have not been resolved. Maybe one hour is too smal, maybe one day rather?

src/gui/servernotificationhandler.cpp
dragotin: This one is difficult. The time-string from server is for example `2016-03-04T16:20:10+00:00`. How is that parsed into a DateTime, given that the default timeSpec of QDateTime is Qt::LocalTime?

That was why I decided to set the timespec explicitely. However, there is another bug report about the same topic that questions it. I added this setTimeSpec to fix #4439 with the activity view. Honestly I am a bit clueless on this.

src/gui/activitywidget.cpp
dragotin: @ckamm there is indeed a guard for that: Check `ActivityWidget::_notificationRequestsRunning` variable. Only if that one is 0 a check for new notifications is started in ActivityWidget::slotRefresh(AccountState *ptr)

src/gui/activitywidget.cpp
dragotin: I forgot to add that to the TODO list in the first comment.

src/gui/activitywidget.cpp
ckamm: @dragotin Yes, that protects against *starting* to fetch new activities while a confirm job is running. It doesn't protect against *finishing* fetching of new activities while a confirm job is running. And I wonder whether that's a problem.

dragotin: @ckamm good point. I do not think it is a problem, because: The notification widget is reused if it already existed. The buttons are re-created, if new notifications appear, but the notification job end request does not touch the button pointers, but gets the pointers from the button box.

Theoretically the buttons could change for the same notification. But even than, if the notification was answered, that would not be a problem. 

I conclude that I think it is fine :-)

ckamm: Okay! :)

dragotin:
@ckamm can you have a look again on the new commits? @danimo ?

src/gui/servernotificationhandler.cpp
ogoffart: I hope that Qt parses the "+00:00" properly and properly sets the timeSpec.

But setTimeSpec is not what you want to do here. I think the line can simply be removed.

dragotin:
@ckamm @ogoffart please. I'd like to get this to master if possible.

ckamm:
@dragotin Ah, I didn't see this early enough and now it's easter. Will review on Tuesday morning!

src/libsync/utility.cpp
```diff
@@ -434,15 +434,17 @@ QByteArray Utility::versionOfInstalledBinary( const QString& command )
...
 QString Utility::timeAgoInWords(const QDateTime& dt, const QDateTime& from)
 {
-    QDateTime now = QDateTime::currentDateTime();
+    QDateTime now = QDateTime::currentDateTimeUtc();
 
     if( from.isValid() ) {
         now = from;
     }
 
+    Q_ASSERT(dt.timeSpec() == Qt::UTC);
```
ckamm: Remove this

src/gui/activitydata.h
```diff
@@ -0,0 +1,88 @@
...
+    QDateTime _dateTime;
+    QString   _accName;
+
+    QVector <ActivityLink> _links;
+    /**
+     * @brief Sort operator to sort the list youngest first.
+     * @param val
+     * @return
+     */
+    bool operator<( const Activity& val ) const;
```
ogoffart: nitpick:  in general it is better to have operator< and operator==  as non member function (for symetry reason)

src/gui/activitylistmodel.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+// current strategy: Fetch 100 items per Account
+// ATTENTION: This method is const and thus it is not possible to modify
+// the _activityLists hash or so. Doesn't make it easier...
+bool ActivityListModel::canFetchMore(const QModelIndex& ) const
+{
+    if( _activityLists.count() == 0 ) return true;
+
+    QMap<AccountState*, ActivityList>::const_iterator i = _activityLists.begin();
+    while (i != _activityLists.end()) {
```
ogoffart: for() loops are better since they are more readable and make it obvious it's an iteration.

 for(auto i = _activityLists.begin() ; i != _activityLists.end(); ++i) {

src/gui/activitylistmodel.cpp
```diff
@@ -0,0 +1,231 @@
...
+    beginInsertRows(QModelIndex(), 0, resultList.count());
+    _finalList = resultList;
+    endInsertRows();
+}
+
+void ActivityListModel::fetchMore(const QModelIndex &)
+{
+    QList<AccountStatePtr> accounts = AccountManager::instance()->accounts();
+
+    foreach (AccountStatePtr asp, accounts) {
```
ogoffart: const AccountStatePtr &

src/gui/activitylistmodel.cpp
```diff
@@ -0,0 +1,231 @@
...
+    beginRemoveRows(QModelIndex(), 0, _finalList.count() );
+    _finalList.clear();
+    endRemoveRows();
+
+    beginInsertRows(QModelIndex(), 0, resultList.count());
+    _finalList = resultList;
+    endInsertRows();
+}
+
+void ActivityListModel::fetchMore(const QModelIndex &)
```
ogoffart: Normaly, fetchMore is too be used when there are more items to fetch that are hidden.
So that can be used in order to fetch older items form example. 
But here it's used to fetch other accounts despite it's not the next item.
fetchMore will be called when the scrollbar reach the end. 
I don't think it's a good use of the fetchMore capabilities

src/gui/activitylistmodel.cpp
```diff
@@ -0,0 +1,231 @@
...
+{
+    ActivityList resultList;
+
+    foreach( ActivityList list, _activityLists.values() ) {
+        resultList.append(list);
+    }
+
+    std::sort( resultList.begin(), resultList.end() );
+
+    beginRemoveRows(QModelIndex(), 0, _finalList.count() );
```
ogoffart: since you are removing everything and recreating everything you can as well reset the model.
beginResetModel()  / endResetModel()

ogoffart: note that removing everything implies that all state or selection will be list when this happen.

doc/conffile.rst
```diff
@@ -23,3 +23,5 @@ You can change the following configuration settings (must be under the ``[ownClo
 - ``maxLogLines`` (default:  ``20000``) -- Specifies the maximum number of log lines displayed in the log window.
 
 - ``chunkSize`` (default:  ``5242880``) -- Specifies the chunk size of uploaded files in bytes.
+
+- ``notificationRefreshInterval`` (default``300,000``)  -- Specifies the default interval of checking for new server notifications.
```
ckamm: The unit is seconds, as usual?

src/gui/activitywidget.cpp
```diff
@@ -360,8 +415,73 @@ void ActivityWidget::slotNotifyServerFinished( const QString& reply, int replyCo
...
+    NotificationWidget *widget = _widgetForNotifId[ blacklistActivity.ident() ];
+    scheduleWidgetToRemove(widget);
+}
+
+void ActivityWidget::scheduleWidgetToRemove(NotificationWidget *widget, int milliseconds)
+{
+    if( !widget ) {
+        return;
+    }
+    // in fife seconds from now, remove the widget.
```
ckamm: five

src/gui/activitywidget.h
```diff
@@ -160,8 +91,21 @@ public slots:
     QPushButton *_copyBtn;
 
     QSet<QString> _accountsWithoutActivities;
+    QMap<Activity::Identifier, NotificationWidget*> _widgetForNotifId;
+    QElapsedTimer _guiLogTimer;
+    QSet<int> _guiLoggedNotifications;
+    ActivityList _blacklistedActivities;
```
ogoffart: Maybe this needs to be a QSet otherwise your code might get very slow if there are many blacklisted activities (O(n²))

src/gui/activitywidget.cpp
```diff
@@ -400,6 +224,264 @@ void ActivityWidget::slotOpenFile(QModelIndex indx)
...
+        // Assemble a tray notification
+        QString msg = tr("You received %n new notification(s) from %2.", "", accNotified[accNotified.keys().at(0)]).
+                arg(accNotified.keys().at(0));
+
+        if( newGuiLogCount >= 2 ) {
+            QString acc1 = accNotified.keys().at(0);
+            QString acc2 = accNotified.keys().at(1);
+            if( newGuiLogCount == 2 ) {
+                int notiCount = accNotified[ acc1 ] + accNotified[ acc2 ];
+                msg = tr("You received %1 new notifications from %2 and %3.").arg(notiCount).arg(acc1).arg(acc2);
```
ogoffart: msg = tr("You received %n new notifications from %1 and %2.", notiCount).arg(acc1, acc2);

 - use the plural form even if it's always more than 1. There are languages in which the name are different for 2 or more than 10, or so.
 - use the arg with multiple argument as it otherwise might cause problem id there is a %1 in the name of one account.

src/gui/activitywidget.cpp
```diff
@@ -400,6 +224,264 @@ void ActivityWidget::slotOpenFile(QModelIndex indx)
...
+            } else {
+                msg = tr("You received new notifications from %1, %2 and other accounts.").arg(acc1).arg(acc2);
```
ogoffart: arg(acc1, acc2)

src/gui/activitywidget.cpp
```diff
@@ -400,6 +224,264 @@ void ActivityWidget::slotOpenFile(QModelIndex indx)
...
+            }
+        }
+
+        emit guiLog(Theme::instance()->appNameGUI() + QLatin1String(" ") + tr("Notifications - Action Required"),
```
ogoffart: tr("%1 Notifications - Action Required").arg(Theme::instance()->appNameGUI())

ogoffart: and a translation comment explaining what %1 is

src/gui/activitywidget.cpp
```diff
@@ -400,6 +224,264 @@ void ActivityWidget::slotOpenFile(QModelIndex indx)
...
+                    msg);
+    }
+}
+
+void ActivityWidget::slotSendNotificationRequest(const QString& accountName, const QString& link, const QString& verb)
```
ogoffart: why use QString for the verb.  It should be QByteArray

src/gui/notificationconfirmjob.cpp
```diff
@@ -0,0 +1,81 @@
...
+{
+    if( !_link.isValid() ) {
+        qDebug() << "Attempt to trigger invalid URL: " << _link.toString();
+        return;
+    }
+    QNetworkRequest req;
+    req.setRawHeader("Content-Type", "application/x-www-form-urlencoded");
+
+    QIODevice *iodevice = 0;
+    setReply(davRequest(_verb.toAscii(), _link, req, iodevice));
```
ogoffart: toAscii is deprecated, don't use it.
Ac tually, the _verb  should be a QByteArray already.

src/gui/activitywidget.h
```diff
@@ -87,10 +91,17 @@ private slots:
     QPushButton *_copyBtn;
 
     QSet<QString> _accountsWithoutActivities;
-    QMap<int, NotificationWidget*> _widgetForNotifId;
+    QMap<Activity::Identifier, NotificationWidget*> _widgetForNotifId;
     QElapsedTimer _guiLogTimer;
     QSet<int> _guiLoggedNotifications;
-    int _notificationRequests;
+    ActivityList _blacklistedActivities;
```
ckamm: We don't persist this information - does it mean users have to acknowledge the server notification every time the client is restarted?

I assume that for notifications with actions, the server will keep track of them and that's how we avoid presenting them to the user multiple times...

src/gui/notificationwidget.cpp
```diff
@@ -0,0 +1,147 @@
...
+    QString timeStr = locale.toString(QTime::currentTime());
+
+    // the ocs API returns stat code 100 if it succeeded.
+    if( statusCode != OCS_SUCCESS_STATUS_CODE  ) {
+        qDebug() << Q_FUNC_INFO << "Notification Request to Server failed, leave button visible.";
+        for( i = 0; i < _buttons.count(); i++ ) {
+            _buttons.at(i)->setEnabled(true);
+        }
+        //* The second parameter is a time, such as 'failed at 09:58pm'
+        doneText = tr("%1 request failed at %2").arg(_actionLabel).arg(timeStr);
```
ogoffart: use the two arguments   arg()

src/gui/notificationwidget.cpp
```diff
@@ -0,0 +1,147 @@
...
+    } else {
+        // the call to the ocs API succeeded.
+        _ui._buttonBox->hide();
+
+        //* The second parameter is a time, such as 'selected at 09:58pm'
+        doneText = tr("'%1' selected at %2").arg(_actionLabel).arg(timeStr);
```
ogoffart: here too

src/gui/notificationwidget.cpp
```diff
@@ -0,0 +1,147 @@
...
+        // of buttons that were added to the button box before.
+        for( int i = 0; i < _buttons.count(); i++ ) {
+            if( _buttons.at(i) == buttonWidget ) {
+                index = i;
+            }
+            _buttons.at(i)->setEnabled(false);
+        }
+
+        // if the button was found, the link must be called
+        if( index > -1 && _myActivity._links.count() == 0 ) {
```
ogoffart: nitpick: You could have a different slot for the close button.

src/gui/notificationwidget.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+    QString timeStr = locale.toString(QTime::currentTime());
+
+    // the ocs API returns stat code 100 if it succeeded.
+    if( statusCode != OCS_SUCCESS_STATUS_CODE  ) {
+        qDebug() << Q_FUNC_INFO << "Notification Request to Server failed, leave button visible.";
+        for( i = 0; i < _buttons.count(); i++ ) {
+            _buttons.at(i)->setEnabled(true);
+        }
+        //* The second parameter is a time, such as 'failed at 09:58pm'
```
ogoffart: it's  //:    for comment that should be shown to the translators

src/gui/notificationwidget.cpp
```diff
@@ -0,0 +1,147 @@
...
+        doneText = tr("%1 request failed at %2").arg(_actionLabel).arg(timeStr);
+    } else {
+        // the call to the ocs API succeeded.
+        _ui._buttonBox->hide();
+
+        //* The second parameter is a time, such as 'selected at 09:58pm'
```
ogoffart: //:

src/gui/servernotificationhandler.cpp
ogoffart: remove the setTimeSpec(Qt::UTC)

src/gui/activitylistmodel.cpp
dragotin: Ok. let me handle that in a new PR together with a proxy model maybe.

src/gui/activitywidget.h
dragotin: Well, the "blacklisted activities" are actually these notifications that come without any action, which means that they do not display a button. These are only a few.

src/gui/activitywidget.h
dragotin: Yes, we present the notifications from the server again, IF the notification is of the special form that has no action, ie. no button defined. I think that is the desired behaviour, as the admin will disable the notifification if it is not longer important.

src/gui/notificationwidget.cpp
dragotin: It's not that easy because the available buttons per notification is not fixed, it is part of the notification description. It has to be dynamic.

src/gui/notificationwidget.cpp
dragotin: Thanks, good catch!

dragotin:
Thanks @ogoffart and @ckamm I think I reworked all according to your feedback.

ogoffart:
:+1:

PR #4580, opened 2016-03-17
Implement EP 1136

danimo:


guruz:
:+1: from me too.
Shall I merge and cherry-pick into 2.1?

dragotin:
Please put the implementations of the Theme methods into the cpp file.

danimo:
Done. Do still not merge, as the current commit only addresses one out of two issues for owncloud/enterprise/issues/1136. I'll add the second one in a second commit.

dragotin:
With the postfix entry field, what happens if the user enters the domain, as she is used to do it?  Does the postfixlineedit add the postfix automagically when I call text() on it?

Also, please add at least a bit of documentation to the postfixlineedit class. I know its third party code, but still a bit of usage information and what it does in general would be good.

danimo:
With the postfix entry field, what happens if the user enters the domain, as she is used to do it? Does the postfixlineedit add the postfix automagically when I call text() on it?

Please give it a try. It always has the .static.domain in the text field, and will not allow a situation where it is not there. There is no special handing for typing away a full url, but pasting is possible. Question is how to prevent this? disallow typing '.'? then again, this might be allowed. I think it's fine the way it is.

>Also, please add at least a bit of documentation to the postfixlineedit class. I know its third party code, but still a bit of usage information and what it does in general would be good.

It's not third party code, will add some documentation.

dragotin:
ok, thanks @danimo , awesome patch. :+1: please merge.

PR #4584, opened 2016-03-18
Fix crash due to destruction order

lultimouomo:
`ExcludedFiles::~ExcludedFiles` calls `c_strlist_destroy()`, so it must be run before `csync_destroy()`; field destructors are run after the destructor body, so we cannot use a smart pointer for `SyncEngine::_excludedFiles`.

### DIFF
--- a/src/libsync/syncengine.cpp
+++ b/src/libsync/syncengine.cpp
@@ -99,6 +99,7 @@ SyncEngine::SyncEngine(AccountPtr account, const QString& localPath,
 
 SyncEngine::~SyncEngine()
 {
+    _excludedFiles.reset();
     csync_destroy(_csync_ctx);
     _thread.quit();
     _thread.wait();


guruz:
Are you sure this is how it crashes? ```c_strlist_destroy``` looks quite safe..

Shouldn't the c_ functions be independent from the context no matter what? 

e.g. look at ```csync_excluded_no_ctx``` it does not have a context but still uses ```c_strlist_t*```

src/libsync/syncengine.cpp
```diff
@@ -99,6 +99,7 @@ SyncEngine::SyncEngine(AccountPtr account, const QString& localPath,
 
 SyncEngine::~SyncEngine()
 {
+    delete _excludedFiles;
```
jturcotte: Thanks for the fix!

I'd prefer to keep the smart pointer, you can just call `_excludedFiles.reset();` to achieve the same.

lultimouomo:
@guruz `csync_destroy` frees `SyncEngine::_csync_ctx`, therefore `_csync_ctx->excludes` that was passed to `SyncEngine::_excludedFiles` in the constructor is not valid anymore. MinGW in debug mode helpfully sets it to an invalid value during `free()` so that you get a reliable segfault instead of random undefined behaviour.

lultimouomo:
This contribution is MIT licensed

PR #4591, opened 2016-03-21
AccountManager: take theURL from the Theme rather than from the confi…

ogoffart:
…g if the theme specify it

That way an upgrade of the client can actually change the URL
Issue https://github.com/owncloud/enterprise/issues/1113
https://github.com/owncloud/enterprise/issues/1126

In addition to restoring commit 7e5d89293d7bd52ac42d026d34a8154fbbdf33bb, this
add a way to override the auth type


guruz:
:+1:

ogoffart:
I did not try that with a real setup.  This might cause problem because of the username.

PR #4605, opened 2016-03-23
Change note to warning in documentation

RealRancor:
Follow-up of: https://github.com/owncloud/documentation/issues/1981

Warnings currently looks like seen here: https://doc.owncloud.org/server/9.0/admin_manual/issues/code_signing.html#errors

### DIFF
--- a/doc/troubleshooting.rst
+++ b/doc/troubleshooting.rst
@@ -78,7 +78,7 @@ Other issues can affect synchronization of your ownCloud files:
 - If you are operating your own server, and use the local storage backend (the
   default), make sure that ownCloud has exclusive access to the directory.
 
-  .. note:: The data directory on the server is exclusive to ownCloud and must not be modified manually.
+  .. warning:: The data directory on the server is exclusive to ownCloud and must not be modified manually.
 
 - If you are using a different file backend on the server, you can try to exclude a bug in the
   backend by reverting to the built-in backend.


dragotin:
@carlaschroder your call, please.

guruz:
@icewind1991 Can you help us with this one?

I wonder if the (original part) is actually true... or shall we mention that the original part is especially valid if ```filesystem_check_changes``` is 0?

RealRancor:
@guruz The original part is there since ages and long before filesystem_check_changes was set to 0 as the default. So a clarification definitely makes sense.

guruz:
Would you like to add the occ scan command or a link to it? Like "if you modified, you need to run..." or so?                                                                                                                                                                                                                                                                                                                                        Sent from my BlackBerry 10 smartphone.                                                                                                                                                                                                                From: RealRancorSent: Mittwoch, 30. März 2016 11:42To: owncloud/clientReply To: owncloud/clientCc: Markus GoetzSubject: Re: [owncloud/client] Change note to warning in documentation (#4605)@guruz The original part is there since ages and long before filesystem_check_changes was set to 0 as the default. So a clarification definitely makes sense.

—You are receiving this because you were mentioned.Reply to this email directly or view it on GitHub

RealRancor:
Thanks for merging. I think this is getting even more important as discussed here: https://github.com/owncloud/core/issues/23783

guruz:
oh well.. :-|

Issue #3122, opened 2015-04-16
Investigate possible hang during discovery


guruz:
```
04-15 16:46:16:473 0x8bb290 OCC::SocketApi::sendMessage: SocketApi:  Sending message:  "STATUS:NEW:C:\Users\xxx\OwnCloud2\Music\LISTEN RETAG AND SORT & CHECK LEVELS\Mp3's\Music\E\Eric Johnson\Ah via musicom\zzz.mp3"
04-15 16:46:16:473 0x8bb290 OCC::SocketApi::command_RETRIEVE_FILE_STATUS: void OCC::SocketApi::command_RETRIEVE_FILE_STATUS(const QString&, QLocalSocket*) "C:\Users\xxx\OwnCloud2\Music\LISTEN RETAG AND SORT & CHECK LEVELS\Mp3's\Music\Incoming\Mary Mary\Incredible\yyy.mp3"
04-15 16:46:16:474 0x9bbcdb0 _csync_detect_update: Database entry found, compare: 1008103388 <-> 1008103388, etag: 8f91279ed7c5ba7213794d966585605e <-> 8f91279ed7c5ba7213794d966585605e, inode: 445127 <-> 445127, size: 4192634 <-> 4192634, perms:  <-> RDNVW
04-15 16:46:16:474 0x9bbcdb0 _csync_detect_update: file: Music/LISTEN RETAG AND SORT & CHECK LEVELS/Mp3's/Music/C/Collective Soul/xxx.mp3, instruction: INSTRUCTION_NONE <<=
... some more lines of csync_update in thread 0x9bbcdb0
```
Note how the thread 0x8bb290 does not continue with sending the result for RETRIEVE_FILE_STATUS
See also discussion on IRC.

The user reports that if he deinstalls the Explorer extension it works fine.
(Note that this not necessarily means this is OS-specific)

eagles051387:
Suggest leaving the explorer extension disabled by default.

ogoffart:
`SyncEngine::estimateState`  is O(n) in the number of files.  I believe it could be made O(log n) since the vector is sorted.  I don't know if this is the bottleneck.

eagles051387:
From what I have seen it seems to be an issue with windows integration the
hang and crash. I have downoaded the 1.8.1 beta of the client and installed
it without explorer integration that eliminates the issue all together.
could there be a race condition maybe between discovery and windows
explorer integration?

[...]

PR #4615, opened 2016-03-28
Socket API refactor

jturcotte:
- Make the SyncEngine persistent in the Folder, in order to be able to move backed logic into libsync
- Move any backend logic from SocketApi into a new SyncFileStatusTracker in libsync
- Refactor SyncFileStatusTracker to simplify the logic, make it match 100% the status of files in the UI and support bubbling up error statuses to parent folders as a warning
- Blacklisted files are now kept marked as error in the overlay icons even though they are a simple information in the activity log

Next step is to try adding auto tests for this, but this branch was tested manually during development and can land without.

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -34,6 +34,7 @@
 
 #include <QDebug>
 #include <QDesktopServices>
+#include <QDir>
 #include <QListWidgetItem>
 #include <QMessageBox>
 #include <QAction>
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -114,7 +114,6 @@ Folder::Folder(const FolderDefinition& definition,
     connect(_engine.data(), SIGNAL(transmissionProgress(ProgressInfo)), this, SLOT(slotTransmissionProgress(ProgressInfo)));
     connect(_engine.data(), SIGNAL(itemCompleted(const SyncFileItem &, const PropagatorJob &)),
             this, SLOT(slotItemCompleted(const SyncFileItem &, const PropagatorJob &)));
-    connect(_engine.data(), SIGNAL(syncItemDiscovered(const SyncFileItem &)), this, SLOT(slotSyncItemDiscovered(const SyncFileItem &)));
     connect(_engine.data(), SIGNAL(newBigFolder(QString)), this, SLOT(slotNewBigFolderDiscovered(QString)));
 }
 
@@ -911,11 +910,6 @@ void Folder::slotItemCompleted(const SyncFileItem &item, const PropagatorJob& jo
         _syncResult.setWarnCount(_syncResult.warnCount()+1);
     }
     emit ProgressDispatcher::instance()->itemCompleted(alias(), item, job);
-}
-
-void Folder::slotSyncItemDiscovered(const SyncFileItem & item)
-{
-    emit ProgressDispatcher::instance()->syncItemDiscovered(alias(), item);
 }
 
 void Folder::slotNewBigFolderDiscovered(const QString &newF)
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -56,9 +56,10 @@ static void csyncLogCatcher(int /*verbosity*/,
 
 
 Folder::Folder(const FolderDefinition& definition,
+               AccountState* accountState,
                QObject* parent)
     : QObject(parent)
-      , _accountState(0)
+      , _accountState(accountState)
       , _definition(definition)
       , _csyncError(false)
       , _csyncUnavail(false)
@@ -92,16 +93,6 @@ Folder::~Folder()
         _engine->abort();
         _engine.reset(0);
     }
-}
-
-void Folder::setAccountState( AccountState *account )
-{
-    _accountState = account;
-}
-
-AccountState* Folder::accountState() const
-{
-    return _accountState;
 }
 
 void Folder::checkLocalPath()
@@ -202,9 +193,6 @@ QString Folder::remotePath() const
 
 QUrl Folder::remoteUrl() const
 {
-    if (!_accountState) {
-        return QUrl("http://deleted-account");
-    }
     return Account::concatUrlPath(_accountState->account()->davUrl(), remotePath());
 }
 
@@ -255,11 +243,6 @@ void Folder::prepareToSync()
 void Folder::slotRunEtagJob()
 {
     qDebug() << "* Trying to check" << remoteUrl().toString() << "for changes via ETag check. (time since last sync:" << (_timeSinceLastSyncDone.elapsed() / 1000) << "s)";
-
-    if (!_accountState) {
-        qDebug() << "Can't run EtagJob, account is deleted";
-        return;
-    }
 
     AccountPtr account = _accountState->account();
 
@@ -328,9 +311,7 @@ void Folder::etagRetreived(const QString& etag)
         emit scheduleToSync(this);
     }
 
-    if( _accountState ) {
-        _accountState->tagLastSuccessfullETagRequest();
-    }
+    _accountState->tagLastSuccessfullETagRequest();
 }
 
 void Folder::etagRetreivedFromSyncEngine(const QString& etag)
@@ -665,11 +646,6 @@ bool Folder::estimateState(QString fn, csync_ftw_type_e t, SyncFileStatus* s)
 
 void Folder::saveToSettings() const
 {
-    if (!_accountState) {
-        qDebug() << "Can't save folder to settings, account is deleted";
-        return;
-    }
-
     auto settings = _accountState->settings();
     settings->beginGroup(QLatin1String("Folders"));
     FolderDefinition::save(*settings, _definition);
@@ -680,11 +656,6 @@ void Folder::saveToSettings() const
 
 void Folder::removeFromSettings() const
 {
-    if (!_accountState) {
-        qDebug() << "Can't remove folder from settings, account is deleted";
-        return;
-    }
-
     auto  settings = _accountState->settings();
     settings->beginGroup(QLatin1String("Folders"));
     settings->remove(_definition.alias);
@@ -692,24 +663,12 @@ void Folder::removeFromSettings() const
 
 bool Folder::isFileExcludedAbsolute(const QString& fullPath) const
 {
-    if (!fullPath.startsWith(path())) {
-        // Mark paths we're not responsible for as excluded...
-        return true;
-    }
-
-    QString myFullPath = fullPath;
-    if (myFullPath.endsWith(QLatin1Char('/'))) {
-        myFullPath.chop(1);
-    }
-
-    QString relativePath = myFullPath.mid(path().size());
-    auto excl = ExcludedFiles::instance().isExcluded(myFullPath, relativePath, _definition.ignoreHiddenFiles);
-    return excl != CSYNC_NOT_EXCLUDED;
+    return ExcludedFiles::instance().isExcluded(fullPath, path(), _definition.ignoreHiddenFiles);
 }
 
 bool Folder::isFileExcludedRelative(const QString& relativePath) const
 {
-    return isFileExcludedAbsolute(path() + relativePath);
+    return ExcludedFiles::instance().isExcluded(path() + relativePath, path(), _definition.ignoreHiddenFiles);
 }
 
 void Folder::watcherSlot(QString fn)
@@ -814,11 +773,6 @@ bool Folder::proxyDirty()
 
 void Folder::startSync(const QStringList &pathList)
 {
-    if (!_accountState) {
-        qDebug() << "Can't startSync, account is deleted";
-        return;
-    }
-
     Q_UNUSED(pathList)
     if (proxyDirty()) {
         setProxyDirty(false);
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -71,6 +71,9 @@ Folder::Folder(const FolderDefinition& definition,
       , _consecutiveFollowUpSyncs(0)
       , _journal(definition.localPath)
 {
+    qRegisterMetaType<SyncFileItemVector>("SyncFileItemVector");
+    qRegisterMetaType<SyncFileItem::Direction>("SyncFileItem::Direction");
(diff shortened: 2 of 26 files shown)


jturcotte:
cc: @dragotin @ogoffart @guruz @ckamm

ckamm:
:+1: on df386b64ba88c70f13cae2514e3dacda8540fb6a

src/libsync/syncfilestatustracker.h
```diff
@@ -0,0 +1,51 @@
...
+#include <csync.h>
+
+namespace OCC {
+
+class SyncEngine;
+
+class OWNCLOUDSYNC_EXPORT SyncFileStatusTracker : public QObject
+{
+public:
+    SyncFileStatusTracker(SyncEngine *syncEngine);
```
ckamm: explicit?

src/libsync/syncfilestatustracker.h
```diff
@@ -0,0 +1,51 @@
...
+#include "syncfileitem.h"
+#include "syncfilestatus.h"
+#include <QSet>
+#include <csync.h>
+
+namespace OCC {
+
+class SyncEngine;
+
+class OWNCLOUDSYNC_EXPORT SyncFileStatusTracker : public QObject
```
ckamm: doc?

ckamm:
On  	6d3fe9d and tainted folders: What if the sync will take a while because a different folder is being synced currently? Sure, the existing code wasn't fully correct as it didn't push the notification, but don't we want something like this?

src/gui/socketapi.h
```diff
@@ -44,7 +44,7 @@ class SocketApi : public QObject
 Q_OBJECT
 
 public:
-    SocketApi(QObject* parent);
+    SocketApi(QObject* parent = 0);
```
ckamm: explicit?

jturcotte:
@ckamm Yes I've been wondering if that would be necessary, I didn't think about the multiple folder scenario. At the same time, this hasn't been working and nobody complained (and it's pretty rare that you'll have a sync taking more than 5 seconds on two folders at the same time), so I would argue that the complexity and the reduced code reliability wouldn't be worth the benefit. What do you think?

src/libsync/syncfilestatustracker.cpp
```diff
@@ -225,6 +227,32 @@ void SyncFileStatusTracker::slotItemCompleted(const SyncFileItem &item)
...
+void SyncFileStatusTracker::slotItemDiscovered(const SyncFileItem &item)
+{
+    QString systemFileName = _syncEngine->localPath() + item.destination();
+
+    // the trailing slash for directories must be appended as the filenames coming in
+    // from the plugins have that too. Otherwise the matching entry item is not found
+    // in the plugin.
+    if( item._type == SyncFileItem::Type::Directory ) {
+        systemFileName += QLatin1Char('/');
+    }
```
ckamm: should systemFileName() be a function somewhere?

jturcotte: Yes you're probably right, some kind of `toShellPath` and `toSyncEnginePath` functions that would add/remove the slash from directories.

src/gui/accountsettings.cpp
```diff
@@ -34,6 +34,7 @@
 
 #include <QDebug>
 #include <QDesktopServices>
+#include <QDir>
```
ogoffart: needed?

jturcotte: Yes, since I removed it from `src/gui/folder.h`.

src/gui/folder.cpp
```diff
@@ -84,24 +88,37 @@ Folder::Folder(const FolderDefinition& definition,
...
+            SLOT(slotAboutToRestoreBackup(bool*)));
+    connect(_engine.data(), SIGNAL(folderDiscovered(bool,QString)), this, SLOT(slotFolderDiscovered(bool,QString)));
+    connect(_engine.data(), SIGNAL(transmissionProgress(ProgressInfo)), this, SLOT(slotTransmissionProgress(ProgressInfo)));
+    connect(_engine.data(), SIGNAL(itemCompleted(const SyncFileItem &, const PropagatorJob &)),
+            this, SLOT(slotItemCompleted(const SyncFileItem &, const PropagatorJob &)));
+    connect(_engine.data(), SIGNAL(newBigFolder(QString)), this, SLOT(slotNewBigFolderDiscovered(QString)));
 }
 
-AccountState* Folder::accountState() const
+Folder::~Folder()
```
ogoffart: Is it pos

src/libsync/syncfilestatustracker.h
```diff
@@ -0,0 +1,62 @@
...
+
+struct Problem {
+    QString path;
+    SyncFileStatus::SyncFileStatusTag severity = SyncFileStatus::StatusError;
+
+    Problem(const QString &path) : path(path) { }
+    Problem(const QString &path, SyncFileStatus::SyncFileStatusTag severity) : path(path), severity(severity) { }
+
+    // Assume that each path will only have one severity, don't care about it in sorting
+    bool operator<(const Problem &other) const { return path < other.path; }
```
ogoffart: it is better if the  operator< and operator== are non-member  (friends).  For simetry reason.

(eg,  you can type    "theProblem < myString"  but not  "myString < theProblem")

src/libsync/syncfilestatustracker.h
```diff
@@ -17,38 +18,43 @@
...
-
-    SyncEngine *_syncEngine;
-    // SocketAPI: Cache files and folders that had errors so that they can
-    // get a red ERROR icon.
-    QSet<QString>   _stateLastSyncItemsWithErrorNew; // gets moved to _stateLastSyncItemsWithError at end of sync
-    QSet<QString>   _stateLastSyncItemsWithError;
+    SyncFileStatus fileStatus(const SyncFileItem& item);
+    void invalidateParentPaths(const QString& path);
+    SyncEngine* _syncEngine;
+    std::set<Problem> _syncProblems;
```
ckamm: Wouldn't a `map<path, severity>` be more convenient here? Then `lookupProblem` could use `lower_bound` too.

src/libsync/syncfilestatustracker.h
```diff
@@ -17,38 +18,43 @@
...
 
 namespace OCC {
 
 class SyncEngine;
 
+struct Problem {
+    QString path;
+    SyncFileStatus::SyncFileStatusTag severity = SyncFileStatus::StatusError;
+
+    Problem(const QString &path) : path(path) { }
```
ckamm: explicit

src/gui/folder.cpp
jturcotte: I don't think that it's possible (it should stop the sync as part of the destruction), but in any case, I moved the abort() logic to the SyncEngine's destructor (owned by the Folder through a QScopedPointer) and unless I missed anything, the previous behavior should be preserved.

src/libsync/syncfilestatustracker.cpp
```diff
@@ -31,229 +40,143 @@ namespace OCC {
...
+    for (auto it = oldProblems.begin(); it != oldProblems.end(); ++it) {
+        if (it->severity == SyncFileStatus::StatusError)
+            invalidateParentPaths(it->path);
+        emit fileStatusChanged(_syncEngine->localPath() + it->path, fileStatus(it->path));
+    }
 }
 
 void SyncFileStatusTracker::slotItemCompleted(const SyncFileItem &item)
 {
+    qDebug() << Q_FUNC_INFO << item.destination() << item._status;
```
ckamm: There are this, and a bunch of other, qDebug()s left in this patch - intentional?

src/libsync/syncfilestatustracker.cpp
```diff
@@ -0,0 +1,182 @@
...
+#include "syncfilestatustracker.h"
+#include "syncengine.h"
+#include "syncjournaldb.h"
+#include "syncjournalfilerecord.h"
+
+namespace OCC {
+
+static SyncFileStatus::SyncFileStatusTag lookupProblem(const QString &pathToMatch, const std::set<Problem> &set)
+{
+    for (auto it = set.cbegin(); it != set.cend(); ++it) {
```
ogoffart: I suppose you can use std::lower_bound

src/libsync/syncfilestatustracker.cpp
```diff
@@ -0,0 +1,182 @@
...
+    SyncFileItem* item = _syncEngine->findSyncItem(fileName);
+    if (item)
+        return fileStatus(*item);
+
+    // If we're not currently syncing that file, look it up in the database to know if it's shared
+    SyncJournalFileRecord rec = _syncEngine->journal()->getFileRecord(fileName);
+    if (rec.isValid())
+        return fileStatus(rec.toSyncFileItem());
+
+    // Must be a new file, wait for the filesystem watcher to trigger a sync
```
ogoffart: shouldn't new file automatically show a blue icon?

src/libsync/syncfilestatustracker.cpp
```diff
@@ -0,0 +1,182 @@
...
+    return SyncFileStatus();
+}
+
+void SyncFileStatusTracker::slotAboutToPropagate(SyncFileItemVector& items)
+{
+    std::set<Problem> oldProblems;
+    std::swap(_syncProblems, oldProblems);
+
+    foreach (const SyncFileItemPtr &item, items) {
+        qDebug() << Q_FUNC_INFO << "Investigating" << item->destination() << item->_status;
```
ogoffart: Is that not a bit of flood?

src/libsync/syncfilestatustracker.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+        if (showErrorInSocketApi(*item))
+            _syncProblems.insert({item->_file, SyncFileStatus::StatusError});
```
ogoffart: I don't think this will compile in CentOS 6

ckamm:
@jturcotte About tainted folders: I think we should defer this issue. I think we will want to maintain a list of 'dirty' files and folders anyway to allow us to sync files without a full discovery. Once we have that, it would have to hook into the SyncFileStatusTracker.

src/libsync/syncengine.cpp
```diff
@@ -1301,33 +1306,14 @@ void SyncEngine::restoreOldFiles()
...
-bool SyncEngine::estimateState(QString fn, csync_ftw_type_e t, SyncFileStatus* s)
+SyncFileItem* SyncEngine::findSyncItem(const QString &fileName) const
 {
-    Q_UNUSED(t);
-    QString pat(fn);
-    if( t == CSYNC_FTW_TYPE_DIR && ! fn.endsWith(QLatin1Char('/'))) {
-        pat.append(QLatin1Char('/'));
-    }
-
     Q_FOREACH(const SyncFileItemPtr &item, _syncedItems) {
```
ogoffart: i fear this is slow..  linear search  on this function that might be called often.
Is the _syncedItems vector not sorted by filename so a binary search is in order?

src/libsync/excludedfiles.h
```diff
@@ -70,7 +67,6 @@ public slots:
     // but the pointer can be in a csync_context so that it can itself also query the list.
     c_strlist_t** _excludesPtr;
     QStringList _excludeFiles;
-    mutable QReadWriteLock _mutex;
```
ogoffart: Why do you remove the mutex?

src/libsync/syncfilestatustracker.cpp
jturcotte: I tried it and it felt right that no icon is shown until the sync kicks in (which took 1-2 seconds). It make the behavior consistent with the rest that if the icon is blue, it's because the sync is happening, and this would always be the case.

Otherwise, we would show a blue icon for new files, but the tray icon would stay green, and we will keep showing the OK status for modified files unless we implement the tainted folder logic.

Personally I feel that if the tray icon shown as syncing, the files should also not be shown as "syncing".

src/libsync/syncengine.cpp
jturcotte: Yes this is slow, this is the main reason that caused #3122, which we worked around by reducing the calls.
It's sorted by destination, so binary search will work most of the time, but not for a rename made by the remote, since we will need to match SyncFileItem::_file in this case instead of SyncFileItem::_renameTarget (which will be the value returned by destination().

This patch isn't changing this though, it's as slow as before.

src/libsync/excludedfiles.h
jturcotte: See a4e0899af4eec5883bb16011c5760ca0c3707f76, it's always called from the GUI thread.

jturcotte:
New commits should address all of the comments that I didn't contest.

ckamm:
:+1:

jturcotte:
The 3 extra commits fix issues I've seen on Windows, if anybody can have a quick look, I think that the branch is ready to merge after this.

If somebody has time to quickly try it on Linux just to make sure that it's not totally broken it would be nice. I guess that it's on Linux that the matching suffix slash consistency for directories is important?

guruz:
> Remove the tainted folder logic

"a sync will be triggered within the next 5 seconds and the
modified folder will be shown as SYNC at that point anyway."

Uff... you know that our discovery jobs are known for taking arbitarily long and the state would only then be updated?

guruz:
(...although if @jturcotte says the tainted folder logic was broken anyway?)

guruz:
> SyncEngine::aboutToPropagate is used in favor of SyncEngine::syncItemDiscovered
since the latter is emitted before file permission warnings are set on the
SyncFileItem.

Same problem: Discovery might take a long time..

guruz:
ok discussed with @ogoffart .. it's not so bad for now, we keep the PR in and fix it when @jturcotte comes back from holiday...

jturcotte:
Yes it wasn't adding much in that the new status wasn't pushed to the socketapi client, so removing this logic isn't really regressing anything. I agree that the functionality makes sense though, and we would have to reimplemented it slightly differently with the new class setup by pushing the tainted items on a file level and store it in a map like the sync problems, making sure that it affects parents like errors do, and emit the file status changed signal accordingly.

PR #4645, opened 2016-04-06
SelectiveSync: Verify if the list could be read from journal.

### LINKED ISSUE #4633: Selective Sync: Handle DB Read Error
### Expected behaviour

If database access fails for whatever reason while reading the selective sync information and reading of the info fails, we should bail out with an critical error.

### Current behaviour

If the select fails, it looks to the user as if the selective sync information is just lost.

dragotin:
If there is a read error from the database while trying to get
the list from database, make sure to not behave badly because
the list is empty.

Supposed to fix #4633

### DIFF
--- a/src/gui/selectivesyncdialog.cpp
+++ b/src/gui/selectivesyncdialog.cpp
@@ -391,12 +391,17 @@ qint64 SelectiveSyncTreeView::estimatedSize(QTreeWidgetItem* root)
 
 
 SelectiveSyncDialog::SelectiveSyncDialog(AccountPtr account, Folder* folder, QWidget* parent, Qt::WindowFlags f)
-    :   QDialog(parent, f), _folder(folder)
+    :   QDialog(parent, f), _folder(folder),
+      _okButton(0) // defined in init()
 {
+    bool ok;
     init(account, tr("Unchecked folders will be <b>removed</b> from your local file system and will not be synchronized to this computer anymore"));
-    _treeView->setFolderInfo(_folder->remotePath(), _folder->alias(),
-                             _folder->journalDb()->getSelectiveSyncList(SyncJournalDb::SelectiveSyncBlackList));
-
+    QStringList selectiveSyncList = _folder->journalDb()->getSelectiveSyncList(SyncJournalDb::SelectiveSyncBlackList, &ok);
+    if( ok ) {
+        _treeView->setFolderInfo(_folder->remotePath(), _folder->alias(),selectiveSyncList);
+    } else {
+        _okButton->setEnabled(false);
+    }
     // Make sure we don't get crashes if the folder is destroyed while we are still open
     connect(_folder, SIGNAL(destroyed(QObject*)), this, SLOT(deleteLater()));
 }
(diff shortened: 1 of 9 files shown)


src/gui/selectivesyncdialog.cpp
```diff
@@ -391,12 +391,17 @@ qint64 SelectiveSyncTreeView::estimatedSize(QTreeWidgetItem* root)
...
+    bool ok;
     init(account, tr("Unchecked folders will be <b>removed</b> from your local file system and will not be synchronized to this computer anymore"));
-    _treeView->setFolderInfo(_folder->remotePath(), _folder->alias(),
-                             _folder->journalDb()->getSelectiveSyncList(SyncJournalDb::SelectiveSyncBlackList));
-
+    QStringList selectiveSyncList = _folder->journalDb()->getSelectiveSyncList(SyncJournalDb::SelectiveSyncBlackList, &ok);
+    if( ok ) {
+        _treeView->setFolderInfo(_folder->remotePath(), _folder->alias(),selectiveSyncList);
+    } else {
+        _okButton->setEnabled(false);
```
ogoffart: i must be missing something but i think _okButton is allaws null here.

ogoffart: ah yes, i missed the call to init()

src/libsync/syncengine.cpp
```diff
@@ -762,7 +766,13 @@ void SyncEngine::startSync()
     DiscoveryJob *discoveryJob = new DiscoveryJob(_csync_ctx);
     discoveryJob->_selectiveSyncBlackList = selectiveSyncBlackList;
     discoveryJob->_selectiveSyncWhiteList =
-        _journal->getSelectiveSyncList(SyncJournalDb::SelectiveSyncWhiteList);
+        _journal->getSelectiveSyncList(SyncJournalDb::SelectiveSyncWhiteList, &ok);
+    if (!ok) {
+        delete discoveryJob;
+        qDebug() << Q_FUNC_INFO << "Unable to read selective sync list, aborting.";
+        return;
```
ogoffart:
```
        emit csyncError(tr("Unable to read from the sync journal."));
        finalize(false);
        return
```

src/libsync/syncengine.cpp
```diff
@@ -733,10 +733,14 @@ void SyncEngine::startSync()
...
-    bool usingSelectiveSync = (!selectiveSyncBlackList.isEmpty());
-    qDebug() << (usingSelectiveSync ? "====Using Selective Sync" : "====NOT Using Selective Sync");
-
+    bool ok;
+    auto selectiveSyncBlackList = _journal->getSelectiveSyncList(SyncJournalDb::SelectiveSyncBlackList, &ok);
+    if (ok) {
+        bool usingSelectiveSync = (!selectiveSyncBlackList.isEmpty());
+        qDebug() << Q_FUNC_INFO << (usingSelectiveSync ? "====Using Selective Sync" : "====NOT Using Selective Sync");
+    } else {
+        qDebug() << Q_FUNC_INFO << "Could not retrieve selective sync list from DB";
```
ogoffart: Then you don't want to continue the sync or it will sync everything !

```
        emit csyncError(tr("Unable to read the blacklist from the local database"));
        finalize(false);
        return;
```

guruz:
> Supposed to fix #4633 

(Only if it verifies ALL DB reads..)

dragotin:
Thanks @ogoffart, I pushed a fix for the stuff you mentioned.

ogoffart:
:+1:

PR #4658, opened 2016-04-11
Fix qt4 build

ogoffart:


ogoffart:
@danimo  for the CMakeLists change,  @dragotin for the ActivityWidget changes

src/gui/activitywidget.cpp
```diff
@@ -440,32 +442,35 @@ void ActivityWidget::scheduleWidgetToRemove(NotificationWidget *widget, int mill
...
-
-    QPair<QDateTime, NotificationWidget*> removeInfo = qMakePair(removeTime, widget);
-    if( !_widgetsToRemove.contains(removeInfo) ) {
-        _widgetsToRemove.insert( removeInfo );
-        if( !_removeTimer.isActive() ) {
-            _removeTimer.start();
-        }
+    // in five seconds from now, remove the widget.
+    QDateTime removeTime = QDateTime::currentDateTimeUtc().addMSecs(milliseconds);
+    QDateTime &it = _widgetsToRemove[widget];
```
dragotin: Minor: Usually we use the variable name it for iterators, here it's a DateTime

dragotin:
Thanks @ogoffart for cleaning up :+1: 
Why do we do the CMake changes?

ogoffart:
without the cmake changes, the dolphin plugin does not compile because the Qt4 include paths are injected. That's why i moved the Qt detection to src/

Issue #4638, opened 2016-04-05
Checksum Mismatch on Download Must be Handled Gracefully


dragotin:
### Expected behaviour
If the client detects a mismatch of checksum after download of a file, it must not completely drop the file but only display a warning that the checksums did not match and the file might be corrupted.

### Actual behaviour
Because of owncloud/core#23783 we must not handle the checksum mismatch as a hard error. Because the server potentially delivers wrong checksums for files, there will be files where the checksums will not match even for **not** corrupted download. 

### Steps to reproduce
1. Have a wrong checksum stored in servers file cache
2. Have it synced to the client 
3. Client will complain about broken checksum and try again. 

Client version 2.2.0 git.

danimo:
Lets disable checksumming completely then. Displaying warnings will do more harm than good, i.e. lead to more bug reports that will wonder why there are warnings, and if they should worry.

dragotin:
Hm. Yes. Probably true.

guruz:
Linking https://github.com/owncloud/core/issues/23783

AnrDaemon:
What use is for such a warning?
Checksums are expected to be correct all along. This is the very reason they exists.
I suggest closing this report as invalig. There must be no wrong checksum in the server storage in first place. If it happened at the server side, the bug is in the server.

rullzer:
As discussed today once again. We just pass around checksums and do no actual checksum verification on the server. Which makes them not very useful. As such it is also somewhat dangerous for the client to rely on them.

AnrDaemon:
That's.… a new low…

AnrDaemon:
If you don't do checksum verification, then don't pass them around. Period.
As stated, the problem is in the server, not client.

moscicki:
Hello,

I would second the above comment: if this checksumming feature is enabled on the client then you should assume that checksums are reliable and re-download automatically. If the checksum problem persists for several retries then this is another problem and should be reported to a user (most likely a bug in the stack somewhere). This would be useful to report to the user.

ckamm:
It sounds like we need to do something like this:
* Disable hard checksum validation for servers that don't have a fix for owncloud/core#23783. It looks like there might be a patch in 9.0.2, but it's not certain yet. This is easy because we already have `OWNCLOUD_DISABLE_CHECKSUM_DOWNLOAD` and `downloadChecksumEnabled()`.
[...]

PR #4663, opened 2016-04-12
Disable validation of transmission checksums by default #4638

ckamm:
The server may send invalid checksums when files change on disk,
see owncloud/core#23783.

Introduces a new validateChecksumOnDownload configuration and
branding option.

### DIFF
--- a/src/libsync/checksums.cpp
+++ b/src/libsync/checksums.cpp
@@ -17,6 +17,7 @@
 #include "syncfileitem.h"
 #include "propagatorjobs.h"
 #include "account.h"
+#include "configfile.h"
 
 #include <qtconcurrentrun.h>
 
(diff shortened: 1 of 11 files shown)


src/libsync/checksums.cpp
```diff
@@ -17,6 +17,7 @@
 #include "syncfileitem.h"
 #include "propagatorjobs.h"
 #include "account.h"
+#include "configfile.h"
```
ogoffart: I wish we don't make dependency between the sync engine and the config file.

ogoffart:
I did not see the validateChecksumOnDownload

Should we really get rid of the branding option? branding who enables this option do it because their server is capable? Or is it assumed that the branding would also have their own server?

src/libsync/checksums.cpp
ogoffart: Since if it's for a debugging purpose, maybe an environment variable would work.  But I think we could also leave it out. Who is this debug option for?  Developers can as well comment the code that does the checking.

dragotin:
@ogoffart yes, there is only one branding using that. And they agreed on setting the server capability.

ckamm:
@dragotin @ogoffart Could I ask you to potentially fix and merge this? I won't get around to it before Monday, unfortunately!

moscicki:
Didn't we agree that this option is not needed? And to use capabilities and not the configuration file? 

https://github.com/owncloud/client/issues/4638#issuecomment-210369951

dragotin:
Yes, we did, I guess that is exactly what Christian means.

ckamm:
@dragotin @moscicki I didn't update this pull request description as the discussion in the ticket unfolded. The patch ea40e314d404c81878f902a60e1a0688ca82a15e removes the config and branding options, as should be expected.

Issue #4655, opened 2016-04-11
[Linux] ignores changes to some files (.git folder): when the client is restarted it notices the changed files and uploads them OK


joysfera:
### Expected behaviour
Whenever a file is modified locally it is uploaded to server

### Actual behaviour
Changes in files (in a .git folder) are sometimes ignored

### Steps to reproduce
1. commit something to a local git repository
2. notice that files changed in the .git folder are not being uploaded to server, not even after two hours of waiting to see if the owncloud-client catches up
3. happens sometimes only, started like 4 days ago, before it's been working fine for 3 months, AFAIK

### Server configuration
Operating system: Ubuntu 12.04 LTS

Web server: Apache

Database: MySQL

PHP version: 5.4.45

ownCloud version: 8.2.2-1.1

Storage backend: filesystem

### Client configuration
Client version: 2.1.1-2.1

Operating system: Ubuntu 14.04 LTS

OS language: Czech

Installation path of client: from your debian package

### Logs
I don't have logs yet as the client is running without logs now and it seems that for capturing log to a file I'd have to restart it - and **when the client is restarted it notices the changed files and uploads them OK** so such log wouldn't help.

What I can observe is that the last owncloud client upload happened at 8:43:51 even though there have newly modified files since then, see below:
```
$ LC_ALL=C ls -lt .git
total 140
-rw-rw-r--   1 joy joy  4463 Apr 11 09:09 gitk.cache
drwxrwxr-x   8 joy joy  4096 Apr 11 09:09 ./
-rw-rw-r--   1 joy joy   643 Apr 11 09:09 index
drwxrwxr-x 135 joy joy 20480 Apr 11 09:09 objects/
drwxrwxr-x   4 joy joy  4096 Apr 11 08:43 ../
```
Please notice that this happens irregularly and I don't know if it is related to the particular .git folder only - I just noticed it here first.

guruz:
@ckamm file watcher issue somehow?

dragotin:
@joysfera could you check the setting of the checkbox "Sync hidden files" in your client? See https://doc.owncloud.org/desktop/2.1/navigating.html#using-the-ignored-files-editor

joysfera:
@dragotin it's of course enabled. It sometimes syncs the .git folder so it's not a settings issue (and it did sync it everytime in the past three months). I hope it answered also eventual question about having .git in the ignore list (no, it's not there).

ckamm:
I'll try to reproduce it later.

[...]

PR #4664, opened 2016-04-12
Fix hidden file handling #4655

ckamm:
There were two issues:

* With the refactoring of how Folder and SyncEngine relate, the
  ignore_hidden_files flag on the CSync context was reset after
  each sync run and not updated from the configuration again.

* The folder watcher failed to enumerate hidden folders and thus
  didn't watch for changes inside them. (linux only)

### DIFF
--- a/src/gui/folderwatcher_linux.cpp
+++ b/src/gui/folderwatcher_linux.cpp
@@ -58,7 +58,7 @@ bool FolderWatcherPrivate::findFoldersBelow( const QDir& dir, QStringList& fullL
     } else {
         QStringList nameFilter;
         nameFilter << QLatin1String("*");
-        QDir::Filters filter = QDir::Dirs | QDir::NoDotAndDotDot|QDir::NoSymLinks;
+        QDir::Filters filter = QDir::Dirs | QDir::NoDotAndDotDot | QDir::NoSymLinks | QDir::Hidden;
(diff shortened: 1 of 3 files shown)


src/gui/folderwatcher_linux.cpp
```diff
@@ -58,7 +58,7 @@ bool FolderWatcherPrivate::findFoldersBelow( const QDir& dir, QStringList& fullL
     } else {
         QStringList nameFilter;
         nameFilter << QLatin1String("*");
-        QDir::Filters filter = QDir::Dirs | QDir::NoDotAndDotDot|QDir::NoSymLinks;
+        QDir::Filters filter = QDir::Dirs | QDir::NoDotAndDotDot | QDir::NoSymLinks | QDir::Hidden;
         const QStringList pathes = dir.entryList(nameFilter, filter);
```
ogoffart: can we add the Hidden flag only if we are syncing them?

ckamm: Yes, but the current pattern here is to generate a full list and then call pathIsIgnored() on each entry afterwards. If we wanted to apply the "is hidden" part of the filter early, we'd need to make the "ignoreHidden" flag available which is currently an implementation detail of pathIsIgnored().

Instead I'd prefer to skip building that list altogether. Agreed?

ogoffart:
ok :+1:

Issue #4383, opened 2016-01-21
Disable share permission checkboxes that are unavailable when resharing


ckamm:
In #4357 #4358 we fixed the problem of resharing being impossible when some permissions had been revoked. Now it works, but users can still try to enable the unavailable permissions manually. Since the client knows which permissions are available, we should disable the permission checkboxes that are forbidden.

To reproduce:
* Share user1's folder/ with user2, give only 'reshare' permission
* As user2, try to share folder/ with user3 using the desktop client
* Notice the 'edit'/'change'/'delete' etc permission checkboxes can be toggled, but won't work

rullzer:
I see no clean way to be able to do this properly for the current API. But if we get something like https://github.com/owncloud/core/pull/22834 in at the server side then we could easily make this happen.

rullzer:
https://github.com/owncloud/core/pull/22834 will only be in 9.1. However we can already start with implementing this in the client. So that once 9.1 is out this will work directly.

I'll prepare a patch that makes this happen soonish.

rullzer:
https://github.com/owncloud/core/pull/22834 is in. So this will be available in 9.1.

This introduces a new webdav property. `share-permissions`. All clients could already query this property now but it would just return a 404 (and then we fall back to the current crappy behaviour). But if the property is there we can do much better.

ckamm:
Pull request: #4673

mcastroSG:
Hi hi:

**Steps executed**

1. user1 shares a folder with user2 giving just reshare privilege.
2. user2 shares the same folder to user3 with "Share with ownCloud" option and choosing user3

**Result**
 
![screen shot 2016-04-19 at 18 03 24](https://cloud.githubusercontent.com/assets/18436398/14646258/18b27366-0659-11e6-8cbb-2cd6906ffee4.png)

The checkbox are shown. In case we check any of them we obtain:

![screen shot 2016-04-19 at 18 13 14](https://cloud.githubusercontent.com/assets/18436398/14646627/76d8e816-065a-11e6-9a67-14c196edd7fe.png)

So after that we could just close the window.

**MacOS**
Client Version: Version 2.2.0-nightly20160419 (build 3282)
Client OS: MacOS X El Capitan 10.11.4
Server Version: 9.0.1
Server OS: Ubuntu

**Windows**
Unable to test due to [https://github.com/owncloud/client/issues/4701](url)

**LOGS**

[...]

PR #4673, opened 2016-04-13
Disable unavailable sharing permissions #4383

ckamm:
Users can't reshare with more permissions than they have themselves.

### DIFF
--- a/src/gui/owncloudgui.cpp
+++ b/src/gui/owncloudgui.cpp
@@ -870,14 +870,27 @@ void ownCloudGui::slotShowShareDialog(const QString &sharePath, const QString &l
 
     const auto accountState = folder->accountState();
 
+    // As a first approximation, set the set of permissions that can be granted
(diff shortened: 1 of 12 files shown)


src/gui/owncloudgui.cpp
```diff
@@ -870,14 +870,26 @@ void ownCloudGui::slotShowShareDialog(const QString &sharePath, const QString &l
 
     const auto accountState = folder->accountState();
 
+    // As a first approximation, set the set of permissions that can be granted
+    // either to everything (resharing allowed) or nothing (no resharing).
+    //
+    // The correct value will be found with a propfind from ShareDialog.
+    SharePermissions maxSharingPermissions =
+            SharePermissions(SharePermissionRead)
```
ogoffart: i believe you can use Q_DECLARE_OPERATORS_FOR_FLAGS  and then you don't need the cast

ogoffart:
:+1:

dragotin:
:+1:

Issue #4657, opened 2016-04-11
Allow to use the character & on the Application Name


rperezb:
### Expected behaviour
I'd like to use the character "&" on the APPLICATION_NAME  variable so that i can call my app "Raquel & cia" :wink: https://github.com/owncloud/client/blob/d7bd1300a8397c2782e8d75cf7c595b1ada70d88/OWNCLOUD.cmake#L1

### Actual behaviour
There is a compilation error

````
** BUILD SUCCEEDED **

[ 99%] Built target legacy_mac_overlayplugin
[100%] icons
/bin/sh: V Cloud: command not found
make[2]: *** [shell_integration/MacOSX/CMakeFiles/mac_overlayplugin] Error 127
make[1]: *** [shell_integration/MacOSX/CMakeFiles/mac_overlayplugin.dir/all] Error 2
make: *** [all] Error 2
````

cc @jturcotte 

related to https://github.com/owncloud/ownbrander/issues/581

guruz:
The problem in the shell is easy to fix by adding " but then then ```Info.plist``` is broken because ```&``` is not valid in XML..
checking how to escpae this..

guruz:
PR: https://github.com/owncloud/client/pull/4683

rperezb:
works! thx
![screen shot 2016-05-10 at 12 52 22](https://cloud.githubusercontent.com/assets/1846927/15144186/1b43a2b6-16ae-11e6-806f-a62e51f59c0c.png)

PR #4683, opened 2016-04-14
OS X: Allow ampersand in APPLICATION_NAME #4657

guruz:


ogoffart:
While you are at it, i think you should also replace '<' and '>'
Is there other fields that would need to be escaped?

guruz:
@ogoffart I grepped for ```APPLICATION_NAME``` but did not find anything more..
I also tested via the build server and the OS X builds worked..

I will add the gt/lt

guruz:
Updated commit

dragotin:
For XML you need to escape 
```
"   &quot;
'   &apos;
<   &lt;
>   &gt;
&   &amp;
```
And once we build a client for *"Claude's Client Cloud"* we will need the quotes excaped.

ogoffart:
:+1:   

Quotes need not to be escaped in this case since APPLICATION_NAME is inside an element and not in an atribute

dragotin:
but would it hurt?

guruz:
@dragotin I'll merge it for now then, we indeed don't use it inside attributes.
If I add the support for quotes, I'll need to escape the quotes _inside cmake_ :-) I'm worried to fall into an inception trap.

dragotin:
ok, I see. Thanks.

Issue #1603, opened 2014-03-22
ownCloud Client ignores files with mtime = 0


Aentfs:
### Expected behaviour
ownCloud Client should should sync files with mtime = 0 as it syncs files with mtime = 1.

### Actual behaviour
ownCloud Client ignores files, with mtime = 0, but synces files with mtime = 1. 
This happend, as i wanted to sync a file, and for some reason the mtime was not copyed over with the file.

### Steps to reproduce
1. Create a file outside of owncloud
2. Use ```touch -t 19700101010000.00 file```
3. Copy or move file to the sync dir
4. ownCloud Client ignores file
5. Now use ```touch -t 19700101010000.01 file``` on file in Sync dir
6. ownCloud Client syncs file

### Server configuration
Operating system: [will be provided if necessary]

Web server: [will be provided if necessary]

Database: [will be provided if necessary]

PHP version: [will be provided if necessary]

ownCloud version: ```ownCloud 6.0.2 (stable)``` 

Storage backend: [will be provided if necessary]

### Client configuration
Client version: 
* ```Version 1.5.3 (Git revision c8a8eb; OCsync 0.19.5; Qt 4.8.5)```
* ```qtkeychain: 0.3-1```
* ```qtwebkit: 2.3.3-1```
* ```neon: 0.30.0-1```

Operating system: ```Archlinux x86_64 with KDE 4.12.3``` 
```Linux 3.13.6-1-ARCH #1 SMP PREEMPT Fri Mar 7 22:47:48 CET 2014 x86_64 GNU/Linux```

OS language: ```de_DE.UTF-8``` (generated locals: ```de_DE.UTF-8``` ```en_US.UTF-8```)

Installation path of client: ```$ whereis owncloud
owncloud: /usr/bin/owncloud [binary] /usr/share/owncloud [folder with l10n inside]```

### Logs

Please use Gist (https://gist.github.com/) or a similar code paster for longer
logs.

```Template for output < 10 lines```

1. Output of `owncloud --logwindow` or `owncloud --logfile log.txt`

https://gist.github.com/Aentfs/b9fd9a07c0d75ea5b181

2. Web server error log: [will be provided if necessary]

3. ownCloud log (data/owncloud.log): [will be provided if necessary]

stefaanderoeck:
I have the same issue. The zero mtime files (they're pictures) are also on an ext4 filesystem, and originate from my android phone, probably transferred through mtp to Windows and then through either smb or scp to this remote Linux box. 

[...]

PR #4686, opened 2016-04-14
CSync: Allow files with 0 mtime #1603

guruz:
I don't see a reason we should not allow this. We detect
a "failed stat reply" in other ways. Maybe this was a legacy
csync decision somewhow that is not valid anymore?


ogoffart:
I believe when the stat files, type == CSYNC_FTW_TYPE_UNKOWN and this code is not reach.
So yes, we can just remove that code.
If it was me i'd also remove the log entry.

anyway, :+1:

PR #4687, opened 2016-04-15
The share-permissions namespace was moved

rullzer:
See https://github.com/owncloud/core/pull/24017

This was done since other services we want to intergrate with don't use
the ownCloud namespace.

Once https://github.com/owncloud/core/pull/24017 is in. Pleae give it a quick test.


guruz:
That's unreleased code right? So we don't need any special stuff for old/new namespace?

guruz:
If yes then: :+1:

rullzer:
yes unreleased :)

guruz:
:+1: 
(didnt test it..)

rullzer:
Merged in master.
Quick test worked for me so merging.

Issue #3634, opened 2015-08-17
[Overlays] Client folder error status not shown in higher-level icon status


phil-davis:
### Expected behaviour
If a folder has a problem syncing from client to server then the problem status will be shown in the icons of parent folders and in the overall sync status icon in the icon box.

### Actual behaviour
If a folder has a problem syncing from client to server it (correctly) shows a little red "x" in the folder icon.
But any folders above it show green,
The overall client status icon shows green.

If the problem folder is several folders down in the folder tree then the user is unlikely to notice it and thus will not know to investigate.

It would be nice to show some non-green status for the folders above and the overall status icon.

### Steps to reproduce
1. On the client create FolderA.
2. Let it sync up to the server.
3. From the folders-to-sync dropdown, unselect FolderA and apply.
4. A sync starts, FolderA is deleted on the client - good - and remains on the server - good.
5. Create another FolderA in the same place.
FolderA is shown with red "x" icon - good.
Activity log says: Ignored because of the "choose what to sync" blacklist
That is also reasonable, because there is already FolderA on the server with some (presumably different) content, and it is not being synced to this client.

But the overall sync status is green.

There is the same sort of behaviour when FolderA is inside some other folder/s. The higher level folders all show green status.

### Server configuration
ownCloud version: 8.0

### Client configuration
Client version: Version 2.0.0-nightly20150817 (build 5403)

Operating system: Windows 10

OS language: English (US)

Installation path of client: default

MTRichards:
@guruz are you working on this now for 2.0, or does it slide to 2.0.1?

jturcotte:
I haven't investigated it yet but unless this is a regression from 1.8 (it's not, is it?) then we shouldn't put pressure on 2.0.0. If this requires bigger changes it might have to go in 2.1.

See also https://github.com/owncloud/client/issues/3171 for a glimpse of how error statuses aresn't preserved as long as they should. The moment that a file ends up in the black list, its status becomes invisible to the user. My knowledge of that area isn't quite good though, @ogoffart could prove me wrong if I am.

jturcotte:
Another question is: Is it correct to propagate anything else than the SYNC status to parent folders?

If I have the structure:
A
|- B
 |- C

And that C has an ignored or error status, should A also be marked as ignored/error? A isn't ignored completely, A isn't completely in error either. For SYNC, however we can say that "A isn't up-to-date", implied by C not being up-to-date.

[...]

PR #4689, opened 2016-04-15
SyncFileStatusTracker: Add a method to handle the root item.

dragotin:
This uses some  heuristics to estimate the root status. Not complete as it completely neglects the database items, but maybe a start.

@jturcotte @ckamm please.

### DIFF
--- a/src/gui/socketapi.cpp
+++ b/src/gui/socketapi.cpp
@@ -293,16 +293,8 @@ void SocketApi::command_RETRIEVE_FILE_STATUS(const QString& argument, QIODevice*
         const QString file = QDir::cleanPath(argument).mid(syncFolder->cleanPath().length()+1);
 
         // future: Send more specific states for paused, disconnected etc.
-        if( syncFolder->syncPaused() ){
+        if( syncFolder->syncPaused() || !syncFolder->accountState()->isConnected() ) {
             statusString = nopString;
-        } else if( !syncFolder->accountState()->isConnected() ) {
-            if( file.isEmpty() || file == QLatin1String("/") ) {
-                // only the root folder
-                statusString = nopString;
-            } else {
-                // all other files and dirs in unconnected sync
-                statusString = nopString;
-            }
         } else {
             SyncFileStatus fileStatus = syncFolder->syncEngine().syncFileStatusTracker().fileStatus(file);
 
--- a/src/gui/socketapi.cpp
+++ b/src/gui/socketapi.cpp
@@ -274,6 +274,8 @@ void SocketApi::command_RETRIEVE_FOLDER_STATUS(const QString& argument, QIODevic
 
 void SocketApi::command_RETRIEVE_FILE_STATUS(const QString& argument, QIODevice* socket)
 {
+    const QString nopString("NOP");
+
     if( !socket ) {
         qDebug() << "No valid socket object.";
         return;
@@ -286,12 +288,26 @@ void SocketApi::command_RETRIEVE_FILE_STATUS(const QString& argument, QIODevice*
     Folder* syncFolder = FolderMan::instance()->folderForPath( argument );
     if (!syncFolder) {
         // this can happen in offline mode e.g.: nothing to worry about
-        statusString = QLatin1String("NOP");
+        statusString = nopString;
     } else {
         const QString file = QDir::cleanPath(argument).mid(syncFolder->cleanPath().length()+1);
-        SyncFileStatus fileStatus = syncFolder->syncEngine().syncFileStatusTracker().fileStatus(file);
 
-        statusString = fileStatus.toSocketAPIString();
+        // future: Send more specific states for paused, disconnected etc.
+        if( syncFolder->syncPaused() ){
+            statusString = nopString;
+        } else if( !syncFolder->accountState()->isConnected() ) {
+            if( file.isEmpty() || file == QLatin1String("/") ) {
+                // only the root folder
+                statusString = nopString;
+            } else {
+                // all other files and dirs in unconnected sync
+                statusString = nopString;
+            }
+        } else {
+            SyncFileStatus fileStatus = syncFolder->syncEngine().syncFileStatusTracker().fileStatus(file);
+
+            statusString = fileStatus.toSocketAPIString();
+        }
     }
 
     const QString message = QLatin1String("STATUS:") % statusString % QLatin1Char(':') %  QDir::toNativeSeparators(argument);
--- a/src/libsync/syncfilestatustracker.cpp
+++ b/src/libsync/syncfilestatustracker.cpp
@@ -77,6 +77,43 @@ SyncFileStatusTracker::SyncFileStatusTracker(SyncEngine *syncEngine)
             this, SLOT(slotItemCompleted(const SyncFileItem&)));
 }
 
+SyncFileStatus SyncFileStatusTracker::rootStatus()
+{
+    /* Possible values for the status:
+    enum SyncFileStatusTag {
+        StatusNone,
+        StatusSync,
+        StatusWarning,
+        StatusUpToDate,
+        StatusError,
+    };
+    */
+    SyncFileStatus status =  SyncFileStatus::StatusUpToDate;
+
+    if( !_syncEngine ) return SyncFileStatus::StatusNone;
+
+    if( _syncEngine->isSyncRunning() ) {
+        status = SyncFileStatus::StatusSync;
+    } else {
+        // sync is not running. Check dirty list and _syncProblems
+        int errs = 0, warns = 0;
+        for (auto it = _syncProblems.begin(); it != _syncProblems.end(); ++it) {
+            if( it->second == SyncFileStatus::StatusError ) {
+                errs ++;
+                break; // stop if an error found at all.
+            } else {
+                warns ++;
+            }
+        }
+        if( errs ) {
+            status = SyncFileStatus::StatusError;
+        } else if( warns ) {
+            status = SyncFileStatus::StatusWarning;
+        }
+    }
+    return status;
+
+}
 
 SyncFileStatus SyncFileStatusTracker::fileStatus(const QString& systemFileName)
 {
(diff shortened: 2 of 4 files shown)


jturcotte:
This will break the bubbling up of blacklisted errors into warnings up to the root. Is there a way that we can find the root problem instead?

I think that it would be nice to avoid  the root folder having any special logic, the status of the whole sync should go in the UI or tray icon instead, it is more a "this sync" info than "this file" info IMO.

dragotin:
Yes, I agree having no special logic for root would be better. OTOH this already does it, and does not waste too much performance. 

Well, and maybe the tray icon and the root sync folder should be aligned what they show. That is true.

I think the root cause of our problem here is that the whole stuff is not really specified. We should solve that first, and until that is done stay with what we had maybe? At least work the same way...

jturcotte:
I meant that the root folder and the sync status should not be the same, or do you see a reason why they should be?

The real issue is that something is off between the root item and the database, any idea why this is the case? Fixing this would avoid adding more logic in this class and make everything easier to test (as fetching that info from the folder would break layers and make it impossible to test).

jturcotte:
We just got started with pull requests and I'm not very available right now so maybe it's not the best time to discuss this. If you can't find another way to fix this, need that fix in and that you would allow me to update the solution later, feel free to ignore my concerns.

All of this would be clearer once I managed to cover this with auto tests, that's the next thing on my list. But at least meanwhile you were able to find a few bugs with the new code.

dragotin:
I looked at that again today. I was trying to create a SyncFileItem object for the sync root that comes in and pass it through the normal  implementation that all other SyncFileItems go through to find the status.

That can work, but I think it is not useful, because: The basic logic for the root folder is fairly easy: It  either sync or not (visible by syncEngine->isSyncRunning() ) or has errors or not. Since all sync problems affect the root dir, it is sufficient to analyze the `_syncProblems` data structure, and if that is not empty, the root has to show error or warning.  

Because of that, I think this pr is not that bad.

dragotin:
@ogoffart @ckamm mind to review this one?

jturcotte:
Ah I understand the issue now, the root item is not actually synced, it's only a container. So how about using the forged SyncFileItem approach you have, but not caring about the error or sync status at all, and treat it only so: a container that inherently always has StatusUpToDate.

This would be a few lines fix, and would make the warning bubbling up work as it does for other folders.

dragotin:
@jturcotte I am not sure I understand. Well, it uses the bubbled up status, as all errors or warnings bubble into the data structure `_syncProblems`, so regardless of the hierarchie, it tells me about the error status.

src/libsync/syncfilestatustracker.cpp
```diff
@@ -77,6 +77,43 @@ SyncFileStatusTracker::SyncFileStatusTracker(SyncEngine *syncEngine)
...
+        for (auto it = _syncProblems.begin(); it != _syncProblems.end(); ++it) {
+            if( it->second == SyncFileStatus::StatusError ) {
+                errs ++;
+                break; // stop if an error found at all.
+            } else {
+                warns ++;
+            }
+        }
+        if( errs ) {
+            status = SyncFileStatus::StatusError;
```
jturcotte: We only bubble up errors as a warning for parents folder. Since there can be no error with the root folder itself (it is not synced), it should never show an error. The tray icon error on the other hand applies to all files, not just the root.

src/libsync/syncfilestatustracker.cpp
```diff
@@ -77,6 +77,43 @@ SyncFileStatusTracker::SyncFileStatusTracker(SyncEngine *syncEngine)
...
+        StatusUpToDate,
+        StatusError,
+    };
+    */
+    SyncFileStatus status =  SyncFileStatus::StatusUpToDate;
+
+    if( !_syncEngine ) return SyncFileStatus::StatusNone;
+
+    if( _syncEngine->isSyncRunning() ) {
+        status = SyncFileStatus::StatusSync;
```
jturcotte: On OS X we need to push every new status to invalidate the FinderSync cache. This might cause the root folder status to stagnate in some case (like in your multiple root folders case?).

src/libsync/syncfilestatustracker.cpp
```diff
@@ -77,6 +77,43 @@ SyncFileStatusTracker::SyncFileStatusTracker(SyncEngine *syncEngine)
...
+        StatusNone,
+        StatusSync,
+        StatusWarning,
+        StatusUpToDate,
+        StatusError,
+    };
+    */
+    SyncFileStatus status =  SyncFileStatus::StatusUpToDate;
+
+    if( !_syncEngine ) return SyncFileStatus::StatusNone;
```
jturcotte: _syncEngine will never be null.

src/libsync/syncfilestatustracker.cpp
dragotin: I have a completely different understanding. That is: If there is an error somewhere in the sync directory, the root folder should show an error. From the users perspective: If my root sync folder is red, I know that something is wrong and I can click through the dirs with red-X to find the issue file.

dragotin: Ah, the difference to the tray icon status is of course that the tray needs to consider all sync connections, not only one as the root in this discussion.

dragotin:
Ok, this statement by @MTRichards clearifies how we want to propagate the status:
https://github.com/owncloud/client/issues/3634#issuecomment-132355093

Bottomline: Errors of individual files cause parent folders to be yellow, general sync errors with the account cause it to be red.

It does not clearify however what to do if there are only warnings (symlinks or ignored chars) in the sync, are those propagated up?

ogoffart:
Warnings don't propagate up: Ignored files are supposed to be intended. Ignore chars do no logner exists. Symlinks maybe should be an error then?

dragotin:
Ok, I pushed code that if the account is not connected the icons have no icon any more. As discussed.

src/gui/socketapi.cpp
```diff
@@ -286,12 +288,26 @@ void SocketApi::command_RETRIEVE_FILE_STATUS(const QString& argument, QIODevice*
...
     } else {
         const QString file = QDir::cleanPath(argument).mid(syncFolder->cleanPath().length()+1);
-        SyncFileStatus fileStatus = syncFolder->syncEngine().syncFileStatusTracker().fileStatus(file);
 
-        statusString = fileStatus.toSocketAPIString();
+        // future: Send more specific states for paused, disconnected etc.
+        if( syncFolder->syncPaused() ){
+            statusString = nopString;
+        } else if( !syncFolder->accountState()->isConnected() ) {
+            if( file.isEmpty() || file == QLatin1String("/") ) {
```
ogoffart: Why this condition if it's the same anyway in both branch. 

The more condition, the harder it is to read and maintain the code.

src/gui/socketapi.cpp
```diff
@@ -286,12 +288,26 @@ void SocketApi::command_RETRIEVE_FILE_STATUS(const QString& argument, QIODevice*
...
         // this can happen in offline mode e.g.: nothing to worry about
-        statusString = QLatin1String("NOP");
+        statusString = nopString;
     } else {
         const QString file = QDir::cleanPath(argument).mid(syncFolder->cleanPath().length()+1);
-        SyncFileStatus fileStatus = syncFolder->syncEngine().syncFileStatusTracker().fileStatus(file);
 
-        statusString = fileStatus.toSocketAPIString();
+        // future: Send more specific states for paused, disconnected etc.
+        if( syncFolder->syncPaused() ){
```
ogoffart: So i'd write this condition:
 
  if(syncFolder->syncPaused()  || !syncFolder->accountState()->isConnected())

So you simplify a lot this code.

src/gui/socketapi.cpp
dragotin: because after 2.2.0 we want to return different strings like "PAUSED" and "DISCONNECTED"

src/libsync/syncfilestatustracker.cpp
```diff
@@ -68,6 +68,12 @@ static inline bool showWarningInSocketApi(const SyncFileItem& item)
         || status == SyncFileItem::Restoration;
 }
 
+static inline bool showSyncInSocketApi( const SyncFileItem& item)
+{
+    const auto inst = item._instruction;
```
ogoffart: Is it really worth making a function for a one liner?  And why introduce this variable inst?

And anyway, what's so special about CSYNC_INSTRUCTION_NEW ?  Then it should also be the case for CSYNC_INSTRUCTION_SYNC should it not? 

Unless I misunderstand this.

src/libsync/syncfilestatustracker.cpp
```diff
@@ -77,6 +83,42 @@ SyncFileStatusTracker::SyncFileStatusTracker(SyncEngine *syncEngine)
...
+        status = SyncFileStatus::StatusSync;
+    } else {
+        // sync is not running. Check dirty list and _syncProblems
+        int errs = 0, warns = 0;
+        for (auto it = _syncProblems.begin(); it != _syncProblems.end(); ++it) {
+            if( it->second == SyncFileStatus::StatusError ) {
+                errs ++;
+                break; // stop if an error found at all.
+            } if( it->second == SyncFileStatus::StatusWarning ) {
+                warns ++;
```
ogoffart: why do you even count the warnings if you don't use them?  dead code?

src/libsync/syncfilestatustracker.cpp
```diff
@@ -116,11 +167,13 @@ void SyncFileStatusTracker::slotAboutToPropagate(SyncFileItemVector& items)
...
 
-        if (showErrorInSocketApi(*item))
+        if (showErrorInSocketApi(*item)) {
             _syncProblems[item->_file] = SyncFileStatus::StatusError;
-        else if (showWarningInSocketApi(*item))
+        } else if (showWarningInSocketApi(*item)) {
             _syncProblems[item->_file] = SyncFileStatus::StatusWarning;
-
+        } else if( showSyncInSocketApi(*item)) {
+            _syncProblems[item->_file] = SyncFileStatus::StatusSync;
```
ogoffart: i think _syncProblems should be only for errors and warning.  Any reason to put this in?

src/libsync/syncfilestatustracker.cpp
dragotin: You do :-)

it is explained in the commit message https://github.com/owncloud/client/commit/361ebf5464077ce08049ea9e1f834e6fe56cade9

ogoffart:
:+1:  for what i can see. But i cannot say i fully understand the whole logic as of now.

dragotin:
lets keep this branch alive until jturcotte is back

guruz:
> lets keep this branch alive until @jturcotte is back

OS X: For me the sync folder itself does not change its icon. 
mirall always sends ```STATUS:OK:/Users/guruz/ownCloud/```, somethines ```UPDATE_VIEW:/Users/guruz/ownCloud/``` but then afterwards the plugin does not re-request the status.

I think ```_syncEngine->isSyncRunning()``` becomes true too late can this be?

Issue #4573, opened 2016-03-16
GoogleDrive sync fails due to unsyncable "My Saved Places." file with files returning -2 as getcontentlength from ownCloud


atroxix:
### Expected behaviour
When some objects from GoogleDrive cannot be synced then the sync process should just ignore them, not abort.

### Actual behaviour
Only some files/folders are synced. The failure happens with files located at the root folder of GoogleDrive, so I cannot even sync one certain folder because the error occurs before sync gets there.

Interesting is that when client gives error about unsyncable things (Google Maps files, and conflicting names of folders and files) then it says it downloaded that one certain folder which I'm interested in, but I don't see it in my local disk.

### Steps to reproduce
1. Add GoogleDrive external storage.
2. Configure GoogleDrive folder in client. You can also specify certain sub-folder only, it doesn't actually matter.

### Server configuration
Operating system: FreeBSD

Web server: Apache 2.4

Database: MariaDB 10.0

PHP version: 5.6

ownCloud version: 9.0

Storage backend: normal directory

### Client configuration
Client version: 2.1.1

Operating system: OpenSUSE 42.1

OS language: English

Installation path of client: /usr/bin/owncloud

### Logs

Please use Gist (https://gist.github.com/) or a similar code paster for longer
logs.

**Output of `owncloud --logfile -`**

About a failing file in root:

```
3-16 09:28:47:694 9561 OCC::Folder::slotSyncStarted: #### Propagation start #################################################### >>
03-16 09:28:47:694 9561 OCC::ownCloudGui::slotSyncStateChange: Sync state changed for folder  "http://oc.kodu/remote.php/webdav/GoogleDrive" :  "Sync Running"
03-16 09:28:47:694 9561 OCC::PropagateDownloadFileQNAM::start: virtual void OCC::PropagateDownloadFileQNAM::start() "My Saved Places." 0
03-16 09:28:47:696 9561 OCC::SyncJournalDb::setDownloadInfo: "INSERT OR REPLACE INTO downloadinfo (path, tmpfile, etag, errorcount) VALUES ( ?1 , ?2, ?3, ?4 )" "My Saved Places." ".My Saved Places..~1d5541ff" "56e8fb1dac33f" 0
03-16 09:28:47:697 9561 OCC::SyncJournalDb::commitInternal: void OCC::SyncJournalDb::commitInternal(const QString&, bool) Transaction commit  "download file start" and starting new transaction
[...]

PR #4696, opened 2016-04-18
Discovery: Be more explicit about files with unknown size #4573

guruz:
Also related https://github.com/owncloud/core/issues/23468

### DIFF
--- a/src/libsync/discoveryphase.cpp
+++ b/src/libsync/discoveryphase.cpp
@@ -258,8 +258,12 @@ static csync_vio_file_stat_t* propertyMapToFileStat(const QMap<QString,QString> 
             file_stat->mtime = oc_httpdate_parse(value.toUtf8());
             file_stat->fields |= CSYNC_VIO_FILE_STAT_FIELDS_MTIME;
         } else if (property == "getcontentlength") {
-            file_stat->size = value.toLongLong();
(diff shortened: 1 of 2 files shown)


guruz:
I tested this with .odt (=Google Documents) files return -2 as content length..

src/libsync/discoveryphase.cpp
```diff
@@ -258,8 +258,12 @@ static csync_vio_file_stat_t* propertyMapToFileStat(const QMap<QString,QString>
             file_stat->mtime = oc_httpdate_parse(value.toUtf8());
             file_stat->fields |= CSYNC_VIO_FILE_STAT_FIELDS_MTIME;
         } else if (property == "getcontentlength") {
-            file_stat->size = value.toLongLong();
-            file_stat->fields |= CSYNC_VIO_FILE_STAT_FIELDS_SIZE;
+            bool ok = false;
+            qlonglong ll = value.toLongLong(&ok);
+            if (ok && ll >= 0) {
+                file_stat->size = value.toLongLong();
```
dragotin: nitpicking: I think you should use the already computed value `ll` rather than computing it again. 

In discovery we need to be picky about ever CPU cycle.

guruz: good catch. commit updated..

dragotin:
:+1:

PR #4700, opened 2016-04-19
Use expireDate if returned by the OCS Share API

### LINKED ISSUE #4409: [Password Policy] Windows Desktop Client does not adhere to the expiration date rules set in the share link password policy
### Steps to reproduce
1. Set an expiration date in the share link password policy
2. From the Windows Desktop client generate a public shared link with an expiration date
3. The expiration date for the Desktop generated public link can be set beyond the one set in the share link password policy (need however to verify if the public link generated from Desktop is still accessible after the expiration date set in the share link password policy)

Customer communication as side note (see screenshot below):

> the Windows Desktop Client does not adhere to the rules that are set in the "share link password policy" app. We've set the link to expire after 3 days without secure password and to 14 days with a secure password.
But in the Desktop Client, you can also create a link without password expiring after 14 days.

### Expected behaviour

Windows Desktop Client should adhere to the rules set in the share link password policy

### Actual behaviour

See above

### Server configuration
**Operating system**:

SLES 12

**Web server:**

Apache 2.4

**ownCloud version:** (see ownCloud admin page)

8.2.2

**Updated from an older ownCloud or fresh install:**

Updated from 8.2.1

### Client configuration

**Operating system:**

Windows

**Sync client version:**

branded version 2.1.0 build 452

unbenannt2

rullzer:
If an app modifies the expiration date (for example the password policy
app) then on more recent versions of the server we will get the share
object back REST style. We should use that info!

Fixes #4409

### DIFF
--- a/src/gui/share.cpp
+++ b/src/gui/share.cpp
@@ -180,9 +180,19 @@ void LinkShare::setExpireDate(const QDate &date)
     job->setExpireDate(getId(), date);
 }
 
-void LinkShare::slotExpireDateSet(const QVariantMap&, const QVariant &value)
+void LinkShare::slotExpireDateSet(const QVariantMap& reply, const QVariant &value)
 {
-    _expireDate = value.toDate();
+    auto data = reply.value("ocs").toMap().value("data").toMap();
+
+    /*
+     * If the reply provides a data back (more REST style)
+     * they use this date.
+     */
+    if (data.value("expiration").isValid()) {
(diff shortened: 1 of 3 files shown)


rullzer:
@dragotin @guruz @ckamm pleae have a look

ogoffart:
:+1:

dragotin:
:+1:

rullzer:
Thanks!

PR #4704, opened 2016-04-19
Doc: Draft overlay icons

dragotin:
@owncloud/desktop-developers please check if that is common understanding.

@MTRichards does that reflect your understanding of the overlay icons 

@carlaschroder draft input for the docs, please do whatever you want with it ;-)

### DIFF
--- a/doc/navigating.rst
+++ b/doc/navigating.rst
@@ -11,7 +11,7 @@ in the system tray (Windows, KDE), status bar (Mac OS X), or notification area
 .. figure:: images/icon.png
    :alt: Status icon, little cloud with green circle and white checkmark 
 
-The status indicator uses overlay icons to indicate the current status of your 
+The status indicator uses icons to indicate the current status of your 
 synchronization. The green circle with the white checkmark tells you that your 
 synchronization is current and you are connected to your ownCloud server.
 
@@ -24,7 +24,7 @@ The blue icon with the white semi-circles means synchronization is in progress.
    :alt: Status icon, little cloud with yellow circle and vertical parallel 
     lines
 
-The yellow overlay icon with the parallel lines tells you your synchronization 
+The yellow icon with the parallel lines tells you your synchronization 
 has been paused. (Most likely by you.)
 
 .. figure:: images/icon-offline.png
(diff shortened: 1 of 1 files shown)


carlaschroder:
Hurrah, I love when someone else writes stuff! Editing is so much easier. Thanks @dragotin!

carlaschroder:
Hmm, Github doesn't show a diff of my changes. I hope that doesn't cause any confusion.

doc/navigating.rst
```diff
@@ -133,6 +133,44 @@ appear as a new tab in the settings dialog, where you can adjust its settings
...
+
+If there is a sync error, or the file is blacklisted, it displays an 
+eye-catching red X.
+
+If the file is waiting to be synced, or is currently syncing, the overlay 
+icon displays a blue cycling icon.
+
+When the client is offline, the icons remain in the state of the 
+last successful sync, as long as they are not touched locally. Once a file
+is edited locally, the icon will change to the 'waiting to sync' icon.
```
jturcotte: The idea of showing a waiting to sync icon while offline is incompatible with the waiting to sync status bubbling up to parent directories between syncs, unless you'd like to do a local discovery on startup only to show what is waiting for sync vs the database state (since we can't afford a recursive scan of the whole tree).

Just hiding all overlay icons while offline/paused would allow things to be simpler and bug-free. The whole point of refactoring this code was to improve the reliability on OS X since the file status was often getting stuck in a sync or OK state.

dragotin: @jturcotte understood.  Lets do what you suggest and display no icon when being disconnected.

For the future, I think we should display the disconnected icon at least on the root sync folder. I will create a separate issue.

MTRichards: @dragotin read through it and it makes sense to me.

doc/navigating.rst
```diff
@@ -133,6 +133,44 @@ appear as a new tab in the settings dialog, where you can adjust its settings
...
+
+The overlay icon of a synced directory indicates the status of the files in the 
+directory. If there are any sync errors, the directory is marked with a warning 
+icon.
+
+If the files inside the directory include ignored files that are marked with 
+warning icons, the parent directory remains in green mode.
```
ogoffart: There will be a warning sign in case of error, even if there is ignored files.  The current wording seems to indicate that if there are ignored files, then the parent directory will always be green (even if there are other errors.)

Maybe reword it like : "include ignored files that are marked with warning icons does not change the status of the parent directies"

doc/navigating.rst
```diff
@@ -133,6 +133,44 @@ appear as a new tab in the settings dialog, where you can adjust its settings
...
+
+Normally-synced directories will not show an error sign, with the exception of 
+that root sync directory. That will turn red if there is a serious error 
+anywhere inside the root directory.
```
ogoffart: I don't understand this part.   Any directory turns red if there is a "serious" error anywhere into them.  how is the root different?

dragotin:
Applied the feedback from the review here. @carlaschroder feel free to merge please.

doc/navigating.rst
```diff
@@ -133,6 +133,39 @@ appear as a new tab in the settings dialog, where you can adjust its settings
 at 
 any time. Use **Account** > **Remove** to delete accounts. 
 
+File Manager Overlay Icons
+--------------------------
+
+The ownCloud sync client provides overlay icons, in addition to the normal file 
+type icons, for your system file manager (Explorer on Windows, Finder on Mac and
```
guruz: Dolphin?

doc/navigating.rst
```diff
@@ -133,6 +133,39 @@ appear as a new tab in the settings dialog, where you can adjust its settings
...
+file is in sync with the server version, it displays a green checkmark.
+
+If the file is ignored from syncing, for example because it is on your 
+exclude list, or because it is a symbolic link, it displays a warning icon.
+
+If there is a sync error, or the file is blacklisted, it displays an 
+eye-catching red X.
+
+If the file is waiting to be synced, or is currently syncing, the overlay 
+icon displays a blue cycling icon.
```
guruz: I'm not a native speaker, but I'd have said blue icon with white arrows?

MTRichards: >icon displays a blue cycling icon

You could say a blue circular icon, that may be best so it is parallel with previous icon descriptions.

dragotin: almost... ;-)

dragotin:
Thanks, but I'll leave the language fine print to @carlaschroder - I shouldn't even try...

carlaschroder:
Merging. I am wondering if more file managers are supported? @guruz said Dolphin. I'll revise the file manager list later.

dragotin:
Yes, dolphin is also supported, but only in very new versions of KDE Frameworks 5 (>5.16). Only openSUSE Tumbleweed ships that in the range of distros we support, so I wonder if we should already mention it or wait a bit.

carlaschroder:
Meh, wait until it propagates more widely.

Issue #4460, opened 2016-02-10
check quota against the default remote local folder (instead of /) 


moscicki:
Currently the quota check (https://github.com/cernbox/smashbox/blob/master/protocol/protocol.md#quota-check-call) is done against / on the server.

In the case of branded client which defines the remote home folder to something different than /  (for us: /home) -- the quota check should be done against /home folder on the server.

That would only affect branded clients which redefine the default remote sync folder.

ogoffart:
But that folder is only a default folder, users can change that.
The quota-used should be querried on "/" no matter what i think, because the quota is for the whole account.

danimo:
@jancborchardt can you chime in here? We designed the quota to be a property of the account. But if we only sync a sub part as outlined by @moscicki, this number will probably deviate from the number shown on the web interface which gives you the full view. This is because the quota is not uniform, but can be different on certain folders.

moscicki:
Hello, 

I agree with you: quota attribute is a property of the account to which the folder belongs and not of the individual subfolders inside. In other words, no matter which folder you query, if it "belongs" to the same account (or quota node) there should be no difference and the result should always be the same (used space vs total space). Currently it seems there is a problem with implementation on the PHP server which does not respect this assumption.

Please refer to this: https://github.com/owncloud/client/issues/4459

Currently owncloud PHP server seems wrong as it gives the total space occupied by that subfolder only (and it was actually abused by the client instead of oc:size as explained in the ticket above). If this is fixed then there is no problem for you with this behaviour also for branded clients which override the remote directory.

The reason for my request is that in our case, we map URIs to different paths on the backend storage namespace (and these folders do NOT belong to one account). The user home directory is mapped as /home (which is actually a shortcut in the context of authenticated user "bob" to /eos/user/b/bob). We transparently expose the full namespace behind so a user can also define a remote sync folder pair against any other path on the backend (if they have access and not necessarily in their home directory). It is not the same as in the case of opaque storage in the PHP owncloud server, where you assume "/" is the top level home directory. In our case it is not, "/home" is. Yes, user may remove that "/home" sync folder pair and redefine any other. So as such this one number to represent quota information does not make sense but needs to be quantified that it applies to "/home" directory. Then we give a clear and useful piece of information to the user about the quota on his personal storage: /home. Maybe you need another branding option internally for that, I do not know.

[...]

PR #4714, opened 2016-04-21
Quota: Add branding option for the base folder

ogoffart:
As discussed on issue #4460
Having the quote to be queried on subfolder is wrong in the generic case,
so add a branding option to configure it.

This partially reverts commit ff4cdc3161ddbb74a0fe10f969043ff898fbb93a


dragotin:
This is nice :+1:

PR #4717, opened 2016-04-22
Only 1 : between namespace and property

rullzer:
Apparently I did not push the correct final version to https://github.com/owncloud/client/pull/4687 :speak_no_evil: 

Here it is...


rullzer:
Actually let me just merge this...

PR #4724, opened 2016-04-24
Removing propagator jobs dependencies on internal NetworkJob objects

orion1024:
Hello,

This is ongoing work but I create the PR now so that I can get feedback early.

Since with ownCrypt, I want the propagator classes to interact with network jobs exactly the same way whether the given folder/file is encrypted, I need to make sure there is no direct access to the internal state of the jobs.

In practice, it mostly means avoiding direct use of the job's QNetworkReply _reply, and instead getting the needed property (such as URL, HTTP return code, and so on) through a method that can be overloaded later for the encrypted version of the NetworkJob.

Status on the progress :
- [x] LsColJob
- [x] MkColJob
- [x] PropfindJob
- [x] ProppatchJob
- [x] EntityExistsJob
- [x] GETFileJob
- [x] PUTFileJob
- [x] PollJob
- [x] MoveJob
- [x] DeleteJob

The network jobs that are not expected to deal with encrypted data (such as CheckServerJob) should remain unchanged.

### DIFF
--- a/src/libsync/abstractnetworkjob.h
+++ b/src/libsync/abstractnetworkjob.h
@@ -38,6 +38,14 @@ const char owncloudContentLengthHeaderName[] = "Content-Length";
 const char owncloudContentRangeHeaderName[] = "Content-Range";
 const char owncloudSTSHeaderName[] = "Strict-Transport-Security";
 const char owncloudFileIDHeaderName[] = "OC-FileId";
+const char owncloudOCETagHeaderName[] = "OC-ETag";
+const char owncloudETagHeaderName[] = "ETag";
+const char owncloudOCFinishPollHeaderName[] = "OC-Finish-Poll";
+const char owncloudOCMTimeHeaderName[] = "X-OC-MTime";
+#if QT_VERSION < QT_VERSION_CHECK(5, 4, 2)
+const char owncloudShouldSoftCancelPropertyName[] = "owncloud-should-soft-cancel";
+#endif
+const char owncloudCheckSumHeaderName[] = "OC-Checksum";
 
 /**
  * @brief The AbstractNetworkJob class
@@ -91,13 +99,26 @@ public:
     virtual QByteArray replyContentLength();
     virtual bool replyCustomSoftErrorStringIsValid();
     virtual QString replyCustomSoftErrorString();
+    virtual  bool replyHasOCETag();
+    virtual QByteArray replyOCETag();
+    virtual  bool replyHasETag();
+    virtual QByteArray replyETag();
+    virtual  bool replyHasOCFinishPoll();
+    virtual QByteArray replyOCFinishPoll();
+    virtual  bool replyHasOCMTime();
+    virtual QByteArray replyOCMTime();
+#if QT_VERSION < QT_VERSION_CHECK(5, 4, 2)
+    virtual bool replyShouldSoftCancelIsValid();
+    virtual QString replyShouldSoftCancel();
+#endif
+    virtual  bool replyHasOCChecksum();
+    virtual QByteArray replyOCChecksum();
+
 public slots:
     void setTimeout(qint64 msec);
     void resetTimeout();
 signals:
-    //TODO MBO : remove this one if it's not used anymore once all dependencies are fixed.
     void networkError(QNetworkReply *reply);
-    void networkError();
     void networkActivity();
 protected:
     void setupConnections(QNetworkReply *reply);
--- a/src/libsync/abstractnetworkjob.h
+++ b/src/libsync/abstractnetworkjob.h
@@ -59,11 +59,32 @@ public:
 
     qint64 timeoutMsec() { return _timer.interval(); }
 
+    // These methods are used to replace the direct calls that were used to fetch the same information.
+    // Can and should be overloaded when needed by specific implementation.
+
+    virtual void abortNetworkReply();
+
+    virtual QUrl getUrl();
+    virtual QNetworkReply::NetworkError getError();
+    virtual int getHttpStatusCode();
+    virtual QString getHttpReasonPhrase();
+    virtual QString getErrorString();
+    virtual bool getHasOCErrorString();
+    virtual QByteArray getOCErrorString();
+    virtual  bool getHasOCFileID();
+    virtual QByteArray getOCFileID();
+    virtual QByteArray replyReadAll();
+    virtual QString getContentTypeHeader();
+    virtual QUrl getRedirectionTarget();
+    virtual bool getHasSTS();
+
 public slots:
     void setTimeout(qint64 msec);
     void resetTimeout();
 signals:
+    //TODO MBO : remove this one if it's not used anymore once all dependencies are fixed.
     void networkError(QNetworkReply *reply);
+    void networkError();
     void networkActivity();
 protected:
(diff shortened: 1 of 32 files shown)


src/3rdparty/owncrypt
```diff
@@ -0,0 +1 @@
+Subproject commit 54d45ce105bb91ed9ec565a44daa6f583f3a826d
```
ogoffart: Did you already intent to commit that?

src/libsync/abstractnetworkjob.h
```diff
@@ -59,11 +59,32 @@ class OWNCLOUDSYNC_EXPORT AbstractNetworkJob : public QObject {
 
     qint64 timeoutMsec() { return _timer.interval(); }
 
+    // These methods are used to replace the direct calls that were used to fetch the same information.
+    // Can and should be overloaded when needed by specific implementation.
+
+    virtual void abortNetworkReply();
+
+    virtual QUrl getUrl();
```
ogoffart: The client's coding style is losely based on the Qt coding style. There should not be "get" in front of the function.

src/libsync/networkjobs.cpp
```diff
@@ -328,21 +337,21 @@ bool LsColJob::finished()
         connect( &parser, SIGNAL(directoryListingIterated(const QString&, const QMap<QString,QString>&)),
                  this, SIGNAL(directoryListingIterated(const QString&, const QMap<QString,QString>&)) );
         connect( &parser, SIGNAL(finishedWithError(QNetworkReply *)),
-                 this, SIGNAL(finishedWithError(QNetworkReply *)) );
+                 this, SIGNAL(parserFinishedWithError(QNetworkReply *)) );
```
ogoffart: s/SIGNAL/SLOT/     parserFinishedWithError is a slot.

But actually, you can connect signal with many arguments to slot with less argument. So you can simply do:

```
  connect( &parser, SIGNAL(finishedWithError(QNetworkReply *)),
                 this, SIGNAL(finishedWithError()) );
```

inless you intended to do tomething with the QNetworkReply

ogoffart: Actually, looking at the code, it seems that this signal is unused anyway, is it not?  So maybe just remove it.

ogoffart:
Looks promising.

Please squash the two commits to reduce the noise in the history.

src/3rdparty/owncrypt
orion1024: No, it's supposed to be isolated in another branch for now

src/libsync/networkjobs.cpp
orion1024: It isn't right now, because the parser does its job in one pass synchronously, and the LsColJob knows about errors through LsColXMLParser::parse() return code.
```
 if( !parser.parse( reply()->readAll(), &_sizes, expectedPath ) ) {
            // XML parse error
            emit finishedWithError()
```
But from the comment below if/when the parser works asynchronously, the signal will be needed. So it's better to keep it imho.
```
// TODO: Instead of doing all in this slot, we should iteratively parse in readyRead(). This
// would allow us to be more asynchronous in processing while data is coming from the network,
// not all in one big blob at the end.
```

I'll correct the signal/slot stuff.

orion1024:
Regarding squashes : I wanted to avoid that since it's rewriting history and rewriting public history is supposed to be bad ; are you saying it's fine to do so in this case ?
(yes, still learning my way in github workflow)

Anyway, thanks for the quick feedback

ogoffart:
You are right, rewriting public history is bad.  This depends how much you think your branch is public history.   Because clean history also has value.  I don't strongly insists that you squash.

dragotin:
looks good, thanks. One coding style thing that I'd ask you to consider: Please **always**, just always, use parenthesis after `if()` and friends, even if there is only one line following.

Not 
```
  if( foo ) 
      bar();
```
but 
```
  if( foo ) {
      bar();
  }
```

That would be great!

orion1024:
@dragotin  : no problem, I'd also rather use it that way too.
I only did it because I saw it several times without brackets in the code, so i thought that was the norm.

I squashed the cleanup commits ; there is next to no chance that somebody fetched that branch and coded on top of it after all.

src/libsync/propagatedownload.cpp
```diff
@@ -432,24 +432,23 @@ void PropagateDownloadFileQNAM::setDeleteExistingFolder(bool enabled)
...
+    QNetworkReply::NetworkError err = job->replyError();
 
-    qDebug() << Q_FUNC_INFO << job->reply()->request().url() << "FINISHED WITH STATUS"
-             << job->reply()->error()
-             << (job->reply()->error() == QNetworkReply::NoError ? QLatin1String("") : job->reply()->errorString())
-             << _item->_httpErrorCode
+    qDebug() << Q_FUNC_INFO << job->replyUrl() << "FINISHED WITH STATUS"
+             << err
+             << (err == QNetworkReply::NoError ? QLatin1String("") : job->replyErrorString())
+             << _item->_httpErrorCode // change to job->replyHttpStatusCode() ?
```
orion1024: Is `_item->_httpErrorCode` the same that the HTTP Status code of the reply object in this context ?
If yes, we can standardise and use `job->replyHttpStatusCode()`

orion1024:
Ok, this branch is ready for review @ogoffart

ogoffart:
There is lots of virtual function in AbstractNetworkJob that does not really belong on the AbstractNetworkJob but rather on the subclass where they belong.

ogoffart:
We are wondering if such abstraction could be used to umprove testing.

orion1024:
> There is lots of virtual function in AbstractNetworkJob that does not really belong on the AbstractNetworkJob but rather on the subclass where they belong.

True. When I began adding those methods, some of them were used in several subclasses, so I added them in the parent class in order to avoid code duplication. For these, I think they can stay where they are ?
And then for the ones that were only used in one child class, I thought there was value in providing them for every class, in case they are needed later by another class than the current one.
I checked and about half of the methods are only used on one subclass and could be moved to a child class :
`virtual QUrl replyRedirectionTarget();
virtual QString replyContentTypeHeader();
virtual QUrl replyRedirectionTarget();
virtual bool replyHasSTS();
virtual QByteArray replyContentRange();
virtual QByteArray replyContentLength();
virtual bool replyCustomSoftErrorStringIsValid();
virtual QString replyCustomSoftErrorString();
virtual QByteArray replyOCETag();
virtual QByteArray replyETag();
virtual QByteArray replyOCFinishPoll();
virtual QByteArray replyOCMTime();
virtual bool replyShouldSoftCancelIsValid();
virtual QString replyShouldSoftCancel();
virtual QByteArray replyOCChecksum();`
Do you think some of them might be needed in more than one child class later ? if not, I'll move them.

> We are wondering if such abstraction could be used to improve testing.

Interesting, how so ? are you just talking about this PR or about the other one too ?

orion1024:
any news on this ? thanks

ckamm:
@orion1024 It looks like we've failed to make this happen, unfortunately. Is this PR still relevant or shall we close it?

orion1024:
@ckamm Nope you're right there's not much point into letting this opened, I'm closing it

Issue #3249, opened 2015-05-19
[Nautilus] [Overlays] Multiple status indicator icons overlaid on file/folder


jancborchardt:
Sometimes I see this:
![capture du 2015-05-19 17 23 01](https://cloud.githubusercontent.com/assets/925062/7706623/cef35260-fe4b-11e4-9c42-05ed963c8bf3.png)

And this
![capture du 2015-05-19 17 24 06](https://cloud.githubusercontent.com/assets/925062/7706664/1d2eeb9c-fe4c-11e4-89cf-69ceed52e3e9.png)

Even though the folder / file is perfectly synced. Shouldn’t we make sure that there is always only _one_ indicator on top of each file? Otherwise it’s really confusing and makes me not trust the app. cc @dragotin @danimo

guruz:
Please edit the bug title to mention which file manager etc

jancborchardt:
Sure, sorry. This is with:
* GNOME nautilus 3.14.2
* client 1.8.0
* ownCloud latest stable 7

GuillaumeAmat:
Is anyone know if it will be fixed soon? I see this bug dozen times a day...

sebastien-villemot:
Same here with GNOME nautilus 3.14.1 and ownCloud client 2.0.2.

ckamm:
Pull request #4736

mcastroSG:
Hi hi:

**Steps executed**

1. Create folder1 at Shared sync folder 
2. Copy inside folder1 several files 

**Results**

![doubleoverlay](https://cloud.githubusercontent.com/assets/18436398/14855480/74edcab8-0c94-11e6-9d3e-8d6588078e05.png)

Double overlay icons are shown. This situation remains till the sync is finished.

Client Version: 2.2.0nightly2016427 (build 1571)
Client OS: Ubuntu 14.04 LTE

guruz:
Did you verify if the nightly really contains the merged commit?
Did you try to re-login?

ckamm:
@mcastroSG Build 1571 didn't include the patch, but build 1573 does. I checked by looking at the .tag file in the nightly-build archive and then checking whether the merge of the pull request was included there.

mcastroSG:
Hi hi:

It works finally.

Client Version: Version 2.2.0beta2 (build 1582)
Client OS: Ubuntu 14.04 LTE

PR #4736, opened 2016-04-26
Overlay icons: Fix invalidation with nautilus #3249

ckamm:
This avoids an extra socketapi call after invalidate_extension_info
is called from a STATUS message.

### DIFF
--- a/shell_integration/nautilus/syncstate.py
+++ b/shell_integration/nautilus/syncstate.py
@@ -66,6 +66,7 @@ class SocketConnect(GObject.GObject):
         GObject.timeout_add(5000, self._connectToSocketServer)
 
     def sendCommand(self, cmd):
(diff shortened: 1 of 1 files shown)


Naereen:
Seems fine to me.

dragotin:
:+1:

shell_integration/nautilus/syncstate.py
```diff
@@ -282,13 +301,28 @@ def update_file_info(self, item):
         if item.is_directory():
             filename += '/'
 
+        inScope = False
```
hefee: You can get rid of inScope by the else statement for for loops:
```
   for reg_path in socketConnect.registered_paths:
     if filename.startswith(reg_path):
       [...]
   else: # this is only called, when the for loop ends without a break
     return
```

Issue #3411, opened 2015-07-06
Explorer Integration on Windows Terminal Server only for one user?


it64-markus:
I have a Windows Server 2008 R2 as terminal server with about 20 Users. Some of them use the owncloud client (currently 1.8.3.5213). During installation process i double checked, that the Windows Explorer integration was activated.

But only the first user who starts the owncloud client gets the Explorer integration and can share files with a rightclick on a synced file or folder directly in the Windows Explorer. All other users on the same server don't have the integration.

To make things a bit clear assume we have user A, B and C. All three are using the owncloud client and it is configured to start at login. Let's say currently no users are logged in. User A logs in, the owncloud client gets started and she can use the Explorer integration right away. Then user B logs in, the owncloud client is also started, we see it in the tray. But she can't use the Explorer integration. The same is true for user C, when she logs in after A and/or B.

If User B and A close the owncloud client and B starts it manually after A closed it, then User B suddenly can use the Explorer integration. When A restarts the client after B, then she doesn't have the integration activated.

The users without the integration activated even don't see the green little additions to the folder- or file-icons in the explorer, showing whether the file is in sync with the server or not.

I already de- and re-installed the client numerous times so far.

Is there someone out there with the same issue? Any help possible? Is this targeted for a future release of the client?

guruz:
Have you tried raising this via the enterprise/edu support (if you have a license)?
https://owncloud.org/support/

it64-markus:
No, i have no license for owncloud. I wonder that i am obviously the only one that hits this issue.

danimo:
Please try  https://download.owncloud.com/desktop/daily/ownCloud-2.2.0.6006-nightly20160420-setup-fix-3411.exe

guruz:
@it64-markus Hi! Were you able to test the above .exe?

it64-markus:
Sorry, i only have the production terminalserver of a client. I am sure you understand i can't install a nightly owncloud client there. Unfortunately i have no test environment for this currently.

Bottswana:
I have a testing terminal server environment at work that I may be able to
try and reproduce this issue with if it is still open in a few days.
On 25 Apr 2016 10:07 p.m., "Markus Mann" <notifications@github.com> wrote:

Sorry, i only have the production terminalserver of a client. I am sure you
understand i can't install a nightly owncloud client there. Unfortunately i
have no test environment for this currently.

—
You are receiving this because you are subscribed to this thread.
[...]

PR #4738, opened 2016-04-26
Disambiguate socket API pathes on Windows with user name

danimo:
Addresses #3411

### DIFF
--- a/src/gui/socketapi.cpp
+++ b/src/gui/socketapi.cpp
@@ -67,7 +67,8 @@ SocketApi::SocketApi(QObject* parent)
 
     if (Utility::isWindows()) {
         socketPath = QLatin1String("\\\\.\\pipe\\")
-        + QLatin1String("ownCloud");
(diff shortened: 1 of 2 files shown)


src/gui/socketapi.cpp
```diff
@@ -67,7 +67,7 @@ SocketApi::SocketApi(QObject* parent)
 
     if (Utility::isWindows()) {
         socketPath = QLatin1String("\\\\.\\pipe\\")
-        + QLatin1String("ownCloud");
+        + QLatin1String("ownCloud") + '\\' + qgetenv("USERNAME");
```
ogoffart: I don't know if the username can contains non-ascii, but if that's the case I think this should be  QString::fromLocal8Bit(qgetenv("USERNAME"))

danimo: Done.

guruz:
👍

danimo:
Wait! This is not tested yet!

Issue #4730, opened 2016-04-25
[Overlays] Investigate if we need file watcher based changing of icons


guruz:
@ckamm has made a WiP here: https://github.com/ckamm/owncloud-client/commit/dff74a626cefc3d10117a6812d741cbc2bff23ed

While discussing with @ogoffart and @dragotin we think that it could be that this is not necessary because the file managers will re-request overlay icons if they themselves (via their own watchers) detect a file change.

Important to consider is us acting on our own modifications, we might not want that (timing issues etc..)

FYI @dragotin @jturcotte

ckamm:
@dragotin @jturcotte 

From the discussion in #4681 I gather that we want the blue icon to mean something like "file was was changed locally and no sync run finished yet; or file is currently syncing".

Here is my current impression of what we want:
1. File watcher changes turn a file blue (this ticket).
2. As soon as we detect a file needs changes during a sync run, turn it blue. We might already do this well enough. This is discussed in #4731.
3. File is also blue while it's being propagated. Pretty sure we do that.
4. After a sync run it's no longer blue.

@guruz Why does file managers re-requesting the icon help? If they did that immediately, before the sync run has finished discovery, the client would reply with the green icon. From what I understood we'd like to see the blue icon immediately, without being dependent on a sync run.

We could do this like suggested in ckamm@dff74a6.

guruz:
>  If they did that immediately, before the sync run has finished discovery, the client would reply with the green icon. From what I understood we'd like to see the blue icon immediately, without being dependent on a sync run.

Yes we want to see the blue icon immediatly. I was under the assumption we had code that (when querying a file status from the file manager plugin) queries the DB mtime and compares it to the fs mtime and if not match then it returns the blue icon.
But I think that was removed for performance reasons (which is fine with me..)

guruz:
About your patch https://github.com/ckamm/owncloud-client/commit/dff74a626cefc3d10117a6812d741cbc2bff23ed : You clean the dirty paths before actually syncing (good). But what about new dirty watches during the sync? E.g. the OS might notify you 1 sec after you think you're finished with downloading a new file. And then you think the file is dirty/blue again?

ckamm:
@guruz Yes, that's not addressed yet. We could use the same mechanism we use for preventing that stray self-caused notifications trigger a sync. (see Folder::slotWatchedPathChanged)

ckamm:
Pull request: #4739

mcastroSG:
[...]

PR #4739, opened 2016-04-26
Overlay icons: Track touched files #4730

ckamm:
This uses the file watcher to keep track of files that were modified
in order to assign them the blue icon.

This is transient state that's not persisted across restarts.

### DIFF
--- a/src/libsync/syncfilestatustracker.cpp
+++ b/src/libsync/syncfilestatustracker.cpp
@@ -81,6 +81,8 @@ SyncFileStatusTracker::SyncFileStatusTracker(SyncEngine *syncEngine)
               this, SLOT(slotAboutToPropagate(SyncFileItemVector&)));
     connect(syncEngine, SIGNAL(itemCompleted(const SyncFileItem&, const PropagatorJob&)),
             this, SLOT(slotItemCompleted(const SyncFileItem&)));
+    connect(syncEngine, SIGNAL(started()),
+            SLOT(slotClearDirtyPaths()));
 }
 
 SyncFileStatus SyncFileStatusTracker::rootStatus()
@@ -143,6 +145,9 @@ SyncFileStatus SyncFileStatusTracker::fileStatus(const QString& systemFileName)
         return SyncFileStatus(SyncFileStatus::StatusWarning);
     }
 
+    if ( _dirtyPaths.contains(fileName) )
+        return SyncFileStatus::StatusSync;
+
     SyncFileItem* item = _syncEngine->findSyncItem(fileName);
     if (item) {
         return fileStatus(*item);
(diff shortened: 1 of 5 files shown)


guruz:
Code looks good to me 👍 
Let's wait until @jturcotte gives his OK to merge this..

src/libsync/syncfilestatustracker.cpp
```diff
@@ -157,6 +162,20 @@ SyncFileStatus SyncFileStatusTracker::fileStatus(const QString& systemFileName)
...
+    QString fileName = systemFileName.normalized(QString::NormalizationForm_C);
+
+    QString folderPath = _syncEngine->localPath();
+    if (!fileName.startsWith(folderPath))
+        return;
+
+    QString localPath = fileName.mid(folderPath.size());
+    _dirtyPaths.insert(localPath);
+
+    emit fileStatusChanged(systemFileName, SyncFileStatus::StatusSync);
```
ogoffart: `systemFileName`?    Are you sure this should not be `fileName`
At least that's what all the other stuff are doing.

src/libsync/syncfilestatustracker.cpp
```diff
@@ -157,6 +162,20 @@ SyncFileStatus SyncFileStatusTracker::fileStatus(const QString& systemFileName)
     return SyncFileStatus();
 }
 
+void SyncFileStatusTracker::slotPathTouched(const QString& systemFileName)
+{
+    QString fileName = systemFileName.normalized(QString::NormalizationForm_C);
```
ogoffart: Is that even required?  I think the file system watcher already make sure that the file names have the right normalisation.

ckamm: I don't know, @jturcotte? I added this because `fileStatus` used it and I couldn't determine whether it was necessary or not.

src/libsync/syncfilestatustracker.cpp
```diff
@@ -157,6 +162,20 @@ SyncFileStatus SyncFileStatusTracker::fileStatus(const QString& systemFileName)
...
+
+    QString folderPath = _syncEngine->localPath();
+    if (!fileName.startsWith(folderPath))
```
jturcotte: This should normally always be the case since the folderwatcher watches only files within that folder, so this could be a Q_ASSERT

ckamm: I'm uncomfortable about Q_ASSERTing there, what if due to some reason the paths don't match exactly? missing drive letter, / vs \ or something like that? It's indicative of a bug, yes. Do we ship binaries with Q_ASSERT enabled?

jturcotte: No we don't ship with assert enabled, in that case it will either break something or not work properly, only developers would get the sig abort if something is wrong.

That's however a risk that we should usually be comfortable taking, it forces us to patch a whole product and not just a single file. No strong opinion however, so feel free to follow your guts.

Issue #4737, opened 2016-04-26
Three dots button does not work when sync was was removed and another one was added


mcastroSG:
### Expected behaviour
In case of click over "three dots"  button  that could be found at main tab of Desktop Client a new contextual menu should appear showing some  options as "Open Shared Folder"

### Actual behaviour

In case of click over "three dots" nothing is shown.

### Steps to reproduce
1. Log into ownCloud Desktop Server client
2. Click into systray icon and select Settings option
3. Click into "three dots" button
4. Remove current Shared sync folder
5. Add a new folder sync connection
6. After configure this new folder and wait till sync ends, click over "three dots" button

### Server configuration

Server OS: Ubuntu
Server Version: 9.1.0 Alpha
Storage backend:

### Client configuration

Client OS: W10
Client version: Version 2.2.0beta1 (build 6020)
### Logs

At Client log file:

`04-26 11:38:23:586 0x4dec0c8 unknown: Failed to resolve EGL function eglGetPlatformDisplayEXT (The specified procedure could not be found.)
04-26 11:38:23:602 0x4dec0c8 unknown: static QWindowsEGLStaticContext* QWindowsEGLStaticContext::create(): Could not initialize EGL display: error 0x3001

04-26 11:38:23:602 0x4dec0c8 unknown: static QWindowsEGLStaticContext* QWindowsEGLStaticContext::create(): When using ANGLE, check if d3dcompiler_4x.dll is available
`

This error is found several times at log.

Checked also at W7 with the same behaviour

guruz:
The ANGLE part can be ignored, it's https://github.com/owncloud/client/issues/3968

I had this bug on OS X too some days ago. FYI @ogoffart

ogoffart:
The problem is that the alias of new folder is empty.  As a result, many things are not working with new folder. For example saving them to the config file

mcastroSG:
Hi hi:

**Steps Executed**

1. Log into ownCloud Desktop Server client
2. Click into systray icon and select Settings option
3. Click into "three dots" button
4. Remove current Shared sync folder
5. Add a new folder sync connection
6. After configure this new folder and wait till sync ends, click over "three dots" button

**Results**

It works fine. So Issue closed ;)

Server OS: Ubuntu
Server Version: 9.1.0 Alpha
Storage backend:

Client configuration

Client OS: W10
Client version: Version 2.2.0-nightly20160427 (build 6028)

PR #4741, opened 2016-04-26
Accountsettings: Generate an alias for new folders

ogoffart:
Before commit 1a51b6718a345f1e5f28dd1f4e0c04bdaeeb5357, the wizard was
making sure folder had an alias but this is no longer the case.
So generate still an unique alias.

Alias is not used in the UI any longer, it's just use for internal purposes.

For issue #4737


guruz:
👎 Please consolidate with the code in ```slotAssistantFinished``` by adding a helper method to ```FolderMan```..

ogoffart:
pushed new commit

PR #4743, opened 2016-04-26
Fix aliases once more

guruz:


ogoffart:
Wriong branch!

src/gui/accountsettings.cpp
```diff
@@ -335,12 +335,12 @@ void AccountSettings::slotRemoveCurrentFolder()
         qDebug() << "Remove Folder alias " << alias;
         if( !alias.isEmpty() ) {
             FolderMan *folderMan = FolderMan::instance();
-            QString aliasGui = folderMan->folder(alias)->aliasGui();
+            QString shortGuiLocalPath = folderMan->folder(alias)->shortGuiLocalPath();
 
             QMessageBox messageBox(QMessageBox::Question,
                                    tr("Confirm Folder Sync Connection Removal"),
                                    tr("<p>Do you really want to stop syncing the folder <i>%1</i>?</p>"
```
ogoffart: Maybe the message should be reworded to    "Stop Syncing folder %1 to %2" and whos both the local and the remote paths

ogoffart:
looks good altough maybe we should reword the messages a bit

Issue #4577, opened 2016-03-17
[2.1.1] [Linux] Folder removal confirmation dialog contains wrong folder name


maxxer:
client version 2.1.1, Ubuntu Linux.

1. setup oC with a single account
1. add a second oC account with a new local folder
1. on the second account press the three dots on the folder and then *Remove folder from sync*

The popup which appears contains the name *ownCloud1*, while the locally synced folder name is *oc3j* (in my case)

![folder_name_delete](https://cloud.githubusercontent.com/assets/240201/13844224/c1fd251e-ec38-11e5-8258-9143ea74f0a2.png)

mcastroSG:
Hi hi:

**Steps executed**

1. Install ownCloud Client setting and account.
2. Add a second account setting another shared folder.
3. On second account shared folder click at tree dots button and choose remove folder sync connection.

**Results**

Client shows at popup : "Do you really want to stop syncing the folder ownCloud Testpilot Edition?" . Does not matter which shared folder we want to remove, always shows same text.

Client Version: Version 2.2.0nightly2016421 (build 1557)
Client OS: Ubuntu 14.04 LTE

guruz:
https://github.com/owncloud/client/blob/f443377978f76fc8c141ee6d525ff8076bd5c896/src/gui/folder.cpp#L145

You have two accounts configured or two folders at the same account?
So for both accounts it shows  "Do you really want to stop syncing the folder ownCloud Testpilot Edition?"?

mcastroSG:
@guruz two different accounts and each one ,with one different shared sync folder. ;)

mcastroSG:
Hi hi:

**Steps executed**

1. Install ownCloud Client setting and account.
2. Add a second account setting another shared folder.
3. On second account shared folder click at tree dots button and choose remove folder sync connection.

**Results**

Client shows at popup : "Do you really want to stop syncing the folder XXXX ?" . 
Client Version: Version 2.2.0nightly2016427 (build 1570)
Client OS: Ubuntu 14.04 LTE

@maxxer may you test again with later nightly ?

PR #4744, opened 2016-04-26
UI: Fix account name/alias display oddities #4577

guruz:


guruz:
Was OKed by @ogoffart on https://github.com/owncloud/client/pull/4743#issuecomment-214770319

Regarding the wording: I don't want to create trouble for translators now before the release..

PR #4748, opened 2016-04-27
vio for windows: Use more graceful create file params.

dragotin:
shouldn't we relax the attributes a bit?

### DIFF
--- a/csync/src/vio/csync_vio_local_win.c
+++ b/csync/src/vio/csync_vio_local_win.c
@@ -232,8 +232,7 @@ int csync_vio_local_stat(const char *uri, csync_vio_file_stat_t *buf) {
 
     h = CreateFileW( wuri, 0, FILE_SHARE_WRITE | FILE_SHARE_READ | FILE_SHARE_DELETE,
                      NULL, OPEN_EXISTING,
-                     FILE_ATTRIBUTE_NORMAL+FILE_FLAG_BACKUP_SEMANTICS+
-                     FILE_FLAG_OPEN_REPARSE_POINT+FILE_FLAG_OVERLAPPED,
+                     FILE_ATTRIBUTE_NORMAL | FILE_FLAG_BACKUP_SEMANTICS | FILE_FLAG_OPEN_REPARSE_POINT,
                      NULL );
     if( h == INVALID_HANDLE_VALUE ) {
         CSYNC_LOG(CSYNC_LOG_PRIORITY_CRIT, "CreateFileW failed on %s", uri );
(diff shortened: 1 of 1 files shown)


csync/src/vio/csync_vio_local_win.c
```diff
@@ -230,8 +230,11 @@ int csync_vio_local_stat(const char *uri, csync_vio_file_stat_t *buf) {
...
     mbchar_t *wuri = c_utf8_path_to_locale( uri );
 
-    h = CreateFileW( wuri, 0, FILE_SHARE_READ, NULL, OPEN_EXISTING,
-                     FILE_ATTRIBUTE_NORMAL+FILE_FLAG_BACKUP_SEMANTICS+FILE_FLAG_OPEN_REPARSE_POINT, NULL );
+    h = CreateFileW( wuri, 0, FILE_SHARE_WRITE | FILE_SHARE_READ | FILE_SHARE_DELETE,
+                     NULL, OPEN_EXISTING,
+                     FILE_ATTRIBUTE_NORMAL+FILE_FLAG_BACKUP_SEMANTICS+
+                     FILE_FLAG_OPEN_REPARSE_POINT+FILE_FLAG_OVERLAPPED,
+                     NULL );
     if( h == INVALID_HANDLE_VALUE ) {
```
guruz: I don't think FILE_FLAG_OVERLAPPED is needed, afaik it's only relevant for the opener (=us)..

csync/src/vio/csync_vio_local_win.c
```diff
@@ -230,8 +230,11 @@ int csync_vio_local_stat(const char *uri, csync_vio_file_stat_t *buf) {
     ULARGE_INTEGER FileIndex;
     mbchar_t *wuri = c_utf8_path_to_locale( uri );
 
-    h = CreateFileW( wuri, 0, FILE_SHARE_READ, NULL, OPEN_EXISTING,
-                     FILE_ATTRIBUTE_NORMAL+FILE_FLAG_BACKUP_SEMANTICS+FILE_FLAG_OPEN_REPARSE_POINT, NULL );
+    h = CreateFileW( wuri, 0, FILE_SHARE_WRITE | FILE_SHARE_READ | FILE_SHARE_DELETE,
```
ogoffart: Why do you specify the WRITE flag?

ogoffart:
What is this commit fixing?

I'd say "if it is not borken, don't fix it"

ckamm:
Looks good to me.

dragotin:
@ogoffart why shouldn't we allow another process to write the file? I was hoping that this change makes the client a better citizen, and not preventing other programs from doing their job.

ogoffart:
@dragotin your commit try to open the file for writing if i'm not mistaken, which will not work in read only share. Or does it?

ogoffart:
[I deleted the previous comment because i was misunderstanding your commit, confused access flag and shared mode]

So then I have no longer anything against this commit.

However it still is a risky change because it may have undesired side effect (say, when syncing files that are being modified)

PR #4751, opened 2016-04-27
Shibboleth: Load username from config for UI

guruz:
For https://github.com/owncloud/enterprise/issues/1034

### DIFF
--- a/src/gui/creds/shibbolethcredentials.cpp
+++ b/src/gui/creds/shibbolethcredentials.cpp
@@ -67,8 +67,13 @@ void ShibbolethCredentials::setAccount(Account* account)
 {
     AbstractCredentials::setAccount(account);
 
+    // This is for existing saved accounts.
+    if (_user.isEmpty()) {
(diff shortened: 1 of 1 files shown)


src/gui/creds/shibbolethcredentials.cpp
```diff
@@ -129,9 +134,6 @@ bool ShibbolethCredentials::ready() const
 
 void ShibbolethCredentials::fetchFromKeychain()
 {
-    if (_user.isEmpty()) {
-        _user = _account->credentialSetting(QLatin1String(userC)).toString();
```
ogoffart: The HTTP crendatials still does fetch the user from this point.  I think we should still be consistent with the HTTP credentials.  So perhaps you should keep it there.  Or remove it as well from the HTTP credentials.

guruz:
@ogoffart commit modified..

guruz:
@ogoffart I've tested logging in with another existing username, it properly asks me to use the $correct_username via a dialog

ogoffart:
perfect!  :+1:

Issue #4583, opened 2016-03-18
PDF Data synced down as 0 byte file


pjrobertson:
### Expected behaviour
Desktop client should correctly download synced files from the server
This is a follow on from #1218 since I am still having this problem with the latest software. 

### Actual behaviour
Synced files are downloaded, but have a 0 byte filesize and are empty.

### Steps to reproduce
Unknown 

### Server configuration
Operating system: CentOS 7
Linux 3.10.0-327.4.5.el7.x86_64

Web server: 
Apache 2.4, Owncloud 8.2.3

Database:
MySQL
PHP version:
7.0.3

ownCloud version:
8.2.3

Storage backend:
local

### Client configuration
Client version:
2.1.1

Operating system:
OS X 10.11

### Logs

Client:
Files with 0 bytes is `Paddy工作.pdf` and `护照.pdf`

Here are the relevant logs:
```
03-18 17:18:04:648 0x7fd3b3c45b00 csync_walker: file: /path/Paddy工作.pdf [inode=31548281 size=0]
03-18 17:18:04:648 0x7fd3b3c45b00 _csync_detect_update: Database entry found, compare: 1458285394 <-> 1458285394, etag: (null) <-> 2067f5bbefe4cd83835092bffc761d47, inode: 31548281 <-> 31548281, size: 0 <-> 0, perms:  <-> RDNVW, ignore: 0
03-18 17:18:04:648 0x7fd3b3c45b00 _csync_detect_update: file: path/Paddy工作.pdf, instruction: INSTRUCTION_NONE <<=
[...snip...]
03-18 17:21:39:196 0x7fb4eabc35b0 csync_walker: file: path/护照.pdf [inode=31431473 size=0]
03-18 17:21:39:196 0x7fb4eabc35b0 _csync_detect_update: Database entry found, compare: 1458123862 <-> 1458123862, etag: (null) <-> aa89e77bf74806c8f318ace2c11ed623, inode: 31431473 <-> 31431473, size: 0 <-> 0, perms:  <-> RDNVW, ignore: 0
03-18 17:21:39:196 0x7fb4eabc35b0 _csync_detect_update: file: path/护照.pdf, instruction: INSTRUCTION_NONE <<=
[...snip...CLIENT]
03-18 17:21:42:281 0x7fb4e84215d0 _csync_merge_algorithm_visitor: INSTRUCTION_NONE     client file: path/护照.pdf
[...snip...]
[...]

PR #4753, opened 2016-04-28
PropagateDownload: Throw an error if the file is empty while it shoul…

ogoffart:
…d not have been

If the downloaded file is empty but the PROPFIND previously announced it
should not have been empty, this might mean the file was somehow corrupted
because of a bug on the server and that we should therefore not accept
the file.

Normaly we accept a change between the actual size of the file and what we
got during discovery because the file might have been updated to a new version
inbetween. But after this patch we won't accept the file if it was replaced
by an empty file.

Will help for issue #4583
Also requested by IL for issue 548

### DIFF
--- a/src/libsync/propagatedownload.cpp
+++ b/src/libsync/propagatedownload.cpp
@@ -543,6 +543,14 @@ void PropagateDownloadFileQNAM::slotGetFinished()
         return;
     }
 
+    if (_tmpFile.size() == 0 && _item->_size > 0) {
+        FileSystem::remove(_tmpFile.fileName());
+        done(SyncFileItem::NormalError,
+             tr("The downloaded file is empty despite the server announced it should have been %1.")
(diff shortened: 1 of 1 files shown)


guruz:
👎  I don't like the usage of octetsToString, for debuggability it is better to have the real size (to compare in DB and fs and whatnot)

I'm OK with the code..
What about false positives?

Summoning @dragotin FYI

ogoffart:
@guruz   it is a user visible string.  The actual value is visible in the log.  I think  octetsToString is a perfect.

guruz:
You can display both, add the (realBytesSize) in brackets..

dragotin:
:+1:  I also think the octets thing is fine here.

dragotin:
@ogoffart can you make sure to merge that to 2.2?

PR #4756, opened 2016-04-28
Revert "Disambiguate socket API pathes on Windows with user name"

danimo:
Reverts owncloud/client#4738


danimo:
This should have not been merged at this point.

ogoffart:
There should be a comment in the commit message explaining the reson of a revert.  It looks strange looking at the history

PR #4757, opened 2016-04-28
csync_fnmatch: Use the wide char variant of the win32 api.

dragotin:
@ckamm


ogoffart:
:-1:   This would revert 71827549d6bc8083baa61d8c2a5f3f3198526002

csync_fnmatch is a funciton that is really called A LOT.  and the cost of the allocation is significant!

dragotin:
ok, thanks @ogoffart for catching this.

PR #4764, opened 2016-04-29
Add spinner during sharee search

### LINKED ISSUE #4740: Client does not show a spinner when search for users might take longer
opening an issue as reported in https://github.com/owncloud/core/issues/22652#issuecomment-212309038

If I am incorrect pleaso tick the corresponding box in https://github.com/owncloud/core/issues/22652#issuecomment-211354727

rullzer:
Fixed #4740

When searching for sharees we should display a loading spinner.

Easy one
CC: @ckamm @guruz @dragotin

### DIFF
--- a/src/gui/shareusergroupwidget.h
+++ b/src/gui/shareusergroupwidget.h
@@ -129,6 +129,8 @@ private:
     bool _isFile;
     bool _disableCompleterActivated; // in order to avoid that we share the contents twice
     ShareManager *_manager;
+
+    QProgressIndicator _pi_sharee;
 };
 
 }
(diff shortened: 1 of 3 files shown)


src/gui/shareusergroupwidget.h
```diff
@@ -129,6 +129,8 @@ private slots:
     bool _isFile;
     bool _disableCompleterActivated; // in order to avoid that we share the contents twice
     ShareManager *_manager;
+
+    QProgressIndicator *_pi_sharee;
```
guruz: Please use a normal member, not a pointer..

ckamm: Yes, or set a parent during creation.

rullzer:
Ok normal member now

PR #4769, opened 2016-04-30
Adding CURL check in t6.pl

orion1024:
As t6 uses curl to run, and curl is not installed by default, it's better to check at the beginning and abort then, with a meaningful message, than later in the script, with an obscure one.


ogoffart:
thanks

Issue #4773, opened 2016-05-02
[Win] [2.2.0 beta1] Server messages in share view too pale


michaelstingl:
@dragotin The error message is nearly invisible:

![c988db4a-0d33-11e6-9edf-02f231289337 2](https://cloud.githubusercontent.com/assets/214010/14952115/e184bca0-105d-11e6-86cc-d56fb7cd3a34.png)

Text color switches to black after the user clicks to another application window.

ogoffart:
What are the step to reproduce this error?  My server accept any password.  How do i configure password restrictions?

rullzer:
@ogoffart password_policy app

dragotin:
@ogoffart that is a generic error message. It can be any reply message from the server I guess.

ogoffart:
yes, i'm just trying to get one of these message from the server

mcastroSG:
Tested at OS X and Ubuntu and works .. but blocked by [https://github.com/owncloud/client/issues/4781](url) for W testing

mcastroSG:
Tested at W10 also and working fine

PR #4775, opened 2016-05-02
ShareDialog: show the error message in red

ogoffart:
Do the visual stuff from designer.
The previous code that was ment to change the color in red did not work
and changed it to gray instead.
Also I don't see why there should be a frame.

Issue #4773


rullzer:
Do we really want red?
Anyways. Looks good! :+1:

dragotin:
At one point we should have a class AlertLabel or so.

Issue #1314, opened 2013-12-22
1.5 Desktop client periodic CPU spikes


brycecovert:
Hi,

I'm seeing the 1.5 OSX desktop client spike CPU to 100% every 30 seconds or so. I have about 25k files totaling 47GB. This issue occurs even when there are no recent changes (my guess is it's actually checking to see if anything has changed.)

I have tried running with --logwindow, but the log window freezes after about 2 seconds, so I can't give any more useful information.

I'm running the server on debian on mysql.

Any thoughts?

Thanks,
Bryce

PR #4780, opened 2016-05-03
Fix nautilus and nemo plugin branding strings.

dragotin:
Fix for enterprise #1314


jturcotte:
FYI: You can write "owncloud/enterprise#1314" instead of "enterprise #1314" to make it clickable in github.

https://guides.github.com/features/issues/#notifications

jturcotte:
LGTM

PR #4784, opened 2016-05-04
[shell/windows] Fix the windows status push not working

jturcotte:
Since the windows implementation first does cache lookups using the
path string, directories need to be passed identically as through
RETRIEVE_FILE_STATUS.

Change the convention to never have a trailing slash for directories
in the protocol. This allows the convention to be applied without
having to access the disk (since we'd need to know if the path is
represented by a directory) and also matches the convention of the
rest of the sync engine. Individual file manager plugins are then
responsible of handling pushed paths as not ending with a trailing
slash.

This also:
- Moves the trailing slash removal logic from the SyncFileStatusTracker
  to the SocketApi class
- Remove the unneeded QString::normalized call in fileStatus, since
  this should already be done by the FolderWatcher and plugins

### DIFF
--- a/src/gui/socketapi.cpp
+++ b/src/gui/socketapi.cpp
@@ -56,6 +56,13 @@
 // The second number should be changed when there are new features.
 #define MIRALL_SOCKET_API_VERSION "1.0"
 
+static inline QString removeTrailingSlash(QString path)
+{
+    Q_ASSERT(path.endsWith(QLatin1Char('/')));
+    path.truncate(path.length()-1);
+    return path;
+}
+
 namespace OCC {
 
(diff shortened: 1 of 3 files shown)


ogoffart:
I did a smoke test with dolphin and nautilous and it is working.

jturcotte:
Thanks @ogoffart. I still need @dragotin's blessing, I'm not sure how he still feels about the change.

dragotin:
I tested the branch with nautilus and it works fine. Please merge. 

@mcastroSG please, remember to give the overlays a good test again on Windows, once this is merged and compiled in - probably with a nightly build next week. Thanks.

jturcotte:
Thanks!

PR #4816, opened 2016-05-09
Adding network job factories

orion1024:
Hello,

This is a new PR that lays down more groundwork towards adding ownCrypt.

The goal is to provide 2 factory classes that will handle the construction of NetworkJob objects (one for libsync jobs, one for gui jobs).

A quick example :
Before :
`MkColJob *job = new MkColJob( [args] )`
After :
`MkColJob *job = _factory->createMkColJob([args])`

For now, the factories always creates normal network jobs. But later on, they will enable the creation of either normal or ownCrypt network jobs, depending on whether the node is protected with ownCrypt. The idea is that the class requesting the creation **won't (and won't need to) know whether they are using a normal or ownCrypt version of the job**.
Polymorphism will be needed to achieve that but that's for another PR.

After this PR is merged, direct construction of network job should no longer be possible for most of the jobs : the network jobs constructors are made `protected` instead of `public`, and only the `friend` factory classes can call them. The goal is to enforce all job creations through factories.

Some jobs are spared from the factory requirement when they won't need an ownCrypt version, such as `CheckServerJob`.
But I suppose that for consistency sake I could include them in the factory too, what do you think ?

Note : I still need to finish the GUI factory, the PR is not yet finished.
But since there might be some discussion on my approach, the sooner it starts the better.

### DIFF
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -91,6 +91,7 @@ set(client_SRCS
     notificationwidget.cpp
     notificationconfirmjob.cpp
     servernotificationhandler.cpp
+    guinetworkjobfactory.cpp
     creds/credentialsfactory.cpp
     creds/httpcredentialsgui.cpp
     creds/shibbolethcredentials.cpp
--- a/src/gui/activitylistmodel.cpp
+++ b/src/gui/activitylistmodel.cpp
@@ -25,12 +25,14 @@
 
 #include "activitydata.h"
 #include "activitylistmodel.h"
+#include "networkjobfactory.h"
 
 namespace OCC {
 
 ActivityListModel::ActivityListModel(QWidget *parent)
     :QAbstractListModel(parent)
 {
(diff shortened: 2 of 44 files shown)


orion1024:
I discussed this PR a bit with @ogoffart on IRC and it seems there are concerns about merging this now vs later.
I'd rather merge these changes now in order to drastically reduce the amount of time I spend on catching up with the changes made by other devs.
I also understand that merging this now will mean that others devs will need to adjust how they use and maintain network jobs. The current design was chosen with this fact in mind, and one of the goals was to minimize this overhead as much as possible.
But I might have missed something ; in any case I'm willing to spend the time needed to change/tune things at this stage it if it means an early merge can happen. In the long run, it will save me a lot of time.
Not sure who is involved in this discussion internally so here are some random mentions : @ogoffart @jturcotte @dragotin @danimo @ckamm .

orion1024:
Any news on this ? thanks

orion1024:
Closing the PR, no more work is being done on this.

PR #4818, opened 2016-05-10
For tests we do not need a xserver running.

hefee:
Use for tests QTEST_GUILESS_MAIN and QTEST_APPLESS_MAIN to get rid of a running xserver for tests.


dragotin:
QTEST_GUILESS_MAIN was introduced with Qt5 it seems, hence this breaks the Qt4 build. Can we do that in a way that this does not happen?

Thanks!!

hefee:
Well to have the same experience on qt4 we need to remove the linkage against QtGui. But I see no way to get rid of GUI, cause they are hardcoded in the QtVersionAbstraction.
But this is still only a "nice addon" to have no xserver running, so I think it is okay to leave qt4 with xserver at the moment.

ckamm:
@hefee :+1: 
@dragotin It does not break the Qt4 build, it only removes the gui dependency for Qt5 builds.

dragotin:
yeah, ok. :+1:

thanks @hefee

hefee:
pushed change to 2.2 branch

PR #4835, opened 2016-05-13
A final release version number breaks regex

hefee:
f.ex. a valid verson nr: 2.2.0
-> fist \d+ = 2
-> second \d+ = 2
-> the last \d+ = 0
-> .+ in not matching, because there is nothing to match


dragotin:
Thank you @hefee  I really feel a bit ...stupid ;-)

Issue #4859, opened 2016-05-18
[MultiOS] "Shared" folder in selective sync dialog does not sort folders in the same order than OS


mcastroSG:
### Expected behaviour

Owncloud should show isn the same order as OS folders included at shared sync folder.

### Actual behaviour

Owncloud  does not show in the same order as OS folders included at shared sync folder.

![issueorder](https://cloud.githubusercontent.com/assets/18436398/15358736/3c32f2f2-1d06-11e6-9638-f5fe81eb2864.png)

### Server configuration

Server OS: Ubuntu
Server Version: 9.1.0 Alpha

### Client configuration

Client Version: Version 2.2.0%nil (build 1605).
Client OS: Ubuntu 14.04

Client Version: Version 2.2.0 (build 3358)
Client OS: OSX 10.11 El capitán

ckamm:
@mcastroSG The sort order in the file browser is configurable. The client just can't be consistent with that.

What surprises me is that the capitalized folders are sorted before the lower case ones. I'll check that out.

ckamm:
Actually that's correct given the ordering of QString. @mcastroSG Do you think something needs to be done here?

mcastroSG:
@ckamm as a guide to check proper order I was comparing to default order at SO and it seems that most common format is lower - capital for any letter but current behaviour is lower capital for all letters and after that lower for all letters. So in case this is not implemented or require a big effort. this issue could be changed to minor enhancement

ckamm:
Pull request: #4864

ckamm:
@mcastroSG The patch was merged to master, I think the nightlies are built from there right now.

mcastroSG:
@ckamm tested with Version 2.2.1nightly2016520 (build 1633) and still same behaviour

mcastroSG:
@ckamm after generate a new client from master it seems to be working fine :smile:

PR #4864, opened 2016-05-19
Use natural sorting for filenames if available #4859

ckamm:
Still need to test building against Qt4.


ckamm:
@ogoffart Could I ask you for a review?

ogoffart:
:+1:

PR #4873, opened 2016-05-20
Upload: put the job in the active list while doing the checksum compu…

ogoffart:
…tation.

This fixes an issue in which too many jobs are started un parallel
while uploading many files, which could cause too much memory usage as the
chunks are stored in memory.


jturcotte:
LGTM 👍...

PR #4900, opened 2016-05-25
SyncEngine: invalid the blacklist entry when the rename destination c…

ogoffart:
…hange

IL issue 558

### DIFF
--- a/src/libsync/syncengine.cpp
+++ b/src/libsync/syncengine.cpp
@@ -183,6 +183,12 @@ QString SyncEngine::csyncErrorToString(CSYNC_STATUS err)
 
 }
 
+/**
+ * Check if the item is in the blacklist.
+ * If it should not be sync'ed because of the blacklist, update the item with the error instruction
+ * and proper error message, and return true.
+ * If the item is not in the blacklist, or the blacklist is stale, return false.
+ */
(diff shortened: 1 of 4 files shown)


src/libsync/syncengine.cpp
```diff
@@ -214,6 +214,9 @@ bool SyncEngine::checkErrorBlacklisting( SyncFileItem &item )
         } else if( item._modtime != entry._lastTryModtime ) {
             qDebug() << item._file << " is blacklisted, but has changed mtime!";
             return false;
+        } else if( item._renameTarget != entry._renameTarget) {
+            qDebug() << item._file << " is blacklisted, byt renameTarget changed from" << entry._renameTarget;
```
guruz: byt?

guruz: Could you also add a comment explaining this? Why is it not blacklisted then? for future...

ogoffart: that's what every other branch in this function does.  I don't think one need a comment for that in this branch and not the others.

guruz:
👍 

Also needs to be put into 2.2

guruz:
So you had merged 2.1 in 2.2/master after all? So this is fine now in master?

ogoffart:
yes, it's in 2.2

PR #4902, opened 2016-05-25
Search results are filtered by server. Don't filter a second time.

scolebrook:
Also allows searching on attributes other than displayname in ldap configurations. Search results match web and mobile apps.

### DIFF
--- a/src/gui/shareusergroupwidget.cpp
+++ b/src/gui/shareusergroupwidget.cpp
@@ -72,9 +72,7 @@ ShareUserGroupWidget::ShareUserGroupWidget(AccountPtr account,
 
     _completer->setModel(_completerModel);
     _completer->setCaseSensitivity(Qt::CaseInsensitive);
-#if QT_VERSION >= QT_VERSION_CHECK(5, 2, 0)
-    _completer->setFilterMode(Qt::MatchContains);
-#endif
+    _completer->setCompletionMode(QCompleter::UnfilteredPopupCompletion);
     _ui->shareeLineEdit->setCompleter(_completer);
 
     _manager = new ShareManager(_account, this);


scolebrook:
Currently, search results are only displayed if the label returned by the server contains the search string. If user_ldap has searching by another attribute such as email address, if you enter someones email address, the correct results are returned to the client but the client displays nothing, not even a no results found message if the search string doesn't match the users displayname (which it probably won't).

The server already filters the search results. There's no reason to filter them again in the client.

scolebrook:
Oh. And this works with all versions of QT, not just 5.2+.

scolebrook:
@bboule @gig13

ogoffart:
What are the adventage of it with Qt >5.2 ?   
because te server might not answer immediatly, so it's good to fiter in advance before the answer from the server

guruz:
@ogoffart I guess the main part is actually "Also allows searching on attributes other than displayname in ldap configurations."

@blizzz @rullzer Is this from all server versions?

rullzer:
*should* work on all supported server versions.

The main reason if I remember correctly was indeed that we want to do 'instant' filtering. Waiting for the server is slow.

We only send of the request after 400ms or so (as to not bombard the server with requests)

So now if there is stuff we can filter live it will actually do that.

scolebrook:
The quick filtering in the current version doesn't display search results for the string you've entered. It just displays the results from it's last search that also match the current search string until the server replies with new results. At which point the result list is updated. This is very different from the behaviour in the web, iOS and Android clients which fetch new results from the server.

Have a user with a displayname of user1 and user2. Do a search for user2. You see user2. Delete one character. You see user2 only. Wait a moment for the server to reply and you see user1 and user2. This PR doesn't change that behaviour, but it does at least display the results the server sends instead of nothing at all. The current behaviour produces no search results box, no "no results" type message, no spinner to indicate that anything has happened at all. It looks to a user that the search functionality is broken so they don't use it. Since it works for iOS, Android and web, it's hard to make an argument that it's not broken.

Since you have to wait a half second for the server anyway to get results based on your actual search, wouldn't it just be better to not cache stale search results, do what the user expects and see the same results as in the other searching interfaces with the same search string?

ogoffart:
ok, :+1:    In which  branch should that go?   2.2  or master?

scolebrook:
I'm running this patch in 2.1.1 atm for our users. Working fine. Ideally I'd like to see it in the next 2.2 release rather than wait for 2.3.0. But I don't have a handle on the big picture and will defer to the judgement of more knowledge people :smile:.

guruz:
(Cherry picked into 2.2 by @dragotin https://github.com/owncloud/client/commit/af2b712dc6c8ee2ef45bc15237b36ede92396ba0 )

Issue #2791, opened 2015-02-07
[Meta bug] Clarify and clean HTTP and HTTPS / SSL redirection issues


guruz:
- [x] https://github.com/owncloud/client/issues/1735
- [x] https://github.com/owncloud/client/issues/1118
- [x] https://github.com/owncloud/client/issues/1616

Also:
- [x] https://github.com/owncloud/client/issues/1478

guruz:
Also https://github.com/owncloud/client/issues/3226

FYI @danimo @ckamm @LukasReschke

ckamm:
#3226 is done, but #3251 is a related enhancement proposal

guruz:
Moving to 2.1 as this is not doable in the current timeline

guruz:
Another candidate https://github.com/owncloud/client/issues/3284

guruz:
Discussed with @danimo @ogoffart 

This is interweaved with the captive portal issues

* We always want to follow redirects in the setup Wizard (Done!)

In future:
When having a http:// server URL:
* If we when requesting status.php get a redirect to https:// with same host (port doesn't matter) then we follow it.
* If we when requesting status.php get any other redirect we go offline.

When having a https:// server URL:
* If we get a redirect when requesting status.php (to a URL also mentioning status.php) to https:// we also accept it (assuming there is no SSL errors..)
* If we in normal operation get a SSL warning (e.g. self signed certificate by captive portal) we go offline and retry the connection in some minutes (captive portal) (Still unclear how to best implement this with Qt's API) (note we must still show the SSL warning dialog in case the user manually sets his/her account to login again)

FYI @LukasReschke

LukasReschke:
> If we when requesting status.php get a redirect to https:// with same host (port doesn't matter) then we follow it.

Should we maybe also store this information to prevent MITM once the client gets restarted?

guruz:
We were thinking when we receive a redirect that we accept as "new URL" then we also store that in the config.

LukasReschke:
:+1: :rocket:

SamuAlfageme:
Somehow *related* with this happens when trying to access an https endpoint of the server via `http://` protocol, server just replies with 400: Bad Request [instead of redirecting to the SSL endopoint](http://stackoverflow.com/questions/4083221/how-to-redirect-all-http-requests-to-https). 

[...]

PR #4905, opened 2016-05-26
Always follow redirects in network jobs

danimo:
This is a move away from the original policy where jobs would only follow
redirects in special cases.

Two restrictions are in place:

1. We do not allow protocol downgrades (https -> http) 
2. We stop redirects after we find them looping (e.g. old = new url, or indirectly when looping 10
times).

This is closer to RFC conforming behavior, although currently we will treat
301 replies like they were 302. This is for a separate commit.

### DIFF
--- a/src/libsync/abstractnetworkjob.cpp
+++ b/src/libsync/abstractnetworkjob.cpp
@@ -43,7 +43,7 @@ namespace OCC {
 AbstractNetworkJob::AbstractNetworkJob(AccountPtr account, const QString &path, QObject *parent)
     : QObject(parent)
     , _timedout(false)
-    , _followRedirects(false)
+    , _followRedirects(true)
     , _account(account)
     , _ignoreCredentialFailure(false)
     , _reply(0)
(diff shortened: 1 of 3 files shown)


ogoffart:
:+1:

ogoffart:
Actually, since _folloRedirects  is always true,  we could get rid of it.

danimo:
@ogoffart I intentionally left it there for jobs to customize, but we could just as well kill it.

danimo:
Olivier also mentioned that this change makes HTTP only connections prey to redirects from mobile hotspots, which could mess with data integrity. We need to find a good solution here, even if it means disabling redirects for HTTP only.

guruz:
👎  Please mention the issue ID in the commit and check if this confirms..
https://github.com/owncloud/client/issues/2791#issuecomment-170584575

guruz:
Behaviour change, move to 2.3?

ogoffart:
:+1:

guruz:
We discussed irl that this is not as dangerous as we thought and that we should go for it.

danimo:
@guruz: build has passed. +1?

guruz:
I don't want to take responsibility for this
@ogoffart already approved.
I guess it's OK.

Issue #4856, opened 2016-05-17
[2.2.0] Upload progress bar displays confusing values for filesize and number of files


michaelstingl:
![upload-progress-error](https://cloud.githubusercontent.com/assets/214010/15334955/9ff5553c-1c70-11e6-87ad-a0f62950ad1c.gif)

### Expected behaviour
If I upload one 26MB file, the upload progress bar should display exactly "file 1 of 1" and "X MB of 26MB"

### Actual behaviour
The 2.2.0 client shows wrong values. (see animation)

It looks like the client sums up the filesizes and number of files of all files that already exist in my account.

### Steps to reproduce
1. Initial sync download of 25 items with 60MB in total
2. upload one 26 MB file with sync client ==> "X MB of 82 MB, file 25 of 25"
3. upload one more 26 MB file with sync client ==> "X MB of 107 MB, file 26 of 26"
4. upload one more 26 MB file with sync client ==> "X MB of 132 MB, file 27 of 27"

### Server configuration
Operating system:
Debian

Web server: 
Apache

Database:
SQLite

PHP version:
5

ownCloud version:
9.0.2

Storage backend:
local storage

### Client configuration
Client version:
ownCloud-2.2.0.3358.pkg

Operating system:
Mac OS X 10.11.1

OS language:
English

Installation path of client:
/Applications/owncloud.app

@dragotin What happens here? Maybe related: https://github.com/owncloud/client/issues/4832

00005496

guruz:
@michaelstingl There is no "Show package contents" in the right-mouse-button menu of that file on your desktop?

michaelstingl:
@guruz No, the [ownCloud-2.2.0.3358.pkg from our website](https://download.owncloud.com/desktop/stable/ownCloud-2.2.0.3358.pkg) is one compressed file, not a Mac OS X bundle.

guruz:
Similar https://github.com/owncloud/client/issues/4832

FYI @enoch85

enoch85:
@guruz Yes, I saw this issue before, and I can confirm this as well.

ckamm:
Pull request: #4872

frank9999:
This issue seems to be partially fixed. The counter seems to reset after uploads and doesn't sum everything together, which is good.
However the counter still starts with a random number (possibly failed uploads?).

For example:
If I upload 1 file, it says uploading 11 of 11 files.
If I upload 2 files, it says uploading 11 of 12 files.

On another test machine:
[...]

Issue #4832, opened 2016-05-13
[2.2.0]Upload speed and remaining upload time is not shown correctly


Daimonion1980:
### Expected behaviour
Tell us what should happen
Tooltip should show the upload speed. Remaining Upload time gives adequate values.

### Actual behaviour
Tell us what happens instead
Tooltip shows 4kb/s upload although the upload is more than 200kb/s
Remaining upload time shows years on remaining upload although it is just hours

### Steps to reproduce
1. upload a file with client 2.2.0

### Server configuration
Operating system:
debian
Web server: 
apache2
Database:
---
PHP version:
----
ownCloud version:
8.2.4
Storage backend:
----
### Client configuration
Client version:
2.2.0

Operating system:
Win10 x64 prof
OS language:
german
Installation path of client:
---
### Logs
see screenshots below

First screenshot shows the owncloud clietn windows with visible tooltip. second screenshot shows the router upstream where you can see it's fully occupied.
![zwischenablagebild 4](https://cloud.githubusercontent.com/assets/12880413/15239650/9323bbda-18e4-11e6-81e7-6789b36de78a.jpg)
![zwischenablagebild 5](https://cloud.githubusercontent.com/assets/12880413/15239651/9326350e-18e4-11e6-891d-003e041ac7b7.jpg)

enoch85:
I can confirm this. 

**PHP**
7

**MySQL**
5.7

**Server OS**
Ubuntu 16.04

**Client**
2.2.0

**ownCloud version**
9.0.2

guruz:
Are you uploading 350 files as the UI text claims?
If you move the mouse does the tooltip change?

enoch85:
> If you move the mouse does the tooltip change?

Yes.

In my case I upload 1 file and the time left says everything between 5 seconds to 1000 years...

enoch85:
@guruz Can it have something to do with you changing the chunking? Maybe the "time-left-code" needs an update?

Daimonion1980:
In my case 10 files are uploaded. The tooltip changes and shows the files which are uploaded. Double checked this with the logs.

guruz:
@ckamm and @ogoffart think this is because of the way the ignored files are counted(?)

ckamm:
* The remaining time estimate was completely wrong because of #4856, that is fixed for 2.2.1.
* There are issues with remaining-time estimation, see #4428. That's not in scope for 2.2.1.
[...]

PR #4910, opened 2016-05-27
Progress: Don't count dirs without propagation jobs #4856

ckamm:
These directory SyncFileItems are necessary for bookkeeping
but should not influence the progress display at all.


guruz:
👍 Looks fine. There were several issues about wrong progress counts/estimates/displays, maybe you can have a quick look if this fixes all of them..

ckamm:
I only know that it fixes #4856. #4832 is about the upload bandwidth estimate and that's completely separate from this (unfortunately!).

Issue #4898, opened 2016-05-24
OC Client crashes when new certificate is rejected (wifi hotspot)


giovi321:
### Expected behaviour
When owncloud client asks me whether to accept or not a new certificate from the server (it happens every time I connect to a wifi hotspot that requires authentication in the browser) and I answer no (reject the certificate), it should either wait to get a new certificate (the original one from the server and not the one from the wifi hotspot) and ask me again if I accept (if it is a new certificate) or just wait for the new certificate to arrive and start automatically syncing.

### Actual behaviour
When I connect to a wifi hotspot that requires a login, owncloud client prompts me whether to accept or not the new ssl certificate (the wifi hotspot sends its own ssl certificate). If I click no (ie: I reject the certificate), after I login in the wifi hotspot, owncloud needs to be restarted to make sure syncing begins again.

### Steps to reproduce
1. Connect to a wifi hotspot which requires authentication
2. OC Client asks whether to accept or not the new certificate
3. Click no
4. Log in the wifi hotspot
5. Oc doesn't begin to sync again until restarted (and sometimes crashes)

### Server configuration
Operating system:
Debian 8
Web server: 
Apache
Database:
MySql
PHP version:
5.6.20-0+deb8u1
ownCloud version:
9.0.2
Storage backend:

### Client configuration
Client version:
2.1.1 (build 5837)
Operating system:
Windows 7
OS language:
Italian
Installation path of client:
C:\Program Files (x86)\ownCloud\

guruz:
Did it ask you to send this to the crash reporter?

FYI @danimo

guruz:
@giovi321 I just noticed you are on 2.1.1.
Would you please try a 2.2.1 nightly? We're about to release 2.2.1, but want to be sure this crash is fixed.
http://download.owncloud.com/desktop/daily/ownCloud-2.2.1.6112-nightly20160525-setup.exe

giovi321:
Well, it doesn't actually crash every time. Sometimes it just stops syncing and the program has to be restarted. I'll download and try it anyway.                                                                                                                                                                                                                                                                                                                                                        From:notifications@github.comSent:25 May 2016 7:59 amTo:client@noreply.github.comReply to:reply@reply.github.comCc:write@giovi.cr; mention@noreply.github.comSubject:Re: [owncloud/client] OC Client crashes when new certificate is rejected (wifi hotspot) (#4898)  @giovi321 I just noticed you are on 2.1.1.
[...]

PR #4911, opened 2016-05-27
Certs: Re-ask for different cert after rejection #4898

ckamm:
Previously rejecting any kind of certificate meant that the user
was never asked again, even if the certificate changed.

Now we keep track of which certificates were rejected and ask again
if the ones mentioned in the ssl errors change.

mitmproxy is excellent for testing this.


danimo:
Well spotted, :+1:

Issue #4840, opened 2016-05-14
[Linux] [2.2.0] text-/iconsize problems


hefee:
With v2.2.0 the texsize/iconsize is changed when moving the window from one screen to another. I have a multi screen setup (a laptop screen with a external monitor). This behaviour I only seen with owncloud-client.

My Desktop env is a plasma5/kf5/qt5 debian unstable.

![occ-resize](https://cloud.githubusercontent.com/assets/1238146/15267221/0b0f867e-19bc-11e6-9047-8511b48c8cf4.png)
![occ-resize2](https://cloud.githubusercontent.com/assets/1238146/15267222/0b346e44-19bc-11e6-93e3-a20db68a47c0.png)

danimo:
I guess that one screen has a different DPI count than the other? In that case, the problem is that the interface _does not_ change size.

hefee:
No OCC is changing its size itself. I can move the window half on the one screen half on the other. And if i come from the "small text" desktop the text is very small also on the other screen. And if I come from the other side the text is big also on the other screen. It only changed if i moved the window completly.

As I told before. OCC is the only application with that behaviour. All other applications have the same textsize on every screen.

lckarssen:
I observe something similar (but not identical). After upgrading my (Ubuntu) Linux client to v2.2.0, the buttons and icons 'have grown'. The font size seems OK (like it was before upgrading), but the buttons and icons appear as in @hefee's second screenshot.
Note that this is on a laptop with only a single screen. The screen isn't HiDPI: 1920x1080 in a 14" laptop. Other applications are fine.

dennisheitmann:
I can confirm @lckarssen 's observation. It is the same on XFCE 4.12 on xubuntu 16.04. The screen is a HDMI connected 1920x1080 monitor.

dennisheitmann:
![image](https://cloud.githubusercontent.com/assets/8471574/15334924/7b5ad01c-1c70-11e6-81f6-f26310f77256.png)

guruz:
Which Qt versions do you guys use? This kind of stuff is supposed to be better with 5.6

piratenpanda:
5.6.0 here and the same problem on arch

lckarssen:
Just upgraded the client from 2.1.1 to 2.2.0 on a different machine (this one runs Ubuntu 15.10 instead of 16.04). Here the size issue is not present. 

[...]

PR #4922, opened 2016-05-30
Don't try to determine pixel ratio with Qt < 5.6.0

danimo:
This seems to be broken. Worst case users can still set it manually

Addresses #4840

### DIFF
--- a/src/gui/main.cpp
+++ b/src/gui/main.cpp
@@ -53,9 +53,7 @@ int main(int argc, char **argv)
     // want to use sizes relative to the font size everywhere.
     // This is automatic on OS X, but opt-in on Windows and Linux
     // https://doc-snapshots.qt.io/qt5-5.6/highdpi.html#qt-support
-#if QT_VERSION < QT_VERSION_CHECK(5, 6, 0)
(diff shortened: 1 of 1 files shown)


src/gui/main.cpp
```diff
@@ -53,9 +53,10 @@ int main(int argc, char **argv)
     // want to use sizes relative to the font size everywhere.
     // This is automatic on OS X, but opt-in on Windows and Linux
     // https://doc-snapshots.qt.io/qt5-5.6/highdpi.html#qt-support
-#if QT_VERSION < QT_VERSION_CHECK(5, 6, 0)
-    qputenv("QT_DEVICE_PIXEL_RATIO", "auto");
-#else
+    //
+    // Note: QT_DEVICE_PIXEL_RATIO = auto for Qt < 5.6.0 is broken,
```
jturcotte: You don't need this line since you already wrote it in the commit message and the comment isn't useful without that code. The comment also only applies to Linux.

danimo:
@jturcotte Updated

jturcotte:
Thanks, LGTM! 👍

PR #4923, opened 2016-05-30
Fix Qt4 build

RobinGeuze:
Since a signal is protected in Qt4 and we are triggering a signal from upload as well now this is required to unbreak Qt4 builds.


danimo:
LGMT. @ogoffart.?

RobinGeuze:
For reference the commit that broke it was bf02ccc1e8d4649fdb880fa46a8d7196db198dd2.

dragotin:
:+1: pls merge.

PR #4935, opened 2016-06-01
Fix memleak in updater

danimo:
Please review carefully


src/gui/updater/ocupdater.cpp
```diff
@@ -253,11 +254,13 @@ void NSISUpdater::slotWriteFile()
     if(_file->isOpen()) {
         _file->write(reply->readAll());
     }
+    reply->deleteLater();
```
ogoffart: i believe this one is not necessary because it will be done in slotDownloadFinished

Actually, it's wrong because it might delete the reply too early.

ogoffart:
:+1:

PR #4938, opened 2016-06-02
GUI: don't force the environment variable for high DPI on linux.

ogoffart:
Setting the Environment variable only for owncloud makes in inconsistant with
other Qt application running at the same time.

The users can still set it themself for the whole desktop if they wish.

Addresses #4840


danimo:
:+1:

PR #4939, opened 2016-06-02
OS X filewatcher fix

guruz:
I got syncs triggered for even opening a file by double clicking it in Finder
and having launched Preview.app (e.g. for JPEG files).


danimo:
That was to say: LGTM

guruz:
I'm wondering why the mention bot did not trigger

@jturcotte FYI this fixes most cases for me(?). I still got a strange timing issue where a sync was running still or so and then the blue icon stayed like this. Hm, we'll see.

Issue #4927, opened 2016-05-31
[W10][Overlay]Overlay icon is not properly updated in case of sync files uploaded to the server


mcastroSG:
### Expected behaviour
In case I upload a file to the server on browser, Desktop Client should show a proper overlay icon on that file once its synced.

### Actual behaviour
Once Desktop client has synced a file previously uploaded to the server, blue icon remains instead of green check.

### Steps to reproduce
1. On a fresh install set a new Shared sync folder
2. Wait till Desktop client ends sync
3. Use browser to upload a file to the folder synced
4. Wait till Desktop Client ends this new sync

Blue icon remains on uploaded file

### Server configuration
Server OS: Ubuntu
Server Version: 9.1.0 RC1

### Client configuration

ClientOS: W10. Unable to reproduce it for W81 or W7
Client Version: Versión 2.2.1-nightly20160531 (build 6128) also tested for v2.2.1-rc1 branch

rperezb:
having in mind the timing and how critic this is, I'd suggest to move to the next version, how does it sound @mcastroSG ?

mcastroSG:
Sure @danimo @dragotin @ogoffart , what do you think ¿?

ogoffart:
The problem is that the file system watcher sees a change.

guruz:
@ogoffart Did you see my commit yesterday for OS X?

I also remember @ckamm had implemented a grace period somewhere for "own changes" so they are not re-detected

ogoffart:
The "own change" step is only working while the OwnCloudPropagator is still alive.  But apparently we get these notifications after that.  From the log i have:

```
06-03 09:45:49:780 0x375e8 OCC::SyncEngine::slotItemCompleted: void OCC::SyncEngine::slotItemCompleted(const OCC::SyncFileItem&, const OCC::Pr
opagatorJob&) "folder2/wait_il.cur" INSTRUCTION_NEW 4 ""
06-03 09:45:49:785 0x375e8 OCC::SocketApi::sendMessage: SocketApi:  Sending message:  "STATUS:OK+SWM:C:\Users\pepin\ownCloud\folder2\wait_il.c
[...]
06-03 09:45:49:824 0x375e8 OCC::SyncEngine::finalize: CSync run took  1026
06-03 09:45:49:825 0x375e8 OCC::SocketApi::sendMessage: SocketApi:  Sending message:  "STATUS:OK:C:\Users\pepin\ownCloud\"
[...]

PR #4942, opened 2016-06-03
Folder: Ignore change from the watcher if the file has not changed it…

ogoffart:
…s mtime or size

For issue #4927:
On Windows 10, we get a notification after the sync is finished for file that were
just downloaded. The guard we have against our "own changes" are only working when
the sync is running and the OwncloudPropagator still alive.


ogoffart:
Since mention-bot is not mentioning anyone: @ckamm @guruz

guruz:
Looks ok to me 👍 
I will merge and create a build for @mcastroSG

PR #4943, opened 2016-06-03
Folder: fix removing folder with escaped characters

ogoffart:
FolderDefinition::save and load escapes the alias. We also need to escape
it when we remove it.
New folder can't be created with alias that needs escaping, but old folder
from old config may still exist, and we must allow user to delete them.


ogoffart:
Since mention-bot is not mentioning anyone: @ckamm

guruz:
Looks ok, i will merge 👍 

Better might have been to move it to FolderDefinition where the other write operations are.

PR #4946, opened 2016-06-07
Move touched-files tracking to SyncEngine #4927

ckamm:


guruz:
Looks OK to me but I'll let @ogoffart decide then

Issue #3261, opened 2015-05-21
Feature Request: Enable right click-mailto function for Links in the Windows client.


Kawohl:
Windows client. A right click on a Link should open the dropdown context menu and display  "add this file to a new e-mail" and open the e-mail application and attach the link to the file.
@hodyroff

EDITED: Add  mail subject "$userName has shared $filename with you"

Dianafg76:
@Kawohl Thanks for your feedback, cc @jancborchardt. Would this work for you?

jancborchardt:
Instead of right-click which is hard to discover, this should be done with a button. Next to »Copy link«, there should be »Mail link« – which automatically opens the email app defined as standard and pastes the link into the message body, ideally with a predefined message (as we do in the mobile apps cc @rperezb @MTRichards)

Kawohl:
@dragotin

rperezb:
@jancborchardt being possible i'd add to prepolulate the mail subject in iOS we use:

"$userName has shared $filename with you"

@dragotin we talked about this in Nuremberg 😉

dragotin:
@rperezb yes, but not all I talk about is implemented immediately.... unfortunately :o)

dragotin:
@MTRichards something to consider for the next version maybe? Not too big effort.

MTRichards:
Lol. I just put this on our sheet for discussions...

jancborchardt:
@rperezb @dragotin yeah, sounds good, also with the predefined message.

rperezb:
@Kawohl i edited your issue so that the mail subject criteria is not lost over the comments

ckamm:
Pull request: #4961

mcastroSG:
Hi:

**Steps Executed**
1.- Right click on a file at Shared Sync folder 
2.- Select Share with ownCloud option
3.- Click on Mail Link

**Results**
A new pop-up window appears associated to our default mail app. Subject and Body contains proper info.

PR #4961, opened 2016-06-10
Share dialog: Add 'Mail link' button #3261

ckamm:


ckamm:
It'd be great if someone could test this on Mac also... @guruz ?

rullzer:
:+1:

src/gui/sharelinkwidget.cpp
```diff
@@ -473,6 +485,20 @@ void ShareLinkWidget::slotPushButtonCopyLinkPressed()
...
 
+void ShareLinkWidget::slotPushButtonMailLinkPressed()
+{
+    QString fileName = _sharePath.mid(_sharePath.lastIndexOf('/') + 1);
+
+    QDesktopServices::openUrl(QUrl(QString(
+            "mailto: "
+            "?subject=%1 has shared %2 with you"
+            "&body=%3").arg(
+            _account->credentials()->user(),
```
scolebrook: Is _account->credentials()->user() the sender's username as entered in the application or is it their display name provided by the server?

guruz: From checking the code, it should be the name fetched from the server (e.g. in case of Shibboleth)

guruz:
Works on my 10.11.4, it opens the Apple Mail application.

src/gui/sharelinkwidget.cpp
```diff
@@ -473,6 +485,20 @@ void ShareLinkWidget::slotPushButtonCopyLinkPressed()
 #endif
 }
 
+void ShareLinkWidget::slotPushButtonMailLinkPressed()
+{
+    QString fileName = _sharePath.mid(_sharePath.lastIndexOf('/') + 1);
+
+    QDesktopServices::openUrl(QUrl(QString(
+            "mailto: "
+            "?subject=%1 has shared %2 with you"
```
guruz: Is there some URL percent encoding stuff needed?  
Maybe not since it's openUrl

guruz:
👍

src/gui/sharelinkwidget.cpp
rullzer: Could also just be "I shared %2 with you" right... since you are normally typing in the 1st person..

scolebrook: @rullzer Nice catch. Since the email is coming from your personal email account and client rather than the server, the template text should be in the first person unlike other email templates.

Issue #4877, opened 2016-05-20
Use tokens on client


rullzer:
We want to support tokens in the client.
Users logs in and the client gets a token from the server. And uses that.

This should be in the 9.1 server. We will know more details on monday.

ChristophWurst:
related https://github.com/owncloud/core/issues/24763

mcastroSG:
Hi hi:

I´ve made some test and have find something that does not fit, we are providing a link to token section but you have to be logged on default browser previous to click on link to get there. In other case you´ll be redirected to login window and after login you are not redirected to token section. Is this the expected behaviour ¿?

ogoffart:
No: The expected behaviour is that you are redirected to the token section even if you were not loged in before. This was fixed last week in the server. What version of the server were you using for the testing?

mcastroSG:
@ogoffart, you are right, I´ve tested it on a docker with 9.1.0 RC1 and you are redirected to login window and in case of proper username and password you are redirected to token section. Thanks a lot!!

rperezb:
@SamuAlfageme fyi we were talking about this one in Nuremberg

PR #4963, opened 2016-06-14
GUI: Show link to the page that allow to add a new token

ogoffart:
If owncloud >= 9.1 is detexted:
instead of "Password:" ask for "Password or Token"
and add a link to the ownCloud page that allow to add device token

Issue #4877

### DIFF
--- a/src/gui/creds/httpcredentialsgui.cpp
+++ b/src/gui/creds/httpcredentialsgui.cpp
@@ -14,6 +14,7 @@
  */
 
 #include <QInputDialog>
+#include <QLabel>
 #include "creds/httpcredentialsgui.h"
 #include "theme.h"
 #include "account.h"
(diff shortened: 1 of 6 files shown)


ogoffart:
cc @danimo @dragotin @ChristophWurst 

Note that the ownCloud server's GUI says "device password" and i'm using "device token"  

Maybe there is no need for https://github.com/owncloud/core/issues/25091 after all.

guruz:
> Note that the ownCloud server's GUI says "device password" and i'm using "device token"

Euhm.. can you guys agree to a common naming? 
I'd even say "application token" might make more sense

@PVince81

PVince81:
Yeah, maybe let's also call it "Device password" or "Password" in the client.
The link could be called "Create device specific password"

ogoffart:
After discussion with @ChristophWurst,  we decided to rename it "App Password"

ogoffart:
The difference now is that there is a message: "Click here to request an app password from the browser."

src/gui/wizard/owncloudhttpcredspage.cpp
```diff
@@ -114,6 +114,16 @@ void OwncloudHttpCredsPage::initializePage()
             _ui.lePassword->setText(password);
         }
     }
+    bool hasTokenAuth = ocWizard->account()->serverVersionInt() >= 0x090100;
+    if (hasTokenAuth) {
+        _ui.tokenLabel->setText(tr("<a href=\"%1/index.php/settings/personal#apppasswords\">Click here</a>"
+                        " to requestn app password from the browser.")
```
guruz: "requestn"

src/gui/creds/httpcredentialsgui.cpp
```diff
@@ -31,22 +32,38 @@ void HttpCredentialsGui::askFromUser()
...
+                     "User: %2<br>"
+                     "Account: %3<br>")
+                  .arg(Utility::escape(Theme::instance()->appNameGUI()),
+                       Utility::escape(_user),
+                       Utility::escape(_account->displayName()));
+
+    bool hasTokenAuth = _account->serverVersionInt() >= 0x090100;
+    if (hasTokenAuth) {
+        msg += tr("<br><a href=\"%1/index.php/settings/personal#apppasswords\">Click here</a>"
+                     " to request an app password from the browser.<br>")
```
guruz: s/browser/web interface/ ?

src/gui/wizard/owncloudhttpcredspage.cpp
guruz: s/browser/web interface/ ?

make sure both strings are the same for i18n?

guruz:
👍

dragotin:
:+1: nice

PVince81:
Requires core's https://github.com/owncloud/core/pull/25134

guruz:
Change URL pls
https://github.com/owncloud/core/pull/25134#issuecomment-227462405

ogoffart:
Updated with the new URL

guruz:
👍

PR #4968, opened 2016-06-15
Checksum speedup.

ogoffart:
No need to allocate (and initialize to 0) a 10 MiB buffer for each files, even
when most files are much smaller than that.
So make sure the buffer that we allocate is not bigger than the file size.
And Also 10 MiB is a bit big for a buffer. 500 KiB should be more than enough.
(Too big allocations can cause problem because of memory fragmentation and such)

### DIFF
--- a/src/libsync/filesystem.cpp
+++ b/src/libsync/filesystem.cpp
@@ -496,16 +496,16 @@ QString FileSystem::fileSystemForPath(const QString & path)
 }
 #endif
 
-#define BUFSIZE 1024*1024*10
+#define BUFSIZE qint64(500*1024)  // 500 KiB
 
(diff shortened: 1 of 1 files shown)


ogoffart:
Profiling showed that lot of time was spend initializing the 10MiB QByteArray buffer with 0

ckamm:
:+1:

guruz:
What about https://doc.qt.io/qt-5/qcryptographichash.html#addData-2 ?

ogoffart:
> What about https://doc.qt.io/qt-5/qcryptographichash.html#addData-2 ?

Was not available in Qt4

(But yes, we could use it,  it uses a 1024 bytes buffer on the stack.)

PR #4969, opened 2016-06-15
SyncEngine: make abort and closing more robust

ogoffart:
We first need to set the abort flag to csync and then aborting the discovery
job, otherwise, the discovery thread could start a new job in the mean time.

We also need to make sure that the thread has existed before we destroy the
exclude list.


jturcotte:
LGTM 👍

ogoffart:
Pushed as 92309013dae728b66c597697de7572b6d8e50f1a

PR #4970, opened 2016-06-15
SocketAPI: Push new status of dirty files regardless when not synced

jturcotte:
The FolderWatcher inserts files to be marked as SYNC and we
currently assume that all file statuses will be updated by the
following sync. It's however possible that the FolderWatcher
notify us of a change that csync won't consider necessary to
propagate, in which case a new status wouldn't be pushed and
the file manager would continue showing this file as syncing.

Re-push the file status when emptying the dirty files list
before propagating to avoid this issue, most likely the OK
status.


guruz:
👍

PR #4979, opened 2016-06-17
Memory improvements during propagation

jturcotte:


src/libsync/propagateupload.cpp
```diff
@@ -268,6 +270,8 @@ void PropagateUploadFileQNAM::slotComputeTransmissionChecksum(const QByteArray&
 
     connect(computeChecksum, SIGNAL(done(QByteArray,QByteArray)),
             SLOT(slotStartUpload(QByteArray,QByteArray)));
+    connect(computeChecksum, SIGNAL(done(QByteArray,QByteArray)),
+            computeChecksum, SLOT(deleteLater()));
     const QString filePath = _propagator->getFilePath(_item->_file);
```
ogoffart: I beieve the download job has the same problem.

src/libsync/utility.h
```diff
@@ -121,7 +121,7 @@ namespace Utility
 
     class OWNCLOUDSYNC_EXPORT StopWatch {
     private:
-        QHash<QString, quint64> _lapTimes;
+        QMap<QString, quint64> _lapTimes;
```
ogoffart: maybe a QVector<QPair<QString, int>>

src/libsync/propagateupload.cpp
ckamm: Yes. But note that the instances are parented to the propagation jobs, so it shouldn't be a big deal.

ckamm:
Looks good to me.

src/libsync/propagateupload.cpp
dragotin: Do we have a double free now? Windows is more picky about that, in case Qt is not taking care of this... which I don't know.

jturcotte: All those are optimizations for within propagation, we don't have real memory leaks.
@dragotin You can mix deleteLater, call them multiple times and have them mixed with QObject parenting without problem as long as you don't have a dangling pointer (there isn't any in this case).

Issue #4967, opened 2016-06-15
Excessive CPU usage ownCloud 2.2.1


jabdoa2:
### Expected behaviour
When I change one file, ownCloud should synchronise that file. CPU usage should stay low.

### Actual behaviour
When I change one file, ownCloud starts to scan the sync folder and uses 100% of CPU for about some seconds. If you changed another file in the meantime it rescans right away and stays at 100% CPU for minutes. This is annoying if you keep editing files (which is not uncommon).

This did not happen in older clients (at least they were not keeping the CPU at 100% all the time).

### Steps to reproduce
1. Start OwnCloud client
2. Edit a file in the sync folder

### Server configuration
Operating system: Ubuntu 14.04

Web server: Apache

Database: MySQL

PHP version: 5.6

ownCloud version: 8.2.5

Storage backend: Files

### Client configuration
Client version: 2.2.1

Operating system: (L)Ubuntu 14.04

OS language: DE

Installation path of client: /usr/bin/owncloud

### Logs

1. Output of `owncloud `
```
06-15 13:44:26:513 0x22e2460 detected changes in paths: QSet("/folder/xxx.myfile")
06-15 13:44:26:513 0x22e2460 Schedule folder  "ownCloud"  to sync!
06-15 13:44:26:514 0x22e2460 Sync state changed for folder  "https://url.xx/remote.php/webdav/" :  "Not yet Started"
06-15 13:44:26:514 0x22e2460 void OCC::ServerNotificationHandler::slotFetchNotifications(OCC::AccountState*) Account "xxx@url.xx" does not have notifications enabled.
06-15 13:44:26:514 0x22e2460 Scheduling a sync in 2 seconds
[...]
06-15 13:44:39:859 0x22e2460 No database Transaction to commit
06-15 13:44:39:878 0x22e2460 CSync run took  11355
```
2. Web server error log: No errors

3. ownCloud log (data/owncloud.log): No problem here

guruz:
@jabdoa2 This is interesting. We always rescan all paths if a single file changes and on startup, this is nothing that changed over the different versions.

How many files/folders do you have? It should not be at 100% for minutes, maybe for some seconds?

jabdoa2:
Maybe it did it less efficient/less often before? As you see in the log it takes about 13s for one run. On this machine some folders are excluded. I got about 62k files to synchronise:
```
$ find | wc -l
62025
```

ogoffart:
[...]

PR #4988, opened 2016-06-20
Performance: Don't redundantly add the same exclude files #4967

ckamm:


src/gui/folder.cpp
```diff
@@ -695,18 +695,6 @@ void Folder::wipe()
 
 bool Folder::setIgnoredFiles()
 {
-    ConfigFile cfg;
```
guruz: This commit is bogus, where is the loading done? Did you forget a git add?

ckamm: Loading is done in application.cpp, but something isn't working. Need to check.

guruz:
👍

csync/src/csync_exclude.c
```diff
@@ -44,6 +44,18 @@
...
+    // in the list first.
+    if (*inList) {
+        for (i = 0; i < (*inList)->count; ++i) {
+            char *pattern = (*inList)->vector[i];
+            if (c_streq(pattern, string)) {
+                return 0;
+            }
+        }
+    }
     return c_strlist_add_grow(inList, string);
```
dragotin: The return value of  `c_strlist_add_grow()` is either 0 (success) or -1. So we can not distinguish if it was added or not. I suggest to return -2 in case the pattern is already in the list. That way we can avoid a misleading log entry in `csync_exclude_load()`. 

I'll add a patch.

guruz: I agree... although the log message is _before_ the add call, so you need to move/rephrase the log message too..

dragotin: no, not working, we can not add a reply < 0 because that would let the loading stop with error condition. +2 works.

dragotin: 99159049e9a5ca38f993925386a17ef13f47caaf should regulate that.

PR #4989, opened 2016-06-20
excludes: Only log if the pattern was really logged.

dragotin:
Otherwise we have the log line even though the pattern was not logged.


guruz:
Code looks OK.
Also tested it and works here, it does not log duplicate entries

PR #4991, opened 2016-06-21
Setup csync logging earlier

ckamm:
We were missing some csync related log output during startup.

Discovered in #4967


guruz:
Works!
Now of course the startup logging is quite verbose when having multiple folders/accounts but I guess we can live with this..

PR #4993, opened 2016-06-21
Fix a deadlock when shutting down during discovery

jturcotte:
Since the SyncEngine now quits and waits for the discovery thread,
the main thread can enter a deadlock where the discovery thread waits
for its directory result.

Add a 2 seconds timer to the discovery thread wait condition
to limit the deadlock time.

### DIFF
--- a/src/libsync/discoveryphase.cpp
+++ b/src/libsync/discoveryphase.cpp
@@ -564,7 +564,7 @@ csync_vio_handle_t* DiscoveryJob::remote_vio_opendir_hook (const char *url,
         discoveryJob->_vioMutex.lock();
         const QString qurl = QString::fromUtf8(url);
         emit discoveryJob->doOpendirSignal(qurl, directoryResult.data());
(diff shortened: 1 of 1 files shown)


guruz:
But 2 seconds are not enough to do a whole network request in the discovery?

jturcotte:
Humm I didn't know this was handled by a network request, and setting a 30 seconds timeout is a bit too much, so we might need a cleaner fix.

jturcotte:
Changed the timeout to 30s so that we at least don't have to kill the process.
I also have a hard time reproducing the issue, so I thing that this is the best value/cost I can provide for now.

ogoffart:
Note that there is also commit 92309013dae728b66c597697de7572b6d8e50f1a which should help.

jturcotte:
When I tried it was with that patch in. I was wondering if it could be related so I reverted it and it didn't change the behavior for that specific scenario.

Issue #4977, opened 2016-06-17
client takes roughly 24 seconds to start (show GUI or systray icon) on windows


ArthurChiao:
### Expected behaviour
Client takes few seconds to launch (show GUI or systray icon) on windows.

### Actual behaviour
Clients takes roughly 24 seconds to launch (show systray icon), which seems to be abnormal.  The owncloud process is created rightly, which could be seen in windows task manager.

### Steps to reproduce
1.  Install windows client, version 2.2.1.6146
2. Double click to launch client

### Client configuration
Client version: 2.2.1.6146

Operating system: Windows 7, 64 bit

OS language: Chinese

Installation path of client: `C:\Program Files (x86)\ownCloud`

### Debug
I debugged into source code, and found that line 108 **takes 22 seconds to return**, in file `src/gui/socketapi.cpp`,
```c
 106     QFileInfo info(socketPath);
 107     if (!info.dir().exists()) {
 108         bool result = info.dir().mkpath(".");
 109         DEBUG << "creating" << info.dir().path() << result;
 110         if( result ) {
 111             QFile::setPermissions(socketPath,
 112                                   QFile::Permissions(QFile::ReadOwner+QFile::WriteOwner+QFile::ExeOwner));
 113         }
 114     }
```
it tries to create a file `\\.\pipe\ownCloud\<user name>` on windows, then creates a `QLocalServer` to listen on this file/address. And although the return result is always `false`, my client (at least seemingly) works well. 

If i hard code this line to `bool result = false;`, client will launch in 2 seconds, and everything seems fine. My question is: **what does this local socket used for? and this is a problem needs to be fixed?**

guruz:
FYI @danimo @jturcotte @ogoffart

guruz:
Thanks for digging so far.

I know that @danimo added another directory component in e833d01288b40ab72de9559755b05a327e923f50

Did you use previous ownCloud client versions where this did not happen?

guruz:
@ArthurChiao The SocketApi is used for communication between the DLLs in the Explorer for showing the sync overlay icons. 

The question is indeed why does it behave like this on your machine.
Do you have any firewalls/antivirus that could cause this?

ArthurChiao:
@guruz thanks for detailed reply!

My windows firewall is turned off and i think i have no antivirus softwares installed. I'll try more previous versions tomorrow. 

[...]

PR #4997, opened 2016-06-22
SocketAPI: Use different pipe name #4977

ckamm:
Named pipe names that contain backslashes sometimes lead to problems.

### DIFF
--- a/shell_integration/windows/OCUtil/CommunicationSocket.cpp
+++ b/shell_integration/windows/OCUtil/CommunicationSocket.cpp
@@ -45,7 +45,7 @@ std::wstring getUserName() {
 std::wstring CommunicationSocket::DefaultPipePath()
 {
 	auto pipename = std::wstring(L"\\\\.\\pipe\\");
-	pipename += L"ownCloud\\";
+	pipename += L"ownCloud-";
 	pipename += getUserName();
 	return pipename;
 }
--- a/src/gui/socketapi.cpp
+++ b/src/gui/socketapi.cpp
@@ -74,7 +74,7 @@ SocketApi::SocketApi(QObject* parent)
 
     if (Utility::isWindows()) {
         socketPath = QLatin1String("\\\\.\\pipe\\")
-                + QLatin1String("ownCloud") + '\\'
+                + QLatin1String("ownCloud-")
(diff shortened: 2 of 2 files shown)


dragotin:
Any rationale why this will help?

guruz:
@dragotin https://github.com/owncloud/client/issues/4977#issuecomment-227656955

guruz:
@jturcotte @danimo After reviewing/merging could you update the binary submodule and create a nightly for @ArthurChiao to test?

danimo:
:-1: Do not apply magic thinking. Backslashes are common in pipe names.

guruz:
@danimo This is not magic thinking. Check the other bug report, @ArthurChiao tested this.

```The pipe name string specified by PipeName can include any character other than a backslash, including numbers and special characters. ```
https://msdn.microsoft.com/en-us/library/windows/desktop/aa365783(v=vs.85).aspx

jturcotte:
See also http://stackoverflow.com/questions/3571422/can-named-pipe-names-have-backslashes

According to the question and answer, it seems like it could fail of something creates a pipe with the name up to the first slash, which happens to be the old pipe name. So an older version of the client would prevent the new version from creating its pipe, even though the new one would have the extension installed, leaving neither of them working.

Not sure if that's the issue that the user is seeing though.

jturcotte:
Humm reading https://github.com/owncloud/client/issues/4977#issuecomment-227412471, he built the windows client himself and commenting out the backslash from the path does make `mkpath` return faster.

The question is why are we trying to do mkpath on `\\.\pipe\ownCloud\username` on windows and set permissions?

But even then, replacing that backslash with a dash is likely going to fix it too according to his investigation. So since I don't see an advantage to use a backslash over a dash, ideally we'd do both removing that mkpath on windows, and remove the backslash from the pipename to avoid mixed client/shellintegration versions to prevent newer versions from creating their pipe.

jturcotte:
Tested the change and it doesn't break anything, I'll merge this pull request and create a new one to update binaries.

ArthurChiao:
Tested with `ownCloud-2.2.3.6203-nightly20160627-setup.exe`, problem [#4977](https://github.com/owncloud/client/issues/4977#issuecomment-227412471) is fixed.

Issue #4994, opened 2016-06-21
[Win][v2.2.2][Regression] HiDPI not working


michaelstingl:
### 2.1.1

<img width="711" alt="211-server" src="https://cloud.githubusercontent.com/assets/214010/16230436/b9ce3872-37c1-11e6-99ac-6f38a4108959.png">
<img width="726" alt="211-credentials" src="https://cloud.githubusercontent.com/assets/214010/16230437/b9cf8222-37c1-11e6-81c2-17e0ee75dfc4.png">

### 2.2.2

<img width="464" alt="221-server" src="https://cloud.githubusercontent.com/assets/214010/16230448/c8d8fdac-37c1-11e6-9b49-dbe4d8f1058a.png">
<img width="433" alt="221-credentials" src="https://cloud.githubusercontent.com/assets/214010/16230449/c8d95edc-37c1-11e6-87d4-f19700d72f06.png">

### Expected behaviour
HiDPI under Windows would be very beautiful

### Actual behaviour
HiDPI seems to not work

### Steps to reproduce
1. Run Win VM (modern.ie) in VMware Fusion 8.1.1 on Retina MBP 15"
2. Install ownCloud-2.2.2.6192-setup.exe
3. Start ownCloud client and configure account

### Client configuration
Client version:
v2.2.2

Operating system:
Win 8.1 32-bit

OS language:
eng

guruz:
@danimo removed the ```qputenv("QT_DEVICE_PIXEL_RATIO", "auto");``` call in 0194ebb2229adc33f5b9790231a3384b46474bde
This was for https://github.com/owncloud/client/issues/4840 which only is reported on Linux?

The patch 6ccd9e872d53f5a7b3ce2d82be0c0e2d99c013f7 by @ogoffart might not matter (even though the comment at the ```#endif``` is wrong)

ckamm:
@michaelstingl Do you get the correct behavior if you set the `QT_DEVICE_PIXEL_RATIO` environment variable to `auto`?

michaelstingl:
@ckamm I've only little Windows experience. Where/how can I set the QT_DEVICE_PIXEL_RATIO environment variable to auto?

ckamm:
@michaelstingl I think this is one way: Run `cmd`.
```
$ set QT_DEVICE_PIXEL_RATIO=auto
$ "C:\Program Files (x86)\ownCloud\owncloud.exe"
```
The actual path to the owncloud binary may be different for you.

[...]

PR #4998, opened 2016-06-23
Fix windows HiDPI #4994

ckamm:
This was removed in 0194ebb2229adc33f5b9790231a3384b46474bde
because it breaks on Linux. However, it looks like it is correct
for Windows. In the meantime the surrounding ifdef has changed
from !Q_OS_MAC to Q_OS_WIN, so reverting it makes sense.

### DIFF
--- a/src/gui/main.cpp
+++ b/src/gui/main.cpp
@@ -58,6 +58,8 @@ int main(int argc, char **argv)
     // not set this envoronment variable)
 #if QT_VERSION >= QT_VERSION_CHECK(5, 6, 0)
     qputenv("QT_AUTO_SCREEN_SCALE_FACTOR", "1");
+#else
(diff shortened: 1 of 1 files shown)


guruz:
Let's have @michaelstingl test this with the next nightly

michaelstingl:
@guruz link or it didn't happen

guruz:
http://download.owncloud.com/desktop/daily/ownCloud-2.2.3.6209-nightly20160628-setup.exe

michaelstingl:
@guruz [ownCloud-2.2.3.6209-nightly20160628-setup.exe](http://download.owncloud.com/desktop/daily/ownCloud-2.2.3.6209-nightly20160628-setup.exe) looks good:

ie11_-_win8_1_1
ie11_-_win8_1

PR #4999, opened 2016-06-23
Excludes: Use QRegularExpression for speedup WiP

guruz:
Unfortunately this does not speed up thins significantly yet, i only see csync_exclude_traversal going from 14.5% of time to 14.0% of time or so

But maybe I need to check if the JIT is really used and also add some of the other patterns that hardcodedly use fnmatch to the hook.

FYI @ckamm 

It changes our exclude list to those two lists:

```
".*~|~\\$.*|\\.~lock\\..*|~.*\\.tmp|\\._.*|desktop\\.ini|\\..*\\.sw.|\\..*\\..*sw.|\\.fseventd|\\.apdisk|\\.htaccess|\\.directory|.*\\.part|.*\\.filepart|.*\\.crdownload|.*\\.kate-swp|.*\\.gnucash\\.tmp-.*|\\.synkron\\..*|\\.sync\\.ffs_db|\\.symform|\\.symform-store|\\.fuse_hidden.*|.*\\.unison|\\.nfs.*|My Saved Places\\."
".*\\.~.*|Icon\r.*|\\.DS_Store|\\.ds_store|Thumbs\\.db|\\.TemporaryItems|\\.Trashes|\\.DocumentRevisions-V100|\\.Trash-.*"
```

Maybe I can avoid having two regexp objects by using capture groups or something like that?

### DIFF
--- a/src/libsync/discoveryphase.cpp
+++ b/src/libsync/discoveryphase.cpp
@@ -610,6 +610,74 @@ void DiscoveryJob::remote_vio_closedir_hook (csync_vio_handle_t *dhandle,  void 
     }
 }
 
+static QString convertToRegexpSyntax(char *exclude)
+{
+    //QString::fromUtf8(exclude);
+    QString s;
+    for (unsigned int i = 0; i < strlen(exclude); i++) {
+        char &c = exclude[i];
+        if (c == '*') s += ".*";
+        else if (c == '.') s += "\\.";
+        else if (c == '$') s += "\\$";
+        else if (c == '?') s += ".";
+        else s += c;
+        // FIXME: [ ] { } ( ) etc
+    }
+    qDebug() << s;
+    return s;
+}
+
+CSYNC_EXCLUDE_TYPE DiscoveryJob::excluded_traversal_hook (const char *path, int filetype, void *userdata)
+{
+    DiscoveryJob *discoveryJob = static_cast<DiscoveryJob*>(userdata);
+
+    // FIXME: csync_e
+
+    // FIXME file match vs directory match?
+
+    qDebug() << discoveryJob->_exclude_traversel_regexp_exclude.isValid() << path << discoveryJob->_exclude_traversel_regexp_exclude.errorString();
+
+    if (discoveryJob->_exclude_traversel_regexp_exclude.pattern().length() == 0) {
+        // env QT_ENABLE_REGEXP_JIT 1
+        discoveryJob->_exclude_traversel_regexp_exclude.setPatternOptions(QRegularExpression::OptimizeOnFirstUsageOption);
+        discoveryJob->_exclude_traversel_regexp_exclude_and_remove.setPatternOptions(QRegularExpression::OptimizeOnFirstUsageOption);
+        QString _exclude_traversel_regexp_exclude;
+        QString _exclude_traversel_regexp_exclude_and_remove;
+        for (unsigned int i = 0; i < discoveryJob->_csync_ctx->excludes->count; i++) {
+            char *exclude = discoveryJob->_csync_ctx->excludes->vector[i];
+            QString *builderToUse = & _exclude_traversel_regexp_exclude;
+            if (exclude[0] == '\n') continue; // empty line
+            if (exclude[0] == '\r') continue; // empty line
+            if (exclude[0] == ']'){
+                exclude++;
+                builderToUse = &_exclude_traversel_regexp_exclude_and_remove;
+            }
+            if (builderToUse->size() > 0) {
+                builderToUse->append("|");
+            }
+            builderToUse->append(convertToRegexpSyntax(exclude));
+        }
+        qDebug() << _exclude_traversel_regexp_exclude;
+        qDebug() << _exclude_traversel_regexp_exclude_and_remove;
+        discoveryJob->_exclude_traversel_regexp_exclude.setPattern(_exclude_traversel_regexp_exclude);
+        discoveryJob->_exclude_traversel_regexp_exclude_and_remove.setPattern(_exclude_traversel_regexp_exclude_and_remove);
+    }
+
+    QString p = QString::fromUtf8(path);
+    if (discoveryJob->_exclude_traversel_regexp_exclude.match(p).hasMatch()) {
+        qDebug() << "WOULD EXCLUDE";
+        return CSYNC_FILE_EXCLUDE_LIST;
+    }
+    // FIXME
+    if (discoveryJob->_exclude_traversel_regexp_exclude_and_remove.match(p).hasMatch()) {
+        qDebug() << "WOULD EXCLUDE AND REMOVE";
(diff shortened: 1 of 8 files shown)


guruz:
No big change if I put ```qputenv("QT_ENABLE_REGEXP_JIT", "1");```

src/libsync/discoveryphase.cpp
```diff
@@ -610,6 +610,74 @@ void DiscoveryJob::remote_vio_closedir_hook (csync_vio_handle_t *dhandle,  void
     }
 }
 
+static QString convertToRegexpSyntax(char *exclude)
```
ogoffart: QRegExp::escape(exclude).replace("\\*", ".*").replace("\\?", ".")

jvillafanez:
Might be worthy to read: https://www.loggly.com/blog/five-invaluable-techniques-to-improve-regex-performance/

I think we can split the list into several groups:
* For fixed files, such as "Thumbs.db" or ".DS_Store": I'd avoid using regular expressions for these files and compare as a fixed strings.
* For files ending in something, such as "*.part": I'm not sure if leaving it as "\\.part$" would perform better, or if the engine would do some kind of optimization. It's worth to try it.
* For files starting with samething: such as ".fuse_hidden.*": the "^" anchor should be used.

And according to the blog post, the order of the expression could help too

danimo:
Note: This is not Qt4 compatible!

guruz:
@jvillafanez thanks, will check it out.

FYI wrapping in ```^``` and ```$``` did not help.

Maybe I should test with a bigger dataset (testing 3400 files/folders now) to make the constant overhead part be less dominant.

guruz:
Oh well. I removed all the qDebug() I had added, now ```csync_exclude_traversal``` is at ~6.5% vs 14.5% before :)
I'll try to add the hardcoded patterns also into the regexp.

dragotin:
I have to say that I do not like this patch. Which problem are we really solving with this? The code does not really look straightforward, especially not with the hook thing.

Which code would we replace by this?

guruz:
@dragotin I need to clean it up and have it tested, indeed.
For now it's a proof that we can lower the CPU load caused by the discovery by using libpcre because it compiles the excludes into a decision tree instead of going linearly over each call with a slow fnmatch system call.

dragotin:
@guruz there are unit tests for the exclude files, please enhance them...

ckamm:
I've done some work on this here, but it's not done https://github.com/ckamm/owncloud-client/tree/exclude_traversal_qregexp

ckamm:
Note, I've also done performance tests on my version. It re-uses the performance test from check_csync_exclude that runs the default list of excludes against a non-matching path a couple of thousand of times. (it's a good test case, because it's close to what happens a lot in practice)

In my tests I could optimize the regex version until it reached these values:
```
baseline fnmatch: csync_excluded_traversal: 0.001308 ms per call
regex version: csync_excluded_traversal: 0.002224 ms per call
```

So it's still almost 2x slower. :/

The regexes that get used are approximately:
```
against the full path: (?!x)x|/[^/]*/[^/]*\.out$|/latex[^/]*/[^/]*\.run\.xml$|/latex/[^/]*/[^/]*\.tex\.tmp$
against the basename only: (?!x)x|^.*~$|^~\$.*$|^\.~lock\..*$|^~.*\.tmp$|^\._.*$|^desktop\.ini$|^\..*\.sw\.$|^\..*\..*sw\.$|^\.fseventd$|^\.apdisk$|^\.htaccess$|^\.directory$|^.*\.part$|^.*\.filepart$|^.*\.crdownload$|^.*\.kate-swp$|^.*\.gnucash\.tmp-.*$|^\.synkron\..*$|^\.sync\.ffs_db$|^\.symform$|^\.symform-store$|^\.fuse_hidden.*$|^.*\.unison$|^\.nfs.*$|^My Saved Places\.$|^.*\.💩$|^пятницы\..*$
```

guruz:
```
with regexp:
csync_excluded: 0.506731 ms per call
csync_excluded_traversal: 0.006164 ms per call
csync_excluded: 0.507286 ms per call
csync_excluded_traversal: 0.006089 ms per call
csync_excluded: 0.520669 ms per call
csync_excluded_traversal: 0.006532 ms per call

Without:
csync_excluded: 0.516179 ms per call
csync_excluded_traversal: 0.006471 ms per call
csync_excluded: 0.505810 ms per call
csync_excluded_traversal: 0.006022 ms per call
csync_excluded: 0.523512 ms per call
csync_excluded_traversal: 0.005959 ms per call
```

So it improves!
For me on OS X with Qt 5.6.2 with the libpcre shipped with Qt and ckamm's branch on https://github.com/ckamm/owncloud-client/commit/34df8971fa1ed5f50ffe474d89d07c7335d40c8a

We're finding out right now why @ckamm has different behaviour (worse) on Linux.

ckamm:
@guruz Here are new timings for Win7, Qt 5.3.2 in a VM:

```
"fnmatch"
excluded:  0.45
traversal: 0.023

regex
excluded:  0.22 (50% speedup)
traversal: 0.007 (70% speedup)
```

ckamm:
Collected timings

table              | linux (ckamm) | osx (guruz) | win7 (ckamm) | linux (ogoffart) | linux2 (ckamm)
-------------------|---------------|-------------|--------------|------------------|----------------
full (fnmatch)     | 0.038         | 0.506       | 0.45         | 0.031            | 0.039
full (regex)       | 0.106         | 0.507       | 0.22         | 0.039            | 0.11
traversal (fnmatch)| 0.0014        | 0.0060      | 0.023        | 0.0011           | 0.0015
traversal (regex)  | 0.0017        | 0.0061      | 0.007        | 0.0017           | 0.0019

ckamm:
Note on the state of my branch at https://github.com/ckamm/owncloud-client/tree/exclude_traversal_qregexp: It does some very hacky things and has outstanding issues. It was only developed far enough for performance testing.

* The way the csync_exclude test is `#include`d into testexcludedfiles with preprocessor tricks isn't pretty and should be reworked.
* I think there is an outstanding issue with / at the end of a pattern matching directories only (maybe related to * at the end of a pattern? I don't remember, but there was some issue there...)
* More test patterns and general cleanup everywhere.

guruz:
BTW Qt 5.9 mentions

> QRegularExpression now requires the PCRE2 library, at least version 10.20. Support for the PCRE1 library has been dropped. A copy of PCRE2 is shipped with Qt and will automatically be used on those platforms which lack it.

guruz:
I'm trying to work on this again..

guruz:
Mergeable version in https://github.com/owncloud/client/pull/6063
There is a lot of extra things done by @ckamm in his branch in this PR, so keeping this open so we know how to improve upon this
(e.g. using regexp also for _no_ctx)

ckamm:
@guruz I don't think we need to keep this open. I doubt we can salvage much of the experimental code in my branch.

Issue #4996, opened 2016-06-22
[FedShare] Can share permission are not granted for folders by default in case of FedShare


mcastroSG:
### Expected behaviour
"Can Share" permission should be granted by default in case of fedshare for folders

### Actual behaviour
To grant "Can Share" permission, we have to uncheck and check "can edit" checkbox. 

### Steps to reproduce
1. As user A connected to server A right click over folder1 on Shared sync folder
2. Choose "Share with owncloud" option
3. Type userb@serverb and wait till operation ends
4. As user B connected to server B right click over folder1 that has been previously synced from server
5. Choose "Share with owncloud" option

A new message saying that you are not able to reshare is shown

### Server configuration

Server OS: Ubuntu 14.04
Server Version: 9.1.0 Beta

### Client configuration

Client OS: OSX El capitán 10.11.5
Client Version: Version 2.2.2 (build 3472)

ckamm:
My current analysis is: due to https://github.com/owncloud/core/issues/22122#issuecomment-185637344 we explicitly removed reshare permissions in e4ec09dd874bd84e913d7e22c801b691537d4297. But then https://github.com/owncloud/core/pull/23918 happened and now we might need to send reshare permissions again. Needs more testing.

ckamm:
I've tested this with 9.1.0beta1 and the server does indeed create federated shares that /include/ the reshare permission. @rullzer Do we need to preserve the 9.0 behavior of not creating federated shares with reshare permissions?

SergioBertolinSG:
cc @PVince81

PVince81:
From what I remember, in OC < 9.0 we didn't want to introduce a change of behavior, so there the permissions were "automatic" and not really CRUDS.

Then in 9.1 with https://github.com/owncloud/core/pull/23918 we introduced CRUDS for federated shares. When you share in the web UI with a user user1@localhost/owncloud, you have the ability to set CRUDS permissions, including reshare. This means the permissions are explicit.

The clients should align on that for 9.1.

PVince81:
Okay, I'm starting to remember... it's a bit messy.

From my understanding the clients would need to do as follows:
- if OC < 9.1: send 1 for readonly permissions, or 7 for readwrite (others will be rejected by the API)
- if OC >= 9.1: send more granular permissions as needed: readonly 1, readwrite 15, readonly + share 17, readwrite + share 31. Not sure what options the GUI shows ?

[...]

PR #5001, opened 2016-06-24
Share: Permissions for federated shares on servers >=9.1 #4996

ckamm:
This disables the workaround 487e1fdca5ee04fc98c1ed77898df70d740967c8
for servers that are new enough to support fine grained permissions
on federated shares.

The consequence is that the 'reshare' permission is now granted by
default and that users can edit permissions on the usual fine-grained
level again.

The way the client deals with servers <9.1 is unchanged.


guruz:
👍

PR #5008, opened 2016-06-27
Fix a QAction leak

jturcotte:
Attach the accout menu QAction to the menu rather than the
ownCloudGui since it should be destroyed when the context
menu is cleared.


jturcotte:
The leak is tiny, like a few bytes on each account status update. But it's the only leak I could find by doing successive syncs (empty or not) using one account and one folder.

guruz:
👍

PR #5016, opened 2016-06-28
AccountManager: Don't force the server URL while loading the config.

ogoffart:
While loading the account, only override the server url if Theme::forceConfigAuthType
is set. This restore the behavior from the client 2.2.0 for theme that did not
use Theme::forceConfigAuthType.

### DIFF
--- a/src/gui/accountmanager.cpp
+++ b/src/gui/accountmanager.cpp
@@ -213,14 +213,12 @@ AccountPtr AccountManager::loadAccountHelper(QSettings& settings)
 
     QString authType = settings.value(QLatin1String(authTypeC)).toString();
     QString overrideUrl = Theme::instance()->overrideServerUrl();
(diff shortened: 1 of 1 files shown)


ogoffart:
This is for issue https://github.com/owncloud/enterprise/issues/1418
It does not follow redirect, but it does not "undo" them anymore.

Should this be pushed to the 2.2 branch or not?

felixboehm:
It is mandatory for owncloud/enterprise#1418 to follow a redirect, provided on the preconfigured url.

ogoffart:
The client 2.1 follows redirects in the wizard for the first connection, then save the redirected URL in the config
The client 2.2 however is not loading the redirected URL from the config. Making this bug appear.

Following redirect on every request is maybe something we could do, but not as part of 2.2.3

felixboehm:
tested, works.

PR #5034, opened 2016-07-04
Only accept notification API Capability if endpoint is OCS-enabled

### LINKED ISSUE #5033: Properly detect notifications app OCS capability
```
<PVince81> danimo: I saw that there is a slight capability difference between 8.2 and 9.0
<PVince81> so depending on what exactly they check
<PVince81> it was observed by cdamken
<dragotin> later, I have to run
<dragotin> sorry
<danimo> if( ! ptr->account()->capabilities().notificationsAvailable())
<PVince81> danimo: might need to go deeper:
<danimo> which is implemented as: "return _capabilities.contains("notifications");"
<PVince81> this is in 9.0+ https://github.com/owncloud/notifications/blob/master/lib/Capabilities.php#L41
<PVince81> and this is in 8.2.6: https://github.com/owncloud/notifications/blob/v8.2.6/lib/capabilities.php#L41
<PVince81> not sure why it existed in 8.2.6... but might be possible to check the "ocs-endpoint" vs "endpoint" string
<PVince81> so if you add ".contains("ocs-endpoint")" it should work correctly
```

Basically, you need to check for "ocs-endpoint" inside the "notifications" block in the capabilities.

Currently the client "wastes" HTTP requests by querying notifications for 8.2.x versions.

@dragotin @danimo @cdamken

danimo:
Fixes #5033

### DIFF
--- a/src/libsync/capabilities.cpp
+++ b/src/libsync/capabilities.cpp
@@ -73,7 +73,8 @@ bool Capabilities::shareResharing() const
 
 bool Capabilities::notificationsAvailable() const
 {
-    return _capabilities.contains("notifications");
+    // We require the OCS style API in 9.x, can't deal with the REST one only found in 8.2
+    return _capabilities.contains("notifications") && _capabilities["notifications"].toMap().contains("ocs-endpoints");
 }
 
 bool Capabilities::isValid() const


PVince81:
- [x] TEST: doesn't request notifications for OC 8.2.6
- [x] TEST: requests notifications for stable9

:+1: looks good

Issue #3764, opened 2015-09-05
Allow multiple accounts to sync the same local folder


icewind1991:
Currently when using the multi-account feature it's not possible to sync the same folder to multiple servers.

It would be nice to have the options to sync a folder to multiple servers.

guruz:
@icewind1991 You mean sync the same local folder? The problem is we use exactly one ```.csync_journal.db``` per local folder and server..

icewind1991:
> You mean sync the same local folder

Yes

> The problem is we use exactly one .csync_journal.db per local folder and server..

Yes, you would have to change the csync.db filename to be based on the server

RalfWW:
I run this scenario like this under Win7:

folder-owncloud1
  Symlink-to_folder_ABC >---|
                                             |----> folder_ABC
folder-owncloud2                  |
  Symlink-to_folder_ABC >---|

Having two local folders synced to a dedicated owncloud server (owncloud1, owncloud2) each.
Adding a symbolicLink to the "shared" folder (folder_ABC).

The only drawback: changes within folder_ABC will not be detected by the client "in time". It can take a long time (up to one hour, I guess) until the client performs a "check over all" run to get notice about the changes.
Stop and start syncing at the client will start a "check over all" manually in case you are in a hurry :-)

regards

gebhard73:
I've just created the workaround RalfWW mentioned - hopefully it will work :-)

Nevertheless I'd like to "vote" for this feature, too.

cheers,
Gebhard

RalfWW:
The actual clients don't support symlinks anymore.
The upcoming release 2.1.1 might perform it again!?

I'm staying with Version 2.0.1, the last that worked with symlinks (directories only!)

Ralf

gebhard73:
Ah - thanks for the feedback, I'll have a look into it next weekend latest.
Have you tried junctions?

Cheers,
Gebhard

gebhard73:
Any information if 2.1.1. works with this workaround?

Thanks!

RalfWW:
It seems that Version 2.2.2 is back with the old functionality to handle symlinked directories.

gebhard73:
Thanks, seems indeed to work again ... :-)

guruz:
@ckamm @dragotin We seem to have a conflict here with regards to https://github.com/owncloud/client/pull/5036
@gebhard73 and @RalfWW want a behaviour that we're disallowing (again).

gebhard73:
[...]

PR #5045, opened 2016-07-10
Dbjournal per account

dragotin:
This patch changes the name of the sync journal. It adds  a component that is built from the remote account. That is supposed to fix the root cause of #3764 and makes it possible to sync a local folder to multiple servers.

Note that this is **experimental** and still lacks these things:
- [x] Fix of the tests
- [x] Maybe more tests
- [x] Manual testing
- [x] An update strategy
- [x] Some code in owncloudcmd and FolderMan, see FIXMEs

Comments are welcome @ogoffart @danimo @ckamm


guruz:
@danimo @DeepDiver1975 Where did @mention-bot go?

ckamm:
@dragotin Looks good to me so far! The exclude pattern for the journal file also needs updating. There are a couple of other places that also have the "csync_journal" string (like folderwatcher_linux).

guruz:
How will this work when the client picks a new sync directory for an unrelated server?
E.g. I add an account to test something else and oC by default picks ~/ownCloud and I accidently click OK.
Will it mess up everything?

dragotin:
Plus csync test fixes, and adoption to the new cmocka API.

dragotin:
@ogoffart @guruz @jturcotte what do you guys think?

src/gui/folderman.cpp
```diff
@@ -1203,37 +1202,50 @@ QString FolderMan::checkPathValidityForNewFolder(const QString& path, bool forNe
         }
         if( ! folderDir.endsWith(QLatin1Char('/')) ) folderDir.append(QLatin1Char('/'));
 
-        if (QDir::cleanPath(f->path()) == QDir::cleanPath(userInput)
-                && QDir::cleanPath(QDir(f->path()).canonicalPath()) == QDir(userInput).canonicalPath()) {
```
ogoffart: I suppose there was an error in the old code.  the `&&` should have been a `||` 


Anyway, this shows a flaw in the code that we should check for cannonical path everywhere, really.
Also we should use Case insensitive comparison depending on the file system.

src/libsync/syncengine.cpp
```diff
@@ -91,7 +91,12 @@ SyncEngine::SyncEngine(AccountPtr account, const QString& localPath,
     url_string = Utility::toCSyncScheme(url_string);
 
     csync_create(&_csync_ctx, localPath.toUtf8().data(), url_string.toUtf8().data());
-    csync_init(_csync_ctx);
+
+    const QString dbFile = this->journalDbFilePath();
+    _journal->setDatabaseFilePath( dbFile );
```
ogoffart: Why should the sync engine set this ?   The journal is created already and might have been opened before.
IMHO, the creator of the journal should set this (i.e: the folder)

test/testfolderman.cpp
```diff
@@ -69,11 +89,21 @@ private slots:
         // A file -> Error
         QVERIFY(!folderman->checkPathValidityForNewFolder(dir.path() + "/sub/file.txt").isNull());
 
-        // There are folders configured in those folders: -> ERROR
-        QVERIFY(!folderman->checkPathValidityForNewFolder(dir.path() + "/sub/ownCloud1").isNull());
-        QVERIFY(!folderman->checkPathValidityForNewFolder(dir.path() + "/ownCloud2/").isNull());
-        QVERIFY(!folderman->checkPathValidityForNewFolder(dir.path() + "/sub").isNull());
```
ogoffart: Why did you remove this test and the next one?

src/libsync/syncengine.cpp
dragotin: pushed a fix, path is now built in the syncjournaldb class, getting the pathes from the folder.

src/gui/folderman.cpp
dragotin: I pushed two commits doing that.

test/testfolderman.cpp
dragotin: because they are superseeded by the tests that consider the url parameter, as it is now allowed to have the local dir used twice, as long as the urls are different.

src/gui/folderman.cpp
```diff
@@ -1194,46 +1193,64 @@ QString FolderMan::checkPathValidityForNewFolder(const QString& path, bool forNe
...
-        if (QDir::cleanPath(QString(userInput)).startsWith( QDir::cleanPath(folderDir)+'/')) {
+        if ( (forNewDirectory || differentPathes) && userDirClean.startsWith( folderDirClean, cs )) {
             return tr("The local folder %1 is already contained in a folder used in a folder sync connection. "
                       "Please pick another one!")
-                .arg(QDir::toNativeSeparators(userInput));
+                    .arg(QDir::toNativeSeparators(path));
         }
 
-        if (absCleanUserFolder.startsWith( QDir::cleanPath(folderDir)+'/')) {
+        bool cleanUserEqualsCleanFolder = Utility::fileNamesEqual(absCleanUserFolder, folderDirClean );
```
dragotin: `differentPathes`: comparison if the folder dir with followed symlinks is the same as the users dir without followed symlinks
`cleanUserEqualsCleanFolder`: both compared pathes follow symlinks

So there is a difference

src/gui/folderman.cpp
```diff
@@ -1194,46 +1193,64 @@ QString FolderMan::checkPathValidityForNewFolder(const QString& path, bool forNe
...
+        if (differentPathes && folderDirClean.startsWith(absCleanUserFolder, cs) &&
+                !cleanUserEqualsCleanFolder && !forNewDirectory ) {
+            return tr("The local folder %1 contains a symbolic link. "
+                      "The link target contains an already synced folder "
+                      "Please pick another one!")
+                    .arg(QDir::toNativeSeparators(path));
+        }
+
+
+        if( serverUrl.isValid() && Utility::fileNamesEqual(absCleanUserFolder,folderDir ) ) {
```
dragotin: `FolderDir` is canonicaled and got the trailing slash above. So the user dir is equal to the folder dir. That was forbidden before and still is forbidden if the account is the same. That is what this if() checks for. I added a comment.

On a general note, I think this function is terrible. I wonder if it could be done differently, the key thing is the checking for symlinks. However there is a unit test for this function which allows test driven development (you define the behaviour through the test) and that is how I worked on it.

ckamm:
@dragotin This looks good to me. (I agree about that checking function being terrible though!)

Does anything else need to happen before this can be merged?

guruz:
So you can't easily switch back to old client version which has the previous naming scheme? but maybe that's ok.

ckamm:
@guruz That's an important point! 

What happens when people switch back and forth? You'd have both dbs coexisting, which would lead to bad behavior. Maybe the new client needs to keep using the old db name for old folders and only use the new db for new folders while storing the config in a way that old clients don't see it?

ogoffart:
Yes, we should keep using the old journal if it exists. When adding new folder, we an then just add a new db.

guruz:
Assigning @ckamm ...
@dragotin if you still want to do this please speak up before @ckamm or we do it..:)

ckamm:
@ogoffart I think that might not be sufficent. When you create a new folder in the new client and then go back to an old client version, a new journal would be created. If you then switched back, you'd be in the two-journal situation again.

I'll think about whether there's a way to do this safely. Maybe the first journal on a folder could still be called csync_journal and only the second gets another name? That way we'd only need to worry about what happens when a config that uses the same folder on two different accounts gets used by an older client. (which might very well be trouble and we probably need to make sure older clients don't see these)

jturcotte:
@ckamm Could we use a table prefix in the same sqlite DB?

ckamm:
@jturcotte I think you'd have similar interoperability issues with old clients that way.

Here's a path that would work:
* New folders that have the same root as old folders get added to a new section of the config, to make sure they are invisible to older clients. That's required, because if an old client attempted to sync the same folder to two accounts using the same db, we'd get bad behavior.
* Only these new folders use the new database file names.

That's ugly but would work, preserving most interoperability.

We could also do the following steps to make our lives easier in the future:
* Every Folder could check for a syncjournaldb at the new path and use it if found. That way we can easily switch to the new names once compatibility with <2.3.0 is no longer required.
* We already have Accounts/version. Add Accounts/N/version and Accounts/N/folders/M/version. If we encounter a version that's bigger than ours, skip that account/folder, but preserve it in the settings. Maybe show a dialog, warning the user about this.

@dragotin @jturcotte @guruz @ogoffart Thoughts?

jturcotte:
@ckamm How would it work to simply remove the old journal once you've "ported" to the new name? This means that the old client would redo a full-sync, not that bad. And if you ever see an old DB with the new client, you always ignore the new one and "re-port" from the old name and delete it again.

This would be a big problem if somebody runs both at the same time, but this is less likely and not something we've ever supported, so maybe that's OK.

ckamm:
@jturcotte Yes, I agree that that should work too. (always in addition to making sure old clients never have the same folder bound to multiple accounts; that would also allow us to tell which of the two folders should take over the old database)

dragotin:
Moving the old csync_journal.db to the new name is exactly how it is implemented in my patch. Yes, going back to an old version would completely ignore the new dbs and do a full sync. Maybe, however, we should check in the old code again if the new fancy names follow a pattern that is on the ignore list of old clients.

guruz:
Another issue came up when discussing last week: What about sysadmins in big deployments that do pre-injection of data into the .csync_journal.db? For example the selective sync list?
But maybe not a problem, now they have to just use the new filename.

ckamm:
Okay, I'll update this patch to:
* Store double-booked folder configs in a new section of the config file.
* always copy the old csync_journal.db to sync_ABC.db, even if sync_ABC.db exists already
* check whether the new db names will be ignored by old clients

ckamm:
Also, it seems that in my tests the big-folder choices got lost and they were displayed as new. Will have to investigate.

ckamm:
There's also currently a problem where having two different accounts on the same server will end up with the same sync journal db name, because the credentials are not part of the hash.

ckamm:
@dragotin What's the rationale for deriving the DB name from the url? The main advantage I see over just using a random value is that the same db might be reused even if the settings are lost. And that seems relatively small. If we used a random value we could completely sidestep the question of whether the user is available at that time already.

ckamm:
One complication is owncloudcmd: If we determine the db filename automatically from url, path and username, we can avoid having to pass the db name explicitly.

ckamm:
Meh, merging in master to resolve all the conflicts seems to have broken this for github. The main thing that still needs doing (besides more thorough tests!) is to update the manual tests: they still use .csync_journal.db explicitly in a couple of places.

dragotin:
@ckamm I was not thinking into the direction of using a random name and storing in the settings. Might be a good idea, I am not sure...

ckamm:
@dragotin I've gone with the hash-based approach, because that allows owncloudcmd to determine the right database from the host, folder, username arguments. Otherwise people would have had to pass the db path explicitly to it and we'd have broken all existing uses of owncloudcmd...

ckamm:
@ogoffart This is ready for a final round of review. All looks good from my point of view now. I'll spend time on manual tests now and see if I can make anything fail.

src/gui/folderman.cpp
```diff
@@ -1231,46 +1271,71 @@ QString FolderMan::checkPathValidityForNewFolder(const QString& path, bool forNe
...
-            return tr("The local folder %1 is already used in a folder sync connection. "
-                      "Please pick another one!")
-                .arg(QDir::toNativeSeparators(userInput));
-        }
-        if (!forNewDirectory && QDir::cleanPath(folderDir).startsWith(QDir::cleanPath(userInput)+'/')) {
+        const QString folderDirClean = QDir::cleanPath(folderDir)+'/';
+        const QString userDirClean = QDir::cleanPath(path)+'/';
+
+        // folderDir follows sym links, path not.
+        bool differentPathes = !Utility::fileNamesEqual(QDir::cleanPath(folderDir), QDir::cleanPath(path));
```
ogoffart: should we not canonicalize?

src/gui/folderwizard.h
```diff
@@ -49,7 +49,7 @@ class FolderWizardLocalPath : public FormatWarningsWizardPage
 {
     Q_OBJECT
 public:
-    FolderWizardLocalPath();
+    FolderWizardLocalPath(AccountPtr account);
```
ogoffart: const &

ogoffart:
Actually, -1,  because:

```
/var/lib/jenkins/workspace/cloud-client_client_PR-5045-XH7FKQPHY3F2PM4KGS2TZBW2XEBGFRDAIIQA4PWGEYSSPLSHKLIA/test/testutility.cpp:163:50: error: ‘qunsetenv’ was not declared in this scope
         qunsetenv("OWNCLOUD_TEST_CASE_PRESERVING");
                                                  ^
```

ckamm:
Bug:
* [ ] Sync journals and the wal and shm files appear in the not-synced tab

src/gui/folderman.cpp
ckamm: Utility::fileNamesEqual canonicalizes both arguments. Which makes me wonder why we call cleanPath. Should be no problem to drop that, right?

ckamm:
@ogoffart Comments addressed.

ogoffart:
:+1: :

Issue #2325, opened 2014-10-17
Don't delete files when a backup on the server is restored


jospoortvliet:
A simple and easy way of creating and restoring backups is using virtual machines or file system snapshots. Unfortunately, this is very likely to result in the deletion of all files on the client side which have been changed since the server snapshot was created!

It would be great to deal with that a little more gracefully.

### What happens is this (from Olivier on IRC):
Server snapshot is restored. The client sees that the file are not on the server.  And because the client knows that the file were there before from its database, it will remove the files localy as if they were deleted on the server.

Also it will download the files from the server erasing the ones in the client because what is on the server is considered new. The client has no way to know that there was a rollback.  It just thinks the files were deleted or updated.

### Dealing with it as end user/admin
You can prevent this from happening by deleting the local cache of the client before connecting to your ownCloud again. The client will then think all files locally are new, and after checking with the server, notice that they are newer than what is there and send them there, too (except for the ones with identical file size/name/change time).

### Fixing the problem
A potential direction for solving this would be to either look better at timestamps (not just == but look if time stamps are older or newer too).

Another idea might be to have counter incremented whenever the client talks to the server, server-side or synchronized between client and server. When the counter suddenly drops, the client knows the server isn't who it used to be ;-)

Obviously both ideas above aren't mine but Olivier's ;-)

tflidd:
There was a slightly different case on the forum: https://forum.owncloud.org/viewtopic.php?t=26498&p=79941

Is it possible to make the client more "intelligent"? In this case an external drive was unmounted (unknown reason), owncloud detects an empty folder, the client "thinks" files were deleted and deletes all files on the client. In this case it would be great if the client asked if it should as well delete all files from the client.

jospoortvliet:
@tflidd I guess the client doesn't even know the files are on external storage ... so it just sees a folder has disappeared and will synchronize that change. To make it warn, the client would have to learn what is external storage from the server, or the server needs a generic way of telling the client "hey, these files are at the moment not here, but they might come back so keep em around a bit longer".

I can actually see why that might be needed some day - not all external storage is equally reliable and if you are checking the server every few minutes, see it appear-disappear-appear-disappear and start to remove-re-download-remove-re-download files, that's not particularly smart ;-)

tflidd:
It seems like you fixed something: https://forum.owncloud.org/viewtopic.php?f=29&t=29613#p93199

guruz:
@tflidd Thanks for notifying..
[...]

Issue #3742, opened 2015-09-03
unique storage identifier to the remote top level sync directory


moscicki:
Think of persistent identifiers in RAID or unique filesystem labels. It is possible to detect configuration mismatches because there is a peristent and universally unique piece of information stored together with the data.

Maybe some of common misconfiguration problems for individual owncloud users could be solved if a sync client could query the server for a storage endpoint identifier which would be universally unique. For example, if there was a persistent uuid stored on the server in the top level sync directory and available as webdav property then a client could detect a silent reconfiguration of a server (which normally results in an empty folder and all files deleted on the server). Such identifier could be generated automatically in a unique way first time user creates the account. For larger deployments, it could maybe help in scenarios when we have test and production environments and sometimes we reconfigure the test proxy to talk to different underlying storage. In this case the server implementation of this identifier could be different but the principle would be the same: it would allow the sync client to find out if it is still talking to the same storage endpoint on the server or there was some kind of reconfiguration. It could result in a better handling of such situations.

Thoughts?

guruz:
I label this as bug because I think this kind of checking is mandatory to do.

@moscicki I agree with you, we should save the root's ```FileId``` also in our ```.csync_journal.db``` and on mis-match throw an error.
The question is do we allow the user to then override the error?

@PVince81 @icewind1991 thoughts? Is ```FileId``` good to use?

Related https://github.com/owncloud/client/issues/2325

moscicki:
@guruz: thanks! I am thinking about the sync protocol: shouldn't we be able to query the "unique storage id" as a webdav property for any directory (also to account for the fact that someone may be adding a sync folder inside the folder tree on the server rather than syncing everything via / ). So imagine that in the current server implementation you have all files in the localfilesystem on the server. I would guess that "unique storage id" would be a UUID stored somewhere in the data directory of the user (next to user/files) and that value could be served by PROPFIND for all directories which sit on this local filesystem. In the sync client you may combine UUID:FileId of the directory in question to uniquely identify a directory (this assumes that FileId never repeats which may be a good assumption for now but may not be a good assumption in the future [e.g. if you wanted to use real inodes as FileId -- but maybe we should not worry about this for now).

IMO FIleId alone is not good enough because if you start with a fresh database then the FileId counts again from zero and there is a possibility that the value repeats with what you have stored earlier.

icewind1991:
> Is FileId good to use?

Yes

ogoffart:
[...]

PR #5056, opened 2016-07-13
SyncEngine: Reads the data-fingerprint property

ogoffart:
When it changes, assume a backup was recovered, and keep conflict files.

Issues: #2325 and https://github.com/owncloud/enterprise/issues/966

### DIFF
--- a/src/libsync/discoveryphase.cpp
+++ b/src/libsync/discoveryphase.cpp
@@ -210,7 +210,7 @@ int get_errno_from_http_errcode( int err, const QString & reason ) {
 
 
 DiscoverySingleDirectoryJob::DiscoverySingleDirectoryJob(const AccountPtr &account, const QString &path, QObject *parent)
-    : QObject(parent), _subPath(path), _account(account), _ignoredFirst(false)
+    : QObject(parent), _subPath(path), _account(account), _ignoredFirst(false), _isRootPath(false)
 {
 }
 
@@ -218,10 +218,15 @@ void DiscoverySingleDirectoryJob::start()
 {
     // Start the actual HTTP job
     LsColJob *lsColJob = new LsColJob(_account, _subPath, this);
-    lsColJob->setProperties(QList<QByteArray>() << "resourcetype" << "getlastmodified"
-                        << "getcontentlength" << "getetag" << "http://owncloud.org/ns:id"
-                        << "http://owncloud.org/ns:downloadURL" << "http://owncloud.org/ns:dDC"
-                        << "http://owncloud.org/ns:permissions");
+
+    QList<QByteArray> props;
+    props << "resourcetype" << "getlastmodified" << "getcontentlength" << "getetag"
+          << "http://owncloud.org/ns:id" << "http://owncloud.org/ns:downloadURL"
+          << "http://owncloud.org/ns:dDC" << "http://owncloud.org/ns:permissions";
+    if (_isRootPath)
+        props << "http://owncloud.org/ns:data-fingerprint";
+
+    lsColJob->setProperties(props);
 
     QObject::connect(lsColJob, SIGNAL(directoryListingIterated(QString,QMap<QString,QString>)),
                      this, SLOT(directoryListingIteratedSlot(QString,QMap<QString,QString>)));
@@ -299,12 +304,14 @@ void DiscoverySingleDirectoryJob::directoryListingIteratedSlot(QString file, con
 {
     //qDebug() << Q_FUNC_INFO << _subPath << file << map.count() << map.keys() << _account->davPath() << _lsColJob->reply()->request().url().path();
     if (!_ignoredFirst) {
-        // First result is the directory itself. Maybe should have a better check for that? FIXME
+        // The first entry is for the folder itself, we should process it differently.
         _ignoredFirst = true;
         if (map.contains("permissions")) {
             emit firstDirectoryPermissions(map.value("permissions"));
         }
-
+        if (map.contains("data-fingerprint")) {
+            _dataFingerprint = map.value("data-fingerprint").toUtf8();
+        }
     } else {
         // Remove <webDAV-Url>/folder/ from <webDAV-Url>/folder/subfile.txt
         file.remove(0, _lsColJob->reply()->request().url().path().length());
@@ -426,6 +433,11 @@ void DiscoveryMainThread::doOpendirSlot(const QString &subPath, DiscoveryDirecto
                      this, SIGNAL(etagConcatenation(QString)));
     QObject::connect(_singleDirJob, SIGNAL(etag(QString)),
                      this, SIGNAL(etag(QString)));
+
+    if (!_firstFolderProcessed) {
+        _singleDirJob->setIsRootPath();
+    }
+
     _singleDirJob->start();
 }
 
@@ -441,7 +453,12 @@ void DiscoveryMainThread::singleDirectoryJobResultSlot(const QList<FileStatPoint
     _currentDiscoveryDirectoryResult->list = result;
     _currentDiscoveryDirectoryResult->code = 0;
     _currentDiscoveryDirectoryResult->listIndex = 0;
-     _currentDiscoveryDirectoryResult = 0; // the sync thread owns it now
+    _currentDiscoveryDirectoryResult = 0; // the sync thread owns it now
+
+    if (!_firstFolderProcessed) {
+        _firstFolderProcessed = true;
+        _dataFingerprint = _singleDirJob->_dataFingerprint;
+    }
 
     _discoveryJob->_vioMutex.lock();
     _discoveryJob->_vioWaitCondition.wakeAll();
--- a/src/libsync/syncengine.cpp
+++ b/src/libsync/syncengine.cpp
@@ -915,7 +915,16 @@ void SyncEngine::slotDiscoveryJobFinished(int discoveryResult)
             return;
         }
     }
-    if (!_hasForwardInTimeFiles && _backInTimeFiles >= 2) {
+
+    auto databaseFingerprint = _journal->dataFingerprint();
+    // If databaseFingerprint is null, this means that there was no information in the database
+    // (for example, upgrading from a previous version, or first sync)
+    // Note that an empty ("") fingerprint is valid and means it was empty on the server before.
+    if (!databaseFingerprint.isNull()
+            && _discoveryMainThread->_dataFingerprint != databaseFingerprint) {
+        qDebug() << "data fingerprint changed, assume restore from backup" << databaseFingerprint << _discoveryMainThread->_dataFingerprint;
+        restoreOldFiles();
+    } else if (!_hasForwardInTimeFiles && _backInTimeFiles >= 2) {
         qDebug() << "All the changes are bringing files in the past, asking the user";
         // this typically happen when a backup is restored on the server
         bool restore = false;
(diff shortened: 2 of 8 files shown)


ogoffart:
Smashbox test: https://github.com/owncloud/smashbox/pull/149

src/libsync/discoveryphase.cpp
```diff
@@ -299,12 +304,14 @@ void DiscoverySingleDirectoryJob::directoryListingIteratedSlot(QString file, con
...
     //qDebug() << Q_FUNC_INFO << _subPath << file << map.count() << map.keys() << _account->davPath() << _lsColJob->reply()->request().url().path();
     if (!_ignoredFirst) {
-        // First result is the directory itself. Maybe should have a better check for that? FIXME
+        // The first entry is for the folder itself, we should process it differently.
         _ignoredFirst = true;
         if (map.contains("permissions")) {
             emit firstDirectoryPermissions(map.value("permissions"));
         }
-
+        if (true || map.contains("data-fingerprint")) {
```
guruz: true || ?

guruz: If this is anyway server global according to @PVince81 we could always read it maybe?

guruz: Aren't we wasting bandwidth by requesting it all the time so it is returned with each file/directory in each PROPFIND in the discovery?

src/libsync/syncengine.cpp
```diff
@@ -912,7 +912,16 @@ void SyncEngine::slotDiscoveryJobFinished(int discoveryResult)
             return;
         }
     }
-    if (!_hasForwardInTimeFiles && _backInTimeFiles >= 2) {
+
+    auto databaseFingerprint = _journal->dataFingerprint();
+    // If databaseFingerprint is null, this means that there was no information in the database
+    // (for example, upgrading from a previous version, or first sync)
+    // If it is empty (""), then that meanst that it was empty on the server before.
```
guruz: Isn't empty also isNull? I'm not sure anymore since doing PHP :)
Maybe make it more clear by having an else if(isEmpty) ?

src/libsync/syncjournaldb.cpp
```diff
@@ -286,6 +286,13 @@ bool SyncJournalDb::checkConnect()
         return sqlFail("Create table version", createQuery);
     }
 
+    // create the checksumtype table.
+    createQuery.prepare("CREATE TABLE IF NOT EXISTS datafingerprint("
+                               "fingerprint TEXT UNIQUE"
+                               ");");
```
guruz: So it is really a per-oC global fingerprint?
And this will never change per storage or anything?

FYI @moscicki

src/libsync/discoveryphase.cpp
ogoffart: oops, leftover from debugging.  well spot.

But we don't request it all the time,  only for the first folder, see previous hunk (_isRootPath)

src/libsync/syncengine.cpp
ogoffart: no, empty is not isNull   hence the isNull   which is different from isEmpty because the isEmpty needs to be covered by the normal case.
My comment is just there so someone does not change isNull to isEmpty,  or at least understands why it is isNull and not isEmpty

src/libsync/syncjournaldb.cpp
guruz: I found the @moscicki discussion.. https://github.com/owncloud/client/issues/3742

src/libsync/ownsql.cpp
```diff
@@ -303,6 +303,10 @@ void SqlQuery::bindValue(int pos, const QVariant& value)
                 res = sqlite3_bind_null(_stmt, pos);
             }
             break; }
+        case QVariant::ByteArray: {
+            auto ba = value.toByteArray();
+            res = sqlite3_bind_text(_stmt, pos, ba.constData(), ba.size(), SQLITE_TRANSIENT);
```
guruz: Is this correct? Should it use sqlite3_bind_blob(64)?

src/libsync/syncengine.cpp
guruz: I don't understand yet. Can't we just ignore it if it is empty or null? I don't think we can properly distinguish between empty and null in the PROPFIND property response can we?
(yes i know there is 404 properties but still).
I don't understand the reason to not just check for notEmpty?

src/libsync/syncengine.cpp
```diff
@@ -912,7 +912,16 @@ void SyncEngine::slotDiscoveryJobFinished(int discoveryResult)
...
+    if (!databaseFingerprint.isNull()
+            && _discoveryMainThread->_dataFingerprint != databaseFingerprint) {
+        qDebug() << "data fingerprint changed, assume restore from backup" << databaseFingerprint << _discoveryMainThread->_dataFingerprint;
+        restoreOldFiles();
```
guruz: Why is it assumed and not asked like in the previous implementation?
https://github.com/owncloud/client/blob/4c17df6be4d6406f207614bf09897bb87d57046b/src/libsync/syncengine.cpp#L928

src/libsync/syncengine.cpp
```diff
@@ -1023,18 +1032,23 @@ void SyncEngine::slotItemCompleted(const SyncFileItem &item, const PropagatorJob
     emit itemCompleted(item, job);
 }
 
-void SyncEngine::slotFinished()
+void SyncEngine::slotFinished(bool success)
 {
     _anotherSyncNeeded = _anotherSyncNeeded || _propagator->_anotherSyncNeeded;
 
+    qWarning() << success << _discoveryMainThread->_dataFingerprint;
+    if (success) {
```
guruz: Why is it only done in success case?

src/libsync/syncengine.cpp
```diff
@@ -1023,18 +1032,23 @@ void SyncEngine::slotItemCompleted(const SyncFileItem &item, const PropagatorJob
...
+
     // emit the treewalk results.
     if( ! _journal->postSyncCleanup( _seenFiles, _temporarilyUnavailablePaths ) ) {
         qDebug() << "Cleaning of synced ";
     }
 
     _journal->commit("All Finished.", false);
     emit treeWalkResult(_syncedItems);
-    finalize(true); // FIXME: should it be true if there was errors?
+    finalize(success);
```
guruz: behaviour change. did you check the implications?

src/libsync/syncjournaldb.cpp
```diff
@@ -436,6 +443,14 @@ bool SyncJournalDb::checkConnect()
     _insertChecksumTypeQuery.reset(new SqlQuery(_db));
     _insertChecksumTypeQuery->prepare("INSERT OR IGNORE INTO checksumtype (name) VALUES (?1)");
 
+    _getDataFingerprintQuery.reset(new SqlQuery(_db));
+    _getDataFingerprintQuery->prepare("SELECT fingerprint FROM datafingerprint");
+
+    _setDataFingerprintQuery1.reset(new SqlQuery(_db));
+    _setDataFingerprintQuery1->prepare("DELETE FROM datafingerprint;");
```
guruz: Do you actually need this query? Why not ```INSERT OR REPLACE```?
If you need it: Why is it not named _deleteDataFingerprintQuery?

src/libsync/syncengine.cpp
ogoffart: Because in this case we know it's the case.
The other case is just heuristics.

So i feel it's better not to ask.

src/libsync/syncengine.cpp
ogoffart: If it fails (say, the connection was closed)  we want the next sync to continue the restoration.

I need to remove the qWarning.

src/libsync/syncengine.cpp
ogoffart: yes, this has to do when the white list is cleared.

src/libsync/syncjournaldb.cpp
ogoffart: because that's a table that can contain several entries, but i only want one.

src/libsync/syncengine.cpp
ogoffart: we need to ignore it if it's null, see the comment.

If it's empty, it is not ignored, it is considered as empty.  The server always return empty by default.  I'll try to improve the comment

src/libsync/discoveryphase.cpp
dragotin: I do not like the isRootPath parameter. It's cluttering of the interface.

I think it would be nicer if you would either set all props from the class that creates the job using `setProperties()`, or even better add a method `addProperty( a_property)` that is called in case it is a root object.

ogoffart: DiscoverySingleDirectoryJob is an internal class to the discovery job, it's not really a public class with an exported interface. Its role is to abstract the protocol part of the discovery.  In that sens, the name of the properties are part of the protocol and should be IMHO kept inside this class.

src/libsync/syncjournaldb.cpp
guruz: Why not make it contain only one by using ```INSERT OR REPLACE```?
Why is this not commented?

ogoffart: How would you write it with INSERT OR REPLACE?

src/libsync/syncengine.cpp
```diff
@@ -912,7 +912,16 @@ void SyncEngine::slotDiscoveryJobFinished(int discoveryResult)
             return;
         }
     }
-    if (!_hasForwardInTimeFiles && _backInTimeFiles >= 2) {
+
+    auto databaseFingerprint = _journal->dataFingerprint();
+    // If databaseFingerprint is null, this means that there was no information in the database
+    // (for example, upgrading from a previous version, or first sync)
+    // Note that an empty ("") fingerprint is valid and means it was empty on the server before.
```
guruz: i still don't understand why there is a difference between empty and null. Why can't they be treated the same?

src/libsync/syncengine.cpp
guruz: Can't it be that the client is the old replica and we'd want the files from the server?

src/libsync/syncengine.cpp
ogoffart: Null means there was no entry before.

Empty means there was one, and it was empty.
By default, on the server, it is empty.

src/libsync/ownsql.cpp
ogoffart: I actually don't know what's the difference here.  I guess it's maybe related to \0 in the string?  Then maybe indeed blob could be better (more generic)  but currently, in owncloud, we don't have any field of type blob,  only text.

ogoffart: And this is just an optimisation, maybe i could leave it out and then it would use the text16 fallback as toString which is just a bit of a waste.

src/libsync/syncengine.cpp
guruz: Ah. Weeeeird. But I guess this is because it was added later to the server

FYI @rullzer

src/libsync/ownsql.cpp
guruz: (I meant because of encoding issues.. but since the data-fingerprint is from WebDAV it is probably anyway safe ascii or so)

guruz:
OK 👍 
Needs good QA then. @rperezb @SergioBertolinSG @davitol 
Will @ogoffart add it to https://github.com/owncloud/client/wiki/Testing-Scenarios ?

ogoffart:
There is a smashbox test: owncloud/smashbox#149
But it does not replace manual QA.

rperezb:
@davitol this is already tested, isn't it?

davitol:
@rperezb No, it isn't. The scenarios tested were related to Fix detection of backup or full resync, but not to the data-fingerprint property.

Right now i'm catching up how all this stuff of data-fingerprint works in oC. ASAP it is tested, i'll  write it here.

davitol:
https://github.com/owncloud/client/issues/5185

Issue #5046, opened 2016-07-11
Progress Calculcations should have a maximum value to be displayed


DeepDiver1975:
### Expected behaviour
Having a progress indicator telling the user that is will take hundreds of thousands of year to up/download has zero value from my pov

see https://www.facebook.com/photo.php?fbid=976914802406122&set=o.366301826358&type=3&theater

![13592161_976914802406122_2971566735647043855_n](https://cloud.githubusercontent.com/assets/1005065/16721299/6ed715d2-473e-11e6-85d7-b30046150164.jpg)

dragotin:
whats the version? There were fixes for 2.2.2 IIRC

DeepDiver1975:
> whats the version? There were fixes for 2.2.2 IIRC

I just asked for the version information ....

DeepDiver1975:
2.2.2 according to the reporter on facebook

ckamm:
Pull request setting a 5kb/s lower limit on the bandwidth estimate: https://github.com/owncloud/client/pull/5066

ckamm:
@dragotin @guruz @ogoffart To continue the discussion from the pull request: If we wanted to hide the time estimate sometimes, what exactly would be a sensible criterion?
* "If the estimated time is longer than N days" -> What if users want to upload TBs of data and it genuinely takes that long?
* "If the bandwidth drops below a critical threshold" -> Unfortunately bandwidth often drops to shockingly small values when a lot of small files are uploaded or downloaded, or many files are moved or deleted.

The current code already tries to detect cases where the remaining time isn't dominated by bandwidth but by per-request overhead. But it doesn't seem to have caught the particular case that's being reported.

Suggestion: Hide the time estimate when the current estimate is way above (>100x) the optimistic estimate (which is calculated using the highest files-per-second and bandwidth values that were observed so far).

davitol:
@ckamm Which is supposed to be the maximum value that can be displayed?

ckamm:
@davitol There is no fixed limit. Time estimates are displayed as long as the estimate isn't larger than 100x the optimistic estimate (which uses the highest-ever seen files-per-second and bytes-per-second values).

This means you should be able to reproduce this, for example, when uploading a large file. Start out normal and then throttle your bandwidth severely (few bytes per second). After a while (a minute?) the current-bandwidth based time estimate should become huge and the time estimate should stop being displayed.

davitol:
[...]

PR #5066, opened 2016-07-19
Progress: Minimum bandwidth to avoid silly estimates #5046

ckamm:


ogoffart:
It would still be quite a silly estimate.   Maybe better not to show anything in that case.

dragotin:
I agree, maybe it is better to show nothing or something like *stalled* or so if the rate is so slow.

guruz:
"stalled" is wrong, "stalled" implies no bytes.
I think @ckamm 's goal is to smoothen estimations.
Showing nothing might blink the UI?

ckamm:
I agree, not showing a time estimate in these cases would be neater.

ckamm:
New patchset, please check it out.

src/gui/folderstatusmodel.cpp
```diff
@@ -934,8 +934,12 @@ void FolderStatusModel::slotSetProgress(const ProgressInfo &progress)
...
+
         //: Example text: "5 minutes left, 12 MB of 345 MB, file 6 of 7"
-        overallSyncString = tr("%5 left, %1 of %2, file %3 of %4")
+        QString msg = progress.trustEta()
+                ? tr("%5 left, %1 of %2, file %3 of %4")
+                : tr("%1 of %2, file %3 of %4");
+        overallSyncString = msg
             .arg(s1, s2)
             .arg(currentFile).arg(totalFileCount)
             .arg( Utility::durationToDescriptiveString1(progress.totalProgress().estimatedEta) );
```
ogoffart: I think arg() will show a warning if there is not a corresponding '%',  doesn't it?

src/gui/folderstatusmodel.cpp
```diff
@@ -934,8 +934,12 @@ void FolderStatusModel::slotSetProgress(const ProgressInfo &progress)
     if (totalSize > 0) {
         QString s1 = Utility::octetsToString( completedSize );
         QString s2 = Utility::octetsToString( totalSize );
+
         //: Example text: "5 minutes left, 12 MB of 345 MB, file 6 of 7"
```
ogoffart: doesn't the translator comment need to be exactly one line above the tr(),  in any case, since there are two tr now, that might not be attached to the second tr which would still need a comment

src/libsync/progressdispatcher.cpp
```diff
@@ -282,9 +282,6 @@ ProgressInfo::Estimates ProgressInfo::totalProgress() const
     // This assumes files and transfers finish as quickly as possible
```
ogoffart: move this comment too

ckamm:
@ogoffart I've explicitly unpacked these tr() calls, please double-check.

ogoffart:
:+1:

Issue #4990, opened 2016-06-21
Cinnamon applet using 90% CPU


mlda065:
I've recently noticed that my computer has been quite slow recently. My System Monitor often shows CPU usage up near 90% when I'm doing nothing but writing emails and reading pdfs.
I initially thought this was a Cinnamon issue ([comment thread in cinnamon repo](https://github.com/linuxmint/Cinnamon/issues/1154#issuecomment-227053785)), since the System Monitor attributes the high CPU usage to cinnamon. However now I'm pretty certain it's Owncloud. Specifically I think it's an issue with the panel applet, because that would explain why it's being counted towards Cinnamon's CPU usage and not owncloud.

### Symptoms
 * My CPU is often running at 90% (as shown with `top` and the *System Monitor* GUI) when I'm not anything intensive (reading pdfs, browsing facebook, etc). Both tools attribute the high CPU usage to cinnamon.
* I cannot watch videos or animated GIFs without visual stuttering
* when I type, what I see on the screen is often several characters behind what I've pressed
* Scrolling through pdfs is not smooth
* switching between windows is laggy
* etc

### Why I blame OC
I realise that 'my computer is slow' is a really vague claim, which is hard to address.

I tried the following things to figure out the cause:
 * Removing cinnamon panel widgets
 * Resetting all Cinnamon settings to default
 * Using one monitor instead of two
 * Not using a VPN
 * Quitting background programs (dropbox, syncthing, bitmeter OS, KDE Connect, etc)

The only thing which made the lag go away was quitting the owncloud client.
I even did a clean install. It didn't lag when the system was clean. Then I installed owncloud and it got laggy again.

### Other Info

 * I have 15 different folders synced through Owncloud, which might be an unusually high number.
 * This issue only started a few weeks ago
 * My laptop is fairly high end, so client hardware performance is not an issue.

### Server configuration
Operating system: Debian 8.5 Jessie

Web server: Apache 2

Database: MySQL

PHP version: PHP 5.6.20-0+deb8u1

ownCloud version: 9.0

Storage backend: USB Hard drive

### Client configuration
Client version: 2.2.1

Operating system: Mint 17.3, Cinnamon

OS language: English (AU)

Installation path of client: /usr/bin/owncloud

### Logs

Please use Gist (https://gist.github.com/) or a similar code paster for longer
logs.

[...]

Issue #4985, opened 2016-06-20
unity-panel-service crashes because of owncloud


nazar-pc:
### Expected behaviour
ownCloud icon displayed constantly

### Actual behaviour
ownCloud icon causes all icons to blinks few times per minute (like slow re-rendering of all icons) and eventually unity-panel-service crashes multiple times per day (once it didn't crash for a long time and its memory consumption grown to ridiculous 2.4 GiB). As a side effect, when I click on icon and wait few seconds, menu will be closed on menu re-rendering, so it is not always easy to even open settings.

### Steps to reproduce
1. Install ownCloud client nightly 2.1.1+
2. Open it
3. Observe tray icons

80 seconds video shows how icons blink, how memory consumption of `unity-panel-service` grows and it crashes eventually: https://owncloud.mokrynskyi.com/index.php/s/GWDpN3lexMgSOUG

### Client configuration
Client version: 2.2.1~nightly2016617-1.1

Operating system: Ubuntu 16.04 x64 (Unity)

OS language: Ukrainian

### Logs

Output of `owncloud --logfile log.txt` during blinking:
```
06-20 04:44:43:051 0xad87b0 csync_reconcile: Reconciliation for remote replica took 1,03 seconds visiting 329 files.
06-20 04:44:43:052 0xad87b0 csync_statedb_close: sqlite3_close=0
06-20 04:44:43:053 9675 OCC::SyncEngine::slotDiscoveryJobFinished: <<#### Reconcile end ####################################################  7070
06-20 04:44:43:054 9675 OCC::SyncEngine::slotDiscoveryJobFinished: Permissions of the root folder:  "RDNVCK"
06-20 04:44:43:055 9675 OCC::SyncJournalDb::commitInternal: void OCC::SyncJournalDb::commitInternal(const QString&, bool) Transaction commit  "post treewalk" and starting new transaction
06-20 04:44:43:057 9675 OCC::SyncJournalDb::commitInternal: void OCC::SyncJournalDb::commitInternal(const QString&, bool) Transaction commit  "post stale entry removal" and starting new transaction
06-20 04:44:43:058 9675 OCC::OwncloudPropagator::start: Using QNAM/HTTP parallel code path
06-20 04:44:43:059 9675 OCC::SyncEngine::slotDiscoveryJobFinished: <<#### Post-Reconcile end ####################################################  7076
[...]

PR #5072, opened 2016-07-21
Tray menu: Update only on demand #4990 #4985

ckamm:
The tray menu is now only updated when it becomes visible or while
it is visible.

### DIFF
--- a/src/gui/owncloudgui.cpp
+++ b/src/gui/owncloudgui.cpp
@@ -92,9 +92,9 @@ ownCloudGui::ownCloudGui(Application *parent) :
              this,SLOT(slotSyncStateChange(Folder*)));
 
     connect( AccountManager::instance(), SIGNAL(accountAdded(AccountState*)),
-             SLOT(setupContextMenu()));
+             SLOT(setupContextMenuIfVisible()));
     connect( AccountManager::instance(), SIGNAL(accountRemoved(AccountState*)),
-             SLOT(setupContextMenu()));
+             SLOT(setupContextMenuIfVisible()));
 
     connect( Logger::instance(), SIGNAL(guiLog(QString,QString)),
              SLOT(slotShowTrayMessage(QString,QString)));
@@ -193,7 +193,7 @@ void ownCloudGui::slotTrayClicked( QSystemTrayIcon::ActivationReason reason )
 void ownCloudGui::slotSyncStateChange( Folder* folder )
 {
     slotComputeOverallSyncStatus();
-    setupContextMenu();
+    setupContextMenuIfVisible();
 
     if( !folder ) {
         return; // Valid, just a general GUI redraw was needed.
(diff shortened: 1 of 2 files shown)


src/gui/owncloudgui.cpp
```diff
@@ -699,8 +709,6 @@ void ownCloudGui::slotUpdateProgress(const QString &folder, const ProgressInfo&
             _recentItemsActions.takeFirst()->deleteLater();
         }
         _recentItemsActions.append(action);
-
-        slotRebuildRecentMenus();
```
jturcotte: If this is removed, how is _recentActionsMenu now updated while the menu is shown during sync?

guruz:
(Wondering if this would influence #5074 for @RalfJung )

RalfJung:
Hm, I could try this... provided compiling and running owncloud from source somewhere in $HOME is not too difficult. ;)

Should I wait until this is merged (since it seems to have other problems still)?

src/gui/owncloudgui.cpp
ckamm: Yes this is a mistake, it should be called when the menu is visible!

danimo:
What's happening here?

ckamm:
@danimo Is your question about the contents of the patch or about "when will this be merged?"

Now that I'm back from vacation, I'll update this patch and then it can be merged.

ckamm:
@jturcotte Does it look good to you now?

src/gui/owncloudgui.cpp
```diff
@@ -552,6 +558,11 @@ void ownCloudGui::setupContextMenu()
     }
 }
 
+void ownCloudGui::setupContextMenuIfVisible()
+{
+    if (!_contextMenu || _contextMenu->isVisible())
```
jturcotte: Is it correct to enter the "if visible" code path when `!_contextMenu`?

ckamm: Hm, I added this with intent and it does work: `setupContextMenu()` builds the context menu if necessary. But one might just as well argue that setting up the context menu when it's not being shown yet is pointless. And `_contextMenu &&` would be less confusing to read. I'll update accordingly.

src/gui/owncloudgui.cpp
```diff
@@ -552,6 +558,11 @@ void ownCloudGui::setupContextMenu()
     }
 }
 
+void ownCloudGui::setupContextMenuIfVisible()
+{
+    if (_contextMenu && _contextMenu->isVisible())
```
mitya57: For platform tray icons, `_contextMenu->isVisible()` will always return `false`.

Though I am not saying that this change is wrong. I am going to cherry-pick it to the Ubuntu packaging anyway, because without it the context menu is updated every 30 seconds, and Qt [quickly runs out](https://bugs.launchpad.net/ubuntu/+source/unity/+bug/1635577/comments/31) of items IDs.

guruz: > will always return false.

isn't that a bug?

I remember there was some follow up work by me and@ckamm to set our own visible variable to true and false when showing or hidding menu. But maybe we factored that out again

ckamm: @mitya57 Note that there was significant follow up work done on context menu issues, at least 98efb075357fa64f424070e723db1786faa89d25, but possibly more.

PR #5080, opened 2016-07-25
ShibbolethView: Open a debug windows that shows cipher info.

dragotin:
It opens a window and connects to a cipher test
page, showing the output from there, that helps for debugging.

The window is enabled by setting the environment variable
OWNCLOUD_SHIBBOLETH_DEBUG

I wonder if we want that in 2.2 branch to be released with 2.2.3?

### DIFF
--- a/src/gui/creds/shibboleth/shibbolethwebview.cpp
+++ b/src/gui/creds/shibboleth/shibbolethwebview.cpp
@@ -20,6 +20,7 @@
 #include <QMessageBox>
 #include <QNetworkReply>
 #include <QSettings>
+#include <QMainWindow>
 
 #include "creds/shibboleth/shibbolethwebview.h"
 #include "creds/shibbolethcredentials.h"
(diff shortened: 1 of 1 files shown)


src/gui/creds/shibboleth/shibbolethwebview.cpp
```diff
@@ -74,10 +75,24 @@ ShibbolethWebView::ShibbolethWebView(AccountPtr account, QWidget* parent)
...
+
     page->mainFrame()->load(account->url());
     this->setPage(page);
     setWindowTitle(tr("%1 - Authenticate").arg(Theme::instance()->appNameGUI()));
 
+    // Debug view to display the cipher suite
+    if( !qgetenv("OWNCLOUD_SHIBBOLETH_DEBUG").isEmpty() ) {
+        // open an additional window to display some cipher debug info
+        QWebPage *debugPage = new UserAgentWebPage(this);
+        account->lendCookieJarTo(debugPage->networkAccessManager());
```
guruz: do we need this line if it's anyway a totally unrelated SSL check page?

src/gui/creds/shibboleth/shibbolethwebview.cpp
```diff
@@ -74,10 +75,24 @@ ShibbolethWebView::ShibbolethWebView(AccountPtr account, QWidget* parent)
...
+        debugPage->mainFrame()->load( QUrl("https://cc.dcsec.uni-hannover.de/"));
```
danimo: Why do we need an external web service for that? we can query for the supported cipher suites, see http://doc.qt.io/qt-4.8/qsslconfiguration.html#ciphers.

danimo:
This patch looks like its solving a very specific problem where a more general solution (`--supported-ciphers`, or some link in the about page) would be perferable. Also, let's not depend on 3rd party services.

dragotin:
Yes, it solves the problem that we want to check the ciphers **that Webkit that we use for Shibboleth** uses currently (and not very long any more) at users sites, where this has caused problems again and again. That is indeed a very specific problem that we solve.

That is why I think it is completely ok to not nerd into a more generic solution. The external web service is provided by an german university, that does not look like something that is evil or going away soon.

And again: Absolute edge case.

guruz:
I'm still fine with the commit except for the cookie jar line 👍

danimo:
It's not nerdy:

1. Distros will patch it out (no third party service policies)
2. The service may go away at any time
3. A good patch is not hard

dragotin:
I pushed two other commits that should make this PR more useful.

src/gui/creds/shibboleth/shibbolethwebview.cpp
```diff
@@ -61,6 +61,8 @@ ShibbolethWebView::ShibbolethWebView(AccountPtr account, QWidget* parent)
     setAttribute(Qt::WA_DeleteOnClose);
 
     QWebPage* page = new UserAgentWebPage(this);
+    page->setNetworkAccessManager( account->networkAccessManager() );
```
guruz: Sorry 👎 
This is a too critical change for a patch release.
As we re-create the QNAM in some other code path I don't know how crashy this could be

dragotin: ok

dragotin:
Now I force-pushed a finally nice version.

guruz:
👍

PR #5082, opened 2016-07-26
Logs: Add the name of the platform we're running on to the log.

dragotin:


src/libsync/utility.cpp
```diff
@@ -424,6 +424,21 @@ bool Utility::isBSD()
 #endif
 }
 
+QString Utility::platformName()
```
guruz: Is there no Qt command outputting the same?

dragotin: Thanks @guruz of course super Qt comes with something awesome. See additional patch.

guruz:
I can 👍 this if you also add the platform to https://github.com/owncloud/client/blob/28c12a3ca0769877e7b900c5bc4b1e9148dd5abf/src/libsync/syncengine.cpp#L730 :-)

On start of sync

dragotin:
@guruz good idea, I added stuff and now it looks like this:
```
07-28 14:45:29:966 15822 OCC::SyncEngine::startSync: "===== Using Qt 5.5.1 SSL library OpenSSL 1.0.1i-fips 6 Aug 2014 on openSUSE Leap 42.1 (x86_64)"
```

PR #5097, opened 2016-08-01
ownsql: handle QByteArray without converting to QString

ogoffart:
QByteArray is used for checksum

### DIFF
--- a/src/libsync/ownsql.cpp
+++ b/src/libsync/ownsql.cpp
@@ -303,6 +303,11 @@ void SqlQuery::bindValue(int pos, const QVariant& value)
                 res = sqlite3_bind_null(_stmt, pos);
             }
             break; }
+        case QVariant::ByteArray: {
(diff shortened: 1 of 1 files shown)


ogoffart:
( extracted from https://github.com/owncloud/client/pull/5056 )

dragotin:
:+1: 

Maybe we should remove the default case and error log so that we notice next time in case we introduce a new data type.

guruz:
Why not use sqlite3_bind_blob ?

ogoffart:
> Why not use sqlite3_bind_blob ?

I don't know? Why use it?  Currently all the QByteArray only contains text.

guruz:
OK 👍  if you take the risk that if we ever change that something bad happens.
Although the function has a size parameter so maybe it's OK.
(And we anyway always get text from WebDAV)

ogoffart:
Any other than a string was not working before because we were converting to QString anyway.

Issue #5092, opened 2016-07-29
client 2.2.3-nightly stop sync on large shared folder - 


PenIv:
### Expected behaviour
Client should sync all files

### Actual behaviour
Client start "Checking for Changes" in large folder and stop with error "An error occurred while opening a folder"

### Steps to reproduce
1. Upgrade to client 2.2.3 (build 6279)
2. Start client 
3. Wait for client to rescan large folder (7.75 GB with 11 325+ files)

### Server configuration
Operating system:
  Ubuntu 14.04
Web server: 
  Nginx/1.4.6
Database:
  MySQL
PHP version:
  5.5.9
ownCloud version:
  9.1.0

### Client configuration
Client version: 2.2.3 (build6279)

Operating system:
  Windows 10
OS language:
  English
Installation path of client:
  C:\Program Files (x86)\ownCloud
### Logs

1. Output of `owncloud --logwindow` or `owncloud --logfile log.txt`
[ocLogwindow.zip](https://github.com/owncloud/client/files/392333/ocLogwindow.zip)

2. Web server error log:
[owncloud.log.txt](https://github.com/owncloud/client/files/390502/owncloud.log.txt)

3. ownCloud log (data/owncloud.log):
...
|0|Attach|INST_SYNC|Down|1469102824|5790baea1823c|0|00386230ocaf51bd2515|4||0|0|1409387492|||INST_NONE|
|0|Attach/Oudtshoorn|INST_SYNC|Down|1469102824|5790baea07d75|0|00386268ocaf51bd2515|4||0|0|1467468787|||INST_NONE|
|0|Attach/Oudtshoorn/Photos|INST_SYNC|Down|1469102824|5790bae9ebe22|0|00388822ocaf51bd2515|4||0|0|1469028218|||INST_NONE|
12:12:15|7945|Attach/Oudtshoorn/Photos/10550_0 DSCN6019.JPG|INST_NEW|Down|1469102795|5156d4494e74949b7f9513b12e656d56|406135|00616484ocaf51bd2515|4||0|0|0|||INST_NONE|
12:12:24|4949|Attach/Oudtshoorn/Photos/10550_0 DSCN6020.JPG|INST_NEW|Down|1469102789|eb784ebf1ad1d9e55084d958fec49b87|252649|00616485ocaf51bd2515|4||0|0|0|||INST_NONE|
#=#=#=# Syncrun started  until  (0 msec)
#=#=#=# Syncrun started  until  (0 msec)
[...]

PR #5100, opened 2016-08-03
Revert "Fix a deadlock when shutting down during discovery"

jturcotte:
Reverts owncloud/client#4993, could be the cause of #5092 and the cost is higher than the benefit if this is the case. A network request taking more than 30 seconds isn't something unlikely in this world and shouldn't be a good reason to abort. We should try to untangle the threads dependencies to properly fix this if possible instead.


ogoffart:
:+1:

PR #5102, opened 2016-08-03
Chunking ng

ogoffart:


src/libsync/account.h
```diff
@@ -75,6 +75,9 @@ class OWNCLOUDSYNC_EXPORT Account : public QObject {
     void setSharedThis(AccountPtr sharedThis);
     AccountPtr sharedFromThis();
 
+    /// The user that can be used in dav url
```
guruz: please comment how this relates to the shibboleth user or the user name entered in the wizard

src/libsync/owncloudpropagator.cpp
```diff
@@ -271,11 +271,17 @@ PropagateItemJob* OwncloudPropagator::createJob(const SyncFileItemPtr &item) {
             }
             {
                 if (item->_direction != SyncFileItem::Up) {
-                    auto job = new PropagateDownloadFileQNAM(this, item);
+                    auto job = new PropagateDownloadFile(this, item);
                     job->setDeleteExistingFolder(deleteExisting);
                     return job;
                 } else {
-                    auto job = new PropagateUploadFileQNAM(this, item);
+                    static const bool isNg  = !qgetenv("OWNCLOUD_CHUNK_NG").isEmpty(); // FIXME! use server version
```
guruz: it's called chunkING NG everywhere else I thought?
rename?

src/libsync/propagateremotemove.h
```diff
@@ -25,8 +25,12 @@ namespace OCC {
 class MoveJob : public AbstractNetworkJob {
     Q_OBJECT
     const QString _destination;
+    const QUrl _url; // Only used (instead of path) when the constructor taking an URL is used
+    QMap<QByteArray, QByteArray> _extraHeaders;
```
guruz: I'm wondering if this should somehow be in AbstractNetworkJob? but maybe it can't if the subclasses are the ones instaciating the reply..

so not important.

src/libsync/propagateupload.cpp
```diff
@@ -887,12 +509,58 @@ void PropagateUploadFileQNAM::abort()
...
+        headers["OC-Tag"] = ".sys.admin#recall#";
+    }
+
+    if (!_item->_etag.isEmpty() && _item->_etag != "empty_etag"
+        && _item->_instruction != CSYNC_INSTRUCTION_NEW  // On new files never send a If-Match
+        && _item->_instruction != CSYNC_INSTRUCTION_TYPE_CHANGE
+        && !_deleteExisting
+    ) {
+        // We add quotes because the owncloud server always adds quotes around the etag, and
+        //  csync_owncloud.c's owncloud_file_id always strips the quotes.
```
guruz: function owncloud_file_id does not exist anymore

src/libsync/propagateupload.h
```diff
@@ -176,49 +258,62 @@ class PropagateUploadFileQNAM : public PropagateItemJob {
...
+ *
+ * Propagation job, impementing the new chunking agorithm
+ *
+ */
+class PropagateUploadFileNG : public PropagateUploadFileCommon {
+    Q_OBJECT
 private:
-    void startPollJob(const QString& path);
-    void abortWithError(SyncFileItem::Status status, const QString &error);
+    quint64 _sent; /// amount of data that was already sent
```
guruz: amount of bytes of data

src/libsync/propagateupload.h
```diff
@@ -176,49 +258,62 @@ class PropagateUploadFileQNAM : public PropagateItemJob {
...
+    uint _transferId; /// transfer id (part of the url)
+    int _currentChunk; /// Id of the next chunk that will be sent
+    QMap<int, quint64> _serverChunks; // Map chunk number with its size  from the PROPFIND on resume
```
guruz: is this map also updated by us when we sent a chunk?

src/libsync/propagateupload.h
```diff
@@ -176,49 +258,62 @@ class PropagateUploadFileQNAM : public PropagateItemJob {
...
+
+    quint64 chunkSize() const { return _propagator->chunkSize(); }
+    /**
+     * Return the URL of a chunk.
+     * If chunk == -1, returns the URL of the parent folder containing the chunks
```
guruz: mhm, sure you want to introduce this kind of magic value? maybe better to have a chunkingUrl() for the parent dir?

src/libsync/propagateuploadng.cpp
```diff
@@ -0,0 +1,401 @@
...
+#include <QFileInfo>
+#include <QDir>
+#include <cmath>
+#include <cstring>
+
+namespace OCC {
+
+QUrl PropagateUploadFileNG::chunkUrl(int chunk)
+{
+    // FIXME! we should not use the user from the credential, we should have it in the account
```
guruz: is this comment still valid? it is using _propagator->account()->user() not the credentials

src/libsync/propagateuploadng.cpp
```diff
@@ -0,0 +1,401 @@
...
+        _transferId = progressInfo._transferid;
+        auto url = chunkUrl();
+        auto job = new LsColJob(_propagator->account(), url, this);
+        _jobs.append(job);
+        job->setProperties(QList<QByteArray>() << "resourcetype" << "getcontentlength");
+        connect(job, SIGNAL(finishedWithoutError()), this, SLOT(slotPropfindFinished()));
+        connect(job, SIGNAL(finishedWithError(QNetworkReply*)),
+                this, SLOT(slotPropfindFinishedWithError()));
+        connect(job, SIGNAL(destroyed(QObject*)), this, SLOT(slotJobDestroyed(QObject*)));
+        //TODO: port to Qt4
```
guruz: Not sure what this means, but IMHO you could have the new chunking be Qt 5 only?

src/libsync/propagateuploadng.cpp
```diff
@@ -0,0 +1,401 @@
...
+
+void PropagateUploadFileNG::slotPropfindFinished()
+{
+    auto job = qobject_cast<LsColJob *>(sender());
+    slotJobDestroyed(job); // remove it from the _jobs list
+    _propagator->_activeJobList.removeOne(this);
+
+    _currentChunk = 0;
+    _sent = 0;
+    while (_serverChunks.contains(_currentChunk)) {
```
guruz: Can we get into an out-of-order situation here if a junk in the beginning is missing? (because the server timed it out)
And then we upload something "in between" and totally mess up the file on the server?

src/libsync/propagateuploadng.cpp
```diff
@@ -0,0 +1,401 @@
...
+    QNetworkReply::NetworkError err = job->reply()->error();
+    auto httpErrorCode = job->reply()->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
+    auto status = classifyError(err, httpErrorCode, &_propagator->_anotherSyncNeeded);
+    if (status == SyncFileItem::FatalError) {
+        _propagator->_activeJobList.removeOne(this);
+        QString errorString = errorMessage(job->reply()->errorString(), job->reply()->readAll());
+        abortWithError(status, errorString);
+        return;
+    }
+    startNewUpload();
```
guruz: So if the server intermittently returns 503 (maintenance mode or whatever) we break the whole upload and restart?

Might be OK though..

FYI @dragotin

src/libsync/propagateuploadng.cpp
```diff
@@ -0,0 +1,401 @@
...
+}
+
+
+void PropagateUploadFileNG::startNewUpload()
+{
+    Q_ASSERT(_propagator->_activeJobList.count(this) == 1);
+    _transferId = qrand() ^ _item->_modtime ^ (_item->_size << 16) ^ qHash(_item->_file);
```
guruz: Is this after the checksum computation? one could argue that it could be an ingredient..

src/libsync/propagateuploadng.cpp
```diff
@@ -0,0 +1,401 @@
...
+    connect(job, SIGNAL(uploadProgress(qint64,qint64)),
+            this, SLOT(slotUploadProgress(qint64,qint64)));
+    connect(job, SIGNAL(uploadProgress(qint64,qint64)),
+            device, SLOT(slotJobUploadProgress(qint64,qint64)));
+    connect(job, SIGNAL(destroyed(QObject*)), this, SLOT(slotJobDestroyed(QObject*)));
+    job->start();
+    _propagator->_activeJobList.append(this);
+    _currentChunk++;
+
+    // FIXME! parallel chunk?
```
guruz: It's sad to lose the feature, but it's also nice to have less complexity. Maybe if the scaling to a bigger chunk size is good enough you don't need to do it.

src/libsync/propagateuploadng.cpp
```diff
@@ -0,0 +1,401 @@
...
+    qDebug() << job->reply()->request().url() << "FINISHED WITH STATUS"
+             << job->reply()->error()
+             << (job->reply()->error() == QNetworkReply::NoError ? QLatin1String("") : job->reply()->errorString())
+             << job->reply()->attribute(QNetworkRequest::HttpStatusCodeAttribute)
+             << job->reply()->attribute(QNetworkRequest::HttpReasonPhraseAttribute);
+
+    _propagator->_activeJobList.removeOne(this);
+
+    if (_finished) {
+        // We have sent the finished signal already. We don't need to handle any remaining jobs
```
guruz: comment how this could happen?

src/libsync/propagateuploadng.cpp
```diff
@@ -0,0 +1,401 @@
...
+        _item->_httpErrorCode = job->reply()->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
+        QByteArray replyContent = job->reply()->readAll();
+        qDebug() << replyContent; // display the XML error in the debug
+        QString errorString = errorMessage(job->errorString(), replyContent);
+
+        if (job->reply()->hasRawHeader("OC-ErrorString")) {
+            errorString = job->reply()->rawHeader("OC-ErrorString");
+        }
+
+        // FIXME!  can tth peneunking?
```
guruz: is this klingon?

src/libsync/account.h
ogoffart: I  don't know exactly.  the answer from @DeepDiver1975 on #engineering (17:26) was not enough for me to make a better comment.

src/libsync/owncloudpropagator.cpp
ogoffart: mmh, OK.  
This is only temporary anyway

src/libsync/propagateuploadng.cpp
ogoffart: Oops, another outdated comment :-/

src/libsync/propagateupload.h
ogoffart: yeah... That's private API anyway.
Having another function mean either duplicating code, or call UrlConcat twice.

src/libsync/propagateuploadng.cpp
ogoffart: I think then we just overwrite the chunks on the server and that should be fine.

guruz: Then we must make sure to keep the same chunk size etc? Maybe this function needs to be more defensive then..
(Because we were going to dynamically adjust it based on bandwidth measurement).

FYI @DeepDiver1975 @dragotin

src/libsync/propagateuploadng.cpp
ogoffart: yes...
Most likely the error is because of 404 or something,  maybe we could have a special case for 503,  but i guess these are already taken care of by classifyError  to return a FatalError

src/libsync/propagateuploadng.cpp
```diff
@@ -0,0 +1,401 @@
...
+            _propagator->_anotherSyncNeeded = true;
+        }
+    }
+
+    // Check whether the file changed since discovery.
+    if (! FileSystem::verifyFileUnchanged(fullFilePath, _item->_size, _item->_modtime)) {
+        _propagator->_anotherSyncNeeded = true;
+        if( !finished ) {
+            abortWithError(SyncFileItem::SoftError, tr("Local file changed during sync."));
+            // FIXME:  the legacy code was retrying for a few seconds.
```
guruz: Maybe this is important because the discovery is kind of expensive, so it's cheaper to retry here than retrying the while discovery?

src/libsync/propagateuploadng.cpp
ogoffart: with parallel chunk which is not implemented yet,  but this code was based on the V1

src/libsync/propagateuploadng.cpp
```diff
@@ -0,0 +1,401 @@
...
+        // Deletes an existing blacklist entry on successful chunk upload
+        if (_item->_hasBlacklistEntry) {
+            _propagator->_journal->wipeErrorBlacklistEntry(_item->_file);
+            _item->_hasBlacklistEntry = false;
+        }
+    }
+    startNextChunk();
+}
+
+void PropagateUploadFileNG::slotMoveJobFinished()
```
guruz: ah so this is where we could implement asynchronicity in future..

src/libsync/propagateuploadng.cpp
ogoffart: not sure what hapenned there.   but i think this was something like  "can this happen for the chunk?"
In fact, this cannot happen.  But the server implementation for the If-Match header is not done yet. So maybe it could

src/libsync/propagateuploadng.cpp
```diff
@@ -0,0 +1,401 @@
...
+        SyncFileItem::Status status = classifyError(err, _item->_httpErrorCode,
+                                                    &_propagator->_anotherSyncNeeded);
+        QString errorString = errorMessage(job->errorString(), job->reply()->readAll());
+        abortWithError(status, errorString);
+        return;
+    }
+
+    QByteArray fid = job->reply()->rawHeader("OC-FileID");
+    if(fid.isEmpty()) {
+        qWarning() << "Server did not return a OC-FileID" << _item->_file;
```
guruz: This should be fatal IMHO?

src/libsync/propagateuploadng.cpp
```diff
@@ -0,0 +1,401 @@
...
+        // the old file id should only be empty for new files uploaded
+        if( !_item->_fileId.isEmpty() && _item->_fileId != fid ) {
+            qDebug() << "WARN: File ID changed!" << _item->_fileId << fid;
+        }
+        _item->_fileId = fid;
+    }
+
+    _item->_etag = getEtagFromReply(job->reply());;
+    if (_item->_etag.isEmpty()) {
+        qWarning() << "Server did not return an ETAG" << _item->_file;
```
guruz: Please re-introduce the world famous ```Server did not send an ETag``` error :P

src/libsync/propagateuploadv1.cpp
```diff
@@ -0,0 +1,375 @@
...
+        // Don't do parallel upload of chunk if this might be the last chunk because the server cannot handle that
+        // https://github.com/owncloud/core/issues/11106
+        // We return now and when the _jobs are finished we will proceed with the last chunk
+        // NOTE: Some other parts of the code such as slotUploadProgress also assume that the last chunk
+        // is sent last.
+        return;
+    }
+    quint64 fileSize = _item->_size;
+    auto headers = PropagateUploadFileCommon::headers();
+    headers["OC-Total-Length"] = QByteArray::number(fileSize);
```
guruz: Why is this not needed anymore with the new chunking? Does the server not make sanity checks anymore about file size?

src/libsync/owncloudpropagator.cpp
```diff
@@ -270,7 +270,13 @@ PropagateItemJob* OwncloudPropagator::createJob(const SyncFileItemPtr &item) {
                 job->setDeleteExistingFolder(deleteExisting);
                 return job;
             } else {
-                auto job = new PropagateUploadFile(this, item);
+                static const bool isNg  = !qgetenv("OWNCLOUD_CHUNK_NG").isEmpty(); // FIXME! use server version
```
dragotin: I think it should be rather decided by a server capability. @DeepDiver1975 does the new chunking already add something to the capabilities?

src/libsync/propagateupload.h
```diff
@@ -155,10 +160,87 @@ class PollJob : public AbstractNetworkJob {
...
+    Q_OBJECT
+
+protected:
+    QElapsedTimer _duration;
+    QVector<AbstractNetworkJob*> _jobs; /// network jobs that are currently in transit
+    bool _finished; /// Tells that all the jobs have been finished
+    bool _deleteExisting;
+
+    // measure the performance of checksum calc and upload
+    Utility::StopWatch _stopWatch;
```
dragotin: Maybe we can drop the stopwatch here, it was more to proof that checksum calculation does not take long, but these discussions are gone thankfully.

src/libsync/propagateupload.h
dragotin: ...and shouldn't the variable rather be called `_serverChunkSize` ? It actually stores the sizes.

src/libsync/propagateuploadng.cpp
```diff
@@ -0,0 +1,401 @@
...
+
+ */
+
+void PropagateUploadFileNG::doStartUpload()
+{
+    _duration.start();
+    _propagator->_activeJobList.append(this);
+
+    const SyncJournalDb::UploadInfo progressInfo = _propagator->_journal->getUploadInfo(_item->_file);
+    if (progressInfo._valid && Utility::qDateTimeToTime_t(progressInfo._modtime) == _item->_modtime ) {
```
dragotin: Don't you have to handle the cases *"progressInfo not valid"* and *"modtimes are not equal any more"* differently? There is no else branch at all, is that correct?

src/libsync/propagateuploadng.cpp
dragotin: But with the lscol job info from before we should exactly know which parts are already on the server. But for that, the upload can only be started after the lscol job info is parsed. Not sure if that is the case with the current code flow.

guruz:
We need to somehow check for the minimum versions (or capability? @PVince81 should it be a capability?) that fixes the performance
https://github.com/owncloud/core/pull/26062

PVince81:
@guruz if needed we can add a new capability value to identify that the new DAV endpoint + chunking is available

src/libsync/propagateupload.h
ogoffart: > is this map also updated by us when we sent a chunk?

No it's not.
It's only used for slotPropfindIterate

src/libsync/propagateuploadng.cpp
ogoffart: Yes, what klaas said.

We need to be sure that there is not _more_ files in the server than in the end.  that's true that this case is not yet handled.

src/libsync/propagateuploadv1.cpp
ogoffart: This is the uploadV1  (so that's the legacy code.)
We still support old server.  And I never tried if the bug was actually really fixed. There might still be race conditions, Also in the client as it may expect that when the last chunk finishes, the while file is transmtted.

src/libsync/owncloudpropagator.cpp
ogoffart: Yes of course, this is debug only right now.

src/libsync/propagateuploadng.cpp
ogoffart: There is an "else case"   that's the "startNewUpload"

guruz:
@PVince81 Not sure what you decided with @ogoffart but indeed a capability would be ideal.
This is also what @mrow4a is doing for bundling.

PVince81:
PRs to core welcome, adding a capability should be easy: apps/files/lib/Capabilities.php

ogoffart:
I thin the only missing part is to querry the capability flag.

guruz:
> I think the only missing part is to querry the capability flag.

FYI @DeepDiver1975

ogoffart:
Ready to be reviewed

src/libsync/propagateuploadng.cpp
guruz: Then handle it?

guruz:
I think this code is not defensive enough, see my comment with regards to dynamic chunk sizes etc.

Does the server always check transmission checksums by now?

PVince81:
> Does the server always check transmission checksums by now?

This is not implemented yet.

ogoffart:
Latests commits added tests,  and remove the extra chunks before starting an upload

src/libsync/propagateuploadng.cpp
```diff
@@ -0,0 +1,480 @@
...
+            _propagator->_anotherSyncNeeded = true;
+        }
+
+        SyncFileItem::Status status = classifyError(err, _item->_httpErrorCode,
+                                                    &_propagator->_anotherSyncNeeded);
+        abortWithError(status, errorString);
+        return;
+    }
+
+    bool finished = _sent >= _item->_size;
```
ckamm: Shouldn't sent > _item->_size be a fatal error? How would that happen?

src/libsync/propagateuploadng.cpp
ckamm: @guruz This is an old comment referencing the legacy-legacy code. It's the same behavior in v1.

I don't see why one would retry. If the file size/mtime changed, why would waiting and trying again be helpful? I'd be fine with removing this comment here and in v1.

src/libsync/account.h
```diff
@@ -68,6 +68,10 @@ class OWNCLOUDSYNC_EXPORT Account : public QObject {
 
     AccountPtr sharedFromThis();
 
+    /// The user that can be used in dav url
```
ogoffart: I don't know.  I was just told it could be different and that i had to do an API call even when using the HTTPCrendentials.

@DeepDiver1975   Can you clarify?

src/libsync/propagateupload.h
```diff
@@ -176,51 +261,66 @@ class PropagateUploadFile : public PropagateItemJob {
...
+    quint64 _sent; /// amount of data (bytes) that was already sent
+    uint _transferId; /// transfer id (part of the url)
+    int _currentChunk; /// Id of the next chunk that will be sent
+    bool _removeJobError; /// If not null, there was an error removing the job
 
+    // Map chunk number with its size  from the PROPFIND on resume.
+    // (Only used from slotPropfindIterate/slotPropfindFinished because the LsColJob use signals to report data.)
+    QMap<int, quint64> _serverChunks;
+
+    quint64 chunkSize() const { return _propagator->chunkSize(); }
```
ogoffart: Well, the idea would be that we could do dynamic chunking, so they would be different.

src/libsync/propagateupload.h
```diff
@@ -176,51 +261,66 @@ class PropagateUploadFile : public PropagateItemJob {
...
     /**
-     * Whether an existing entity with the same name may be deleted before
-     * the upload.
-     *
-     * Default: false.
+     * Return the URL of a chunk.
+     * If chunk == -1, returns the URL of the parent folder containing the chunks
      */
-    void setDeleteExisting(bool enabled);
+    QUrl chunkUrl(int chunk = -1);
```
ogoffart: yeah, that's true. But it's internal function anywaéy

src/libsync/propagateuploadng.cpp
```diff
@@ -0,0 +1,480 @@
...
+    emit progress(*_item, 0);
+
+    SyncJournalDb::UploadInfo pi;
+    pi._valid = true;
+    pi._transferid = _transferId;
+    pi._modtime =  Utility::qDateTimeFromTime_t(_item->_modtime);
+    _propagator->_journal->setUploadInfo(_item->_file, pi);
+    _propagator->_journal->commit("Upload info");
+    QMap<QByteArray, QByteArray> headers;
+    headers["OC-Total-Length"] = QByteArray::number(_item->_size);
```
ogoffart: Yes, because the _transferId would be different  (the size is xor'ed with it)

guruz:
@PVince81 @ogoffart What's the state about checksum checking on the server side?

ogoffart:
@guruz the client send the checksum header (for the whole file) in the MOVE operation.
I think it would be overkill to do checksum check on every chunk.

guruz:
I'm working on a server patch that does checksum verifying for chunking ng (only).
Then the client code could check if Capabilities::supportedChecksumTypes returns something and only then enable chunking ng.

test/testchunkingng.cpp
```diff
@@ -0,0 +1,78 @@
...
+ *
+ */
+
+#include <QtTest>
+#include "syncenginetestutils.h"
+#include <syncengine.h>
+
+using namespace OCC;
+
+class TestChunkingNG : public QObject
```
DeepDiver1975: I see the tmp folder being polluted with folders named ChunkingNG-* - is there some cleanup missing?

```sh
root@ci:/tmp/ChunkingNgTest-AQJojC/A# ls -lh
total 287M
-rw-r--r-- 1 jenkins jenkins 287M Nov 11 02:34 a0
-rw-r--r-- 1 jenkins jenkins    4 Nov  4 02:35 a1
-rw-r--r-- 1 jenkins jenkins    4 Nov  4 02:35 a2

ogoffart: The cleanup is there if the test does not crash.  But it seems that the test is crashing.   need to investigate why.

ogoffart:
Checksuming is a server issue and is orthogonal to this problem

guruz:
@SamuAlfageme This needs to be tested with a server 9.2 (e.g. a quite current master branch of the server), previous servers don't have this chunking ng support.

SamuAlfageme:
@guruz noted, will do 😉

PR #5104, opened 2016-08-04
SyncEngine: Fix detection of backup or full resync

ogoffart:
Once upon a time, the SyncEngine was instantiated once per sync. But now that the SyncEngine is kept between sync, we need to reset all these variable between syncs.

### DIFF
--- a/src/libsync/syncengine.cpp
+++ b/src/libsync/syncengine.cpp
@@ -847,6 +847,8 @@ void SyncEngine::slotDiscoveryJobFinished(int discoveryResult)
 
     _hasNoneFiles = false;
     _hasRemoveFile = false;
+    _hasForwardInTimeFiles = false;
+    _backInTimeFiles = 0;
     bool walkOk = true;
     _seenFiles.clear();
     _temporarilyUnavailablePaths.clear();


ogoffart:
Does that go into 2.2.3 ?

src/libsync/syncengine.cpp
```diff
@@ -681,6 +681,10 @@ void SyncEngine::startSync()
     s_anySyncRunning = true;
     _syncRunning = true;
     _anotherSyncNeeded = false;
+    _hasNoneFiles = false;
+    _hasRemoveFile = false;
```
jturcotte: Those two are already reset in slotDiscoveryJobFinished

src/libsync/syncengine.cpp
```diff
@@ -681,6 +681,10 @@ void SyncEngine::startSync()
...
+    _hasForwardInTimeFiles = false;
+    _backInTimeFiles = 0;
```
jturcotte: Could we reset those two in slotDiscoveryJobFinished as well?

guruz:
OK let's have it in 2.2.3 and make a rc tomorrow..

jturcotte:
👍

guruz:
👍  from me.
Waiting for second 👍 from @jturcotte then @ogoffart can merge

guruz:
@davitol FYI nightlies should have it

davitol:
@guruz @ogoffart Tested and works fine, Great job 💃

PR #5108, opened 2016-08-05
Limit Qt and OpenSSL Library pathes to the install directory

danimo:


danimo:
@guruz pls

src/gui/main.cpp
```diff
@@ -30,6 +30,7 @@
 
 #include <QTimer>
 #include <QMessageBox>
+#include <QFileInfo>
```
guruz: remove?

PR #5122, opened 2016-08-16
Syncfilestatustracker tests

jturcotte:


ogoffart:
2507ba9 looks good.  	0e94022 looks good.    	e194176 looks good at least for the changes in the sync engine.  I did not read the full tests in details.

	3507bcc is a bit more complicated to review. Looks fine, but also dangerous. maybe @ckamm can also have a look.

75e0e93 also look fine.

src/libsync/syncfilestatustracker.cpp
```diff
@@ -168,9 +169,7 @@ void SyncFileStatusTracker::decSyncCount(const QString &relativePath, EmitStatus
 
 void SyncFileStatusTracker::slotAboutToPropagate(SyncFileItemVector& items)
 {
-    // New sync, clear the sync counts to reduce the impact of a bug that would cause unsymetrical inc/dec
     Q_ASSERT(_syncCount.isEmpty());
-    _syncCount.clear();
```
ogoffart: you would keep it.

jturcotte: I don't think anything can increment this between slotSyncFinished and here. I was even considering removing the assert.

ckamm:
Can review in detail tomorrow!

ckamm:
I've carefully reviewed 3507bcc and like it, particularly the way it makes metadata updating much more explicit and easy to understand. It's also a nice step towards being able to potentially push metadata-only changes to the remote eventually.

I agree with @ogoffart that it's tricky and somewhat dangerous, there are many corner cases.

test/syncenginetestutils.h
```diff
@@ -0,0 +1,658 @@
...
+    void insert(const QString &relativePath, qint64 size = 64, char contentChar = 'W') override {
+        create(relativePath, size, contentChar);
+    }
+    void setContents(const QString &relativePath, char contentChar) override {
+        FileInfo *file = findInvalidatingEtags(relativePath.split('/', QString::SkipEmptyParts));
+        Q_ASSERT(file);
+        file->contentChar = contentChar;
+    }
+    void appendByte(const QString &relativePath) override {
+        FileInfo *file = findInvalidatingEtags(relativePath.split('/', QString::SkipEmptyParts));
```
ckamm: Minor: `.split('/', QString::SkipEmptyParts)` and `.mid(0, pathComponents.size() - 1)` happen often enough that they might be functions.

test/syncenginetestutils.h
```diff
@@ -0,0 +1,658 @@
...
+    }
+    FileInfo *create(const QString &relativePath, qint64 size, char contentChar) {
+        const QStringList pathComponents = relativePath.split('/', QString::SkipEmptyParts);
+        const QString childName = pathComponents.last();
+        FileInfo *parent = findInvalidatingEtags(pathComponents.mid(0, pathComponents.size() - 1));
+        Q_ASSERT(parent);
+        FileInfo &child = parent->children[childName] = FileInfo{childName, size};
+        child.parentPath = parent->path();
+        child.contentChar = contentChar;
+        child.etag = QString::number(QDateTime::currentDateTime().toMSecsSinceEpoch(), 16);
```
ckamm: createDir() + fi->contentChar = contentChar? I appreciate this might not hold true forever.

test/syncenginetestutils.h
```diff
@@ -0,0 +1,658 @@
...
+    }
+    bool operator<(const FileInfo &other) const {
+        return name < other.name;
+    }
+
+    bool operator==(const FileInfo &other) const {
+        return name == other.name
+            && isDir == other.isDir
+            && size == other.size
+            && children == other.children;
```
ckamm: Where's this used? Possibly comment that the other fields are not compared intentionally?

test/syncenginetestutils.h
```diff
@@ -0,0 +1,658 @@
...
+            // Ignore outgoingData always returning somethign good enough, works for now.
+            return new FakePropfindReply{_remoteRootFileInfo, op, request, this};
+        else if (verb == QLatin1String("GET"))
+            return new FakeGetReply{_remoteRootFileInfo, op, request, this};
+        else if (verb == QLatin1String("PUT"))
+            return new FakePutReply{_remoteRootFileInfo, op, request, outgoingData->readAll(), this};
+        else if (verb == QLatin1String("MKCOL"))
+            return new FakeMkcolReply{_remoteRootFileInfo, op, request, this};
+        else if (verb == QLatin1String("DELETE"))
+            return new FakeDeleteReply{_remoteRootFileInfo, op, request, this};
```
ckamm: Initially I was a bit confused that these are not only Fake*Replys* but they actually act on and modify the FileInfo. That's of course neat and much nicer, just surprised me given the name.

src/libsync/syncfilestatustracker.cpp
```diff
@@ -140,22 +127,76 @@ void SyncFileStatusTracker::slotPathTouched(const QString& fileName)
...
             _syncProblems[item->_file] = SyncFileStatus::StatusWarning;
         }
+
+        // Mark this path as syncing for instructions that will result in propagation,
+        // but DontEmitStatusChange since we're going to emit for ourselves using the
+        // info in the SyncFileItem we received, parents will still be emit if needed.
+        if (item->_instruction != CSYNC_INSTRUCTION_NONE
+            && item->_instruction != CSYNC_INSTRUCTION_UPDATE_METADATA
+            && item->_instruction != CSYNC_INSTRUCTION_IGNORE
+            && item->_instruction != CSYNC_INSTRUCTION_ERROR) {
```
ckamm: I think this same condition ("does this need propagation?") is in progressdispatcher.cpp:shouldCountProgress - could this be a shared function?

test/syncenginetestutils.h
jturcotte: I'm not sure I understand what you mean, this one is create() for files, createDir is above (but could use an empty line).

test/syncenginetestutils.h
jturcotte: Added a PathComponent class, let me know what you think.

test/syncenginetestutils.h
jturcotte: Added a comment, and a missing check for contentChar (and fix for the test failure that happened after adding that check).

test/syncenginetestutils.h
jturcotte: The "fake" comes from the "fake object" term used in testing. If I understand the term correctly myself, they aren't real replies, and that they do more than mocking, they implement a seam which you can use as a test harness.

src/libsync/syncfilestatustracker.cpp
jturcotte: Humm not sure, I think that "should count as progress" is a different requirement than "should be shown as syncing in the overlay icons". Moreover, the overlay icon situation is also restricted by what can give us a symetrical inc/dec, so I wouldn't want a change in the progress logic to inadvertently affect the status tracker. It's also not a big deal if they diverge in the future, so I would only see saving a few similar lines of code as a benefit.

ckamm:
:+1:

I've also run smashbox tests lib/owncloud/* on this branch. lib/owncloud/test_remoteShareFile.py failed, but it also fails on master, maybe a setup issue.

jturcotte:
Merged in 2ff7b63551d342ef6747e2723ca7007f57c90b83, thanks for the reviews!

Issue #5120, opened 2016-08-15
smashbox testDirBecomesFile fails


PVince81:
Run: `bin/smash -a lib/owncloud/test_dirBecomesFile.py`
Check the log result in the smashdir.

With the desktop client 2.2.3 the test will fail.
With 2.1.1 it passes (compiled from source). I didn't try other versions, a bisect should be possible.

From what I understand the test tries to sync a dir, then replaces it with a file.

Result:
```
********************************************************************************
(2) dir_to_file:  TURN THE FOLDER INTO A FILE LOCALLY AND PROPAGATE TO THE SERVER
********************************************************************************
2016-08-15 16:12:45,120 - INFO - dir_to_file - running '/usr/bin/owncloudcmd --trust /home/vincent/work/smashdir/test_dirBecomesFile/dir_to_file/folder2/ owncloud://test:test@127.0.0.1/owncloud/remote.php/webdav/ >> /home/vincent/work/smashdir/test_dirBecomesFile/dir_to_file-ocsync.step02.cnt000.log 2>&1'
2016-08-15 16:12:46,947 - INFO - dir_to_file - sync cmd is: /usr/bin/owncloudcmd --trust /home/vincent/work/smashdir/test_dirBecomesFile/dir_to_file/folder2/ owncloud://test:test@127.0.0.1/owncloud/remote.php/webdav/ >> /home/vincent/work/smashdir/test_dirBecomesFile/dir_to_file-ocsync.step02.cnt000.log 2>&1
2016-08-15 16:12:46,947 - INFO - dir_to_file - sync finished: 0:00:01.827289
2016-08-15 16:12:46,947 - INFO - dir_to_file - running 'mv /home/vincent/work/smashdir/test_dirBecomesFile/dir_to_file/folder2/dirtofile/file-move.txt /home/vincent/work/smashdir/test_dirBecomesFile/dir_to_file/folder2/file-move.txt'
2016-08-15 16:12:46,951 - INFO - dir_to_file - running 'mv /home/vincent/work/smashdir/test_dirBecomesFile/dir_to_file/folder2/dirtofile/sub_move /home/vincent/work/smashdir/test_dirBecomesFile/dir_to_file/folder2/sub_move'
2016-08-15 16:12:46,955 - INFO - dir_to_file - createfile /home/vincent/work/smashdir/test_dirBecomesFile/dir_to_file/folder2/dirtofile character='N' count=1000 bs=50
[...]

PR #5123, opened 2016-08-17
Fix local to remote file type change propagation #5120

ckamm:
Addresses the failing test_dirBecomesFile test in smashbox.


ckamm:
Let's merge #5122 instead, which also fixes the problem.

Issue #5116, opened 2016-08-10
OS X client fails to see watcher change if sync directory is a symlink


wrigjz:
OS X client Sync to Linux server has issues - appears to be related to the OS X users home directory being a symlink.

[b]Steps to reproduce[/b]
1. Add a file to ~/ownCloud directory on client
2. Wait - (10 mins+) no sync 0ccurs
3. Click Pause/UnPause sync and it will start to sync

[b]Expected behaviour[/b]
Adding a file should sync without needing to Pause/UnPause

[b]Actual behaviour[/b]
No sync - we need to Pause/Unpause the client (perhaps 10 mins later) for the sync to happen to the server

[b]Client configuration[/b]
Client version: 2.2.3 (also 2.1, 2.2)
Client operating system: OSX 10.11.6

[b]Server configuration[/b]
Operating system: Centos 7.2
Web server: Apache 2.4.6
Database: mariadb 1.5.5
PHP version: php5.5
ownCloud version (see ownCloud admin page): 9.1
Updated from an older ownCloud or fresh install: Updated from 9.0
ownCloud log (data/owncloud.log):
N/A

<b> Special configurations (external storage, external authentication, reverse proxy, server-side-encryption): </b> On the client the home directory is in one place but symlinked to somewhere else in order to match our other Linux machines. Perhaps this is messing owncloud client up?

<b>Issue</b>
The client log indicates that when the file is added the first time it is ignored (although no ignore options have been set), the ignore files is given as '*' i.e. everything. This however can not be the case because pause/unpause then starts the sync.

<b>Initial problem</b>
I removed the ~/ownCloud directory and then mkdir ~/ownCloud (this is set in the symlinked home directory to be /home/disk3/jon/ownCloud), I then restarted the owncloud client and everything sync'd fine from the server. using this new ownCloud directory I then added a file (cp ... ~/owncloud) and it was not sync'ed to the server.

<b>Debugging attempts</b>
On adding the file:
08-10 13:11:33:366 0x7fe941f13500 OCC::FolderMan::slotFolderSyncFinished: <===================================== sync finished for  "https://xxx.xxx.org/owncloud/remote.php/webdav/"
08-10 13:11:35:205 0x7fe941f13500 -[DelegateObject updaterDidNotFindUpdate:]: -[DelegateObject updaterDidNotFindUpdate:]
08-10 13:11:39:794 0x7fe941f13500 OCC::callback: FolderWatcherPrivate::callback by OS X
[...]

PR #5153, opened 2016-09-05
FolderWatcher: Act on relative paths #5116

guruz:


dragotin:
Can you explain why this patch helps? Thx.

guruz:
See @ckamm 's comments in the original issue.
Basically if the sync directory or the directory containing the sync directory is a symlink, the folder watcher fired events for that link target)
E.g. on OS X /tmp is actually /private/tmp so if you have a sync directory in /tmp you'd get the watcher events for /private/tmp which will then be ignored by the sync client

guruz:
@ckamm Tested, please review and squash-merge and add to https://github.com/owncloud/client/issues/5168

ckamm:
:+1:

PR #5154, opened 2016-09-05
Selective Sync: Fix request loop and show error in view

guruz:
I got into a situation where the model would endlessly request the directory
contents from the server because we did not notice yet that the server
is actually in maintenance mode while we were expanding the tree view when
changing the tab to the account or when just expanding it by clicking.


ckamm:
:+1:

PR #5155, opened 2016-09-05
prototype of bundled requests in the client

mrow4a:
- [x] Client feature is able to create bundled INSTRUCTION_NEW request.
- [x] Bundle could contain only file which is smaller than chunking size
- [x] Bundle may contain only files with total is smaller than chunking size
- [x] Bundle is compatibile with https://github.com/owncloud/core/pull/25760
- [x] Bundle correctly adds the files to both remote and local database - checked by manual tests
- [ ] Bundle correctly adds the files to both remote and local database - checked by unit tests
- [ ] Bundle correctly adds the files to both remote and local database - checked by integration tests
- [x] Bundle passes prove of concept tests
- [ ] Integrated error handling
- [x] Boundary generator
- [ ] Add support for Upload Device to limit bandwidth usage

### DIFF
--- a/src/cmd/cmd.cpp
+++ b/src/cmd/cmd.cpp
@@ -58,6 +58,7 @@ struct CmdOptions {
     QString password;
     QString proxy;
     bool silent;
+    bool bundleRequests;
     bool trustSSL;
     bool useNetrc;
     bool interactive;
@@ -157,6 +158,7 @@ void help()
     std::cout << std::endl;
     std::cout << "Options:" << std::endl;
     std::cout << "  --silent, -s           Don't be so verbose" << std::endl;
+    std::cout << "  --bundle-requests, -b  Bundle Requests if supported" << std::endl;
     std::cout << "  --httpproxy [proxy]    Specify a http proxy to use." << std::endl;
     std::cout << "                         Proxy is http://server:port" << std::endl;
     std::cout << "  --trust                Trust the SSL certification." << std::endl;
@@ -224,6 +226,8 @@ void parseOptions( const QStringList& app_args, CmdOptions *options )
             options->silent = true;
         } else if( option == "--trust") {
             options->trustSSL = true;
+        } else if( option == "-b" || option == "--bundle-requests") {
+            options->bundleRequests = true;
         } else if( option == "-n") {
             options->useNetrc = true;
         } else if( option == "-h") {
@@ -292,6 +296,7 @@ int main(int argc, char **argv) {
 
     CmdOptions options;
     options.silent = false;
+    options.bundleRequests = false;
     options.trustSSL = false;
     options.useNetrc = false;
     options.interactive = true;
@@ -416,6 +421,7 @@ int main(int argc, char **argv) {
 
     loop.exec();
 #endif
+    account->setBundleRequestsIfCapable(options.bundleRequests);
 
     // much lower age than the default since this utility is usually made to be run right after a change in the tests
     SyncEngine::minimumFileAgeForUpload = 0;
--- a/src/libsync/account.h
+++ b/src/libsync/account.h
@@ -33,6 +33,7 @@ class QSettings;
 class QNetworkReply;
 class QUrl;
 class QNetworkAccessManager;
+class QHttpMultiPart;
 
 namespace OCC {
 
@@ -63,6 +64,26 @@ public:
 class OWNCLOUDSYNC_EXPORT Account : public QObject {
     Q_OBJECT
 public:
+    /**
+     * @brief The possibly themed dav path for the account. It has
+     *        a trailing slash.
+     * @returns the (themeable) dav path for the account.
+     */
+    QString davPath() const;
+
+    /**
+     * @brief The possibly themed files dav path for the account. It has
+     *        a trailing slash.
+     * @returns the (themeable) files dav path for the account.
+     */
+    QString davFilesPath() const;
+    void setDavPath(const QString&s) { _davPath = s; }
+    void setNonShib(bool nonShib);
+
+    /// Functions for bundle support
+    void setBundleRequestsIfCapable(bool bundleRequests);
+    bool bundledRequestsEnabled();
+
     static AccountPtr create();
     ~Account();
 
@@ -77,15 +98,6 @@ public:
     /** Server url of the account */
     void setUrl(const QUrl &url);
     QUrl url() const { return _url; }
-
-    /**
-     * @brief The possibly themed dav path for the account. It has
-     *        a trailing slash.
-     * @returns the (themeable) dav path for the account.
-     */
-    QString davPath() const;
-    void setDavPath(const QString&s) { _davPath = s; }
-    void setNonShib(bool nonShib);
 
     /** Returns webdav entry URL, based on url() */
     QUrl davUrl() const;
@@ -103,6 +115,8 @@ public:
     QNetworkReply* deleteRequest( const QUrl &url);
     QNetworkReply* davRequest(const QByteArray &verb, const QString &relPath, QNetworkRequest req, QIODevice *data = 0);
     QNetworkReply* davRequest(const QByteArray &verb, const QUrl &url, QNetworkRequest req, QIODevice *data = 0);
+    QNetworkReply* multipartRequest(const QString &relPath, QNetworkRequest req, QHttpMultiPart *multiPart);
+    QNetworkReply* multipartRequest(const QUrl &url, QNetworkRequest req, QHttpMultiPart *multiPart);
 
 
     /** The ssl configuration during the first connection */
@@ -216,6 +230,8 @@ private:
     QByteArray _pemCertificate; 
     QString _pemPrivateKey;  
     QString _davPath; // defaults to value from theme, might be overwritten in brandings
+    bool _wasMigrated;
+    bool _bundleRequests;
     friend class AccountManager;
 };
 
--- a/src/libsync/owncloudpropagator.cpp
+++ b/src/libsync/owncloudpropagator.cpp
@@ -303,6 +303,13 @@ void OwncloudPropagator::start(const SyncFileItemVector& items)
     directories.push(qMakePair(QString(), _rootJob.data()));
     QVector<PropagatorJob*> directoriesToRemove;
     QString removedDirectory;
+
+    PropagateBundle *bundledUploadRequestsJob = new PropagateBundle(this);
+    quint64 chunkSize = OwncloudPropagator::chunkSize();
+
+    // TODO: here we should also check somehow if bundle is not broken for specific sync - bug recovery to standard PUTs
+    bool bundledRequestsEnabled = _account->bundledRequestsEnabled();
+
     foreach(const SyncFileItemPtr &item, items) {
 
         if (!removedDirectory.isEmpty() && item->_file.startsWith(removedDirectory)) {
@@ -379,6 +386,12 @@ void OwncloudPropagator::start(const SyncFileItemVector& items)
                 currentDirJob->append(dir);
             }
             directories.push(qMakePair(item->destination() + "/" , dir));
+        } else if (bundledRequestsEnabled
+                  && (item->_instruction == CSYNC_INSTRUCTION_NEW)
+                  && (item->_direction == SyncFileItem::Up)
+                  && (item->_size < chunkSize)) {
+            //this will create list of bundle files to sync for that bundlejob
+            bundledUploadRequestsJob->append(item);
         } else if (PropagateItemJob* current = createJob(item)) {
             if (item->_instruction == CSYNC_INSTRUCTION_TYPE_CHANGE) {
                 // will delete directories, so defer execution
@@ -392,6 +405,21 @@ void OwncloudPropagator::start(const SyncFileItemVector& items)
 
     foreach(PropagatorJob* it, directoriesToRemove) {
         _rootJob->append(it);
+    }
+
+    // Root job is DirectoryJob with FullParallelizm, so we need to halt bundle till all other jobs are finished.
+    // If there are no running jobs in root folder it means we can start with sending the bundle
+    // If there are running jobs, we halt and wait for finish of the other job to trigger sending
+    if (bundledRequestsEnabled && !bundledUploadRequestsJob->empty()) {
+        bundledUploadRequestsJob->_item->_direction = SyncFileItem::Direction::Up;
+        bundledUploadRequestsJob->_item->_type = SyncFileItem::Type::RequestsContainer;
+        bundledUploadRequestsJob->_item->_instruction = CSYNC_INSTRUCTION_NEW;
+        bundledUploadRequestsJob->_item->_size = bundledUploadRequestsJob->syncItemsSize();
+        bundledUploadRequestsJob->_item->_originalFile = tr("%1 file(s)").arg(bundledUploadRequestsJob->syncItemsNumber());
+        _rootJob->append(bundledUploadRequestsJob);
+    }
+    else{
+        delete bundledUploadRequestsJob;
     }
 
     connect(_rootJob.data(), SIGNAL(itemCompleted(const SyncFileItem &, const PropagatorJob &)),
@@ -624,6 +652,7 @@ bool PropagateDirectory::scheduleNextJob()
         Q_ASSERT(_subJobs.at(i)->_state == Running);
 
         auto paral = _subJobs.at(i)->parallelism();
+
         if (paral == WaitForFinished) {
             return false;
         }
--- a/src/libsync/propagateupload.cpp
+++ b/src/libsync/propagateupload.cpp
@@ -28,10 +28,12 @@
 
 #include <json.h>
 #include <QNetworkAccessManager>
+#include <QHttpMultiPart>
 #include <QFileInfo>
 #include <QDir>
 #include <cmath>
 #include <cstring>
+#include <QXmlStreamReader>
 
 #if QT_VERSION < QT_VERSION_CHECK(5, 4, 2)
 namespace {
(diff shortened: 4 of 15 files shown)


mrow4a:
I am appending first prove of concept test results:

- both client and server on the same machine
- localhost
- Intel® Core™ i7-5500U CPU @ 2.40GHz × 4 
- 8GB RAM

resultsbundling

@cdamken @dragotin @DeepDiver1975 @PVince81

mrow4a:
@guruz If you could add some bottom line here or review, please tell me. I need all the desktop sync guys experience now. It is not that easy to integrate it into the GUI client and handle errors.

src/cmd/cmd.cpp
```diff
@@ -216,7 +218,9 @@ void parseOptions( const QStringList& app_args, CmdOptions *options )
         } else if( option == "-s" || option == "--silent") {
             options->silent = true;
         } else if( option == "--trust") {
-            options->trustSSL = true;
```
ogoffart: involuntary change?

mrow4a: Ah, yes, I did not notice that, thanks!

src/libsync/account.h
```diff
@@ -225,6 +239,7 @@ protected Q_SLOTS:
     QString _pemPrivateKey;  
     QString _davPath; // defaults to value from theme, might be overwritten in brandings
     bool _wasMigrated;
+    bool _bundleRequests;
```
ogoffart: Must be initialized in the constrcutor

src/libsync/owncloudpropagator.cpp
```diff
@@ -390,6 +400,12 @@ void OwncloudPropagator::start(const SyncFileItemVector& items)
         }
     }
 
+    //TODO Add support for more bundled jobs if bundleJob->append fails, please refer to class declaration
+    if (bundledRequestsEnabled && !bundledRequestsJob->empty()) {
+        _rootJob->append(bundledRequestsJob);
+    }
```
ogoffart: else, you shall delete the job, otherwise you get memory leak

src/libsync/propagateupload.h
```diff
@@ -220,5 +259,38 @@ private slots:
     void abortWithError(SyncFileItem::Status status, const QString &error);
 };
 
+/**
+ * @brief The PropagateBundleQNAM class
+ * @ingroup libsync
+ */
+class PropagateBundleQNAM : public PropagateItemJob {
```
ogoffart: don't add QNAM in the job t-name.  This was just to distinguish the older joobs from the legacy neon one.  But today, no reason to keep QNAM in the name.

src/libsync/propagateupload.h
```diff
@@ -220,5 +259,38 @@ private slots:
...
+    Q_OBJECT
+
+private:
+    QVector<AbstractNetworkJob*> _jobs; /// network jobs that are currently in transit
+    QVector<SyncFileItemPtr> _bundledFiles;
+    QJsonObject _bundledFilesMetadata;
+    quint64 _contentID = 1;
```
ogoffart: Initializing things here is a C++11 extention that we can't unfortunately use.
(Unless we finaly decide to drop old compilers)
It needs to be initialized in the constructor.

src/libsync/propagateupload.h
```diff
@@ -220,5 +259,38 @@ private slots:
...
+    quint64 chunkSize() const { return _propagator->chunkSize(); }
+
+    bool _finished; // Tells that all the jobs have been finished
```
ogoffart: Need to be initialized in the constructor.

src/libsync/propagateupload.h
```diff
@@ -220,5 +259,38 @@ private slots:
...
+
+public:
+    PropagateBundleQNAM(OwncloudPropagator* propagator)
+        : PropagateItemJob(propagator, SyncFileItemPtr(new SyncFileItem)) {}
+    void start() Q_DECL_OVERRIDE;
+    bool append(SyncFileItemPtr bundledFile);
```
ogoffart: object that hae a copy constructor should preferably be taken as const reference.
So `const SyncFileItemPtr &bundledFile`

src/libsync/propagateupload.h
```diff
@@ -220,5 +259,38 @@ private slots:
...
+    QString getRemotePath(QString filePath);
```
ogoffart: same here,  `const QString &`

src/libsync/propagateupload.cpp
```diff
@@ -894,5 +898,392 @@ void PropagateUploadFile::abortWithError(SyncFileItem::Status status, const QStr
     done(status, error);
 }
 
+void PropagateBundleQNAM::start()
+{
+    QHttpMultiPart *multiPart = new QHttpMultiPart(QHttpMultiPart::RelatedType);
+    multiPart->setBoundary("boundrary123");
```
ogoffart: This looks wierd. I suppose this should not be hardcoded, but be a sufficiently long random string.

src/libsync/propagateupload.cpp
```diff
@@ -894,5 +898,392 @@ void PropagateUploadFile::abortWithError(SyncFileItem::Status status, const QStr
...
+QString PropagateBundleQNAM::getRemotePath(QString filePath){
+    return QString((_propagator->_remoteFolder) + filePath);
+}
+
+bool PropagateBundleQNAM::append(SyncFileItemPtr bundledFile){
+    //TODO: validate here that
+    _bundledFiles.append(bundledFile);
+    QJsonObject fileMetadata;
+    fileMetadata["oc-path"] = getRemotePath(bundledFile->_file);
+    fileMetadata["x-oc-mtime"] = QString::number(qint64(bundledFile->_modtime));
```
ogoffart: don't you set the mtime already in the other function?

mrow4a: No, it has to be done here. What you do , you create here the JSON file describing the structure of created bundle.

src/libsync/propagateupload.cpp
```diff
@@ -894,5 +898,392 @@ void PropagateUploadFile::abortWithError(SyncFileItem::Status status, const QStr
...
+            //TODO ERROR, it means response did not return any status for the file!!
+            continue;
+        }
+
+        if (!_propagator->_journal->setFileRecord(SyncJournalFileRecord(*item, _propagator->getFilePath(item->_file)))) {
+            done(SyncFileItem::FatalError, tr("Error writing metadata to the database"));
+            //return;
+        }
+        // Remove from the progress database:
+        _propagator->_journal->setUploadInfo(item->_file, SyncJournalDb::UploadInfo());
```
ogoffart: I don't think you need this.

ogoffart:
Looks quite good!   The TODO items still need to be done. In particular the error reporting, and also we should split so that the size don't exceed the chunk size.

ogoffart:
> /var/lib/jenkins/workspace/owncloud-client/client/PR-5155/src/libsync/propagateupload.h:22:23: fatal error: QJsonObject: No such file or directory

Yeah, QJsonObject is Qt5 only.  There is another QJson library which we use all over mirall.

mrow4a:
Isnt it 20th century? Why do we support all old C++ and old Qt?

guruz:
I'm fine with having the feature Qt 5 only :)

src/libsync/propagateupload.cpp
guruz: Could be that it's not really necessary because it's already set: https://code.woboq.org/qt5/qtbase/src/network/access/qhttpmultipart.cpp.html#443
Maybe qDebug() the boundary to verify..

src/libsync/propagateupload.cpp
guruz: Why?

src/libsync/owncloudpropagator.cpp
```diff
@@ -379,6 +383,12 @@ void OwncloudPropagator::start(const SyncFileItemVector& items)
                 currentDirJob->append(dir);
             }
             directories.push(qMakePair(item->destination() + "/" , dir));
+        } else if (bundledRequestsEnabled
+                  && (item->_instruction == CSYNC_INSTRUCTION_NEW)
```
guruz: Good that it is only for new files, maybe for updating files there is too many other things to consider for now, e.g. If-Etag-match and permissions and concurrency/locks etc.

src/libsync/propagateupload.h
guruz: ...but you should add "Upload" somewhere to the name?

src/libsync/propagateupload.h
```diff
@@ -124,6 +124,45 @@ private slots:
 };
 
 /**
+ * @brief The MultipartJob class
+ * @ingroup libsync
+ */
+class MultipartJob : public AbstractNetworkJob {
```
guruz: Should this have "Upload" in the name?

src/libsync/propagateupload.cpp
```diff
@@ -894,5 +898,392 @@ void PropagateUploadFile::abortWithError(SyncFileItem::Status status, const QStr
...
+        // Abort the job and try again later.
+        // This works around a bug in QNAM wich might reuse a non-empty buffer for the next request.
+        qDebug() << "Forcing job abort on HTTP connection reset with Qt < 5.4.2.";
+        _propagator->_anotherSyncNeeded = true;
+        abortWithError(SyncFileItem::SoftError, tr("Forcing job abort on HTTP connection reset with Qt < 5.4.2."));
+        return;
+    }
+#endif
+
+    //TODO might need verifying that it is multistatus?
```
guruz: Would be good indeed just as a sanity check

src/libsync/propagateupload.cpp
```diff
@@ -894,5 +898,392 @@ void PropagateUploadFile::abortWithError(SyncFileItem::Status status, const QStr
...
+                const QString fullFilePath(_propagator->getFilePath(item->_file));
+                if( !FileSystem::fileExists(fullFilePath) ) {
+                    //abortWithError(SyncFileItem::SoftError, tr("The local file was removed during sync."));
+                }
+
+                if (! FileSystem::verifyFileUnchanged(fullFilePath, item->_size, item->_modtime)) {
+                    _propagator->_anotherSyncNeeded = true;
+                }
+
+                //OC-FileID section
```
guruz: I would say it needs to have fileID, Etag and mtime:accepted else it's an error.. 
@ogoffart ?

src/libsync/propagateupload.h
mrow4a: Not convinced, the goal with bundling is to bundle all requests in Bunlemetadata JSON Object

guruz: Ambitious :-) But cool..

guruz:
Great patch so far. Let's be safe with this and mandate the use of content-checksum.
https://github.com/owncloud/core/pull/25760#issuecomment-239110848

mrow4a:
I discussed that with Olivier. Checksums is the next pull request.

mrow4a:
Update on bundling prototype on central:
https://central.owncloud.org/t/files-bundling-extension-multiple-files-update-creation-in-various-locations-on-the-server/1295/20

mrow4a:
@owncloud/desktop-developers Ok, This version of bundle has fixed the problem with very nested directories:

What it will do is:
1. Execute all root folder _rootJob and skip the bundle job (or execute it if it is the only job)
https://github.com/owncloud/client/pull/5155/files#diff-a8e0763851310eb229ea750c1f9b62aeR287

2. After all directory biased jobs will be executed in parallel- MKDIR, MOVE, DELETE, GET, PUT(big files) - being executed the old sync way, it detects that bundling upload is the only one left, and will execute bundled PUT requests.

TODO: unit tests and error handling verification. 

This is sync of around 300 files in nested directories with cross directory bundled PUTs in few POST - 156 MB:
selection_039

mrow4a:
Hey guys. I recently been thinking how to avoid keeping files in the memory on the server side parsing whole the bundle. 

In all cases, we expect that if we send a request with 500 files, we expect 500 responses or one saying parsing error at the begining

In current implementation, when each component is separate part, I have to in advance parse each body or seek to the position of the header of that file. Then, I have to either take this file from memory after parsing or seek back to the request beginning (which is not working well for me at current ownCloud implementation, since you can safely read the request body once - to be honest it is not working even though it is php5.6, and this complicates implementation a lot).

This means I have to keep all the files in memory after parsing. This is also quite inefficient as you know, and will not give the clear expected gain.

I propose to get back to the previous idea, however lets make each file metadata separate part, and the last part will be part containing all the files together. Each file part metadata will contain an information what is the content-length and what is an offset at which I can find a file. This will give me a clear advantage of reading and adding a request on-fly directly from request body, not saving it additionaly in memory or temporary file.

@ogoffart @jturcotte @guruz @DeepDiver1975 @butonic

jturcotte:
I'm far from knowing how that works in PHP, but why can't you do each request one-by-one, keeping the Content-Length to know the potential size of the body, and send the response on the socket after each request is complete?

I thought that you're basically re-implementing HTTP Connection: keep-alive, if the HTTP server can do it, what in the design prevents you from doing the same on the part level?

mrow4a:
This will be nice, but in PHP you have one request one response. It does not work like that. You have to finish processing whole request with emits to send a response.

jturcotte:
I thought that you were encoding each WebDAV request in a multipart message of the same HTTP request body? I mean that if you have 100 parts encoding 100 requests, why can't you send each response on the socket as you process each part?

mrow4a:
As far as I know, it is imposible in the current ownCloud implemetation @PVince81

PVince81:
I'm not sure but we might be bumping against limitations of Sabre DAV. If that's the case then there is the question whether we want to create a separate handler for such requests and if that's even possible with PHP.

PVince81:
I'm not sure but we might be bumping against limitations of Sabre DAV. If that's the case then there is the question whether we want to create a separate handler for such requests and if that's even possible with PHP.

PVince81:
I'm not sure but we might be bumping against limitations of Sabre DAV. If that's the case then there is the question whether we want to create a separate handler for such requests and if that's even possible with PHP.

PVince81:
I'm not sure but we might be bumping against limitations of Sabre DAV. If that's the case then there is the question whether we want to create a separate handler for such requests and if that's even possible with PHP.

PVince81:
I'm not sure but we might be bumping against limitations of Sabre DAV. If that's the case then there is the question whether we want to create a separate handler for such requests and if that's even possible with PHP.

PVince81:
I'm not sure but we might be bumping against limitations of Sabre DAV. If that's the case then there is the question whether we want to create a separate handler for such requests and if that's even possible with PHP.

guruz:
Why do you have to keep everything in memory?
The POST/PUT data you can read in a streaming way and then send a response for each batched file. Your response will also be multipart. The server will send the response out immediately using HTTP chunked encoding.

To wireshark/proxies/etc this will all look like a single request getting a single response.
Also it is a single PHP invocation.
The multipart is _inside_ the request/response, it's opaque to the outside.

guruz:
Why do you have to keep everything in memory?
The POST/PUT data you can read in a streaming way and then send a response for each batched file. Your response will also be multipart. The server will send the response out immediately using HTTP chunked encoding.

To wireshark/proxies/etc this will all look like a single request getting a single response.
Also it is a single PHP invocation.
The multipart is _inside_ the request/response, it's opaque to the outside.

PVince81:
Not sure about PUT and Sabre DAV.

But for POST it is built-in in PHP that it will first store the whole payload in memory and/or a temporary file before reaching it to the PHP code. Basically the PHP code doesn't even start running before the file is completely uploaded to that temp space. Not sure if there is a way around it.

guruz:
@PVince81 It's also a web server configuration issue, e.g. seehttps://github.com/owncloud/core/issues/21338

But I would not worry too much about the memory usage, the bundling is for small files anyway.

mrow4a:
I just want to avoid doing unnecesary reads to memory during parsing, and currently I am forced to do that, cause I had to ensure that everything was correct. Having Multiparts with metadata at the beginign and big data block at the end, I will parse it and evaluate if fine, read from block of data, add to database, emit, get to next read, add to database etc. This will be very lightweight, and with PUT it could be really nice looking. 

I really like the idea about web socket sync, but this is not the reality for PHP i think, as I recall the code I worked with there.

jturcotte:
What do you mean by "ensure that everything was correct"? How is this different than having separate HTTP requests that you can't verify with one-another anyway?

The thing that I'm afraid is that if this tries to be too "smart", it will be more complex and therefore more buggy and difficult to maintain. If you can't keep it simple, then you can't, alright. But we have to do everything we can to make sure that the protocol and the design themselves are as simple as possible to keep the whole thing reliable enough to be enabled everywhere without worries from administrators.

mrow4a:
How is it more complex? It is much less code and operations on the server side, and much clearer than saving something in-memory and then removing on fly. On the client the only change is that you will not add content to the parts and insert only to the last one at specific offsets. On the server side I will also add checksums.

jturcotte:
I never mentioned WebSockets, and by complexity I mean the need to do anything different than if the requests were passed as separate HTTP requests. For example mixing "sync" logic with "transport" logic can add unnecessary coupling. That's assuming that this can be implemented by reading the input stream only once without having to keep everything in memory.

However I don't really know if that's possible, and I won't be able to make my point unless I try to implement it myself, so do what you think is best.

mrow4a:
Hmm, if you want to keep it simple thinking about having separate parts, we would need to add separate multipart part at begining declaring what files to expect, to be able to return responses for the files.

E.g.:
1st part consists of list of files to expect, this will be parsed first - if this fails, whole request fails. 
2nd-Nth part should be parts, and after reading headers it will read contents one by one. 

This will give a clear option of returning single bundled response for each of the files being send, even if parsing of the parts failed for some reason, because you have a list of files to expect in the bundle.

@PVince81 @jturcotte

ogoffart:
Why do you need part 1?  Is it not redundent information from what's in the rest of the parts?

Is it because you want to do early check to report an error early? Is there so much to be gained by erroring early?  The check need to be done again for all the files later anyway.

mrow4a:
Imagine, that you are syncing 500 files. Each file is separate part. At some point, at file 100, you dont have a path to file or content-length or whatever else goes wrong. How do you proceed with parsing, if you dont know what is further, and how to you return an error for that file if you dont know the path for this file? If you have first part specifying the expected files, you have a clear way of defining the response based on that.. read, if you did not parse correctly path headers for file 123,233,444, you know its paths by at the end checking the checklist list from first part.

I know guys that everything sounds easy and straightforward from client perspective, but it is not from server.

ogoffart:
> how to you return an error for that file if you dont know the path for this file? 

If thes server don't include a success reply for a file, the client will know that there was an error for this file.
If there is a parse error for the multi part, you can report success for all the files that succeeded before the parse error, then an error for the rest, the client will understand the error is for all the remaining ones.

> I know you think.

no you don't :-)

mrow4a:
As mentioned above, @DeepDiver1975 did not agreed on that, I proposed that long time ago. If you send 500 HTTP PUTs, do you expect only 300 and you figure out it later? No, you send 500, you receive 500 or 1. I cannot work-around this.

guruz:
You will need a path and a content-length for each batched file. This is a hard fact of the protocol, because of bad webserver support you can't do HTTP chunked encoding for PUT/POST requests, you can only do it for the PUT/POST reply. So the request length must be known.
Else you can't compute the content-length for the whole PUT/POST request. Qt calculates the content-length for the whole request here: https://code.woboq.org/qt5/qtbase/src/network/access/qhttpmultipart.cpp.html#_ZNK22QHttpMultiPartIODevice4sizeEv
(Yes, in theory then the boundary is useless, but that boundary stuff had been defined for E-Mail MIME not HTTP Multipart)

> Now, with multipart/mixed, there has to be compromise, not to keep parsed files in memory or seek in the request body back-and-forth.

I agree that the server should not need to seek back and forth.

It should receive a not-too-big-request on a batch-specific URL, then try to read the first batched request from there. Read a checksum header for this request body and verify it. Then invoke the existing DAV/Storage/Whatever logic in PHP on this. Then formulate the response as a single multi-part-ish response for that first batched request. 
For normal errors (e.g. storage access denied) it will formulate this error as part of the multi-part-ish response for that request.
Then it should attempt to go to the next batched request.
When all batched requests have been received and processed, the server would send a last multi-part-ish "ALL FINE!" response so the sync client knows that its request was OK.
If there is a perceived protocol corruption or body checksum error, the server could just closes the connection. The sync client then should retry those batched requests as unbatched individual requests.
(In my opinion, this can really only happen with a programming error from our side which we will fix or if there is a weird webproxy/firewall/etc mangling with the requests)

So I don't even think there needs to be a big preable to describe each request in the batch. Just send batched request after batched request (with headers, http method etc), encapsulated in multipart.

mrow4a:
@guruz 

> You will need a path and a content-length for each batched file. This is a hard fact of the protocol, because of bad webserver support you can't do HTTP chunked encoding for PUT/POST requests, you can only do it for the PUT/POST reply. So the request length must be known.
Else you can't compute the content-length for the whole PUT/POST request. Qt calculates the content-length for the whole request here: https://code.woboq.org/qt5/qtbase/src/network/access/qhttpmultipart.cpp.html#_ZNK22QHttpMultiPartIODevice4sizeEv
(Yes, in theory then the boundary is useless, but that boundary stuff had been defined for E-Mail MIME not HTTP Multipart)

Boundrary is a part of Multipart request always, not only for email. Thanks to boundrary virtualy you don't need content-length, and you could read the body with fget(), but this is extremely inefficient. 

> 

> It should receive a not-too-big-request on a batch-specific URL, then try to read the first batched request from there. Read a checksum header for this request body and verify it. Then invoke the existing DAV/Storage/Whatever logic in PHP on this. Then formulate the response as a single multi-part-ish response for that first batched request. 
For normal errors (e.g. storage access denied) it will formulate this error as part of the multi-part-ish response for that request.
Then it should attempt to go to the next batched request.
When all batched requests have been received and processed, the server would send a last multi-part-ish "ALL FINE!" response so the sync client knows that its request was OK.
If there is a perceived protocol corruption or body checksum error, the server could just closes the connection. The sync client then should retry those batched requests as unbatched individual requests.
(In my opinion, this can really only happen with a programming error from our side which we will fix or if there is a weird webproxy/firewall/etc mangling with the requests)

It is impossible to implement that in PHP. And still you send 500 files and dont necessairly get 500 responses. Again, what if one of the parts does not have path header? How do you handle that problem? It is impossible not defining it at the begining, otherwise you end up processing 300/500 files and closing connection.

butonic:
@mrow4a Reading through this thread, https://github.com/owncloud/core/pull/25760#issuecomment-240689668 and https://central.owncloud.org/t/files-bundling-extension-multiple-files-update-creation-in-various-locations-on-the-server/1295/29 I think you are trapped in the multipart/related implementation.

With multipart/related you are forced to describe the whole sync. As a result you need to think about additional error conditions that can arise from invalid relations in he request.

I still think multipart/mixed is the right way to go. Every file PUT is contained in a dedicated part. The part can contain the related headers (Content-type, Content-lenght, Content-MD5, X-OC-MTime).

As @DeepDiver1975 pointed out in https://github.com/owncloud/core/pull/25760#issuecomment-238945002 all file uploads should go to the same folder, limiting the batch to a single folder. This limitation can be adressed by adding a [Destination header](http://www.webdav.org/specs/rfc4918.html#HEADER_Destination) to each part.

Regarding the operation / HTTP method several solutions are possible: Use separate bundles for MKCOL, PUT and DELETE in that order. The bundle methot is then the same for all parts. Or allow a bundle to contain multiple operations / HTTP methods by adding a X-OC-Method header. The Bundle operation then MUST be a POST request.

The response should be a HTTP/1.1 207 Multi-Status with a response for every part in the request.

Finally, we could use `Transfer-Encoding: chunked` to send and receive each part / status to the server / client in a dedicated chunk to fail early.

butonic:
in case a part does not contain a path header send a badrequest status for that part in the 207 response but try to continue uploading / handling the other parts. Maybe a user deleted a parent to a certain path. if so all fileuploads to that url will fail. yes. corner case, because it can only happen when the user deleted it between the sync start and the actual upload.

butonic:
if a user specifies garbage with curl all we can do is ignore the garbage. if he eg creates a batch with

1. create /some/foo
2. upload /some/foo/f1.txt
3. upload /some/foo/f2.txt
4. upload /some/foo/f3.txt

but /some never existed. then curl will try to upload the files und transmit them ... but get a 207 response with 4 errors.

why is that a problem? because we may be transferring multiple (in this case 3) files for nothing? It won't happen during normal operations. If you still want to optimize do the folder creation phase in a dedicated batch. Simple graph theory: first MKCOL, (then MOVE and COPY,) then PUT, then DELETE.

mrow4a:
@butonic yes I am also with multipart mixed. The problem which I am currently facing is how do you return a response for the part without the path header? So the solution is to just ignore this part ? We assume that if someone puts the part which is anynomous for the server we just ignore it and let the client figure that out?

butonic:
btw we could allow the server to announce if chunked encoding should be used for batch uploads ... http://sabre.io/dav/0bytes/#webserver-and-chunked-transfer-encoding

mrow4a:
I was only about that, if someone includes 500 parts, and like 3 had no path header(so basicaly anonymousy) we will return 497 bundled responses. I wanted to not allow this situation and if you include 500 parts relating to some files, you will return 500 responses related to files, otherwise 1 BadRequest for whole bundle not even starting adding files. Or have a way to specify that these 3 paths are bad request at least.

jturcotte:
Do you need the path to return an error response? Response could be bound to their request by the order they are sent rather than the path they map to. The client can make sure that the response count matches the requests and raise an error if this condition isn't met to make it a hard requirement.

guruz:
> Boundrary is a part of Multipart request always, not only for email. Thanks to boundrary virtualy you don't need content-length, and you could read the body with fget(), but this is extremely inefficient.

Please see the link I pasted above. Qt needs the content-length of each part and not each webserver supports chunked upload. So just do it like Qt asks you, it's not a downside, it's good to have those restrictions to make the code cleaner.

> Again, what if one of the parts does not have path header?

That could be declared invalid based on the protocol you define, so just let the webserver error out.

butonic:
@mrow4a If a request part contains no path header we can add a response status at the correct position in the 207 response, because multipart/mixed is ordered. even if the part does not have a name / path.

Why can't we use `Content-Disposition: inline; filename="foobar.txt"` when using PUT or
`Content-Disposition: form-data; name="name"; filename="foobar.txt"` when using POST? We coud even use `Content-Disposition: modification-date ...`

mrow4a:
OK: You have the following request in the current implementation:
````
2016-11-07 17:27:14 POST http://localhost:9999/remote.php/dav/files/admin
                                                                                                                                                          
Content-Type:    multipart/mixed; boundary=boundary_3b3ec596a59cadd9bbabd89864e88775                                                                                                                       
content-length:  24277                                                                                                                                                                                     
                                                                                                                                                                                       
--boundary_3b3ec596a59cadd9bbabd89864e88775
X-OC-Path: test/test1.txt
X-OC-Mtime: 1471254375
Content-length: 4
Content-MD5: e2fc714c4727ee9395f324cd2e7f331f

abcd
--boundary_3b3ec596a59cadd9bbabd89864e88775
X-OC-Path:  bundling_tests.sh
X-OC-Mtime: 1471254375
Content-length: 4
Content-MD5: e2fc714c4727ee9395f324cd2e7f331f

abcd
--boundary_3b3ec596a59cadd9bbabd89864e88775
X-OC-Path:  test/zombie1.jpg
X-OC-Mtime: 1471254375
Content-length: 4
Content-MD5: e2fc714c4727ee9395f324cd2e7f331f

abcd
--boundary_3b3ec596a59cadd9bbabd89864e88775--
````

This results in following response:
````
<?xml version='1.0' encoding='UTF-8'?>
<d:multistatus xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns" xmlns:cal="urn:ietf:params:xml:ns:caldav" xmlns:cs="http://calendarserver.org/ns/" xmlns:card="urn:ietf:params:xml:ns:carddav"
xmlns:oc="http://owncloud.org/ns">
  <d:response>
    <d:href>/remote.php/dav/files/admin</d:href>
    <d:propstat>
      <d:prop>
        <d:error>
          <s:exception>Sabre\DAV\Exception\BadRequest</s:exception>
          <s:message>File creation on not existing or without creation permission parent folder is not permitted</s:message>
        </d:error>
        <d:oc-path>test/test.txt</d:oc-path>
      </d:prop>
      <d:status>HTTP/1.1 400 Bad Request</d:status>
    </d:propstat>
  </d:response>
  <d:response>
    <d:href>/remote.php/dav/files/admin</d:href>
    <d:propstat>
      <d:prop>
        <d:x-oc-mtime>accepted</d:x-oc-mtime>
        <d:etag>"d3c3cfe2ec676419f57ff68f210098ce"</d:etag>
        <d:oc-etag>"d3c3cfe2ec676419f57ff68f210098ce"</d:oc-etag>
        <d:oc-fileid>00000029ocs8a4v26ffq</d:oc-fileid>
        <d:oc-path>bundling_tests.sh</d:oc-path>
      </d:prop>
      <d:status>HTTP/1.1 200 OK</d:status>
    </d:propstat>
  </d:response>
  <d:response>
    <d:href>/remote.php/dav/files/admin</d:href>
    <d:propstat>
      <d:prop>
        <d:error>
          <s:exception>Sabre\DAV\Exception\BadRequest</s:exception>
          <s:message>File creation on not existing or without creation permission parent folder is not permitted</s:message>
        </d:error>
        <d:oc-path>test/zombie1.jpg</d:oc-path>
      </d:prop>
      <d:status>HTTP/1.1 400 Bad Request</d:status>
    </d:propstat>
  </d:response>
</d:multistatus>
````

------------------------------------
Now, assume that you have different request like that, so one part has no header:
````
2016-11-07 17:27:14 POST http://localhost:9999/remote.php/dav/files/admin
                                                                                                                                                          
Content-Type:    multipart/mixed; boundary=boundary_3b3ec596a59cadd9bbabd89864e88775                                                                                                                       
content-length:  24277                                                                                                                                                                                     
                                                                                                                                                                                       
--boundary_3b3ec596a59cadd9bbabd89864e88775
X-OC-Path: test/test1.txt
X-OC-Mtime: 1471254375
Content-length: 4
Content-MD5: e2fc714c4727ee9395f324cd2e7f331f

abcd
--boundary_3b3ec596a59cadd9bbabd89864e88775
X-OC-Mtime: 1471254375
Content-length: 4
Content-MD5: e2fc714c4727ee9395f324cd2e7f331f

abcd
--boundary_3b3ec596a59cadd9bbabd89864e88775
X-OC-Path:  test/zombie1.jpg
X-OC-Mtime: 1471254375
Content-length: 4
Content-MD5: e2fc714c4727ee9395f324cd2e7f331f

abcd
--boundary_3b3ec596a59cadd9bbabd89864e88775--
````

In the current implementation, it will return the following response, but this requires parsing whole request before even start processing the files.

````
<?xml version="1.0" encoding="utf-8"?>
<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
  <s:exception>Sabre\DAV\Exception\BadRequest</s:exception>
  <s:message>File header does not contain X-OC-Path. Unable to parse whole bundle request</s:message>
</d:error>
````

What is your suggestion for response guys?

**Comment by jturcotte** [comment]
Just my perception of things, but individual bundled request could have individual <d:multistatus> responses, with one <d:response> each.

And you return them as:

```
--boundary_3b3ec596a59cadd9bbabd89864e88775
X-OC-StatusCode: 200
X-OC-Etc: something

<?xml version="1.0" encoding="utf-8"?>
<d:multistatus xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns" xmlns:cal="urn:ietf:params:xml:ns:caldav" xmlns:cs="http://calendarserver.org/ns/" xmlns:card="urn:ietf:params:xml:ns:carddav"
xmlns:oc="http://owncloud.org/ns">
  <d:response>
</...>

--boundary_3b3ec596a59cadd9bbabd89864e88775
X-OC-StatusCode: 400
X-OC-Etc: something

<?xml version="1.0" encoding="utf-8"?>
<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
  <s:exception>Sabre\DAV\Exception\BadRequest</s:exception>
  <s:message>File header does not contain X-OC-Path. Unable to parse whole bundle request</s:message>
</d:error>
```

This way you just wrap HTTP in multipart, and the whole protocol, the same thing, no difference in payload vs. normal WebDAV since you just encode the HTTP headers and body differently. If your wrapping/unwrapping code works and that we don't get request timing or size issues, everything works as it did before.

**Comment by mrow4a** [comment]
@jturcotte Sounds like a plan, you dont even need status code, because you can have:
````
--boundary_3b3ec596a59cadd9bbabd89864e88775
<d:response>
    <d:href>/remote.php/dav/files/admin</d:href>
    <d:propstat>
      <d:prop>
        <d:x-oc-mtime>accepted</d:x-oc-mtime>
        <d:etag>"d3c3cfe2ec676419f57ff68f210098ce"</d:etag>
        <d:oc-etag>"d3c3cfe2ec676419f57ff68f210098ce"</d:oc-etag>
        <d:oc-fileid>00000029ocs8a4v26ffq</d:oc-fileid>
        <d:oc-path>bundling_tests.sh</d:oc-path>
      </d:prop>
      <d:status>HTTP/1.1 200 OK</d:status>
    </d:propstat>
</d:response>

--boundary_3b3ec596a59cadd9bbabd89864e88775
<d:response>
    <d:href>/remote.php/dav/files/admin</d:href>
    <d:propstat>
      <d:prop>
        <d:error>
          <s:exception>Sabre\DAV\Exception\BadRequest</s:exception>
          <s:message>File creation on not existing or without creation permission parent folder is not permitted</s:message>
        </d:error>
        <d:oc-path>test/zombie1.jpg</d:oc-path>
      </d:prop>
      <d:status>HTTP/1.1 400 Bad Request</d:status>
    </d:propstat>
</d:response>

--boundary_3b3ec596a59cadd9bbabd89864e88775
<?xml version="1.0" encoding="utf-8"?>
<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
  <s:exception>Sabre\DAV\Exception\BadRequest</s:exception>
  <s:message>File header does not contain X-OC-Path. Unable to parse whole bundle request</s:message>
</d:error>

--boundary_3b3ec596a59cadd9bbabd89864e88775--
````

@butonic @guruz @ogoffart ?

**Comment by butonic** [comment]
It does not seem to be explicitly forbidden to send a body thet does not contain a multistatus element. From https://tools.ietf.org/html/rfc4918#section-13:

> A Multi-Status response conveys information about multiple resources
   in situations where multiple status codes might be appropriate.  The
   default Multi-Status response body is a text/xml or application/xml
   HTTP entity with a 'multistatus' root element. [...]

I'd say go for it. Semantically, it makes a lot of sense to use the parts of a multipart/mixed response similar to a multistatus element.

A different approach would be using chunked transfer encoding and sending the multistatus opening tag in the first chunk, followed by multiple response tags - each in its own chunk - and finally a chunk containing the closing tag. This would allow us to send results as we have them, without having to preparing the complete response to calculate the response content-length.

**Comment by DeepDiver1975** [comment]
> @jturcotte Sounds like a plan, and it is no longer multistatus then,

please keep the multistatus response

**Comment by guruz** [comment]
> it will return the following response, but this requires parsing whole request before even start processing the files.

Instead you could do like https://github.com/owncloud/client/pull/5155#issuecomment-258805609 and kill the connection or so :-) But that requires having some kind of termination of the XML. Although XML is anyway terminated.
Or you could do like @butonic proposed and just go by the natural ordering of your requests, If request n is bogus then you just write that in the nth response and then continue normally.

About the response format:
@butonic @mrow4a You can just write out data there, the webserver will automatically decide if he wants to buffer it and send it with content-length or he will use chunked encoding for the response. You don't even need to think where the webserver does the chunks. Form the client side (Qt) they are anyway hidden and the response appears as a stream/array.

(I don't mind if it's a single XML response or the multipart thing. Much easier to implement as a single XML response, so maybe go for that. Even though the protocol feels less symmetric then)

**Comment by DeepDiver1975** [comment]
> Even though the protocol feels less symmetric then

Not really - we might even accept different request body types one day - but the response will stay constant. At least this is my view on this.

**Comment by mrow4a** [comment]
I had quite long talk with @DeepDiver1975 from the oC architectural point of view, and he gave the implementation freedom to these feature, but only while the following conditions are fullfiled, without work-arounding this (this is actualy what was estabilished writing specification for bundling at the begining with Thomas and Klaas):

1. Request response should be 207 Multistatus in case of successfull request, including individual statuses for files in the bundle or 4xx-5xx in case of the error in the request

2. Request body can be any mime-type, with full implementation freedom. 

3. Request is finishing with successfull response after all linked operations has been successfuly finished, or aborted immedietaly with single error message. 

4. If request cannot be executed or request is wrongly constructed, request has to be aborted and error 4xx-5xx has to be returned for whole the request. 

5. Error and success messages for files should be included in the multistatus response, for each file included in the request.

We already had implemented both multipart/mixed and multipart/related, discussed it a lot and tried out:

I found following limitations for each of the request, starting with the order of implementation:

Multipart/related:
- This type of mime type includes in the first part the list of files to be created, in the key->value manner, where key is the path and value is metadata for that file. Response is created based on the keys in the metadata part, and files are reconstructed from binary contents in the request body, referenced by Content-ID
- This mime-type allows to to easily return response for the file, because key-value structure and validation at the begining allows you to correctly construct the response for each of the keys-files(even if content-id is missing, you can return the response for the specific file that binary content for that file has not been found) or return parsing error at the begining. 
- This is high efficiency solution, where files are being added to OC as they are being read from the request body and allows you to use chunked transfer encoding for the response while files are being added.
- Disadvantage is that list of files is specified in the first part, and the binary contents are anynomous reading the request without first part.
- Disadvantage is that, in case of parsing error, request has occupied bandwidth for nothing - this should however not happen in practice.

Multipart/mixed:
- This mime types includes in each part headers (metadata for a file), and in the part body, the actual file body. 
- As in multipart/related, we can use chunked transfer encoding for the response.
- Advantage here is that reading pure request, we have an independent parts in the request. 
- This mime-type requires you to parse the request body on fly in order to read the body of the request and simultaneously add it to OC. In order to correctly construct the response for each of the files, each part has to be parsed and validated, since it serves as a container for the file.
- In this mime-type, in order to validate the request at the beginning, one would need to parse the request body and save it in-memory or on the disk. The other option is to seek in the request body, however this is dangerous and unpredictable. This results in the fact, that the bundle has to occupy in the peak moment the memory equal to chunk size - typicaly 10MB for request itself and 10MB for in-memory  storage, and will decrease with each file added to OC. This gives 20MB per request, and 20GB per 1000 simultaneous request in peak moment. 
- In this mime-type in one need to validate and parse independent parts on fly. In case of single part parsing error, one would need to raise the full request error, since the construction for that file might not be possible due to lacking URL header for that file, which will result in lack of the response of the previous files already added in the OC for that bundle. Having half of the files acknowledged for that request, and afterwards return error for the whole request invalidates the architectural concept for request in OC (refer to point 3). 
- Disadvantage is that, in case of parsing error, request has occupied bandwidth for nothing - this should however not happen in practice.

This is the final overview of this feature, and the implementation decision has to be taken. RFC - Request For Comments @ogoffart @butonic @jturcotte @guruz, on which trade-off should we decide on the client. 

Bottom note from my side:
I am ok with any type of implementation, however I see no sense in bundle witch adds additional operations and consumes a lot of resources in high-scale systems. By design, it should reduce resources and number of operation/processes per file. Please refer to description at: https://conference.owncloud.org/conference/oCC2016/program/proposal/180

**Comment by mrow4a** [comment]
I forgot to add something that @jturcotte mentioned one day, that in case of multipart/related the metadata should be the same type as response, so XML. The same way as PROPFIND is being done.

**Comment by DeepDiver1975** [comment]
> however I see no sense in bundle which adds additional operations and consumes a lot of resources in high-scale systems.

The benefit is already within reducing requests and responses.
Finally to note: we really should analyze how http/2 pipelining compares to this. To me both approaches are mechanisms to speedup file operations.
If we get almost the same benefit by just switching to http/2 - I see no need in investing into our own implementations.

**Comment by mrow4a** [comment]
I definitely agree. However, I will also look into combining both http/2 and bundling. Http/2 is not a server feature and requires proxy in front of the server. I am looking that up.

**Comment by butonic** [comment]
both request types (mimepart/mixed and mimepart/related) need to be parsed to fulfill Requirement 3. Neither can start storing any received data because a later part might be malformed. In that case the whole request needs to be aborted with a single error message.

I think rquirement 3 contradicts requirement 1. 207 I there to allow multiple response statuses. Requirement 1 will lead to all response statuses being 200 OK or a single 40x response.

Take a step back and look at our curren sync which uses individual request. It also continues if a file could not be uploaded or a folder could not be created. In the latter case it can also skip syncing the files contained in the folder, which the multipart/related request cannot do. It is all or nothing.

If we remove Requirement 3 we can use chunked encoding for the multipart/mixed POST request and a 207 multi-status response. After sending the initial header the client can send a chunk containing a complete part (I would end every chunk with the boundary). The server could also use chunked encoding and immediately send the 207 response with headers and the
```
<?xml version='1.0' encoding='UTF-8'?>
<d:multistatus xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns" xmlns:cal="urn:ietf:params:xml:ns:caldav" xmlns:cs="http://calendarserver.org/ns/" xmlns:card="urn:ietf:params:xml:ns:carddav"
xmlns:oc="http://owncloud.org/ns">
```
Start as the first chunk. Immediately after processing a request chunk containing a part he can then send the corresponding `<response>...</response>` element.

@DeepDiver1975 @mrow4a where does requirement come from. It kind of defeats the purpose of a 207 response.

mrow4a:
> both request types (mimepart/mixed and mimepart/related) need to be parsed to fulfill Requirement 3. Neither can start storing any received data because a later part might be malformed. In that case the whole request needs to be aborted with a single error message.

@butonic In case of multipart/related, you are able to create response for each of the files. You specify that for specific file, content is not existing and you are done, go to next part. This could be a case of corrupted header for example. In case of multipart/mixed you cannot ceate the correct response if something is wrong with your headers.

> I think rquirement 3 contradicts requirement 1. 207 I there to allow multiple response statuses. Requirement 1 will lead to all response statuses being 200 OK or a single 40x response.

I dont see any contradiction there. 207 is just requirement for bundled request, you are not expecting 200. By all linked operations I meant that you are not throwing exception somewhere in the middle for whole request.

butonic:
@mrow4a a multipart/related request may very well get corrupted at a boundary. 

> Request is finishing with successfull response after all linked operations has been successfuly finished, or aborted immedietaly with single error message.

This says to abort *immediately* unless *all* linked operations have been *successfully* finished. As in a 207 multi-stats with all ok responses or a single error status.

If that is not the requirement, please rephrase. Maybe:

A bundle Request receives a 207 Multi-Status Response with the individual 20x, 30x, 40x, 50x statuses. It may receive a 400 Bad Request response with an error message if the Request was malformed

I don't see a problem with the multipart/mixed request when a part has invalid or missing headers. The response XML can still contain a response tag... ah now I understand part of the problem. From the RFC:

>  Each 'response' element MUST have an 'href' element to identify the resource.

Thats obviously not possible if the path is missing from the part. 

But how do we treat that case in multipart/related: an extra part may be present that has no matching entry in the initial description. Do we just ignore it? do we have to respond with a single 400 BadResponse?

guruz:
> If we get almost the same benefit by just switching to http/2 - I see no need in investing into our own implementations.

As far as I know this was already benchmarked some time ago and the answer is that bundling still helps since oC/PHP has a big per-request overhead which will still occur with HTTP2 since that is only about avoiding TCP/SSL connection establishing and saving some bandwidth for headers.

So what @mrow4a is doing here with bundling is great. I like to see this getting in, especially given how far the code is already.

jturcotte:
One thing I think is a bit off focus is that I don't think that parsing errors or malformed request is something we have to optimize. Aborting the whole thing with a 500 and a cryptic error will be enough to get this ready for production; if there is any way that we think this will happen in practice it means that the code isn't ready to ship.

What matters more I think is what happens if:
- The connection is interrupted in the middle of getting the request
- The request contains an non-existant/already deleted path
- A 500 error happens in the middle of the processing
- The connection is interrupted in the middle of sending back the response of one of the bundled request

In all of those cases, the client should be able as much as possible to parse the response and know exactly what was successfully processed, what failed, and what wasn't even touched. And I don't see those mentioned in the requirements, and even conflict with a few of them (1, 3, 4 and 5 all assume that the feedback is all or nothing, while some of it might have succeeded before the PHP or transmission error occurred).

mrow4a:
Obsolete with https://github.com/owncloud/client/pull/5319

PR #5160, opened 2016-09-07
Wizard: Do the system proxy lookup in a thread

guruz:
This is analogous to the code in ConnectionValidator. Otherwise Qt did it internally and blocked the UI.

I had this pending in a branch for another issue.

@danimo Sorry that this might collide with your new wizard code.

### DIFF
--- a/src/gui/owncloudsetupwizard.cpp
+++ b/src/gui/owncloudsetupwizard.cpp
@@ -30,6 +30,7 @@
 #include "networkjobs.h"
 #include "sslerrordialog.h"
 #include "accountmanager.h"
+#include "clientproxy.h"
 
 #include "creds/credentialsfactory.h"
 #include "creds/abstractcredentials.h"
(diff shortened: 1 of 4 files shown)


src/gui/owncloudsetupwizard.cpp
```diff
@@ -128,15 +129,46 @@ void OwncloudSetupWizard::slotDetermineAuthType(const QString &urlString)
...
+    if (ClientProxy::isUsingSystemDefault()) {
+        qDebug() << "Trying to look up system proxy";
+        ClientProxy::lookupSystemProxyAsync(account->url(),
+                                            this, SLOT(slotSystemProxyLookupDone(QNetworkProxy)));
+    } else {
+        // We want to reset the QNAM proxy so that the global proxy settings are used (via ClientProxy settings)
+        account->networkAccessManager()->setProxy(QNetworkProxy(QNetworkProxy::DefaultProxy));
+        // use a queued invocation so we're as asynchronous as with the other code path
+        QMetaObject::invokeMethod(this, "slotContinueDetermineAuth", Qt::QueuedConnection);
+    }
```
ckamm: Optional: It'd be nice if this code wasn't duplicated with connectionvalidator.cpp somehow. A function that takes the two target slots? A minimal QObject that has two signals to connect to? It's still significant overhead for what is essentially four lines of code, but it might be worthwhile.

ckamm:
@guruz Basically good, one code reuse question.

src/gui/owncloudsetupwizard.cpp
```diff
@@ -128,15 +129,46 @@ void OwncloudSetupWizard::slotDetermineAuthType(const QString &urlString)
...
+
     // Set fake credentials before we check what credential it actually is.
     account->setCredentials(CredentialsFactory::create("dummy"));
     CheckServerJob *job = new CheckServerJob(_ocWizard->account(), this);
     job->setIgnoreCredentialFailure(true);
     connect(job, SIGNAL(instanceFound(QUrl,QVariantMap)), SLOT(slotOwnCloudFoundAuth(QUrl,QVariantMap)));
     connect(job, SIGNAL(instanceNotFound(QNetworkReply*)), SLOT(slotNoOwnCloudFoundAuth(QNetworkReply*)));
     connect(job, SIGNAL(timeout(const QUrl&)), SLOT(slotNoOwnCloudFoundAuthTimeout(const QUrl&)));
-    job->setTimeout(10*1000);
+    job->setTimeout((account->url().scheme() == "https") ? 30*1000 : 10*1000);
```
ogoffart: Why different timeout values?

src/gui/owncloudsetupwizard.cpp
```diff
@@ -128,15 +129,46 @@ void OwncloudSetupWizard::slotDetermineAuthType(const QString &urlString)
...
+        account->networkAccessManager()->setProxy(QNetworkProxy(QNetworkProxy::DefaultProxy));
+        // use a queued invocation so we're as asynchronous as with the other code path
+        QMetaObject::invokeMethod(this, "slotContinueDetermineAuth", Qt::QueuedConnection);
+    }
+}
+
+void OwncloudSetupWizard::slotSystemProxyLookupDone(const QNetworkProxy &proxy)
+{
+    if (proxy.type() != QNetworkProxy::NoProxy) {
+        qDebug() << "Setting QNAM proxy to be system proxy" << printQNetworkProxy(proxy);
```
ogoffart: CMakeFiles/owncloud.dir/objects.a(owncloudsetupwizard.cpp.obj):owncloudsetupwizard.cpp:(.text+0x165e): undefined reference to `OCC::printQNetworkProxy(QNetworkProxy const&)'



The OWNCLOUDSYNC_EXPORT macro is missing for this function.

src/gui/owncloudsetupwizard.cpp
guruz: For SSL one might need to show a cert dialog.
For unencrypted, we'd want the timeout to be as fast as possible in case the user miss-typed an IP (and it gets host unreachable etc..)

ckamm:
@guruz Do you still need more review?

PR #5170, opened 2016-09-12
add server capabilities to owncloudcmd

mrow4a:
@owncloud/desktop-developers

### DIFF
--- a/src/cmd/cmd.cpp
+++ b/src/cmd/cmd.cpp
@@ -29,6 +29,7 @@
 #include "syncengine.h"
 #include "syncjournaldb.h"
 #include "config.h"
+#include "connectionvalidator.h"
 
 #include "cmd.h"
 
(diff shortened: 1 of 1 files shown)


dragotin:
please at least ifdef the C++ 11 stuff to keep it compileable with older systems and/or Qt4

src/cmd/cmd.cpp
```diff
@@ -68,6 +69,10 @@ struct CmdOptions {
 // So we have to use a global variable
 CmdOptions *opts = 0;
 
+// Make sure the timeout for this job is less than how often we get called
+// This makes sure we get tried often enough without "ConnectionValidator already running"
+static qint64 timeoutToUseMsec = qMax(1000, ConnectionValidator::DefaultCallingIntervalMsec - 5*1000);
```
ogoffart: should be const

(to avoid runtime initialisation)

src/cmd/cmd.cpp
```diff
@@ -68,6 +69,10 @@ struct CmdOptions {
 // So we have to use a global variable
 CmdOptions *opts = 0;
 
+// Make sure the timeout for this job is less than how often we get called
```
ogoffart: this comment does not make sens here.

ogoffart:
Looks good, thanks   :+1: :

guruz:
👍 Please merge

guruz:
Next time format the message like:
```owncloudcmd: Add server capability check```

rullzer:
Ah that would be usefull when debugging indeed. :+1:

PR #5172, opened 2016-09-12
Dolphin Plugin: Use the Application name for the socket path

### LINKED ISSUE #5165: [2.2.3] [Overlays] dolphin plugin fails with branding
As can be seen in: 
https://github.com/owncloud/client/blob/master/shell_integration/dolphin/ownclouddolphinpluginhelper.cpp#L70
the socket is hardcoded to
`ownCloud/socket`
in the XDG-path which naturally will fail if the APPLICATION_SHORTNAME is set via OEM.cmake. 

### Expected behaviour
Overlay icons and integration works with dolphin. 

### Actual behaviour
If owncloud-client and branded client installed, overlay icons are duplicated on owncloud folder, branded directory has no overlays. 

### Client configuration
Client version: 2.2.3

Operating system: Gentoo Linux

dragotin:
do not hardcode owncloud here.

This should fix #5165

### DIFF
--- a/shell_integration/dolphin/ownclouddolphinpluginhelper.cpp
+++ b/shell_integration/dolphin/ownclouddolphinpluginhelper.cpp
@@ -21,6 +21,7 @@
 #include <qcoreevent.h>
 #include <QFile>
 #include "ownclouddolphinpluginhelper.h"
+#include "config.h"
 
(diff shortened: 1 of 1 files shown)


dragotin:
maybe candidate for 2.2 branch?

ogoffart:
:+1: :

Issue #5163, opened 2016-09-09
File Watcher: sync is triggered when files/directories under hidden directories are edited.


pfoo:
### Expected behaviour

owncloud client should not trigger a sync check when a hidden directory or any of its subfiles/subdirectories is modified when "sync hidden files" is unchecked.
### Actual behaviour

owncloud client seems to be triggering a sync check when a directory inside a hidden directory is modified even if "sync hidden files" is unchecked.
This is not happenning when adding ".*" to sync-exclude.lst
### Steps to reproduce
1. sync your home folder (~), uncheck sync hidden files, do not add .\* to sync-exclude.lst
2. launch firefox
3. wait for firefox to change something in its profile (launching firefox itself does, creating new tabs also works)

This is also triggering it :
touch .local/share/klipper/test
touch .local/share/test

this is however not triggering it :
touch .local/test
### Client configuration

Client version: 2.2.3

Operating system: linux ubuntu 16.04 4.4.0-36-generic
### Logs

```
09-09 14:51:38:159 15058 OCC::FolderWatcher::changeDetected: detected changes in paths: QSet("/home/pfoo/.mozilla/firefox/1mtnhnyr.default/xulstore.json.tmp")
09-09 14:51:38:160 15058 OCC::FolderMan::slotScheduleSync: Schedule folder  "pfoo"  to sync!
[...]
09-09 15:13:29:857 16490 OCC::FolderWatcher::changeDetected: detected changes in paths: QSet("/home/pfoo/.local/share/klipper/history2.lst.TJ3570")
09-09 15:13:29:858 16490 OCC::FolderMan::slotScheduleSync: Schedule folder  "pfoo"  to sync!
```

ckamm:
I can reproduce the problem.

ckamm:
Pull request: https://github.com/owncloud/client/pull/5173

SamuAlfageme:
@ckamm Seems to be fixed (tried in OS X, Ubuntu and Windows with OC Client 2.3.0~nightly20161111) and it's not happening anymore. Although I found a minor bug somehow related with windows special hidden files I've logged in #5309 and a small detail that came up while selecting `$HOME` as local dir to sync, the relative path is cropped and shows `Open folder ''` under Managed Folders, instead of showing 'Open folder /User/<username>`:

![relative path os x](https://cloud.githubusercontent.com/assets/2644445/20211375/86ab5166-a7fe-11e6-9034-b47b75bb9da6.png)

[...]

PR #5173, opened 2016-09-12
Excludes: Consider files in hidden folders excluded #5163

ckamm:
Previously, we only checked the hiddenness of the target file and
ignored the hiddenness of the containing folders. This lead to
undesired behavior when people synced their home folders and there
was a folder watcher notification for a non-hidden file in one of
the hidden folders.

I'm not fully sure why, but sometimes notifications for .foo/bar were
already ignored, but notifications for .foo/bar/car were not. This may
be because of how we set up the FolderWatchers on Linux.

The new behavior is to check all path components for hiddenness, up
until the base path (but excluding the base path, so using a hidden
folder as a sync folder will work).

### DIFF
--- a/src/libsync/excludedfiles.cpp
+++ b/src/libsync/excludedfiles.cpp
@@ -76,13 +76,22 @@ bool ExcludedFiles::isExcluded(
         return true;
     }
 
-    QFileInfo fi(filePath);
     if( excludeHidden ) {
-        if( fi.isHidden() || fi.fileName().startsWith(QLatin1Char('.')) ) {
-            return true;
+        QString path = filePath;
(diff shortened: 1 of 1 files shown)


ogoffart:
:+1:

ckamm:
@jturcotte @guruz If you also have feedback on this...

ogoffart:
:1:

dragotin:
The PR is great, yet I do not understand why it is needed. I thought this method is always called recursively, which would mean that a file `foo/.bar/car` would be ignored because `.bar` was ignored before the function sees the `car` file? No?

ogoffart:
it s needed because for the socket api, it is not called recrsively.

PR #5174, opened 2016-09-12
Better fix for #5116

ogoffart:


guruz:
@ogoffart I can confirm this still fixes the bug

ogoffart:
Updated.   Now we will not change the config.

ckamm:
:+1:

Issue #5164, opened 2016-09-09
[Xubuntu 16.04.1] [2.3] Systray icon disappear briefly when a file or icon changes state, tray menu opens and dissapears instantly


babs:
### Expected behaviour
Systray icon to stay steady and be right clickable to open owncloud's menu

### Actual behaviour (on the packaged version)
Systray icon disappear briefly when it changes state making the whole systray to flickers.
Icon menu opens as expected.

### Actual behaviour (on the master commit 524220d0907713408298a36df18bd04635ce5563)
Systray icon stay steady as expected but unable to has access to right click menu.
Whe the icon is clicked, the menu show up an instant and disappear instantaneously.

### Steps to reproduce
1. install the client on a 64bit xunbutu
2. update or touch a file in a monitored directory 

### Client configuration
Client version: ownCloud version 2.2.3 Using Qt 5.5.1

Operating system: Xubuntu 16.04.1 LTS amd64

Window manager: Xfce 4.12

OS language: fr_FR.UTF-8

Installation path of client: packaged version or run from source dir

ckamm:
@babs Thanks for the nice report. The difference on master is probably due to 88cd5421bfcfafa9f4aaa46cff6166dfce1f516d, which intended to work around flickering and unreasonable resource usage by updating the tray menu only on demand. Apparently that somehow breaks with your setup :/

ckamm:
A fix to this would also address or change #5158

babs:
Ok, tag me if you want me to pull/compile/check anything on my setup.

ckamm:
@babs I assume you do see the string "Enabled QDBusPlatformMenu workaround" in your log output? (--logfile)

babs:
@ckamm yes, on both versions.

ckamm:
@babs Are you up to compiling the client from source to test something?

babs:
@ckamm Of course, give me a branch or commit :)

ckamm:
@babs at owncloudgui.cpp:477 in the master branch, comment out `_qdbusmenuWorkaround = true;` - does it work then?

Can you still click the tray menu to show the settings dialog? Does it still update when you add an account?

babs:
Menu is available and doesn't even close when icon change.
No flicker

guruz:
@nazar-pc has the same symptom with the nightlies..
https://github.com/owncloud/client/issues/4985#issuecomment-246185323
And also memory growing? (because of menu changing constantly?)

[...]

PR #5179, opened 2016-09-13
Tray: Workaround collection

ckamm:
* QDBus workaround for Qt 5.5.0 only, there were reports of the tray
  working fine with 5.5.1. #5164
* OWNCLOUD_FORCE_QDBUS_TRAY_WORKAROUND to force the workaround on an off
* OWNCLOUD_TRAY_UPDATE_WHILE_VISIBLE to enable or disable updating of
  the menu while it's visible
* Track the visibility of the tray menu with aboutToShow/aboutToHide
  only on OSX - the aboutToHide signal doesn't trigger reliably on linux
* Refactor such that setupContextMenu is different from updateContextMenu
* Don't use on-demand updating of the tray menu when the qdbus workaround
  is active, instead to occasional (30s) updates of the tray menu.

### DIFF
--- a/src/gui/owncloudgui.cpp
+++ b/src/gui/owncloudgui.cpp
@@ -56,7 +56,7 @@ ownCloudGui::ownCloudGui(Application *parent) :
     _settingsDialog(new SettingsDialog(this)),
 #endif
     _logBrowser(0),
-    _contextMenuVisible(false),
+    _contextMenuVisibleOsx(false),
     _recentActionsMenu(0),
     _qdbusmenuWorkaround(false),
     _folderOpenActionMapper(new QSignalMapper(this)),
@@ -93,9 +93,9 @@ ownCloudGui::ownCloudGui(Application *parent) :
              this,SLOT(slotSyncStateChange(Folder*)));
 
     connect( AccountManager::instance(), SIGNAL(accountAdded(AccountState*)),
-             SLOT(setupContextMenuIfVisible()));
+             SLOT(updateContextMenuNeeded()));
     connect( AccountManager::instance(), SIGNAL(accountRemoved(AccountState*)),
-             SLOT(setupContextMenuIfVisible()));
+             SLOT(updateContextMenuNeeded()));
 
     connect( Logger::instance(), SIGNAL(guiLog(QString,QString)),
              SLOT(slotShowTrayMessage(QString,QString)));
@@ -194,7 +194,7 @@ void ownCloudGui::slotTrayClicked( QSystemTrayIcon::ActivationReason reason )
 void ownCloudGui::slotSyncStateChange( Folder* folder )
 {
     slotComputeOverallSyncStatus();
-    setupContextMenuIfVisible();
+    updateContextMenuNeeded();
 
     if( !folder ) {
         return; // Valid, just a general GUI redraw was needed.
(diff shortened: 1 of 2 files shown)


src/gui/owncloudgui.cpp
```diff
@@ -581,17 +651,30 @@ void ownCloudGui::setupContextMenu()
...
+        if (!_workaroundBatchTrayUpdate.isActive()) {
+            _workaroundBatchTrayUpdate.start();
+        }
+        return;
+    }
+
 #ifdef Q_OS_MAC
     // https://bugreports.qt.io/browse/QTBUG-54845
+    // We cannot update on demand or while visible -> update when invisible.
     if (!_contextMenuVisible) {
```
guruz: _contextMenuVisibleOsx

src/gui/owncloudgui.cpp
```diff
@@ -400,41 +400,159 @@ void ownCloudGui::addAccountContextMenu(AccountStatePtr accountState, QMenu *men
...
+#endif
+    }
+
+    // Populate the context menu now.
+    updateContextMenu();
+}
+
+void ownCloudGui::updateContextMenu()
+{
+    qDebug() << "ZZZ CONTEXT";
```
guruz: ^^

src/gui/owncloudgui.cpp
guruz: Tested patch on OS X, compiles and works fine.

src/gui/owncloudgui.cpp
```diff
@@ -400,41 +400,159 @@ void ownCloudGui::addAccountContextMenu(AccountStatePtr accountState, QMenu *men
...
+    // In particular QTBUG-47863 #3672 (tray menu fails to update and
+    // becomes unresponsive) and QTBUG-48068 #3722 (click signal is
+    // emitted several times)
+    // The Qt version check intentionally uses 5.0.0 (where platformMenu()
+    // was introduced) instead of 5.5.0 to avoid issues where the Qt
+    // version used to build is different from the one used at runtime.
+    // If we build with 5.6.1 or newer, we can skip this because the
+    // bugs should be fixed there.
+#ifdef Q_OS_LINUX
+#if (QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)) && (QT_VERSION < QT_VERSION_CHECK(5, 5, 1))
```
ogoffart: Technicly, one can compile an app against 5.5.1 and run it against 5.5.0.  so i guess this should be < 5.6

src/gui/owncloudgui.cpp
```diff
@@ -400,41 +400,159 @@ void ownCloudGui::addAccountContextMenu(AccountStatePtr accountState, QMenu *men
...
+    if (qVersion() == QByteArray("5.5.0")) {
+        QObject* platformMenu = reinterpret_cast<QObject*>(_tray->contextMenu()->platformMenu());
+        if (platformMenu
+                && platformMenu->metaObject()->className() == QLatin1String("QDBusPlatformMenu")
+                && !disableQDBusTrayWorkaround()) {
```
ckamm: This function was removed.

ckamm:
@guruz @ogoffart Addressed your comments.

guruz:
I can confirm it still seems to work on OS X.

guruz:
👍 let's merge it after @ogoffart agrees too.
Did someone test on Windows?

ogoffart:
Tested on linux with the xembed protocol.  Works but changing the menu while it is visible causes flickering of the menu.  I don't know if it is such a good idea.

ckamm:
@ogoffart So you're saying we should disable update-while-visible altogether? It'd be easy to test, just set OWNCLOUD_TRAY_UPDATE_WHILE_VISIBLE=0.

PR #5182, opened 2016-09-14
Fix the MSVC build

jturcotte:
- Replace functions that are provided by MinGW with a Win32-based
  implementation
- Explicitly export needed symbols from ocsync.dll
- Rename share.h to sharemanager.h since the name clashes with one
  of the Windows headers and get included from there
- Remove the timestamp from the fallback csync stderr logging, it's
  not used since we always provide a log callback

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,9 @@
 
 cmake_minimum_required(VERSION 2.6)
 cmake_policy(VERSION 2.8.0)
+if(POLICY CMP0020)
+  cmake_policy(SET CMP0020 NEW)
+endif()
 
 project(client)
 
@@ -57,7 +60,7 @@ endif()
 
 include(GNUInstallDirs)
 include(DefineInstallationPaths)
-
+include(GenerateExportHeader)
 
 include(GetGitRevisionDescription)
 
@@ -166,6 +169,9 @@ find_package(SQLite3 3.8.0 REQUIRED)
 # On some OS, we want to use our own, not the system sqlite
 if (USE_OUR_OWN_SQLITE3)
     include_directories(BEFORE ${SQLITE3_INCLUDE_DIR})
+    if (WIN32)
+        add_definitions(-DSQLITE_API=__declspec\(dllimport\))
+    endif()
 endif()
 
 find_package(ZLIB)
@@ -180,6 +186,7 @@ add_definitions(-DUNICODE)
 add_definitions(-D_UNICODE)
 if( WIN32 )
 add_definitions( -D__USE_MINGW_ANSI_STDIO=1 )
+add_definitions( -DNOMINMAX )
 endif( WIN32 )
 
 # Handle Translations, pick all client_* files from trans directory.
--- a/cmake/modules/Warnings.cmake
+++ b/cmake/modules/Warnings.cmake
@@ -2,21 +2,23 @@
 # Redistribution and use is allowed according to the terms of the BSD license.
 # For details see the accompanying COPYING* file.
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -pedantic -Wno-long-long")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++0x")
+if (CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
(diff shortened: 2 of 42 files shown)


CMakeLists.txt
```diff
@@ -166,6 +169,9 @@ find_package(SQLite3 3.8.0 REQUIRED)
 # On some OS, we want to use our own, not the system sqlite
 if (USE_OUR_OWN_SQLITE3)
     include_directories(BEFORE ${SQLITE3_INCLUDE_DIR})
+    if (WIN32)
```
ogoffart: That's both for MSVC and mingw,  right?

cmake/modules/Warnings.cmake
```diff
@@ -2,21 +2,23 @@
 # Redistribution and use is allowed according to the terms of the BSD license.
 # For details see the accompanying COPYING* file.
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -pedantic -Wno-long-long")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++0x")
+if (CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -pedantic -Wno-long-long")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++0x")
 
-if(CMAKE_COMPILER_IS_GNUCXX)
```
ogoffart: oh, was that not more correct?

csync/src/csync_exclude.c
```diff
@@ -102,7 +107,7 @@ int csync_exclude_load(const char *fname, c_strlist_t **list) {
   int fd = -1;
   int i = 0;
   int rc = -1;
-  int64_t size;
+  off_t size;
```
ogoffart: We have had problems with off_t before  (see commit 4f5bfbdb784fb4c589ff04c439fb82749b3b54e5 )

csync/src/csync_log.c
```diff
@@ -37,46 +31,11 @@ CSYNC_THREAD int csync_log_level;
...
-    int rc;
-
-    rc = current_timestring(1, date, sizeof(date));
-    if (rc == 0) {
-        fprintf(stderr, "[%s, %d] %s:", date+5, verbosity, function);
-    } else {
-        fprintf(stderr, "[%d] %s", verbosity, function);
-    }
-
+    fprintf(stderr, "[%d] %s", verbosity, function);
```
ogoffart: Yes we don"t really need the date...
Then the two fprintf can be merged.

csync/src/std/asprintf.c
```diff
@@ -0,0 +1,88 @@
+/*
```
ogoffart: Should the file go in 3rdparty?
Should there be a URL from the readme on where this came from, and which revision. ?

CMakeLists.txt
jturcotte: Yes, same for _WIN32 in the code, it applies to both MSVC and MinGW

cmake/modules/Warnings.cmake
jturcotte: They just recommend to use the compiler ID if we want to target both Clang and GCC, my last commit to that file was getting stuff out of that block for Clang, but now I need to keep MSVC out of it.

csync/src/csync_exclude.c
jturcotte: OK that one was only a warning so I'll remove it.

csync/src/std/asprintf.c
jturcotte: It's unlikely we'll update that code ever again so I think it's not worth it to start referencing files across the filesystem to 3rdparty in the cmake files.

Where should I put the URL, in the file itself?

csync/src/csync_exclude.c
jturcotte: done

csync/src/std/asprintf.c
jturcotte: I've put the URL in the file headers.

jturcotte:
Please look at the commit message for the reason why renaming share.h was necessary.

ogoffart:
:+1:

Issue #5180, opened 2016-09-13
Licence incompatibilities


XRevan86:
src/3rdparty/QProgressIndicator/ is licensed under LGPLv3+ which [cannot mix](https://gnu.org/licenses/gpl-faq.en.html#AllCompatibility) with GPLv2 (different from GPLv2+).

There are multiple GPLv2 licensed files in src/cmd/, src/libsync/, src/gui/ and src/3rdparty/certificates/,
you can see them when using `grep -R "version 2 of the License\."` in the tree root (dot at the end means there's no "`, or (at your option) any later version.`" afterwards).

GPLv2 is also a problem for src/3rdparty/libcrashreporter-qt/ which is GPLv3+ licensed.

michaelstingl:
@jnweiger @hodyroff ?

jnweiger:
This was not intentional. 
I suggest to correct the situation by  broadening the LGPL-V2 license declaration to always say LGPL-V2+ instead.

hodyroff:
Approved. Need to add to the dependency list. libcrashreporter-qt is only used in the Windows built and if needed could be taken out. Any other GPL code included needs special approval.

PR #5184, opened 2016-09-14
Update copy of QProgressIndicator.cpp in 3rdparty.

dragotin:
It was relicensed to MIT according to
https://github.com/mojocorp/QProgressIndicator/commit/14bb9d10e29407175fdf4b9f03ed9cc7bb124f93

This is part of the fix for #5180

### DIFF
--- a/src/3rdparty/QProgressIndicator/LICENSE
+++ b/src/3rdparty/QProgressIndicator/LICENSE
@@ -1,165 +1,21 @@
-                   GNU LESSER GENERAL PUBLIC LICENSE
-                       Version 3, 29 June 2007
+The MIT License (MIT)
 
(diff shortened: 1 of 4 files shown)


ogoffart:
Now, some files do no longer have the license header. That's bad!

(I guess this should also be fixed upstream)

dragotin:
Yes, but came like this from upstream. I will do a PR there.

dragotin:
PR upstream: https://github.com/mojocorp/QProgressIndicator/pull/8

guruz:
Can we just merge this? The license is displayed in the LICENSE file in that directory.

ogoffart:
No we can't we need to add the license to every files.
We can however merge @dragotin's patch from mojocorp/QProgressIndicator#8

ogoffart:
Done in 766d9ae

PR #5188, opened 2016-09-16
ConnectionValidator: properly handle error in status.php

ogoffart:
We wwer enot connecting to the right signal from the check server
job, and therefore we were not catching the condition in which the
json was invalid. We would then never terminate the ConnectionValidator job.

Note that instanceNotFound is also emited if there is a network error.

The log looked like this:

10:25:51.247 OCC::CheckServerJob::finished: status.php from server is not valid JSON!
10:25:51.248 OCC::CheckServerJob::finished: status.php returns:  QMap()   QNetworkReply::NetworkError(NoError)  Reply:  QNetworkReplyHttpImpl(0x2b6a790)
10:25:51.248 OCC::CheckServerJob::finished: No proper answer on  QUrl("http://localhost/~owncloud/status.php")
10:26:23.235 OCC::AccountState::checkConnectivity: ConnectionValidator already running, ignoring "owncloud@localhost"
10:26:55.235 OCC::AccountState::checkConnectivity: ConnectionValidator already running, ignoring "owncloud@localhost"
[...]

### DIFF
--- a/src/libsync/connectionvalidator.cpp
+++ b/src/libsync/connectionvalidator.cpp
@@ -106,7 +106,7 @@ void ConnectionValidator::slotCheckServerAndAuth()
     checkJob->setTimeout(timeoutToUseMsec);
     checkJob->setIgnoreCredentialFailure(true);
     connect(checkJob, SIGNAL(instanceFound(QUrl,QVariantMap)), SLOT(slotStatusFound(QUrl,QVariantMap)));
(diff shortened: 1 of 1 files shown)


ogoffart:
...

guruz:
👍 
Can you cherry pick this into 2.2 too?

> Note that instanceNotFound is also emited if there is a network error.

I sure hope so :D

guruz:
Cherry pick still needs to be done

ogoffart:
Done.

Issue #5185, opened 2016-09-14
Changing data-fingerprint  and syncing, client does not assume a backup was recovered


davitol:
### Steps to reproduce
1. Disconnect the client
2. Do changes on some files of the server
3. Change the fingerprint (run ./occ maintenance:data-fingerprint)
4. Connect to the client

### Expected behaviour

Client should create conflict files

### Actual behaviour

No conflicts are created and the files are modified with the new content

### Server configuration
Operating system: OSX 10.10.5 (14F1909) client; Ubuntu 14.04 server

Web server: Apache

Database: MySQL

PHP version: PHP 5.5.9

ownCloud version: "version":"9.1.0.15"

@ogoffart

davitol:
:+1: Tested and WFM with Version 2.3.0-nightly20160922 (build 3694)

![screen shot 2016-09-22 at 10 25 20](https://cloud.githubusercontent.com/assets/10596560/18741349/0446e224-80af-11e6-9b56-e4800a4946c7.png)
![screen shot 2016-09-22 at 10 25 37](https://cloud.githubusercontent.com/assets/10596560/18741351/07131112-80af-11e6-8926-9e7bd997afa8.png)

PR #5189, opened 2016-09-16
Propagator: properly sned the success flag

ogoffart:
In case of the root directory, it may happen that the _item
is empty and the _item->_status is NoStatus.  But we still need to report
the proper success or error of the whole propagation. We should really
use _hasError for that. However, _hasError is also defined to NoStatus
if there was no error, so in that case we need to set Success.

This fixes the problem in which the data-fingerprint is not saved on the
database because the SyncEngine think that the sync failed. (Issue #5185)


guruz:
👎  for spelling mistake ;-)

Didnt check the code yet

ogoffart:
...

ogoffart:
The reason why the test is not completing:

```
Could not update commit status. Message: GitHub API rate limit exceeded
```

Issue #5192, opened 2016-09-18
File is deleted by syncing client, if it was modified and his parent folder has moved


lfontaine35:
Sorry for my english:
have described an issue on forum:
https://central.owncloud.org/t/file-delete-by-mistake-when-it-was-modified-and-his-parents-directory-was-moved-or-renamed/2476

### Expected behaviour

When i've the following tree folder who is synced with a client 2.2.3,

/FOLDERA/file1.txt
/FOLDERB

if i edit file1.txt on web interface, during a period who my sync-client is disconnect
and if i move FOLDERA in FOLDERB, or if i rename FOLDERA

when the sync-client reconnect, it moved FOLDERA in FOLDERB, or rename FOLDERA, it download the file with modification and during the next verficiation for change, it declare that te file1.txt in the older location was deleted
the server misunderstand and delete the file in the new location
the client sync delete the file in the new location
 
### Actual behaviour
i prefer the file was not deleted...

### Steps to reproduce
1. syncing a file in a folder
2. desactivate syncing client
3. editing the file for modification and saved. Rename the folder or move it
4. reactivate syncing client and wait
5. the file is delete and in the trashbin folder

### Server configuration
Operating system:  8.6

Web server: 2.4.10

Database: mysql

PHP version: 5.6.24

ownCloud version: 9.1, it's true with 9.02 and 8.1

Storage backend (external storage): no

### Client configuration
Client version: 2.2.3 

Operating system: windows 7 - 8 and 10 in x64.

OS language: french

Installation path of client: %userprofile%owncloud

### Logs

Please use Gist (https://gist.github.com/) or a similar code paster for longer
logs.

```Template for output < 10 lines```

1. Client logfile: Output of `owncloud --logwindow` or `owncloud --logfile log.txt`
(On Windows using `cmd.exe`, you might need to first `cd` into the ownCloud directory)
(See also http://doc.owncloud.org/desktop/2.2/troubleshooting.html#client-logfile )

2. Web server error log:

3. Server logfile: ownCloud log (data/owncloud.log):

dragotin:
I am not able to reproduce that (at least on Linux), and there is the code that handles this:
https://github.com/owncloud/client/blob/master/csync/src/csync_update.c#L440

It compares the ETags and if they are not equal, the file is downloaded again. You should see that GET that re-downloads the file from the new location  in your servers `access_log`.

tflidd:
[...]

PR #5195, opened 2016-09-20
SyncEngine: Fix renaming of folder when file are changed

ogoffart:
Two bugs:
 - The change filed are not considered as move, they are re-downloaded
   but the old file was not removed from the database. The change in
   owncloudpropagator.cpp takes care of removing the old entries.

 - Next sync would then remove the file in the server in the old folder
   This was not a problem until we start reusing the sync engine, and
   that the _renamedFolders map is not cleared. We were before deleting
   a non-existing file. But now we delete the actual file.

Also improve the tests to be able to do move on the server.
This include support for file id.

Issue #5192

### DIFF
--- a/test/syncenginetestutils.h
+++ b/test/syncenginetestutils.h
@@ -14,12 +14,16 @@
 
 #include <QDir>
 #include <QNetworkReply>
+#include <QMap>
 #include <QtTest>
 
 static const QUrl sRootUrl("owncloud://somehost/owncloud/remote.php/webdav/");
 
-static QString generateEtag() {
+inline QString generateEtag() {
     return QString::number(QDateTime::currentDateTime().toMSecsSinceEpoch(), 16);
+}
+inline QByteArray generateFileId() {
+    return QByteArray::number(qrand(), 16);
 }
 
 class PathComponents : public QStringList {
@@ -44,6 +48,7 @@ public:
     virtual void setContents(const QString &relativePath, char contentChar) = 0;
     virtual void appendByte(const QString &relativePath) = 0;
     virtual void mkdir(const QString &relativePath) = 0;
+    virtual void rename(const QString &relativePath, const QString &relativeDestinationDirectory) = 0;
 };
 
 class DiskFileModifier : public FileModifier
@@ -85,6 +90,9 @@ public:
     void mkdir(const QString &relativePath) override {
         _rootDir.mkpath(relativePath);
     }
+    void rename(const QString &, const QString &) override {
+        Q_ASSERT(!"not implemented");
+    }
 };
 
 class FileInfo : public FileModifier
@@ -125,6 +133,7 @@ public:
         for (const auto &source : children) {
             auto &dest = this->children[source.name] = source;
             dest.parentPath = p;
+            dest.fixupParentPathRecursively();
         }
     }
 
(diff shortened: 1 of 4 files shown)


test/syncenginetestutils.h
```diff
@@ -14,13 +14,17 @@
...
 #include <QNetworkReply>
+#include <QMap>
 #include <QtTest>
 
 static const QUrl sRootUrl("owncloud://somehost/owncloud/remote.php/webdav/");
 
 static QString generateEtag() {
     return QString::number(QDateTime::currentDateTime().toMSecsSinceEpoch(), 16);
 }
+QByteArray generateFileId() {
```
jturcotte: Might need to be static if this header ever gets built more than once per linked program (even though something else will probably fail).

test/syncenginetestutils.h
```diff
@@ -85,6 +90,14 @@ class DiskFileModifier : public FileModifier
     void mkdir(const QString &relativePath) override {
         _rootDir.mkpath(relativePath);
     }
+    void move(const QString &relativePath, const QString &relativeDestinationDirectory) override {
+        auto orig = _rootDir.filePath(relativePath);
+        auto dest = _rootDir.filePath(relativeDestinationDirectory);
+        QVERIFY(QFileInfo::exists(orig));
+        QVERIFY(QFileInfo::exists(dest));
+        QVERIFY(QFileInfo(dest).isDir());
+        QVERIFY(_rootDir.rename(relativePath, relativeDestinationDirectory));
```
jturcotte: Move "A/a1" --> "B/a1" should probably not be rename to "B", or am I missing something? Is it a lot of work to add a test that moves a file to make sure it works too?

test/syncenginetestutils.h
```diff
@@ -156,6 +170,20 @@ class FileInfo : public FileModifier
...
 
+    void move(const QString &relativePath, const QString &relativeDestinationDirectory) override {
+        FileInfo *dir = findInvalidatingEtags(relativeDestinationDirectory);
+        Q_ASSERT(dir);
+        Q_ASSERT(dir->isDir);
+        const PathComponents pathComponents{relativePath};
+        FileInfo *parent = find(pathComponents.parentDirComponents());
+        Q_ASSERT(parent);
+        FileInfo fi = parent->children.take(pathComponents.fileName());
+        qWarning() << "MOVE!!!!!!! " << relativePath << relativeDestinationDirectory << *parent;
```
jturcotte: Might be better not to leave the debugging stuff there.

test/syncenginetestutils.h
```diff
@@ -228,6 +257,19 @@ class FileInfo : public FileModifier
     FileInfo *findInvalidatingEtags(const PathComponents &pathComponents) {
         return find(pathComponents, true);
     }
+
+    void fixupParentPathRecursively() {
+        auto p = path();
```
jturcotte: You could use that variable, or remove it.

test/syncenginetestutils.h
```diff
@@ -228,6 +257,19 @@ class FileInfo : public FileModifier
...
+        for (auto it = children.begin(); it != children.end(); ++it) {
+            Q_ASSERT(it.key() == it->name);
+            it->parentPath = path();
+            it->fixupParentPathRecursively();
+        }
+    }
+
+    friend inline QDebug operator<<(QDebug dbg, const FileInfo& fi) {
```
jturcotte: Could be removed too if you remove the qWarning above.

test/testsyncengine.cpp
```diff
@@ -119,6 +119,34 @@ private slots:
...
+    void testRemoteChangeInMovedFolder() {
+        // issue #5192
+        FakeFolder fakeFolder{FileInfo{ QString(), {
+            FileInfo { QStringLiteral("folder"), {
+                FileInfo{ QStringLiteral("folderA"), { { QStringLiteral("file.txt"), 400 } } },
+                QStringLiteral("folderB")
+            }
+        }}}};
+
+        fakeFolder.syncOnce();
```
jturcotte: There is already a syncOnce in the FakeFolder constructor, this shouldn't be necessary in every test.

test/syncenginetestutils.h
ogoffart: this is only for move,  not renames.

test/syncenginetestutils.h
ogoffart: oops, that not intentional

test/syncenginetestutils.h
ogoffart: This i wanted to leave there for convinience.

test/syncenginetestutils.h
jturcotte: What I mean is that there is nothing in the rename documentation that you can achieve "A/a1" --> "B/a1" by doing `rename("A/a1", "B/")`. You want to move a1 into B, and to do that wouldn't the correct way to do `rename("A/a1", "B/a1")`?

PR #5199, opened 2016-09-22
OwnSQL: avoid crashing in SqlQuery::reset_and_clear_bindings

ogoffart:
The crash reporter shows a lot of crashes in sqlite3_clear_bindings
which seems to indicate that _stmt is null. We should guard against
a null value in order to avoid crashing.

This should only happen if the prepare call fails. We don't usually
check the return value of the prepare call, but if _stmt is null, the
exec call should return false, not true. We check the result of the
exec call, so this should then abort the sync with an error, rather
than crashing.

### DIFF
--- a/src/libsync/ownsql.cpp
+++ b/src/libsync/ownsql.cpp
@@ -230,8 +230,12 @@ bool SqlQuery::isPragma()
 
 bool SqlQuery::exec()
 {
+    if (!_stmt) {
+        return false;
+    }
+
     // Don't do anything for selects, that is how we use the lib :-|
-    if(_stmt && !isSelect() && !isPragma() ) {
+    if( !isSelect() && !isPragma() ) {
         int rc, n = 0;
         do {
             rc = sqlite3_step(_stmt);
@@ -376,8 +380,10 @@ void SqlQuery::finish()
 
 void SqlQuery::reset_and_clear_bindings()
 {
-    SQLITE_DO(sqlite3_reset(_stmt));
-    SQLITE_DO(sqlite3_clear_bindings(_stmt));
+    if (_stmt) {
+        SQLITE_DO(sqlite3_reset(_stmt));
+        SQLITE_DO(sqlite3_clear_bindings(_stmt));
+    }
 }
 
 } // namespace OCC


ogoffart:
cc @jturcotte

jturcotte:
This will avoid a crash, but if this is really the case, it would be nice if checkConnect would return false the next time so that the statements are prepared again, else you might end up with problems worse than a crash.

ogoffart:
@jturcotte   hence the change in exec to return false. 

But yes, it would be better to check the value of prepare all the time.

jturcotte:
A failed exec usually just means that a qWarning is output in SyncJournalDB. If some exec fail and some succeed, this could mean an inconsistent DB and this is worse than a crash IMO, this could result in data loss.

jturcotte:
@ogoffart Could you add an `sqlFail`  in `SyncJournalDb::setSelectiveSyncList`, `SyncJournalDb::avoidRenamesOnNextSync`, `SyncJournalDb::avoidReadFromDbOnNextSync`, `SyncJournalDb::forceRemoteDiscoveryNextSyncLocked`, `SyncJournalDb::getChecksumType`? Covering the ones we see in most crash reports would probably help already.

ogoffart:
> A failed exec usually just means that a qWarning is output in SyncJournalDB.

No, it means in some cases sqlFail, and in other cases that the calling function will report the error. Which if it happens during the sync will abort the sync with FatalError and an adequate error message

jturcotte:
Seems like all the setters have proper error handling, so that would only leave something like `SyncJournalDb::getSelectiveSyncList` to cause problems. The `ok` flag isn't checked enough to fail a sync and it doesn't close the database. I don't know what the consequences for this could be, but I guess it won't break the database directly.

lohner:
I don't know if it is of interest, but I can reliably reproduce that crash whenever the filesystem of my sync-folder is mounted read-only (Having a dual-boot setup with Windows/Linux and the sync-folder on a shared NTFS partition this happens if Windows was hibernated instead of a proper shutdown).

Here is the gdb backtrace:
```
Thread 1 "owncloud" received signal SIGSEGV, Segmentation fault.
0x00007fffefe21892 in sqlite3_clear_bindings () from /usr/lib/x86_64-linux-gnu/libsqlite3.so.0
(gdb) bt
#0  0x00007fffefe21892 in sqlite3_clear_bindings () from /usr/lib/x86_64-linux-gnu/libsqlite3.so.0
#1  0x00007ffff5740b26 in OCC::SqlQuery::reset_and_clear_bindings() ()
   from /usr/lib/x86_64-linux-gnu/libowncloudsync.so.0
#2  0x00007ffff57324b4 in OCC::SyncJournalDb::getSelectiveSyncList(OCC::SyncJournalDb::SelectiveSyncListType, bool*) () from /usr/lib/x86_64-linux-gnu/libowncloudsync.so.0
#3  0x0000000000480806 in OCC::AccountSettings::refreshSelectiveSyncStatus() ()
#4  0x000000000048149b in OCC::AccountSettings::AccountSettings(OCC::AccountState*, QWidget*) ()
#5  0x00000000004fcf82 in OCC::SettingsDialog::accountAdded(OCC::AccountState*) ()
#6  0x00000000004fdd36 in OCC::SettingsDialog::SettingsDialog(OCC::ownCloudGui*, QWidget*) ()
#7  0x00000000004d53b1 in OCC::ownCloudGui::ownCloudGui(OCC::Application*) ()
#8  0x00000000004887fa in OCC::Application::Application(int&, char**) ()
#9  0x0000000000475f48 in main ()
```

And the tail of `strace owncloud`:
```
fstat(29, {st_mode=S_IFREG|0777, st_size=77824, ...}) = 0
open("/path/to/sync/folder/.csync_journal.db-shm", O_RDWR|O_CREAT|O_CLOEXEC, 0777) = -1 EROFS (Read-only file system)
fstat(29, {st_mode=S_IFREG|0777, st_size=77824, ...}) = 0
open("/path/to/sync/folder/.csync_journal.db-shm", O_RDWR|O_CREAT|O_CLOEXEC, 0777) = -1 EROFS (Read-only file system) 
fstat(29, {st_mode=S_IFREG|0777, st_size=77824, ...}) = 0
open("/path/to/sync/folder/.csync_journal.db-shm", O_RDWR|O_CREAT|O_CLOEXEC, 0777) = -1 EROFS (Read-only file system) 
fstat(29, {st_mode=S_IFREG|0777, st_size=77824, ...}) = 0
open("/path/to/sync/folder/.csync_journal.db-shm", O_RDWR|O_CREAT|O_CLOEXEC, 0777) = -1 EROFS (Read-only file system) 
fstat(29, {st_mode=S_IFREG|0777, st_size=77824, ...}) = 0
open("/path/to/sync/folder/.csync_journal.db-shm", O_RDWR|O_CREAT|O_CLOEXEC, 0777) = -1 EROFS (Read-only file system) 
--- SIGSEGV {si_signo=SIGSEGV, si_code=SEGV_MAPERR, si_addr=0} ---
+++ killed by SIGSEGV (core dumped) +++
Speicherzugriffsfehler (Speicherabzug geschrieben)
```

guruz:
@lohner Thanks for the backtrace! You can also run the client with parameters ```--logfile -``` to make it log to stdout/stderr.

@ogoffart Does this mean that exec() should check if prepare() failed? e.g. see ```_error = QString::fromUtf8(sqlite3_errmsg(_db));``` in prepare()? Or is the issue somehow else?

ogoffart:
@guruz exec already checks if prepare() failed.  The problem (that i am fixing in this patch) is that reset_and_clear_bindings() does not check it.

ogoffart:
Anyway, i'll merge this patch already.  Further imporovements to add the sqlFail  can be added later.

PR #5211, opened 2016-09-28
Update changelog for 2.2.4 final

davitol:
@guruz

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,15 @@
 ChangeLog
 =========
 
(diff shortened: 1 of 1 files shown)


guruz:
Hey!
thanks, it should not be necessary, @ogoffart will merge 2.2 into master.

```
<guruz> olivier: do you want to merge? can you also set 2.2 branch to 2.2.5?
<olivier> I'm not home now
<olivier> But I can do it later
```

PR #5215, opened 2016-09-29
csync: Do not wipe the db file name when cleaning csync_ctx.

dragotin:
The name of the db is only set in the constructor of the SyncEngine
and never again. If we wipe it, subsequent operations can not open
the database through csync. For example csync_update does that.

This problem was found after the client reported a csync error
after changing the selective sync list.

### DIFF
--- a/csync/src/csync.c
+++ b/csync/src/csync.c
@@ -521,7 +521,6 @@ static void _csync_clean_ctx(CSYNC *ctx)
     c_rbtree_free(ctx->local.tree);
     c_rbtree_free(ctx->remote.tree);
 
-    SAFE_FREE(ctx->statedb.file);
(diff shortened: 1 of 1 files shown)


dragotin:
@ogoffart please check. Do you agree on this?

ogoffart:
:-1:   It will then be leaking!
the free would  need to be put in csync_destroy and  csync_update before it reallocates it.

But really, this should not need to be changed at all.

ogoffart:
@dragotn  csync_update calls asprintf that re-allocates statedb.file.

Maybe this could be optimized if statedb.file is already set, not to reallocate it again.

dragotin:
Ok, this patch does only make sense in the branch context.

PR #5229, opened 2016-10-07
OS X: Use monochrome tray icon by default

guruz:
Review @tsparber  ?

### DIFF
--- a/src/libsync/configfile.cpp
+++ b/src/libsync/configfile.cpp
@@ -611,7 +611,12 @@ void ConfigFile::setPromptDeleteFiles(bool promptDeleteFiles)
 bool ConfigFile::monoIcons() const
 {
     QSettings settings(configFile(), QSettings::IniFormat);
-    return settings.value(QLatin1String(monoIconsC), false).toBool();
+    bool monoDefault = false; // On Mac we want bw by default
+#ifdef Q_OS_MAC
+    // OEM themes are not obliged to ship mono icons
+    monoDefault = (0 == (strcmp("ownCloud",APPLICATION_NAME)));
+#endif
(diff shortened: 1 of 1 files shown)


src/libsync/configfile.cpp
```diff
@@ -611,7 +611,12 @@ void ConfigFile::setPromptDeleteFiles(bool promptDeleteFiles)
 bool ConfigFile::monoIcons() const
 {
     QSettings settings(configFile(), QSettings::IniFormat);
-    return settings.value(QLatin1String(monoIconsC), false).toBool();
+    bool monoDefault = false; // On Mac we want bw by default
+#ifdef Q_OS_MAC
+    // OEM themes are not obliged to ship mono icons
+    monoDefault = (0 == (strcmp("ownCloud",APPLICATION_NAME)));
```
tsparber: I think it would be better to check if there are mono icons (black/white). If the OEM theme includes black&white icons it would be good to also default to those.
Similar to [src/gui/generalsettings.cpp#L80](https://github.com/owncloud/client/blob/91b6b888835852450e521c3a6bf5f2366e85feef/src/gui/generalsettings.cpp#L80):
```cpp
    // OEM themes are not obliged to ship mono icons, so there
    // is no point in offering an option
    QString themeDir = QString::fromLatin1(":/client/theme/%1/")
            .arg(Theme::instance()->systrayIconFlavor(true));
    _ui->monoIconsCheckBox->setVisible(QDir(themeDir).exists());
```
Maybe factor this check out into a separate function...

Btw: maybe this option should also only be shown if black **and** white icons are present.

jturcotte: Ahh, so we hide the option in that case, I didn't know.

Issue #4986, opened 2016-06-20
sync mutiple small files from windows client to server, slow and every time 3 files uploaded


joeyjy:
### Expected behaviour
upload smoothly as single file

### Actual behaviour
upload slow and every time 3 objects uploaded (according to the log)

### Steps to reproduce
1. move a folder contains 100 small files to owncloud sync dir

### Server configuration
Operating system: centos7.0

Web server: apache 2.4

Database: MariaDB 5.5

PHP version: PHP 5.4.16 (cli)

ownCloud version: 9.0.1 (with encryption)

Storage backend: ceph object storage

### Client configuration
Client version: 2.2

Operating system: window7 basic

OS language: Chinese

Installation path of client: C:\Program Files (x86)

### Logs

Please use Gist (https://gist.github.com/) or a similar code paster for longer
logs.

```Template for output < 10 lines```

1. Output of `owncloud --logwindow` or `owncloud --logfile log.txt`
(On Windows using `cmd.exe`, you might need to first `cd` into the ownCloud directory)
(See also http://doc.owncloud.org/desktop/2.2/troubleshooting.html#client-logfile )

2. Web server error log:

3. ownCloud log (data/owncloud.log):

Hi OC team,

I want to know if the encryption process is simultaneous when handling multiple files? or it only encrypts file one by one?
I notice that when sync big file, it uses chunk uploading, 10M for each chunk and send the chunk to object storage first, when all  the file is uploaded, OC gets all the chunks from object storage, combines them, encrypts the file, finally sends the complete to object storage.

guruz:
Each file is an independent HTTP request (for small files). So it indeed happens one by one.

Does this answer your question?

joeyjy:
@guruz thank you for reply!
I see, so, multiple REQUESTS will be sent to the server at one time, can we config the number?

guruz:
@joeyjy You can increase it to a maximum of 5 with the environment variable ```OWNCLOUD_MAX_PARALLEL=5```.

Note to self: Maybe we should do the auto-scaling also for uploads in 2.3 (https://github.com/owncloud/client/pull/4529)

joeyjy:
@guruz Thanks! I also noticed that "if !max max = 3" in the client source code.

fossxplorer:
@joeyjy 
[...]

PR #5230, opened 2016-10-07
Propagator: Also upload more in parallel

guruz:
Fix for https://github.com/owncloud/client/issues/4986#issuecomment-227071801

This is before we have bundling at some point.

### DIFF
--- a/src/libsync/owncloudpropagator.cpp
+++ b/src/libsync/owncloudpropagator.cpp
@@ -529,7 +529,7 @@ void OwncloudPropagator::scheduleNextJob()
             }
         }
         if (_activeJobList.count() < maximumActiveJob() + likelyFinishedQuicklyCount) {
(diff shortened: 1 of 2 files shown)


ckamm:
@ogoffart Was there a reason uploads were never marked as likely to finish quickly?

guruz:
@ogoffart If there was one (I don't remember for sure) it was: "let's try first if this does not mess up the server load too much"

ogoffart:
Why the magic value of 10KiB ?  Where does that number comes from?

guruz:
@ogoffart it's 100 not 10.
I took less than for the download magic value because I'm assuming a PUT is more costly time wise on the server

ogoffart:
Yeah, i just wonder why this value and not 200 or 50 ?
Did you just select randomly?

Issue #3449, opened 2015-07-13
[Windows, OS X] Switch to Qt 5.6.2 for 2.3 client


guruz:
Right now we're based on 5.4.0+patches

https://github.com/qtproject/qtbase/blob/dev/dist/changes-5.4.1 (incomplete usually)
https://github.com/qtproject/qtbase/blob/dev/dist/changes-5.4.2 (incomplete usually)
https://github.com/qtproject/qtbase/blob/dev/dist/changes-5.5.0 (incomplete usually)
https://wiki.qt.io/New_Features_in_Qt_5.5

dragotin:
I'd vote for staying with the current build that we had for 1.8.4 (5.4.0+patches) for 2.0.0 and go to hopefully 5.5.1 wtih the next bugfix release, ie. oCC 2.0.1. One moving target less.

hefee:
for downstream it would be nice to to nail to newest qt version and don't bump the minimum version for no reason. 

And changing the minimum version within bugfix releases is really not a good thing, cause normally bugfix releases should be a minimal changeset and therefore the dependency shouldn't changed. 

For sure you can ship a new qt version within obs at any time - i'm just talking about the CMakefile.txt

danimo:
There is no time for a 5.5 switch in the 2.0 timeframe. @hefee I don't think we'll soon drop support for older Qt versions, but e.g. 5.5. solves the systray issue. But that's a different conversation, since we can't have 5.5. everywhere in the forseeable future.

guruz:
-> 2.1 then

guruz:
5.5.1 is about-to-be-released..

ogoffart:
5.5.1 was released

dragotin:
This is the repo to update: https://build.opensuse.org/project/show/isv:ownCloud:toolchains:mingw:win32:stable
and this is how its done:
https://en.opensuse.org/openSUSE:Build_Service_Collaboration#How_do_I_contribute_my_changes_to_someone_else_packages.3F

guruz:
5.6 (unreleased) has some network fixes, analyzing those now and then maybe we stay a bit on our patched 5.4 plus those patches..

scolebrook:
Is the patched version of qt you use for official builds of ownCloud available anywhere? Or are the mentioned patches in the client repo and get applied during the build process?

I'd prefer to make sure that our build is as close as possible to being the same as yours (apart from branding of course).

ogoffart:
[...]

PR #5233, opened 2016-10-10
Move check for mono icons into separate function

tsparber:
Refactor the code to use a separate function to check if mono icons are available.
This could be used in #5229.

Open questions:
- This only checks for a single color (black or white). Should this be changed to check for both?
  (macOS currently requires black&white icons, this could also be changed to white only if build on >= Qt 5.6.0).


guruz:
Let's evaluate that when we have changed the build to 5.6
https://github.com/owncloud/client/issues/3449

Issue #5224, opened 2016-10-04
Rename a folder downloads entire content again


daorte:
### Expected behaviour

When renaming a folder, it should remember the sync configuration of its subfolders and not download everything again.
### Steps to reproduce
1. Create a folder 'ParentFolder'
2. Create two subfolders of 'ParentFolder', 'SubFolderOne' and 'SubFolderTwo'
3. Put any content on both subfolders
4. Sync with the desktop application
5. At this point, in the Desktop Application main screen, you should have 'ParentFolder' and all its subfolders with marked sync checks
6. Uncheck 'SubFolderTwo'
7. The Desktop Application deletes the folder from your local disk
8. Go to OC Web, login with your account and rename 'ParentFolder' to 'ParentFolderRenamed'
9. The Desktop Application renames 'ParentFolder' to 'ParentFolderRenamed' and download again 'SubFolderTwo'
### Server configuration

Operating system: Ubuntu 16.04
Web server: Apache 2.4
Database: MySQL
PHP version: 7
ownCloud version: 9.1.1
Storage backend (external storage): NO
### Client configuration

Client version: 2.2.3
Operating system: Windows 8.1
OS language: Spanish
Installation path of client: C:\Program Files (x86)\ownCloud

ogoffart:
Right,  the entries in the selective sync black list are not renamed.  This is actually a bug.

daorte:
Tested on Windows Client Version 2.2.4 (build 6408)

does not work...

SamuAlfageme:
Not happening anymore. Tested on [Windows Client Version 2.3.0 (build 6501)](http://download.owncloud.com/desktop/daily/ownCloud-2.3.0.6501-nightly20161110-setup.exe) 
![img](http://i.imgur.com/wFzBRWl.gif)

SamuAlfageme:
👍

PR #5235, opened 2016-10-10
SyncEngine: Fix renaming a folder should keep the selective sync state

ogoffart:
Issue #5224

Two problems:

 - In the discovery phase, we need to check the selective sync entries of
   the source path in case of renames.

 - When the rename is done, we need to actually update the black list in the
   database.

### DIFF
--- a/test/testsyncengine.cpp
+++ b/test/testsyncengine.cpp
@@ -7,6 +7,7 @@
 
 #include <QtTest>
 #include "syncenginetestutils.h"
+#include <syncengine.h>
 
 using namespace OCC;
 
@@ -137,8 +138,8 @@ private slots:
         fakeFolder.syncOnce();
         QCOMPARE(fakeFolder.currentLocalState(), fakeFolder.currentRemoteState());
         auto oldState = fakeFolder.currentLocalState();
-        QVERIFY(oldState.find(PathComponents("folder/folderB/folderA/file.txt")));
-        QVERIFY(!oldState.find(PathComponents("folder/folderA/file.txt")));
+        QVERIFY(oldState.find("folder/folderB/folderA/file.txt"));
+        QVERIFY(!oldState.find("folder/folderA/file.txt"));
 
         // This sync should not remove the file
         fakeFolder.syncOnce();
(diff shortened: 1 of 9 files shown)


test/testsyncengine.cpp
```diff
@@ -146,6 +147,69 @@ private slots:
...
+        {
+            auto remoteState = fakeFolder.currentRemoteState();
+            // The subFolderA should still be there on the server.
+            QVERIFY(remoteState.find(PathComponents("parentThirdName/subFolderA/fileA.txt")));
+            // But not on the client because of the selective sync
+            remoteState.remove("parentThirdName/subFolderA");
+            QCOMPARE(fakeFolder.currentLocalState(), remoteState);
+
+            expectedServerState = fakeFolder.currentRemoteState();
+            fakeFolder.syncOnce(); // This sync should do nothing
```
jturcotte: You could verify this further with a `QSignalSpy::isEmpty()` on itemCompleted, if it's the doing part that matters like your comment suggests.

test/testsyncengine.cpp
```diff
@@ -146,6 +147,69 @@ private slots:
...
+        // Rename parentFolder on the server
+        fakeFolder.remoteModifier().rename("parentFolder", "parentFolderRenamed");
+        expectedServerState = fakeFolder.currentRemoteState();
+        fakeFolder.syncOnce();
+
+        {
+            QCOMPARE(fakeFolder.currentRemoteState(), expectedServerState);
+            auto remoteState = fakeFolder.currentRemoteState();
+            // The subFolderA should still be there on the server.
+            QVERIFY(remoteState.find(PathComponents("parentFolderRenamed/subFolderA/fileA.txt")));
```
ogoffart: I just copy pasted from other tests.

ogoffart:
From jenkins: 

> CMakeFiles/owncloudsync.dir/objects.a(discoveryphase.cpp.obj):discoveryphase.cpp:(.text+0x37d): undefined reference to `_imp__csync_rename_count'
> CMakeFiles/owncloudsync.dir/objects.a(discoveryphase.cpp.obj):discoveryphase.cpp:(.text+0x3a3): undefined reference to `_imp__csync_rename_adjust_path_source'

I don't understand.  What's the `_imp__`  ? The export macro is there, and the same file also do calls to `csync_vio_file_stat_new` and that works.

PR #5239, opened 2016-10-11
Add a note that the source repositories need to be enabled.

dragotin:
@carlaschroder please check. 

I had a user here who was struggling with the description because he did not have the source repos enabled on Ubuntu.


carlaschroder:
@dragotin Ah yes, that is a rather necessary step! Please change " is enabled" to "are enabled" and then I can merge. Thanks!

carlaschroder:
Groovy. Thanks!

Issue #4628, opened 2016-03-31
sync-exclude.lst not working with full folder patterns


kaeser-sosae:
### Expected behaviour

The "Music/iTunes/Album Artwork" folder should be excluded from Sync.
### Actual behaviour

The "Music/iTunes/Album Artwork" folder is synced, including it's entire contents.
### Steps to reproduce
1. Add the following lines to the ~/Library/Application Support/ownCloud/sync-exclude.lst file:

`]*iTunes/Album Artwork/`
`]iTunes/Album Artwork`
`]iTunes/Album Artwork/*`
`]iTunes/Album Artwork/`
`]*iTunes/Album Artwork*`
`]*iTunes/Album Artwork/*`
### Server configuration

Operating system: Ubuntu 14.04.4

Web server: Apache 2.4.7

Database: MySQL 5.5.47

PHP version: 5.5.9

ownCloud version: 9.0.0

Storage backend: SMB
### Client configuration

Client version: 2.1.1 Build 3107

Operating system: OS X 10.10.5

OS language: English

Installation path of client: /Applications/owncloud.app
### Logs

I haven't included any logs, as the folders simply sync and are not excluded.

It's entirely possible that I'm doing something wrong, but the documentation here: https://doc.owncloud.org/desktop/2.1/navigating.html#using-the-ignored-files-editor is not specific enough (for me!).

Cheers

guruz:
What if you just put ```Album Artwork```?

dragotin:
rather `Album Artwork/` to make it only match directories.

kaeser-sosae:
Because I only want the Album Artwork folder inside the iTunes folder to be excluded.

I might have a folder called ~/Desktop/Album Artwork that I want synced. If I do as you suggest, any folder called Album Artwork could be excluded.

Remember that this folder is one example of the Sync not working.

guruz:
@kaeser-sosae Yes, we know that. We want to find out if you have a general exclude problem or only this specific pattern with '/'

dragotin:
Can it be a problem with the space in "Album Artwork"? Does a directory without space in between work as expected?

dragotin:
@kaeser-sosae you need to add the pattern in `sync-exclude.lst` it with a leading slash:
```
/Music/iTunes/Album Artwork
```
will do it. Tested here: If you skip the leading slash, it fails, with it works.

Keeping the issue open because it might make sense to fix that to also work without slash.

tuxayo:
[...]

PR #5245, opened 2016-10-12
FolderWatcher (linux): optimize slotAddFolderRecursive

ogoffart:
- Instead of putting all the pathes in a big QStringList and then process them,
  do the processing as we iterate with QDirIterator
- Since we know that the parent directory is not ignored, we can simplify the
  ignore check by not checking the part before (i.e: we call
  ExcludedFiles::isExcluded with the parent 'path' as base path) This is what
  speeds up considerably the function.

Note that the OWNCLOUD_TEST check is already there in FolderWatcher::pathIsIgnored
so the behavior is not changed

### DIFF
--- a/src/gui/folderwatcher_linux.cpp
+++ b/src/gui/folderwatcher_linux.cpp
@@ -23,6 +23,7 @@
 #include <QStringList>
 #include <QObject>
 #include <QVarLengthArray>
+#include <syncengine.h>
 
 namespace OCC {
 
@@ -48,31 +49,6 @@ FolderWatcherPrivate::~FolderWatcherPrivate()
 
 }
 
-// attention: result list passed by reference!
-bool FolderWatcherPrivate::findFoldersBelow( const QDir& dir, QStringList& fullList )
-{
-    bool ok = true;
-    if( !(dir.exists() && dir.isReadable()) ) {
-        qDebug() << "Non existing path coming in: " << dir.absolutePath();
-        ok = false;
-    } else {
-        QStringList nameFilter;
-        nameFilter << QLatin1String("*");
-        QDir::Filters filter = QDir::Dirs | QDir::NoDotAndDotDot | QDir::NoSymLinks | QDir::Hidden;
-        const QStringList pathes = dir.entryList(nameFilter, filter);
-
-        QStringList::const_iterator constIterator;
-        for (constIterator = pathes.constBegin(); constIterator != pathes.constEnd();
-               ++constIterator) {
-            const QString fullPath(dir.path()+QLatin1String("/")+(*constIterator));
-            fullList.append(fullPath);
-            ok = findFoldersBelow(QDir(fullPath), fullList);
-        }
-    }
-
-    return ok;
-}
-
 void FolderWatcherPrivate::inotifyRegisterPath(const QString& path)
 {
     if( !path.isEmpty()) {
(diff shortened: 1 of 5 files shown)


ckamm:
@ogoffart I haven't read this in detail yet, but wouldn't the change in exclude check break patterns that include the path ("foo/bar/file" wouldn't match if we feed only "file" inside "foo/bar/")? I'm not sure, in general, if we aim to support patterns like that.

ogoffart:
@ckamm  good point

guruz:
I'm actually thinking if we should stop supporting this kind of exclusion pattern?

1) My https://github.com/owncloud/client/pull/4999 does not support it either (yet)
2) There is a bug about this that it is broken anyway: https://github.com/owncloud/client/issues/4628

ckamm:
@ogoffart @guruz But wouldn't this change still break full-absolute paths, like ignoring "/A/B/C"? Because "/A/B/" is not excluded, and "C" doesn't match any pattern on its own.

test/testinotifywatcher.cpp
```diff
@@ -11,18 +11,23 @@
...
+{
+    Q_OBJECT
 private slots:
-    void initTestCase() {
-        qsrand(QTime::currentTime().msec());
+    // Test the recursive path listing function lists everything
+    void testAddFolderRecursiveHelper() {
+        QTemporaryDir tmpDir;
 
-        _root = QDir::tempPath() + "/" + "test_" + QString::number(qrand());
```
ckamm: Why did you remove the test_1231 subdirectory?

src/gui/folderwatcher_linux.cpp
```diff
@@ -88,41 +64,47 @@ void FolderWatcherPrivate::inotifyRegisterPath(const QString& path)
...
+
+    QDir::Filters filter = QDir::Dirs | QDir::NoDotAndDotDot | QDir::NoSymLinks | QDir::Hidden;
+    const QStringList pathes = dir.entryList(filter);
+    for (auto constIterator = pathes.constBegin(); constIterator != pathes.constEnd(); ++constIterator) {
+        const QString subfolder = path + QLatin1String("/") + (*constIterator);
+        QDir folder(subfolder);
+        if (folder.exists() && !watchedFolders.contains(subfolder)) {
+#ifndef OWNCLOUD_TEST // InotifyWatcherTest is not interested in ignored files and does not link against the folder
+            if( _parent->_folder->syncEngine().excludedFiles().isExcluded(
+                    subfolder, path, _parent->_folder->ignoreHiddenFiles())) {
```
ckamm: I think for correctness this must always pass in the full folder-relative path. (see other comment)

guruz:
@ckamm I've never seen full absolute paths in the exclude. What is it absolute to? The system root or the sync folder root?

ckamm:
@guruz I meant absolute to the sync folder root. For example the "/Music/iTunes/Album Artwork" pattern that @dragotin recommended in #4628.

test/testinotifywatcher.cpp
ogoffart: I'm using QTemporaryDir which take care of that instead of manually creating and destroying the directory.

ckamm:
I've created #5275 for discussion about what the desired behavior for exclude patterns even is.

guruz:
I wrote my personal conclusion as comment on https://github.com/owncloud/client/issues/5275#issuecomment-262519711
How would we integrate this with this PR here? (Because we want to fix https://github.com/owncloud/client/issues/4628 too)

ckamm:
@ogoffart I'd close this?

PR #5252, opened 2016-10-13
add note about deleting .csync_journal.db for "CSync unknown error"

carlaschroder:


carlaschroder:
@cdamken

cdamken:
Thanks Carla :)

guruz:
I'm OK with this.

However: @ogoffart This resets the selective sync list right? So it syncs everything then?

ogoffart:
Deleting the sync journal is dangerous and may cause fyle to be re-sync'ed that we don't want, or conflicts that the user don't want.

I don't think it's good to suggest the user to do it for an error that vague.

Or at least a word of caution should be there. Maybe something like: “Warning: this will delete the synchronisation state, and may result in the client producing conflicts or synchronizing extra files. The Choose what you sync state will also be lost.”

Not that sometimes, corruption is temporary because the disk was removed or temporarily unreadable.

carlaschroder:
> The Choose what you sync state will also be lost

Ay yi yi that is not good.

Can anyone give me some specific circumstances that justify deleting the sync journal? If we can narrow it down to some specific use cases then it's a useful tip, and not so much creating more problems.

 @dragotin @ogoffart @guruz

ogoffart:
In fact, when the journal is corrupted, the error is
"CSync failed to load the journal file. The journal file is corrupted."

The problem in issue 5226 seems to be some other corruption

carlaschroder:
Thanks @ogoffart

guruz:
I'm confused, why can't i assign @settermjd here?
@DeepDiver1975 ?

guruz:
I suggest to add a further sentence "Please note that this will also erase some of your settings about which files to download"?

settermjd:
@guruz not to worry. I assigned myself.

settermjd:
@guruz I added the note as an admonition, given the potential seriousness of the action, if taken. If you're happy with how this reads, then give it a thumbs up and I'll merge the PR.

guruz:
Thanks. Please try 'rebase and merge' or so so we can avoid merge commits
('squash and merge' would remove @carlaschroder 's commit which might be not nice to her :-)

PR #5258, opened 2016-10-19
FolderMan cleanup and improvements

ckamm:


src/gui/folderman.cpp
```diff
@@ -766,10 +757,54 @@ void FolderMan::slotServerVersionChanged(Account *account)
...
+        f->scheduleThisFolderSoon();
+    }
+}
+
+void FolderMan::slotScheduleFolderByTime()
+{
+    foreach (auto& f, _folderMap) {
+        // Never schedule if syncing is disabled or when we're currently
+        // querying the server for etags
+        if (!f->canSync() || f->etagJob()) {
```
ogoffart: Discard this comment, when i wrote it i tought the other timer was runing every 5 minutes.

src/gui/folderman.cpp
```diff
@@ -64,11 +64,17 @@ FolderMan::FolderMan(QObject *parent) :
     connect(&_startScheduledSyncTimer, SIGNAL(timeout()),
             SLOT(slotStartScheduledFolderSync()));
 
+    _timeScheduler.setInterval(5000);
```
ckamm: I think it's okay, the slot doesn't do anything intense.

src/gui/folderman.cpp
ckamm: @ogoffart Wouldn't it still make sense to remove the check? Then it'd work even if someone set the etag check interval to 5 seconds.

Issue #2799, opened 2015-02-08
[SSL] [Client certs] Actually display/parse HTML error message


guruz:
Seen this when testing https://github.com/owncloud/client/issues/69 some days ago. Might not be that important anymore since I think we show a dialog now in any case?

```
02-04 17:50:18:182 0x7fabfae05610 networkjobs.cpp:543 error: status.php replied  400 "<html>
<head><title>400 No required SSL certificate was sent</title></head>
<body bgcolor="white">
<center><h1>400 Bad Request</h1></center>
<center>No required SSL certificate was sent</center>
<hr><center>nginx</center>
</body>
</html>
```

jdhoek:
Yes, a dialog is shown when any `https` URL returns any error. Parsing the actual error message might be complicated if HTML is returned (how will you display it?), but perhaps it is sufficient to settle for the HTTP error code?

For example, when a `403` code is returned, add the canonical description for it in the error message presented to the user:

> Could not connect to https://owncloud.example.org/, the server said `403: Forbidden`.

That would at least tell the user something of the nature of the error.

guruz:
@ogoffart would probably recommend to use a simple ```QRegExp``` to take everything in the body that is not a ```<tag>``` (or ```</tag>```)

guruz:
PR is pending..

PR #5261, opened 2016-10-19
Wizard: Show if server asks for client certificate #2799 #69

guruz:
This also nicely displays the 'Untrusted domain' message of oC.
The link to add a trusted domain (via web browser) is clickable.

### DIFF
--- a/src/libsync/networkjobs.cpp
+++ b/src/libsync/networkjobs.cpp
@@ -447,7 +447,7 @@ bool CheckServerJob::finished()
     }
 
     bool success = false;
-    QByteArray body = reply()->readAll();
+    QByteArray body = reply()->peek(4*1024);
     int httpStatus = reply()->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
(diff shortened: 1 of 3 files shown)


guruz:
@ogoffart Please re-evaluate. I extract the message now early so reply is not deleted before.
I also think the sync dialog is ok for now as some lines below we already do a sync dialog too.

src/libsync/networkjobs.cpp
```diff
@@ -447,7 +447,7 @@ bool CheckServerJob::finished()
     }
 
     bool success = false;
-    QByteArray body = reply()->readAll();
+    QByteArray body = reply()->peek(1024);
```
ogoffart: 1024 bytes should be enough for every status.php  replies?

guruz: Not sure. Right now we're at around 110 bytes. Shall I increase it or not?

guruz:
> Why do you show a messagebox instead of displaying the error within the wizard?

No need to mess with the old wizard logic until @danimo has given us his code for the new wizard.

Issue #5260, opened 2016-10-19
Backup Detected message - Client stops working until selection


LukasRo:
Hello. From time to time I get an error message "Backup Detected" while syncing files. Here I can choose "Keep Local Files as Conflict" or "Normal Sync". I can guarantee there was no backup or other tasks that resyncs "old" files, so this seems to be an error.

The bigger problem here is, that the client stops working until I choose one of the options above.

I've installed the client on a Windows Server where it runs and puts the data into a shared folder. When the sync stops because of the message I doesn't get a notice on one of my other clients.. and thats why I have to check the server several times a day.

Maybe there is a hidden config-key like _promptBackupMessage=false_ or sth like this?!
### Expected behaviour

Synching files instead of stoping the sync and waiting for input. Maybe there could ne a switch in the Settings?
### Actual behaviour

stop synching and waiting for me to choose
### Steps to reproduce
1. Waiting for the "Backup Detected"... don't know how to reproduce... but I see it every day.
2. now the sync won't stop until I choose between "keep local files as conflict" or "normal sync"
### Server configuration

Operating system: ‪Ubuntu 14.04.5 LTS‬
Web server: apache
Database: mysql
PHP version: 5.5.9
ownCloud version: ownCloud 9.1.0 (stable)
### Client configuration

Client version: 2.2.4 (build 6408)
Operating system: Windows Server 2008 R2 (same problem on Windows 10 x64)
OS language: German
Installation path of client: C:\Program Files (x86)\ownCloud
### Logs

Server logfile: ownCloud log (data/owncloud.log):
Latest error (many times teh same text in one sec):
[...]

Issue #5109, opened 2016-08-07
Backup Detected message with no accompaning data


mgaman:
Every now & then I get a message pop-up "Backup Detected" due to date conflicts and given a choice between keeping the local file or allowing the sync to go forward.

![untitled](https://cloud.githubusercontent.com/assets/3582286/17460543/2735fc34-5c74-11e6-94e3-36c7706b950b.png)

The pathname of the offending file is not given, so I can't make an informed decision as how to act.
### Server configuration

Operating system:

Web server: 
Linux DietPi 4.4.13-v7+ #894 SMP Mon Jun 13 13:13:27 BST 2016 armv7l GNU/Linux
Database:
Mysql

PHP version:
PHP 5.6.24-0+deb8u1 (cli) 

ownCloud version:
8.1.9

Storage backend:
### Client configuration

Client version:
Version 2.2.2 (build 6192)
Operating system:
Window 10 64 bit

OS language:

Installation path of client:
C:\Program Files (x86)\ownCloud>
### Logs

Please use Gist (https://gist.github.com/) or a similar code paster for longer
logs.

`Template for output < 10 lines`
1. Output of `owncloud --logwindow` or `owncloud --logfile log.txt`
   (On Windows using `cmd.exe`, you might need to first `cd` into the ownCloud directory)
   (See also http://doc.owncloud.org/desktop/2.2/troubleshooting.html#client-logfile )

No ouput
1. Web server error log:
   /var/log/apache2/error.log is empty
2. ownCloud log (data/owncloud.log):
   {"reqId":"Q\/jga\/a221ql6y7bMiMN","remoteAddr":"79.177.220.46","app":"core","message":"imagecreatefrombmp: imagecreatetruecolor failed for file \"\/var\/www\/owncloud\/data\/henryd\/files\/Photos\/mumpics\/Scan.BMP\" with dimensions 466x-663","level":2,"time":"2016-08-07T04:44:46+00:00"}
   {"reqId":"R8d5qBKgi0HzBFI9OWch","remoteAddr":"79.177.220.46","app":"PHP","message":"imagecreatetruecolor(): Invalid image dimensions at \/var\/www\/owncloud\/lib\/private\/image.php#700","level":3,"time":"2016-08-07T04:44:47+00:00"}
[...]

PR #5263, opened 2016-10-21
SyncEngine: disable heuristics for backup restoration for server >= 9.1

ogoffart:
The ownCloud 9.1 server has a data-fingerprint property that the admin must
change in case of backup restoration. When this change, the client understands
that a backup was restored, and will generate conflict files and re-upload
new files.

The heuristics based system checks that there is at least two files wose mtime
is put back in the past and no files that goes forward. In that case we ask the
user before creating the conflicts.

This commit disable the heuristics for newer server that have the data-fingerpint.
And change the heuristics to two hours because we want to avoid false positive due
to some clock error, and that 2 hours of lost due to backup restoration is probably
not so bad.

We only ask the user in the heuristics based aproach so in practice this mean that
the "backup-detected" dialog will no longer appear with newer server.

Relates issues #5260, #5109

### DIFF
--- a/src/libsync/syncengine.cpp
+++ b/src/libsync/syncengine.cpp
@@ -589,11 +589,15 @@ int SyncEngine::treewalkFile( TREE_WALK_FILE *file, bool remote )
             // This counts as a NONE for detecting if all the files on the server were changed
             _hasNoneFiles = true;
         } else if (!isDirectory) {
-            if (std::difftime(file->modtime, file->other.modtime) < 0) {
+            auto difftime = std::difftime(file->modtime, file->other.modtime);
+            if (difftime < -3600*2) {
                 // We are going back on time
+                // We only increment if the difference is more than two hours to avoid clock skew
+                // issues or DST changes. (We simply ignore files that goes in the past less than
+                // two hours for the backup detection heuristics.)
                 _backInTimeFiles++;
                 qDebug() << file->path << "has a timestamp earlier than the local file";
-            } else {
+            } else if (difftime > 0) {
                 _hasForwardInTimeFiles = true;
             }
         }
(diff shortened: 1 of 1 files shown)


guruz:
Looks OK except for the comment i had on IRC

src/libsync/syncengine.cpp
```diff
@@ -589,11 +589,12 @@ int SyncEngine::treewalkFile( TREE_WALK_FILE *file, bool remote )
             // This counts as a NONE for detecting if all the files on the server were changed
             _hasNoneFiles = true;
         } else if (!isDirectory) {
-            if (std::difftime(file->modtime, file->other.modtime) < 0) {
-                // We are going back on time
+            auto difftime = std::difftime(file->modtime, file->other.modtime);
+            if (difftime < -3600) {
+                // We are going back on time (more than two hours to avoid clock skew issues or DST)
```
ckamm: It's only more than one hour!

src/libsync/syncengine.cpp
```diff
@@ -589,11 +589,12 @@ int SyncEngine::treewalkFile( TREE_WALK_FILE *file, bool remote )
...
                 _backInTimeFiles++;
                 qDebug() << file->path << "has a timestamp earlier than the local file";
-            } else {
+            } else if (difftime > 0) {
                 _hasForwardInTimeFiles = true;
             }
```
ckamm: Is it correct to do nothing for the -3600, 0 case? Could you add a comment about that?

ckamm:
@ogoffart The idea is that this situation shouldn't happen anymore with new servers because admins should use the new data-fingerprint based backups? I'm a bit worried about disabling our accidental-deletion heuristic like that... Would there be an easy way to disable it only if a data-fingerprint backup restoration is used?

ogoffart:
> Would there be an easy way to disable it only if a data-fingerprint backup restoration is used?

No, because by default, data-fingerprint is not set.


The problem i'm trying to fix is users complaining about this message that it appears too often when it shouldn't.
We want to rely on the admin setting the data-fingerprint. This heuristics is already flawed as there might be new files already after the backup was recovered.

PR #5266, opened 2016-10-22
Fix find_library FindCMocka.cmake

rmekarni:


ogoffart:
The patch looked trivial enough to accespt it despite the CLA was not signed.

Issue #4608, opened 2016-03-24
Right-click »Share with ownCloud« not available on folders which have not synced 100% yet


jancborchardt:
Version 2.1.1

I have a folder with some vacation pictures I want to share with my friends. I just added some new ones which are syncing.

Now I want to right-click and share the link to that folder already, because that’s possible on the web interface as well and there’s really no reason not to. But it’s not possible, the entry is not there in the menu.

For fully synced folders it works.

cc @rullzer

guruz:
(not that it might matter but..) Which plugin do you use? Nautilus?

rullzer:
I remember adding this on purpose... let me think about why again...

jancborchardt:
It’s in Nautilus.

@rullzer even when it’s on purpose in some cases, this where many of the files are already synced is most definitely a case where generating the link should be possible.

rullzer:
Ah, I remember. It was to avoid erros with files. Because you can't share files before they are properly synced. So we need different file and folder handling. Agreed.

jancborchardt:
Even for files though I would say it should be possible to share it when it’s still about to sync. It’s a slightly different issue, but syncing might just take a few seconds or minutes, while the person actually opening the link might take minutes to hours or days.

ogoffart:
We need to make sure that if the file, or folder, is not fully sync'ed, the user rxperience should be as good as possible.

ckamm:
I tested this and it works after minor further adjustments.

jnweiger:
@ckamm can you please re-check....

jnweiger:
# linux mint tara
daily client 20181014 beta1
![grafik](https://user-images.githubusercontent.com/1108546/44100576-ac1c1832-9fe5-11e8-93a2-0245d34bdfe6.png)

BAD
Unchanged: Still seen in 2.5.0daily20181014: The 'Share and 'create public link' entries are not there, when the folder is still doing its initial sync.

ckamm:
This must be an odd corner case, I'm looking into it. The code intends for this ui state to only be shown if the file: is on the server, does have permissions set, and does not have the reshare permission bit.

ckamm:
@jnweiger Works for me now!

michaelstingl:
LGTM

![cursor_and_item-0_and_applevolumeextra_and_airportextra_and_displaysextra](https://user-images.githubusercontent.com/214010/44998359-4e283780-afb5-11e8-9727-3f8afa223904.png)

PR #5267, opened 2016-10-22
nautilus integration: "share" extension for syncing folders

payload:
Issue #4608
- folder state is either OK or SYNC usually
- this extension does not know about if a folder is shareable
- a OK folder is shareable
- a SYNC folder could be uploaded or not uploaded
- this commit looks into file entries below this folder
- if one exists with state OK or SYNC
  this folder must have been uploaded
- better would be if the server (gui) tells us
  if the folder is uploaded

### DIFF
--- a/shell_integration/nautilus/syncstate.py
+++ b/shell_integration/nautilus/syncstate.py
@@ -18,6 +18,7 @@
 import os
 import urllib
 import socket
+import tempfile
 
 from gi.repository import GObject, Nautilus
 
@@ -43,7 +44,7 @@ def get_runtime_dir():
     try:
         return os.environ['XDG_RUNTIME_DIR']
     except KeyError:
-        fallback = '/tmp/runtime-' + os.environ['USER']
+        fallback = os.path.join(tempfile.gettempdir(), 'runtime-' + os.environ['USER'])
         return fallback
 
 
@@ -86,9 +87,9 @@ class SocketConnect(GObject.GObject):
     def _connectToSocketServer(self):
         try:
             self._sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
-            postfix = "/" + appname + "/socket"  # Should use os.path.join instead
-            sock_file = get_runtime_dir() + postfix
-            print ("Socket: " + sock_file + " <=> " + postfix)
+            postfix = os.path.join(appname, "socket")
+            sock_file = os.path.join(get_runtime_dir(), postfix)
+            print ("Socket: " + sock_file + " <=> /" + postfix)
             if sock_file != postfix:
                 try:
                     print("Socket File: " + sock_file)
@@ -187,9 +188,9 @@ class MenuExtension(GObject.GObject, Nautilus.MenuProvider):
         filename = get_local_path(file.get_uri())
         # Check if its a folder (ends with an /), if yes add a "/"
         # otherwise it will not find the entry in the table
-        isDir = os.path.isdir(filename + "/")
+        isDir = os.path.isdir(filename + os.sep)
         if isDir:
-            filename += "/"
+            filename += os.sep
 
         # Check if toplevel folder, we need to ignore those as they cannot be shared
         topLevelFolder, internalFile = self.check_registered_paths(filename)
@@ -336,7 +337,7 @@ class SyncStateExtension(GObject.GObject, Nautilus.ColumnProvider, Nautilus.Info
 
         filename = get_local_path(item.get_uri())
         if item.is_directory():
-            filename += '/'
+            filename += os.sep
 
         inScope = False
         for reg_path in socketConnect.registered_paths:
--- a/shell_integration/nautilus/syncstate.py
+++ b/shell_integration/nautilus/syncstate.py
@@ -157,32 +157,65 @@ class MenuExtension(GObject.GObject, Nautilus.MenuProvider):
     def __init__(self):
         GObject.GObject.__init__(self)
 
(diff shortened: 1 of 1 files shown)


payload:
Ultimate problem is that the server (GUI) is not responding with a detailed status. It tells us OK and SYNC but not if it is already present on the ownCloud and shareable. Correct? How to attack this problem?

shell_integration/nautilus/syncstate.py
```diff
@@ -157,32 +157,65 @@ class MenuExtension(GObject.GObject, Nautilus.MenuProvider):
...
-            # ignore top level folders
-            if filename.startswith(reg_path) and topLevelFolder == False and socketConnect.nautilusVFSFile_table[filename]['state'].startswith('OK'):
-                syncedFile = True
-
-        # If it is neither in a synced folder or is a directory
-        if not syncedFile:
+        filename = get_local_path(file.get_uri())
+        # Check if its a folder (ends with an /), if yes add a "/"
+        # otherwise it will not find the entry in the table
+        isDir = os.path.isdir(filename + "/")
```
Naereen: Please consider using `os.path.join(a, "/")` instead of `a + "/"`.

shell_integration/nautilus/syncstate.py
```diff
@@ -157,32 +157,65 @@ class MenuExtension(GObject.GObject, Nautilus.MenuProvider):
...
+        elif state_sync and isDir:
+            # some file below is OK or SYNC
+            for key, value in socketConnect.nautilusVFSFile_table.items():
+                if key != filename and key.startswith(filename):
+                    state = value['state']
+                    if state.startswith('OK') or state.startswith('SYNC'):
+                        shareable = True
+                        break
+
+        if not shareable:
```
Naereen: All this seems good to me.

shell_integration/nautilus/syncstate.py
payload: I know what you mean. `os.path.isdir(filename + os.sep)` or `os.path.isdir(os.path.join(filename, ""))`. Can do.

shell_integration/nautilus/syncstate.py
```diff
@@ -86,9 +87,9 @@ def addListener(self, listener):
     def _connectToSocketServer(self):
         try:
             self._sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
-            postfix = "/" + appname + "/socket"  # Should use os.path.join instead
-            sock_file = get_runtime_dir() + postfix
-            print ("Socket: " + sock_file + " <=> " + postfix)
+            postfix = os.path.join(appname, "socket")
```
Naereen: Does this work like you think?
It was `/appname/...` now it is probably `appname/...`: the leading `/` is not there?

payload: Ah yes, I overlooked the sock_file test, sorry. Well it is difficult to say what was meant the `if`. The original checks if the concat with `get_runtime_dir()` introduces some change, else it is an invalid `sock_file`. Here are examples for the edge cases:

>>> os.path.join("", "foo/bar")
'foo/bar'
>>> os.path.join("/", "foo/bar")
'/foo/bar'
>>> os.path.join("/", "/foo/bar")
'/foo/bar'

I should probably be on the safe side and don't introduce a behaviour change. Can I amend the fix to this commit and force push this branch?

payload: I looked more closely on this and the check for a valid "Sock-File" isn't making sense. There is a `try` to catch illegal accesses to an invalid socket file (for example permission denied). The code just checked if `get_runtime_dir()` returned empty string and deemed it invalid. All else returns of `get_runtime_dir()` would succeed. I kill the code and make it clean.

Naereen:
@ckamm agreed about the possible expensive operation.
It should be done differently, I guess it can be done by looking at the owncloud special files (`.owncloudsync.log` ?).

ckamm:
@Naereen No, `.owncloudsync.log` is only in the root folder. The database knows, of course, but we probably don't want to extend the protocol for this. I wonder whether we should just ignore the problem, always show the context menu button for SYNC folders and depend on error messages in the folder-share dialog.

@payload What do you think? Did you make a fix for this issue because it's something that is regularly problematic for you?

payload:
@ckamm It is a regular problem I had. It is a fix I made and personally use for newly created folders with media files inside of them. My workflow is like moving a folder of videos into a ownCloud folder. I am used to use the web interface for sharing but since I discovered the nautilus context menu option to share a file without logging into the web interface, I want to use it.

The expensiveness is something I don't notice in my usage, but yes, it is definitely not efficient and won't scale for huge folders. Good point, that it only checks it on selecting the folder if I observed it correctly.

I don't know of special files, so I can't comment on that but that parsing such a file would be a hidden dependency on its layout of content and therefore a dirty hack.

IMHO the protocol seems not easily extendible in the RETRIEVE_FOLDER_STATUS response. Some parts of some software depend on exact status strings and not only on "starts with" or "contains" some string. I don't know if this is intended.

I prefer to always show the context menu button and handle shareability in the dialog. It makes the feature always visible, so it can be discovered. It isn't just disappearing without notice, so less possibility for confusion by users who already discovered the feature but don't know the connection to the SYNC state of a folder or file. The dialog has the opportunity to show a status message if the file is not yet registered, sheduled, syncing or ok or whatever. The dialog can share a folder which has syncing files in it.

ogoffart:
Thanks for the pull request!

Issue #1633, opened 2014-04-02
Propagator: Balance concurrent up/down-load


guruz:
When synchronizing, it makes sense to try to keep at least 1 PUT and 1 GET running at the same time so we don't actually waste time by not trying to use the connection both upstream and downstream at its fullest.

(MOVE, MKCOL, DELETE do not matter in this issue I'd say)

mrow4a:
Awesome, so the concept exists. Nice! Will do that surely.

mrow4a:
@guruz @ogoffart One question, why do we actually keep a list of all finished jobs, instead of just destroying them?

ogoffart:
> why do we actually keep a list of all finished jobs

We probably should delete them indeed.

mrow4a:
Ok. will see what we can do about it. 

@ogoffart @guruz 
I actualy have also anotheridea about the file scheduling. This could be done both inside the folder and outside folders. 

During the loop over all files to synchronise, we structurise our directories so that they are being synced in order of insertion, does not matter is specific tree contains 100GB of data to sync, and another 1MB of very important document, but added to sync list later. 

We could do it a little bit smarter. After we have all the "directories" object, we can give the folder ordered sync instructions, with which files/folders to start within that specific folder. 

Imagine you have following folder structure to sync:

In current implementation, these files will be synced as they are met in the structure below:

ownCloud Root
   ├── less-important-folder-10MB/
   │   └── less-important-file
   ├── Home&Work-10GB/               
   │   └── folder with nested 10GB of photos to be sync
   │   └── very-very-important folder with single documents to sync
   │   └── less-important-document - 1MB
   ├── important-folder-12MB/
   ├── less-important-folder - 5MB/
   │   └── less-important-document 
   └── important-file-1MB 

However, post-loop could quickly restructurise it as follows, looking of size of files within to sync (of course not by priority, because ownCloud does not support sth like FOLDER PRIORITIES:

ownCloud Root
   └── important-file-1MB 
   ├── less-important-folder-5MB/
   ├── less-important-folder-10MB/
   │   └── less-important-file
   ├── important-folder-12MB/
      └── less-important-document 
   ├── Home&Work-10GB/
        └── less-important-document - 1MB               
        └── very-very-important folder with single documents to sync - 2MB
        └── folder with nested 10GB of photos to be sync

PR #5269, opened 2016-10-23
Remove excesive loops and add sync files scheduler class - 1000 files, reduction from half milion to three thousand loops

mrow4a:
Hello guys: 

While working on bundling, I analysed a little bit of code, and I found these little flower there, which will reduce the number of loops significantly, in this example from 506000 to 3000 for 1000 files. However, this implementation also has something more in mind, but I will explain it later, please verify and do some QA on this small solution: 

MOTIVATION:
- mkdir, move, delete, ignore jobs, and specialy mkdir jobs should be running with checking status of subjobs (I think), while new and update jobs can be being run in the manner proposed in the code

The only change here is that we actualy extract jobs from the queue and execute them, not checking the status of all jobs which were already finished - please refer to the code at first picture.

- This will give me an easy way to insert bundling for files, or some nice schedulers which will create 3 parallel pipes of files - upload, download and mixed upload/download of small files. If this will pass, I will propose new code quite quickly - I will just copy paste this concept: https://en.wikipedia.org/wiki/Round-robin_scheduling

Test on localhost, I have added one folder and 1000 files of 1kB:
In the old version you have required to run around half a million loops to execute this code:
loop_half_milion

This is my version with around three thousands loops and 1000 files of 1kB, sync also took a little bit shorter, from 58s to 50s on my machine, and took much less CPU
looped_3000

@dragotin @DeepDiver1975 @ogoffart @guruz


mrow4a:
Anyways we should get rid of this ecpobential/kogaritmic looo vecause for 1k files it is 500k loops, but for 10k files it is 50mln loops

felixboehm:
Good catch @mrow4a.

michaelstingl:
@guruz @ogoffart Could we add this to [2.3.0](https://github.com/owncloud/client/milestone/31) ?

src/libsync/owncloudpropagator.h
```diff
@@ -187,34 +188,42 @@ class OWNCLOUDSYNC_EXPORT PropagateDirectory : public PropagatorJob {
     // e.g: create the directory
     QScopedPointer<PropagateItemJob>_firstJob;
 
-    // all the sub files or sub directories.
-    QVector<PropagatorJob *> _subJobs;
+    // all the new and changed files without conflicts.
+    QLinkedList<PropagatorJob *> _syncJobs;
```
guruz: AFAIK not ideal because of heap fragmentation, but @ogoffart can say more to this.

guruz:
Amazing. High CPU usage has plagued us there for quite some time!
I don't know how easy it is to break stuff here though.

How does the memory usage compare? Because @jturcotte had worked hard to try to lower this.

Can you look at this please?
```
The following tests FAILED:
	 16 - SyncEngineTest (Failed)
	 17 - SyncFileStatusTrackerTest (Failed)
Errors while running CTest
```

@ogoffart needs to have a look at this from the correctness perspective :)

src/libsync/owncloudpropagator.h
mrow4a: It should be fine, I have choosen LinkedList to be able to easily put jobs both at the end and begining of the list. I need it for bundling and stuff. You know, everywhere there are pros and cons. Lets discuss it, but this is implementation detail

mrow4a:
I guess there are more flowers like that in the code, I will look for more later. I guess another friends like that are in the discovery.

src/libsync/owncloudpropagator.cpp
```diff
@@ -608,22 +610,23 @@ bool PropagateDirectory::scheduleNextJob()
 
     bool stopAtDirectory = false;
     // FIXME: use the cached value of finished job
```
ogoffart: Here is what you need to do ro remove the O(N^2)   cache the value of the first unfinished job and start from here instead of always starting from 0

src/libsync/owncloudpropagator.h
ogoffart: In that case, use QList, which reserve space at the begining and at the end.

ogoffart:
I did not review the wole thing, but it seems you are leaking when the sync abort prematuraly. 

There is indeed a O(n^2) complexity. But if i understand correctly what you are doing is to having two lists to reduce the n.  But the real fix is what the FIXME is saying:  cache the value so one does not starts always at 0

mrow4a:
@ogoffart I did not touch this implementation of starting from 0 since I though you had some idea behind it/bug scenario. I can also add this caching thing if you say that does not matter. I would anyways go with the other class syncing files at the end, since I want to add another feature there - job scheduler which will try to fill upload/download bandwidth, running this jobs in parallel. In the PropagateSyncFIles job I want to create like few queues from which I will extract in round robin scheduler way.

mrow4a:
Ok, I pushed new version. This should solve most of the concerns and implementation is very easy and lightweight. It should also reduce memory usage. 

Caching (so CPU saving) I will add in another PR. 

@guruz @jturcotte @DeepDiver1975 @ogoffart

mrow4a:
@ogoffart I actualy have a problem caching this _lastFinishedJob for this loop in the other branch. Did you already try to do that? I am running in very strange race conditions.

ckamm:
@mrow4a I've read through the patch and don't understand yet why having the split into subJobs and syncJobs is necessary. Wouldn't removing all those finished jobs from _subJobs (or, less invasively, keeping track of the first interesting index) have the same effect?

I'm assuming the main thing you want to get rid of is iterating over all these finished jobs all the time?

mrow4a:
Caching is yet another problem, separate PR. In this PR it was by the way reducing this loops,  actually want to have clear separation of download/upload jobs and other jobs, which will ease pain of adding new features.

ckamm:
@mrow4a Thanks for explaining, I thought reducing pointless looping was your main objective.

About concurrency: Only the GUI thread calls scheduleNextJob. All threaded concurrency happens purely inside the QNetworkManager when propagation jobs run network tasks.

michaelstingl:
@mrow4a Could you build a Win testclient with your improvements on rotor? A user offered help to test-drive this?

ogoffart:
Is this still relevant now that #5274 was integrated?

mrow4a:
@ogoffart You suggested to do everything in the PropagateDirectory, I will try to fit everything there, along with Bundling. Lets see how it will work. Otherwise it will end up with sth like here if it will be too big mess in PropagateDirectory. This class is already complicated, not including anything else.

mrow4a:
https://github.com/owncloud/client/issues/1633

PR #5271, opened 2016-10-24
Linktocentral

Kawohl:
added a link to central owncloud.org in owncloudtheme.cpp for the settings overview

### DIFF
--- a/src/libsync/owncloudtheme.cpp
+++ b/src/libsync/owncloudtheme.cpp
@@ -45,8 +45,9 @@ QString ownCloudTheme::about() const
 {
     QString devString;
     devString = trUtf8("<p>Version %2. For more information visit <a href=\"%3\">%4</a></p>"
-               "<p><small>By Klaas Freitag, Daniel Molkentin, Jan-Christoph Borchardt, "
-               "Olivier Goffart, Markus Götz and others.</small></p>"
+               "<p>For known issues and help, please visit: <a href=https://central.owncloud.org/c/help/desktop-file-sync>central.owncloud.org</a></p>"
+               "<p><small>By Klaas Freitag, Daniel Molkentin, Olivier Goffart, Markus Götz, "
+               " Jan-Christoph Borchardt, and others.</small></p>"
                "<p>Copyright ownCloud GmbH</p>"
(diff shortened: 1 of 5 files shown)


src/libsync/owncloudtheme.cpp
```diff
@@ -45,8 +45,9 @@ QString ownCloudTheme::about() const
 {
     QString devString;
     devString = trUtf8("<p>Version %2. For more information visit <a href=\"%3\">%4</a></p>"
-               "<p><small>By Klaas Freitag, Daniel Molkentin, Jan-Christoph Borchardt, "
-               "Olivier Goffart, Markus Götz and others.</small></p>"
+               "<p>For known issues and help, please visit: <a href=https://central.owncloud.org/c/help/desktop-file-sync>central.owncloud.org</a></p>"
+               "<p><small>By Klaas Freitag, Daniel Molkentin, Olivier Goffart, Markus Götz, "
```
dragotin: Might be just me being old-school, but I more prefer reading urls including the protocol in user visible strings, that would mean `central.o.o` => `https://central.o.o`.... just if you want.

Kawohl: fine with me!

guruz: The "" are missing around the https:// .. :)

guruz:
(To whoever merging this after @Kawohl fixes the " issue: Please use "Squash and merge" for clean history)

Kawohl:
imho the : ""  are unnecessary and confusing... The Links work and are clickable
please explain what can be done better and why.

Best John

guruz:
@Kawohl but it's not valid XML/HTML.. I can fix it myself and merge https://github.com/owncloud/client/pull/5271/files#diff-5c233b67dd98ce7e8ea3ffbe0a4b3547R48

felixboehm:
@Kawohl welcome to the show!

src/libsync/owncloudtheme.cpp
Kawohl: @guruz i get the line but i just can't figure out the right way... could you please point me somewhere for an example?
Thanks!

PR #5272, opened 2016-10-25
License: Adjust license of GPLv2 source files to GPLv2+

ckamm:
This change of license needs approval by someone from owncloud @jnweiger @hodyroff? 

See #5180: It seems to be the case that mixing GPLv2 and GPLv2+ was an accidental mistake. This change relicenses every source file that used to be GPLv2 to GPLv2+.

### DIFF
--- a/admin/osx/macdeployqt.py
+++ b/admin/osx/macdeployqt.py
@@ -5,7 +5,8 @@
 #
 #  ownCloud is free software: you can redistribute it and/or modify
 #  it under the terms of the GNU General Public License as published by
-#  the Free Software Foundation; version 2 of the License.
+#  the Free Software Foundation; either version 2 of the License, or
+#  (at your option) any later version.
 #
 #  ownCLoud is distributed in the hope that it will be useful,
 #  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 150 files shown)


ckamm:
@jnweiger Note that your comment in #5180 said LGPL. Currently none of the code is LGPL.

guruz:
@hodyroff ping..

hodyroff:
Approved. Need to add to the dependency list. libcrashreporter-qt is only used in the Windows built and if needed could be taken out. Any other GPL code included needs special approval.

ckamm:
@hodyroff I don't fully understand. Are there implications for this change?
* What needs to be added to the dependency list? Is this something that needs to be done along with this change?
* Your statement is that including any foreign GPL code that is not covered by the contributor agreement needs approval, correct? (contributors would be adding GPLv2+ code all the time)

jnweiger:
@ckamm Holger explained me the background yesterday. We have two objectives:

* A) we want the client to be an open source offer with consistent licensing, both within the code base, and with regards to 3rd party components we want to link with. (this PR addresses a GPLv2-only versus GPLv3-only issue).

* B) the contributors agreement is supposed to allow us to offer client code (or parts of the client code) *also* under a closed source license. The scenario I have in mind is a partner who wants to integrate prorietary code with the client for his specific use case. GPL would ask the partner to open source his proprietary code, making this offer impossible. 

The crashreporter component would not work with B) as GPL applies there. We consider crashreporter an optional component and would drop it to make B) possible. We need to keep track of all components that are under a strong license but not under CLA. That is why we need a review/approval step and a list.

@hodyroff please double check, if that describes what we discussed.

PR #5274, opened 2016-10-25
Cache the value of last unfinished job in the PropagateDerectory scheduleNextJob 

mrow4a:
This PR solves the problem discussed in this PR https://github.com/owncloud/client/pull/5269, implementing the solution of @ckamm which appears to work :

I did not run in any race condition and manualy tested most of the cases. Hope this will work out

@guruz @ogoffart

### DIFF
--- a/src/libsync/owncloudpropagator.cpp
+++ b/src/libsync/owncloudpropagator.cpp
@@ -606,9 +606,15 @@ bool PropagateDirectory::scheduleNextJob()
         return false;
     }
 
+    // cache the value of first unfinished subjob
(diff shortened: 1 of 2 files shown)


dragotin:
On a more general note: We're using `int` type as index here. If we have a *really* huge directory, couldn't we have more entries than MAX(int)? Not sure tough if QVector would survive that, or everything....

ogoffart:
Looks good (appart from the typo in the comment)

mrow4a:
Ok, done. Ready to merge if checks will pass.

PR #5277, opened 2016-10-26
Doc: Add sync algorithm overview and comments

ckamm:


doc/dev/sync-algorithm.md
```diff
@@ -0,0 +1,79 @@
...
+    - csync_update() uses csync_ftw() on the local and remote trees, one after the other.
+    - csync_ftw() iterates through the entities in a tree and calls csync_walker() for each.
+    - csync_walker() calls _csync_detect_update() on each.
+    - _csync_detect_update() compares the item to its db entry (if any) to detect new, changed or renamed files. This is the main function of this pass.
+
+
+
+### Reconcile
+
+The reconcile phase compares and adjusts the local and remote trees (in both directions), detecting conflicts.
```
dragotin: For me, the key thing of the reconcile phase is the compilation of the instruction set for the propagator out of the two states of the local and remote tree. So maybe this might be changed to 
"The reconcile phase compares the local and remote tree (in both directions), and computes an instruction for the propagation phase for each file or directory. It also detects conflicts."

dragotin:
This is awesome. :+1:

Two remarks: keep in mind that there is the sync protocol documentation at https://github.com/cernbox/smashbox/blob/master/protocol/protocol.md and also the normal documentation in the clients source repo. Merging all into one might be useful on the longer run.

Apart from that, you use the word *database*, however, I think the term *syncjournal* is widely used, which I personally like more (which means nothing) and which might be a bit more specific for the usage.

doc/dev/sync-algorithm.md
ogoffart: That's the "post reconcile" phases which does give information for the propagator.

ckamm:
@dragotin @ogoffart Thank you for your comments.

* I've replaced db/database with sync journal.
* Some comments mention 'replica' - is that what I call 'tree' or something else?
* I'm not wedded to the location of the documentation - would there be a better place? I like it to be in-source.
* It'd be great to have similar technical documentation on the propagator (including job scheduling) and then link the sync protocol documentation from there. I don't think it fits with the sync algorithm documentation.

ogoffart:
> I've replaced db/database with sync journal.

I think database is more appropriate than journal. Journal implies chronilogical records of changes. More like a log. 
Despite the class being named "SyncJournalDB", it is just a database and does not have a record of the changes.  If anything, the class should be renamed to SyncDatabase.

> Some comments mention 'replica' - is that what I call 'tree' or something else?

"Tree" is correct because it's store in a red-black tree (with a hash as a key). But this is an implementation detail. And it convey the idea that the tree-structure of the file system is preserved in the data structure, while it is not.

Replica is not that bad. Or maybe "Representation" ?

>I'm not wedded to the location of the documentation - would there be a better place? I like it to be in-source.

I think it's fine.

> I don't think [the protocol documentation] fits with the sync algorithm documentation.

I agree, the protocol documentation is targeted at anyone implementing a server or client (sync or not), compatible with ownCloud.  This documentation only explain the algorithms for people trying to hack on the client.

guruz:
Good initiative :)

> I don't think [the protocol documentation] fits with the sync algorithm documentation.

Could you still link to it? To make it easier for newcomers to dive into all areas..?

ckamm:
@ogoffart @dragotin 

journal/database: let's keep "journal" for now, because that's what's in the code. I'm up for renaming it to database everywhere, but I don't want to add inconsistent docs.

tree/replica: I meant the local tree of files and directories. I'll make this more explicit.

@guruz Will add the protocol doc link.

ckamm:
Ok to merge?

PR #5285, opened 2016-10-29
Add manifest file on Windows to make the application UAC aware.

proog128:
Added RT_MANIFEST resource on Windows to make the application UAC aware. This disables UAC virtualization which caused problems on my machine.

### DIFF
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -199,6 +199,9 @@ IF( WIN32 )
       ${CMAKE_CURRENT_BINARY_DIR}/version.rc
       @ONLY)
     set(client_version ${CMAKE_CURRENT_BINARY_DIR}/version.rc)
+    IF(NOT MSVC)
+        set(client_manifest ${CMAKE_CURRENT_SOURCE_DIR}/manifest-mingw.rc)
+    ENDIF()
 ENDIF()
 
 set( final_src
@@ -206,6 +209,7 @@ set( final_src
     ${client_SRCS}
     ${client_UI_SRCS}
     ${client_version}
+    ${client_manifest}
     ${guiMoc}
     ${client_I18N}
     ${3rdparty_SRC}
--- a/src/gui/manifest-mingw.rc
+++ b/src/gui/manifest-mingw.rc
@@ -1,0 +1,3 @@
+#include "winuser.h"
+
+CREATEPROCESS_MANIFEST_RESOURCE_ID RT_MANIFEST "owncloud.exe.manifest-mingw"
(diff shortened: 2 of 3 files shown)


dragotin:
@proog128 thanks for your contribution. Before we can merge this it would be awesome if you could explain a bit why it is needed. And doesn't that need to be included into the NSIS packaging?

proog128:
If the manifest is missing, UAC (User Account Control) on Windows virtualizes access to some files and registry keys. This allows legacy applications to run without permission to write e.g. to the Program Files folder, because the file access is redirected to the VirtualStore folder in AppData\Local. The application will not fail with Access Denied.

The ownCloud desktop client does not access resoures it is not allowed to access, for example settings are stored in AppData\Local. Therefore, UAC virtualization should not have any negative effects. However, on my machine I got Access Denied errors when the client wanted to write the settings file. It was not possible to store settings, I had to configure the server URL each time the client started. I could not find out why this is the case on this machine (I could not reproduce it anywhere else), but adding the manifest to disable UAC virtualization solved the problem. Generally, it seems to be a good idea to always embed a manifest, for example MS Visual Studio does that automatically.

The manifest is embedded into owncloud.exe as a resource, so I think it should not affect packaging. Right? I did not try building the installer.

guruz:
Why does @CLAassistant think that @proog128 is not a github user and did not sign the CLA?

> Tobias Haeussler seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA.

guruz:
Thanks for the contribution.

@dschmidt @danimo Do you have input on those topics?

dschmidt:
Tbh I dont know anything about it, but the explanations sound fine.

Do we need it just for the main .exe? I dont know, is there anything implemented as external process or something right now?

danimo:
LGTM. I had this as a patch once but it turned out not to fix the issue at hand. That said, virtualization is nasty, and causes unexpected effects (because writes to the registry that should fail will not, etc).

danimo:
Btw: Qt adds a manifest for MSVC, but does not seem to do that for MinGW, at least for our cross-compile setup. Keep in mind when you consider switching.

ogoffart:
Thanks for the merge request.

guruz:
@SamuAlfageme something to have a look when you have access to different Windows versions :)
1) Does oC still install nicely on all those versions?
3) Does creating the sync folder and syncing work?

SamuAlfageme:
@guruz I'm on it. Have a couple of those versions ready 😄 

(Thanks for assigning, dunno why I missed the mention)

SamuAlfageme:
Used `ownCloud-2.3.0.6552-nightly20161207-setup.exe` to test and everything is looking bright and green: 

|                 | Windows 7 - SP1 | Windows 8.1 | Windows 10 |
|-----------------|:---------------:|:-----------:|:----------:|
| 1. Installation |       ✅        |      ✅      |     ✅     |
| 2. Basic syncing|       ✅        |      ✅      |     ✅     |

Any reasons to extend this to XP/Vista? (Haven't included those 'cause MS is pulling support from them)

guruz:
@SamuAlfageme Not needed IMHO, thanks. If it makes problems there people can still sync with 2.2.4

Issue #5213, opened 2016-09-28
Replace p12topem.* with QSslCertificate::importPkcs12


guruz:
See https://code.woboq.org/qt5/qtbase/src/network/ssl/qsslcertificate.cpp.html#583

FYI @qknight @nocteau @Raptormagnum @jdhoek @dragotin 

Any volunteers? I don't have a client cert setup.
Does anyone want to do this?
Or give me a cert/key and account?

This is to clean up for https://github.com/owncloud/client/issues/69

guruz:
Done in https://github.com/owncloud/client/pull/5289
Please everyone review this PR

PR #5289, opened 2016-10-31
Fix up SSL client certificates #5213 #69

guruz:
The re-enables the UI, uses Qt API for importing and
stores the certificate/key in the keychain.


src/libsync/connectionvalidator.cpp
```diff
@@ -144,7 +144,12 @@ void ConnectionValidator::slotStatusFound(const QUrl&url, const QVariantMap &inf
 // status.php could not be loaded (network or server issue!).
 void ConnectionValidator::slotNoStatusFound(QNetworkReply *reply)
 {
-    qDebug() << Q_FUNC_INFO << reply->error() << reply->errorString();
+    qDebug() << Q_FUNC_INFO << reply->error() << reply->errorString() << reply->peek(1024);
```
guruz: @ogoffart @ckamm @dragotin Would you judge this as OK? On startup..
The server (nginx at least) sends a HTTP 400 if there is no cert sent (for status.php). Then we'll load the credentials and retry (the connection validator seems to retry if reportResult that) when the cert (and other stuff) is fetched from keychain. Then it works..

Does this loading-from-keychain somehow interfere with SAML/Shibboleth or anything overriding the auth via theme? I guess not since the type of credentials is set in the cfg and not keychain?

src/gui/wizard/owncloudconnectionmethoddialog.cpp
```diff
@@ -29,8 +29,11 @@ OwncloudConnectionMethodDialog::OwncloudConnectionMethodDialog(QWidget *parent)
     connect(ui->btnClientSideTLS, SIGNAL(clicked(bool)), this, SLOT(returnClientSideTLS()));
     connect(ui->btnBack, SIGNAL(clicked(bool)), this, SLOT(returnBack()));
 
-    // DM: TLS Client Cert GUI support disabled for now
+
+#if QT_VERSION < QT_VERSION_CHECK(5, 4, 0)
```
ckamm: Must be >= 5.4.0?!

src/gui/wizard/owncloudsetuppage.cpp
```diff
@@ -320,36 +316,40 @@ QString subjectInfoHelper(const QSslCertificate& cert, const QByteArray &qa)
...
+                                            &clientCaCertificates,
+                                            addCertDial->getCertificatePasswd().toLocal8Bit())){
+        AccountPtr acc = _ocWizard->account();
 
-    resultP12ToPem certif = p12ToPem(addCertDial->getCertificatePath().toStdString() , addCertDial->getCertificatePasswd().toStdString());
-    if(certif.ReturnCode){
-        QString s = QString::fromStdString(certif.Certificate);
-        QByteArray ba = s.toLocal8Bit();
+        // to re-create the session ticket because we added a key/cert
+        acc->setSslConfiguration(QSslConfiguration());
```
ckamm: Does this line do anything?

src/libsync/connectionvalidator.cpp
ckamm: I don't know about this. Could we generally require that credentials must be loaded before we attempt to connect?

src/libsync/creds/httpcredentials.cpp
```diff
@@ -36,8 +37,8 @@ namespace OCC
 namespace
 {
 const char userC[] = "user";
-const char certifPathC[] = "certificatePath";
-const char certifPasswdC[] = "certificatePasswd";
+const char clientCertificatePEMC[] = "_clientCertificatePEM";
+const char clientKeyPEMC[] = "_clientKeyPEM";
```
ckamm: This means this change will force current users of client certificates to re-setup? (probably okay)

src/libsync/creds/httpcredentials.cpp
```diff
@@ -129,35 +143,83 @@ void HttpCredentials::fetchFromKeychain()
...
-        key = QString::fromLatin1( "%1/type" ).arg( kck );
-        settings->remove(key);
-        settings->sync();
-    }
-
     if (_ready) {
         Q_EMIT fetched();
     } else {
+        // Read client cert from keychain
+        const QString kck = keychainKey(_account->url().toString(), _user + clientCertificatePEMC);
```
ckamm: Is it correct that the PEM *and* the key are both considered secrets?

src/libsync/connectionvalidator.cpp
guruz: Yes, maybe it's better if this is generally needed, so we'd always load the credentials at the start from keychain.
@ogoffart @dragotin @danimo Any input?

src/gui/wizard/owncloudconnectionmethoddialog.cpp
guruz: No, for old versions we want to hide this button to not allow the user to configure anything.
This is based on 3dd6bcc3233cd7c4ad88eecb43af726e11869601

src/libsync/creds/httpcredentials.cpp
guruz: Yes... IMHO that's ok because it was always an undocumented feature

ogoffart: on the other hand,  why changing?

src/libsync/creds/httpcredentials.cpp
```diff
@@ -129,35 +143,83 @@ void HttpCredentials::fetchFromKeychain()
...
+}
+
+void HttpCredentials::slotReadClientKeyPEMJobDone(QKeychain::Job* incoming)
+{
+    // Store key in memory
+    ReadPasswordJob *readJob = static_cast<ReadPasswordJob*>(incoming);
+
+    if (readJob->error() == NoError && readJob->binaryData().length() > 0) {
+        QByteArray clientKeyPEM = readJob->binaryData();
+        // FIXME Unfortunately Qt has a bug and we can't just use QSsl::Opaque to let it
```
ogoffart: Has the bug been reported?  If yes, a reference to the bug number here would be nice so we can remove that code in some year when we see it is fixed in the version we support.

src/libsync/creds/httpcredentials.cpp
guruz: Because the config file is not a secure place to store a private key

src/gui/wizard/owncloudsetuppage.cpp
ckamm: @guruz Yes, right. Setting a null configuration will cause `getOrCreateSslConfig` to build a fresh configuration! All good.

PR #5294, opened 2016-11-02
Macro to handle legacy OpenSSL SSLeay_add_all_algorithms()

Germano0:
Hi, I am one of the Fedora Owncloud-Client maintainers.
Fedora Rawhide (26) ships OpenSSL 1.1.0 that no longer permits to call the old `SSLeay_add_all_algorithms()` that has been replaced with `OpenSSL_add_all_algorithms()` since OpenSSL version 0.9.5. So we started [getting build failures](https://kojipkgs.fedoraproject.org//work/tasks/9827/16189827/build.log)
I solved by adding a macro that calls
- `SSLeay_add_all_algorithms()` if the system uses OpenSSL <= 0.9.4
- `OpenSSL_add_all_algorithms()` if the system uses OpenSSL > 0.9.4
The hexadecimal code for OpenSSL 0.9.4 version has been taken from [here] (https://github.com/openssl/openssl/blob/6f0ac0e2f27d9240516edb9a23b7863e7ad02898/include/openssl/opensslv.h#L29)

Best regards
Germano Massullo

### DIFF
--- a/src/3rdparty/certificates/p12topem.cpp
+++ b/src/3rdparty/certificates/p12topem.cpp
@@ -65,7 +65,7 @@ resultP12ToPem p12ToPem(string p12File, string p12Passwd) {
     ret.PrivateKey = "";
     ret.Certificate = "";
     
-    SSLeay_add_all_algorithms();
+    ADD_ALL_ALGORITHMS();
     ERR_load_crypto_strings();
     if(!(fp = fopen(p12File.c_str(), "rb"))) {
         ret.ErrorCode = 1;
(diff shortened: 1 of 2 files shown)


guruz:
Thanks for the PR.
After merging https://github.com/owncloud/client/pull/5289/files we would not need your PR anymore, sorry :-)

I'll keep it here still since we'd need to have it if we ever decide to do a 2.2.5 then we could cherry-pick your PR there.

ogoffart:
@guruz:  maybe the patch that gets rid of p12topem.cpp could be merged now already?

guruz:
2.3.0 is released, thanks anyway :)

Issue #5282, opened 2016-10-28
owncloud folder is shared across different user accounts (permissions issue)


1951FDG:
### Expected behavior

For example, if you create a standard account or admin account, you can see the contents of the owncloud folder for other users, this is of security concern. Proper permissions should be setup for the owncloud folder. Antoher example, if we take a computer to a an Apple store for repair and create a standard account for Apple for testing purposes, Apple techs using that account could still see files and folders within the owncloud folder of the account belonging to the owner.
### Actual behavior

The contents of the owncloud folder is visible to different users, unlike what apple does for the Pictures, Movie, Music, etc folders that are all located in the user folder.
### Steps to reproduce
1. N/A
2. N/A
3. N/A
### Client configuration

Client version: Version 2.2.4 (build 3709)

Operating system: macOS El Capitan

OS language: English

Installation path of client: /Applications

guruz:
@ckamm @ogoffart What is the behaviour on Linux? Is this with regards to umask somehow too?
On my OS X umask returns 0022

ogoffart:
On linux we just use the default which is drwxr-xr-x.  But since it is usually within the user home directory, they are not readable by other users normaly.

ckamm:
@guruz @ogoffart I don't think we explicitly change the permissions at all, do we? So the umask should be respected. There's some code in FileSystem to detect the default-write permissions that the umask implies, but this ticket is about either r permissions on files or x permissions on directories.

ckamm:
I've double checked this on linux with a 002 and 027 umask. It sets the expected file and directory permissions. @1951FDG please clarify!

1951FDG:
I will clarify tomorrow, and do the tests in a MacOS system

1951FDG:
On my OS X umask also returns 0022

~/ownCloud returns 755 which is the standard permission
~/Downloads returns 700 which is the permission that should be set also for ownCloud

`This umask will be applied to all new files and folders created in the system. While the built-in Documents, Desktop, Music, Movies, and other folders in your home directory have custom permissions that keep them private, because of the umask, if you create a new folder next to these in your home folder then its contents can be readable by everyone. Therefore, for privacy, be sure to either change the permissions of files and folders you create in your home folder, or only place them within the existing folders Apple has created there.`

1951FDG:
@guruz  Thanks for fixing this, and by the way do you happen to know which upcoming release will include this fix?

guruz:
[...]

PR #5315, opened 2016-11-15
Folder: Make folder only accessible by user #5282

guruz:
Because on OS X the parent folder might not protect
against access.

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -30,6 +30,7 @@
 #include "owncloudsetupwizard.h"
 #include "creds/abstractcredentials.h"
 #include "tooltipupdater.h"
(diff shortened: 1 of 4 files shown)


ogoffart:
I could not change that on linux.  I think the default permissions should be fine.
Maybe only on OSX ?

ckamm:
I agree with what @ogoffart said: we do respect the umask. We can add this if we need different behavior on OS X, but I wouldn't do this everywhere.

guruz:
Please re-review, I do it on OS X only now

ogoffart:
The code looks good. (assuming that's a feature we want, i can't judge because i'm not used to osx customs)

Maybe add a api documentation?

PR #5316, opened 2016-11-16
libsync: Don't store the remote URI in the csync or in the SyncEngine

ogoffart:
We are going to change the webdav path depending on the capabilities.
But the SyncEngine and csync might have been created before the capabilities
are retrieved.

The main raison why we gave the path to the sync engine was to pass it to csync.
But the thing is that csync don't need anymore this url as everything is done by the
discovery classes in libsync that use the network jobs that use the account for the urls.
So csync do not need the remote URI.

shortenFilename in folderstatusmodel.cpp was useless because the string is the
_file of a SyncFileItem which is the relative file name, that name never
starts with owncloud://.

All the csync test creates the folder because csync use to check if the folder
exists. But we don't need to do that anymore

### DIFF
--- a/csync/src/csync.c
+++ b/csync/src/csync.c
@@ -89,7 +89,7 @@ static int _data_cmp(const void *key, const void *data) {
   return 0;
 }
 
-void csync_create(CSYNC **csync, const char *local, const char *remote) {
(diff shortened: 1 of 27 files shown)


ogoffart:
cc @guruz

PR #5318, opened 2016-11-16
Fix some typing errors in russian translation.

Revertron:
The text is not correct now, proposing a correction.


ckamm:
@Revertron Thank you for your suggested changes! The translations of the owncloud client are maintained in Transifex (https://www.transifex.com/owncloud-org/owncloud/), see https://owncloud.org/contribute/translation/ for documentation on that. Would you be up for incorporating your changes there?

PR #5319, opened 2016-11-17
Prototype: Requests bundling feature implementation

mrow4a:
Bundling feature has been discussed in server PR https://github.com/owncloud/core/pull/25760. For more details, please refer there to the first and last posts in the conversation. This PR is created to integrate bundling into client. The prototype PR is thus obsolete: https://github.com/owncloud/client/pull/5155

- [x] Client feature is able to create bundled INSTRUCTION_NEW request.
- [x] Bundle could contain only file which is smaller than chunking size
- [x] Bundle may contain only files with total is smaller than chunking size
- [x] Bundle is compatibile with https://github.com/owncloud/core/pull/25760
- [x] Bundle correctly adds the files to both remote and local database - checked by manual tests
- [x] Bundle correctly adds the files to both remote and local database - checked by unit tests
- [ ] Bundle correctly adds the files to both remote and local database - checked by integration tests
- [x] Bundle passes prove of concept tests
- [x] Integrated error handling
- [x] Bundle will not be sent if there is limit bandwidth feature on.

### DIFF
--- a/src/libsync/account.cpp
+++ b/src/libsync/account.cpp
@@ -40,8 +40,6 @@ Account::Account(QObject *parent)
     : QObject(parent)
     , _capabilities(QVariantMap())
     , _davPath( Theme::instance()->webDavPath() )
-    , _wasMigrated(false)
-    , _bundleRequests(true)
 {
     qRegisterMetaType<AccountPtr>("AccountPtr");
 }
@@ -91,7 +89,7 @@ void Account::setDavUser(const QString &newDavUser)
 QString Account::davFilesPath() const
 {
     //TODO DO NOT HARCODE PATH, GET IT FROM THE SERVER!!!!
-    QString dfp("remote.php/dav/files/");
+    QString dfp("/remote.php/dav/files/");
     dfp.append(_credentials->user());
     return dfp;
 }
@@ -499,13 +497,9 @@ void Account::setNonShib(bool nonShib)
     } 
 }
 
-void Account::setBundleRequestsIfCapable(bool bundleRequests){
-    _bundleRequests = bundleRequests;
-}
-
-bool Account::bundledRequestsEnabled()
+bool Account::bundledRequestsEnabled() const
 {
-    return (_bundleRequests &&  _capabilities.bundledRequest()) ? true : false;
+    return _capabilities.bundledRequest();
 }
 
 } // namespace OCC
--- a/src/libsync/account.cpp
+++ b/src/libsync/account.cpp
@@ -40,6 +40,8 @@ Account::Account(QObject *parent)
     : QObject(parent)
     , _capabilities(QVariantMap())
     , _davPath( Theme::instance()->webDavPath() )
+    , _wasMigrated(false)
+    , _bundleRequests(true)
 {
     qRegisterMetaType<AccountPtr>("AccountPtr");
 }
@@ -84,6 +86,14 @@ QString Account::davUser() const
 void Account::setDavUser(const QString &newDavUser)
 {
     _davUser = newDavUser;
+}
+
+QString Account::davFilesPath() const
+{
+    //TODO DO NOT HARCODE PATH, GET IT FROM THE SERVER!!!!
+    QString dfp("remote.php/dav/files/");
+    dfp.append(_credentials->user());
+    return dfp;
 }
 
 QString Account::displayName() const
@@ -235,6 +245,20 @@ QNetworkReply *Account::davRequest(const QByteArray &verb, const QUrl &url, QNet
     req.setSslConfiguration(this->getOrCreateSslConfig());
 #endif
     return _am->sendCustomRequest(req, verb, data);
+}
+
+QNetworkReply *Account::multipartRequest(const QString &relPath, QNetworkRequest req, QHttpMultiPart *multiPart)
+{
+    return multipartRequest(Utility::concatUrlPath(url(), relPath), req, multiPart);
+}
+
+QNetworkReply *Account::multipartRequest(const QUrl &url, QNetworkRequest req, QHttpMultiPart *multiPart)
+{
+    req.setUrl(url);
+#if QT_VERSION > QT_VERSION_CHECK(4, 8, 4)
+    req.setSslConfiguration(this->getOrCreateSslConfig());
+#endif
+    return _am->post(req, multiPart);
 }
 
 void Account::setCertificate(const QByteArray certficate, const QString privateKey)
@@ -475,6 +499,13 @@ void Account::setNonShib(bool nonShib)
     } 
 }
 
+void Account::setBundleRequestsIfCapable(bool bundleRequests){
+    _bundleRequests = bundleRequests;
+}
 
+bool Account::bundledRequestsEnabled()
+{
+    return (_bundleRequests &&  _capabilities.bundledRequest()) ? true : false;
+}
 
 } // namespace OCC
--- a/src/libsync/account.h
+++ b/src/libsync/account.h
@@ -82,8 +82,7 @@ public:
     void setNonShib(bool nonShib);
 
     /// Functions for bundle support
-    void setBundleRequestsIfCapable(bool bundleRequests);
-    bool bundledRequestsEnabled();
+    bool bundledRequestsEnabled() const;
 
     static AccountPtr create();
     ~Account();
@@ -241,8 +240,6 @@ private:
     QByteArray _pemCertificate; 
     QString _pemPrivateKey;  
     QString _davPath; // defaults to value from theme, might be overwritten in brandings
-    bool _wasMigrated;
-    bool _bundleRequests;
     friend class AccountManager;
 };
 
--- a/src/libsync/account.h
+++ b/src/libsync/account.h
@@ -34,6 +34,7 @@ class QSettings;
 class QNetworkReply;
 class QUrl;
 class QNetworkAccessManager;
+class QHttpMultiPart;
 
 namespace OCC {
 
@@ -64,6 +65,26 @@ public:
 class OWNCLOUDSYNC_EXPORT Account : public QObject {
     Q_OBJECT
 public:
+    /**
+     * @brief The possibly themed dav path for the account. It has
+     *        a trailing slash.
+     * @returns the (themeable) dav path for the account.
+     */
+    QString davPath() const;
+
+    /**
+     * @brief The possibly themed files dav path for the account. It has
+     *        a trailing slash.
+     * @returns the (themeable) files dav path for the account.
+     */
+    QString davFilesPath() const;
+    void setDavPath(const QString&s) { _davPath = s; }
+    void setNonShib(bool nonShib);
+
+    /// Functions for bundle support
+    void setBundleRequestsIfCapable(bool bundleRequests);
+    bool bundledRequestsEnabled();
+
     static AccountPtr create();
     ~Account();
 
@@ -88,15 +109,6 @@ public:
     void setUrl(const QUrl &url);
     QUrl url() const { return _url; }
 
-    /**
-     * @brief The possibly themed dav path for the account. It has
-     *        a trailing slash.
-     * @returns the (themeable) dav path for the account.
-     */
-    QString davPath() const;
-    void setDavPath(const QString&s) { _davPath = s; }
-    void setNonShib(bool nonShib);
-
     /** Returns webdav entry URL, based on url() */
     QUrl davUrl() const;
 
@@ -113,6 +125,8 @@ public:
     QNetworkReply* deleteRequest( const QUrl &url);
     QNetworkReply* davRequest(const QByteArray &verb, const QString &relPath, QNetworkRequest req, QIODevice *data = 0);
     QNetworkReply* davRequest(const QByteArray &verb, const QUrl &url, QNetworkRequest req, QIODevice *data = 0);
+    QNetworkReply* multipartRequest(const QString &relPath, QNetworkRequest req, QHttpMultiPart *multiPart);
+    QNetworkReply* multipartRequest(const QUrl &url, QNetworkRequest req, QHttpMultiPart *multiPart);
 
 
     /** The ssl configuration during the first connection */
@@ -227,6 +241,8 @@ private:
     QByteArray _pemCertificate; 
     QString _pemPrivateKey;  
     QString _davPath; // defaults to value from theme, might be overwritten in brandings
+    bool _wasMigrated;
+    bool _bundleRequests;
     friend class AccountManager;
 };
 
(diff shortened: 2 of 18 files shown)


mrow4a:
Where can I place documentation and state machines for this?

src/libsync/account.cpp
```diff
@@ -39,6 +39,8 @@ Account::Account(QObject *parent)
     : QObject(parent)
     , _capabilities(QVariantMap())
     , _davPath( Theme::instance()->webDavPath() )
+    , _wasMigrated(false)
```
mrow4a: Not sure how this thing came here, should I delete it?

src/libsync/account.cpp
```diff
@@ -85,6 +87,14 @@ void Account::setUser(const QString &user)
     _user = user;
 }
 
+QString Account::davFilesPath() const
+{
+    //TODO DO NOT HARCODE PATH, GET IT FROM THE SERVER!!!!
+    QString dfp("/remote.php/dav/files/");
```
mrow4a: should I get it from the server?

guruz: I think not because @ogoffart is working in another PR to get this..

src/libsync/propagatebundle.cpp
```diff
@@ -0,0 +1,601 @@
...
+            // when it becomes available again.
+            if (FileSystem::isFileLocked(fullFilePath)) {
+                emit _propagator->seenLockedFile(fullFilePath);
+            }
+
+            // Soft error because this is likely caused by the user modifying his files while syncing
+            itemDone(item, SyncFileItem::SoftError, tr("ERR: Could not prepare upload device"));
+        } else {
+            // add xml part for this specific file
+            xml += "  <d:part>\n"
```
mrow4a: This should be using http://doc.qt.io/qt-5/qstring.html#append, otherwise with each "+=" it will recreate string class.

src/libsync/account.cpp
ogoffart: see https://github.com/owncloud/client/pull/5332

With that change, you don't need this change anymore and you can just use the davPath

src/libsync/account.cpp
```diff
@@ -474,6 +498,13 @@ void Account::setNonShib(bool nonShib)
     } 
 }
 
+void Account::setBundleRequestsIfCapable(bool bundleRequests){
```
ogoffart: I don't think we need this.  We should just use what the capability tells.
For debugging purposes, we might want to have an environment variable.

src/libsync/account.cpp
```diff
@@ -474,6 +498,13 @@ void Account::setNonShib(bool nonShib)
...
+    _bundleRequests = bundleRequests;
+}
 
+bool Account::bundledRequestsEnabled()
+{
+    return (_bundleRequests &&  _capabilities.bundledRequest()) ? true : false;
```
ogoffart: `(condition) ? true : false;`  is not really good style.  You can just right `return condition`

src/libsync/account.h
```diff
@@ -221,6 +235,8 @@ protected Q_SLOTS:
     QByteArray _pemCertificate; 
     QString _pemPrivateKey;  
     QString _davPath; // defaults to value from theme, might be overwritten in brandings
+    bool _wasMigrated;
```
ogoffart: merge error?

src/libsync/owncloudpropagator.cpp
```diff
@@ -115,6 +124,61 @@ static bool blacklistCheck(SyncJournalDb* journal, const SyncFileItem& item)
     return newEntry.isValid();
 }
 
+void PropagateItemJob::itemDone(SyncFileItemPtr item, SyncFileItem::Status status, const QString &errorString)
```
ogoffart: can some code be shared with PropagateItemJob::done() ? I guess you can simply call this function from done() .

src/libsync/owncloudpropagator.cpp
```diff
@@ -386,6 +456,17 @@ void OwncloudPropagator::start(const SyncFileItemVector& items)
...
+        } else if (enableBundledRequests
+                  && (item->_instruction == CSYNC_INSTRUCTION_NEW)
+                  && (item->_direction == SyncFileItem::Up)
+                  && (item->_size < chunkSize)) {
+            //this will create list of bundle files to sync for that bundlejob
+            if (directories.top().second->_bundleJob.isNull()) {
+                PropagateBundle* bundleJob = new PropagateBundle(this);
+                directories.top().second->_bundleJob.reset(bundleJob);
+            }
+            PropagateBundle* bundleJob = dynamic_cast<PropagateBundle*>(directories.top().second->_bundleJob.data());
```
ogoffart: Use qobject_cast instead of dynamic_cast.

But that actually should be static_cast, (otherwise you should check that bundleJob is not null.)

src/libsync/owncloudpropagator.cpp
```diff
@@ -599,6 +680,16 @@ bool PropagateDirectory::scheduleNextJob()
     if (_state == NotYetStarted) {
         _state = Running;
 
+        if(_bundleJob){
+            PropagateBundle* bundle = dynamic_cast<PropagateBundle*>(_bundleJob.take());
```
ogoffart: ditto  (static_cast)
Why don't you make _bundleJob  a `QScopedPointer<PropagateBundle>`  ?

src/libsync/owncloudpropagator.h
```diff
@@ -215,6 +219,8 @@ class OWNCLOUDSYNC_EXPORT PropagateDirectory : public PropagatorJob {
     virtual void abort() Q_DECL_OVERRIDE {
         if (_firstJob)
             _firstJob->abort();
+        if (_bundleJob)
+            _bundleJob->abort();
```
ogoffart: is that necessary,  shouldn't the _bundleJob be in the _subjobs if it was started?

src/libsync/propagateupload.cpp
```diff
@@ -32,6 +32,7 @@
 #include <QDir>
 #include <cmath>
 #include <cstring>
+#include <QXmlStreamReader>
```
ogoffart: mistake?

src/libsync/propagateupload.h
```diff
@@ -16,11 +16,11 @@
 #include "owncloudpropagator.h"
 #include "networkjobs.h"
 
+#include <QHttpMultiPart>
```
ogoffart: Better move it to the .cpp if it's not needed in the header.

src/libsync/syncfileitem.h
```diff
@@ -45,6 +45,7 @@ class SyncFileItem {
 
     enum Type {
       UnknownType = 0,
+      RequestsContainer = 1,
```
ogoffart: That's the same as CSYNC_FTW_TYPE_SLINK  (which is of value 1)  So the same as SoftLink.

(Also UnkownType is the same as File.  So this enum is already broken.  But one should not make it more broken.)

src/libsync/account.cpp
mrow4a: Agreed

src/libsync/account.cpp
mrow4a: Makes sense

src/libsync/owncloudpropagator.cpp
mrow4a: No, in this case we cannot do that, otherwise you will need to overwrite "_item" each time you do "done()". This need to be like that. Please also mind that _item in this case is the whole Bundle.

src/libsync/owncloudpropagator.cpp
mrow4a: Arent we using dynamic_cast for Directories? https://github.com/owncloud/client/blob/master/src/libsync/owncloudpropagator.cpp#L317 Line 317

src/libsync/owncloudpropagator.cpp
mrow4a: ditto directories

mrow4a: For security agains memory leaks. I am anyways removing this scoped pointer later and changing the parent to _subJobs

src/libsync/propagatebundle.cpp
guruz: https://doc.qt.io/qt-5/qstring.html#operator-2b-eq

src/libsync/owncloudpropagator.cpp
ogoffart: we should not use dynamic_cast for QObject. The code you point is now fixed in d22d0e7e302fc61f55b9a14551bb9a873876f934 :-)

src/libsync/syncfileitem.h
ogoffart: Make it a bigger number.

src/libsync/propagatebundle.cpp
mrow4a: Ok, so QString is like StringBuilder, nice.

mrow4a: The important thing I wanted to get here, so it will be amortized O(1), so all fine with QString.

felixboehm:
I assume you mean very technical developer documentation.
Feel free to add it to the development docs and find place for client related stuff:
https://github.com/owncloud/documentation/tree/master/developer_manual

src/libsync/owncloudpropagator.cpp
mrow4a: nice catch, did not notice that I copied the code twice.

src/libsync/propagateupload.h
mrow4a: Cannot, QHttpPart is within that, and I use it in .h file also

src/libsync/propagatebundle.cpp
```diff
@@ -0,0 +1,602 @@
...
+        QByteArray replyContent = job->reply()->readAll();
+        qDebug() << replyContent; // display the XML error in the debug
+        QString errorString = errorMessage(job->errorString(), replyContent);
+
+        if (job->reply()->hasRawHeader("OC-ErrorString")) {
+            errorString = job->reply()->rawHeader("OC-ErrorString");
+        }
+
+        //TODO: soft error till other mechanism will be added - this will retry bundle later.
+        abortWithError(SyncFileItem::SoftError, errorString);
```
mrow4a: @ogoffart @jturcotte For now I am just using SoftError in case of any full Bundle Errors. How can I implement other mechanism for whole the bundle? I mean, the problem here that currently we blacklist by filename, not by other means. How can I blacklist bundle ? I mean, this will be good mechanism on weak networks, if this will fail for some reason (airport network stops working), the client will roll back immedietaly to standard PUTs, not to try upload like 10+ files again.

src/libsync/propagatebundle.cpp
```diff
@@ -0,0 +1,602 @@
...
+
+    connect(this, SIGNAL(networkActivity()), account().data(), SIGNAL(propagatorNetworkActivity()));
+
+    //TODO this uses _device, and we have no access here to that
+//    // For Qt versions not including https://codereview.qt-project.org/110150
+//    // Also do the runtime check if compiled with an old Qt but running with fixed one.
+//    // (workaround disabled on windows and mac because the binaries we ship have patched qt)
+//#if QT_VERSION < QT_VERSION_CHECK(4, 8, 7)
+//    if (QLatin1String(qVersion()) < QLatin1String("4.8.7"))
+//        connect(_device.data(), SIGNAL(wasReset()), this, SLOT(slotSoftAbort()));
```
mrow4a: How should I come with that?

src/libsync/propagatebundle.h
```diff
@@ -0,0 +1,145 @@
...
+    void slotMultipartFinished();
+    void slotJobDestroyed(QObject *job);
+    void abortWithError(SyncFileItem::Status status, const QString &error);
+
+private:
+    quint64 checkBundledRequestsLimits()
+    {
+        //TODO: obtain this value from the server or by other means
+        quint64 maximumBundledFiles = 500;
+        return (maximumBundledFiles/_propagator->maximumActiveJob());
```
mrow4a: what do you think about that?

test/syncenginetestutils.h
```diff
@@ -649,6 +649,223 @@ class FakeErrorReply : public QNetworkReply
...
+            // ERROR CASE
+            xml.writeStartDocument();
+            xml.writeStartElement(davUri, QStringLiteral("error"));
+            xml.writeTextElement(sabUri, QStringLiteral("exception"), QStringLiteral("OCA\\DAV\\Connector\\Sabre\\Exception\\Forbidden"));
+            xml.writeTextElement(sabUri, QStringLiteral("message"), QStringLiteral("URL endpoint has to be instance of \\OCA\\DAV\\Files\\FilesHome"));
+            xml.writeTextElement(ocUri, QStringLiteral("retry"), QStringLiteral("false"));
+            xml.writeTextElement(ocUri, QStringLiteral("reason"), QStringLiteral("URL endpoint has to be instance of \\OCA\\DAV\\Files\\FilesHome"));
+            xml.writeEndElement(); // error
+            xml.writeEndDocument();
+            setAttribute(QNetworkRequest::HttpStatusCodeAttribute, 403);
```
mrow4a: I am trying to generate test, which will throw error for whole the bundle. This does not work. @jturcotte

test/testsyncengine.cpp
```diff
@@ -54,6 +64,71 @@ private slots:
...
+        fakeFolder.localModifier().insert("A/a7");
+        fakeFolder.localModifier().insert("A/a8");
+        fakeFolder.localModifier().insert("B/b4");
+
+        //add the user "erroruser" which is not a FilesHomeCollection
+        fakeFolder.getAccount()->setUrl(QUrl(QStringLiteral("http://erroruser:admin@localhost/owncloud")));
+        fakeFolder.syncOnce();
+
+        //check whole bundle
+        QVERIFY(itemDidCompleteWithStatus(completeSpy, "") == SyncFileItem::SoftError);
```
mrow4a: @jturcotte This does not work for me

mrow4a:
This PR is dependent on :
https://github.com/owncloud/client/pull/5349
https://github.com/owncloud/client/pull/5368

mrow4a:
Should I try to modify the bundling and show-case it on our local environment to try solve https://github.com/owncloud/client/issues/5391? @felixboehm @ogoffart

felixboehm:
No. First finish bundling.

guruz:
This PR would also help to decrease peak memory usage which is correlated with how many files we can sync (without switching to 64bit builds on Windows)

https://github.com/owncloud/client/issues/5451

PR #5325, opened 2016-11-19
OEM shell integration

ftiede:
OEM configurability creates binaries with OEM names which the .desktop
file doesn't take into account and thus dolphin integration does not
work if OEM changes the binary's name(s).
These changes are now integrated into the .desktop file at configure time.

This PR does not introduce changes in code functionality.


ogoffart:
Looks good. 

But there will be many icons when several plugins are installed, i think.

ftiede:
> But there will be many icons when several plugins are installed, i think.

I don't understand this specific remark as this PR is just about consistency in OEM theming and does not add any icons - esp. since an OEM themed oc client is mutually exclusive with itself, whether themed or not and thus can't be installed twice.

ckamm:
@ftiede Also looks good to me, thanks. Please change the target of the pull request to be the master branch though (might need to make a new one). And also check out the CLAassistant message above.

ftiede:
> And also check out the CLAassistant message above.

That comes from an error of mine not using the github account's name and email during my commits.
If that poses a serious problem I can recreate the entire thing, this time with the correct data.

guruz:
@ckamm He changed the branch. I guess now we can merge?

How can we get rid of my 2.2.4 commit that somehow ended up in there?

ogoffart:
The change could not be merged as it because it was made on top 2.2.4.

But i rebased the relevant commit in master. (d090a3330a0273df93bed56daa388d7a8aa67333 and before)

PR #5326, opened 2016-11-19
Use OWNCLOUD_BIN_DIR directly in C++ code.

hefee:
No need to do a STR replacement. It actually broke test execution for
Debian.

See https://bugs.debian.org/844937

### DIFF
--- a/test/owncloud_add_test.cmake
+++ b/test/owncloud_add_test.cmake
@@ -22,7 +22,7 @@ macro(owncloud_add_test test_class additional_cpp)
     )
 
     add_definitions(-DOWNCLOUD_TEST)
-    add_definitions(-DOWNCLOUD_BIN_PATH=${CMAKE_BINARY_DIR}/bin)
+    add_definitions(-DOWNCLOUD_BIN_PATH="${CMAKE_BINARY_DIR}/bin")
     add_test(NAME ${OWNCLOUD_TEST_CLASS}Test COMMAND ${OWNCLOUD_TEST_CLASS}Test)
 endmacro()
 
@@ -51,5 +51,5 @@ macro(owncloud_add_benchmark test_class additional_cpp)
     )
 
     add_definitions(-DOWNCLOUD_TEST)
-    add_definitions(-DOWNCLOUD_BIN_PATH=${CMAKE_BINARY_DIR}/bin)
+    add_definitions(-DOWNCLOUD_BIN_PATH="${CMAKE_BINARY_DIR}/bin")
 endmacro()
--- a/test/testutility.cpp
+++ b/test/testutility.cpp
@@ -11,10 +11,6 @@
 
 #include "utility.h"
 
-#define STR_(X) #X
-#define STR(X) STR_(X)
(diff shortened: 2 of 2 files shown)


test/owncloud_add_test.cmake
```diff
@@ -21,6 +21,6 @@ macro(owncloud_add_test test_class additional_cpp)
     )
 
     add_definitions(-DOWNCLOUD_TEST)
-    add_definitions(-DOWNCLOUD_BIN_PATH=${CMAKE_BINARY_DIR}/bin)
+    add_definitions(-DOWNCLOUD_BIN_PATH="${CMAKE_BINARY_DIR}/bin")
```
jturcotte: This might break on some platforms if the shell eats the quotes when invoking the compiler.

For example http://stackoverflow.com/a/38704229/56315

ogoffart:
What was broken? how does this fxes it?  What's wrong with the current macro?

ogoffart:
@hefee ?

guruz:
@hefee seems vanished..

hefee:
I added a qDebug statement to show what is going on at the function
versionOfInstalledBinary( const QString& command )

and now run the 
versionOfInstalledBinary(BIN_PATH+QLatin1String("/owncloud"));
 and the test awnsers with FAILED:
QDEBUG : TestUtility::testVersionOfInstalledBinary() "/build/owncloud-client-cTKj4V/owncloud-client-2.3.2+dfsg/obj-x86_64-1-gnu/bin/owncloud"
QDEBUG : TestUtility::testVersionOfInstalledBinary() Version of installed ownCloud Binary:  ""

if I run with my add_definitions(-DOWNCLOUD_BIN_PATH="${CMAKE_BINARY_DIR}/bin") i get the expected behaviour:
QDEBUG : TestUtility::testVersionOfInstalledBinary() "/build/owncloud-client-cTKj4V/owncloud-client-2.3.2+dfsg/obj-x86_64-linux-gnu/bin/owncloud"
QDEBUG : TestUtility::testVersionOfInstalledBinary() Version of installed ownCloud Binary:  "ownCloud version 2.3.2"

as you see BUILD_PATH is a different string than OWNCLOUD_BIN_PATH. The easiest fix I found was adding the '"' to the definition of the string. Btw. this is the common way in all other project I know that are using Qt5.

(for the completion, the test are run by:
XDG_CONFIG_HOME=config  xvfb-run bin/UtilityTest

* test/testexcludedfiles.cpp is not more using OWNCLOUD_BIN_PATH

hefee:
btw still an issue with 2.3.2

ogoffart:
Please rebase the commit,  and escape the quotes.

But I now understand why this is needed: `linux`  in the path was replaced by `1` because it is a macro.

test/owncloud_add_test.cmake
hefee: this is actually the way I use a lot while developing for KDE and it never raised a problem in Qt5 days. At Qt4 days I know we had to escape the strings.

hefee:
@ogoffart: rebased for master

ogoffart:
Meged, thank you for the patch.

PR #5332, opened 2016-11-21
libsync: use the new webdav url if the server reports it

ogoffart:
The rules to select the webdav url are now:

 - If the server reports that the new chunking algorithm is working,
   always use remote.php/dav/files/<username>
   This capability can be overriden with an environment variable

 - Otherwise, use the dav path provided by the theme, which defaults to
   remote.php/webdav

This means that with the newer server, the branding can no longer override
the webdav URL. If there is still an usecase for the branding to do so, we
need to find another way to override it. But it is now more complicated to
configure as might need include the username and need different endpoint
depending on the operations (chunks or not)

### DIFF
--- a/src/libsync/account.cpp
+++ b/src/libsync/account.cpp
@@ -57,6 +57,11 @@ Account::~Account()
 
 QString Account::davPath() const
 {
+    if (capabilities().chunkingNg()) {
+        // The chunking-ng means the server prefer to use the new webdav URL
+        return QLatin1String("/remote.php/dav/files/") + davUser() + QLatin1Char('/');
+    }
+
     // make sure to have a trailing slash
     if( !_davPath.endsWith('/') ) {
         QString dp(_davPath);
--- a/src/libsync/capabilities.cpp
+++ b/src/libsync/capabilities.cpp
@@ -110,6 +110,9 @@ QByteArray Capabilities::uploadChecksumType() const
 
 bool Capabilities::chunkingNg() const
 {
+    static const auto chunkng = qgetenv("OWNCLOUD_CHUNKING_NG");
+    if (chunkng == "0") return false;
+    if (chunkng == "1") return true;
     return _capabilities["dav"].toMap()["chunking"].toByteArray() >= "1.0";
 }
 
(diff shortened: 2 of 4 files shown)


src/libsync/account.cpp
```diff
@@ -57,6 +57,11 @@ Account::~Account()
 
 QString Account::davPath() const
 {
+    if (capabilities().chunkingNg()) {
```
mrow4a: Why do you need this if here? I might want to use it for bundling also.

mrow4a: But been mostly thinking about creating just separate function instead of "if"

guruz:
👎  for now. Can you mention the issue ID in the commit? 
https://github.com/owncloud/client/issues/4007

@michaelstingl @rperezb Please see the comment in the commit message
https://github.com/owncloud/client/pull/5332/commits/c659a9fd39c7752d90dc8746ec6dfdd866ac1ccb
Is there still a use case to change the DAV url in branding?

👍 if this is clarified

rperezb:
@guruz in ownBrander we do have a variable for the users to update the webdav end point:
- webDavPath()
 
If i got this right, within the version 2.3.0, this is not going to be possible any more, 
@michaelstingl do we have any customers with the desktop_webdav_url_text modified? If so, we need to handle this before updating ownBrander

mrow4a:
When merge is planned? Is it possible to create new function filesDavPath()?

ogoffart:
@rperezb , @michaelstingl :  With the client 2.3.0 , the webDavPath ownbrander setting will only have any impact if the server is < 9.2

From owncloud 9.2, we will use the new dav system with a new URL scheme.  If we want to keep an option, we need to make it smarter so we can configure it for all the possible end point.

Depending on the use case of this setting, i suggest one of the following:
 - We realize that the reason that forced us to create this settings are no longer applicable with the new scheme, and we drop the option
 - If a custom webdav url is configured, we do not enable the new url scheme (This also means the chunking NG and bundling will not be enabled)
 - We find another way to configure the webdav path

rperezb:
about how to proceed, let's see what @hodyroff or @felixboehm thinks, at the end, this was a requirement from customers.

to sum up, we do have an option on the desktop client that allows clients to modify the webdav end point, instead of using the one by default (remote.php/webdav) is this needed to any customer? / do we want to keep this?
for context, this is on the backlog for mobile apps, still not implented nor planned

ogoffart:
The option will still be working with server older than 9.2

felixboehm:
I can't tell if we need to be able to configure webdav routes.
But we should document this, right? @owncloud/doc

PR #5337, opened 2016-11-25
owncloudcmd: Don't put the password in the URL

ogoffart:
As the URL might be print on the logs.
Also don't change the scheme from http to owncloud.

This was required before when we were using neon through csync, but now that
we use QNAM for everything we don't need it. The credentials from the account
are used.

### DIFF
--- a/src/cmd/cmd.cpp
+++ b/src/cmd/cmd.cpp
@@ -329,8 +329,7 @@ int main(int argc, char **argv) {
     if( !options.target_url.contains( account->davPath() )) {
         options.target_url.append(account->davPath());
     }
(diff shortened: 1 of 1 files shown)


felixboehm:
👍  
Should we test this in detail? Or is behavior not changed with that ...

dragotin:
I tested with my initial testcase and it is good. Thanks @ogoffart 
👍

ogoffart:
> There's a replace owncloud->http further down that can probably also be removed (unless we expect users to pass in owncloud:// urls).

I kept it exactly for that reason, some people might use owncloud:// URL for legacy reason and i did not want to break it.

Issue #5331, opened 2016-11-21
[Placeholder] Implement branding option handling for M permission


guruz:
If this branding option is activated, folders with 'M' should not be auto-selected in the selective sync.
The user would then need to manually click the checkbox to sync the folder.
(No prompt or error message, just not select by default)

guruz:
PR is ready, closing here
https://github.com/owncloud/client/pull/5340

PR #5340, opened 2016-11-29
Discovery: Add branding option to disable default sync of 'M' directo…

ogoffart:
…ries

Issue #5331 and https://github.com/owncloud/enterprise/issues/1594

### DIFF
--- a/src/gui/wizard/owncloudadvancedsetuppage.cpp
+++ b/src/gui/wizard/owncloudadvancedsetuppage.cpp
@@ -215,6 +215,11 @@ QStringList OwncloudAdvancedSetupPage::selectiveSyncBlacklist() const
     return _selectiveSyncBlacklist;
 }
 
+bool OwncloudAdvancedSetupPage::isConfirmBigFolderChecked() const
+{
+    return _ui.rSyncEverything->isChecked() && _ui.confCheckBoxSize->isChecked();
+}
+
 bool OwncloudAdvancedSetupPage::validatePage()
 {
     if(!_created) {
--- a/src/gui/wizard/owncloudadvancedsetuppage.cpp
+++ b/src/gui/wizard/owncloudadvancedsetuppage.cpp
@@ -68,6 +68,15 @@ OwncloudAdvancedSetupPage::OwncloudAdvancedSetupPage()
     _ui.lServerIcon->setPixmap(appIcon.pixmap(48));
     _ui.lLocalIcon->setText(QString());
     _ui.lLocalIcon->setPixmap(QPixmap(Theme::hidpiFileName(":/client/resources/folder-sync.png")));
+
+    if (theme->wizardHideExternalStorageConfirmationCheckbox()) {
+        _ui.confCheckBoxExternal->hide();
+    }
+    if (theme->wizardHideFolderSizeLimitCheckbox()) {
+        _ui.confCheckBoxSize->hide();
+        _ui.confSpinBox->hide();
+        _ui.confTraillingSizeLabel->hide();
+    }
 }
 
 void OwncloudAdvancedSetupPage::setupCustomization()
@@ -118,6 +127,12 @@ void OwncloudAdvancedSetupPage::initializePage()
         _selectiveSyncBlacklist = QStringList("/");
         QTimer::singleShot(0, this, SLOT(slotSelectiveSyncClicked()));
     }
+
+    ConfigFile cfgFile;
+    auto newFolderLimit = cfgFile.newBigFolderSizeLimit();
+    _ui.confCheckBoxSize->setChecked(newFolderLimit.first);
+    _ui.confSpinBox->setValue(newFolderLimit.second);
+    _ui.confCheckBoxExternal->setChecked(cfgFile.confirmExternalStorage());
 }
 
 // Called if the user changes the user- or url field. Adjust the texts and
@@ -207,6 +222,13 @@ bool OwncloudAdvancedSetupPage::validatePage()
         _checking = true;
         startSpinner();
         emit completeChanged();
+
+        if (_ui.rSyncEverything->isChecked()) {
+            ConfigFile cfgFile;
+            cfgFile.setNewBigFolderSizeLimit(_ui.confCheckBoxSize->isChecked(),
+                                                _ui.confSpinBox->value());
+            cfgFile.setConfirmExternalStorage(_ui.confCheckBoxExternal->isChecked());
+        }
 
         emit createLocalAndRemoteFolders(localFolder(), _remoteFolder);
         return false;
--- a/src/libsync/discoveryphase.h
+++ b/src/libsync/discoveryphase.h
@@ -229,7 +229,7 @@ signals:
     void doGetSizeSignal(const QString &path, qint64 *result);
 
     // A new folder was discovered and was not synced because of the confirmation feature
-    void newBigFolder(const QString &folder);
+    void newBigFolder(const QString &folder, bool isExternal);
 };
 
 }
--- a/src/libsync/discoveryphase.h
+++ b/src/libsync/discoveryphase.h
@@ -116,6 +116,8 @@ private:
     bool _ignoredFirst;
     // Set to true if this is the root path and we need to check the data-fingerprint
     bool _isRootPath;
+    // If this directory is an external storage (The first item has 'M' in its permission)
+    bool _isExternalStorage = false;
     QPointer<LsColJob> _lsColJob;
 
 public:
--- a/src/libsync/discoveryphase.h
+++ b/src/libsync/discoveryphase.h
@@ -176,8 +176,8 @@ class DiscoveryJob : public QObject {
      */
     bool isInSelectiveSyncBlackList(const char* path) const;
     static int isInSelectiveSyncBlackListCallback(void *, const char *);
-    bool checkSelectiveSyncNewFolder(const QString &path);
-    static int checkSelectiveSyncNewFolderCallback(void*, const char*);
+    bool checkSelectiveSyncNewFolder(const QString &path, const char *remotePerm);
+    static int checkSelectiveSyncNewFolderCallback(void* data, const char* path, const char* remotePerm);
 
     // Just for progress
     static void update_job_update_callback (bool local,
--- a/src/libsync/discoveryphase.h
+++ b/src/libsync/discoveryphase.h
@@ -34,6 +34,15 @@ class Account;
  * if the files are new, or changed.
  */
 
+struct SyncOptions {
+    /** Maximum size (in Bytes) a folder can have without asking for confirmation.
+     * -1 means infinite */
+    qint64 _newBigFolderSizeLimit = -1;
+    /** If a confirmation should be asked for external storages */
+    bool _confirmExternalStorage = false;
+};
+
+
 /**
  * @brief The FileStatPointer class
  * @ingroup libsync
@@ -197,7 +206,7 @@ class DiscoveryJob : public QObject {
 
 public:
     explicit DiscoveryJob(CSYNC *ctx, QObject* parent = 0)
-            : QObject(parent), _csync_ctx(ctx), _newBigFolderSizeLimit(-1) {
+            : QObject(parent), _csync_ctx(ctx) {
         // We need to forward the log property as csync uses thread local
         // and updates run in another thread
         _log_callback = csync_get_log_callback();
@@ -207,7 +216,7 @@ public:
 
     QStringList _selectiveSyncBlackList;
     QStringList _selectiveSyncWhiteList;
-    qint64 _newBigFolderSizeLimit;
+    SyncOptions _syncOptions;
     Q_INVOKABLE void start();
 signals:
     void finished(int result);
--- a/src/libsync/theme.cpp
+++ b/src/libsync/theme.cpp
@@ -275,6 +275,15 @@ qint64 Theme::newBigFolderSizeLimit() const
     return 500;
 }
 
+bool Theme::wizardHideExternalStorageConfirmationCheckbox() const
+{
+    return false;
+}
+
+bool Theme::wizardHideFolderSizeLimitCheckbox() const
+{
+    return false;
+}
 
 QString Theme::gitSHA1() const
 {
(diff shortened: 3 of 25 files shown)


ogoffart:
(untested)

rperezb:
@SamuAlfageme calling you! 📲

guruz:
👎  doesn't work, debugging

guruz:
@SamuAlfageme You can test this by having an external storage configured in your oC server (files_external). I tried with a system wide mount but you could try with a user mount.
It should then show up unsynced by default in selective sync

SamuAlfageme:
@ogoffart @guruz let me get this straight: `bool dontSyncMountedStorageByDefault() const { return true; }` should **not mount** external storage by default? Cause if that's the case, trying with 3 different external storage opts. configured (Dropbox, Drive & Local), all of them are selected in the selective sync view by default.

SamuAlfageme:
So far what I found:

* Both account setup wizard options (`sync everything` and `choose what to sync`) are still selecting by default all the folders, including the ones from external storage. 
* When deleting `.csync_journal.db`, 3 actions are displayed and the folders unchecked, however none of those are erasing the contents from the local sync folder (plus the message shown is addressing size problems, instead of external-storage reasons):

data_too_big

michaelstingl:
@SamuAlfageme What is the exact test procedure?

SamuAlfageme:
@michaelstingl you need to modify the theme to set `dontSyncMountedStorageByDefault() const { return true; }` and build with the `issue_5331` branch. 

From there, I can provide you with a test server with 2/3 different external storage options configured and running. And you can see how the auto-deselect doesn't work neither on the create account wizard nor when deleting (to force restore) the `.csync_journal.db` on the root folder sync connection. 

Second scenario unchecks the external storage folders but does not remove them from the local filesystem when selecting any of those 3 options.

SamuAlfageme:
Cherrypicking the changes in #5426 for testing purposes...

ckamm:
@SamuAlfageme @ogoffart I've addressed the inconsistency between setup and later synchronization in https://github.com/ckamm/owncloud-client/commit/39d048a8117c2508c8b9ca66300b3912c36910ca - shall I include that patch here?

SamuAlfageme:
@ckamm sure! will re-test as soon as it's ready.

ckamm:
@SamuAlfageme I've put it into this PR for now, but it's pending @ogoffart's opinion.

ogoffart:
I added some fixup commit.

The remaining issues i could think of with this patch: 

 - The rewording, as i wrote on IRC: the reason it was phrased like (with double negative) that is because it is a black list. So it is "unselected item will not be synced" rather than "selected items will be synced". The difference is weak,  but it is to make sure people don't complain that they select  "foo/bar",  but not "foo",  and yet, the file within foo are sync'ed.
I don't have a strong opinion. And if you think the new wording is easier to understand,  let's just change it.

- The 'M' detection will only work for top level directories.  We don't check for sub directories.   I guess that's fine because mount point are always top level, are they not?

SamuAlfageme:
New tests after these changes are looking good so far: 🎉

screenshot 2017-01-11 18 36 23

There's also this one conceptual thing we need to address: https://github.com/owncloud/client/issues/5421#issuecomment-269325277 because `Sync everything` + do not sync ext.sto. will result in syncing ext.sto.

SamuAlfageme:
After including the [message pointing the cause of folders being unchecked](https://github.com/owncloud/client/pull/5340/commits/99cfd4b14f4ccff7300ae73cf6fcf5249e2bf1b3#diff-9ca7819238177b4961cd262343ab72fdR83) there should be the same for newly created storages. cc/ @ckamm 

new_ext_sto

Answering to @ogoffart's https://github.com/owncloud/client/pull/5340#issuecomment-271915388

> The 'M' detection will only work for top level directories. We don't check for sub directories. I guess that's fine because **mount point are always top level, are they not?**

Well, looks like yes, if you specify a path for the mounting point instead of a single, folder name; it's mounted that path (if available; I'm investigating a bug related with this). I don't know if this is the expected behavior though. cc/ @PVince81 

screenshot 2017-01-12 10 06 13

Also, again, we have discrepancies between what happens on the wizards and if changes were made in an already created folder sync connection. In the second scenario, folder is unchecked: 

subfolder_mounting_point

In the first, as addressed in https://github.com/owncloud/client/pull/5340/commits/7344571ae8e4010b9c9e813aaf61e181bb43c4d6 it isn't. So we should have an uniform way of dealing with this.

---

@ogoffart agreed on having a different name for the option as current (double negation antipattern) is too confusing (and was source of my question on https://github.com/owncloud/client/pull/5340#issuecomment-266390102)

ogoffart:
@michaelstingl  Should the branding option be only for new folder,  or also for new sync?

i.e, in the first sync, if the user selected "Sync Everything" in the wizard, should the 'M' folder be sync'ed or not?

ckamm:
@ogoffart How about we merge this to have it in 2.3 and create a new ticket to discuss the implications on the wizards? If merging this soon isn't urgent, we can also keep it like this.

pmaier1:
1. Setup local folder options
- [ ] Add checkbox 'Ask **for** confirmation before **synchronizing** folders larger than X MB'
- [ ] Add checkbox 'Ask **for** confirmation before **synchronizing** external storages'
-> Gives easy ability to configure the two options before initial sync; should apply to initial sync
2. Client default view / new external storage received
- [ ] Add folder icons to indicate external storages
- [ ] Add text below 'There are folders that were not synchronized because they are too big:' (removed 'new'!) => 'There are folders that were not synchronized because they are external storages:'
3. Client 'General' settings
- [ ] Add checkbox 'Ask **for** confirmation before **synchronizing** external storages'
- [ ] Adapt >500MB text accordingly
4. Design for subfolders

When i have a folder foo/bar/baz :  
- can bar be 'M' and not foo? => Yes
- does baz have 'M' too or not? => if 'bar' has 'M', 'baz' should have 'M' as well (as handled server-side!?)
- if foo is checked, should 'bar' be downloaded? => depends on setting in 3. ('bar' should behave analogous to a folder >500MB)
- if 'bar' is checked, should 'baz' be downloaded? => since 'bar' is 'M', 'baz' automatically is 'M' as well and should be synced without further confirmation
5. ownBrander
- [ ] Area 'Optional' => 'Define folder sync limit': Hide checkbox 'Ask for confirmation... folders larger than..' in account setup dialog
- [ ] Area 'Optional' => Top level: Hide checkbox 'Ask for confirmation... external storages' in account setup dialog

SamuAlfageme:
@pmaier1 your comment also works for https://github.com/owncloud/client/pull/5426 regarding all the threshold limit options. 

As for: 

> Setup local folder options
>  - [ ]  Add checkbox 'Ask for confirmation before ...

In https://github.com/owncloud/client/pull/5426#issuecomment-272412550 I suggested using buttons instead of checkboxes to select/deselect all the folders that met this condition:

choose_what_to_sync_button

**The only issue I see is the current behavior of the wizard, since `Sync everything` radio button is selected by default** unless the `Theme::wizardSelectiveSyncDefaultNothing()` (https://github.com/owncloud/client/issues/5421#issuecomment-269325277) is used; in which case, no folder is checked. In which step does it make sense to add these checkboxes/options?

ogoffart:
> 1. Setup local folder options

So Like this?

Should these checkboxes always appear? or should it be dependent of a branding option?

spectacle a31134

> 2. Client default view / new external storage received
> -￼ Add folder icons to indicate external storages

A new icon? What should it look like?
(Note: the current folder icon is provided by the system as the native folder icon on whatever running platform.)

> 4. Design for subfolders

Alright. (It just makes the implementation extra complicated because it cannot be plugged on the folder limit anymore)

> 5. ownBrander

What should the default options be? (What would it be for the official owncloud client?)

ogoffart:
Another question: Are the new checkboxes int he new account wizard dependent of the options from the general tab? (I guess they should).  And should modifying the value also change the general option (for all account)?

In my mockup I made them radio button, but I realize that then it would not be possible to select both at the same time.
But if it is checkboxes it becomes wierd too as you cannot say "syncronize everything" but then not synchronized mounted folder.  Or how should it look?

pmaier1:
wizard-confirmation

phil-davis:
Something like the above post, but maybe the 2 "Ask confirmation" checkboxes can also be applied to "Choose what to sync" if the user selects that - i.e. the folder tree of the "Choose what to sync" dialog would display with large folders and/or external storages unchecked if the user selected like that. Then the user can go ahead and de-select any other stuff they do not want, or re-select any things they do want.

i.e. the "Ask confirmation" checkboxes are orthogonal to the radio button set "Sync everything"/"Choose what to sync".

phil-davis:
The "folders larger than nnn Mb" could display the current general setting, and be clickable to modify. Then if clicked, popup a dialog to change the value, that can have more text that will make it clear that changing the value will change the general setting. (It needs to be able to change the general setting, because on the first account setup the user would have no other way to get to the General tab)

pmaier1:
Added summary to my last comment. Not sure if I covered everything, ping me if there are questions.

SamuAlfageme:
@pmaier1 Think this works to sum up:

wizard

The checkboxes can be used to modify the contents of the "Choose What to Sync" file picker. 

Also, if `Theme::wizardSelectiveSyncDefaultNothing()` is set, both checkboxes would be unselected.

ogoffart:
I pushed my work so far which is ready for further testing.

SamuAlfageme:
After some smoke tests with the new changes, everything seems to be behaving in the way we discussed in previous comments. Here are some insights after the tests:

### Step 3. on the setup wizard:

screenshot 2017-01-24 18 06 59

* The checkboxes do not affect the contents of the `Choose what to sync...` filepicker 
* The value used in this wizard is then not set as the global threshold (i.e. not applied in the General settings view on the client) but it's kept for future "new account" wizards. 
* The checkboxes are not aligned H/V:

screenshot 2017-01-24 18 17 37

* Notifications:
    - [ ] Fix the content of the notifications about external storage:
    - [ ] Group them in one instead of displaying them in batch after clicking in `Connect...`

<p align="center">
  
</p>

### Account tab 

icons

- [ ] Notice some of the folder names are cropped after the line break.

ogoffart:
@SamuAlfageme  Thanks for testing!  I have fixed most issues. I icreased the default size of the wizard so everything fits.

> The checkboxes do not affect the contents of the Choose what to sync... filepicker

That is expected i'd say. (See comment https://github.com/owncloud/client/pull/5340#issuecomment-274756529 )

> Group [notifications] in one instead of displaying them in batch after clicking in Connect...

The API for notifications is "fire and forget".  So we'd need to accumulate them and only show the notification when discovery is finished. We can do that. But that's a bigger change which i might do later.

>  some of the folder names are cropped after the line break.

I could not reproduce that. Maybe it's only on Mac?

SamuAlfageme:
@ogoffart awesome, updated https://github.com/owncloud/client/pull/5340#issuecomment-274878023 and will move some of the things that can be improved to individual issues to address them later so we can merge here. Couple of simple things before so,

- [ ] The wizard window now looks huge: 

screenshot 2017-01-26 11 36 29

- [ ] The new icons introduced in ed15842eb992bd1a77c4f5d60143cc1d90f6097d are not being used in the "Choose what to sync" view: 

screenshot 2017-01-26 10 49 23

SamuAlfageme:
@ogoffart and finally, extracted from https://github.com/owncloud/client/pull/5340#issuecomment-274756529:

> [If checkboxes apply to 'Choose what to sync' as well they should probably be placed below, not in between the both options]

- [ ] Place the the confirmation options below the Sync everything/Choose what to sync radio buttons

ogoffart:
@SamuAlfageme :  Do you think the wizard is now too big?


Regarding the choose what to sync, i'll change their icon in a follow up pull request.
But the checkboxes do not apply to the choose what to sync.

src/libsync/discoveryphase.h
```diff
@@ -34,6 +34,15 @@ class Account;
  * if the files are new, or changed.
  */
 
+struct SyncOptions {
+    /** Maximum size (in Bytes) a folder can have without asking for confirmation.
+     * -1 means infinite */
+    qint64 _newBigFolderSizeLimit = -1;
```
guruz: Just saying in future it would be nice if all variables with time or bytes are called something like ...msec or ..bytes etc. Not important now

src/gui/wizard/owncloudadvancedsetuppage.cpp
```diff
@@ -200,6 +215,11 @@ QStringList OwncloudAdvancedSetupPage::selectiveSyncBlacklist() const
     return _selectiveSyncBlacklist;
 }
 
+bool OwncloudAdvancedSetupPage::confirmBigFolder() const
```
guruz: Would be good to have such functions named "is..." or "are..." to see it is a getter

src/libsync/theme.cpp
```diff
@@ -275,6 +275,15 @@ qint64 Theme::newBigFolderSizeLimit() const
     return 500;
 }
 
+bool Theme::wizardHideExternalStorageConfirmationCheckbox() const
```
guruz: FYI @rperezb

SamuAlfageme:
@ogoffart 

> Do you think the wizard is now too big? 

Kind of, yeah :confused:. Took the whole screen on my Win7 VM, it's bigger than the application main window and there's a lot of unused space on the right. 

> Regarding the choose what to sync, i'll change their icon in a follow up pull request. 

Great! 😉 

> But the checkboxes do not apply to the choose what to sync.

They do if they're set as global settings, so I would rather enable them for both options and place them under the radio buttons.

guruz:
Code looks OK 👍 
I have not tested the actual functionality.

wizard is also too big on my screen, please reduce and merge.

@SamuAlfageme Did you also test this with deeper directory structures? Will creating a new directory on the server side deep inside the mounted storage again ask for confirmation?

SamuAlfageme:
@guruz 
> Did you also test this with deeper directory structures?

Yeah, looks good so far: 

screenshot 2017-01-26 17 56 33

> Will creating a new directory on the server side deep inside the mounted storage again ask for confirmation? 

Tried making **superficial changes** on the external storage and it was ok, program logic could point otherwise?

src/libsync/theme.cpp
rperezb: @guruz thanks!

ogoffart:
I resized the wizard again (750 pixels wide)

guruz:
@ogoffart I accidently merged this to master instead of 2.3.
Do you want to merge to 2.3 manually with git?

PR #5343, opened 2016-11-29
Attempt to fix main crashes reported by the crash reporter

ogoffart:


src/libsync/account.cpp
```diff
@@ -374,8 +374,11 @@ void Account::slotHandleSslErrors(QNetworkReply *reply , QList<QSslError> errors
     // Keep a ref here on our stackframe to make sure that it doesn't get deleted before
     // handleErrors returns.
     QSharedPointer<QNetworkAccessManager> qnamLock = _am;
+    QPointer<QObject> guard = reply;
 
     if (_sslErrorHandler->handleErrors(errors, reply->sslConfiguration(), &approvedCerts, sharedFromThis())) {
+        if (!guard) return;
```
ogoffart: how can `reply->ignoreSslErrors()`  works if reply was deleted?

src/gui/folderstatusmodel.cpp
```diff
@@ -698,17 +696,20 @@ void FolderStatusModel::slotLscolFinishedWithError(QNetworkReply* r)
...
         parentInfo->_lastErrorString = r->errorString();
 
+        parentInfo->resetSubs(this, idx);
+
         if (r->error() == QNetworkReply::ContentNotFoundError) {
             parentInfo->_fetched = true;
         } else {
-            if (!parentInfo->hasLabel()) {
+            bool add = !parentInfo->hasLabel();
+            if (add) {
```
ogoffart: The problem was the lack of remove i think  (altough i tried to reproduce the crash localy but i could not)

So the problem would appear in the case there was a fetchLabel,  but that we get then a 404 so we would remove the label, but not call removeRows.  In this case, it is possible that the treeview still call us with an index that is child of this element, with row 0 (which was corresponding before to the label)  But now that the label is gone, the row 0 correspond to the first element of the _subs vector.  And since the vector is empty, this leads to crashes.


But you areright that we don't need to test for hasLabel

ogoffart:
commit updated.

src/libsync/account.cpp
ckamm: @ogoffart I meant that we should move the check to before `reply->ignoreSslErrors()`. That way we can store the approved cert and save the account. The user dialog probably went through just fine. And if we don't do it, the user will be asked about the same certificate again.

ckamm:
@ogoffart  This still has a pending question at https://github.com/owncloud/client/pull/5343#pullrequestreview-10702967

Issue #4875, opened 2016-05-20
Dynamic chunk sizing with new bigfile chunking


dragotin:
Do dynamic chunk sizes with the new bigfile chunking related #4019

cmonteroluque:
Network Traffic Improvements (for 2.3, would spend a day anazlying this to see if there is something to do)
- Ongoing excercise to make it better and faster on the network
- This includes dynamic chunk sizing and HTTP2 which was already discussed above
- Wanted to do a wireshark trace to analyze http, tcp traffic flow to see if we need to patch Qt
- It seems good, but not critical
- Then we need to provide a set of upstream patches to Qt

guruz:
Let's move to 2.4..

ckamm:
Dynamic chunking is in and happens in practice.

PR #5368, opened 2016-12-08
Prototype: dynamic chunkingNG

mrow4a:
In current implementation, big files e.g. 100MB are being chunked into smaller pieces called 10MB chunks. The problem here is that this is fixed value, and it is not appropriate on all types of the networks (WiFi versus FiberToTheHome versus ETHERNET LAN). For WiFi it makes sense for small chunks, while for fast networks, it makes sense to have very big chunks - optimal maximum of 50MB.

**Server capability**

This PR proposes implementation, which in first phase downloads from the server the following capabilities:
````
	/*
	 * This function will return:
	 *
	 * - <chunking>: version number of chunking on the client
	 *
	 * - <max_single_upload_request_duration_msec>: Dynamic Chunking attribute the maximum number of miliseconds that single request below chunk size can take
	 * 		This value should be based on heuristics with default value 10000ms, time it takes to transfer 10MB chunk on 1MB/s upload link.
	 *
	 * 		Suggested solution will be to evaluate max(SNR, MORD) where:
	 * 	    > SNR - Slow network request, so time it will take to transmit default chunking sized request on the current client version to sync at specific low upload bandwidth
	 *      > MORD - Maximum observed request time, so double the time of maximum observed RTT of the very small PUT request (e.g. 1kB) to the system
	 *
	 * 		Exemplary, syncing 100MB files, with chunking size 10MB, will cause sync of 10 PUT requests which max evaluation was set to <max_single_upload_request_duration_msec>
	 *
	 * 		Dynamic chunking client algorithm is specified in the ownCloud documentation and uses <max_single_upload_request_duration_msec> to estimate if given
	 * 		bandwidth allows higher chunk sizes (because of high goodput)
	 */
	public function getCapabilities() {
		return [
			'dav' => [
				'chunking' => '1.0',
				'max_single_upload_request_duration_msec' => '10000',
			]
		];
	}
````

The above parameter will be used as a reference in client learning of the available host/server bandwidth and capabilities of the server.

**How dynamic chunking works**

Lets assume, that the client wants to synchronise 100MB file. Lets assume, that request RTT is 100ms. This means, that it takes around 100ms to send 1st chunk with 1kB of the data. Recalling, the 1st to nth chunked PUTs will not cause any delay in database (as single file PUTs are doing). Final MOVE will cause file to be added to the ownCloud cache.

 In the current implementation, the following will happen, taking around 3,6 second:
selection_086

File will be equally divided into 10 chunks of 10MB, send to assemply stream and MOVEd (with bookkeeping operation) to ownCloud.

In the PRed implementation, the following will happen, taking around 2,5second (30% faster):
selection_087

What happend, is that 1 PUT carries 10MB, 2nd PUT carries 20MB, 3rd PUT carries 30MB, and 4th PUT carries 40MB. 

**Algorithm description**

Algorithm is based on changes in request time to the reference value, and being corrected using ``ln(change)-1``. 
This results in the following - please mind that this does not includes TCP Congestion Control and in reality growth will be lower:
selection_091

Three colors are showing the 4 consequitive PUTs for static bandwidth of 10MB/s, chunk sizes 10MB->20MB->25MB->30MB, with slowing growth or reset to 10MB again.
selection_091

The above graph presents what will be the next values of chunking sizes for the specific static congestion window and specific values of previous values of chunking sizes. 

Higher the bandwidth, faster will approach maximum value of 50MB. For lower bandwidths it will not rise at all or rise very slowly.

Higher the chunking size, higher bandwidth needed to cause growth. 

You also probably ask yourselves, ok, but this will not work for higher latencies (RTTs). 
selection_092

Yes it works, but until the specific limit of 1s latency. With increasing latency above 1s it will decrease exponentialy. However, this is never the case, since such latencies are rare and are signaling an error. This kind or RTTs are also possible for database bookkeeping operation, however, this is not a case here, since MOVE is a separate operator here, and this one depends on the server. 

**Why Natural Logarithm?**

Natural logarithm has a property that for given given x, when x<1, logarithm is negative, x=1, logarithm is 0, and for x>1, logarithm is possitive. 

If your x is ratio of ``CurrentRequestDuration / ReferenceRequestDuration``, when CurrentRequestDuration is equal to ReferenceRequestDuration, the next chun size value ''_lastChunkSize + log(CurrentRequestDuration / ReferenceRequestDuration)*chunkSize()" is equal to "_lastChunkSize". 

As an example, given network 1MBs, thus 10MB default chunk will take 10s, and having reference value equal to 10s, our change will be exactly 0. 

Next property of logarithm is that, if you the result of ''log()'' you substract 1, you shift the value when logarithm becomes possitive e.g. given network 10MBs, thus 10MB default chunk will take 1s, and having reference value equal to 10s, our correction parameter ''log()-1'' change will be equal to 1, so your next chunk size will be 20MBs. The zero border will be shifted to 3MBs in order to start increasing the chunk size. This can be shifted by substracting log value more. 

RFC - Request for Comments 
@DeepDiver1975 @ogoffart @guruz @jturcotte @hodyroff @felixboehm @butonic 

Excel with calculations:
[dynamic-chunking.ods.zip](https://github.com/owncloud/client/files/638078/dynamic-chunking.ods.zip)

### DIFF
--- a/src/libsync/propagateupload.h
+++ b/src/libsync/propagateupload.h
@@ -304,6 +304,24 @@ private:
     uint _transferId; /// transfer id (part of the url)
     int _currentChunk; /// Id of the next chunk that will be sent
     bool _removeJobError; /// If not null, there was an error removing the job
+    quint64 _lastChunkSize; /// current chunk size
+
+    /*
+    * This is value in ms obtained from the server.
+    *
+    * Dynamic Chunking attribute the maximum number of miliseconds that single request below chunk size can take
+    * This value should be based on heuristics with default value 10000ms, time it takes to transfer 10MB chunk on 1MB/s upload link.
+    *
+    * Suggested solution will be to evaluate max(SNR, MORD) where:
+    * > SNR - Slow network request, so time it will take to transmit default chunking sized request at specific low upload bandwidth
+    * > MORD - Maximum observed request time, so double the time of maximum observed RTT of the very small PUT request (e.g. 1kB) to the system
+    *
+    * Exemplary, syncing 100MB files, with chunking size 10MB, will cause sync of 10 PUT requests which max evaluation was set to <max_single_upload_request_duration_msec>
+    *
+    * Dynamic chunking client algorithm is specified in the ownCloud documentation and uses <max_single_upload_request_duration_msec> to estimate if given
+    * bandwidth allows higher chunk sizes (because of high goodput)
+    */
+    quint64 _requestMaxDuration;
 
     // Map chunk number with its size  from the PROPFIND on resume.
     // (Only used from slotPropfindIterate/slotPropfindFinished because the LsColJob use signals to report data.)
@@ -311,6 +329,12 @@ private:
     QMap<int, ServerChunkInfo> _serverChunks;
 
     quint64 chunkSize() const { return propagator()->chunkSize(); }
+    quint64 maxChunkSize() const { return propagator()->maxChunkSize(); }
+
+    quint64 getRequestMaxDurationDC(){
+        return _requestMaxDuration;
+    }
+
     /**
      * Return the URL of a chunk.
      * If chunk == -1, returns the URL of the parent folder containing the chunks
@@ -318,10 +342,11 @@ private:
     QUrl chunkUrl(int chunk = -1);
 
 public:
-    PropagateUploadFileNG(OwncloudPropagator* propagator,const SyncFileItemPtr& item) :
-        PropagateUploadFileCommon(propagator,item) {}
+    PropagateUploadFileNG(OwncloudPropagator* propagator,const SyncFileItemPtr& item, const quint64& requestMaxDuration) :
+        PropagateUploadFileCommon(propagator,item), _lastChunkSize(0), _requestMaxDuration(requestMaxDuration) {}
 
     void doStartUpload() Q_DECL_OVERRIDE;
+
 private:
     void startNewUpload();
     void startNextChunk();
--- a/src/libsync/propagateupload.h
+++ b/src/libsync/propagateupload.h
@@ -19,6 +19,7 @@
 #include <QBuffer>
 #include <QFile>
 #include <QDebug>
+#include <QElapsedTimer>
 
 
 namespace OCC {
@@ -90,6 +91,7 @@ private:
     QMap<QByteArray, QByteArray> _headers;
     QString _errorString;
     QUrl _url;
+    QElapsedTimer _requestTimer;
 
 public:
     // Takes ownership of the device
@@ -122,6 +124,10 @@ public:
     }
 
     virtual void slotTimeout() Q_DECL_OVERRIDE;
+
+    quint64 msSinceStart() const {
+        return _requestTimer.elapsed();
+    }
 
 
 signals:
@@ -201,7 +207,6 @@ protected:
     QByteArray _transmissionChecksum;
     QByteArray _transmissionChecksumType;
 
-
 public:
     PropagateUploadFileCommon(OwncloudPropagator* propagator,const SyncFileItemPtr& item)
         : PropagateItemJob(propagator, item), _finished(false), _deleteExisting(false) {}
@@ -276,7 +281,7 @@ private:
     int _chunkCount; /// Total number of chunks for this file
     int _transferId; /// transfer id (part of the url)
 
-    quint64 chunkSize() const { return propagator()->chunkSize(); }
+    quint64 chunkSize() const { return propagator()->initialChunkSize(); }
 
 
 public:
@@ -303,37 +308,13 @@ private:
     quint64 _sent; /// amount of data (bytes) that was already sent
     uint _transferId; /// transfer id (part of the url)
     int _currentChunk; /// Id of the next chunk that will be sent
+    quint64 _currentChunkSize; /// current chunk size
     bool _removeJobError; /// If not null, there was an error removing the job
-    quint64 _lastChunkSize; /// current chunk size
-
-    /*
-    * This is value in ms obtained from the server.
-    *
-    * Dynamic Chunking attribute the maximum number of miliseconds that single request below chunk size can take
-    * This value should be based on heuristics with default value 10000ms, time it takes to transfer 10MB chunk on 1MB/s upload link.
-    *
-    * Suggested solution will be to evaluate max(SNR, MORD) where:
-    * > SNR - Slow network request, so time it will take to transmit default chunking sized request at specific low upload bandwidth
-    * > MORD - Maximum observed request time, so double the time of maximum observed RTT of the very small PUT request (e.g. 1kB) to the system
-    *
-    * Exemplary, syncing 100MB files, with chunking size 10MB, will cause sync of 10 PUT requests which max evaluation was set to <max_single_upload_request_duration_msec>
-    *
-    * Dynamic chunking client algorithm is specified in the ownCloud documentation and uses <max_single_upload_request_duration_msec> to estimate if given
-    * bandwidth allows higher chunk sizes (because of high goodput)
-    */
-    quint64 _requestMaxDuration;
 
     // Map chunk number with its size  from the PROPFIND on resume.
     // (Only used from slotPropfindIterate/slotPropfindFinished because the LsColJob use signals to report data.)
     struct ServerChunkInfo { quint64 size; QString originalName; };
     QMap<int, ServerChunkInfo> _serverChunks;
-
-    quint64 chunkSize() const { return propagator()->chunkSize(); }
-    quint64 maxChunkSize() const { return propagator()->maxChunkSize(); }
-
-    quint64 getRequestMaxDurationDC(){
-        return _requestMaxDuration;
-    }
 
     /**
      * Return the URL of a chunk.
@@ -342,8 +323,8 @@ private:
     QUrl chunkUrl(int chunk = -1);
 
 public:
-    PropagateUploadFileNG(OwncloudPropagator* propagator,const SyncFileItemPtr& item, const quint64& requestMaxDuration) :
-        PropagateUploadFileCommon(propagator,item), _lastChunkSize(0), _requestMaxDuration(requestMaxDuration) {}
+    PropagateUploadFileNG(OwncloudPropagator* propagator,const SyncFileItemPtr& item) :
+        PropagateUploadFileCommon(propagator,item), _currentChunkSize(0) {}
 
     void doStartUpload() Q_DECL_OVERRIDE;
 
--- a/src/libsync/propagateupload.h
+++ b/src/libsync/propagateupload.h
@@ -281,7 +281,7 @@ private:
     int _chunkCount; /// Total number of chunks for this file
     int _transferId; /// transfer id (part of the url)
 
-    quint64 chunkSize() const { return propagator()->initialChunkSize(); }
+    quint64 chunkSize() const { return propagator()->syncOptions()._initialChunkSize; }
 
 
 public:
--- a/src/libsync/propagateuploadng.cpp
+++ b/src/libsync/propagateuploadng.cpp
@@ -32,7 +32,7 @@
 #include <QDir>
 #include <cmath>
 #include <cstring>
-
+#include <cmath>
 namespace OCC {
 
 QUrl PropagateUploadFileNG::chunkUrl(int chunk)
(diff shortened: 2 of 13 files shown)

### CODE REVIEW

File: src/libsync/propagateuploadng.cpp [hunk 1]
```diff
@@ -262,7 +263,42 @@ void PropagateUploadFileNG::startNextChunk()
...
+            qreal log = qLn(requestMaxDurationDC / requestDuration) - 1;
+
+            // If logarithm is smaller or equal zero, it means that we exceeded max request duration
+            // If exceeded it will use currentChunkSize = chunkSize()
+            // If did not exceeded, we will increase the chunk size
+            // motivation for logarithm is specified in the dynamic chunking documentation
+            // TODO: give link to documentation
+            if (log>0){
+                currentChunkSize = qMin(_lastChunkSize + (qint64) log*chunkSize(), maxChunkSize());
+            }
```
mrow4a: The above lines are core algorithm implementation (yes, it is simple)

mrow4a:
How can I solve this? ````/var/lib/jenkins/workspace/cloud-client_client_PR-5368-W5SU4XR3JEFUXQZDXYKOO3KHHMWVP7JL5YGKS7WCUZRTDKRDQMZA/src/libsync/propagateuploadng.cpp:35:18: fatal error: QtMath: No such file or directory
 #include <QtMath>````

mrow4a:
I would love to use the same mechnism for Bundling https://github.com/owncloud/client/pull/5319

src/libsync/propagateuploadng.cpp
```diff
@@ -262,7 +262,42 @@ void PropagateUploadFileNG::startNextChunk()
 
     quint64 fileSize = _item->_size;
     Q_ASSERT(fileSize >= _sent);
-    quint64 currentChunkSize = qMin(chunkSize(), fileSize - _sent);
+
+    quint64 currentChunkSize = chunkSize();
+
+    // this will check if getRequestMaxDurationDC is set to 0 or not
+    double requestMaxDurationDC = (double) getRequestMaxDurationDC();
```
mrow4a: This actually checks if capability is turned on or not

src/libsync/propagateuploadng.cpp
```diff
@@ -262,7 +262,42 @@ void PropagateUploadFileNG::startNextChunk()
...
+            double correctionParameter = log(requestMaxDurationDC / requestDuration) - 1;
+
+            // If logarithm is smaller or equal zero, it means that we exceeded max request duration
+            // If exceeded it will use currentChunkSize = chunkSize()
+            // If did not exceeded, we will increase the chunk size
+            // motivation for logarithm is specified in the dynamic chunking documentation
+            // TODO: give link to documentation
+            if (log>0){
+                currentChunkSize = qMin(_lastChunkSize + (qint64) correctionParameter*chunkSize(), maxChunkSize());
+            }
```
mrow4a: this is main algorithm

src/libsync/propagateuploadng.cpp
```diff
@@ -262,7 +262,42 @@ void PropagateUploadFileNG::startNextChunk()
...
+        }
+
+        //remember the value of last chunk size
+        _lastChunkSize = currentChunkSize;
+    }
+
+    // prevent situation that chunk size is bigger then required one to send
+    currentChunkSize = qMin(currentChunkSize, fileSize - _sent);
```
mrow4a: this is main algorithm

ogoffart:
So this needs a change in the server?

mrow4a:
No, it not even needs the capability from the server, it is enought that you set this 10s - so time to upload 10MB on 1MB/s link on the client. But better if this will be in control of sys admin..

src/libsync/propagateuploadng.cpp
```diff
@@ -262,7 +262,42 @@ void PropagateUploadFileNG::startNextChunk()
...
+
+    quint64 currentChunkSize = chunkSize();
+
+    // this will check if getRequestMaxDurationDC is set to 0 or not
+    double requestMaxDurationDC = (double) getRequestMaxDurationDC();
+    if (requestMaxDurationDC != 0) {
+        // this if first chunked file request, so it can start with default size of chunkSize()
+        // if _lastChunkSize != 0 it means that we already have send one request
+        if(_lastChunkSize != 0){
+            //TODO: this is done step by step for debugging purposes
```
guruz: Please leave it "by step". The compiler will anyway optimize. It is very very very important that we also understand this code in some months.

guruz:
Please mention https://github.com/owncloud/client/issues/4875 in the commit message.

> taking around 2,5second (30% faster): 

Cool! :)

> I would love to use the same mechnism for Bundling #5319

Do you mean for the amount of files in a bundle? Let's not complicate the code too much for now please.. the goal is to get something in a solid state.

mrow4a:
@guruz I think we might need to do that, so the bundle will adjust to the server. Otherwise you will give the same amount of files in bundle both for raspery PI and high performance server and high scale server with high write time.

src/libsync/propagateupload.h
```diff
@@ -290,23 +290,48 @@ class PropagateUploadFileNG : public PropagateUploadFileCommon {
...
+    * This is value in ms obtained from the server.
+    *
+    * Dynamic Chunking attribute the maximum number of miliseconds that single request below chunk size can take
+    * This value should be based on heuristics with default value 10000ms, time it takes to transfer 10MB chunk on 1MB/s upload link.
+    *
+    * Suggested solution will be to evaluate max(SNR, MORD) where:
+    * > SNR - Slow network request, so time it will take to transmit default chunking sized request at specific low upload bandwidth
+    * > MORD - Maximum observed request time, so double the time of maximum observed RTT of the very small PUT request (e.g. 1kB) to the system
+    *
+    * Exemplary, syncing 100MB files, with chunking size 10MB, will cause sync of 10 PUT requests which max evaluation was set to <max_single_upload_request_duration_msec>
```
ckamm: I don't understand how this paragraph relates to the previous one. (or what it is an example of)

src/libsync/propagateupload.h
```diff
@@ -290,23 +290,48 @@ class PropagateUploadFileNG : public PropagateUploadFileCommon {
...
+    *
+    * Dynamic chunking client algorithm is specified in the ownCloud documentation and uses <max_single_upload_request_duration_msec> to estimate if given
+    * bandwidth allows higher chunk sizes (because of high goodput)
```
ckamm: minor: throughput

src/libsync/propagateupload.h
```diff
@@ -290,23 +290,48 @@ class PropagateUploadFileNG : public PropagateUploadFileCommon {
     uint _transferId; /// transfer id (part of the url)
     int _currentChunk; /// Id of the next chunk that will be sent
     bool _removeJobError; /// If not null, there was an error removing the job
+    quint64 _lastChunkSize; /// current chunk size
```
ckamm: This is a good start, but it would be nice if this was shared between PropagateUploads, so they don't have to re-learn for each upload.

src/libsync/propagateuploadng.cpp
```diff
@@ -262,7 +262,42 @@ void PropagateUploadFileNG::startNextChunk()
...
+
+            // calculate natural logarithm
+            double correctionParameter = log(requestMaxDurationDC / requestDuration) - 1;
+
+            // If logarithm is smaller or equal zero, it means that we exceeded max request duration
+            // If exceeded it will use currentChunkSize = chunkSize()
+            // If did not exceeded, we will increase the chunk size
+            // motivation for logarithm is specified in the dynamic chunking documentation
+            // TODO: give link to documentation
+            if (log>0){
```
mrow4a: Sorry, I commit --amend the PR, to make the variables more descriptive and did not notice that, will change it now, this should be correctionParameter

mrow4a:
@ckamm Please check now. I commit --ammend the PR again, now should be ok. Sorry I had changed variables to be more descriptive and did not notice that it broke the condition check.

src/libsync/propagateuploadng.cpp
```diff
@@ -262,7 +262,42 @@ void PropagateUploadFileNG::startNextChunk()
...
+            //TODO: this is done step by step for debugging purposes
+
+            //get last request timestamp
+            double lastChunkLap = (double) _stopWatch.durationOfLap(QLatin1String("ChunkDuration"));
+
+            //get duration of the request
+            double requestDuration = (double) _stopWatch.addLapTime(QLatin1String("ChunkDuration")) - lastChunkLap;
+
+            // calculate natural logarithm
+            double correctionParameter = log(requestMaxDurationDC / requestDuration) - 1;
```
mrow4a: You never reduce the chunking size. You either increase or roll-back to default size.

src/libsync/propagateupload.h
mrow4a: http://ethancbanks.com/2015/03/06/what-is-the-difference-between-throughput-goodput/

src/libsync/propagateupload.h
```diff
@@ -290,23 +290,48 @@ class PropagateUploadFileNG : public PropagateUploadFileCommon {
...
+    * Suggested solution will be to evaluate max(SNR, MORD) where:
+    * > SNR - Slow network request, so time it will take to transmit default chunking sized request at specific low upload bandwidth
+    * > MORD - Maximum observed request time, so double the time of maximum observed RTT of the very small PUT request (e.g. 1kB) to the system
+    *
+    * Exemplary, syncing 100MB files, with chunking size 10MB, will cause sync of 10 PUT requests which max evaluation was set to <max_single_upload_request_duration_msec>
+    *
+    * Dynamic chunking client algorithm is specified in the ownCloud documentation and uses <max_single_upload_request_duration_msec> to estimate if given
+    * bandwidth allows higher chunk sizes (because of high goodput)
+    */
+    quint64 _requestMaxDuration;
```
mrow4a: ok

src/libsync/propagateuploadng.cpp
```diff
@@ -262,7 +262,42 @@ void PropagateUploadFileNG::startNextChunk()
...
+
+            // If logarithm is smaller or equal zero, it means that we exceeded max request duration
+            // If exceeded it will use currentChunkSize = chunkSize()
+            // If did not exceeded, we will increase the chunk size
+            // motivation for logarithm is specified in the dynamic chunking documentation
+            // TODO: give link to documentation
+            if (log>0){
+                currentChunkSize = qMin(_lastChunkSize + (qint64) correctionParameter*chunkSize(), maxChunkSize());
```
mrow4a: Is this state possible if we never get lower? We only increase, starting from default chunk size

mrow4a:
@ckamm
> qBound between minChunkSize, newChunkSize and maxChunkSize, otherwise this could be 0 or negative.

Is this state possible if we never get lower? We only increase, starting from default chunk size. If measurement negative, rollback to default chunk size.

src/libsync/propagateupload.h
ckamm: Oh! I hadn't heard of that term, thanks for the link!

src/libsync/propagateuploadng.cpp
ckamm: I see! With the `if (correctionParameter > 0)` check this means that you only adjust the chunk size upwards if `actual_time < target_time / e`. I guess that's okay.

ckamm:
@mrow4a Yep, the qBound is unnecessary if you only ever adjust the chunk size upwards.

ogoffart:
I suggest this simple  algorithm:

```
  chunkSize = qBound(minValue,  previousChunkSize * targetTime / previousChunkTime  , maxValue)
```

Note that, (esp. for fast network)  the time is bounded by the PHP, and increasing the chunk size will not make the request slower. Meaning we will only converge assymtotically towards the target

For this reason, it is important that the target time is much bigger than the time needed for the PHP to proccess.  Otherwise we would always converge quite fast to the minimum value, which would be the worst outcome.

mrow4a:
Should we allow this to work with both ChunkingNG and ChunkingV1?

guruz:
> Should we allow this to work with both ChunkingNG and ChunkingV1?

No, only new chunking. The old chunking cannot easily support dynamic chunk sizes, don't change what works :)

src/libsync/propagateupload.h
guruz: Maybe BandwidthManager could learn the information.
Although the BandwidthManager is more for limiting currently

mrow4a: I am not fully sure how Bandwidth Manager works, I did not touch it. Of course, we can play with it and test what gives the enhancement easier.

src/libsync/propagateuploadng.cpp
guruz: (Please also add your replies that @ckamm had asked as code comment)

src/libsync/propagateupload.h
guruz: Don't worry about it then. I just meant as a place where you could store a cross-file estimated value fro chunk size

ckamm:
@mrow4a Is there something specific you need to make progress on this?

mrow4a:
I need time, https://cs3.surfsara.nl/ (I need to do tests and measurements) and exam session is coming :>

ckamm:
@mrow4a Understood! All the best with exams!

ckamm:
@mrow4a @guruz @ogoffart I've taken the liberty of taking most feedback and cleaning up this commit. It's pushed as a follow-up fixup commit. The whole thing is not yet rebased onto master.

Please re-review. It looks good from my point of view now!

src/libsync/propagateuploadng.cpp
```diff
@@ -387,6 +389,31 @@ void PropagateUploadFileNG::slotPutFinished()
...
+
+        auto correctedSize = static_cast<quint64>(
+                _currentChunkSize / uploadTime * targetDuration);
+
+        // There can be multiple chunk uploads going on at the same time.
+        // So don't force the chunk size to the new predicted best size
+        // and instead move it there gradually.
+        _propagator->_chunkSize = qBound(
+                _propagator->minChunkSize(),
+                (_propagator->_chunkSize + correctedSize) / 2,
```
ogoffart: Why take the average and not just set `correctedSize` ?   

The comment speaks about multiple chunk upload going on, but i don't see how this is related.

ogoffart:
(I don't like the fact that there is a dependency between the config file and the sync engine.  But since it is already the case, i can't really stop this commit for it.  Anyway, it would be nice if it was part of the "SyncOptions"  struct instead

src/libsync/propagateuploadng.cpp
ckamm: Both work. My reasoning was this:

The whole thing is heuristic. It's possibly that the code is executed after a single chunk finishes uploading when nothing else was going on, but it could also be that several chunks were competing for bandwidth when we get here. If we wanted to target a specific duration per chunk upload this would not give us the right result even if network throughput was completely constant.

Thus I expect `correctedSize` to over and undershoot a good size all the time. Using an exponential moving average like is a cheap way of smoothing chunk sizes a bit.

I could put this reasoning into the code or just remove the averaging, both are fine with me.

ckamm:
@ogoffart I had to rebase on top of master to make `SyncOptions` available.

ckamm:
Merged manually as 53c5f03c323d5bb11595c816a9cc1e1fef01fa0e

Issue #5317, opened 2016-11-16
[Windows] [2.2.4 build 6408] does not re-include unavailable local sync folders when they are available again.


TDannhauer:
### Expected behaviour
After adding a folder sync to a removable drive (e.g. stick or Veracrypt container), the client complains after system boot:  It marks the unaccessible folder with a read cross and says "Unable to find local sync folder."

After making the missing folder accessible, ( e.g. mounting the encrypted device), the missing folder should be re-evaluated (time or event based) by the OC client and included into sync activities as soon as it is available again. 

This would make it convenient to setup local sync folders on encrypted devices. 

### Actual behaviour
However, after making the missing folder accessible, ( e.g. mounting the encrypted device) does not re-enable syncronisation for that folder. Only manually pausing and resuming sync triggers the OC client to re-evaluate the folder, to find it and to re-enable its synchronisation.

### Steps to reproduce
1. 
Setup a folder syncronisation on a accessible removable drive (like a Truecrypt/veracrypt container)
2. 
Make drive unaccessible at bootup.
3. 
Setup Owncloud Client to auto start at user login
4.
Reboot system
5. 
Look at Owncloud client: It marks the unaccessible folder with a read cross and says "Unable to find local sync folder."
6.
Make folder accessible again, e.g. by mounting it in Veracrypt
7. 
Look at Owncloud client: It still marks the unaccessible folder with a read cross and says "Unable to find local sync folder.".
8.
Pause sync for that folder manually.
9.
Resume sync for that folder manually.
10.
Look at Owncloud client: It now finds the accessible folder and re-enables sync for that folder.

### Server configuration
Operating system:
Debian Jessie

Web server: 
Apache 2.4.10

Database:
Postgres 9.3

PHP version:
PHP-FPM 5.6.27

ownCloud version:
9.1.2

Storage backend (external storage):
none

### Client configuration
Client version:
2.2.4 build 6408

Operating system:
Win10 64 bit

OS language:
DE

Installation path of client:
C:\Program Files (x86)\ownCloud

SamuAlfageme:
It's not reproducing in Windows 7/10. The folder is automatically found, no need to Pause/Resume the sync. It just takes a bit (<20s) to be found and synced again.

TDannhauer:
Have you used veracrypt with a encrypted container to reproduce? Are you usi g Win10 1607 64 bit?

I waited 120 min and it highly reproducable. Should i make a screen video?

Torben

Von unterwegs gesendet
[...]

PR #5381, opened 2016-12-13
Regularly retry an unavailable folder

ckamm:
Fix for #5317. If the folder isn't there initially, retry occasionally so users don't have to manually restart.

Note that the case of a folder being there, vanishing and then being there again is very likely still broken. (Does the DB reconnect? Do the FolderWatchers do the right thing?)

### DIFF
--- a/src/libsync/syncengine.cpp
+++ b/src/libsync/syncengine.cpp
@@ -73,7 +73,7 @@ SyncEngine::SyncEngine(AccountPtr account, const QString& localPath,
   , _downloadLimit(0)
   , _newBigFolderSizeLimit(-1)
   , _checksum_hook(journal)
-  , _anotherSyncNeeded(false)
+  , _anotherSyncNeeded(NoFollowUpSync)
 {
     qRegisterMetaType<SyncFileItem>("SyncFileItem");
     qRegisterMetaType<SyncFileItem::Status>("SyncFileItem::Status");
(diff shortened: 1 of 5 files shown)


src/libsync/syncengine.cpp
```diff
@@ -1048,7 +1050,9 @@ void SyncEngine::slotItemCompleted(const SyncFileItem &item, const PropagatorJob
 
 void SyncEngine::slotFinished(bool success)
 {
-    _anotherSyncNeeded = _anotherSyncNeeded || _propagator->_anotherSyncNeeded;
+    if (_propagator->_anotherSyncNeeded && _anotherSyncNeeded == NoFollowUpSync) {
```
ogoffart: Don't we want an immediateFollowUp also if the previous flag was WithDelay?

ckamm: Hm. I think I did it this way because I considered WithDelay to be stronger - it will always have an effect, while ImmediateFollowUp only works a couple of times before it is ignored. It's a situation that won't usually come up. I'd keep it as is and add a comment. Agreed?

ckamm:
@ogoffart I think the folder should already be re-scheduled after the long interval expires, this is about making it more responsive.

src/libsync/syncengine.cpp
guruz: agreed

Issue #5370, opened 2016-12-08
Security Issue with sync client and cookies


Peter-Prochaska:

ogoffart:
The ownCloud client clears the session cookies when disconnecting.
But your  `stack_session` is not a session cookie since it has an expires date.
AFAIK the owncloud server does not send this cookie.
What version of owncloud (client and server) are you using. Which web server or proxy? Who is setting this cookie?

Issue #5305, opened 2016-11-09
Desktop client allows to add the same account multiple times


SamuAlfageme:
### Expected behaviour
An account (user@server combination) should be only added once in the same desktop client.

### Actual behaviour
The client allows to input the same url address **and user** (different users should be fine) multiple times. Resulting in duplicate entries on the "Activity" tab, account tabs ... 

![duplicated accounts](http://i.imgur.com/XIbYZpA.png)

And menu bar dropdown:

![menu dropdown](http://i.imgur.com/pbEpZn8.png)

### Steps to reproduce
1. Open the desktop client and add an account
2. Skip the rest of the config 
3. On the client, select: Settings > Account > Add new
4. Add the same server and credentials from step 1

### Server configuration
Operating system:
CentOS 6.6

Web server: 
Apache

Database:
MySQL

PHP version: 
5.6

ownCloud version: 
9.1.2 Enterprise

Storage backend (external storage):
No

### Client configuration
Client version:
2.3.0.3806-nightly20161109 (build 3806)

Operating system:
OS X 10.11.6

OS language:
English

Installation path of client: 
`/Applications/owncloud.app`

guruz:
I think this is by design. It used to be forbidden but now is allowed.

TDannhauer:
It should be allowed. 
It is not allowed to sync the root node and additionally only a subfolder in a second sync folder setup. The recommendation for this setup is to setup the same account twice - so it must be allowed.

hodyroff:
Lets keep it for now the way it is. Obviously if a user does this, there is an easy workaround. ownBrander (part of Enterprise) allows to create branded clients which only allow one account.

SamuAlfageme:
> It is not allowed to sync the root node and additionally only a subfolder in a second sync folder setup.

After https://github.com/owncloud/client/pull/6032 we can now sync the root `owncloud/` folder multiple times per account. Might be worth looking at this again.

PR #5383, opened 2016-12-14
HTTPCreds: remove all cookies when logging out

ogoffart:
Some custom server use persistent cookies with the auth token. So we should
clear all the cookies when disconnecting.
Account::clearCookieJar is only called from the HTTPCredentials. This funciton
is not used for shibboleth.
There is probably no reasons to keep the HTTP cookie anyway.

Issue #5370

### DIFF
--- a/src/libsync/account.cpp
+++ b/src/libsync/account.cpp
@@ -149,10 +149,14 @@ QUrl Account::davUrl() const
     return Utility::concatUrlPath(url(), davPath());
 }
 
+/**
+ * clear all cookies. (Session cookies or not)
+ */
 void Account::clearCookieJar()
 {
     Q_ASSERT(qobject_cast<CookieJar*>(_am->cookieJar()));
-    static_cast<CookieJar*>(_am->cookieJar())->clearSessionCookies();
+    static_cast<CookieJar*>(_am->cookieJar())->setAllCookies(QList<QNetworkCookie>());
+    emit wantsAccountSaved(this);
 }
 
 /*! This shares our official cookie jar (containing all the tasty


SamuAlfageme:
HTTP basic auth. working like a charm now. Have to try a couple more scenarios, but seems to be all right. 

Once it's done, we should cherrypick this into https://github.com/owncloud/client/releases/tag/v2.2.4 So customers can rebuild their branded clients with this version.

cc/ @michaelstingl

SamuAlfageme:
Leaving here a corner case due to the same account being able to log in multiple times in the same client (https://github.com/owncloud/client/issues/5305)

### Actual behavior:
Both cookies are wiped as they refer to the same username@domain combination.

### Expected behavior:
The cookies could be wiped based on which was logged out and which wasn't

### Steps to reproduce:
1. Log in to the same account with the same sync client
2. Log out of one of the accounts - the other account is fully usable 
3. Close the client and reopen
4. Both accounts ask for password.
5. Enter **one** of the passwords, leave the other logged out.
6. Repeat 3.
7. Both accounts are logged in.

As said, this is a corner case, don't know how relevant.

michaelstingl:
@SamuAlfageme From my perspective, same account logged in multiple times, should be out of scope.

SamuAlfageme:
@owncloud/desktop-developers this is ready to be merged & cherrypicked to v2.2.4 tag, anyone?

guruz:
@SamuAlfageme 2.2.4 is already released and tags can't be changed. Was there a decision to do a 2.2.5 for this?

michaelstingl:
@guruz I need a patch that would work on the 2.2.4 tag. No 2.2.5 required.

guruz:
@michaelstingl https://gist.github.com/guruz/f81cbc4cbcc499f9a87ee74f1fc9e86b (haven't tested it on 2.2.4 though but it should work)

michaelstingl:
@SamuAlfageme Could you test-drive https://gist.github.com/guruz/f81cbc4cbcc499f9a87ee74f1fc9e86b and give a 👍  if it could be applied to 2.2.x?

SamuAlfageme:
After struggling a bit because of some unrelated issues to test the patch over 2.2.4, finally did in both Win7 and OS X, and it's working as expected 🎉

Couple of concerns though: 
1. If we don't make a specific tag containing the patch, building a client with it it's going to be hard from ownbrander.
2. If someone uses a v2.2.4 client with the patch applied, the update notification is triggered; even if the patched version is built on top of 2.2.4, updater will try to install the stable: 

updater_win
 
screenshot 2017-01-05 14 56 37

cc/ @michaelstingl

jturcotte:
Maybe we could just make this an unofficial 2.2.5?

michaelstingl:
> Maybe we could just make this an unofficial 2.2.5?

From my perspective, not required.

jturcotte:
If it's only for ownbrander, I guess they won't have auto-update anyway. But if you're going to create a tag in the sources to be able to build it, then a release number would make sense. We don't want to start maintaining a tree of possible patches to apply on top of a release, this is what releases are meant for. That we have a release number/tag doesn't mean that we have to release it, the patch has no value for the rest of the world anyway.

Issue #5390, opened 2016-12-18
Broken sync time estimation - enterprise setup


mrow4a:
### Scenario
Imagine you are just doing your initial sync on one of the sync clients. Target server is enterprise scale setup, meaning so called "1 Byte PUT duration" takes relatively long compared to home scale server. 

Imagine, that your initial sync folder contains 55000 files and 50GB of data. 

### Observed behaviour
Obviously, sync in this case will consist of small files and big files. When sync client syncs big files, it shows estimation for big files. However, when the client syncs very small files, the estimation shows the estimate for whole sync size compared to very small transfer rates. E.g. for 500B-5kB files, 5kB/s observed transfers for 55000 files and 55GB will show something around 2-3 months. This is obviously not true.. of course unless your sync client will sync 1GB file with speed of 5kB/s.. 

### Expected behavious
Client does not show the dramatic and confusing staticstics. It could be very strange for non technical person to see estimate for 3 months, while the sync will take some hours.

Sync client should show estimate for files under "MINIMUM CHUNKING SIZE" which is to be defined for 1MB, and files over this value. Call it "small files and big files estimation". 

Client should show how many big files are to be synced and how many small files to be synced currently. Correspondingly, it should show e.g. "Estimated 7h for 35000 small files - total 10GB" or "Estimation not available" instead of "3 months needed" or "No big files to sync". Or at least something less confusing. 

I would love to drive a discussion about that and try to implement it in case it is accepted. 
@hodyroff @ogoffart @jturcotte @guruz @butonic @felixboehm

mrow4a:
Can be solved with https://github.com/owncloud/client/issues/5391

davidjericho:
This is a must fix, and it doesn't appear too difficult to resolve. 

We routinely have users complain the estimated time is massively broken, As an example, just this morning, I updated a collection of small files. My desktop is 10Gbps connected, I have less than 1ms to the server infrastructure, which is connected at over 360 Gbps, and it was still forecasting 46 minutes to update 157 files totaling 34MB. The estimated speed was 36kbyte/s when I checked.

Of course, it didn't actually take 46 minutes (rather, less than one minute), but I saw the time estimated jump between less than 10 minutes, to at one point, 37 hours. 

This **does** confuse users, and they do complain about it.

felixboehm:
I thought such would be solved with bundling. @mrow4a ??

mrow4a:
@felixboehm Indeed.

mrow4a:
[...]

Issue #4428, opened 2016-02-03
ETA totally unreliable


ghost:
Dear Developers,

While syncing, the ETA is shown. I find it totally unrealiable. It seems the estimate is "total transfer to be made/ current transfer speed". The truth is there is a "constant cost" if transferring a single file, which is easy to estimate (it is around 0.8 sec in my case). Actually in my case I am pretty sure the sync will be over in approximately two hours, just because the remainder of files consists of larger files.

The "transfer time per file" is given by:

a constant value (in my case 0.8 sec) for a small file
filesize/ transfer rate for large files.
There is a smooth crossover between "small" and "large" file regime. Currently OC wrongly assumes all files are large.

Though I guess it would reordering synchronization order (first send a few small files, and then a large one to get the estimate of both the constant component and the actual transfer speed). Or am I wrong?

![untitled](https://cloud.githubusercontent.com/assets/17043732/12785774/66e97644-ca8e-11e5-8c4f-e797b2889e7d.png)

tflidd:
Remaining transfer time = remaining filesize / current transfer speed = 283470s = 3.28 days
Constant cost = 1654 files * 0.8s/file = 22 min.

The constant cost of each file transfer is somehow negligible at this point. There are more effects (larger files often start slowly and get faster) and the overall performance for a large set of files is hard to predict. Do you know a data transfer tool that does a good eta prediction?

mateusz-lacki:
Nope, but I do not really use them.  Still the data transfer took me approximately 3 hours in the end (upload speed approx 500 kB/s). The estimate for the "remaining time" was changing every second between 1 and 7 days (depending whether the "small file" was say a few or a few dozens kB).

Of course if transfer speeds are not stable, then it is not possible to predict the total time at all. 

But:
- assuming the network connection is good
- server is not under heavy load
- there are no hdd issues

The copy time depends mainly on filesize.
  
So after the program has copied say a few hundred files of various sizes, it should be possible to estimate reliable the ETA (by reliably I also mean under these assumptions).

tflidd:
On a given connection with a given client/server-combination, you can probably figure out something more sophisticated. But does this easily apply for the other cases?

You could as well group the small files, put them in a zip-archive and transfer this archive (and unpack on server side). This way, the predictions are better (all files are about the same size, because big files get chunked) and the upload faster.

mateusz-lacki:
[...]

PR #5395, opened 2016-12-20
Increase initial bandwidth estimates #4428 #5390

ckamm:
To 2 MB/s and 10 files/s.


mrow4a:
I think this still does not solve https://github.com/owncloud/client/issues/5390, it will only do it for first 10 requests. Am I mistaken?

ckamm:
Yes, this doesn't solve anything. At best it makes the problem less obvious. It should have nothing to do with the first 10 requests.

PR #5396, opened 2016-12-20
Fix Mac PATH command

phil-davis:
Reported in the doc repo https://github.com/owncloud/documentation/issues/1618

### DIFF
--- a/doc/building.rst
+++ b/doc/building.rst
@@ -83,7 +83,7 @@ To set up your build environment for development using HomeBrew_:
 
 3. Add Qt from brew to the path::
 
-    export PATH=/usr/local/Cellar/qt5/5.x.y/bin/qmake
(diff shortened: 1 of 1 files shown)


doc/building.rst
```diff
@@ -83,7 +83,7 @@ To set up your build environment for development using HomeBrew_:
 
 3. Add Qt from brew to the path::
 
-    export PATH=/usr/local/Cellar/qt5/5.x.y/bin/qmake
+    export PATH=$PATH:/usr/local/Cellar/qt5/5.x.y/bin/qmake
```
jturcotte: Nice catch, just to make sure I'd put the qt5 path first just in case a Qt4 is already in PATH. `/qmake` should probably also not be in PATH.

`export PATH=/usr/local/Cellar/qt5/5.x.y/bin:$PATH`

phil-davis:
See #5399 - easier than having 2 commits here for such a small change.

PR #5399, opened 2016-12-20
Fix Mac PATH command

phil-davis:
Reported in the doc repo owncloud/documentation#1618
This version as suggested by @jturcotte


phil-davis:
Yeh, but it is really easy for small things like this to just do it in the online GitHub browser-based GUI rather than messing about with putting the branch on my local git repo and doing stuff to it.

PR #5400, opened 2016-12-20
Delete finished job in the PropagateDirectory scheduleNextJob. #5269

mrow4a:
Delete finished job in the PropagateDerectory scheduleNextJob. 

This should reduce code complexity from https://github.com/owncloud/client/pull/5274 and solve the issue https://github.com/owncloud/client/pull/5269

### DIFF
--- a/src/libsync/owncloudpropagator.cpp
+++ b/src/libsync/owncloudpropagator.cpp
@@ -588,17 +588,12 @@ OwncloudPropagator *PropagatorJob::propagator() const
 PropagatorJob::JobParallelism PropagateDirectory::parallelism()
 {
     // If any of the non-finished sub jobs is not parallel, we have to wait
-
-    // FIXME!  we should probably cache this result
-
-    if (_firstJob && _firstJob->_state != Finished) {
-        if (_firstJob->parallelism() != FullParallelism)
-            return WaitForFinished;
+    if (_firstJob && _firstJob->parallelism() != FullParallelism) {
+        return WaitForFinished;
     }
 
-    // FIXME: use the cached value of finished job
     for (int i = 0; i < _subJobs.count(); ++i) {
-        if (_subJobs.at(i)->_state != Finished && _subJobs.at(i)->parallelism() != FullParallelism) {
+        if (_subJobs.at(i)->parallelism() != FullParallelism) {
             return WaitForFinished;
         }
     }
@@ -629,15 +624,8 @@ bool PropagateDirectory::scheduleNextJob()
         return false;
     }
 
-    // cache the value of first unfinished subjob
     bool stopAtDirectory = false;
-    int i = _firstUnfinishedSubJob;
-    int subJobsCount = _subJobs.count();
-    while (i < subJobsCount && _subJobs.at(i)->_state == Finished) {
-      _firstUnfinishedSubJob = ++i;
-    }
-
-    for (int i = _firstUnfinishedSubJob; i < subJobsCount; ++i) {
+    for (int i = 0; i < _subJobs.size(); ++i) {
         if (_subJobs.at(i)->_state == Finished) {
             continue;
         }
@@ -665,8 +653,22 @@ bool PropagateDirectory::scheduleNextJob()
 
 void PropagateDirectory::slotSubJobFinished(SyncFileItem::Status status)
 {
+    PropagatorJob *subJob = static_cast<PropagatorJob *>(sender());
+
+    // Delete the job and remove it from our list of jobs.
+    subJob->deleteLater();
+    bool wasFirstJob = false;
+    if (subJob == _firstJob.data()) {
+        wasFirstJob = true;
+        _firstJob.reset();
+    } else {
+        int i = _subJobs.indexOf(subJob);
+        Q_ASSERT(i >= 0);
+        _subJobs.remove(i);
+    }
+
     if (status == SyncFileItem::FatalError ||
-            (sender() == _firstJob.data() && status != SyncFileItem::Success && status != SyncFileItem::Restoration)) {
+            (wasFirstJob && status != SyncFileItem::Success && status != SyncFileItem::Restoration)) {
         abort();
         _state = Finished;
         emit finished(status);
@@ -675,16 +677,10 @@ void PropagateDirectory::slotSubJobFinished(SyncFileItem::Status status)
         _hasError = status;
     }
     _runningNow--;
-    _jobsFinished++;
-
-    int totalJobs = _subJobs.count();
-    if (_firstJob) {
-        totalJobs++;
-    }
 
     // We finished processing all the jobs
     // check if we finished
-    if (_jobsFinished >= totalJobs) {
+    if (!_firstJob && _subJobs.isEmpty()) {
         Q_ASSERT(!_runningNow); // how can we be finished if there are still jobs running now
         finalize();
     } else {
--- a/src/libsync/owncloudpropagator.h
+++ b/src/libsync/owncloudpropagator.h
@@ -192,14 +192,12 @@ public:
 
     SyncFileItemPtr _item;
 
-    int _jobsFinished; // number of jobs that have completed
     int _runningNow; // number of subJobs running right now
     SyncFileItem::Status _hasError;  // NoStatus,  or NormalError / SoftError if there was an error
-    int _firstUnfinishedSubJob;
 
     explicit PropagateDirectory(OwncloudPropagator *propagator, const SyncFileItemPtr &item = SyncFileItemPtr(new SyncFileItem))
         : PropagatorJob(propagator)
-        , _firstJob(0), _item(item),  _jobsFinished(0), _runningNow(0), _hasError(SyncFileItem::NoStatus), _firstUnfinishedSubJob(0)
+        , _item(item), _runningNow(0), _hasError(SyncFileItem::NoStatus)
     { }
 
     virtual ~PropagateDirectory() {
--- a/test/testsyncengine.cpp
+++ b/test/testsyncengine.cpp
@@ -213,6 +213,17 @@ private slots:
         }
     }
 
+    void abortAfterFailedMkdir() {
+        FakeFolder fakeFolder{FileInfo{}};
+        QSignalSpy finishedSpy(&fakeFolder.syncEngine(), SIGNAL(finished(bool)));
+        fakeFolder.serverErrorPaths().append("NewFolder");
+        fakeFolder.localModifier().mkdir("NewFolder");
+        // This should be aborted and would otherwise fail in FileInfo::create.
+        fakeFolder.localModifier().insert("NewFolder/NewFile");
+        fakeFolder.syncOnce();
+        QCOMPARE(finishedSpy.size(), 1);
+        QCOMPARE(finishedSpy.first().first().toBool(), false);
+    }
 };
 
 QTEST_GUILESS_MAIN(TestSyncEngine)


mrow4a:
@felixboehm @guruz

src/libsync/owncloudpropagator.cpp
```diff
@@ -618,30 +623,31 @@ bool PropagateDirectory::scheduleNextJob()
         return false;
     }
 
-    // cache the value of first unfinished subjob
     bool stopAtDirectory = false;
-    int i = _firstUnfinishedSubJob;
-    int subJobsCount = _subJobs.count();
-    while (i < subJobsCount && _subJobs.at(i)->_state == Finished) {
-      _firstUnfinishedSubJob = ++i;
-    }
```
mrow4a: This will fix this sophisticated caching. Is there any reason to keep Finished Jobs?

mrow4a:
The only place client could use the "job" is in https://github.com/owncloud/client/blob/master/src/libsync/propagatedownload.cpp#L425 and https://github.com/owncloud/client/blob/master/src/libsync/owncloudpropagator.cpp#L557

However, it checks only for Running ones, so there is no problem.

src/libsync/owncloudpropagator.h
```diff
@@ -188,18 +188,18 @@ class OWNCLOUDSYNC_EXPORT PropagateDirectory : public PropagatorJob {
     QScopedPointer<PropagateItemJob>_firstJob;
 
     // all the sub files or sub directories.
-    QVector<PropagatorJob *> _subJobs;
+    QList<PropagatorJob *> _subJobs;
```
guruz: Not sure if this is good, https://marcmutz.wordpress.com/effective-qt/containers/
@ogoffart might have to say more

src/libsync/owncloudpropagator.cpp
```diff
@@ -578,16 +578,16 @@ PropagatorJob::JobParallelism PropagateDirectory::parallelism()
...
 
-    // FIXME: use the cached value of finished job
-    for (int i = 0; i < _subJobs.count(); ++i) {
-        if (_subJobs.at(i)->_state != Finished && _subJobs.at(i)->parallelism() != FullParallelism) {
+    QListIterator<PropagatorJob *> subJobsIterator(_subJobs);
+
+    while (subJobsIterator.hasNext()) {
+        subJobsIterator.next();
+        if (subJobsIterator.peekPrevious()->_state != Finished && subJobsIterator.peekPrevious()->parallelism() != FullParallelism) {
             return WaitForFinished;
```
ckamm: minor: Would prefer `const auto job = subJobsIterator.next();` and then use `job` over `subJobsIterator.peekPrevious()`. Usage of `peekPrevious()` makes me think that something special is happening. Same in the other loop below.

I'd even be fine calling `subJobsIterator` `it`.

src/libsync/owncloudpropagator.h
ckamm: You can delete from the front with QVector and we might want to go for memory over speed here. But I don't have confidence of the best choice here. QVector should be fine.

mrow4a:
@ogoffart Do you see anything from logic perspective ? Should we check it in all possible scenarious to find out?

ckamm:
@mrow4a I looked at it from a `_subJobs` logic point of view.

* Previously jobs where deleted by `qDeleteAll(_subJobs)`. I don't think most jobs delete themselves. Probably need explicit deletion when they are removed from the list.
* It is no longer safe to call `append(Job)` while a directory job is running. Nothing uses that as far as I am aware, but maybe it makes sense to adjust the total job count when append is called? From a different point of view: Why do we need to track `_jobsFinished` and compare it to `totalJobs` in the first place? We can probably drop these and detect being finished in a different way.

I didn't see any other issues.

mrow4a:
@ckamm I think the simplest way could be to track _jobsFinished, it is very self explaining and easy to understand for other developer. I did not really think of any other solution because that one was seeming quite nice.

BTW: If it will occur that we can merge this PR, I think we can safaly say that 2.3 decreases both memory and CPU usage, we get rid of unnecessary loops and in-memory objects.

guruz:
@mrow4a Please change QList back to QVector.

Also add a qFatal please on add instead of assert.

(I didn't look at the code itself, just infering from you and @ckamm 's comments)

mrow4a:
@guruz If we would use QVector with iterators, complexity is O(n). If so, I would rather stay with Caching as it is now and dont use this PR.

mrow4a:
@guruz I think we would need to use the class which allows delete from any point in the list e.g. QLinkedList instead of QList(Qt custom implementation?) or QVector

jturcotte:
@mrow4a QList behaves exactly as QVector when sizeof(T) == sizeof(void*). In both cases deleting from the front will only involve a memmove since pointers are POD types without destructors.

A QLinkedList *might* scale better, but we need to be conservative with memory during sync, so as @ckamm mentioned, not sure either if it's worth it. If you can't see QVector operations in a profiler while syncing, it's definitely not worth it.

mrow4a:
@jturcotte I think we would be very conservative with memory if deleting every finished items compared to current Vector implementation which stores everything in memory.

guruz:
@jturcotte @mrow4a One could set the "deleted" pointer in the qlist/vector to 0 instead of using QLinkedList

mrow4a:
@guruz yes this is another approach, however you would need to keep current cache implementation not to perform unnecesary loops

src/libsync/owncloudpropagator.cpp
```diff
@@ -618,30 +623,31 @@ bool PropagateDirectory::scheduleNextJob()
...
-        if (_subJobs.at(i)->_state == Finished) {
+    QMutableListIterator<PropagatorJob *> subJobsIterator(_subJobs);
+
+    while (subJobsIterator.hasNext()) {
+        subJobsIterator.next();
+        // get the state of the state of the sub job pointed by call next()
+        // peekPrevious() will directly access the item through hash in the QList at that subjob
+        if (subJobsIterator.peekPrevious()->_state == Finished) {
+            // if this items is finish, remove it from the _subJobs list as it is not needed anymore
+            subJobsIterator.remove();
```
jturcotte: Won't this leak the PropagatorJob? I'm surprised that you could just delete jobs without crashes, it's possible that you'll start facing issues if you actually delete the job. Please test thoroughly with very large syncs.

src/libsync/owncloudpropagator.cpp
```diff
@@ -604,6 +604,11 @@ bool PropagateDirectory::scheduleNextJob()
     if (_state == NotYetStarted) {
         _state = Running;
 
+        // at the begining of the Directory Job, update expected number of Jobs to be synced
```
jturcotte: Please start your comments with a capital letter, and end them with a period.

src/libsync/owncloudpropagator.cpp
mrow4a: As I understand, we will now just set pointer to 0? As I understand, I should also call `delete` on class pointer to delete?

jturcotte: That sounds about right.

guruz: remove() does not delete, so setting to 0 does also not delete. You need to find out where it is deleted normally and check if you're not breaking anything by deleting here.

ckamm: Yes, I think it is a leak, also mentioned it in https://github.com/owncloud/client/pull/5400#issuecomment-270365450

mrow4a:
Ok, corrected the required things, also did a test with 5000 files in nested directories, deleting, moving, renaming, downloading, uploading etc. Works correctly. 

I also did a test syncing 1000 files in 10 directories on my local owncloud:
Client version 2.2.4: 67s and 68s
Client this branch: 49s, 47s

Of course, increasing bookkeeping time higher the timing will not be that visible, but generaly it works faster, and because we delete jobs it should be also much more memory friendly. 

@felixboehm @jturcotte @guruz @cdamken @ogoffart @ckamm

src/libsync/owncloudpropagator.cpp
```diff
@@ -618,30 +620,34 @@ bool PropagateDirectory::scheduleNextJob()
...
+        subJobsIterator.next();
+        // Get the state of the state of the sub job pointed by call next()
+        // Function value() will directly access the item through hash in the QList at that subjob
+        if (subJobsIterator.value()->_state == Finished) {
+            // If this items is finish, remove it from the _subJobs list as it is not needed anymore
+            // Note that in this case remove() from QVector will just perform memmove of pointer array items.
+            PropagatorJob * jobPointer = subJobsIterator.value();
+            subJobsIterator.remove();
+            delete jobPointer;
             continue;
```
mrow4a: This is a moment in which I am removing the job from the queue and deleting the referenced object since it is not used anymore nowhere.

ckamm:
@mrow4a I'm busy today, but will review tomorrow!

mrow4a:
No need actualy, I included it in https://github.com/owncloud/client/pull/5440. There it only deletes finished Directory jobs, when it detects that they are finished, and all items are queued in _finishedSubJobs queue. This is safer and ensures that slotSubJobFinished doesnt run in any race condition.

ckamm:
@mrow4a Hm, were they dependent somehow? I'll review here anyway, since this PR has all the comments about the current state of this patch.

ckamm:
@mrow4a The changes here look good and I'd be okay with merging. 

I still have a slight worry about someone calling `append()` while the directory job is running. Could you either document or fix that?

mrow4a:
We cannot merge this, argument of @jturcotte that we can safely remove from QVector is not working. I discovered that during tests with HTTP2, removeing from vector is not working on production system as in local environment. 

We either have to use QLinkedList and check, or dont do that at all

jturcotte:
"not working" is broad and I'm pretty sure it can work. OTOH I'd like to look if we can create jobs lazily, so I can have a look how deleting jobs will fit into the code after the refactoring this will require. I'm not a big fan of using QLinkedList except when it has clear benefits.

mrow4a:
@jturcotte If we would refactor the code to create it lazily, it would be really great! I like it!

guruz:
While @mrow4a was probably thinking mostly about CPU, I had a look on peak memory in general 

I've created (and am still adding) a bunch of issues regarding peak memory usage.
https://github.com/owncloud/client/labels/Peak%20Memory

Some seem to be not needed of this here is implemented because it's often about letting go of memory more eagerly when a job is finished.

guruz:
As per IRC, we want something like this in 2.3.
@jturcotte can you take over? @mrow4a is out for the month

src/libsync/owncloudpropagator.cpp
```diff
@@ -665,8 +653,20 @@ bool PropagateDirectory::scheduleNextJob()
...
 {
+    PropagatorJob *subJob = static_cast<PropagatorJob *>(sender());
+
+    // Delete the job and remove it from our list of jobs.
+    subJob->deleteLater();
+    if (subJob == _firstJob.data()) {
+        _firstJob.reset();
+    } else {
+        bool removed = _subJobs.removeOne(subJob);
+        // FIXME: Try debug build
```
ogoffart: remove the comment

src/libsync/owncloudpropagator.cpp
```diff
@@ -665,8 +653,20 @@ bool PropagateDirectory::scheduleNextJob()
...
+        Q_ASSERT(removed);
+    }
+
     if (status == SyncFileItem::FatalError ||
-            (sender() == _firstJob.data() && status != SyncFileItem::Success && status != SyncFileItem::Restoration)) {
+            (subJob == _firstJob.data() && status != SyncFileItem::Success && status != SyncFileItem::Restoration)) {
```
ogoffart: You just cleared _firstJob earlier, so this would always be false!

src/libsync/owncloudpropagator.cpp
jturcotte: Done

src/libsync/owncloudpropagator.cpp
jturcotte: Done, added a test too.

mrow4a:
Can I benchmark it before we merge?

jturcotte:
@mrow4a Unless I messed up the scheduling there is no way that you will notice a difference if you involve a PHP server that takes >100ms to respond for each file. I'll merge it tomorrow, but if you think you caught some issue let me know and we can revert it (could be non-perf related too).

mrow4a:
I need several minutes, to rerun the test several times for many files. I wont check it with HTTP2 though, should I?

jturcotte:
@mrow4a HTTP2 is a layer under, it won't have any effect on the scheduling beside the fact that the server might take maybe 10% less time to respond and trigger the next job. I simulated it with the unit tests that don't even involve a server and I couldn't see a difference (fake HTTP response posted to the event loop), so I doubt that it would make any difference whether you involve HTTP or HTTP2 regarding this patch.

mrow4a:
In HTTP2, I am talking about 100 files in parallel.

mrow4a:
@jturcotte Hmm, I think you are right and it wont make a difference, since you do it in slotSubJobFinished. In my version I did it in scheduleNext, maybe this is why I saw a difference. I think you are right, that having it in slotSubJobFinished, backed that you did a local test, should not have any influence here. Will do the tests in case, but dont expect anything wild.

:+1:

jturcotte:
@mrow4a OK, thanks for testing!

mrow4a:
@jturcotte @ogoffart For HTTP2, comparing branches 2.3 and 2.3-pre1, having sync of 1000 files of 1kB, total 1MB, 100 files in parallel to CERNBOX EOS, on WiFi and 52 ms latency, repeating 5 times, in average your PR drops upload from 19s to 21s.

jturcotte:
@mrow4a Please allow me to be skeptical, but I don't see how this patch could lead to those results, not for a vector of less than 1000 elements. Could you paste the 5 run times of each configuration? What's the standard deviation between runs?

mrow4a:
https://cloudstor.aarnet.edu.au/plus/index.php/s/jWBwy6gm2ZtqSYi

But yes, it was done first one set, then the other, so it could be also variation of the system itself (which EOS has a lot during the day). I should do it interleaving runs and at night though to be sure.

mrow4a:
I placed also results for HTTP1 in shared line,  and I indeed can see that especially for http1 with 6 connections the veriations can be quite big, so I dont even try to deduce sth from there. However they oscillate around the one line. 

Bet that this for HTTP1 does not make any difference, maybe for HTTP2 with a lot of files in parallel though

mrow4a:
Anyways, I think the benefit of this is bigger than the use case of not having this PR. I just remember that during test we had a little bigger difference in the previous version of this PR in HTTP2 so it alarmed me. 

Anyways, :+1:

PR #5401, opened 2016-12-20
Correctly handle possible missing GIT_SHA1.

kyrofa:
Currently, if get_git_head_revision() doesn't define GIT_SHA1 (which it doesn't in some cases) it's possible to end up with the following error:

    CMake Error at CMakeLists.txt:69 (if):
      if given arguments:

        "STREQUAL" "GITDIR-NOTFOUND"

      Unknown arguments specified

This PR fixes this issue by making sure both left and right hand arguments to STREQUAL are always strings, even if GIT_SHA1 is undefined.


guruz:
Thank you!

PR #5403, opened 2016-12-20
Add capability to disable parallel chunked upload #5364

mrow4a:
Reviewed https://github.com/owncloud/client/pull/5364 needed technical correction and tests

Tested agains oc92 in following scenarios:
- **chunking capability set to true on server**: triggers chunkingNG
- **chunking capability set to false on server**: triggers old chunking with parallel chunk upload
- **chunkingParallelUploadDisabled capability not existing** : triggers old chunking with parallel chunk upload
- **chunkingParallelUploadDisabled set to true**: triggers old chunking without parallel chunk upload
- **chunkingParallelUploadDisabled set to false**: triggers old chunking with parallel chunk upload
- **chunking capability set to false on server and chunkingParallelUploadDisabled set to true**: triggers old chunking without parallel chunk upload

@moscicki @ogoffart

### DIFF
--- a/src/libsync/capabilities.cpp
+++ b/src/libsync/capabilities.cpp
@@ -116,5 +116,9 @@ bool Capabilities::chunkingNg() const
     return _capabilities["dav"].toMap()["chunking"].toByteArray() >= "1.0";
 }
 
+bool Capabilities::chunkingParallelUploadDisabled() const
+{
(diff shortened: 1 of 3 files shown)


mrow4a:
@guruz Merge?

guruz:
Merged. I didn't do it because we have different opinions in the team about who does the merge.
Let's agree that after it's reviewed the submitter can merge..

Issue #5391, opened 2016-12-18
In the mixed file size scenario bandwidth is hardly underutilized - enterprise setup


mrow4a:
### Setup information
Target server is enterprise scale setup, meaning so called "1 Byte PUT duration" takes relatively long compared to home scale server. This is also server on which "100MB PUT duration" on high-bandwidth client takes relatively faster compared to home scale server.

### Scenario 1
Imagine you are just doing your sync after a time of longer inactivity on one of your devices. You used your web interface, mobile and other devices - e.g. work computer - to add your files. You have many small files below 1MB to be uploaded/downloaded, many between 1MB and 5MB, and some over 5MB. 
Your folder is usualy structured that folders contain usualy files of similar file sizes, with few exceptions. Number of the folders containing changes, is small. Generalizing, typical user case. 

### Scenario 2
Imagine you are just doing your initial sync on one of the sync clients. Imagine, that your initial sync folder contains 55000 files and 50GB of data. You have around 20000 files under 1MB, 35000 files over 1MB. This is contained in around 500 folders

### Actual behaviour
If sync client finds folder with many >5MB files, everything is fine. However, when a sync client visits a folder with many >1MB files and small files PUT takes relatively long, bandwidth is underutilized, since in the same type "raw bytes data" transfer could be performed of bigger files. 

### Solution
The solution is a continuation of Request Scheduler PRs on the client: [Folder items scheduler - evaluation attribute sorted tree](https://github.com/owncloud/client/pull/5349) 

In the implementation, Request Scheduler will construct 2 separate queues (robins) in each of the folders to synchronise:

- list of all files under 1MB
- list of all files over 1MB

Propagator, which is class responsible for scheduling items, will contain new value - queueRobin. 
Directory, which is class responsible for dispatching items within folder that contain changes, will contain new value - maximumFileSize.
 
Request scheduler will perform round robin on each of the queues, visiting them sequentially within one Directory class:
- if scheduler finds in a given directory a file in a queue pointed by robin and the file is not currently in flight, file is processed and robin switched to next robin.
- if scheduler finds in a given directory a file in a queue pointed by robin and the file is currently in flight, robin is switched and scheduler will check another queue.
- if scheduler does not find in a given directory a file in the following robin pointed queue, it calls scheduleNext, to ask another folder if it has the file
- if scheduler does not find a folder which satisfies its robin  (e.g. >1MB) because their minimumFileSize's currently are smaller, it will exclude robin from search. (Only <1MB are left)

[...]

PR #5406, opened 2016-12-21
Discussion - Separate sync specific jobs from data transfer jobs

mrow4a:
This will separate jobs which are directory logic based from jobs which carrry files to be uploaded/donwloaded. As discussed in https://github.com/owncloud/client/issues/5391

This gives following benefits:
- there is a specific class in which you can apply any sync logic you need for uploads/downloads without affecting code complexity of Directory and Propagator classes. 
- all files to be synced are "in-line"

There are following disadvantages:
- sync will first initialize/create/update whole the directory structure on remote/localy, resolve conflicts and ignores, and then start syncing actual files. 

@felixboehm @ogoffart

### DIFF
--- a/src/libsync/owncloudpropagator.cpp
+++ b/src/libsync/owncloudpropagator.cpp
@@ -308,11 +308,18 @@ void OwncloudPropagator::start(const SyncFileItemVector& items)
      * In order to do that we loop over the items. (which are sorted by destination)
      * When we enter a directory, we can create the directory job and push it on the stack. */
 
+    /// root job is initialized as PropagateDirectory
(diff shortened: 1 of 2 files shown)


mrow4a:
This PR just shows how it could be implemented, just try to evalue if this makes/does not make sense.

mrow4a:
Worth noting is that, if implemented this way, you are able to do cross-folder optimisations in sync (Bundling balancing network utilization better etc)

felixboehm:
Using this, what more is needed to bundle across folders?

mrow4a:
Solved by https://github.com/owncloud/client/pull/5440

Issue #5407, opened 2016-12-21
The username in SAML logins doesn't appear next to the URL in the account tab


SamuAlfageme:
### Expected behavior
The username should be shown after the server URL as it does in the regular auth scenario/the account tab tittle itself. 
 
### Current behavior
Just the regular trailing dot used after the username is shown. 

![saml_username](https://cloud.githubusercontent.com/assets/2644445/21389191/a136863a-c780-11e6-996a-5d480fe005d7.png)

### Client configuration
* **Client version:** 2.3.0git self compiled from 0797d5c5254b52dcbf7add5274dee1b10426f2e6 
* **Operating system:** OS X 10.11.6
* **OS language:** English

ckamm:
Same in 2.2.4.

ckamm:
Probably has the same root cause as https://github.com/owncloud/client/issues/5408

ckamm:
Should be fixed by https://github.com/owncloud/client/pull/5409

SamuAlfageme:
👍

PR #5409, opened 2016-12-21
Wizard: Detach saved account from wizard

### LINKED ISSUE #5408: Shibboleth accounts don't work after restart
Create a shibboleth account, restart the client. Afterwards `master` is unable to connect to the account.

The testing propfind gets replies like this:
```
12-21 14:40:12:600 21736 OCC::PropfindJob::finished: PROPFIND request *not* successful, http result code is 200 ""
12-21 14:40:12:600 21736 OCC::ConnectionValidator::slotAuthFailed: ZZZ 200 "\n<html>\n  <head>\n        <title>TestShib Identity Provider Login</title>\n
```
and initial investigation shows that DummyCredentials are used for the shib account after a restart.

ckamm:
Fixes #5408 #5407.

The problem was that cleanup of the credentials page set the
credentials of the account back to dummy, thereby overriding
things like shib usernames.

This should be broken since a932eac832b442cca763197240f036905dd284da.

### DIFF
--- a/src/gui/owncloudsetupwizard.cpp
+++ b/src/gui/owncloudsetupwizard.cpp
@@ -553,6 +553,13 @@ void OwncloudSetupWizard::slotSkipFolderConfiguration()
 AccountState *OwncloudSetupWizard::applyAccountChanges()
 {
     AccountPtr newAccount = _ocWizard->account();
+
+    // Detach the account that is going to be saved from the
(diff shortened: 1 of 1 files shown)


guruz:
👍 From me, however would be nice if you specify the actual clean up. Is it a specific destructor? if yes then just mention the symbol?

PR #5425, opened 2017-01-03
Build: Enable Q_ASSERT in release mode

ckamm:


guruz:
Seems to work fine with the crash reporter on OS X, I get the crash dialogue.

```
01-03 13:41:11:300 36574 unknown: ASSERT: "polltime == 123" in file /Users/guruz/woboq/owncloud/client/mirall/src/gui/accountstate.cpp, line 176
m_crashReporterWChar: /Users/guruz/woboq/owncloud/client/install/owncloud.app/Contents/MacOS/owncloud_crash_reporterDump file was written to: /var/folders/qn/s5d9n8sd45g4sbts7_0m8qxw0000gn/T/8B5722BC-8AC1-496D-B301-C27035C0A7F8.dmp
```

FYI @ogoffart @jturcotte

jturcotte:
Humm is this really a good idea? Qt headers will sometimes assert in situation that would not result in a crash or non-recoverable error. Developers (myself at least) always assume that asserts are not enabled in release builds and are not always all-edge-case-proof.

ogoffart:
I tend to agree with @jturcotte .  Asserts have a performence impact which might not be neglegible in some case.

IMHO asserts should be enabled for nightly and beta builds, and disabled for RC and release.

dragotin:
I would also agree with @jturcotte (for what it is worth). We always said we want to crash in beta and pre-release phase but never in production.

guruz:
@ogoffart @dragotin the discussion should happen in https://github.com/owncloud/client/issues/5429

We will still go with qFatal than to create bogus user data or delete it.

@dschmidt FYI

Issue #5421, opened 2016-12-27
"Ask confirmation before downloading folders larger than X MB" does not work when creating a folder sync connection


SamuAlfageme:
### Expected behavior
After setting a download size limit for folders in the sync client settings:

![larger_than](https://cloud.githubusercontent.com/assets/2644445/21496139/7e59b196-cc1c-11e6-9ffb-62f0cc4a526f.png)

Those remote folders larger than the specified size should be initially unchecked on the wizards.

### Actual behavior
No manual confirmation is asked, everything is selected in the `Choose what to sync` (in both new account and new folder sync connection wizards) and syncs **by default**:

![screenshot_2016-12-27_10_01_36](https://cloud.githubusercontent.com/assets/2644445/21496005/8ccefbec-cc1b-11e6-9919-8497c83d8a9a.png)

However, when the `.csync_journal.db` (or `._sync_[...].db` after #5045) is removed, the expected behavior is triggered:

* The notification is finally displayed:

![screenshot 2016-12-27 09 48 45](https://cloud.githubusercontent.com/assets/2644445/21495744/b739b6a8-cc19-11e6-9fac-2166af283772.png)

* The 3 options to handle these folders are shown: 

![screenshot 2016-12-27 09 52 14](https://cloud.githubusercontent.com/assets/2644445/21495908/c7e88744-cc1a-11e6-8ddf-e82b283cb33f.png)

### Steps to reproduce
1. Configure the sync client to ask confirmation before downloading folders larger than 20MB
2. Create a folder larger than 20MB in the web UI
3. Complete the add account/create new folder sync connection wizard selecting to synchronize a parent the folder just created.

### Client configuration
* Client version: 2.2.4
* Operating system: OS X 10.11.6

This was also reported in https://github.com/owncloud/client/pull/5340#issuecomment-266415969 as this is the workaround used there and causing it to fail on first runs.

SamuAlfageme:
I think this is somehow conflicting with the `Theme::wizardSelectiveSyncDefaultNothing()` concept.

Also related with that: #4581

ckamm:
@SamuAlfageme I don't think this is a bug. Currently the user gets to explicitly choose which folders to sync when the folder sync connection is made. So after the user approves the dialog where all the folders are listed with their sizes, the client doesn't believe that big folders need an additional level of confirmation.

[...]

PR #5426, opened 2017-01-03
Selective sync: Respect big folder threshold #5421

ckamm:
When creating a folder sync connection through the account or folder
wizard we didn't take the big folder threshold into account when
presenting the user the selective sync folder list.

Now big folders are deselected by default. And if that happens a label
is displayed that notifies the user of that fact.

### DIFF
--- a/src/gui/folderwizard.cpp
+++ b/src/gui/folderwizard.cpp
@@ -500,7 +500,11 @@ void FolderWizardSelectiveSync::initializePage()
     QString alias        = QFileInfo(targetPath).fileName();
     if (alias.isEmpty())
         alias = Theme::instance()->appName();
-    _selectiveSync->setFolderInfo(targetPath, alias);
+    QStringList initialBlacklist;
+    if (Theme::instance()->wizardSelectiveSyncDefaultNothing()) {
+        initialBlacklist = QStringList("/");
+    }
+    _selectiveSync->setFolderInfo(targetPath, alias, initialBlacklist);
     QWizardPage::initializePage();
 }
 
--- a/src/gui/folderwizard.cpp
+++ b/src/gui/folderwizard.cpp
@@ -482,9 +482,8 @@ void FolderWizardRemotePath::showWarn( const QString& msg ) const
 FolderWizardSelectiveSync::FolderWizardSelectiveSync(const AccountPtr& account)
 {
     QVBoxLayout *layout = new QVBoxLayout(this);
-    _treeView = new SelectiveSyncTreeView(account, this);
-    layout->addWidget(new QLabel(tr("Choose What to Sync: You can optionally deselect remote subfolders you do not wish to synchronize.")));
-    layout->addWidget(_treeView);
+    _selectiveSync = new SelectiveSyncWidget(account, this);
+    layout->addWidget(_selectiveSync);
 }
 
 FolderWizardSelectiveSync::~FolderWizardSelectiveSync()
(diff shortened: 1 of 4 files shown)


src/gui/folderwizard.cpp
```diff
@@ -482,9 +482,8 @@ void FolderWizardRemotePath::showWarn( const QString& msg ) const
 FolderWizardSelectiveSync::FolderWizardSelectiveSync(const AccountPtr& account)
 {
     QVBoxLayout *layout = new QVBoxLayout(this);
-    _treeView = new SelectiveSyncTreeView(account, this);
-    layout->addWidget(new QLabel(tr("Choose What to Sync: You can optionally deselect remote subfolders you do not wish to synchronize.")));
```
guruz: The SelectiveSyncWidget has an equivalent text?

guruz:
Code looks ok, let's have @SamuAlfageme take a look before merging

src/gui/folderwizard.cpp
ckamm: The new widget is (header label, tree view, big folder notice (if necessary)). The new header text is going to be slightly less specific than before since it always says the same thing. But I think it's okay and nicely unifies some more aspects of selective sync setup.

SamuAlfageme:
From https://github.com/owncloud/client/issues/5421#issuecomment-271523697 -> 👍

ckamm:
@ogoffart Pointed out an issue with the text (phrasing it as a double negative was intentional to highlight the blacklist character of this list) and a problem where, when one expands an explicitly selected large folder, large subfolders can start out as deselected.

ckamm:
@ogoffart I wonder whether we make our lives unnecessarily complicated this way. What if we just considered the selective sync dialog during setup a way for users to create an explicit blacklist. Nothing would be whitelisted: So the first sync would ask users about large folders and possible 'M' permissions.

That has the advantage that only exactly one place has to deal with that logic (namely discovery). It also makes sense from the point of view of the account/folder setup wizards not having all the data we need: If, for instance, it was allowed that a deeply nested folder needs confirmation (say, because it has 'M' permissions) the wizards currently couldn't highlight that because they don't query all folder data. The sync really is the correct place for making these decisions.

ckamm:
@SamuAlfageme: @ogoffart and me discussed this for quite a while. We think the previous behavior was pretty much what we wanted:

* In setup wizards, the user gets presented a list of folders along with their sizes and is free to make decisions about what to sync. (the default-nothing theme fix still makes sense) There's no need to explicitly warn the user about big folders here, since they see them along with the sizes.
* For the first sync, we should treat these choices as given and not pop up warnings about folders the user has just looked at in the wizard.

We could start marking folders that would usually need confirmation in this dialog, but think it's optional and may not be worth the complications.

@ogoffart Do you agree with this summary?

SamuAlfageme:
@ckamm @ogoffart hm, I see your point here. 

In the checklist though, I'd offer some kind of **shortcut/button to check/uncheck all these folders that exceed the threshold without having to go one-by-one** as the reinstall scenario could be tedious in this sense. 

dialog

guruz:
I'm merging this as @SamuAlfageme has tested this.

We need to get 2.3 out of the door so ideally your suggestion https://github.com/owncloud/client/pull/5426#issuecomment-272412550 would go into a new issue so we can do it another time.

guruz:
This was merged as https://github.com/owncloud/client/commit/c8cfb3160efffdcf16a22ec830fd2fc7ae14b070 manually to 2.3 instead of master

PR #5428, opened 2017-01-05
Symlink Support for Dolphin Plugins

octotree:
I use the owncloud client by having a directory `~/.owncloud` which is getting synced and placing symlinks in my home directory `ln -s ~/.owncloud/documents ~/Dokumente`. With the current owncloud client my Dolphin file manager does not show the status of the files when I open the symlinked folders and I cannot use the sharing functionality either.

Therefore, this patch normalizes the directories, to let the plugin know, that the files in the symlinked directories are part of the owncloud sync directory.


guruz:
Thanks! 👍 
squashed into 1 commit

PR #5349, opened 2016-12-01
Prototype: Folder items scheduler - evaluation attribute sorted tree

mrow4a:
Case Scenario 1:
Carlos and his Zombies just went back from holidays, he opens the computer and tries to sync all his photos - 4GB. With his link, to upload all the photos will take around 1 hour. He starts working since next day is monday, and he puts in the various folders his new files. 

Case Scenario 2:
Carlos didnt use sync client for longer time, he want to do the initial sync of his files. He starts sync of a lot of folders with a lot of files. Total sync time will be around 30 minutes. Within all these files, there are some folders on which he collaborates on the document. He want to share the document he is editing during the sync, and this folder is not big (it contains only documents).

EXPECTATIONS:
In the very nested structure, there is a folder with only small changes, and folder with a lot of new files, which sync time will be very long. It is expected, that the folders with small changes will be synced first. If I pause-start sync, my small file schould by synced first

REALITY:
OC client currently is "I just found folder on filesystem and I will try to sync its contents to the cloud" structured. This means, that if you place a small change in some nexted folder tree, and before it there is another nested tree with a lot of files, your document will wait till sync comes to your folder

Solution:
Effect 1:
I removed the caching of the Finished Jobs, and I just remove them from the stack. I am using also Iterator instead of for loop. 

Effect 2:
Oc sync is structurising the sync in the following manner, not regarding on the size of the folders:

THE TREES BELOW ARE TREES OF CHANGES - These folders could contain a lot of unchanged files in each. Tree is being build only basing on size of changes within specific subtree. 

Important note: all file operations(IGNORE, MOVE, ETC) except for PUT and GET, are assigned with HighPriority, it means they will be synced with insertion order (the same as previously). PUT and GET predicates are based on their file sizes and are assigned NormalPriority. Folders (subtrees) are assigned ContainerPriority, which means their predicate is size of the changes within this subtree. REMOVED folders are HighPriority and will be synced old way. 

````
ownCloud Root
├── less-important-folder-10MB/
       ├── nested-folderes-files 9MB
       ├── less-important-file 1MB
       └── nested-folderes-files 4MB
├── Home&Work-10GB/
        ├── folder with nested 10GB of photos to be sync
        ├── very-very-important folder with single documents to sync - 2MB
        └── less-important-document - 1MB
├── important-folder-12MB/
└── less-important-folder - 5MB/
        └── important-file-1MB
````
[...]

Issue #4498, opened 2016-02-22
sync files from smallest first to biggest last


geoffm:
I'm not sure in what order the files are syunced actually, but it would be nice if it ordered them like dropbox, syncing files from smallest to biggest. This allow to complete much faster a bigger part of the multiple files to sync

guruz:
Related: https://github.com/owncloud/client/issues/1633 https://github.com/owncloud/client/issues/2128

Maybe we should do this as part of the activation of pipelining https://github.com/owncloud/client/issues/4454

FYI @ogoffart

Issue #4454, opened 2016-02-10
Investigate HTTP Pipelining


dragotin:
Please investigate
- Is HTTP pipelining a strategy to improve the upload and download performance especially with small files
- Do we enable HTTP pipelining properly in our Qt code? The documentation about `QNetworkRequest::HttpPipeliningAllowedAttribute` says that pipelining is disabled by default, but we do not set this to true.
- Does the ownCloud Server support pipelining, or does it require special configuration. If so, which?
- How can we add a test that verifies that pipelining works?

guruz:
I implemented pipelining in QNAM back in the days. I don't know how well it works these days given our experiences with recent Qt bugs (and given that it is disabled by default it has less testing).

Pipelining is an attribute of the web server, not of the oC/PHP. Web browsers disable it because of possible problems with "middle boxes" but hopefully our customers all use good software/hardware there. Maybe we should only enable it for HTTPS.


As long as oC has unexpected performance characteristics, we cannot enable pipelining in general because we don't know if there is an actual timeout or not because we don't know how long requests take.

We might however want to enable it for downloading (GET) of small files.
QNAM pipelines 3 requests at once into a socket, so if we assume that they all finish in less time than our ```GETFileJob``` timeout we are good.
.oO(maybe ```QNetworkReply``` should have a ``requestSent``` signal)

guruz:
QNAM only supports pipelining for ``GET``` but I'm supe we could upstream a patch for ```DELETE``` and ```MOVE```.

https://code.woboq.org/qt5/qtbase/src/network/access/qhttpnetworkconnection.cpp.html#726

IMHO If we should try to enable this but need to be super careful for things not to break.
.oO(too bad we don't have checksums in general.. @rullzer @ckamm )

dragotin:
@guruz how about PUT?

guruz:
@dragotin Would need a QNAM patch. The QNAM implementation was done back in the days in a very conservative way with focus on usage in a mobile phone web browser.

guruz:
Another caveat: QNAM does not do Pipelining when ```Authorization``` headers are there. But we currently send this header again, see ef9483c82d111464e67291cd134beef99d0afcfc

So maybe we'd need a switch based on the server version which has the cookie-only auth fixed.

guruz:
Moving this out of 2.2
Will investigate instead of to dynamically adjust the parallelism for smaller requests (together with issue #3382 )

guruz:
[...]

PR #5440, opened 2017-01-08
Group requests to ensure balance between bandwidth utilization and bookkeeping #5391 #5390 #4498 #1633 #4454

mrow4a:
This fixes issues: 
#5391 In the mixed file size scenario bandwidth is very underutilized - enterprise setup
#5390 Broken sync time estimation - enterprise setup
#4498 sync files from smallest first to biggest last
#1633 Propagator: Balance concurrent up/down-load
#4454 Investigate HTTP Pipelining

This includes:
https://github.com/owncloud/client/pull/5400 so if this is merged, the other can be closed

This is basic for:
https://github.com/owncloud/client/issues/5319 Bundling
HTTP2

This can be enhanced by:
https://github.com/owncloud/client/pull/5368 Dynamic Chunking
https://github.com/owncloud/client/pull/5349 Sorting folders by modification time

This requires following capability on server side, otherwise will sync old way:
selection_018

### Idea

The idea behind the implemented algorithm is that in the scenario, in which we have a lot of mixed within the same or different folder - big files, under chunking size files of size 100kB or 5MB, files to download, moves, deletes, directory creations etc:

- Separate transfer jobs from jobs which server database interaction is major factor

- Reserve uncoditionaly 4-6 flows for "db sensitive files" like small file uploads and 2 for upload/download >1MB file transfers

- sync normalny if no other files

- Cross folder, this means it will look for "db sensitive" and "data transfer" in all folders.

The above will ensure that there is a balance between bandwidth utilization and bookkeeping, what in turn gives faster synchronisation (for details look below)

Used algorithm:
selection_019

### Checkpoints

- [x] Cross folder, this means it will look for "db sensitive" and "data transfer" in all folders.
- [x] Test manualy to ensure faster sync - passed prove of concept
- [x ] Test on enterprise setup to ensure faster sync
- [x] Write unit tests to cover whole the code 
- [ ] Test with integration tests

### Details
Lets do some maths. This will be our folder to sync:
100 files - average 100kB file size -> total 10MB to be transfered
10 files - average 10MB file size -> total 100MB to be transfered

Assume, that your network is 5MB/s and one "1 Byte PUT" takes 1s (we are not in ideal home case scenario with empty server now guys, it is not that easy, it could take much much more)

The request "latency" consists of 2 components, time it takes for bookkeeping on server and time it takes for data transfer.

Current case

If you do 100 small request in the row, your data transfer is neglible ~0s and bookkeeping time is 100s. Parallelising that maybe you can achieve 33s having 3 parallel flows.

If you do 10 biger files request in a row, your data transfer is 10s bookkeeping. Parallelised say 4s if you are lucky with 3 parallel flows. However, you cannot omit 20s coming from transfer of 100MB, does not matter how many requests you have in parallel 1000 or 1. Your 5MB/s office net bounds you.

Total you need around 33s plus 24s from big, having 57s.

Optimized case

If you do 100 small request in the row, your data transfer is neglible ~0s and bookkeeping time is 100s. Parallelising that maybe you can achieve 50s reserving 2 flow slots. In this 50s, you used neglible bandwidth. If you use the 3rd flow to pump there 30s comming from big files, you just synced your files in 50s, since your 30s are done in parallel filling the bandwidth.

57s vs 50s is like 13% of the time orginal time? In this example it is around 7s. 100 files is not a big deal but shows bigger picture

More big files and more small files, the different is the percentage. Do the math for 55000 files and 55GB. Where you have 40000 of small files in 10GB(avg. size 250kB) and 15000 in the rest 45GB (avg size 3MB). I think we wont be talking about even minutes there :>

@felixboehm @DeepDiver1975 @davidjericho @cdamken @jnweiger


src/libsync/propagateupload.h
```diff
@@ -321,6 +322,99 @@ private slots:
     void slotUploadProgress(qint64,qint64);
 };
 
+/**
+ * @brief The PropagateUploadBundle class is a container class for upload jobs under chunking size.
+ *
+ * It will also ensure proper bandwidth utilization vs bookkeeping balance, and that in case no other items then under chunk uploads are available,
```
phil-davis: I don't understand the detailed meaning of:
"and that in case no other items then under chunk uploads are available,"

phil-davis:
I made some minor edit suggestions for comments in #5441 
The principle of this seems very useful - for example I am a long way (in round-trip-time ping...) from my server and my upload speed is not so good (1 to 3 Mbps range). I often have cases where there are a few big files and a lot of small files. That often results in sub-optimal total transfer time, because a some big files might chug away for a while (3 in parallel), then it gets to a heap of small files and those then have to slowly get created on the server... If we make sure that the upload of lots of small files is running in one stream(s) in parallel with big files in another stream(s) then the things that need server time and the things that need raw transfer time can happen concurrently and the total elapsed time should be minimized.
👍 for the concept. I am not familiar enough with the overall client design to give a good opinion on the detailed implementation.

mrow4a:
Ok, it works awesome, need to check it in enterprise scale. Remember, if you want to test it:

You need a server capability of enviromental capability on the client - otherwise it will sync the old way. 
```
    static const auto bundling = qgetenv("OWNCLOUD_BUNDLING");
    if (bundling == "0") return false;
    if (bundling == "1") return true;

    return _capabilities["dav"].toMap()["bundling"].toByteArray() >= "1.0";
```

@phil-davis Could you rebase and check again! I made a change to be sure that it parallelises correctly for ALL items. It was not previously, so 3 big downloads will block everything if placed in alphabetical order.

src/libsync/capabilities.cpp
```diff
@@ -108,6 +108,15 @@ QByteArray Capabilities::uploadChecksumType() const
     return QByteArray();
 }
 
+bool Capabilities::bundling() const
+{
+    static const auto bundling = qgetenv("OWNCLOUD_BUNDLING");
+    if (bundling == "0") return false;
+    if (bundling == "1") return true;
+
+    return _capabilities["dav"].toMap()["bundling"].toByteArray() >= "1.0";
```
mrow4a: I just reproduced what @ogoffart did with chunking, nearly copy paste from ChunkingNG

src/libsync/owncloudpropagator.cpp
```diff
@@ -389,13 +390,34 @@ void OwncloudPropagator::start(const SyncFileItemVector& items)
...
-        } else if (PropagateItemJob* current = createJob(item)) {
-            if (item->_instruction == CSYNC_INSTRUCTION_TYPE_CHANGE) {
-                // will delete directories, so defer execution
-                directoriesToRemove.prepend(current);
-                removedDirectory = item->_file + "/";
-            } else {
-                directories.top().second->append(current);
+        } else {
+            // Ensure that only new files are inserted into PropagateFiles
+            if (enableBundledRequests && item->_instruction == CSYNC_INSTRUCTION_NEW) {
```
mrow4a: It is basis for dispatching bundles in the future, this is why. You can call this feature bundling 1.0. i can change a name however. I also give a flag, because this is now cross-folder, some people might not like it, so I give them the chance to sync folder by folder just by changing one capability on server.

src/libsync/owncloudpropagator.cpp
```diff
@@ -459,6 +481,20 @@ quint64 OwncloudPropagator::chunkSize()
     return chunkSize;
 }
 
+quint64 OwncloudPropagator::smallFileSize()
+{
+    // A small filesize item is a file whose transfer time
+    // typically will be lower than its bookkeeping time.
+    static uint smallFileSize;
```
mrow4a: Again, copy from ChunkingNG

src/libsync/propagateupload.h
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) by Olivier Goffart <ogoffart@owncloud.com>
+ * Copyright (C) by Piotr Mrowczynski <piotr@owncloud.com>
```
mrow4a: Oh, on server side it makes sense for code maintenance reasons (you know whom to contact for details), can change if it this is different on the client.

src/libsync/propagateupload.h
```diff
@@ -321,6 +322,100 @@ private slots:
...
+ * |                                      |                          |
+ * |                                      |                          |
+ * ^         (Are there at least 2 STANDARD jobs                     |
+ * |          already running within this class?)                    |
+ * |          No  |                      | Yes                       |
+ * |              |                      |                           |
+ *  <-[Schedule STANDARD job]             ->[Try in next container]<-
+ *
+ */
+class PropagateFiles : public PropagatorJob {
```
mrow4a: Because PropagateDirectory is much more complicated thinking about cases to handle, did not want to introduce even more complexity, we will get lost. Also I wanted to make this "pluggable" so you can easily switch it on and off.

mrow4a:
@ckamm Because PropagateDirectory is much more complicated thinking about cases to handle, did not want to introduce even more complexity, we will get lost. 

Also I wanted to make this "pluggable" so you can easily switch it on and off. 

Mind that PropagateFiles is only for new files without any "special" cases, just normal upload/download, chunked or not. No _first job, delete cases, WaitInDirectory etc.

src/libsync/capabilities.cpp
ckamm: Okay, I didn't realize. That doesn't make the comparison any more correct though, even if it does set precedent for having "N.M" versions in the capabilities (which I wasn't aware of). Then I suggest we merge as-is and fix both in a follow-up.

mrow4a:
Hello, 

Changed a little bit a structure of this and been testing it on enterprise setup, details can be found in this document @ogoffart @guruz @jturcotte @felixboehm, which is part of my CS3 presentation : 
SYNC OPTIMIZATIONS - SCHEDULING
[sync-optimizations-mrowczynski-scheduling.pdf](https://github.com/owncloud/client/files/724335/sync-optimizations-mrowczynski-scheduling.pdf)

Features:
- Ensured balance between jobs
- Is entriely plugable, cross-folder and separate from Propagate Directory main logic
- Only New and Updated files are being propagated (data transfers) 
- Jobs are created lazily

src/libsync/propagatefiles.cpp
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    if (_state == NotYetStarted) {
+        _state = Running;
+
+        if (isEmpty()) {
+            finalize();
+            return true;
+        }
+        _subJobs.reserve(_totalItems);
```
mrow4a: I think I need to get rid of that, since I am wasting memory here, dont I?

src/libsync/propagatefiles.cpp
```diff
@@ -0,0 +1,158 @@
...
+        Q_ASSERT(_subJobs.at(i)->_state == Running);
+    }
+
+    return scheduleNextItem();
+}
+
+bool PropagateFiles::scheduleNewJob(QVector<SyncFileItemPtr> &syncJobs){
+    // This function is used to schedule new job and lazily create job from sync items
+    Q_ASSERT(!syncJobs.isEmpty());
+    const SyncFileItemPtr &item = syncJobs.takeFirst();
```
mrow4a: We are planning to use only 5.8 dont we? Or should I still keep compatibility with Qt4?

jturcotte: No the plan is to use 5.6 everywhere we can, and still use Qt4 on linux platforms that we don't ship Qt (unless we manage to bundle Qt). Qt 5.6 is LTS and will be receiving patches until 2019 while 5.8 only until 2018.

mrow4a:
Ok, checked with unit tests and they covered the code, everything is passing, please also mind that https://github.com/owncloud/client/pull/5440/files#diff-7e5082f89a138020f2b1d37fc97d17dbR319 this line has to be adjusted before merging, there is TODO

src/libsync/owncloudpropagator.cpp
```diff
@@ -388,6 +395,9 @@ void OwncloudPropagator::start(const SyncFileItemVector& items)
                 currentDirJob->append(dir);
             }
             directories.push(qMakePair(item->destination() + "/" , dir));
+        } else if (enableScheduledRequests
+              && (item->_instruction == CSYNC_INSTRUCTION_NEW || item->_instruction == CSYNC_INSTRUCTION_SYNC)) {
+            filesJob->append(item);
```
jturcotte: This changes one purpose of the PropagateDirectory structure: update the directory's etag in the database only once child files have been synced properly.

I can't wrap my head about the conditions where this could be an issue, but @ogoffart should know if this could cause concrete problems.

See e.g. https://github.com/owncloud/client/blob/1cec2ca13dc35aba6aaf79f42bde9a6dd00143a7/src/libsync/owncloudpropagator.cpp#L705

mrow4a: Thanks for pointing out, did not know about this, dont think it is difficult to resolve but let @ogoffart see this

mrow4a:
For LAN sync - https://central.owncloud.org/t/gsoc-2017-fast-lan-sync/6271 - we need this PR. Sync clients need to be "on the same page", and be finished with any metadata logic around, so that they can "safely" transfer the data, as in this PR. LAN Sync has to be done only on raw PUT/GETs and this logic has to be abstracted from "Directory sync logic" (basicaly what this PR is doing). @hodyroff @DeepDiver1975

guruz:
@mrow4a: make sure to try to interrupt syncs in the middle and then resume them (also multiple times) and check if this will really still lead to correctly synchronized directories and not to deletes file or files that are not synced.
@mrow4a: please make sure smashbox reliability sync tests pass
@mrow4a: please make sure S3 syncs faster than without the patch

src/libsync/owncloudpropagator.cpp
ckamm: @mrow4a @jturcotte @ogoffart If I'm not mistaken the following is a situation in which that dependency structure could lead to an abort-related problem:

```
Initial state in db and server:
/  - "etag/1"
/A - "etagA1"
/A/F - "etagF1"
```
now when someone touches /A/F on the server, the server state becomes
```
Server state, after touching /A/F
/  - "etag/2"
/A - "etagA2"
/A/F - "etagF2"
```
but if I understand correctly the propagation dependency graph is
```
PropagateDirectory (/)
 |- PropagateDirectory (/A)
 |- PropagateFiles
    |- PropagateItem (/A/F)
```
meaning that /A and /A/F run independently of each other. So it would be possible to completely finish propagating /A before the file transfer /A/F is done. Then aborting the sync run could lead to the db tree
```
/  - "etag/1"
/A - "etagA2"
/A/F - "etagF1"
```
Which means a follow up sync would probably not pick up on /A/F being out of date.

Maybe this hints at a second dependency problem: Currently local and remote MkDir are FullParallelism - so isn't there a chance that with this change one could run into a case where the client wants to download or upload into a non-existant directory?

Possibly I'm missing something, feel free to point out incorrectness.

mrow4a: No it wont upload to non-existent directory, because directory structure / resolving conflicts etc are done before any transfers, and directory deletions are done after https://github.com/owncloud/client/pull/5440/files#diff-c9731f430e8a29b13deaaf73bc4a4e22R56
https://github.com/owncloud/client/pull/5440/files#diff-20b960bb10cf3c0781a80fb3e5775241R493
https://github.com/owncloud/client/pull/5440/files#diff-7e5082f89a138020f2b1d37fc97d17dbR420

About the etag propagation did not look into the problem yet, because we dont have yet smashbox automation for PRs (it is nearly done) and I am busy on the server side before the release.

ckamm: @mrow4a Okay, I didn't see that. (btw, _runningNow is gone, so this branch doesn't compile anymore) In that case the second problem can indeed not happen.

ogoffart:
I removed the milestone because this would still need more work, and I don't think it is worth it at all. But I leave it open for the sake of discussion.

ogoffart:
Closing outdated pull request

Issue #5344, opened 2016-11-29
delete stale chunked upload transfers more aggresively


moscicki:
This is about the current major flow of the V1 chunking algorithm which we need to fix in V1 because of deployment reasons (cannot wait for full NG support).

Currently when an upload of a file fails, it is blacklisted after 3 retries and removed from blacklist and tried again after some time. This is good for plain PUT but not for chunked PUT. If the chunk fails consistently (*), this process will never stop and the client will try indefinitely (I see in the logs 100s of failed PUTs over and over again for several users). What is needed in this case is resetting of the entire chunked transfer from scratch. Manually I can achieve this by: 

    sqlite> delete from uploadinfo where path="xxx";

But I cannot ask users to do it manually.

Looking into the source code, the cleanup of stale upload transfers is already implemented in src/libsync/syncengine.cpp as deleteStaleUploadInfos(). However it will not cleanup such failed chunk transfers in a way which allows the client to restart the whole chunked upload.

Could we declare that after 3 (or more) failed retries the entire chunk transfer is reset, gets the new transfer id and is retried from scratch? Alternatively, at client startup, can we remove all entries for chunked transfers from uploadinfo? At least this would give us a clear operational procedure to restart the client (not as nice as automatic handling of the error but still workable for the time being)? 

(*) consistent chunk failing may happen when the server enters into a bad state for the particular transfer id. In the case of our storage it could happen that chunks came out of order at some point and our server which does not support out-of-order chunk currently will always fail for this transfer id. What needs to be done is to start the transfer from scratch with a new transfer id.

Kawohl:
@guruz @ogoffart

ogoffart:
This should have been fixed a year ago as part of issue #3934

ogoffart:
What is the version of the client with this problem?

ogoffart:
Ah ok, i think i misunderstood the problem.  This was another issue.

moscicki:
Cool. Could I try it out with some nightly?

guruz:
@moscicki Should be in tomorrow morning's nightly

glpatcern:
Hi,

We (I and @moscicki) have tested the fix, but it seems that it unveiled another issue. The commit correctly assumes that the retryCount is updated, but this does not seem to be the case as the uploadinfo.errorcount field stays at 0:
```
$ sqlite3 --header ._sync_985a2109c4ff.db 'select * from uploadinfo;'
path|chunk|transferid|errorcount|size|modtime
[...]

Issue #5366, opened 2016-12-07
Glitchy sqlite3 on ci server? "disk I/O error"


guruz:
Will need to see why this happens.
Maybe because of old Qt version? old sqlite3 version?
I've seen this a few times..

```
********* Start testing of TestOwnSql *********
Config: Using QtTest library 5.3.2, Qt 5.3.2
PASS   : TestOwnSql::initTestCase()
PASS   : TestOwnSql::testOpenDb()
QDEBUG : TestOwnSql::testCreate() Sqlite exec statement error: 10 "disk I/O error" in "CREATE TABLE addresses ( id INTEGER, name VARCHAR(4096), address VARCHAR(4096), entered INTEGER(8), PRIMARY KEY(id));"
FAIL!  : TestOwnSql::testCreate() 'q.exec()' returned FALSE. ()
   Loc: [/var/lib/jenkins/workspace/ncloud-client_client_master-FVKMGQRMEBLP2FT6I5VKJF3NR7ONJIT6DT36KUYQTVUJMTC3LXJA/test/testownsql.cpp(54)]
PASS   : TestOwnSql::testIsSelect()
QDEBUG : TestOwnSql::testInsert() Sqlite exec statement error: 8 "attempt to write a readonly database" in "INSERT INTO addresses (id, name, address, entered) VALUES (1, 'Gonzo Alberto', 'Moriabata 24, Palermo', 1403100844);"
FAIL!  : TestOwnSql::testInsert() 'q.exec()' returned FALSE. ()
   Loc: [/var/lib/jenkins/workspace/ncloud-client_client_master-FVKMGQRMEBLP2FT6I5VKJF3NR7ONJIT6DT36KUYQTVUJMTC3LXJA/test/testownsql.cpp(71)]
QDEBUG : TestOwnSql::testInsert2() Sqlite exec statement error: 8 "attempt to write a readonly database" in "INSERT INTO addresses (id, name, address, entered) VALUES (?1, ?2, ?3, ?4);"
FAIL!  : TestOwnSql::testInsert2() 'q.exec()' returned FALSE. ()
   Loc: [/var/lib/jenkins/workspace/ncloud-client_client_master-FVKMGQRMEBLP2FT6I5VKJF3NR7ONJIT6DT36KUYQTVUJMTC3LXJA/test/testownsql.cpp(83)]
PASS   : TestOwnSql::testSelect()
PASS   : TestOwnSql::testSelect2()
[...]

PR #5443, opened 2017-01-10
Reset stuck chunked uploads eventually #5344

ckamm:
Previously this wasn't happening for errors that were not
NormalErrors because they don't end up in the blacklist.

This revises the resetting logic to be independent of the
error blacklist and make use of UploadInfo::errorCount
instead.

412 errors should reset chunked uploads because they might be
indicative of a checksum error.

Additionally, server bugs might require that additional
errors cause an upload reset. To allow that, a new capability
is added that can be used to advise the client about this.


guruz:
I'm merging this, the Jenkins failures is unrelated https://github.com/owncloud/client/issues/5366

PR #5446, opened 2017-01-11
shell: Only push status for requested files

jturcotte:


jturcotte:
I haven't tested with the Dolphin and Nautilus plugins. I see that Dolphin sends a `RETRIEVE_FILE_STATUS` even for files in the cache, so that shouldn't impact it.

I'm not sure for Nautilus, it should be alright if `UPDATE_VIEW` re-sends a `RETRIEVE_FILE_STATUS`, but I can't figure out if this is the case from reading the code.

shell_integration/windows/OCUtil/RemotePathChecker.cpp
```diff
@@ -172,10 +176,13 @@ RemotePathChecker::~RemotePathChecker()
     CloseHandle(_newQueries);
 }
 
-vector<wstring> RemotePathChecker::WatchedDirectories()
+void RemotePathChecker::UpdateWatchedDirectories(vector<wstring>& localWatchedDirectories, unsigned& localVersion) const
 {
+    if (localVersion == _watchedDirectoriesVersion)
```
ogoffart: Access to "_watchedDirectoriesVersion"  without the lock!
Put it inside the lock or make it std::atomic

shell_integration/windows/OCOverlays/OCOverlay.cpp
```diff
@@ -126,29 +127,27 @@ IFACEMETHODIMP OCOverlay::GetPriority(int *pPriority)
...
     }
 
-    if (!watched) {
+    if (!watched)
         return MAKE_HRESULT(S_FALSE, 0, 0);
-    }
 
     int state = 0;
-    if (!checker->IsMonitoredPath(pwszPath, &state)) {
+    if (!checker->IsMonitoredPath(pwszPath, &state))
```
ogoffart: (style nitpick:  the coding style was that every if should have '{', even if one line. But now that klaas is gone, maybe we can change that :-) )

src/gui/socketapi.cpp
```diff
@@ -56,16 +58,101 @@
 // The second number should be changed when there are new features.
 #define MIRALL_SOCKET_API_VERSION "1.0"
 
+#define DEBUG qDebug() << "SocketApi: "
```
ogoffart: we could start using Qt5's categorized login  (out of scope for this change)

shell_integration/windows/OCUtil/RemotePathChecker.cpp
jturcotte: Ah I thought that reading from an int was safe, I need an atomic even in that case?

shell_integration/windows/OCUtil/RemotePathChecker.cpp
```diff
@@ -90,15 +93,13 @@ void RemotePathChecker::workerThreadLoop()
                     // Remove any item from the cache
                     for (auto it = _cache.begin(); it != _cache.end() ; ) {
                         if (StringUtil::begins_with(it->first, responsePath)) {
+                            SHChangeNotify(SHCNE_UPDATEITEM, SHCNF_PATH | SHCNF_FLUSHNOWAIT, it->first.data(), NULL);
```
jturcotte: The `SHCNF_FLUSHNOWAIT` should normally take care of that, is it possible that it wasn't there when you saw the deadlock?

I don't want to take the chance though, if you're not sure it might be better to get it out of the lock.

shell_integration/windows/OCUtil/RemotePathChecker.cpp
jturcotte: Humm yeah it's safe but I could theoretically miss  the change notify callback if that `int` is still in the CPU cache. I'll put an atomic.

shell_integration/windows/OCUtil/RemotePathChecker.cpp
jturcotte: Done

shell_integration/windows/OCUtil/RemotePathChecker.cpp
jturcotte: After realizing that multiple explorer windows will call `IsMemberOf` concurrently, I've put the vector in a shared_ptr and each call will keep a ref to that copy on the stack (@ogoffart's idea of course).

shell_integration/windows/OCOverlays/OCOverlay.cpp
jturcotte: Reverted that change and put a brace above, let's have that discussion officially before we start doing it.

shell_integration/windows/OCUtil/RemotePathChecker.cpp
```diff
@@ -120,16 +120,23 @@ void RemotePathChecker::workerThreadLoop()
...
-                    if (wasAsked || wasCached) {
-                        auto &it = _cache[responsePath];
-                        changed = (it != state);
-                        it = state;
+                    auto it = _cache.find(responsePath);
+                    if (it == _cache.end()) {
+                        // The client only approximates requested files, if the bloom
+                        // filter becomes saturated after navigating multiple directories we'll start getting
+                        // status pushes that we never requested and fill our cache. Ignore those.
+                        if (!wasAsked) {
```
jturcotte: Updated to bring back wasAsked since the bloom filter will probably get saturated after a while if the user has lots of directories and navigate them over time. We're still keeping the cache forever in each process and should be a bit careful.

shell_integration/windows/OCUtil/RemotePathChecker.cpp
ogoffart: Without std::atomic, this is a undefined behavior according to the C++ standard.

shell_integration/windows/OCUtil/RemotePathChecker.cpp
```diff
@@ -182,9 +187,8 @@ RemotePathChecker::~RemotePathChecker()
     CloseHandle(_newQueries);
 }
 
-vector<wstring> RemotePathChecker::WatchedDirectories()
+std::shared_ptr<const std::vector<std::wstring>> RemotePathChecker::WatchedDirectories() const
 {
-    std::unique_lock<std::mutex> lock(_mutex);
```
ogoffart: Still need the lock!

The refcount is atomic, meaning that different instance of the shared_ptr pointing to the same object can be used in different thread.  But the same shared_ptr instance cannot, it is not thread safe.

jturcotte: Done, used `std::atomic_store` and `std::atomic_load`.

PR #5450, opened 2017-01-13
Updater fixes, disable RTTI

guruz:


src/gui/updater/updater.cpp
```diff
@@ -87,6 +87,10 @@ Updater *Updater::create()
     if (updateBaseUrl.isEmpty()) {
         updateBaseUrl = QUrl(QLatin1String(APPLICATION_UPDATE_URL));
     }
+    if (!updateBaseUrl.isValid() || updateBaseUrl.host() == ".") {
+        qDebug() << "Not a valid updater URL, will not no update check";
```
phil-davis: "will not do..."

ckamm:
@guruz How certain are you that disabling rtti doesn't introduce silent issues? Is it worthwhile to do this for 24 kb?

Also, the chunking ng test doesn't seem to compile right now.

guruz:
Thanks @phil-davis @ckamm  , typo and test compile fixed!

PR #5459, opened 2017-01-13
DB: Also migrate -shm and -wal to new name #3764 #5045

guruz:
FYI @SamuAlfageme

### DIFF
--- a/src/libsync/syncjournaldb.cpp
+++ b/src/libsync/syncjournaldb.cpp
@@ -64,8 +64,12 @@ bool SyncJournalDb::maybeMigrateDb(const QString& localPath, const QString& abso
     if( !FileSystem::fileExists(oldDbName) ) {
         return true;
     }
+    const QString oldDbNameShm = oldDbName + "-shm";
+    const QString oldDbNameWal = oldDbName + "-wal";
 
     const QString newDbName = absoluteJournalPath;
+    const QString newDbNameShm = newDbName + "-shm";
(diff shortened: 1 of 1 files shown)


src/libsync/syncjournaldb.cpp
```diff
@@ -80,12 +84,36 @@ bool SyncJournalDb::maybeMigrateDb(const QString& localPath, const QString& abso
             return false;
         }
     }
+    if( FileSystem::fileExists( newDbNameWal ) ) {
```
SamuAlfageme: Q: could these methods be factored out and extracted to avoid repetition?

SamuAlfageme:
Lookin' neat 👍

src/libsync/syncjournaldb.cpp
```diff
@@ -80,12 +84,36 @@ bool SyncJournalDb::maybeMigrateDb(const QString& localPath, const QString& abso
...
+        if( !FileSystem::remove(newDbNameWal, &error) ) {
+            qDebug() << "Database migration: Could not remove db WAL file" << newDbNameWal
+                     << "due to" << error;
+            return false;
```
ogoffart: i think we could ignore this error and not return false.  (everywhere on these new checks)

Because this is not a fatal error, is it?

guruz: It's not fatal but would keep a stale file there. I don't know, i don't have a further opinion

Issue #4797, opened 2016-05-06
[Overlays] Don't rely on hierarchical SyncFileItems to propagate OK statuses to parent folders during sync


jturcotte:
At the moment we rely on the fact that the SyncEngine will emit itemCompleted for all child tasks before completing parent directory tasks to be able to change those parent status from SYNC to OK.

This has a few problems:
- If the parent folder needs to be created as well, it will have another itemCompleted signal of its own, meaning that the new folder will be in OK state even though its children are still propagated.
- If another child of a parent has an ERROR, the parent will switched to WARN (IGNORE) even though its children aren't done propagating
- For RETRIEVE_FILE_STATUS messages, especially problematic in bulk following an UPDATE_VIEW message, we have to do a O(n) search of the file in the SyncFileItem vector to know if the file is currently syncing and this can slow down the sync for very large initial syncs.

A solution would be to ignore CSYNC_INSTRUCTION_NONE items in SyncFileStatusTracker, and do the child sync tracking directly there with a more adapted data structure to speed up the lookup.

jturcotte:
Idea for an implementation before I forget it:
- Keep a "sync count" for each directory path relative to the folder root in a QHashMap. A file being discovered as needing propagation increases the count of its parent, on completion it reduces the count
- Don't increase the count for all parents to avoid large directory structures being too heavy to process. Instead, increase the count of a directory's immediate parent directory only if the directory's count was 0 (first increase), possibly until the root. And when we decrease from 1 to 0 (last decrease), also reduce the immediate parent, and so on.
- When a directory goes from 0 to 1, send a SYNC status. When it goes from 1 to 0, send an OK status (considering errors that should bubble up, etc).
- A directory that needs propagation for itself (mkdir) increases/decreases its own count following the same rules.
- We can then ignore all INSTRUCTION_NONE items that we receive in SyncFileStatusTracker

SamuAlfageme:
@jturcotte the solution seems to be working after the fix. However, there's some minor delays/glitches in the propagation, see these gifs as example:

* Parents take a bit to change the status in a sync:

![propagation](https://cloud.githubusercontent.com/assets/2644445/20176983/b0da163e-a74a-11e6-88cf-a8add8ff228e.gif)

* A parent appears synced before its children: 

![propagation2](https://cloud.githubusercontent.com/assets/2644445/20177020/e004cfc6-a74a-11e6-98c1-8b202f80616a.gif)

[...]

PR #5467, opened 2017-01-17
StatusTracker: Emit OK for the last child before parents

jturcotte:
The current logic tried to avoid a DB lookup just to fetch whether
the file is shared or not since that info is already in the
SyncFileItem. The implementation would however need to decrease the
sync count for itself (and parents) before emitting the new status,
thus emitting the OK status for parents before that last child that
ended the propagation for that folder.

Change the implementation to achieve what we want: give the
possibility to decSyncCount to use a pre-fetched sharing state while
still doing the emission for all involved files. This ensures that
the leaf file also gets its status emitted before its parents.

Issue #4797

### DIFF
--- a/src/libsync/syncfilestatustracker.cpp
+++ b/src/libsync/syncfilestatustracker.cpp
@@ -129,42 +129,46 @@ void SyncFileStatusTracker::slotPathTouched(const QString& fileName)
     emit fileStatusChanged(fileName, SyncFileStatus::StatusSync);
 }
 
-void SyncFileStatusTracker::incSyncCount(const QString &relativePath, EmitStatusChangeFlag emitStatusChange)
+void SyncFileStatusTracker::incSyncCountAndEmitStatusChanged(const QString &relativePath, SharedFlag sharedFlag)
 {
     // Will return 0 (and increase to 1) if the path wasn't in the map yet
     int count = _syncCount[relativePath]++;
     if (!count) {
-        if (emitStatusChange)
-            emit fileStatusChanged(getSystemDestination(relativePath), fileStatus(relativePath));
+        SyncFileStatus status = sharedFlag == UnknownShared
+            ? fileStatus(relativePath)
+            : resolveSyncAndErrorStatus(relativePath, sharedFlag);
+        emit fileStatusChanged(getSystemDestination(relativePath), status);
 
         // We passed from OK to SYNC, increment the parent to keep it marked as
         // SYNC while we propagate ourselves and our own children.
         Q_ASSERT(!relativePath.endsWith('/'));
         int lastSlashIndex = relativePath.lastIndexOf('/');
         if (lastSlashIndex != -1)
-            incSyncCount(relativePath.left(lastSlashIndex), EmitStatusChange);
+            incSyncCountAndEmitStatusChanged(relativePath.left(lastSlashIndex), UnknownShared);
         else if (!relativePath.isEmpty())
-            incSyncCount(QString(), EmitStatusChange);
+            incSyncCountAndEmitStatusChanged(QString(), UnknownShared);
     }
 }
 
-void SyncFileStatusTracker::decSyncCount(const QString &relativePath, EmitStatusChangeFlag emitStatusChange)
+void SyncFileStatusTracker::decSyncCountAndEmitStatusChanged(const QString &relativePath, SharedFlag sharedFlag)
 {
     int count = --_syncCount[relativePath];
     if (!count) {
         // Remove from the map, same as 0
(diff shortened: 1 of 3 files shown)


src/libsync/syncfilestatustracker.h
```diff
@@ -51,15 +51,14 @@ private slots:
...
     enum PathKnownFlag { PathUnknown = 0, PathKnown };
-    enum EmitStatusChangeFlag { DontEmitStatusChange = 0, EmitStatusChange };
-    SyncFileStatus resolveSyncAndErrorStatus(const QString &relativePath, SharedFlag isShared, PathKnownFlag isPathKnown = PathKnown);
+    SyncFileStatus resolveSyncAndErrorStatus(const QString &relativePath, SharedFlag sharedState, PathKnownFlag isPathKnown = PathKnown);
 
     void invalidateParentPaths(const QString& path);
     QString getSystemDestination(const QString& relativePath);
-    void incSyncCount(const QString &relativePath, EmitStatusChangeFlag emitStatusChange);
-    void decSyncCount(const QString &relativePath, EmitStatusChangeFlag emitStatusChange);
+    void incSyncCountAndEmitStatusChanged(const QString &relativePath, SharedFlag sharedState);
```
guruz: Please add a detailed comment about what those two functions are doing.

jturcotte: Humm, those functions increase the sync count and emit the statusChanged signals, I think the name is already pretty self-explanatory.

The implementation also has detailing comments.

guruz: I don't know what a "sync count" is.

guruz: (e.g. add comment above QHash<QString, int> _syncCount; too..)

jturcotte: Added a comment, the purpose of _syncCount should be clearer.
Let me know if it's still not clear.

I don't want to add to much comments in the header since those methods are private and explanation should rather go at the single call site or in those function implementations.

src/libsync/syncfilestatustracker.cpp
```diff
@@ -129,42 +129,46 @@ void SyncFileStatusTracker::slotPathTouched(const QString& fileName)
...
         // We passed from OK to SYNC, increment the parent to keep it marked as
         // SYNC while we propagate ourselves and our own children.
         Q_ASSERT(!relativePath.endsWith('/'));
         int lastSlashIndex = relativePath.lastIndexOf('/');
         if (lastSlashIndex != -1)
-            incSyncCount(relativePath.left(lastSlashIndex), EmitStatusChange);
+            incSyncCountAndEmitStatusChanged(relativePath.left(lastSlashIndex), UnknownShared);
         else if (!relativePath.isEmpty())
-            incSyncCount(QString(), EmitStatusChange);
+            incSyncCountAndEmitStatusChanged(QString(), UnknownShared);
```
ckamm: Not related to your patch: This confused me while reading. Why does this branch exist? I would guess that the code assumes all paths either start with a / - then this is unnecessary - or don't, in which case this could be an assert.

src/libsync/syncfilestatustracker.cpp
```diff
@@ -177,6 +181,7 @@ void SyncFileStatusTracker::slotAboutToPropagate(SyncFileItemVector& items)
 
     foreach (const SyncFileItemPtr &item, items) {
         // qDebug() << Q_FUNC_INFO << "Investigating" << item->destination() << item->_status << item->_instruction;
+        _dirtyPaths.remove(item->destination());
```
ckamm: I don't understand this yet.

src/libsync/syncfilestatustracker.cpp
jturcotte: Humm, I think because `relativePath.left(-1)` wouldn't work. Does that answer your question?

src/libsync/syncfilestatustracker.cpp
jturcotte: It's related to the `oldDirtyPaths` logic lower, we need to know what file detected by the filesystemwatcher will actually be propagated (vs. e.g. checksum checks by csync).

src/libsync/syncfilestatustracker.cpp
jturcotte: It's all relative paths (so none of them starts with a slash) and we check whether there is a slash or not to know if:

1. It's the folder root (relativePath empty, no parent to update)
1. It's a child of the root (needs to recurse using "" as relativePath, the root)
1. Something lower (in which case, just remove the last path component and recurse)

PR #5479, opened 2017-01-19
Fixed hungarian translation

csgabor:
Log out was translated as quit which can confuse users.


ogoffart:
Hi, thanks for your contribution.

Translation fixes should be done on transifex: https://www.transifex.com/owncloud-org/owncloud/

PR #5480, opened 2017-01-20
Chunking ng cleanup

ogoffart:
Will help for https://github.com/owncloud/core/issues/26981

### DIFF
--- a/src/libsync/propagateuploadng.cpp
+++ b/src/libsync/propagateuploadng.cpp
@@ -97,6 +97,12 @@ void PropagateUploadFileNG::doStartUpload()
                 this, SLOT(slotPropfindIterate(QString,QMap<QString,QString>)));
         job->start();
         return;
+    } else if (progressInfo._valid) {
+        // The upload info is stale. remove the stale chunks on the server
+        _transferId = progressInfo._transferid;
+        // Fire and forget. Any error will be ignored.
+        (new DeleteJob(propagator()->account(), chunkUrl(), this))->start();
+        // startNewUpload will reset the _transferId and the UploadInfo in the db.
     }
 
     startNewUpload();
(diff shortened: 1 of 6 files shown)


guruz:
Is this for 2.3?

guruz:
is this 2.3 material?

src/libsync/propagateuploadng.cpp
```diff
@@ -97,6 +97,12 @@ void PropagateUploadFileNG::doStartUpload()
                 this, SLOT(slotPropfindIterate(QString,QMap<QString,QString>)));
         job->start();
         return;
+    } else if (progressInfo._valid) {
+        // The upload info is stale. remove the stale chunks on the server
+        _transferId = progressInfo._transferid;
+        // Fire and forget. Any error will be ignored.
+        (new DeleteJob(propagator()->account(), chunkUrl(), this))->start();
```
jturcotte: It would be a bit cleaner and safer if `chunkUrl` would take the transferId as parameter (and could be made static) instead of having to set the variable and reset it later.

src/libsync/syncengine.cpp
```diff
@@ -301,7 +302,16 @@ void SyncEngine::deleteStaleUploadInfos()
...
 
     // Delete from journal.
-    _journal->deleteStaleUploadInfos(upload_file_paths);
+    auto ids = _journal->deleteStaleUploadInfos(upload_file_paths);
+
+    // Delete the stales chunk on the server.
+    if (account()->capabilities().chunkingNg()) {
+        foreach (uint transferId, ids) {
+            QUrl url = Utility::concatUrlPath(account()->url(), QLatin1String("remote.php/dav/uploads/")
+                + account()->davUser() + QLatin1Char('/') + QString::number(transferId));
```
jturcotte: Could this use the same `chunkUrl` implementation?

Issue #5237, opened 2016-10-10
Windows and Linux 32-bit clients crashing on large number of files


montanaviking:
### Expected behaviour

Tell us what should happen
Should sync any number of files up to physical memory/disk space of computers
### Actual behaviour

Tell us what happens instead
Attempts to sync more than approximately 300,000 to 400,000 files fails with Windows and 32-bit Linux sync clients. The Windows sync client appears to crash when its RAM requirement exceeds approximately 1.5GB and while it's checking for things to sync - with an error from Visual C++. The 32-bit Ubuntu Linux client also crashes under similar conditions. 
However, the 64-bit Linux sync client is handling > 900,000 files - about 50GB of data
Thanks
### Steps to reproduce
1. Increase number of files to sync to >900,000
2. 
3. 
### Server configuration

Operating system: Ubuntu 14.04 64-bit
3.13.0-96-generic #143-Ubuntu SMP Mon Aug 29 20:15:20 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux

Web server: Apache/2.4.7 (Ubuntu)

Database:mysql  Ver 14.14 Distrib 5.5.52, for debian-linux-gnu (x86_64) using readline 6.3

PHP version: PHP 5.5.9-1ubuntu4.20 (cli) (built: Oct  3 2016 13:00:37)

ownCloud version: 9.1.1

Storage backend (external storage): None
### Client configuration

Client version: 2.2.4 build 6408

Operating system: Windows 7

OS language: English

Installation path of client: C:\Program Files (x86)\ownCloud
### Logs

Please use Gist (https://gist.github.com/) or a similar code paster for longer
logs.

`Template for output < 10 lines`
1. Client logfile: Output of `owncloud --logwindow` or `owncloud --logfile log.txt`
   (On Windows using `cmd.exe`, you might need to first `cd` into the ownCloud directory)
   (See also http://doc.owncloud.org/desktop/2.2/troubleshooting.html#client-logfile )
2. Web server error log:
3. Server logfile: ownCloud log (data/owncloud.log):

guruz:
(Also posted to central on https://central.owncloud.org/t/owncloud-client-crashes-windows/3081/1 )

FYI @jturcotte

michaelstingl:
@guruz What are the plans for building 64-bit for Windows?

00006335

montanaviking:
Hi Michael,
[...]

PR #5488, opened 2017-01-25
Delete SyncFileItems after propagation

jturcotte:
This improves the number of files that can be synced in one shot.

Issue #5237
This depends on #5400 to work properly

### DIFF
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -692,10 +692,9 @@ bool Folder::setIgnoredFiles()
     // a QSet of files to load.
     ConfigFile cfg;
     QString systemList = cfg.excludeFile(ConfigFile::SystemScope);
-    if( QFile::exists(systemList) ) {
-        qDebug() << "==== adding system ignore list to csync:" << systemList;
-        _engine->excludedFiles().addExcludeFilePath(systemList);
-    }
+    qDebug() << "==== adding system ignore list to csync:" << systemList;
+    _engine->excludedFiles().addExcludeFilePath(systemList);
+
     QString userList = cfg.excludeFile(ConfigFile::UserScope);
     if( QFile::exists(userList) ) {
         qDebug() << "==== adding user defined ignore list to csync:" << userList;
@@ -739,6 +738,8 @@ void Folder::startSync(const QStringList &pathList)
     qDebug() << "*** Start syncing " << remoteUrl().toString() << " - client version"
              << qPrintable(Theme::instance()->version());
 
+    _fileLog->start(path());
+
     if (!setIgnoredFiles())
     {
         slotSyncError(tr("Could not read system exclude file"));
@@ -754,8 +755,6 @@ void Folder::startSync(const QStringList &pathList)
     _engine->setNewBigFolderSizeLimit(limit);
 
     _engine->setIgnoreHiddenFiles(_definition.ignoreHiddenFiles);
-
-    _fileLog->start(path());
 
     QMetaObject::invokeMethod(_engine.data(), "startSync", Qt::QueuedConnection);
 
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -52,7 +52,6 @@ Folder::Folder(const FolderDefinition& definition,
     : QObject(parent)
       , _accountState(accountState)
       , _definition(definition)
-      , _csyncError(false)
       , _csyncUnavail(false)
       , _wipeDb(false)
       , _proxyDirty(true)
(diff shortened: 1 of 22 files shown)


jturcotte:
Done, diff on last commit after force-push:

``` diff
diff --git a/src/gui/folder.cpp b/src/gui/folder.cpp
index 096eed568..6aa0fe2fb 100644
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -709,7 +709,7 @@ void Folder::slotSyncFinished(bool success)
 
     bool syncError = !_syncResult.errorStrings().isEmpty();
     if( syncError ) {
-        qDebug() << "-> SyncEngine finished with ERROR, warn count is" << _syncResult.warnCount();
+        qDebug() << "-> SyncEngine finished with ERROR";
     } else {
         qDebug() << "-> SyncEngine finished without problem.";
     }
@@ -724,8 +724,7 @@ void Folder::slotSyncFinished(bool success)
     } else if (_csyncUnavail) {
         _syncResult.setStatus(SyncResult::Error);
         qDebug() << "  ** csync not available.";
-    } else if( _syncResult.warnCount() > 0 ) {
-        // there have been warnings on the way.
+    } else if( _syncResult.foundFilesNotSynced() ) {
         _syncResult.setStatus(SyncResult::Problem);
     } else if( _definition.paused ) {
         // Maybe the sync was terminated because the user paused the folder
diff --git a/src/libsync/syncresult.cpp b/src/libsync/syncresult.cpp
index 558c03fc5..625ebae2c 100644
--- a/src/libsync/syncresult.cpp
+++ b/src/libsync/syncresult.cpp
@@ -20,7 +20,7 @@ namespace OCC
 
 SyncResult::SyncResult()
     : _status( Undefined )
-    , _warnCount(0)
+    , _foundFilesNotSynced(false)
     , _folderStructureWasChanged(false)
     , _numNewItems(0)
     , _numRemovedItems(0)
@@ -93,16 +93,6 @@ QDateTime SyncResult::syncTime() const
     return _syncTime;
 }
 
-int SyncResult::warnCount() const
-{
-    return _warnCount;
-}
-
-void SyncResult::appendErrorStrings( const QStringList& list )
-{
-    _errors.append(list);
-}
-
 QStringList SyncResult::errorStrings() const
 {
     return _errors;
@@ -136,10 +126,9 @@ QString SyncResult::folder() const
 
 void SyncResult::processCompletedItem(const SyncFileItemPtr &item)
 {
-    // FIXME: Why are error counted as warnings? Test.
     if (Progress::isWarningKind(item->_status)) {
-        // Count all error conditions.
-        _warnCount++;
+        // Count any error conditions, error strings will have priority anyway.
+        _foundFilesNotSynced = true;
     }
 
     if (item->_isDirectory && (item->_instruction == CSYNC_INSTRUCTION_NEW
@@ -191,9 +180,9 @@ void SyncResult::processCompletedItem(const SyncFileItemPtr &item)
                 // nothing.
                 break;
             }
-        } else if( item->_direction == SyncFileItem::None ) { // ignored files counting.
+        } else if( item->_direction == SyncFileItem::None ) {
             if( item->_instruction == CSYNC_INSTRUCTION_IGNORE ) {
-                _warnCount++;
+                _foundFilesNotSynced = true;
             }
         }
     }
diff --git a/src/libsync/syncresult.h b/src/libsync/syncresult.h
index 7d694c141..0c630f737 100644
--- a/src/libsync/syncresult.h
+++ b/src/libsync/syncresult.h
@@ -50,10 +50,8 @@ public:
     void reset();
 
     void    appendErrorString( const QString& );
-    void    appendErrorStrings( const QStringList& );
     QString errorString() const;
     QStringList errorStrings() const;
-    int     warnCount() const;
     void    clearErrors();
 
     void setStatus( Status );
@@ -63,6 +61,7 @@ public:
     void setFolder(const QString& folder);
     QString folder() const;
 
+    bool foundFilesNotSynced() const { return _foundFilesNotSynced; }
     bool folderStructureWasChanged() const { return _folderStructureWasChanged; }
 
     int numNewItems() const { return _numNewItems; }
@@ -90,7 +89,7 @@ private:
      * when the sync tool support this...
      */
     QStringList        _errors;
-    int                _warnCount;
+    bool _foundFilesNotSynced;
     bool _folderStructureWasChanged;
 
     // count new, removed and updated items
```

dragotin:
you wanna really merge that in 2.3?

mrow4a:
@jturcotte About deleting jobs, it needs to be carefully done and checked on some latency with real upload timings. 

Reason for that is simple, remove() and delete operations are blocking, so it means it increases the idle gap between finishing the file and scheduling the next one - this transition should be blasting fast, since this moment is performance critical. Even 50ms on one files, causes 1000 to be in 6 connections -> 167 lines of blocking, which could be seconds lost!

jturcotte:
@dragotin Only the last commit is meaty and the risk is pretty low (mostly bubbleUpResults has its logic changed). It's all mostly only about making sure that we don't keep SyncFileItemPtrs around except in the job. The merge request looks scary mainly because of all the harmless `SyncFileItem&` -> `SyncFileItemPtr&` changes.

@mrow4a Those operation are meaningless vs. the disk and sqlite operations. Updating a vector of 10k elements is nothing compared to even a single disk access. I passed this in the profiler and I can't even see a sample spent in the job deletion code.

mrow4a:
@jturcotte On localhost they nearly come in order, so you remove from begining of the list, try to have them out of order. Just saying that this was not working good with HTTP2

jturcotte:
@mrow4a Have a look at this and note that accessing a byte on disk is going to be 10000 times faster than accessing a vector in memory:
https://gist.github.com/jboner/2841832

The first access of each cache line will then load following bytes in the CPU cache, meaning that it will be even faster. Don't worry about this, even if that vector has 10000 subjobs, it will still be *way* faster than accessing your disk (and we do lots of it). We can give you an introduction to perf next time you pass by the office, then you can start thinking about more than loops.

ogoffart:
+1

Issue #5469, opened 2017-01-17
Shibboleth accounts don't need the credentials to log in after wiping its keychain entry


SamuAlfageme:
Found while testing #5408

### Steps to reproduce:
1. Login with a valid shibb account and sync your files
2. Remove keychain entry for the account 
3. Restart the client, it asks for shibb credentials
4. Input **a different but valid account**
5. An error is displayed (see the one for the http://testshib.org down). Close it.
6. Click `Log in` in the account view
7. Account is logged back without entering the credentials again

![error_shib](https://cloud.githubusercontent.com/assets/2644445/21925325/44630654-d97d-11e6-8cf6-8022e4cc79ac.png)

I assume wiping the keychain entry is not enough to force log out and the client is using some session token to login back again, but then, **why it asks for the credentials, if they're not needed?**

--- 

**Edit: Error in step 5 is indeed only displayed when both accounts were previously set and you input the credentials for the wrong one**, otherwise you get the usual:

![screenshot 2017-01-13 11 02 52](https://cloud.githubusercontent.com/assets/2644445/21926144/c8e99980-d980-11e6-90a2-4c706bb96ed2.png)

**Made a screencast explaining this particular issue: https://webmshare.com/play/dEwbv**

--- 

Also, I'm only seeing one entry in the keychain for all the shibboleth accounts. When a shibb account is removed, is this entry updated stripping the right `_shibsession_`?

guruz:
Which version?

SamuAlfageme:
@guruz `Version 2.3.0-nightly20170116 (build 3970)`

ckamm:
Putting into 2.3 since it's a surprising auth-related issue.

ckamm:
Can reproduce:
1. For the same shib server, configure two account.
2. Log out of one, log in, enter credentials for the *other, wrong* one.
3. Close the error that pops up.
4. Click "Log in" again. Now the account is logged in with the wrong credentials, syncing to the wrong account.

This potentially messes up user data.

ckamm:
Pull request: https://github.com/owncloud/client/pull/5486

SamuAlfageme:
@ckamm yes, that was my final concern in the original issue:

> Also, **I'm only seeing one entry in the keychain for all the shibboleth accounts.** When a shibb account is removed, is this entry updated stripping the right _shibsession_?

SamuAlfageme:
Looking good 😎, have to run some more tests but the whole behavior seems fixed.

[...]

PR #5490, opened 2017-01-25
Cookies: Use different DB for different accounts

guruz:
This is a follow up to #5469

### DIFF
--- a/src/libsync/account.h
+++ b/src/libsync/account.h
@@ -172,6 +172,7 @@ public:
 
     void clearCookieJar();
     void lendCookieJarTo(QNetworkAccessManager *guest);
+    QString cookieJarPath();
 
     void resetNetworkAccessManager();
     QNetworkAccessManager* networkAccessManager();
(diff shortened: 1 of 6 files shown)


src/libsync/account.h
```diff
@@ -172,6 +172,7 @@ class OWNCLOUDSYNC_EXPORT Account : public QObject {
 
     void clearCookieJar();
     void lendCookieJarTo(QNetworkAccessManager *guest);
+    QString cookieJarPath();
```
ogoffart: Please make it private.

SamuAlfageme:
Minor detail: there's no migration strategy for `cookies.db` to be split into different `cookiesN.db`: the old cookies file is not removed nor its contents distributed.

src/libsync/account.h
guruz: It's accessed from 2 different files/classes which are not the account, can't do it without friending both.

guruz:
> Minor detail: there's no migration strategy for cookies.db to be split into different cookiesN.db: the old cookies file is not removed nor its contents distributed.

Hm, I just assumed it's fine because the behaviour should be similar to as if the cookies had been forgotten/expired.

PR #5495, opened 2017-01-26
Memory optimisations

jturcotte:


src/gui/syncrunfilelog.cpp
```diff
@@ -145,7 +145,6 @@ void SyncRunFileLog::logItem( const SyncFileItem& item )
 
     const QChar L = QLatin1Char('|');
     _out << ts << L;
-    _out << QString::number(item._requestDuration) << L;
```
jturcotte: **Note:** This changes the format of the log, if anybody was parsing it assuming the position of fields, this might break it. I'm assuming that we don't support a stable format for this file.

dragotin: you could keep two pipe chars, in case a script splits by '|' that still works, only the duration is empty.

jturcotte: Sounds good, did that.

Issue #5494, opened 2017-01-26
Improve message when deleting all files in the local sync folder conection (aboutToRemoveAllFiles)


SamuAlfageme:
- [x] Give more details about why this message occurs, something like: 

> You are trying to remove all the files on your local sync folder `<folder>`. <br> These changes will be synchronized with your server, making your files unavailable unless restored. <br> Are you sure you want to remove all the files? 

- [x] [Windows] Make the dialog have focus over other windows, so it doesn't go minimized without picking a resolution strategy, etc.

SamuAlfageme:
@CaptionF PR in https://github.com/owncloud/client/pull/5503 check this GIF: 

![on_top_dialog](https://cloud.githubusercontent.com/assets/2644445/22422752/61d46ba0-e6ee-11e6-8509-cadfc03b6089.gif)

michaelstingl:
Sticky! 👍

guruz:
Can we close this issue and keep the PR https://github.com/owncloud/client/pull/5503 in 2.3?
And someone (@ogoffart ?) adjusts the msg in the PR to what holger mentioned in 5503?

SamuAlfageme:
@guruz sure!

Issue #5485, opened 2017-01-23
If all the files are deleted in the server, keep files is duplicating the contents in the server


SamuAlfageme:
### Actual behavior
The client is prompted with the usual warning dialog regarding all the files were deleted: 

<p align="center">
  <img src="https://cloud.githubusercontent.com/assets/2644445/22206981/98f7e65a-e17d-11e6-8e8b-dee57e42fb8b.png"/>
</p>

If `Keep files` is selected as resolution strategy, the files are restored in the server but they still appear in the `Deleted files` section (piling up every time this is carried out).
![screenshot 2017-01-23 15 05 45](https://cloud.githubusercontent.com/assets/2644445/22206978/9593a7ce-e17d-11e6-81d3-14716aec00c7.png)

### Expected behavior
One of the following options:
1. The server actions are seen as master -> the client should not prompt the warning dialog 
2. `Keep files` in the dialog should be seen as multiple `Restore` operations on the server

### Steps to reproduce
1. Select all the files in the web UI
2. Delete them

hodyroff:
Option one under expected is absolutely wrong and a no-go. We implemented this feature for the very reason ...
Restore does IMHO not yet have an API, so what we are currently doing is uploading the files to the server again - therefor they stay in the trashbin. Works as expected.
We expect a user to know what they are doing ... this was build in particullary to protect from storage connection errors which is different then actually really deleting things.
This doesn't sound like a real actual challenge and is not a bug as it works as designed.

ogoffart:
I'm closing this issue because I think the client is behaving properly.
The client is not supposed to know what is in the trash.

If anything, the server could find out that the files are duplicated based on the checksum.

PR #5503, opened 2017-01-30
Local massive deletion message improvements

SamuAlfageme:
As discussed in #5494 and https://github.com/owncloud/enterprise/issues/1780 this changes the message and makes the "Remove All Files" dialog to be on top of other windows to avoid it going unnoticed or lost without picking a strategy:

on_top_dialog

### DIFF
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -883,17 +883,22 @@ void Folder::setSaveBackwardsCompatible(bool save)
     _saveBackwardsCompatible = save;
 }
 
-void Folder::slotAboutToRemoveAllFiles(SyncFileItem::Direction, bool *cancel)
+void Folder::slotAboutToRemoveAllFiles(SyncFileItem::Direction dir, bool *cancel)
 {
     ConfigFile cfgFile;
     if (!cfgFile.promptDeleteFiles())
         return;
 
-    QString msg =
-        tr("This sync would remove all the files in the sync folder '%1'.\n"
-           "This might be because the folder was silently reconfigured, or that all "
-           "the files were manually removed.\n"
-           "Are you sure you want to perform this operation?");
+    QString msg = dir == SyncFileItem::Down ?
+        tr("All files in the sync folder '%1' folder were deleted on the server.\n"
+            "These deletes will be synchronized to your local sync folder, making such files "
+            "unavailable unless you have a right to restore. \n"
+            "If you decide to keep the files, they will be re-synced with the server if you have rights to do so.\n"
+            "If you decide to delete the files, they will be unavailable to you, unless you are the owner.") :
+        tr("All the files in your local sync folder '%1' were deleted. These deletes will be "
+            "synchronized with your server, making such files unavailable unless restored.\n"
+            "Are you sure you want to sync those actions with the server?\n"
+            "If this was an accident and you decide to keep your files, they will be re-synced from the server.");
     QMessageBox msgBox(QMessageBox::Warning, tr("Remove All Files?"),
                        msg.arg(shortGuiLocalPath()));
     msgBox.addButton(tr("Remove all files"), QMessageBox::DestructiveRole);


ogoffart:
This message is also displayed when all the files are removed on the server. That is, i think the most likely sceniario and the reason behind this message. I'm afraid the new wording is even more confusing in this case then.
You can use the direction argument of this unction to know where all files were removed.

As for the StayOnTop, I am not sure it is a good idea. Normal dialogs don't do that. It may bother the user.  But if you think it helps more than it hurts, why not...

phil-davis:
I am assuming that if the user clicks "Keep Files" then a sync is started that will send the files from wherever they remain to the other side (i.e. files deleted on server but remain in local sync folder, then a "full copy" is initiated from local sync folder to server, files deleted on local sync folder but remain on server then a "full copy" is initiated from server to local sync folder).
Here are some suggested words for each direction:
```
    QString msg =
        tr("All the files on the server have been removed.\n"
           "These removals will be synchronized to your local sync folder '%1', making your files "
           "unavailable unless restored. \n"
           "Are you sure you want to remove all the files from your local sync folder?");
    QString msg =
        tr("All the files in your local sync folder '%1' have been removed.\n"
           "These removals will be synchronized with your server, making your files "
           "unavailable unless restored. \n"
           "Are you sure you want to remove all the files from the server?");
```
Personally, I would prefer to use the verb "delete" rather than "remove". I think that "remove" sounds a bit temporary and recoverable - you can remove things from a room and then put them back! And in this case, if the server has no large enough deleted files trashbin functioning, and the user has no local backup of the files... then it may be a huge nuisance for the ownCloud admin to recover their files.

SamuAlfageme:
I'd discuss about the dialog staying on top since it requires a fast input on the user over some "important" change made and it's not something you'd have minimized or closed without reading and it also gives you the only two choices available. 

As for what @phil-davis says, I agree ideally we should distinguish the source of the deletion (because it'd also help with some other issues related with this behavior like https://github.com/owncloud/client/issues/5485), but as the current implementation simplifies this, we can change the message to be something more suited and open a separate issue for that (?)

michaelstingl:
> I agree ideally we should distinguish the source of the deletion

👍

hodyroff:
Yes, we need to be precise.
- Agree that "delete" is a better wording then "remove". Normally actions are read, write, delete, move, etc. when we do a sync from the server again its a re-sync, or?
- Hate dominant windows from a UX perspective. In the case of something done on the desktop, give immediate feedback this seems a good choice. In the other case something happened on the server I would suggest the more seamless behaviour. So either split up or focus on error message only, but can be convinced otherwise.
- We need to think about headless also, people are running the client with NSSM and the deletion in such a usecase might be standard! Or is this taken into account (config option?)

hodyroff:
Messages, if 2 are possible pending on the direction, they might yet be perfect, feedback of course welcome!

1) server-side deletion

All files in a folder were deleted on the Server by %2 (user name).\n"
These deletes will be synchronized to your local sync folder '%1', making such files "
unavailable unless you have a right to restore. \n"
If you decide to keep the files, they will be re-synced with the server if you have rights to do so.
If you decide to delete the files, they will be unavailable to you, unless you are the owner.
");
(Keep Files; Delete Files)

2) local deletion

All the files in your local sync folder '%1' were deleted. These deletes will be synchronized with your server, making such files "
unavailable unless restored. \n"
Are you sure you want to sync those actions with the server?");
If this was an accident and you decide to keep your files, they will be re-synced from the Server.
(Keep Files; Delete Files)

ogoffart:
I updated the message to (almost) the wording from @hodyroff .  The main difference is that it does not say which user removed the files. (We don't have this information. And usually this happens because the account was reset to default by the admin)

>We need to think about headless also, people are running the client with NSSM and the deletion in such a usecase might be standard! Or is this taken into account (config option?)

There is an hidden config option for that already.

ogoffart:
Merged in 2.3 (439e6889063b57bf4cf2efab70e7b6251de6bae4)

src/gui/folder.cpp
```diff
@@ -883,17 +883,22 @@ void Folder::setSaveBackwardsCompatible(bool save)
...
+    QString msg = dir == SyncFileItem::Down ?
+        tr("All files in the sync folder '%1' folder were deleted on the server.\n"
+            "These deletes will be synchronized to your local sync folder, making such files "
+            "unavailable unless you have a right to restore. \n"
+            "If you decide to keep the files, they will be re-synced with the server if you have rights to do so.\n"
+            "If you decide to delete the files, they will be unavailable to you, unless you are the owner.") :
+        tr("All the files in your local sync folder '%1' were deleted. These deletes will be "
+            "synchronized with your server, making such files unavailable unless restored.\n"
+            "Are you sure you want to sync those actions with the server?\n"
+            "If this was an accident and you decide to keep your files, they will be re-synced from the server.");
```
dragotin: Just for the record, I think that construct is not nice. It is hard to read. How about setting msg to the `Down`-text, check in a simple if statement if dir is `Up`, and maybe have an assert in the else-branch in case dir can be different than Up and Down later?

And if you still do not want to change that, don't you want to put a `const` in front of the QString msg?

SamuAlfageme:
What about the sticky message? Not focus anymore? See https://github.com/owncloud/enterprise/issues/1780

Issue #2558, opened 2014-11-28
Qt5 everywhere. Drop legacy Qt4 support.


dragotin:
It would be very beneficial if we could build against Qt5 on all platforms that we support. That is no problem for MacOSX and Windows as we have to ship with the Qt. However on Linux, where we want to depend on the upstream packages, we need to double check. From which versions of the distros are Qt5 packages available in the main repositories?

| Distro | first (supported) version that provides Qt5 | has Webkit |
| --- | --- | --- |
| openSUSE | openSUSE 13.1, before in other repo | yes |
| Fedora | ? |  |
| Ubuntu | 14.04 LTS provides Qt5-default, earlier? |  |
| Debian | ? |  |
| CentOS 6/7 | Qt 5.4 via EPEL | yes |

Also, do the version provide Webkit?

https://github.com/owncloud/client/wiki/Qt-5-Dependency-on-Linux

guruz:
Yes, please get all those technology-scared Linux users to a newer Qt version.

Another nice reason for using Qt5: We can use the function pointer signal slot syntax..

http://woboq.com/blog/new-signals-slots-syntax-in-qt5.html

Would easily fix issues as
https://github.com/owncloud/client/commit/b0dbb49ca84a42f701f75270334f3e1714de9074 https://github.com/owncloud/client/commit/12ac9f9aa9cd97d4331aa3c0d59d49defd4c5157 https://github.com/owncloud/client/commit/803cb5d015f998c9d60112d977f5b495730ed320

dragotin:
For the upcoming 1.8.0 client: packages for openSUSE starting with 13.2 now build against Qt5. In 13.1 it uses Qt 5.3.x and thus neon is still required, but in Factory there (currently) is Qt 5.4.0 and we can build without neon.

Please help testing the packages from https://software.opensuse.org/download.html?project=isv%3AownCloud%3Acommunity%3Anightly&package=owncloud-client if possible.

If other distros have Qt5 packages in their default repositories available I am happy to integrate patches into our build service projects.

dragotin:
Fedora 21 is also building against Qt5.

danimo:
I have an experimental OBS build for CentOS 7 (soon 6) that depends on EPEL for Qt 5.4.

I'm looking for testers and input.

danimo:
@moscicki Can you run this solution (Client on CentOS/SCL requiring EPEL) past your deployment guys? It would allow us to jump to Qt 5.

[...]

Issue #5470, opened 2017-01-17
Qt-5.6.2 everywhere


jnweiger:
Public client builds
* [x] Windows https://github.com/owncloud/client/issues/3449
* [x] Mac https://github.com/owncloud/client/issues/3449
* [x] Linux clients should all use the same Qt version: 5.6.2

jnweiger:
@SamuAlfageme https://github.com/owncloud/client/issues/2558#issuecomment-99424884 lists many issues we identified in qt-5.5.1, and older versions (including qt4)
Can you please design a test suite to cover all these issues? This helps to demonstrate that qt-5.6.2 is a good choice.

jnweiger:
@guruz https://github.com/owncloud/client/issues/3449 mentions patches. Are these obsolete with 5.6.2?

guruz:
@jnweiger Yes, no patches needed

moscicki:
Please also remember to take on board this comment: https://github.com/owncloud/client/issues/2558#issuecomment-75262413

Or shall I open another issue for it?

guruz:
What's the status on this? People in https://github.com/owncloud/client/issues/4828 are waiting for it..

jnweiger:
We have set up a devel project in obs for providing Qt-5.6.2 packages for the Linux platforms. https://build.opensuse.org/package/show/isv:ownCloud:devel:Qt562/qt5-qtbase

The plan is:
* [ ] collect all needed packages there
* [ ] define the list of Linux build targets.
* [ ] make these packages build on all targets.
** [ ] SUSE rpms, [ ] Fedora rpms, [ ] centOS rpms
** [ ] Ubuntu DEB packages, [ ] Debian DEB packages.
* [ ] change the packages to install without conflicting with existing system qt packages
* [ ] change the packages to allow proper requires/provides
* [ ] test build with owncloud-client-2.2.4 and 2.3.0

jnweiger:
Qt-5.6.2 completion will not be ready for 2.3.0
the packaging roadmap for the 2.3.0 release are

* [ ] Roll back 2.2.4 in the testing project to using the Qt4/Qt5 packages from the distributions.
* [ ] Test by comparing with 2.2.3 packages that the rollback is complete and without regressions.
* [ ] Re-Release 2.2.4 packages 
* [ ] create 2.3.0 setup in the testing project using the Qt4/Qt5 packages from the distributions.
* [ ] Release 2.3.0 packages.
[...]

PR #5505, opened 2017-01-30
Remove all conditional checks for Qt 4.x

vladimiroff:
Now that #2558 has been closed there is no point of having to check whether Qt is <5.0.


michaelstingl:
https://github.com/owncloud/client/issues/2558 has been replaced by https://github.com/owncloud/client/issues/5470 (2.3.1 release)

ogoffart:
We still have to wait a bit before we can totally stop supporting Qt 4.
Then we will have to decide what's the minimal version of Qt we support. And we can also drop all the old Qt 5.x.

guruz:
@vladimiroff This has merge conflicts now, could you re-base?

guruz:
thanks @vladimiroff

Issue #5499, opened 2017-01-27
Brandings: Don't migrate from ownCloud.cfg if branded config is not readable


guruz:
Unproven but plausible.

Issue found by @jturcotte 

In ```AccountManager::restore()``` we need to not ```restoreFromLegacySettings``` if the branded config is there and can't be read for locking or whatever reasons.

EDIT: We also should have more debug output when doing such a migration

guruz:
Waiting for info from @michaelstingl

michaelstingl:
@guruz plz check https://github.com/owncloud/enterprise/issues/1729#issuecomment-277098702

ckamm:
@jturcotte @guruz If the settings were restored from the old `ownCloud/owncloud.cfg` path, we should see a `Migrate: checking old config` log message. That's the legacy-legacy code path. I'll add a debug message to the legacy path and check `QSettings::status`.

ckamm:
Pull request: https://github.com/owncloud/client/pull/5523

PR #5523, opened 2017-02-08
Settings: Don't migrate settings on access error #5499

ckamm:
Previously, we'd try migrating from legacy settings if reading
the settings failed with an error. Now, we try again after a
couple of seconds and eventually give up.

### DIFF
--- a/src/gui/accountmanager.cpp
+++ b/src/gui/accountmanager.cpp
@@ -46,11 +46,17 @@ AccountManager *AccountManager::instance()
 bool AccountManager::restore()
 {
     auto settings = Utility::settingsWithGroup(QLatin1String(accountsC));
+    if (settings->status() != QSettings::NoError) {
+        qDebug() << "Could not read settings from" << settings->fileName()
+                 << settings->status();
+        return false;
+    }
 
     // If there are no accounts, check the old format.
     if (settings->childGroups().isEmpty()
             && !settings->contains(QLatin1String(versionC))) {
-        return restoreFromLegacySettings();
+        restoreFromLegacySettings();
+        return true;
     }
 
     foreach (const auto& accountId, settings->childGroups()) {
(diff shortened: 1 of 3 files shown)


src/gui/application.cpp
```diff
@@ -156,7 +156,23 @@ Application::Application(int &argc, char **argv) :
...
 
-    AccountManager::instance()->restore();
+    if (!AccountManager::instance()->restore()) {
+        // If there is an error reading the account settings, try again
+        // after a couple of seconds, if that fails, give up.
+        // (non-existence is not an error)
+        Utility::sleep(5);
+        if (!AccountManager::instance()->restore()) {
+            qDebug() << "Could not read the account settings, quitting";
+            QMessageBox::critical(
```
guruz: Will this message actually display if you quit() in next one?

ckamm: Yes, the call blocks, I tested it.

src/gui/application.cpp
```diff
@@ -156,7 +156,23 @@ Application::Application(int &argc, char **argv) :
...
+                        0,
+                        tr("Error accessing the configuration file"),
+                        tr("There was an error while accessing the configuration "
+                           "file at %1.").arg(ConfigFile().configFile()),
+                        tr("Quit ownCloud"));
+            QTimer::singleShot(0, qApp, SLOT(quit()));
```
jturcotte: Is the timer required? quit seems to tell the event loop to exit and not exit immediately.

src/gui/accountmanager.cpp
```diff
@@ -69,6 +75,9 @@ bool AccountManager::restore()
 
 bool AccountManager::restoreFromLegacySettings()
 {
+    qDebug() << "Migrate: restoreFromLegacySettings, checking settings group"
+             << Theme::instance()->appName();
+
     // try to open the correctly themed settings
     auto settings = Utility::settingsWithGroup(Theme::instance()->appName());
```
jturcotte: Could it be worth to also check this one? Maybe also in `saveAccount*` methods, which could mean that other accounts could be lost?

PR #5526, opened 2017-02-10
Fix all unit tests on macOS

jturcotte:
Also fixes the UNIT_TESTING build on Windows.


csync/src/csync_private.h
```diff
@@ -207,7 +207,7 @@ __attribute__ ((packed))
 #endif
 ;
 
-void csync_file_stat_free(csync_file_stat_t *st);
+OCSYNC_EXPORT void csync_file_stat_free(csync_file_stat_t *st);
```
ogoffart: In Qt there is a  Q_XXX_AUTIOTEST_EXPORT  for this use case.

I'm not saying we should do the same.

guruz:
(Haven't looked at the code actually because I assume @ogoffart did)

src/libsync/utility.cpp
```diff
@@ -277,6 +277,9 @@ bool Utility::fsCasePreserving()
     if( isWindows() || isMac() ) {
         re = true;
     } else {
+#ifndef WITH_TESTING
+        static
```
jturcotte: Fixed the test to run on macOS instead.

csync/src/csync_private.h
jturcotte: Qt however tries to be tidy with its export, I think we can live without it unless it poses a concrete problem.

Issue #5484, opened 2017-01-23
Keep files option after aboutToRemoveAllFiles() signal ignores previously unchecked folders


SamuAlfageme:
### Expected behavior
When trying to remove all files under a folder sync connection and cancelling the operation (i.e. the removal was a mistake) the local folder is restored to the original state, previous to the deletion operation. 

### Actual behavior
The previously unchecked folders are ignored and therefore, downloaded. See [this video](https://webmshare.com/play/WJmND) to see it reproduced. 

### Steps to reproduce
1. Uncheck some folders in a folder sync connection 
2. Remove all the files in the local folder sync connection
3. Select `Keep files` as resolution strategy

ogoffart:
Ah, of course, the "keep file" option deletes the database which removes the blacklist.  We need to do something else.

SamuAlfageme:
👍

PR #5531, opened 2017-02-14
Folder: Keep files option after aboutToRemoveAllFiles should not clea…

ogoffart:
…r selective sync list

We were removing the wholme journal db when the user wanted to keep all files,
But that would also remove the selective sync lists.
We should only remove the metadata table.

Issue #5484

### DIFF
--- a/test/testallfilesdeleted.cpp
+++ b/test/testallfilesdeleted.cpp
@@ -1,0 +1,118 @@
+/*
+ *    This software is in the public domain, furnished "as is", without technical
+ *    support, and with no warranty, express or implied, as to its usefulness for
+ *    any purpose.
+ *
+ */
+
+#include <QtTest>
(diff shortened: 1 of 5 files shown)


test/testallfilesdeleted.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+#include <QtTest>
+#include "syncenginetestutils.h"
+#include <syncengine.h>
+
+using namespace OCC;
+
+/*
+ * This test ensure that the SyncEngine::aboutToRemoveAllFiles is correctly called and that when
+ * we the user choose to remove all files SyncJournalDb::clearFileTable makes works as exprected
```
phil-davis: s/exprected/expected

SamuAlfageme:
👍 it is

PR #5532, opened 2017-02-14
Update windows docker file to use 2.3 OBS repo

rullzer:
While building a themed client for 2.3 I noticed your Dockerfile was not up to date.


guruz:
Thanks. I don't think we're using that file anymore.

ogoffart:
@guruz,  if we don't use it, we need to remove it.  Also the documentation still mention it.

rullzer:
@guruz as @ogoffart mentions it is still in the docs :)

ogoffart:
Even if we use another script on obs, the docker image still need to be kept up to date.

PR #5540, opened 2017-02-17
Propagator: Fix t8.pl

ogoffart:
The test sets OWNCLOUD_MAX_PARALLEL to 1 to disable parallelism.
But since the max amount of parallelism is twice as much, that does not
work.
So change the way we compute the hardMaximumActiveJob: Use the value of
OWNCLOUD_MAX_PARALLEL to maximize this amount and base the maximum amount
of transfer jobs on it instead of the other way.

A result of this change is that, in case of bandwidth limit, we keep the
default of 6 non-transfer jobs in parallel. I believe that's fine since
the short jobs do not really use bandwidth, so we can still keep the same
amount of small jobs.

### DIFF
--- a/src/libsync/owncloudpropagator.cpp
+++ b/src/libsync/owncloudpropagator.cpp
@@ -40,6 +40,7 @@
 #include <QObject>
 #include <QTimerEvent>
 #include <QDebug>
+#include <qmath.h>
 
 namespace OCC {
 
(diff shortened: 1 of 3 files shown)


jturcotte:
Why not:
```
    static int max = qgetenv("OWNCLOUD_MAX_PARALLEL").toUInt();
    if (!max) {
        max = maximumActiveJob() * 2;
    }
```

and also continue using OWNCLOUD_MAX_PARALLEL in maximumActiveJob.

ogoffart:
@jturcotte  Because I think it's fine to use more "small" jobs when the bandwith limit is not set.
(And also i wanted to read the env variable in only one place)
(and also because your code does not adapt when the bandwithlimit is changed, as max is static)

PR #5542, opened 2017-02-17
Create PropagateItemJobs on the fly

jturcotte:
This is pull request finishes the work to fix TestSyncEngine::abortAfterFailedMkdir by ensuring that we either go down in the job tree (e.g. scheduleNextJob, abort), or up (finished) on a stack to avoid reentrancy.

This is mixed with refactoring to split the subjob logic from PropagateDirectory in a PropagateCompositeJob class to make it easier to add a dynamic creation of item jobs.

### DIFF
--- a/src/libsync/owncloudpropagator.cpp
+++ b/src/libsync/owncloudpropagator.cpp
@@ -381,22 +381,22 @@ void OwncloudPropagator::start(const SyncFileItemVector& items)
                 }
             } else {
                 PropagateDirectory* currentDirJob = directories.top().second;
-                currentDirJob->append(dir);
+                currentDirJob->appendJob(dir);
             }
             directories.push(qMakePair(item->destination() + "/" , dir));
-        } else if (PropagateItemJob* current = createJob(item)) {
+        } else {
             if (item->_instruction == CSYNC_INSTRUCTION_TYPE_CHANGE) {
                 // will delete directories, so defer execution
-                directoriesToRemove.prepend(current);
+                directoriesToRemove.prepend(createJob(item));
                 removedDirectory = item->_file + "/";
             } else {
-                directories.top().second->append(current);
+                directories.top().second->appendTask(item);
             }
         }
     }
 
     foreach(PropagatorJob* it, directoriesToRemove) {
-        _rootJob->append(it);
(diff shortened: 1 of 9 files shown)


src/libsync/owncloudpropagator.h
```diff
@@ -227,7 +227,7 @@ class OWNCLOUDSYNC_EXPORT PropagateDirectory : public PropagatorJob {
 private slots:
     bool possiblyRunNextJob(PropagatorJob *next) {
         if (next->_state == NotYetStarted) {
-            connect(next, SIGNAL(finished(SyncFileItem::Status)), this, SLOT(slotSubJobFinished(SyncFileItem::Status)), Qt::QueuedConnection);
+            connect(next, SIGNAL(finished(SyncFileItem::Status)), this, SLOT(slotSubJobFinished(SyncFileItem::Status)));
```
ogoffart: Please don't commit this, we established it is wrong.  
Or do you fix it later?

jturcotte: Yes, notably the invokeMethod additions in "Create PropagateItemJobs only before starting them" and all of "Don't abort propagation job abortions synchronously from finished". That's why I've put them in the same pull request.

jturcotte: This makes that commit non-free of regression by itself, but I've spent too much time on this pull request already so I thought I might be OK.

src/libsync/owncloudpropagator.cpp
```diff
@@ -386,33 +390,29 @@ void OwncloudPropagator::start(const SyncFileItemVector& items)
...
-                currentDirJob->append(dir);
+                currentDirJob->appendJob(dir);
             }
             directories.push(qMakePair(item->destination() + "/" , dir));
-        } else if (PropagateItemJob* current = createJob(item)) {
+        } else {
             if (item->_instruction == CSYNC_INSTRUCTION_TYPE_CHANGE) {
                 // will delete directories, so defer execution
-                directoriesToRemove.prepend(current);
+                directoriesToRemove.prepend(createJob(item));
```
ogoffart: directoriesToRemove could actually be a vector of items (no need to create the job before)
(that's just a further optimisation.)

PR #5546, opened 2017-02-20
Add a Vagrant provision configuration

settermjd:
This PR:

- Adds a Vagrant provisioning setup to the client docs

:information_desk_person: While not strictly necessary, having it makes it easier to build and test the documentation before commits are made. It's a copy of the setup from the documentation repository, so should really be separated out into a separate repository and included with git subtree or something similar.


settermjd:
@danimo @dschmidt can one or both of you critique pls?

dschmidt:
Maybe @tboerger has any opinion on this? I haven't worked with vagrant or similar tools so far.

tboerger:
A verhängt setup just to build the docs?

SamuAlfageme:
@settermjd I agree maybe this is too much for the client repository since `sphinx-build` wraps most of the difficult parts. Definitely worthy of the Documentation repo, where setting everything up can be very messy. 

We could include a note on how to build the docs with the community theme on the docs itself. What do you think?

tboerger:
If this works like expected I'm fine with that. But Jenkins fails.

guruz:
> While not strictly necessary

Maybe it could be a git submodule then?
I don't have an opinion here..

settermjd:
Ok, perhaps this is overkill. I'll look for something simpler, perhaps Docker.

Issue #4617, opened 2016-03-29
Problem on building owncloud client !


yavama:
dhouha@dhouha:~/client-build$ cmake -DCMAKE_BUILD_TYPE="Debug" /home/dhouha/client
-- Build of crashreporter disabled.
-- Found Qt5 core, checking for further dependencies...
CMake Error at cmake/modules/QtVersionAbstraction.cmake:22 (find_package):
  By not providing "FindQt5WebKitWidgets.cmake" in CMAKE_MODULE_PATH this
  project has asked CMake to find a package configuration file provided by
  "Qt5WebKitWidgets", but CMake did not find one.

  Could not find a package configuration file provided by "Qt5WebKitWidgets"
  with any of the following names:

    Qt5WebKitWidgetsConfig.cmake
    qt5webkitwidgets-config.cmake

  Add the installation prefix of "Qt5WebKitWidgets" to CMAKE_PREFIX_PATH or
  set "Qt5WebKitWidgets_DIR" to a directory containing one of the above
  files.  If "Qt5WebKitWidgets" provides a separate development package or
  SDK, be sure it has been installed.
Call Stack (most recent call first):
  CMakeLists.txt:54 (include)

-- Configuring incomplete, errors occurred!
See also "/home/dhouha/client-build/CMakeFiles/CMakeOutput.log".

i'm working on ubuntu 14.04

yavama:
@pierrejochem Help plz :)

ogoffart:
You need QtWebkit dev packages.

yavama:
@ogoffart  how to install it plz

yavama:
@ogoffart i have libqt5webkit5 installed and i still have the same message

danimo:
@yavama First of all, you need the -dev package. Secondly, you cannot build the client with Qt 5.1.0 that comes with 14.04. It has too many issues. That's the reason why the official packages still build with 4.8.

yavama:
@danimo what do you mean with the -dev package plz ? 
what should i do exactlyyy

yavama:
i now got this :

dhouha@dhouha:~/client-build$ cmake -DCMAKE_BUILD_TYPE="Debug" /home/dhouha/client
-- Build of crashreporter disabled.
-- Found Qt5 core, checking for further dependencies...
-- Using Qt 5!
CMake Warning at cmake/modules/QtVersionAbstraction.cmake:75 (find_package):
[...]

PR #5549, opened 2017-02-21
Fix documentation build process

settermjd:
This PR:

- Fixes the documentation build process so that it can run in a UNIX/Linux environment. 

:information_desk_person: I see that `doc/Makefile` was removed in 98efab83, but I'm not sure why. May need further investigation to fully appreciate the reasons for the decision, at the time.


guruz:
Might as well be an accidental removal by @dragotin ?

settermjd:
Could well be. After some investigation, the "Generic Build Instructions" are not complete and definitely need some updates. For example, that Ubuntu 14.04 is not supported, as documented in https://github.com/owncloud/client/issues/4617, needs to be mentioned.

settermjd:
I'm going to merge this one to get the build working again, as it's a sufficiently high priority. Then I'm going to update the docs. In the process of doing so, this PR may become unecessary. But, at least we're moving forward.

guruz:
👍

PR #5550, opened 2017-02-21
Update the generic build instructions

settermjd:
**This PR:**

- Updates the generic build instructions with relevant background/dependency information. Without it, the build will fail (unless the user just happens to have all the dependencies installed. It also adds a note letting the user know that Ubuntu 14.04 is not supported. This was an issue which I ran into while attempting to test the instructions. I've referenced the relevant GitHub issue to provide further information, should the user want to know more about it.

This PR needs #5546 to be merged for the user to be able to follow it if they choose to use the referenced VM.

### DIFF
--- a/doc/building.rst
+++ b/doc/building.rst
@@ -224,7 +224,31 @@ Compared to previous versions, building the desktop sync client has become easie
 earlier versions, CSync, which is the sync engine library of the client, is now
 part of the client source repository and not a separate module.
 
-To build the most up-to-date version of the client:
+To build the most up-to-date version of the client, requires four mandatory steps (and one optional one).
+But, before you can run those, make sure you have all the dependencies installed. 
+These are:
+
+* build-essential
+* bzip2
+* cmake
+* doxygen
+* git
+* gzip
+* lacheck
+* libdbusmenu-qt5-dev 
+* libqt5webkit5-dev
+* libsqlite3-dev
+* libssl-dev
+* python3
+* qtkeychain-dev
+
+.. note:: 
+   Feel free to use the included VM which has all of the dependencies available. All you need to run it are the latest version of VirtualBox and Vagrant. These are ``5.0.32 r112930`` and ``1.9.1``, respectively, at the time of writing. 
+   
+.. warning::
+   Please know that Ubuntu 14.04 is NOT supported as the client will not build properly. You need at least version 16.04 (which is available in the Vagrant/VirtualBox VM). For more information, have a read through https://github.com/owncloud/client/issues/4617.
+
+With the dependencies satisfied, you can now build the client.
 
 1. Clone the latest versions of the client from Git_ as follows::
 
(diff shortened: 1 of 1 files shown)


doc/building.rst
```diff
@@ -224,7 +224,31 @@ Compared to previous versions, building the desktop sync client has become easie
 earlier versions, CSync, which is the sync engine library of the client, is now
 part of the client source repository and not a separate module.
 
-To build the most up-to-date version of the client:
+To build the most up-to-date version of the client, requires four mandatory steps (and one optional one).
+But, before you can run those, make sure you have all the dependencies installed.
```
ogoffart: We are in the generic build instruction chapter,  but this seems to be specific to ubuntu.

doc/building.rst
```diff
@@ -224,7 +224,31 @@ Compared to previous versions, building the desktop sync client has become easie
...
+These are:
+
+* build-essential
+* bzip2
+* cmake
+* doxygen
+* git
+* gzip
+* lacheck
+* libdbusmenu-qt5-dev
```
ogoffart: Why do we need libdbusmenu?

doc/building.rst
```diff
@@ -224,7 +224,31 @@ Compared to previous versions, building the desktop sync client has become easie
...
+* git
+* gzip
+* lacheck
+* libdbusmenu-qt5-dev 
+* libqt5webkit5-dev
+* libsqlite3-dev
+* libssl-dev
+* python3
+* qtkeychain-dev
+
```
ogoffart: There are missing also the optional kde dependences to build the Dolphin integration plugin.

doc/building.rst
```diff
@@ -224,7 +224,31 @@ Compared to previous versions, building the desktop sync client has become easie
...
+.. note:: 
+   Feel free to use the included VM which has all of the dependencies available. All you need to run it are the latest version of VirtualBox and Vagrant. These are ``5.0.32 r112930`` and ``1.9.1``, respectively, at the time of writing. 
+   
+.. warning::
+   Please know that Ubuntu 14.04 is NOT supported as the client will not build properly. You need at least version 16.04 (which is available in the Vagrant/VirtualBox VM). For more information, have a read through https://github.com/owncloud/client/issues/4617.
```
ogoffart: Is it not?  I think Piotr is building it in 14.04 and it works fine.

SamuAlfageme: > It also adds a note letting the user know that Ubuntu 14.04 is not supported.

How come? https://github.com/owncloud/client/issues/4617 is a particular problem with packages provided in the bare distro, but it's perfectly possible to build the client in Ubuntu 14.04. 

Also note that Ubuntu 14.04 is in the list of distros supported under https://software.opensuse.org/download/package?project=isv:ownCloud:desktop&package=owncloud-client

doc/building.rst
settermjd: Fair point.

doc/building.rst
settermjd: It seemed to be required when I was working through the instructions. Perhaps I bundled it in with another package, so it seemed like it was needed, when it wasn't.

doc/building.rst
settermjd: Ah ok. The package built successfully, according to the instructions. So I never came across the need for it.

doc/building.rst
settermjd: @ogoffart / @SamuAlfageme I inferred from https://github.com/owncloud/client/issues/4617#issuecomment-202856120 that 14.04 wasn't supported. I wasn't able to successfully build in that release. If that's not the case, then I'll update the docs further.

settermjd:
Closing as the issue seems redundant.

doc/building.rst
SamuAlfageme: @settermjd you can always get a different Qt version than the shipped with the distro to get it working.

Issue #5537, opened 2017-02-16
Make insufficient storage errors visible without attempting too frequent uploads


ckamm:
This is a follow up to #5346 and owncloud/enterprise#1831.

Before 2.3, 507 Insufficent Storage errors used to be blacklisted. That had the positive effect of making the client not reattempt the upload very frequently. But it also hid the error and sometimes users weren't even aware that there was a storage problem.

For 2.3 we removed the blacklisting. That made the error visible, but lead to too-frequent uploads.

The behavior of the client in situations with insufficient storage space on the server needs to be adjusted. We want users to become aware of the problem while not attempting too many uploads that have a high likelihood of failure.

ckamm:
Okay, so what do we want to happen if the server has insufficient storage?

The obvious path would be to trust the quota the server reports and prohibit uploads that would exceed the quota and make the whole sync fail with a "Insufficent storage" normal error if that happens.

SamuAlfageme:
@ckamm +1 to being reactive to the quota, that would solve collateral problems like: https://github.com/owncloud/client/issues/5346#issuecomment-264810401 (changes in user's quota)

ckamm:
@SamuAlfageme Actually, using quota changes to trigger syncs is not part of this commit yet. 

To do that, we'd need to check quota regularly, even if the settings dialog is not visible. We'd also need to track that there are files that couldn't be synced due to quota limitations. It goes beyond what I wanted to accomplish here.

ckamm:
Unfinished pull request: https://github.com/owncloud/client/pull/5552

Current status: We can have different quotas per folder and currenty don't track them. Fully addressing this needs significantly more work.

GrendelWWU:
Hello, I understand the problem. You can have your own folder and a share from another user. But if one account is full, you can't upload to that folder. The client does only see the sum and thinks, that he can upload. So far, so good.

Maybe we can realize a solution in 2 steps?
1. when encountering "Insufficient Storage" error for several times, slowing down the attempts for upload. First a few seconds upto several minutes. Each time let the client put an error message via popup. So we don't fill the logs with more than 80000 tries from just 1 user, as I've seen today. This delay-interval can be decreased again, when the server stops sending error messages about storage.
2. making the server smart, that it reports the quota for several upload paths so the client can stop uploading the folders, which are full. Restart will only begin with this folder, when the server reports free space again for this particular upload path.

BTW... step 1 can stay active too, even when #2 is realized...

Just my proposal to this issue.

[...]

PR #5552, opened 2017-02-22
Don't upload if it would exceed remote quota #5537

ckamm:
Targetting master since it's uncertain whether this should be in 2.3.1 or 2.4.0! Assigning to 2.3.1 for now - opinions?

* Run a Propfind to determine the available quota before each sync run.
* Keep the available quota up to date with each upload and delete job.
* Don't start upload jobs if there's not enough available space.
* If we (unexpectedly) receive a 507 Insufficent Storage error, adjust
  the expectation of available remote space downwards.

### DIFF
--- a/src/libsync/propagateupload.cpp
+++ b/src/libsync/propagateupload.cpp
@@ -331,6 +331,23 @@ void PropagateUploadFileCommon::slotStartUpload(const QByteArray& transmissionCh
         return;
     }
 
+    // Check that there's enough space on the server
+    if (!propagator()->quotaCheck(_item->_size, _item->log._other_size)) {
+        done(SyncFileItem::NormalError, tr("Storage quota on the server is exceeded"));
+        return;
+    }
+
+    // If we will end up needing more space, eat up the available quota
+    // right now. That way we won't start parallel jobs that compete for
+    // the remaining quota if we know we can't fit both.
+    // However, if we will *free up* quota through the PUT request (say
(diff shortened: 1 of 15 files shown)


src/libsync/propagateupload.cpp
```diff
@@ -521,6 +538,21 @@ void PropagateUploadFileCommon::checkResettingErrors()
     }
 }
 
+void PropagateUploadFileCommon::checkInsufficentStorageError()
```
ogoffart: On the other hand, maybe the folder is on a different storage or with different quota requirement that the main quota. So we might mess the quota up.

ckamm:
@ogoffart Hm, you're right about quotas potentially differing per folder - I didn't think about shared folders. So we actually need to track quota-available per-folder for good behavior. I'll think about it. Definitely not for 2.3.1 then.

ckamm:
@ogoffart What if we retrieved quota information of a folder before each upload. Once per folder only, possibly only for larger uploads, or for folders that have seen an insufficent storage error before.

It'd add quite a bit of bookkeeping but is currently the sanest option I see. Thoughts?

The best option would be being able to retrieve the 'mount points' with individual quotas from the server. I doubt something like that exists currently?

ckamm:
We haven't decided about what to do here. I don't see this in 2.4.0.

ckamm:
We'll not do it this way.

PR #5559, opened 2017-02-23
Make sure account state is not NULL before fetch

vladimiroff:
In `slotRefreshActivity` is assumed that given `ast` could be NULL and
yet is still being passed to `startFetchJob` which could segfault while
checking if connected.

This makes sure that `ast` is not NULL before using it.

### DIFF
--- a/src/gui/activitylistmodel.cpp
+++ b/src/gui/activitylistmodel.cpp
@@ -116,7 +116,7 @@ bool ActivityListModel::canFetchMore(const QModelIndex& ) const
 
 void ActivityListModel::startFetchJob(AccountState* s)
 {
-    if( !s->isConnected() ) {
+    if( !s || !s->isConnected() ) {
         return;
     }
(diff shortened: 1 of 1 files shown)


ogoffart:
I don't think ast could be null.  The test in slotRefreshActivity is redundent.
Did you experience any crash?

vladimiroff:
No, I didn't experienced a crash. Just going through the code I saw that check if `ast` is null and decided to learn how is this handled later in `startFetchJob`. Feel free to close this if you're sure that there's no case when `ast` is actually null.

ckamm:
@vladimiroff Yes, that null check there is confusing because it hints towards an edge case that doesn't exist. `slotRefreshActivity` is called from `ActivityWidget::slotRefreshActivities` is called from `ActivitySettings::slotRefresh` only with non-null arguments. It'd be invalid to call it with a null pointer. (so yeah, it could just be a reference, but that's not the style that code is written in)

PR #5561, opened 2017-02-24
Traymenu quota info

rullzer:
Adds a section to the traymenu to show the quota info (or just used space if there is no quota). Just like in the account panel.

Looks like:
sh

### DIFF
--- a/src/gui/owncloudgui.cpp
+++ b/src/gui/owncloudgui.cpp
@@ -34,6 +34,7 @@
 #include "openfilemanager.h"
 #include "accountmanager.h"
 #include "creds/abstractcredentials.h"
+#include "quotainfo.h"
 
 #include <QDesktopServices>
 #include <QDir>
@@ -346,6 +347,13 @@ void ownCloudGui::addAccountContextMenu(AccountStatePtr accountState, QMenu *men
     auto actionOpenoC = menu->addAction(browserOpen);
     actionOpenoC->setProperty(propertyAccountC, QVariant::fromValue(accountState->account()));
     QObject::connect(actionOpenoC, SIGNAL(triggered(bool)), SLOT(slotOpenOwnCloud()));
+
+    menu->addSeparator();
+
+    QString quotaInfo = tr("Fetching Quota");
+    auto actionQuotaInfo = menu->addAction(quotaInfo);
+    actionQuotaInfo->setEnabled(false);
+    new QuotaAction(actionQuotaInfo, accountState);
 
     FolderMan *folderMan = FolderMan::instance();
     bool firstFolder = true;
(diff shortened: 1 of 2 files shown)


src/gui/owncloudgui.cpp
```diff
@@ -347,6 +348,15 @@ void ownCloudGui::addAccountContextMenu(AccountStatePtr accountState, QMenu *men
     actionOpenoC->setProperty(propertyAccountC, QVariant::fromValue(accountState->account()));
     QObject::connect(actionOpenoC, SIGNAL(triggered(bool)), SLOT(slotOpenOwnCloud()));
 
+    if (accountState->state() == AccountState::Connected) {
+        menu->addSeparator();
+
+        QString quotaInfo = tr("Fetching Quota");
+        auto actionQuotaInfo = menu->addAction(quotaInfo);
+        actionQuotaInfo->setEnabled(false);
+        new QuotaAction(actionQuotaInfo, accountState);
```
ckamm: This will mean that a new quota request for all acounts is sent whenever the tray menu is updated. For some collections of tray-workarounds, the update can happen quite frequently. 

I'd prefer if the quota information for accounts could be updated occasionally (or on request, but not too frequently) and then be propagated to all components that are interested in that information, possibly via a signal and a getter on AccountState.

rullzer: The hacks for the tray on OS X you mean? Because I checked and on linux/windows the requests are just every 30 seconds.

rullzer: @ckamm It already happens max every 30 seconds https://github.com/owncloud/client/blob/master/src/gui/quotainfo.h#L33-L36

We could of course have that job run in the background and have the info propegate to other parts of the code. Down side is that we will request more from the server.

rullzer: Ah seems we can drop the OS X magic when 5.6.2 is used: https://github.com/owncloud/client/issues/5470 https://bugreports.qt.io/browse/QTBUG-54845

ckamm: @rullzer You make a new QuotaAction and thereby QuotaInfo every time the menu is updated, resetting QuotaInfo's 30s timer. So there will be a request for each account every time the menu is updated, and every 30s after that. This is in addition to the quota checks that happen when the settings are open.

I think tray menu regenerations are currently on-demand or max-every-30s. That's probably why it looks similar.

I think there should be one thing (probably QuotaInfo) gathering quota information regularly, and other components should make use of that information. Instantiating another quota-retriever to get that data into the tray sets a wrong precedent even if it's probably few enough requests to not matter much.

rullzer: Fair enough. I'll look into it.

ogoffart:
The problem is that requesting the Quota every 30s will mean some more load on the server. (every client requesting the quota every 30 seconds).  

We tried hard before to limit the recurrent call before (one PROPFIND per folder every 30s)
Other stuff are only requested when the UI is visible.  So maybe make it request the quota only when the context menu is shown?

rullzer:
@ogoffart yes that is what I remembered as well. THing is that with the current code on OS X we can't update live it seems. Altough that is fixed with 5.6.2 if I'm not mistaken.

guruz:
Indeed it would be nicer if this is on-menu-click and not possibly every 30 sec

Could you also put the action into a ```QPointer``` and check if it got set to 0 meanwhile? Since this is aynchronous..

ckamm:
@rullzer Are you planning to come back to this?

PR #5571, opened 2017-03-01
Follow redirects only for GET requests

### LINKED ISSUE #5555: Transparent redirect handling forces verb to GET
If `_followRedirects` is enabled - as is the case for nearly all requests - we attempt to follow redirects by automatically issuing follow-up requests. However, these follow ups will always use GET.

This probably makes the redirect handling not work for many requests (we've seen that it can break PROPFINDs) and might have detrimental effects on some of them (what if a PUT is redirected to a GET?)

@ogoffart Do you know if this is in any way intentional?

vladimiroff:
This fixes #5555.

### DIFF
--- a/src/libsync/abstractnetworkjob.cpp
+++ b/src/libsync/abstractnetworkjob.cpp
@@ -173,7 +173,7 @@ void AbstractNetworkJob::slotFinished()
     // get the Date timestamp from reply
     _responseTimestamp = _reply->rawHeader("Date");
 
-    if (_followRedirects) {
+    if (_followRedirects && reply()->operation() == QNetworkAccessManager::GetOperation) {
         // ### the qWarnings here should be exported via displayErrors() so they
         // ### can be presented to the user if the job executor has a GUI
         QUrl requestedUrl = reply()->request().url();


src/libsync/abstractnetworkjob.cpp
```diff
@@ -173,7 +173,7 @@ void AbstractNetworkJob::slotFinished()
     // get the Date timestamp from reply
     _responseTimestamp = _reply->rawHeader("Date");
 
-    if (_followRedirects) {
+    if (_followRedirects && reply()->operation() == QNetworkAccessManager::GetOperation) {
```
ckamm: @guruz Is this correct? Why is `operation` an attribute of the reply? Shouldn't it be part of the request?

ckamm: @vladimiroff It would be better to check after `_redirectCount` was incremented, log something and then not trigger a follow-up request (similar to the other error cases)

ogoffart:
The goal of the redirect was to support hte use case in which a server changes address.
Of course, the sync will be totally broken if we follow only redirection on GET.

ckamm:
Full redirect handling is in https://github.com/owncloud/client/pull/5577, superceding this.

Issue #5573, opened 2017-03-01
Additionnal ')' characters on each line in certifcate popup


ldmpub:
See screen shot:
![bug additionnal character](https://cloud.githubusercontent.com/assets/1452146/23480573/ae5301de-fec8-11e6-96d6-0ceef25d0e63.jpg)

Version 2.3.0rc3 (build 2254)

ogoffart:
Thanks for reporting the issue.
Fixing in https://github.com/owncloud/client/pull/5574

PR #5574, opened 2017-03-02
SslButton: Remove additional ')'

ogoffart:
Regression since 7bfe0613827aa7cee93a77ccbedab9b07ea9ca78
Issue: #5573


ogoffart:
@guruz: I believe this could go in the final without another beta. The one-character change is ok and safe

PR #5577, opened 2017-03-03
AbstractNetworkJob: Improve redirect handling #5555

ckamm:
* For requests:
  - reuse the original QNetworkRequest, so headers and attributes
    are the same as in the original request
  - determine the original http method from the reply and the request
    attributes
  - keep the original request body around such that it can be sent
    again in case the request is redirected

* Simplify the interface that is used for creating new requests in
  AbstractNetworkJob.

### DIFF
--- a/src/gui/notificationconfirmjob.cpp
+++ b/src/gui/notificationconfirmjob.cpp
@@ -54,9 +54,7 @@ void NotificationConfirmJob::start()
     req.setRawHeader("Ocs-APIREQUEST", "true");
     req.setRawHeader("Content-Type", "application/x-www-form-urlencoded");
 
-    QIODevice *iodevice = 0;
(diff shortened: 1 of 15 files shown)


ckamm:
Please review carefully, has the chance of breaking existing network calls. I'll run tests now.

ckamm:
Note: Previously GETFileJob and PUTFileJob were going through `QNetworkAccessManager::sendCustomRequest()` - with this change they are going through `::get()` and `::put()`. That may lead to slightly different behavior because QNAM will not treat it as `CustomOperation` internally.

ckamm:
@ogoffart You approved but your comments are a bit ambiguous, do you agree about merging this?

ckamm:
Updated due to compile failure with older Qt.

Issue #5578, opened 2017-03-06
GLib-CRITICAL: Source ID was not found when attempting to remove it (Ubuntu 14.04, client 2.3.0) 


s-ntech:
### Expected behaviour
After fresh install (Ubuntu 14.04, client 2.3.0 using Qt 4.8.6), client should be able to sync folders and the client should not crash.

### Actual behaviour
After downloading (syncing) a folder from remote for a short time (~10s) the client crashes. In the command line:

> (owncloud:9634): GLib-CRITICAL **: Source ID 251 was not found when attempting to remove it
Aborted (core dumped)

### Steps to reproduce
1. `sudo apt-get install owncloud-client`
2. Connect to remote folder and try to sync remote folder

### Server configuration
Operating system: Ubuntu 14.04 (3.13.0-110-generic)

Web server: Apache2 

Database: MySql

PHP version:

ownCloud version: 9.1.4

Storage backend (external storage):

### Client configuration
Client version: 2.3.0

Operating system: Ubuntu 14.04

OS language: En

Qt version used by client package (Linux only, see also Settings dialog): 4.6.8

Client package (From ownCloud or distro) (Linux only):

Installation path of client:

### Logs

Please use Gist (https://gist.github.com/) or a similar code paster for longer
logs.

```Template for output < 10 lines```

1. Client logfile: Output of `owncloud --logwindow` or `owncloud --logfile log.txt` (I'm showing the last 10 or so lines here, but can upload full log if required as well)
`03-06 10:43:54:952 0x1d07070 void OCC::AbstractNetworkJob::slotFinished() 2 "Connection closed" QVariant(int, 200) 
03-06 10:43:54:952 0x1d07070 void OCC::PropagateDownloadFile::slotGetFinished()  QUrl( "https://******.de/remote.php/webdav/instr/2011-Linear_Algebra_I/Aufgabeblätter/R Vektorräume.pdf" )  FINISHED WITH STATUS 2 "Connection closed" 0 179340 189666 0 "" "189666" 
03-06 10:43:54:952 0x1d07070 void OCC::SyncEngine::slotItemCompleted(const SyncFileItemPtr&) "2011-Linear_Algebra_I/Aufgabeblätter/R Vektorräume.pdf" INSTRUCTION_NEW 1 "Connection closed" 
03-06 10:43:54:971 0x1d07070 void OCC::AbstractNetworkJob::slotFinished() 5 "Operation canceled" QVariant(, ) 
[...]

PR #5579, opened 2017-03-06
Propagator: Fix finished signal of directory being emited twice and c…

ogoffart:
…ausing crash

When there is a FatalError, we ended up emiting the finished signal for
the directory job several times, which would lead to crashes

Issue #5578


src/libsync/owncloudpropagator.cpp
```diff
@@ -721,8 +723,10 @@ void PropagateDirectory::finalize()
             }
         }
     }
-    _state = Finished;
-    emit finished(_hasError == SyncFileItem::NoStatus ?  SyncFileItem::Success : _hasError);
+    if (_state != Finished) {
+        _state = Finished;
+        emit finished(_hasError == SyncFileItem::NoStatus ?  SyncFileItem::Success : _hasError);
+    }
```
jturcotte: Would it make sense to also avoid the DB update in that case?

ogoffart:
@jturcotte   patch updated.

Issue #5544, opened 2017-02-19
Client stuck in upload loop on similarly named files


pylorak:
If you have a folder that contains two files whose names differ only in casing, the Windows client gets stuck in a loop and keeps reuploading one of the files every minute forever. This situation can happen for people dual-booting Windows with Linux and storing an owncloud-synchronized folder on a shared drive. Windows has otherwise no problems accessing or processing the files created this way.

Using Windows client version 2.2.4 on an NTFS partition under Win10.

I renamed one of the files and the upload-loop stopped, so the fix is easy IF you happen to figure it out. This has been happening silently in the background though for the past few days, and I am afraid it will have cost me a lot in mobile data transfer by the end of the month.

The desktop client should handle this situation and *not* keep reuploading the same file over and over again if a differently-cased file exists in the same folder.

hodyroff:
@SamuAlfageme Looks like a classic corner case to look into after 2.3 is out of the door.

SamuAlfageme:
@pylorak I've tried using [`win-sshfs`](https://code.google.com/archive/p/win-sshfs) to mount a remote Linux NTFS partition with a filename case collision (should be the same for dual-boot-like shared drives) and I've not been able to reproduce for the moment since windows is ignoring one of the files (the second queued for download). 

I'll try to reproduce with the same shared folder setup you propose to see if I'm able to see what's going on in the logs (if you have your own, please paste them here too). Thanks for reporting! 😉

ckamm:
I also attempted to reproduce, but it worked fine in Win7 with a case collision on a non-ntfs filesystem. I think we've previously seen problems when an NTFS filesystem is adjusted to states that are usually unreachable in Windows (by mounting in linux, for example).

pylorak:
I'm not sure if it can give any more info than what I already explained, but here is a screenshot of the sync log of the phenomenon. As you can see the affected file was named TcomTcp.cpp, the other was TCOMTCP.cpp. I renamed the latter to something else and the client immediately stopped resyncing the other one. They were both originally created under Linux in the same location, but I only have the sync client running under Windows.

The client is of nextcloud, but its repository tells everybody they should report problems here at owncloud, as the nextcloud repository for the client is only for theming. So both should be exhibiting the same problem.

![nc-issue](https://cloud.githubusercontent.com/assets/4016335/23127765/77fcf3cc-f77c-11e6-9c72-f2daf344a80b.png)

ckamm:
[...]

PR #5583, opened 2017-03-07
Windows/NTFS: Do not attempt to upload inaccessible files #5544

ckamm:
It is possible to create files with filenames that differ
only by case in NTFS, but most operations such as stat and
open only target one of these by default.

When that happens, we want to avoid uploading incorrect data
and give up on the file.

Typically this situation should never occurr during normal use
of Windows. It can happen, however, when a NTFS partition is
mounted in another OS.

### DIFF
--- a/src/libsync/owncloudpropagator.cpp
+++ b/src/libsync/owncloudpropagator.cpp
@@ -509,6 +509,36 @@ bool OwncloudPropagator::localFileNameClash( const QString& relFile )
     return re;
 }
 
+bool OwncloudPropagator::hasCaseClashAccessibilityProblem(const QString &relfile)
+{
+#ifdef Q_OS_WIN
+    bool result = false;
+    const QString file( _localDir + relfile );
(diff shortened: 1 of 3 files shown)


src/libsync/owncloudpropagator.cpp
```diff
@@ -509,6 +509,32 @@ bool OwncloudPropagator::localFileNameClash( const QString& relFile )
...
 
+bool OwncloudPropagator::hasCaseClashAccessibilityProblem(const QString &relfile)
+{
+#ifdef Q_OS_WIN
+    bool result = false;
+    const QString file( _localDir + relfile );
+    WIN32_FIND_DATA FindFileData;
+    HANDLE hFind;
+
+    hFind = FindFirstFileW((wchar_t*)file.utf16(), &FindFileData);
```
ogoffart: I don't like C-casts.  (static_cast or reinterpret_cast)

PR #5585, opened 2017-03-07
Add explanation of how to replace local sync folder to FAQ

settermjd:
This PR:

- Adds an explanation of how to replace the sync directory in the desktop client to the FAQ.

### Relates To

owncloud/documentation/issues/2923.


doc/faq.rst
```diff
@@ -18,3 +18,32 @@ The sync client has been intentionally limited to sync no deeper than
...
+
+image:: images/setup/ownCloud-remove_existing_connection.png
+
+To do so, in the client UI, which you can see above, click the "**Account**" drop-down menu and then click "Remove". 
+This will display a "**Confirm Account Removal**" dialog window.
+
+image:: images/setup/ownCloud-remove_existing_connection_confirmation_dialog.png
+
+If you're sure, click "**Remove connection**".
+
```
phil-davis: At this point you can actually move/copy the old local sync folder (which has all your files locally) to the new place that you want it. Then when you "Add new" sync connection, choose "keep existing data". 
ownCloud client happily checks the files in the newly-added sync folder and finds that they match with what is on the server.
The alternative is to start with an empty new sync folder - which means that the ownCloud client will have to download your files again. That may be many GB that is already in the old sync folder on your system.
I haven't tried, but I guess copying over from the old local folder will also preserve the selection of folders to be ignored... since the "hidden" oC database will come across from the old local folder.

settermjd: Good point. I was only testing with a minimal amount of data. I'll update the PR with your feedback.

settermjd: I've added an admonition to stress the implications of starting a clean sync. Thanks for pointing that out.

PR #5590, opened 2017-03-08
templates > .github and create release_template

michaelstingl:
I would like to have the release steps under version control to constantly improve the process. And we could put all templates to the .github folder same as it's in the core repo.

@guruz @ogoffart @SamuAlfageme What do you think?

**Open tasks:**

- [ ] Remove redundant wiki page [owncloud/client/Release Checklist Template](https://github.com/owncloud/client/wiki/Release%20Checklist%20Template)
- [ ] Merge with "Squash and merge" option


ogoffart:
There are many links to the issue_template.md that points to it in the root location.
So we must move that file carefully. For example there is a link in CONTRIBUTING.md that needs to be updated.
There might be more  (there are links in many issue but i gues they need not to be changed)

michaelstingl:
@ogoffart Good point. I could copy the file for the start to make migration smoother.

guruz:
I wanted to revise the release template a bit. Shall I do this before or after this is merged? Maybe after is better..

michaelstingl:
@guruz Both fine for me. You can update the wiki page, and I will then update the PR…

PR #5591, opened 2017-03-09
Propagator: Attempt to fix a crash in OwncloudPropagator::scheduleNex…

ogoffart:
…tJob

The crash reporter shows many crashes in OwncloudPropagator::scheduleNextJob.
We don't really know what could be the cause, but it's probably because
the _activeJobList contains dangling pointer.
So this patch makes sure to remove all the jobs from this list as they get
destroyed.

### DIFF
--- a/src/libsync/owncloudpropagator.cpp
+++ b/src/libsync/owncloudpropagator.cpp
@@ -94,6 +94,16 @@ int OwncloudPropagator::hardMaximumActiveJob()
     return max;
 }
 
+PropagateItemJob::~PropagateItemJob()
+{
+    if (auto p = propagator()) {
+        // Normally, every job should clean itself from the _activeJobList. So this should not be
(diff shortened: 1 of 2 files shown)


src/libsync/owncloudpropagator.cpp
```diff
@@ -94,6 +94,16 @@ int OwncloudPropagator::hardMaximumActiveJob()
     return max;
 }
 
+PropagateItemJob::~PropagateItemJob()
+{
+    if (auto p = propagator()) {
```
jturcotte: Technically PropagatorJob::propagator could use a static_cast and this null-check wouldn't be necessary since `PropagatorJob::PropagatorJob` takes an `OwncloudPropagator *propagator`. Just a nitpick, looks good anyhow.

ogoffart: Well, it's still undefined behavior to static_cast to the Propagator if we are called from the QObject destructor.

Issue #5592, opened 2017-03-09
OpenSSL version: Output the actually loaded one, not the build version


t-b:
### Expected behaviour
owncloud client should use the latest version of openssl available on the distribution.
On debian jessie this is currently `1.0.1t-1+deb8u6`.

Maybe it is only an issue of compile time openssl version vs. runtime openssl version. But even compiling against an old version is not the right thing to do.
### Actual behaviour
Owncloud client tells me

```
Gebaut von der GIT-Revision 832599 auf Mar 3 2017, 09:28:09 verwendet Qt 4.8.6, OpenSSL 1.0.1k 8 Jan 2015
```
### Client configuration
Client version: 2.3

Operating system:
debian jessie
OS language:
de

Installation path of client:
Packages from https://software.opensuse.org/download/package?project=isv:ownCloud:desktop&package=owncloud-client

```
2.3.0-1.1 (/var/lib/apt/lists/download.opensuse.org_repositories_isv:_ownCloud:_desktop_Debian%5f8.0_Packages) (/var/lib/dpkg/status)
 Description Language: 
                 File: /var/lib/apt/lists/download.opensuse.org_repositories_isv:_ownCloud:_desktop_Debian%5f8.0_Packages
                  MD5: 3ce97b207e6d1f58416ab77929104f12
```

ogoffart:
Yes, that string is using the build-time version.

ogoffart:
We can use QSslSocket::sslLibraryVersionString,  but that's only working with Qt 5, so it would not help in this case

ckamm:
Note that the string says "built from ..." and then lists the compile-time QT and SSL versions. How useful are the build-versions vs runtime-versions? Currently `--version` shows the run-time Qt version.

ckamm:
Adding more info to --version: https://github.com/owncloud/client/pull/5598

t-b:
Why does the build daemon uses an outdated openssl version full of bugs?

ckamm:
@ogoffart I'd expect that the runtime Qt and SSL versions are more useful, if you agree I'd adjust it.

Also, do you know whom to ask regarding the build daemon's openssl version?

ckamm:
Pull request changing what is being shown: https://github.com/owncloud/client/pull/5619

ckamm:
@ogoffart @guruz The remaining questing is: is building against an outdated version of openssl a problem? How do we update the build machine's version?

ogoffart:
[...]

PR #5598, opened 2017-03-10
Add more versions to --version #5592

ckamm:
* Add git revision, if available
* Add SSL runtime/compiletime version
* Add compiletime Qt version

### DIFF
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -37,6 +37,7 @@
 #include "updater/ocupdater.h"
 #include "excludedfiles.h"
 #include "owncloudsetupwizard.h"
+#include "version.h"
 
 #include "config.h"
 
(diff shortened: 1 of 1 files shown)


src/gui/application.cpp
```diff
@@ -508,7 +511,13 @@ void Application::showVersion()
     stream << _theme->appName().toLatin1().constData()
            << QLatin1String(" version ")
            << _theme->version().toLatin1().constData() << endl;
-    stream << "Using Qt " << qVersion() << endl;
+#ifdef GIT_SHA1
+    stream << "Git revision " << GIT_SHA1 << endl;
+#endif
+    stream << "Using Qt " << qVersion() << ", built against Qt " << QT_VERSION_STR << endl;
+#if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
+    stream << "Using '" << QSslSocket::sslLibraryVersionString() << "', built against '" << OPENSSL_VERSION_TEXT << "'" << endl;
```
ogoffart: the version of openssl we are build against does not mean any sense.
Maybe what makes sens is the version Qt is built against.
So i'd leave it out.

Issue #5605, opened 2017-03-13
Update the app passwords url


ckamm:
See https://github.com/owncloud/core/issues/27360 - it is now `/index.php/settings/personal?sectionid=security#apppasswords`

Found by @SamuAlfageme

ckamm:
Pull request: https://github.com/owncloud/client/pull/5606

SamuAlfageme:
👍

PR #5606, opened 2017-03-13
HttpCreds: Update app passwords url #5605

ckamm:
See also owncloud/core#27360

Create Account::makeServerVersion() along the way, to avoid
version literals like 0x0A0000.

### DIFF
--- a/src/gui/creds/httpcredentialsgui.cpp
+++ b/src/gui/creds/httpcredentialsgui.cpp
@@ -75,7 +75,12 @@ QString HttpCredentialsGui::requestAppPasswordText(const Account* account)
         return QString();
     }
 
-    QString path = QLatin1String("/index.php/settings/personal?section=apppasswords");
+    QString path;
+    if (account->serverVersionInt() < Account::makeServerVersion(10, 0, 0)) {
+        path = QLatin1String("/index.php/settings/personal?section=apppasswords");
+    } else {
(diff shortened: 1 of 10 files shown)


src/gui/creds/httpcredentialsgui.cpp
```diff
@@ -71,12 +71,18 @@ void HttpCredentialsGui::askFromUserAsync()
 QString HttpCredentialsGui::requestAppPasswordText(const Account* account)
 {
     if (account->serverVersionInt() < 0x090100) {
```
ogoffart: while you are at it, you can use makeServerVersion there too, then.

ckamm:
@ogoffart I've tested againt 10.0.0 and the URL works. I'll make the other changes you suggest, maybe replacing our hex-version literals everywhere.

ckamm:
@ogoffart Split into two commits, got rid of hex version literals, tested not-logged-in case: works. Approve?

ogoffart:
So you tested with a version that were not logged, and after ligging, the browser show the page at the right location?

ckamm:
@ogoffart Yes, first I get redirected to `index.php/login?redirect_url=%252Findex.php%252Fsettings%252Fpersonal%253Fsectionid%253Dsecurity#apppasswords` and then finally to `index.php/settings/personal?sectionid=security#apppasswords` which is good.

ogoffart:
:+1:

PR #5617, opened 2017-03-16
SyncEngine: fix test regression in TestChunkingNG::testCreateConflict…

ogoffart:
…WhileSyncing

In 8ef11a38c99c6a33b175f1ad792ec02b52f88d25, we started blacklisting
SoftError for 0 seconds.  But if the two sync happen with less than
1s interval, we would still prevent them to happen.
So make sure we expire if 0 seconds have expired


ckamm:
I'd weakly prefer if we'd have a separate case for `ignoreDuration == 0` to make explicit that this error wasn't ever blacklisted, just tracked. (i.e. also skip the qDebug there). But it's okay to merge as-is too.

PR #5619, opened 2017-03-16
About: Show the runtime versions of Qt and OpenSSL #5592

ckamm:
I didn't want to touch this visible translated string in a
patch release so I just switched the data.

For Qt versions before 5, we continue to show the build-time version
of OpenSSL but it's clearly marked as such.


ogoffart:
:+1:

(The version against which Qt is built is also important for the cases where we have ifdef. but less important than the running one)

PR #5623, opened 2017-03-16
Folder::showSyncResultPopup: Fix undefined behavior when there is no …

ogoffart:
…errors

When there is no errors  _syncResult.firstItemError() is NULL, and accessing
it's _file member is an undefined behavior. (Thankfully, createGuiLog did not
use the string when the count was 0, but we are not supposed to create
null references.

Found with the UB sanitizer:

src/gui/folder.cpp:348:49: runtime error: member access within null pointer of type 'OCC::SyncFileItem'
src/gui/folder.cpp:348:19: runtime error: reference binding to null pointer of type 'const QString'


ogoffart:
I ran ownCloud with -fsanitize=undefined,  and this is the only issue it detects. I'm kind of surprised there were no more problems.

PR #5625, opened 2017-03-17
Including some info on the proxy config. in the docs.

SamuAlfageme:
This was a request of @michaelstingl to document how to configure the proxy manually via the .cfg file for config provisioning purposes.
- Restructured the bullet points into 3 different tables; I believe it's a cleaner approach; maybe this is also applicable to other sections.
- Would be logical to include also settings on proxy auth?

It does look like this:

screenshot 2017-03-17 13 21 39

@settermjd maybe some changes would make the tables look better; what do you think? 🤔


ogoffart:
Merged manually in 2.3

ogoffart:
(commit d375e9b86cde0a847cbec88e7818eeb7292c483e )

PR #5636, opened 2017-03-20
Fix incorrect documentation for the ownCloud command-line client

### LINKED ISSUE #5609: [owncloudcmd] the command-line client ignores the keychain entries
This is a follow-up of the thread started on [central](https://central.owncloud.org): https://central.owncloud.org/t/owncloudcmd-on-macos-does-not-authenticate-automatically/6349 about a potential regression or just outdated docs. 

### Expected behavior
According to the ["*Credential handling*" section of the docs](https://doc.owncloud.org/desktop/2.3/advancedusage.html#credential-handling):

> `owncloudcmd` uses the credentials of the GUI synchronization client. If no client is configured

### Actual behavior
The client always asks for user/password combination, regardless of what was set on the GUI / present on the keychain.

### Steps to reproduce
1. Set up an account on the GUI
2. Try to sync the same account URL(*) from the command-line client.

*: *as after multi-account support, it's the only way to indicate the client which account we are referring to*

settermjd:
This fixes #5609.

### DIFF
--- a/doc/owncloudcmd.rst
+++ b/doc/owncloudcmd.rst
@@ -60,10 +60,9 @@ Other command line switches supported by ``owncloudcmd`` include the following:
 Credential Handling
 ~~~~~~~~~~~~~~~~~~~
 
-``owncloudcmd`` uses the credentials of the GUI synchronization client.
-If no client is configured, or if you choose to use a different user to synchronize,
-you can specify the user
-password setting with the usual URL pattern.  For example::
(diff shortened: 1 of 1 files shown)


ogoffart:
Thanks for the fix.

Manually merged in the 2.3 branch. (commit 9840ae9e9671ea6080e9aeaf9f39febbad796b53 )

(Pull request for the documentation can be done against the 2.3 branch)

Issue #5597, opened 2017-03-10
Wizard: Create a new directory name, don't always use "ownCloud"


guruz:
Only use "ownCloud" (or whatever $THEMENAME) for a new install, for existing installs we want users consciously decide that they really want to sync into an existing directory
(Especially now where we allow to sync same directory to multiple servers)

phil-davis:
I agree, it could be too easy to accidentally put multiple sync connections mapped into the same local folder.
It would be possible to have logic so that if the "owncloud" folder is not yet used by any sync connection then it is made the default. But actually if someone already has 1 ormore sync connections that do not use the "owncloud" local folder, then probably when they ad another sync connection they likely do not want to use the (unused) "owncloud" folder anyway.

ckamm:
@guruz @phil-davis This applies to the folder wizard and the account wizard. Should we default to offer syncing to a previously non-existing folder? E.g. owncloud/, owncloud2/, owncloud3/, ... whatever doesn't exist yet.

phil-davis:
I would find a rule like this to be reasonable:
1) If the existing accounts are all synced in the "default" home folder owncloud/, owncloud2/,... then offer the next owncloud_n_/ as the efault for a new sync connection.
2) Otherwise (i.e. if there is any sync connection that the was explicitly chosen non-default in the past by the user) offer no default - so the user cannot "accidentally" click through and end up with what the (probably) do not want.

But such a rule might be a bit surprising to users who notice and ask about why a default was or wasn't presented.

guruz:
Could even have ownCloud_$PARTOFHOSTNAME like ownCloud_UniversityOfBiesdorf or ownCloud_philscloud automagically

phil-davis:
That sounds good. It might be a bit tricky depending on the domain name of the cloud instance.
1) Probably want to have the username in there also? Keep dots in usernames like "phil.davis"?
2) Some people will have a domain like davis.mycloudhoster.com and so you can grab the "davis" part. But others will have cloud.universityofxyz.edu and if you just grab "cloud" then that is of little use.

and if you take the whole thing (and maybe replace dots with "-" or "_"?) then you get long things like:
ownCloud_davis-mycloudhoster-com_phil-davis

Hmmm - is there a nice algorithm possible here?

ckamm:
I'll take care of the first incremental step: Never default to a folder that is already in use. 

[...]

Issue #3632, opened 2015-08-15
Add account wizard "choose what to sync" option also syncs the root without choice


icawebdesign:
When selecting a directory to sync, the client also downloads all files in the root directory too without an option to change this behaviour. This can result in a lot of files and MB being downloaded and sync'd that aren't required.

This is on Windows 8.1 Pro.

ownCloud client 1.8.x works as expected adn will only sync the directories selected.

icawebdesign:
Upon further investigation, this isn't restricted to the root path, but in fact a selected directories _parent_ dir, for example:

I want to sync 'baz':
/foo/bar/baz

Everything in 'baz' gets sync'd as expected, but also everything in 'bar' will too, which is undesired.

ogoffart:
There is no differences with 1.8.x.  Could you elaborate on what is the expected behaviour and what do you use to select directories.

MTRichards:
If I am understanding correctly, this is an issue where selective sync pulls in all the files in the parent directory of the selected folder, rather than just the folder itself.

So if I selectively sync only /root/pictures the sync algorithm pulls in all of the files in root, and also all of the files in pictures. It does not pull in any other directories in /root/.

If this is the comment, it is the designed behavior in ownCloud. We follow the same approach as Dropbox in this case, where the files in all parent folders of a selected directory down in the file tree are synced.

icawebdesign:
@MTRichards that's correct in what I'm seeing (apologies, not had time to follow up with screenshots as I had planned today).

This is different behaviour to 1.8.x. Is there any particular reason for this? Using your example of /root/pictures, if I wanted /root files, I'd check the /root box and uncheck the pictures box if I only wanted the files in the root dir sync'd. By selecting /root/pictures, surely this indicates that I want the pictures directory sync'd and the mere fact I haven't specifically selected the root dir, should indicate that I'm not interested in the contents of that.

If you're going for a dropbox replacement (I've never used any other cloud service for the very reason owncloud exists, I value the fact I can host it in-house but I understand where you're coming from in this respect), could this possibly be an option somewhere to enable / disable this feature?

For example, I wanted to sync some ebooks (~5GB in a 'dev' directory) but I ended up syncing nearer 20GB as there was some "uncategorised" books in the parent directory that I had no interest in syncing with my laptop.

[...]

PR #5638, opened 2017-03-21
Improve folder selection in wizards

ckamm:
Primarily address #5597: Don't propose existing folders as sync folders in wizards to not cause accidents.

### DIFF
--- a/src/gui/folderman.h
+++ b/src/gui/folderman.h
@@ -124,11 +124,24 @@ public:
      * Check if @a path is a valid path for a new folder considering the already sync'ed items.
      * Make sure that this folder, or any subfolder is not sync'ed already.
      *
+     * Note that different accounts are allowed to sync to the same folder.
+     *
      * \a forNewDirectory is internal and is used for recursion.
      *
      * @returns an empty string if it is allowed, or an error if it is not allowed
      */
-    QString checkPathValidityForNewFolder(const QString &path, const QUrl& serverUrl = QUrl(), bool forNewDirectory = false);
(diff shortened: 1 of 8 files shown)


SamuAlfageme:
Note to self; check what happens with the finder integration that creates a sync folder shortcut in the left bar when the default folder name was used.

SamuAlfageme:
@ckamm also one of my wishes for a future complete folder selection _experience_ in the account wizard would be https://github.com/owncloud/client/issues/3632 😇

What do you think?

ckamm:
@SamuAlfageme That is indeed occasionally confusing users, but would need a significant account creation wizard redesign. Currently the ability to only sync a remote subfolder is not being exposed in the folder setup part of the account creation wizard.

src/gui/folderman.h
```diff
@@ -124,11 +124,24 @@ class FolderMan : public QObject
...
+    /**
+     * Attempts to find a non-existing, acceptable path for creating a new sync folder.
+     *
+     * Uses \a basePath as the baseline. It'll return this path if it's acceptable.
+     *
+     * Note that this can fail. If someone syncs ~ and \a basePath is ~/ownCloud, no
+     * subfolder of ~ would be a good candidate. When that happens \a ok will be
+     * false and \a basePath is returned.
+     */
+    QString findGoodPathForNewSyncFolder(const QString &basePath, const QUrl &serverUrl, bool *ok = 0) const;
```
ogoffart: if ok is not used, maybe we should leave it out.

ckamm: True

ogoffart:
:+1:  
Maybe write some tests.

ogoffart:
:+1:

Issue #5642, opened 2017-03-22
The "can edit" checkbox in the share dialog should be tristate


ckamm:
Currently it is checked if *any* of the sub-permissions is granted. This can hide the fact that only a limited set of edit permissions were granted to a user.

Instead, the "can edit" checkbox should be partially checked if only some of the edit permissions were granted.

SamuAlfageme:
👍 now it's even better than the webUI 😎

PR #5643, opened 2017-03-22
ShareDialog: Make "can edit" partially checked sometimes #5642

ckamm:


ckamm:
@ogoffart Good point. Expanded commit message and merged.

PR #5646, opened 2017-03-24
Improve http error messages; cleanup

ckamm:
By default QNetworkReply::errorString() often produces messages like
   `"Error downloading <url> - server replied: <reason>"`
but the "downloading" part invariably confuses people since the
error might very well have been produced by a PUT request.

This commit produces clearer error messages for HTTP errors.

Additionally:
* Remove some unnecessary null checks from slots connected to
  network job signals and document that these signals never send
  null replies.
* There was a bug where AbstractNetworkJob::_timedout wasn't
  set when derived classes overrode slotTimeout. We now ensure
  it's always set by disallowing overrides of slotTimeout.
  Instead it now calls onTimedOut, which allows custom handling.
* Several subclasses declared errorString, isTimedOut. Move
  these to AbstractNetworkJob.
* Unify handling of OC-ErrorString (via the new, general
  Job::errorString)
* Add documentation in various places.

### DIFF
--- a/src/libsync/abstractnetworkjob.cpp
+++ b/src/libsync/abstractnetworkjob.cpp
@@ -146,37 +146,16 @@ QUrl AbstractNetworkJob::makeDavUrl(const QString& relativePath) const
     return Utility::concatUrlPath(_account->davUrl(), relativePath);
 }
 
-QByteArray AbstractNetworkJob::requestVerb(QNetworkReply* reply)
-{
-    switch (reply->operation()) {
-    case QNetworkAccessManager::HeadOperation:
-        return "HEAD";
-    case QNetworkAccessManager::GetOperation:
-        return "GET";
-    case QNetworkAccessManager::PutOperation:
-        return "PUT";
-    case QNetworkAccessManager::PostOperation:
-        return "POST";
-    case QNetworkAccessManager::DeleteOperation:
-        return "DELETE";
-    case QNetworkAccessManager::CustomOperation:
-        return reply->request().attribute(QNetworkRequest::CustomVerbAttribute).toByteArray();
-    case QNetworkAccessManager::UnknownOperation:
-        break;
-    }
-    return QByteArray();
-}
-
 void AbstractNetworkJob::slotFinished()
 {
     _timer.stop();
 
     if( _reply->error() == QNetworkReply::SslHandshakeFailedError ) {
-        qDebug() << "SslHandshakeFailedError: " << reply()->errorString() << " : can be caused by a webserver wanting SSL client certificates";
+        qDebug() << "SslHandshakeFailedError: " << errorString() << " : can be caused by a webserver wanting SSL client certificates";
     }
 
     if( _reply->error() != QNetworkReply::NoError ) {
-        qDebug() << Q_FUNC_INFO << _reply->error() << _reply->errorString()
+        qDebug() << Q_FUNC_INFO << _reply->error() << errorString()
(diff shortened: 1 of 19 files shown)


src/libsync/abstractnetworkjob.cpp
```diff
@@ -315,4 +332,48 @@ QString errorMessage(const QString& baseError, const QByteArray& body)
...
+{
+    QString base = reply.errorString();
+    int httpStatus = reply.attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
+    QString httpReason = reply.attribute(QNetworkRequest::HttpReasonPhraseAttribute).toString();
+
+    // Only adjust HTTP error messages of the expected format.
+    if (httpReason.isEmpty() || httpStatus == 0 || !base.contains(httpReason)) {
+        return base;
+    }
+
```
guruz: FYI https://code.qt.io/cgit/qt/qtbase.git/commit/?id=4efa50a5e37497bc3a3dd30f27db7d8100e3f6a5

guruz: (Also note that we'll have Qt 5.6.2 also on Linux soon because we will ship with it)

src/libsync/abstractnetworkjob.cpp
```diff
@@ -315,4 +332,48 @@ QString errorMessage(const QString& baseError, const QByteArray& body)
...
+    return AbstractNetworkJob::tr("Server replied \"%1 %2\" to \"%3 %4\"").arg(
+            QString::number(httpStatus),
+            httpReason,
+            requestVerb(reply),
+#if QT_VERSION < QT_VERSION_CHECK(5, 0, 0)
+            reply.request().url().toString()
+#else
+            reply.request().url().toDisplayString()
```
ogoffart: Do we need the URL in the error string?  The file will anyway be displayed in the UI.  
If you're overwriting Qt's error message, i'd say we should remove the URL from it.

ckamm: @ogoffart For the error list in the ui we don't need the URL, but it might be handy for errors of other network jobs. I had similar doubts and that's why I put it at the end, but would err on the side of including it.

src/libsync/abstractnetworkjob.cpp
ckamm: @guruz Thanks for the update on the change to Qt, that would already help a bit. This code will not continue to function correctly with that Qt version and I think applying it still makes sense.

guruz: Can you clarify this comment? Is this good to merge or not?

ckamm: @guruz  This commit will continue to work (i.e. replace this error message) even if the Qt error message changes in the way described in https://code.qt.io/cgit/qt/qtbase.git/commit/?id=4efa50a5e37497bc3a3dd30f27db7d8100e3f6a5. I would merge it.

guruz:
```

hudson.plugins.git.GitException: Command "git merge 68f0f1b404485ffc4f5127784eb4ff52919953ad" returned status code 1:
stdout: Auto-merging src/libsync/propagateuploadng.cpp
Auto-merging src/libsync/propagateupload.h
CONFLICT (content): Merge conflict in src/libsync/propagateupload.h
Auto-merging src/libsync/propagateupload.cpp
Automatic merge failed; fix conflicts and then commit the result.
```

Issue #2153, opened 2014-09-03
unix man pages are not fully branded


jnweiger:
genbranding.pl renames the man pages, but the contents always speaks of OWNCLOUD, ownCloud, owncloud. No templates apply.

hodyroff:
oh mighty. @michaelstingl 
Honestly I would just have an option to delete those man pages if so desired, but I don't know if thats more difficult then putting $name or something into them ... there are some borders to branding maybe ;)

michaelstingl:
@guruz Is it easy to add the branding variables here?

guruz:
@michaelstingl afaik you need to replace it with a $FILENAME.in (or .cmake) file and then add a link in the CMakeFiles.txt to have cmake process the file when copying it.
You can look at how it is done for admin/osx/post_install.sh.cmake

michaelstingl:
Who can fix it? @SamuAlfageme ?

SamuAlfageme:
Yes, I can make a small patch to use the `$APPLICATION_EXECUTABLE`[`cmd`] in the manpages, and expand it with Sphinx, it should be that easy. 

But my real concern is that these are currently not being copied/linked to `/usr/[local/]share/man/man1/` in any [post]installation script (or at least I can't see where; also tried installing owncloud through APT in ubuntu and no luck with it).

btw, the pages seem to be a bit outdated/incomplete: 

* [**`owncloud(1)`**](https://github.com/owncloud/client/blob/master/doc/owncloud.1.rst)
* [**`owncloudcmd(1)`**](https://github.com/owncloud/client/blob/master/doc/owncloudcmd.1.rst)

guruz:
I don't even think you need Sphinx, you can do it analog to what I described above: https://github.com/owncloud/client/issues/2153#issuecomment-265714410

hodyroff:
If you like we can also get rid of the man pages completely ... but lets decide on one direction and get it done for 2.4.

jnweiger:
Manpages to be removed ASAP

PR #5649, opened 2017-03-24
doc/CMakeLists.txt no more doc-man

jnweiger:
remove doc-man target in doc/CMakeLists.txt https://github.com/owncloud/client/issues/2153


jnweiger:
@guruz kill man pages with fire. 
they aren't used in mac, are they?

PR #5659, opened 2017-03-28
HTTP/2

ogoffart:
We need Qt 5.9 for HTTP2 because, even if Qt 5.8 already has support
for it, there is some critical bug in the HTTP2 implementation which
make it unusable [ https://codereview.qt-project.org/186050 and
https://codereview.qt-project.org/186066 ]

When using HTTP2, we can use many more parallel network request, this
is especially good for small file handling

### DIFF
--- a/src/libsync/accessmanager.cpp
+++ b/src/libsync/accessmanager.cpp
@@ -94,8 +94,11 @@ QNetworkReply *AccessManager::createRequest(QNetworkAccessManager::Operation op,
     newRequest.setRawHeader("X-Request-ID", requestId);
 
 #if QT_VERSION >= QT_VERSION_CHECK(5, 9, 0)
-    // only enable HTTP2 with Qt 5.9 because Qt 5.8.0 has too many bugs (only use one connection if the server does not support HTTP2)
-    newRequest.setAttribute(QNetworkRequest::HTTP2AllowedAttribute, true);
+    // only enable HTTP2 with Qt 5.9 because Qt 5.8.0 has too many bugs
+    // (only use one connection if the server does not support HTTP2)
+    if (newRequest.url().scheme() == "https") { // Not for "http": QTBUG-61397
+        newRequest.setAttribute(QNetworkRequest::HTTP2AllowedAttribute, true);
+    }
 #endif
 
     return QNetworkAccessManager::createRequest(op, newRequest, outgoingData);
--- a/src/libsync/accessmanager.cpp
+++ b/src/libsync/accessmanager.cpp
@@ -93,6 +93,11 @@ QNetworkReply *AccessManager::createRequest(QNetworkAccessManager::Operation op,
     qInfo(lcAccessManager) << op << verb << newRequest.url().toString() << "has X-Request-ID" << requestId;
     newRequest.setRawHeader("X-Request-ID", requestId);
 
+#if QT_VERSION >= QT_VERSION_CHECK(5, 9, 0)
+    // only enable HTTP2 with Qt 5.9 because Qt 5.8.0 has too many bugs (only use one connection if the server does not support HTTP2)
+    newRequest.setAttribute(QNetworkRequest::HTTP2AllowedAttribute, true);
+#endif
+
     return QNetworkAccessManager::createRequest(op, newRequest, outgoingData);
 }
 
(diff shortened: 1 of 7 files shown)


ogoffart:
I have not tried this on any real http2 server yet.  We anyway will have to wait for Qt 5.9
I'm just creating this pull request for reference.

ogoffart:
Another problem i foresee is that this might cause the sycning "long-running" job to make the other short lived control job queued and timeout. I'm thinking about the quota check for the UI, or the connection validator, or the job to expend the selective sync  view.

Currently, without HTTP2, we only limit the transfer of long running (big PUT or GET) jobs to 3. But Qt keeps 6 TCP connection open. So the short selective sync jobs happen in parallel and are still finishing well on time.

However, with HTTP2, there is only one connection and the big job might take the whole bandwidth of this connection, meaning that the smaller jobs would be queued and will not finish in time to have a responsible UI.

The solution would be to use higher priority for such jobs. But the Qt API does not support setting a priority on the request, as far as i know.

guruz:
As mentioned on IRC, we need to set maintenance jobs to HighPriority (QNetworkRequest property).

guruz:
We talked to the developer of HTTP2 in Qt, he did not check if certain PUT requests can stall each other, e.g. one request gets all the send window and other requests don't progress.
This would be bad since our timeout handling would trigger for the non-sending PUTs

src/libsync/owncloudpropagator.cpp
```diff
@@ -80,18 +80,18 @@ int OwncloudPropagator::maximumActiveTransferJob()
...
     static int max = qgetenv("OWNCLOUD_MAX_PARALLEL").toUInt();
-    if (!max) {
-        max = 6; //default (Qt cannot do more anyway)
-        // TODO: increase this number when using HTTP2
-    }
-    return max;
+    if (max)
+        return max;
+    if (_account->isHttp2Supported())
+        return 20;
```
ogoffart: That number is for smaller jobs (mkcol, small uploads, small downloads, move, delete)

the maxium amount of longer jobs is still 3

mrow4a: But anyways, it might need some tests probably.

ogoffart:
Using Qt 5.9 beta, I did some basic tests against an HTTP/2 enabled server, and everything seemed to work.
I did not do any performence testing tough.

mrow4a:
I will happily test it again looking on performance if we would decide to integrate http2 in the client officialy @labkode @moscicki @DeepDiver1975

ogoffart:
@mrow4a   Yes, we integrate HTTP2 in the client officially, and this pull request does it.

Now that Qt 5.9 is released, we could already integrate this patch.

guruz:
Was https://github.com/owncloud/client/pull/5659#issuecomment-291487229 adressed?

ogoffart:
> Was #5659 (comment) adressed?

We set a low priority on the PUT and GET jobs:

```
req.setPriority(QNetworkRequest::LowPriority);
```

ogoffart:
[Conflicts resolved]

Any objections merging this to master now?

src/libsync/accessmanager.cpp
```diff
@@ -79,6 +79,15 @@ QNetworkReply *AccessManager::createRequest(QNetworkAccessManager::Operation op,
     if (verb == "PROPFIND") {
         newRequest.setHeader(QNetworkRequest::ContentTypeHeader, QLatin1String("text/xml; charset=utf-8"));
     }
+
+#if QT_VERSION >= QT_VERSION_CHECK(5, 9, 0)
```
guruz: wasn't it 5.9.1 with all necessary bugfixes?

src/libsync/owncloudpropagator.cpp
guruz: 20 sounds scary :-) let's see if the servers can handle it.. @DeepDiver1975 @PVince81 @butonic

SamuAlfageme:
@ogoffart are the upload/download propagator low priorities introduced here also applicable for non-HTTP/2 requests?

No questions/objections apart from that - just tested it on macOS with Qt 5.9.1 and everything seem to be smooth. 

We might want to start setting a plan to have a stable Qt 5.9 toolchain on the build system.

src/libsync/accessmanager.cpp
ogoffart: There are necessary bugfixes in 5.9.1.

ogoffart:
@SamuAlfageme   Yes, this also apply to Non-HTTP/2 requests, that should not change much actually for we never have more than 6 requests at the same time normally (and if we have, this is actually a good change)

Issue #5515, opened 2017-02-05
sharing client - ubuntu - dialogue in background


herrdeh:
When opening the sync client GUI, it will not appear in the foreground, but send a message: "Sync client is ready". Clicking this will bring the GUI into the foreground.
The same happens with the "share with owncloud..." dialogue from a file manager.

Normally, a programme will launch in the foreground.

ubuntu 16.10, gnome, the same with debian 8.6
OC client 2.2.4

SamuAlfageme:
@herrdeh

> When opening the sync client GUI, it will not appear in the foreground, but send a message: "Sync client is ready". Clicking this will bring the GUI into the foreground.

This is expected behavior if you don't have anything to configure (i.e. if the connection with your server was set and successful). Since it's running smooth already you don't need co configure anything on the client GUI.

>  The same happens with the "share with owncloud..." dialogue from a file manager.

This however should not happen, is the dialogue popping behind the file manager?

herrdeh:
> This is expected behavior if you don't have anything to configure (i.e. if the connection with your server was set and successful). Since it's running smooth already you don't need co configure anything on the client GUI.

Yes. But I'd not open the GUI when everything is running smoothly. Only in the case adjustments are required, errors occur etc - then I'd open the GUI, but then I'd like to have it in the foreground immediately. 

> is the dialogue popping behind the file manager?

Yes, it does so; but I found very rare exeptions.

guruz:
Let's try to send the sharing dialogue more to foreground, I know @ogoffart @SamuAlfageme had been trying something last week.

ogoffart:
The way to open the "GUI" is by clicking on the systemtray icon.

With plasma, the sharing dialog always appear on the top.

ckamm:
I've tested this on linux with the account wizard and several share dialogs open. It worked fine for me.

Issue #5566, opened 2017-02-27
[Windows 10] [2.2.4] New account wizard gets positioned under all other Windows


guruz:
This report is when adding a second account and having other windows programs windows open.
For me the new wizard was then hidden.

FYI @SamuAlfageme @ogoffart

guruz:
... now I'm thinking this might have been fixed in 2.3 already

ckamm:
@guruz I cannot reproduce it on Windows 10. When I click the "Add account" menu entry, the account settings window hides and the new account wizard pops up in its place and is properly covering the other windows.

guruz:
This is fixed with my patch above..

PR #5664, opened 2017-03-29
Share Dialog: Bring to top on tray click #5515

guruz:
On my OS X, it might get hidden under other apps while I opened it and then want to quickly
verify something in another app.

### DIFF
--- a/src/gui/owncloudgui.cpp
+++ b/src/gui/owncloudgui.cpp
@@ -174,22 +174,31 @@ void ownCloudGui::slotTrayClicked( QSystemTrayIcon::ActivationReason reason )
         last_click.start();
     }
 
-    // A click on the tray icon should only open the status window on Win and
-    // Linux, not on Mac. They want a menu entry.
-#if !defined Q_OS_MAC
+    // Left click
     if( reason == QSystemTrayIcon::Trigger ) {
-        // Start settings if config is existing.
-        slotOpenSettingsDialog();
-    }
+        if (OwncloudSetupWizard::bringWizardToFrontIfVisible()) {
+            // brought wizard to front
+        } else if (_shareDialogs.size() > 0) {
+            // Share dialog(s) be hidden by other apps, bring them back
+            Q_FOREACH(const QPointer<ShareDialog> &shareDialog, _shareDialogs) {
+                Q_ASSERT(shareDialog.data());
+                raiseDialog(shareDialog);
+            }
+        } else {
+#ifdef Q_OS_MAC
+            // on macOS, a left click always opens menu.
+            // However if the settings dialog is already visible but hidden
+            // by other applications, this will bring it to the front.
+            if (!_settingsDialog.isNull() && _settingsDialog->isVisible()) {
+                raiseDialog(_settingsDialog.data());
+            }
 #else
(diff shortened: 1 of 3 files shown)


src/gui/owncloudgui.cpp
```diff
@@ -190,6 +190,12 @@ void ownCloudGui::slotTrayClicked( QSystemTrayIcon::ActivationReason reason )
         }
     }
 #endif
+
+    // Might be hidden by other apps, bring them back
+    Q_FOREACH(const QPointer<ShareDialog> shareDialog, _shareDialogs) {
```
ogoffart: missing a '&'

ogoffart:
I guess if there is a share dialog, then we should not raise the settings at the same time.

Which means the user can't go to the settings while using a share dialog.

ogoffart:
Do we really want such changes in 2.3 branch?  maybe....

guruz:
> Also, I tested on linux, and when the account settings window is previously hidden, it pops up on top of the share dialog...

If you look at the existing surrounding code, it's a bit confusing. Why do we do different things for macOS and non-macOS. (the patch was even by me..meh)

guruz:
> Why Q_FOREACH?

Because we used it somewhere else too. Am I old-fashioned?

> Also, I tested on linux, and when the account settings window is previously hidden, it pops up on top of the share dialog...

The linux code path is different (see around patch), I can have a look and unify the code base.

> I guess if there is a share dialog, then we should not raise the settings at the same time.

ack

guruz:
@ckamm Should be better now?

guruz:
Or maybe not, I wanted to do https://github.com/owncloud/client/issues/5566 too.. sec

guruz:
Done

src/gui/owncloudgui.cpp
```diff
@@ -174,22 +174,31 @@ void ownCloudGui::slotTrayClicked( QSystemTrayIcon::ActivationReason reason )
...
     if( reason == QSystemTrayIcon::Trigger ) {
-        // Start settings if config is existing.
-        slotOpenSettingsDialog();
-    }
+        if (OwncloudSetupWizard::bringWizardToFrontIfVisible()) {
+            // brought wizard to front
+        } else if (_shareDialogs.size() > 0) {
+            // Share dialog(s) be hidden by other apps, bring them back
+            Q_FOREACH(const QPointer<ShareDialog> &shareDialog, _shareDialogs) {
+                Q_ASSERT(shareDialog.data());
```
ckamm: Minor: I still think Q_FOREACH is awkward and, unrelatedly, wonder why _shareDialogs has QPointer<...> values, since deleted dialogs are removed from the map anyway...

src/gui/owncloudsetupwizard.cpp
```diff
@@ -64,10 +65,10 @@ OwncloudSetupWizard::~OwncloudSetupWizard()
     _ocWizard->deleteLater();
 }
 
+static QPointer<OwncloudSetupWizard> wiz = 0;
```
ckamm: Minor: I would prefer a better name and documentation for a global like this.

ckamm:
I tested the behavior on linux and it does what it's supposed to do.

Issue #3936, opened 2015-10-09
Make owncloudcmd return code depend on sync result


nickvergessen:
Smashbox currently has code (added from cern) that ignores the returncode on the sync.
After speaking with @guruz it should be checked, so the return code is reliable and we can notify the user when the returncode is not-successful.

moscicki:
Yes. Definitely a good idea. The question is what it means that the sync failed. I guess it only means for example that the client failed to connect to the server (so there was a critical error which prevented the sync to run). But probably a sync run for which there was an error downloading one file out of 100  is OK. Or isn't it?

dragotin:
@moscicki yes, that's a fine line. I think I agree with what you said above, as the fail of  a sync might be part of an successful sync.

nickvergessen:
Okay at least the return codes of --help and --version should be changed.
They currently return 1 instead of non-zero

also multiple options can't be combined appearantly? I tried `--trust --version`:
```
nickv@nickv-think ~ $ /usr/bin/owncloudcmd --trust --version
Source dir '--trust/' does not exist.
```

ogoffart:
After a sync it always return 0.  It doe snot matter if the sync success or fails.

We could change that, but under which criterias?
Is any SoftError a failure? or is failures oinly FatalError?

schnidrig:
What about this:
0: no error at all. All files synchend and up to date.

Then if some kind of error occurs, use specific values to indicate what it was. 
For example: 
10: connection error
11:  server gave us some 5XX http return code

20: some files failed to download

30: some files failed to upload

40: ... you name it

SamuAlfageme:
I'm playing around with #5669, and noticed that network errors do not stop the program after they happen:

e.g.: Authentication Failed: 

```
$ cmd_return_codes owncloudcmd . demo.owncloud.com
Please enter user name: asdasd
Password for user asdasd: !!! OCC::JsonApiJob created for "http://demo.owncloud.com" + "ocs/v1.php/cloud/capabilities" ""
Redirecting "GET" QUrl("http://demo.owncloud.com/ocs/v1.php/cloud/capabilities?format=json") QUrl("https://demo.owncloud.com:443/ocs/v1.php/cloud/capabilities?format=json")
Stop request: Authentication failed for  "https://demo.owncloud.com:443/ocs/v1.php/cloud/capabilities?format=json"
[...]

PR #5669, opened 2017-03-31
owncloudcmd: return code depend on sync result

ogoffart:
Issue #3936

### DIFF
--- a/src/cmd/cmd.cpp
+++ b/src/cmd/cmd.cpp
@@ -478,7 +478,12 @@ restart_sync:
 
     SyncEngine engine(account, options.source_dir, folder, &db);
     engine.setIgnoreHiddenFiles(options.ignoreHiddenFiles);
+#if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
+    QObject::connect(&engine, &SyncEngine::finished,
+                     [&app](bool result) { app.exit(result ? EXIT_SUCCESS : EXIT_FAILURE); });
+#else
     QObject::connect(&engine, SIGNAL(finished(bool)), &app, SLOT(quit()));
+#endif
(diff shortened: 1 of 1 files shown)


ogoffart:
(This is made against 2.3,  but i don't know exactly if it should go into 2.3 or master)

ogoffart:
cc @SamuAlfageme

src/cmd/cmd.cpp
```diff
@@ -478,7 +478,12 @@ int main(int argc, char **argv) {
 
     SyncEngine engine(account, options.source_dir, folder, &db);
     engine.setIgnoreHiddenFiles(options.ignoreHiddenFiles);
+#if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
+    QObject::connect(&engine, &SyncEngine::finished,
+                     [&app](bool result) { app.exit(result ? EXIT_SUCCESS : EXIT_FAILURE); });
```
jturcotte: Is it sure that we will have c++11 on all our platforms already? Otherwise maybe better guard it with the same defines as in folderstatusmodel.cpp.
`QCoreApplication::exit` is static, so you would probably also not need to capture `app`.

guruz: Jenkins does not seem to complain yet.
(but i dont know if it checks all platforms..)

Maybe where we have Qt5 we also have that?

guruz:
👍 although maybe it's safer in master where we for sure will go with a bundled Qt 5.6.2?
(FYI @dschmidt )

guruz:
👎 until https://github.com/owncloud/client/issues/3936#issuecomment-292921020 is adressed somehow?

ogoffart:
@guruz:  that problem (Not starting a sync if the capability call faills)  has nothing to do with this pull request which is about returining a different code when the sync finished with an error.

Issue #5647, opened 2017-03-24
Provide severity, categories and account IDs in log files


jturcotte:
- Use [Qt categorized logging](http://doc.qt.io/qt-5/qloggingcategory.html) to improve the readability and grepability of client log files
- It should be possible to filter by category and severity to control the size of log files a bit (using Qt environment variables or command line arguments)
- Try to make sync protocol events easy to find (special category) in the log to provide a replacement for `.owncloudsync.log`
- A bit of changes to log window could also be implied, either add some filtering/search options, or removing it altogether in favor of enabling file logging and opening the file explorer in the right folder when F12 is pressed.

SamuAlfageme:
Some related issues that could be addressed: 
- [ ] Format: https://github.com/owncloud/client/issues/5634
- [ ] Size: https://github.com/owncloud/client/issues/3019
- [ ] logwindow invocation: https://github.com/owncloud/client/issues/4098

butonic:
If the server sent the requestID as a response header, could we add it to the log, so we can link requests between server and client logs? see https://github.com/owncloud/documentation/issues/2967

jturcotte:
@butonic If you tell us which header to put in the log, this should be very easy to do at least for the most important WebDAV methods.

butonic:
Not yet decided. I suppose `X-OC-RequestId` makes sense? Lets ask the Internet ...

It seems Heroku allows clients to send a request id in the `X-Request-ID` header. That of course makes sense because the client can then already log the id even when the server fails to respond. I think this maskes more sense then letting the server generate the uuid. That might be a fallback in case no `X-Request-ID` header has been sent: `RequestHeader setifempty X-Request-ID %{UNIQUE_ID}e` in the .htaccess and in oc core we then only check X-Request-ID.

@DeepDiver1975 @PVince81 @jturcotte sounds like a plan?

jturcotte:
I'm putting ReadyToTest but just any kind of QA feedback would be nice regarding the output to see if anything is missing or could be improved.

SamuAlfageme:
- I really like this whole logging concept, it's way cleaner than the previous one and super handy when it comes to track down flows and issues with it.  🎉 😊
    - It'd be really useful to provide checkboxes with the categories on the bottom part of the Log Output window to filter error/warnings and discard `info` messages and avoid some noise 
[...]

PR #5671, opened 2017-03-31
Use Qt logging categories

jturcotte:
Issue #5647


src/libsync/logger.cpp
```diff
@@ -67,10 +67,10 @@ Logger *Logger::instance()
 Logger::Logger( QObject* parent) : QObject(parent),
   _showTime(true), _logWindowActivated(false), _doFileFlush(false), _logExpire(0)
 {
-#ifndef NO_MSG_HANDLER
```
ogoffart: :-1:  

I want to keep my own message pattern which i set in an environment variable.

(My message pattern has color code for warning which is very isefull.)

So please don't overwrite it when NO_MSG_HANDLER is defined.

Or maybe

#ifdef NO_MSG_HANDLER
if (!qEnvironmentVariableIsSet("QT_MESSAGE_PATTERN"))
#endif

src/libsync/logger.cpp
```diff
@@ -67,10 +67,10 @@ Logger *Logger::instance()
...
 #if QT_VERSION >= QT_VERSION_CHECK(5, 4, 0)
-    qSetMessagePattern("%{time MM-dd hh:mm:ss:zzz} %{threadid} %{function}: %{message}");
+    qSetMessagePattern("%{time MM-dd hh:mm:ss:zzz} [ %{type} %{category} ]%{if-debug}\t(%{threadid}) %{function}%{endif}:\t%{message}");
```
ogoffart: (i wonder if threadid is so usefull anyway. It should be clear what message runs in what thread.)

src/gui/accountsettings.cpp
```diff
@@ -58,6 +57,8 @@
 
 namespace OCC {
 
+Q_LOGGING_CATEGORY(lcAccountSettings, "gui.account.settings", QtInfoMsg)
```
ogoffart: or should it be gui.settings.account ?

src/gui/folder.cpp
```diff
@@ -46,6 +45,9 @@
 
 namespace OCC {
 
+// Start with "sync." since this is mostly sync logic at the wrong layer
```
ogoffart: i disagree with that.   it is scheduling logic,  but not sync logic.

But that does not matter much.

src/gui/folderstatusmodel.cpp
```diff
@@ -29,6 +29,8 @@ Q_DECLARE_METATYPE(QPersistentModelIndex)
 
 namespace OCC {
 
+Q_LOGGING_CATEGORY(lcFolderStatus, "gui.folder.status", QtInfoMsg)
```
ogoffart: gui.floder.model  maybe ?   (FolderStatusModel is a bad name anyway)

src/libsync/bandwidthmanager.cpp
```diff
@@ -85,12 +88,11 @@ BandwidthManager::BandwidthManager(OwncloudPropagator *p) : QObject(),
 
 BandwidthManager::~BandwidthManager()
 {
-    qDebug() << Q_FUNC_INFO;
 }
 
 void BandwidthManager::registerUploadDevice(UploadDevice *p)
 {
-    //qDebug() << Q_FUNC_INFO << p;
+    qCDebug(lcBandwidthManager) << p;
```
ogoffart: do we need all these debug?  if they were commented out it probably does not matter

src/libsync/logger.cpp
```diff
@@ -20,9 +20,12 @@
 #include <qmetaobject.h>
 
 #include "csync.h"
+#include "csync_log.h"
 
 namespace OCC {
 
+Q_LOGGING_CATEGORY(lcCsync, "sync.csync", QtInfoMsg)
```
ogoffart: maybe just "csync" ?

src/libsync/logger.cpp
jturcotte: From the [doc](http://doc.qt.io/qt-5/qloggingcategory.html):
```
Order of evaluation:

[QLibraryInfo::DataPath]/qtlogging.ini
QtProject/qtlogging.ini
setFilterRules()
QT_LOGGING_CONF
QT_LOGGING_RULES
```

src/libsync/logger.cpp
jturcotte: Agreed, removed it.

src/gui/accountsettings.cpp
jturcotte: We don't have any other gui.settings yet, so not sure. I however doubt that people will tweak the rules in `gui.` except for the specific part they work on.

src/gui/folder.cpp
jturcotte: The reason is that I wanted to be able to get all sync extra stuff by doing `QT_LOGGING_RULES="sync.*=true"`. But the `Folder` doesn't have much impact on the sync, so I guess `info` will be enough most of the time anyway, so changed it.

src/gui/folderstatusmodel.cpp
jturcotte: Done

src/libsync/bandwidthmanager.cpp
jturcotte: I kept some, remove others, but that one won't be so useful investigating problems I guess. Removed it.

src/libsync/logger.cpp
jturcotte: That one I kept since I'd like to be able to set `QT_LOGGING_RULES="sync.*=true"`, and I think csync plays a role at the same level as `sync.engine` and `sync.propagator`.

Doing `QT_LOGGING_RULES="*=true"` is a bit annoying because it also includes `qt.*`, so I opted for `gui` and `sync` as top-level rules and didn't parent everything under `oc.`.

src/libsync/logger.cpp
ogoffart: ah ok, no problems then!

ogoffart:
:+1:   from me
but we need to fix jenkins

jturcotte:
The CI should be updated with a better version of Qt within a few weeks. I won't merge this pull request until we can get it to pass the build anyway.

ckamm:
I also think this is good, though I haven't reviewed in depth. I find stuff like `qCDebug(lcAccountSettings)` a bit verbose - too much noise for just logging - but it's not a major complaint and I'm excited to get log categories.

jturcotte:
The thing is that the macro converts `lcAccountSettings` to a singleton function, and the name needs to be unique across the binary. So I think the name could be shorter too, but I don't know what we could use that is both expressive and unique. Let me know if you have idea, I'll change them when I rebase the branch.

guruz:
> The CI should be updated with a better version of Qt within a few weeks.

This is done now right? I think we should get this in before it becomes too messy to merge.
Is there any change in the default behaviour? Will people still get all logs by default?

jturcotte:
I pushed a rebase and now the build passes. This branch will become outdated pretty fast, could somebody approve it?

I also added the last commit that adds the `--logdebug` command line option and a checkbox in the log window to enable debug messages if "info" isn't detailed enough in some cases.

ckamm:
@jturcotte Will review/approve/merge tomorrow.

guruz:
Needs also my fixup from https://github.com/owncloud/client/commits/qloggingcategory

guruz:
I don't understand the indentation.. Can this be fixed somehow to align the actual log lines?

```
05-11 12:23:54:610 [ info sync.database ]:	sqlite3 journal_mode= "wal"
05-11 12:23:54:610 [ info sync.engine ]:	There are 4710969344 bytes available at "/Users/guruz/owncloud10git/" and at least 50000000 are required
05-11 12:23:54:611 [ info sync.engine ]:	Sync with existing sync journal
05-11 12:23:54:611 [ info sync.engine ]:	"Using Qt 5.6.2 SSL library OpenSSL 1.0.2j  26 Sep 2016 on OS X El Capitan (10.11)"
05-11 12:23:54:611 [ info sync.engine ]:	NOT Using Selective Sync
05-11 12:23:54:611 [ info sync.engine ]:	#### Discovery start ####################################################
05-11 12:23:54:611 [ info sync.engine ]:	Server "10.0.0.6" "rootEtagChangesNotOnlySubFolderEtags=1"
05-11 12:23:54:611 [ info sync.csync ]:	file: Containing mp3 file.zip, instruction: INSTRUCTION_NEW <<=
05-11 12:23:54:611 [ info sync.csync ]:	file: subfolder, instruction: INSTRUCTION_NEW <<=
05-11 12:23:54:612 [ info sync.csync ]:	file: welcome copy.txt, instruction: INSTRUCTION_NEW <<=
05-11 12:23:54:612 [ warning sync.csync ]:	Invalid characters in file/directory name, please rename: "💩.txt" (/Users/guruz/owncloud10git)
```

src/libsync/propagateuploadv1.cpp
```diff
@@ -176,12 +176,6 @@ void PropagateUploadFileV1::slotPutFinished()
 
     slotJobDestroyed(job); // remove it from the _jobs list
 
-    qCDebug(lcPropagateUpload) << job->reply()->request().url() << "FINISHED WITH STATUS"
```
guruz: this one is usually very useful. Or is that now printed out somewhere else?

src/gui/application.cpp
```diff
@@ -357,10 +360,11 @@ void Application::setupLogging()
     Logger::instance()->setLogDir(_logDir);
     Logger::instance()->setLogExpire(_logExpire);
     Logger::instance()->setLogFlush(_logFlush);
+    Logger::instance()->setLogDebug(_logDebug);
 
     Logger::instance()->enterNextLogFile();
 
-    qDebug() << QString::fromLatin1( "################## %1 %2 (%3) %4 on %5").arg(_theme->appName())
+    qCDebug(lcApplication) << QString::fromLatin1( "################## %1 %2 (%3) %4 on %5").arg(_theme->appName())
```
guruz: Info please, this one needs to be shown all the time

src/gui/folderstatusmodel.cpp
```diff
@@ -730,7 +732,7 @@ void FolderStatusModel::slotLscolFinishedWithError(QNetworkReply* r)
     }
     auto parentInfo = infoForIndex(idx);
     if (parentInfo) {
-        qDebug() << r->errorString();
+        qCDebug(lcFolderStatus) << r->errorString();
```
guruz: Info please, unless it is shown by another caller?

src/gui/folderwatcher.cpp
```diff
@@ -57,7 +58,7 @@ bool FolderWatcher::pathIsIgnored( const QString& path )
 
 #ifndef OWNCLOUD_TEST
     if (_folder->isFileExcludedAbsolute(path)) {
-        qDebug() << "* Ignoring file" << path;
+        qCDebug(lcFolderWatcher) << "* Ignoring file" << path;
```
guruz: I'm undecided on this one, but think it should be shown all the time.
@ogoffart @ckamm opinions?

src/gui/lockwatcher.cpp
```diff
@@ -32,7 +34,7 @@ LockWatcher::LockWatcher(QObject* parent)
 
 void LockWatcher::addFile(const QString& path)
 {
-    qDebug() << "Watching for lock of" << path << "being released";
+    qCDebug(lcLockWatcher) << "Watching for lock of" << path << "being released";
```
guruz: I think it's important to have this shown always, same below..

@ckamm ?

src/libsync/abstractnetworkjob.cpp
```diff
@@ -260,13 +260,13 @@ void AbstractNetworkJob::start()
...
     QString parentMetaObjectName = parent() ? parent()->metaObject()->className() : "";
-    qDebug() << "!!!" << metaObject()->className() << "created for" << displayUrl << "+" << path() << parentMetaObjectName;
+    qCInfo(lcNetworkJob) << metaObject()->className() << "created for" << displayUrl << "+" << path() << parentMetaObjectName;
 }
 
 void AbstractNetworkJob::slotTimeout()
 {
     _timedout = true;
-    qDebug() << this << "Timeout" << (reply() ? reply()->request().url() : path());
+    qCDebug(lcNetworkJob) << this << "Timeout" << (reply() ? reply()->request().url() : path());
```
guruz: Info/Warning..

src/libsync/syncengine.cpp
```diff
@@ -1469,7 +1466,7 @@ void SyncEngine::abort()
 {
     // Sets a flag for the update phase
     csync_request_abort(_csync_ctx);
-    qDebug() << Q_FUNC_INFO << _discoveryMainThread;
+    qCDebug(lcEngine) << "Aborting sync, _discoveryMainThread:" << _discoveryMainThread;
```
guruz: Info..

src/libsync/propagatorjobs.cpp
```diff
@@ -159,7 +162,7 @@ void PropagateLocalMkdir::start()
     }
 
     if( Utility::fsCasePreserving() && propagator()->localFileNameClash(_item->_file ) ) {
-        qDebug() << "WARN: new folder to create locally already exists!";
+        qCDebug(lcPropagateLocalMkdir) << "WARN: new folder to create locally already exists!";
```
guruz: warning?

guruz: ... ^^ maybe also with filename if this one is warning but the one above with the filename is only debug?

src/libsync/propagateuploadv1.cpp
```diff
@@ -175,12 +176,6 @@ void PropagateUploadFileV1::slotPutFinished()
 
     slotJobDestroyed(job); // remove it from the _jobs list
 
-    qDebug() << Q_FUNC_INFO << job->reply()->request().url() << "FINISHED WITH STATUS"
```
guruz: This is removed here but not in other jobs? Maybe keep it there then too? Even if this is v1..

```
src/libsync/networkjobs.cpp:    qCInfo(lcEtagJob) << "Request Etag of" << reply()->request().url() << "FINISHED WITH STATUS"
src/libsync/networkjobs.cpp:    qCInfo(lcMkColJob) << "MKCOL of" << reply()->request().url() << "FINISHED WITH STATUS"
src/libsync/networkjobs.cpp:    qCInfo(lcLsColJob) << "LSCOL of" << reply()->request().url() << "FINISHED WITH STATUS"
src/libsync/networkjobs.cpp:    qCInfo(lcPropfindJob) << "PROPFIND of" << reply()->request().url() << "FINISHED WITH STATUS"
src/libsync/networkjobs.cpp:    qCInfo(lcProppatchJob) << "PROPPATCH of" << reply()->request().url() << "FINISHED WITH STATUS"
src/libsync/networkjobs.cpp:    qCInfo(lcJsonApiJob) << "JsonApiJob of" << reply()->request().url() << "FINISHED WITH STATUS"
src/libsync/propagatedownload.cpp:            qCInfo(lcGetJob) << "GET of" << reply()->request().url().toString() << "FINISHED WITH STATUS"
src/libsync/propagateremotedelete.cpp:    qCInfo(lcDeleteJob) << "DELETE of" << reply()->request().url() << "FINISHED WITH STATUS"
src/libsync/propagateremotemove.cpp:    qCInfo(lcMoveJob) << "MOVE of" << reply()->request().url() << "FINISHED WITH STATUS"
src/libsync/propagateupload.h:        qCInfo(lcPutJob) << "PUT of" << reply()->request().url().toString() << "FINISHED WITH STATUS"
```

src/libsync/propagateuploadv1.cpp
```diff
@@ -209,7 +204,7 @@ void PropagateUploadFileV1::slotPutFinished()
         }
         QByteArray replyContent;
         QString errorString = job->errorStringParsingBody(&replyContent);
-        qDebug() << replyContent; // display the XML error in the debug
+        qCDebug(lcPropagateUpload) << replyContent; // display the XML error in the debug
```
guruz: I wonder if this should be upgraded too since it's about an error condition?

src/libsync/propagateuploadv1.cpp
```diff
@@ -45,7 +46,7 @@ void PropagateUploadFileV1::doStartUpload()
     if (progressInfo._valid && Utility::qDateTimeToTime_t(progressInfo._modtime) == _item->_modtime ) {
         _startChunk = progressInfo._chunk;
         _transferId = progressInfo._transferid;
-        qDebug() << Q_FUNC_INFO << _item->_file << ": Resuming from chunk " << _startChunk;
+        qCDebug(lcPropagateUpload) << _item->_file << ": Resuming from chunk " << _startChunk;
```
guruz: Hmmmm it's not really an error condition but resuming uploads is something crucial..

src/libsync/propagateuploadng.cpp
```diff
@@ -346,12 +346,6 @@ void PropagateUploadFileNG::slotPutFinished()
 
     slotJobDestroyed(job); // remove it from the _jobs list
 
-    qDebug() << job->reply()->request().url() << "FINISHED WITH STATUS"
```
guruz: You kept it in other jobs, maybe keep it here too?
Else how can you debug out all those codes?

```
src/libsync/networkjobs.cpp:    qCInfo(lcEtagJob) << "Request Etag of" << reply()->request().url() << "FINISHED WITH STATUS"
src/libsync/networkjobs.cpp:    qCInfo(lcMkColJob) << "MKCOL of" << reply()->request().url() << "FINISHED WITH STATUS"
src/libsync/networkjobs.cpp:    qCInfo(lcLsColJob) << "LSCOL of" << reply()->request().url() << "FINISHED WITH STATUS"
src/libsync/networkjobs.cpp:    qCInfo(lcPropfindJob) << "PROPFIND of" << reply()->request().url() << "FINISHED WITH STATUS"
src/libsync/networkjobs.cpp:    qCInfo(lcProppatchJob) << "PROPPATCH of" << reply()->request().url() << "FINISHED WITH STATUS"
src/libsync/networkjobs.cpp:    qCInfo(lcJsonApiJob) << "JsonApiJob of" << reply()->request().url() << "FINISHED WITH STATUS"
src/libsync/propagatedownload.cpp:            qCInfo(lcGetJob) << "GET of" << reply()->request().url().toString() << "FINISHED WITH STATUS"
src/libsync/propagateremotedelete.cpp:    qCInfo(lcDeleteJob) << "DELETE of" << reply()->request().url() << "FINISHED WITH STATUS"
src/libsync/propagateremotemove.cpp:    qCInfo(lcMoveJob) << "MOVE of" << reply()->request().url() << "FINISHED WITH STATUS"
src/libsync/propagateupload.h:        qCInfo(lcPutJob) << "PUT of" << reply()->request().url().toString() << "FINISHED WITH STATUS"
```

guruz: AhhhhH! you removed it because it's in propagateupload.h :)

src/libsync/propagateuploadng.cpp
```diff
@@ -376,7 +370,7 @@ void PropagateUploadFileNG::slotPutFinished()
         _item->_httpErrorCode = job->reply()->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
         QByteArray replyContent;
         QString errorString = job->errorStringParsingBody(&replyContent);
-        qDebug() << replyContent; // display the XML error in the debug
+        qCDebug(lcPropagateUpload) << replyContent; // display the XML error in the debug
```
guruz: same as for v1, i think those errors should be displayed more easily as the log is important for people to find out what's going wrong

src/libsync/propagateupload.cpp
```diff
@@ -525,7 +523,7 @@ void PropagateUploadFileCommon::abort()
 {
     foreach(auto *job, _jobs) {
         if (job->reply()) {
-            qDebug() << Q_FUNC_INFO << job << this->_item->_file;
+            qCDebug(lcPropagateUpload) << job << this->_item->_file;
```
guruz: hmm @ckamm ?

src/libsync/propagatedownload.cpp
```diff
@@ -626,13 +623,13 @@ static QString makeRecallFileName(const QString &fn)
 
 void handleRecallFile(const QString& filePath, const QString& folderPath, SyncJournalDb& journal)
 {
-    qDebug() << "handleRecallFile: " << filePath;
+    qCDebug(lcPropagateDownload) << "handleRecallFile: " << filePath;
```
guruz: The recall file is an important expert feature, i think the log should definitely show by default..

src/libsync/propagateuploadv1.cpp
guruz: (ignore the above..)

guruz:
I think it looks fine except for the log alignment mentioned above and the error promotions I mentioned.

So we don't want the ```--logdebug``` as default?

src/libsync/propagateupload.cpp
ckamm: I've never needed this log line.

src/gui/lockwatcher.cpp
ckamm: Agreed, info please.

src/gui/folderwatcher.cpp
ckamm: It's useful when debugging folder watcher issues, but I'm okay with it being in debug.

guruz: I more meant from a user perspective "oh my file is not synced, why?". But maybe that happens too seldomly..

src/libsync/propagateuploadv1.cpp
jturcotte: It moved in src/libsync/networkjobs.cpp for all network jobs, instead of reporting it in the propagation job (so that the right category is used).

src/gui/folderstatusmodel.cpp
jturcotte: LsColJob will report its result.

src/gui/folderwatcher.cpp
jturcotte: If the file is excluded you'll see it in the sync log, and I can't really think of other cases where the user would need this to understand the behavior of the client. Let me know if you think of any.

src/libsync/propagatedownload.cpp
jturcotte: More descriptive relevant stuff is qCInfo/Warning lower in that function.

jturcotte:
> The categories are useful for grepping etc, but I don't like that so many important things that helped us in the last 5 years to fix things are hidden now behind the --debuglog switch :[
> 
> I wonder what is the usecase then for this switch. Because then everything could go behind this switch and you'd not log anything without anymore.

If you want to fix an issue in the client (debug it) then use --debuglog or check the checkbox in the log window. I added those to make sure that support can still get them, they're not gone. For the rest of the time (>80%?) you get a log easier to follow and understand. qDebug is the most verbose category and I think it is useful to be able to put logging in the code "just in case" without having to worry that it clutters the log, and without having to commit logging as comments like we did before which won't compile when you need it.

src/libsync/propagateuploadv1.cpp
jturcotte: `PropagateItemJob::done` will normally print errorString, I changed it to a warning when it's an error status.

SamuAlfageme:
@jturcotte can we reduce somehow the noise from using functional stuff? e.g.`[ ()::(anonymous class)::operator()()::(anonymous class)::operator()():: ]:` doesn't look handy at first sight.

jturcotte:
@SamuAlfageme Function names only appear for qCDebug statements. I think that --debug will probably become more noisy than before and we can use logging rules to enable only what we want. If the rest of the line is useful to you, I think that we could either:

- Remove function names from the default message pattern (but this would mean that we'd need to explicitly add it to every statement in the code). qCDebug statements tend to be quickly written and are less expressive, so the context is usually useful.
- You can [https://woboq.com/blog/nice-debug-output-with-qt.html](provide your own message pattern through an environment variable)
- Maybe we could put the function name at the end, or try to support some kind of tool to read logs that we'd output to a file, that would place this info in a resizable column that you would make the rest more readable.

Let me know what you think.

PR #5677, opened 2017-04-03
Set proper application name, icon name for freedesktop notifications for OEM builds

juliushaertl:
This PR will set proper application name and icon name for OEM builds.


CMakeLists.txt
```diff
@@ -178,6 +178,12 @@ endif()
 
 find_package(ZLIB)
 
+if ( EXISTS ${OEM_THEME_DIR}/OEM.cmake )
+    set(ICON_APP_NAME ${APPLICATION_SHORTNAME})
+else()
+    set(ICON_APP_NAME "owncloud")
+endif()
+
```
dschmidt: How about 
```cmake
if (NOT DEFINED APPLICATION_ICON_NAME)
    set(APPLICATION_ICON_NAME ${APPLICATION_SHORTNAME})
endif()
``` 
underneath here https://github.com/juliushaertl/client/blob/2d1a210cd1f9de6059a244d9bb0ee4e8de204eac/CMakeLists.txt#L25?

juliushaertl: Sorry for the delay. I basically used the same as in here: https://github.com/owncloud/client/blob/master/src/gui/CMakeLists.txt#L226-L232

But I guess we could also move to the change you proposed and add the APPLICATION_ICON_NAME to the OWNCLOUD.cmake file.

juliushaertl:
@dschmidt I've updated the PR to use APPLICATION_ICON_NAME.

rugk:
Any news here?

dschmidt:
Good to go from my POV. Sorry, for missing the update. Can you resolve the conflict?

dschmidt:
Wow - that GitHub resolve-conflicts-on-the-web-ui is damn nice :-D

juliushaertl:
@dschmidt Rebased and fixed the conflicts.

That github ui sounds interesting, I didn't knew that it was there until now. :wink:

dschmidt:
Woah, that was quick!

PR #5682, opened 2017-04-04
GUI Testing

dschmidt:
This is still WIP but by now it feels like it's worth a first round of reviewing.

The implementation of the command parser is still fragile (splitting by | is not exactly error-proof)
I'm not completely settled on the commands we want to support: this largely depends on feedback from @SamuAlfageme 

As most of it is hidden inside the `SocketApi` class, it should not affect any other part of the application with only one excception:
We just need to give a few widgets names to be able to find them, are you fine with that?

Also let me know if the code does not speak for itself enough and you need any explanation.

### DIFF
--- a/src/gui/socketapi.cpp
+++ b/src/gui/socketapi.cpp
@@ -75,15 +75,65 @@
 
 namespace {
 #if GUI_TESTING
-QWidget *findWidget(const QString &objectName)
+
+QList<QObject*> allObjects(const QList<QWidget*> &widgets) {
+    QList<QObject*> objects;
+    std::copy(widgets.constBegin(), widgets.constEnd(), std::back_inserter(objects));
+
+    objects << qApp;
+
+    return objects;
+}
+
+QObject *findWidget(const QString &queryString, const QList<QWidget*> &widgets = QApplication::allWidgets())
 {
-    auto widgets = QApplication::allWidgets();
+    auto objects = allObjects(widgets);
 
-    auto foundWidget = std::find_if(widgets.constBegin(), widgets.constEnd(), [&](QWidget *widget) {
-        return widget->objectName() == objectName;
-    });
+    QList<QObject*>::const_iterator foundWidget;
 
-    if (foundWidget == widgets.constEnd()) {
+    if (queryString.contains('>')) {
+        DEBUG << "queryString contains >";
+
+        auto subQueries = queryString.split('>', QString::SkipEmptyParts);
+        Q_ASSERT(subQueries.count() == 2);
+
+        auto parentQueryString = subQueries[0].trimmed();
+        DEBUG << "Find parent: " << parentQueryString;
+        auto parent = findWidget(parentQueryString);
+
+        if(!parent) {
+            return nullptr;
+        }
+
+        auto childQueryString = subQueries[1].trimmed();
+        auto child = findWidget(childQueryString, parent->findChildren<QWidget*>());
+        DEBUG << "found child: " << !!child;
+        return child;
+
+    } else if(queryString.startsWith('#')) {
+        auto objectName = queryString.mid(1);
+        DEBUG << "find objectName: " << objectName;
+        foundWidget = std::find_if(objects.constBegin(), objects.constEnd(), [&](QObject *widget) {
+            return widget->objectName() == objectName;
+        });
+    } else {
+        QList<QObject*> matches;
+        std::copy_if(objects.constBegin(), objects.constEnd(), std::back_inserter(matches), [&](QObject* widget) {
+            return widget->inherits(queryString.toLatin1());
+        });
+
+        std::for_each(matches.constBegin(), matches.constEnd(), [](QObject* w) {
+            if(!w) return;
+            DEBUG << "WIDGET: " << w->objectName() << w->metaObject()->className();
+        });
+
+        if(matches.empty()) {
+            return nullptr;
+        }
+        return matches[0];
+    }
+
+    if (foundWidget == objects.constEnd()) {
         return nullptr;
     }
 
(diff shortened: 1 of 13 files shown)


dschmidt:
Jenkins is unhappy about Qt5 code ... but as we're about to drop Qt5 support, I'd consider this a non issue.

ogoffart:
Maybe a small documentation of what the commands do could be nice.  So from that documetation, @SamuAlfageme  would also be able to tell if that is enough for what he wants to do

ogoffart:
We could merge the last commit "Disable Qt4 CI" already!  that'd be useful for other things

ogoffart:
@SamuAlfageme it would be nice to have your input so we can move forward.

SamuAlfageme:
@ogoffart no worries, I'll post here my comments when I sort them out. So far I haven't had time to play as much as I would like with this and most of it was setting a reproducible test env. and conversations with @dschmidt about what can be tested with this, etc.

First real step will probably be to test all the translations and string labels in the client using this.

ckamm:
@SamuAlfageme Is this one still alive?

SamuAlfageme:
@ckamm yeah; it's waiting for further development and some documentation once @dschmidt and I find some time to sit together. It has been pushed for a while now 😕 

See it in action in the talk I gave in the conference: https://www.youtube.com/watch?v=H3FuMeRIR2c & https://www.youtube.com/watch?v=nxQkM6nxW2s and the proof of concept in https://gitlab.com/alfageme/client-GUI-tests/-/jobs/14114030

src/gui/socketapi.cpp
```diff
@@ -506,6 +511,109 @@ void SocketApi::command_SHARE_MENU_TITLE(const QString &, SocketListener* listen
...
+    if (!widget) {
+        job->reject(QLatin1String("widget not found"));
+        return;
+    }
+
+    auto propertyName = job->arguments()[QLatin1String("property")].toString();
+
+    job->resolve(widget->property(propertyName.toLocal8Bit().constData())
+                     .toString()
+                     .toLocal8Bit()
```
ogoffart: Should be toUtf8()  here and everywhere.  We use UTF8 in the protocol.

src/gui/socketapi.cpp
```diff
@@ -506,6 +511,109 @@ void SocketApi::command_SHARE_MENU_TITLE(const QString &, SocketListener* listen
...
+{
+    auto &arguments = job->arguments();
+
+    auto widget = findWidget(arguments["objectName"].toString());
+    if (!widget) {
+        job->reject(QLatin1String("widget not found"));
+        return;
+    }
+
+    QMetaObject::invokeMethod(widget, arguments["method"].toString().toLocal8Bit().constData());
```
ogoffart: toString().toLocal8Bit() ?  If that's already a QByteArray, it should have the right encoding.

guruz:
in 5 days, this PR will be 2 years old

michaelstingl:
@guruz old, but still interesting. After Drone setup, we can continue …

dschmidt:
Yes, this is totally still relevant and we are very close to actually being able to use this.

dschmidt:
Updated and rebased. 

If there's nothing fundamentally wrong with the gui testing commands, I'd vote for merging it now. It's pretty much work in progress, but we already use it for our limited tests and it works.
Anyhow I would like to base further feature PRs on this and not end up with a giant branch. Also this has gone out of date several times already.

src/gui/socketapi.cpp
```diff
@@ -58,10 +66,80 @@
 #include <CoreFoundation/CoreFoundation.h>
 #endif
 
+
 // This is the version that is returned when the client asks for the VERSION.
 // The first number should be changed if there is an incompatible change that breaks old clients.
 // The second number should be changed when there are new features.
 #define MIRALL_SOCKET_API_VERSION "1.1"
+#define DEBUG qDebug() << "SocketApi: "
```
ogoffart: Don't do that.  Use a debug category.  (As it allow to turn the log on and off with a env variable.

But I think you shoud cleanup most debug output.

src/gui/socketapi.cpp
```diff
@@ -1027,6 +1091,177 @@ void SocketApi::command_GET_MENU_ITEMS(const QString &argument, OCC::SocketListe
...
+void SocketApi::command_ASYNC_SET_WIDGET_PROPERTY(const QSharedPointer<SocketApiJob> &job)
+{
+    auto &arguments = job->arguments();
+    auto widget = findWidget(arguments["objectName"].toString());
+    if (!widget) {
+        job->reject(QLatin1String("widget not found"));
+        return;
+    }
+    widget->setProperty(arguments["property"].toString().toUtf8().constData(),
+                        arguments["value"].toString().toUtf8().constData());
```
ogoffart: The property name is a `char*`  ,  but the property itself is a QVariant,  and you should keep using QString here, not convert to char*

src/gui/socketapi.cpp
```diff
@@ -1027,6 +1091,177 @@ void SocketApi::command_GET_MENU_ITEMS(const QString &argument, OCC::SocketListe
...
+
+            DEBUG << "VALUE: " << value;
+            break;
+        }
+
+        auto tmpObject = var.value<QObject*>();
+        if(tmpObject) {
+            currentObject = tmpObject;
+        } else {
+            DEBUG << "TODO: object not found, what should happen here now?";
```
ogoffart: Maybe reply with an error on the socket?

src/gui/socketapi.cpp
```diff
@@ -1027,6 +1091,177 @@ void SocketApi::command_GET_MENU_ITEMS(const QString &argument, OCC::SocketListe
...
+        }
+    }
+
+    job->resolve(value.toUtf8().constData());
```
ogoffart: Resolve takes a QString.  No need to convert.

src/gui/socketapi.cpp
```diff
@@ -1027,6 +1091,177 @@ void SocketApi::command_GET_MENU_ITEMS(const QString &argument, OCC::SocketListe
...
+
+    QObject* currentObject = widget;
+    QString value;
+    for(int i = 0;i<segments.count(); i++) {
+        auto segment = segments.at(i);
+        auto var = currentObject->property(segment.toUtf8().constData());
+
+        if(var.canConvert<QString>()) {
+            var.convert(QMetaType::QString);
+            value = var.value<QString>();
```
ogoffart: you know that QDebug prints the content of the QVariant even if it is not a QString...

Anyway, this looks like leftover debug that should be cleaned

dschmidt: It's not converted to print it, it's converted to pass it to `resolve()`

jnweiger:
Yeah merge!

PR #5689, opened 2017-04-09
Labels / Translations polishing: adding a dot to properly end status messages

Quatrejas:


ldmpub:
See the related discussion to this "missing dot" here: https://central.owncloud.org/t/owncloud-client-unable-to-find-a-string-to-improve-french-translation/6164

phil-davis:
When the "full stop" is included in a string passed to tr() then the translator gets a chance to make the translation have the language-specific "sentence terminator" (full-stop equivalent). e.g. in Nepali sentences are ended with the vertical bar "|". So in places that append just a Qchar(".") to a string, it would be good to run the "." through the translator and then the translation team have a chance to choose a language-specific equivalent.

src/gui/accountsettings.cpp
```diff
@@ -571,7 +571,7 @@ void AccountSettings::slotAccountStateChanged(int state)
         } else if (state == AccountState::SignedOut) {
             showConnectionLabel( tr("Signed out from %1.").arg(serverWithUser) );
         } else {
-            showConnectionLabel(tr("No connection to %1 at %2.")
```
ckamm: Why did you remove that dot?

ckamm:
@Quatrejas Adding the dot like this is fine, I'm just confused why one of the error messages had its dot removed as well.

src/gui/accountsettings.cpp
Quatrejas: Hi,
I removed the dot because it is re-added by the other modification I made (the join adds the dot).
However based on phil-davis comment above and the discussions we are having on the French translation team, I understand this is not a satisfying PR, so I think I should update it (more below in the discussion thread)

Quatrejas:
Hi, based on @phil-davis comment above and the discussions we are having on the French translation team, I understand this is not a satisfying PR, so I think I should update it.

The discussions that took place on the French translation ML have raised the 3 following options:

- easiest one would be to follow phil-davis advice by enclosing the dot in a tr() call. I've tested this locally on my computer, and also switched the way the string is built in order not to have to remove the dot from the first message (this will save time for the translators if the former string is not modified).
- option 2 would be to request Qt maintainers to update the error messages and add a dot to end them, although we tend to think that it might be unlikely they accept / they might be skipping dots intentionnally
- option 3 would be to catch the Qt error messages in owncloud client and rephrase owncloud's own error messages. However I don't feel up to such a task because I'm unsure of how many errors could be caught and how they are caught. When reading the code, I have not understood how the QStringList errors was populated.

So if that seems fine to you, I'd like to go for option 1. I believe I can add a new commit (https://github.com/Quatrejas/client/commit/e8d2901ebf9d3113de9701c03f7ac68f8382ec66) from my branch to the PR to update it.

Quatrejas:
Hi @ckamm, I made a new fix proposal, which I hope is better:
- I restored the accountsettings file contents
- I close the sentence of the error message when the message is generated.
Up to now I had not understood where the message was built because I am not familiar with Qt signals, but  now that I've understood the concept I could trace back to where the original message is built.

src/gui/accountsettings.cpp
```diff
@@ -405,8 +405,8 @@ void AccountSettings::showConnectionLabel( const QString& message, QStringList e
         ui->connectLabel->setToolTip(QString());
         ui->connectLabel->setStyleSheet(QString());
     } else {
-        errors.prepend(message);
-        const QString msg = errors.join(QLatin1String("\n"));
+		errors.prepend(message);
+		const QString msg = errors.join(QLatin1String("\n"));
```
ckamm: indent

ckamm:
@Quatrejas I think this is almost ready, thanks for the update (and sorry for the delay)!

There's one indent I'd like you to fix. I'm fine with adding trailing dots to connectionvalidator.cpp errors, that was inconsistent before. 

Making sentences out of the errors in abstractnetworkjob seems odd. I'd prefer you did the `QObject::tr("%1.").arg(job->errorString())` in connectionvalidator. That way we have error sentences there, without affecting other error reporting.

guruz:
@Quatrejas any news?

PR #5690, opened 2017-04-10
When creating explorer favorite use more specific windows functions

remixtj:
Fixes https://github.com/owncloud/client/issues/2719

### DIFF
--- a/src/libsync/utility_win.cpp
+++ b/src/libsync/utility_win.cpp
@@ -36,9 +36,9 @@ static void setupFavLink_private(const QString &folder)
     if(SHGetKnownFolderPath(FOLDERID_Links, 0, NULL, &path) == S_OK) {
         QString links = QDir::fromNativeSeparators(QString::fromWCharArray(path)); 
         linkName = QDir(links).filePath(folderDir.dirName() + QLatin1String(".lnk"));
+    	CoTaskMemFree(path);
     }
     qDebug() << Q_FUNC_INFO << " creating link from " << linkName << " to " << folder;
-    CoTaskMemFree(path);
     if (!QFile::link(folder, linkName))
         qDebug() << Q_FUNC_INFO << "linking" << folder << "to" << linkName << "failed!";
 
--- a/src/libsync/utility_win.cpp
+++ b/src/libsync/utility_win.cpp
@@ -12,9 +12,11 @@
  * for more details.
  */
 
+#define WINVER 0x0600
 #include <shlobj.h>
 #include <winbase.h>
 #include <windows.h>
+#include <winerror.h>
 #include <shlguid.h>
 #include <string>
 #include <QLibrary>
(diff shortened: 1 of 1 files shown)


src/libsync/utility_win.cpp
```diff
@@ -15,21 +15,56 @@
...
+    static SHGetKnownFolderPathFun SHGetKnownFolderPathPtr = NULL;
+    QString linkName; 
     QDir folderDir(QDir::fromNativeSeparators(folder));
-    QString linkName = profile+QLatin1String("/Links/") + folderDir.dirName() + QLatin1String(".lnk");
+    if (!SHGetKnownFolderPathPtr)
+    {
+      QLibrary kernel32Lib("shell32.dll");
+      if(kernel32Lib.load())
+      {
+        SHGetKnownFolderPathPtr = (SHGetKnownFolderPathFun) kernel32Lib.resolve("SHGetKnownFolderPath");
```
jturcotte: This function seems to be supported since Windows Vista, but we require Windows 7 at least. Would it be possible to use normal dynamic linking to access it rather than doing a manual symbol lookup?

remixtj: Last time i tried this i had errors with direct lookup, but maybe because was testing also XP clients. If you say that Windows 7 is the minimum supported, i can try rewriting it

ckamm: @remixtj I'd be great if you could try rewriting to use normal dynamic linking. Feel free to ping me if you need someone to test on Win7.

src/libsync/utility_win.cpp
```diff
@@ -15,21 +15,34 @@
...
-    SHGetSpecialFolderPath(0, path, CSIDL_PROFILE, FALSE);
-    QString profile =  QDir::fromNativeSeparators(QString::fromWCharArray(path));
+    
+    QString linkName; 
     QDir folderDir(QDir::fromNativeSeparators(folder));
-    QString linkName = profile+QLatin1String("/Links/") + folderDir.dirName() + QLatin1String(".lnk");
+     
+    /* Use new WINAPI functions */
+    wchar_t *path = NULL;
+    if(SHGetKnownFolderPath(FOLDERID_Links, 0, NULL, &path) == S_OK) {
```
jturcotte: From the documentation about the `path` argument:

> The calling process is responsible for freeing this resource once it is no longer needed by calling CoTaskMemFree.

Could you verify or fix it?

src/libsync/utility_win.cpp
```diff
@@ -15,21 +15,36 @@
...
+     
+    /* Use new WINAPI functions */
+    PWSTR path;
+    
+    if(SHGetKnownFolderPath(FOLDERID_Links, 0, NULL, &path) == S_OK) {
+        QString links = QDir::fromNativeSeparators(QString::fromWCharArray(path)); 
+        linkName = QDir(links).filePath(folderDir.dirName() + QLatin1String(".lnk"));
+    }
+    qDebug() << Q_FUNC_INFO << " creating link from " << linkName << " to " << folder;
+    CoTaskMemFree(path);
```
jturcotte: This will crash if path is not null and yet not pointing to a valid allocation, which can happen if `SHGetKnownFolderPath` doesn't touch it. You could either move CoTaskMemFree into the `if(){}`, or initialize `path` to `NULL`.

jturcotte:
Argh...

> /var/lib/jenkins/workspace/cloud-client_client_PR-5690-ICN3YAYSR6SITVLPHS5TYBHNKKE7PAVK3ABYGFA5SQ67JL2DZEJA/src/libsync/utility_win.cpp:36:59: error: 'SHGetKnownFolderPath' was not declared in this scope

The CI might not be able to cross-build this...

remixtj:
Hi,

i'm compiling on mingw32-w64 on opensuse.

Maybe my environment is not much up to date, but works.

jturcotte:
That's perfect then, that's our official toolchain. Thanks for the patch and for fixing the build, I'll merge it 👍

Issue #5655, opened 2017-03-27
[OC 10.0] Include multiple-link share capabilities on the sharing dialog


SamuAlfageme:
Since version 1.8, the client includes a sharing dialog that makes fast file/folder sharing available from the file explorer. Since then, many new sharing capabilities have been included to the server, the growing complexity calls for a redesign.

After the inclusion of multiple-link shares on OC 10.0 (https://github.com/owncloud/core/issues/22327) the client could adapt to include this feature in the sharing file/folder dialog. Some options on how to implement this:

### List both user- and link- sharing in the same dialog

This respects the design already used in the sharing dialog: 

<p align="center">
  <img src="https://cloud.githubusercontent.com/assets/2644445/24356874/edfe5e9a-12fb-11e7-993b-60e4c8afc7f4.png"/>
</p>

... however, this many options make the dialog a bit overwhelming and confusing. Notice the inclusion of the permalink (https://github.com/owncloud/client/issues/5023) button to copy the local link and the redistribution of the checkboxes in tables to make the layout cleaner and avoid issues like the `change` button that does nothing here:

<p align="center">
  <img src="https://cloud.githubusercontent.com/assets/2644445/24357903/88bbff48-12ff-11e7-88a1-8897c2191e44.png"/>
</p>

### Server-like design: separate in tabs

This would follow the same idea (list all the created links on a scrollable list) but applying the server's sidebar same approach: separate the two sections (currently done by a horizontal rule) in two tabs user- and link- sharing:

![share_public links_tabs](https://cloud.githubusercontent.com/assets/2644445/24356865/e61b0b56-12fb-11e7-9429-ddc124c4d601.png)

This layout is the best way to further extend the information shown in the dialog (e.g. history, comments, etc...) if considered in the future.

Notice in the list view I've included 2 download buttons to try to address also https://github.com/owncloud/client/issues/5627 and I've removed the hyperlinks everywhere from the mockups since with all the copy-link buttons these are no longer needed (indirectly solves #5626)

## Option B. Hybrid solution

This would open the sharing dialog directly on the browser, abstracting all the sharing logic (both users and links) for a cleaner experience on the desktop. The reason to support this option would be the growing complexity on the dialog, having more and more information to edit and display: 

* Name of the custom link
* Expiration date 
* Copy (Direct download link/In the context)
* Option to delete
* Option to share ... 

[...]

PR #5695, opened 2017-04-11
Multiple link shares and UI adjustments for share dialog

ckamm:
See https://github.com/owncloud/client/issues/5655

For servers that support it:
screen1

For older servers:
screen2

### DIFF
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -27,7 +27,7 @@ set(client_UI
     sharedialog.ui
     sharelinkwidget.ui
     shareusergroupwidget.ui
-    sharewidget.ui
+    shareuserline.ui
     sslerrordialog.ui
     addcertificatedialog.ui
     proxyauthdialog.ui
--- a/src/gui/ocssharejob.cpp
+++ b/src/gui/ocssharejob.cpp
@@ -84,6 +84,16 @@ void OcsShareJob::setPublicUpload(const QString &shareId, bool publicUpload)
     start();
 }
 
+void OcsShareJob::setName(const QString &shareId, const QString &name)
+{
+    appendPath(shareId);
+    setVerb("PUT");
+    addParam(QString::fromLatin1("name"), name);
+    _value = name;
+
+    start();
+}
+
 void OcsShareJob::setPermissions(const QString &shareId, 
                                  const Share::Permissions permissions)
 {
--- a/src/gui/ocssharejob.cpp
+++ b/src/gui/ocssharejob.cpp
@@ -96,7 +96,8 @@ void OcsShareJob::setPermissions(const QString &shareId,
     start();
 }
 
-void OcsShareJob::createLinkShare(const QString &path, 
+void OcsShareJob::createLinkShare(const QString &path,
+                                  const QString &name,
                                   const QString &password)
 {
     setVerb("POST");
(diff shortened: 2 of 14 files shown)


pmaier1:
Great work! Fine for me as it is.

ckamm:
@SamuAlfageme Thanks for checking it out, here are detail comments.

> Server sets the name of the file as default name of the public link.

I didn't do this because then the nice help text in the edit box typically isn't visible. We could have a label and pre-fill the edit box as suggested. I'll try that.

> Links can be created in the web UI with no-name: some users might want just to create an old-school link share and now they're forced to set an explicit name.

Why would someone go "I absolutely want a nameless link share!"? I doubt this is a real use-case and think having a straightforward default-name as you suggested above is sufficient.

> The "Create" button is enabled even when it does nothing

Will fix.

> display the name from the "token" field on the response like the server does? (Also note that the server uses this value to list the public links when no name was set; the new dialog shows "Public link")

I didn't realize it was still possible to create nameless links on the server. Then we must be consistent with the names displayed there. I didn't like the token-based name because it looked too technical, but believe consistency is more important.

> the name field still applies only for creating new links (update the textbox on click and change the button label to "Update" instead of "Create"?)

The flow for creating/updating is hard to get right. If I did as you suggested and made the creation box do double duty as the name-edit box, one could not create new links unless there also was a way to deselect the current share: and it wouldn't be discoverable that deselecting was necessary to create more shares. I think the current way is comparatively clear. Adding a "Link properties" label might be helpful.

> Also, updating the link name by directly clicking its name on the list does not perform the PUT request to update the name on the server.

Yes, there currently is no way of editing the name of a share. I'll look at it.

> Actions for each link are stacked in a drop menu (open, copy, send link), when there's enough room for them in the list view

I used the menu because in my opinion it's easier to discover what the actions do this way, because there's obvious text. The button should probably get a "..." label to make it clearer more actions are hiding there. I have no strong opinion here.

> The detailed permissions are still shown on the "Users and Groups" tab (ref. ckamm@3d92c2c)

This patch was merged to 2.3 and hasn't yet been merged to master. It's just not included in this branch.

ckamm:
@SamuAlfageme Updated with some fixes!

SamuAlfageme:
@ckamm lookin' great 😎 I think it's good to be merged. Last comments: 

> The flow for creating/updating is hard to get right. [...]  Adding a "Link properties" label might be helpful.

+1 to try how a label that separates somehow explicitly a "create" section from an "update" section; It might be also useful if possible to have the ability to deselect any entry on the table by clicking outside/on empty space. 

> I used the menu because in my opinion it's easier to discover what the actions do this way, because there's obvious text. The button should probably get a "..." label to make it clearer more actions are hiding there. I have no strong opinion here.

Agreed on the menu; however I think the most common use case for sharing links is to get copied, so a shortcut could also be nice.

And last difference with the web UI would be that there they have consecutive numbering on the name (by appending `(n)` to the link's default name) and we reuse the same name. I don't think this is necessary to be addressed at all. 

As said, from an user's pov, this is good to go.

guruz:
Code looks ok to me 👍 

However

```
hudson.plugins.git.GitException: Command "git merge 8cb3a77022dfc8c688e3aeaf9fb08b6ddd305521" returned status code 1:
stdout: Auto-merging src/gui/shareusergroupwidget.cpp
Auto-merging src/gui/sharelinkwidget.ui
CONFLICT (content): Merge conflict in src/gui/sharelinkwidget.ui
Auto-merging src/gui/sharedialog.ui
CONFLICT (content): Merge conflict in src/gui/sharedialog.ui
Auto-merging src/gui/sharedialog.cpp
Automatic merge failed; fix conflicts and then commit the result.

```

ckamm:
@guruz Rebased on top of master with some minor adjustments due to conflicting changes.

Issue #5666, opened 2017-03-30
Force sync re-scanning the server contents 


SamuAlfageme:
Sometimes the information on the sync DB is incorrect and the client needs to manually discard parts of this information (e.g. server reported a folder empty) in order to recover. 

We need something like the "Force Sync" button (https://github.com/owncloud/client/issues/2158) but to discard the saved metadata if that's requested. And also not to drop **all** the client's DB (we don't want to loose e.g. the selective sync information if we remove the sync journal) but just the pieces we are interested in discarding while we improve the overall performance to avoid the client reporting different contents than the server. 

cc/ @hodyroff

ckamm:
@SamuAlfageme In which situations would we need this command? All situations where it is useful would be due to bugs, right?

I'm not against having an action like this, as long as it's a hidden debug one. If it were prominently visible, users would press this button in situations that don't warrant it and either get into trouble because of it, or not report bugs because clicking the button fixes the symptoms temporarily.

hodyroff:
Any type of server2client communication issues/bugs, yes. And to recover from them. Like in my situation where the folder exists but the files are missing and somehow the client thinks all is good. To trigger a resync it would be worthwhile. Hidden is fine, but then there is no real harm donw by pressing the button either ... just some load on both desktop and server ...

SamuAlfageme:
@ckamm exactly, maybe launch the client with an additional `-parameter` that adds the option, when right clicking in a folder on the selective sync list (since this would only be necessary for folders), to force this behavior.

dragotin:
reading what @hodyroff says, I would be highly reluctant to hand the user such a tool/button. It could easily lead to unexpected data loss. Experience from the past tell that these situations mostly stem from server problems, and fixing that on the client with such a brutal method is dangerous.

I would rather analyze the problem again, and fix it, rather than covering it by a parameter that users will use desperately.

ogoffart:
We just need to add a (hidden) shortcut or command that calls `SyncJournalDb::forceRemoteDiscoveryNextSync`


For reference, this is the SQL command that one can run directly on the database if the client don't have this command: `UPDATE metadata SET md5='_invalid_' WHERE type=2;`
It clears the etag field for directories and will force a full remote discovery on the next sync.

ckamm:
I'll add something like `Ctrl-F6` for this along the lines of @ogoffart's description.

PR #5696, opened 2017-04-12
Add a "force re-scanning" key sequence

ckamm:
As suggested in #5666

The first commit simplifies related code first.


src/gui/accountsettings.cpp
```diff
@@ -163,6 +163,14 @@ void AccountSettings::createAccountToolbox()
     slotAccountAdded(_accountState);
 }
 
+QString AccountSettings::selectedFolderAlias() const
+{
+    QModelIndex selected = ui->_folderList->selectionModel()->currentIndex();
+    if( !selected.isValid() )
+        return "";
```
ogoffart: You could also return a null string  (QString())

Issue #5691, opened 2017-04-10
status.php: Be lenient when version parsing, also parse version from server capabilities


guruz:
https://github.com/owncloud/core/pull/27473

guruz:
Closing in favor of the PR

PR #5698, opened 2017-04-12
Server: Parse version from capabilities too #5691

guruz:


guruz:
I only parse in ConnectionValidator not in wizard.. is that a problem? I didn't see the version used earlier

src/libsync/connectionvalidator.cpp
```diff
@@ -119,6 +119,8 @@ void ConnectionValidator::slotStatusFound(const QUrl&url, const QVariantMap &inf
              << url << " with version "
              << CheckServerJob::versionString(info)
              << "(" << CheckServerJob::version(info) << ")";
+    // Newer servers don't disclose any version in status.php anymore
+    // https://github.com/owncloud/core/pull/27473/files
 
     QString version = CheckServerJob::version(info);
     _account->setServerVersion(version);
```
ckamm: This should be guarded by a `if (!version.isEmpty())` - I think we run this periodically and it means that we might reset the server version stored in the account to the empty string occasionally!

ogoffart:
We use the version info in the wizzard (httpcredspage) before connecting, and in the password dialog.
So we need to make sure that if no version is reported by status.php, then these behave as if the server is >= 10.0

src/libsync/connectionvalidator.cpp
```diff
@@ -235,6 +237,13 @@ void ConnectionValidator::slotCapabilitiesRecieved(const QVariantMap &json)
     auto caps = json.value("ocs").toMap().value("data").toMap().value("capabilities");
     qDebug() << "Server capabilities" << caps;
     _account->setCapabilities(caps.toMap());
+
+    QString serverVersion = caps.toMap()["core"].toMap()["status"].toMap()["version"].toString();
+    if (!serverVersion.isEmpty()) {
+        _account->setServerVersion(serverVersion);
+        qDebug() << "Version from capabilities instead of status.php:" << serverVersion;
```
ogoffart: Do we need this debug?  the version will already be shown in the caps earlier.

src/libsync/networkjobs.cpp
```diff
@@ -477,7 +477,8 @@ bool CheckServerJob::finished()
         if( status.contains("installed")
                 && status.contains("version")
                 && status.contains("versionstring") ) {
-
+            // Newer servers don't disclose anything in that empty string "version" anymore
```
ogoffart: So then we will show instanceNotFound and we can't connect.  This needs to be changed as well!

ckamm: The server still sends "version" and "versionstring", but it will be empty. I had to look this up too. Expand comment?

guruz: That's why I wrote "empty string" and not "does not contain this specific key"

src/libsync/connectionvalidator.cpp
guruz: You can try reading the capabilities, but to most people it will look like

ogoffart: The thing is that adding debug output only make it looks more like that matrix screen, even for me.   While people who care to see the server version will be able to parse it from other debug output already

src/libsync/networkjobs.cpp
ogoffart: ah ok, so it still has the version string, but it is empty?
Shoud we not make it works also if the version string is not there at all?  (maybe future server remove it completely?)

ckamm:
@ogoffart Could you re-review? In particular you said "We use the version info in the wizzard (httpcredspage) before connecting, and in the password dialog." - but I only found it being used in httpcredspage!

src/libsync/networkjobs.cpp
```diff
@@ -474,10 +474,7 @@ bool CheckServerJob::finished()
         }
 
         qDebug() << "status.php returns: " << status << " " << reply()->error() << " Reply: " << reply();
-        if( status.contains("installed")
-                && status.contains("version")
-                && status.contains("versionstring") ) {
-
+        if( status.contains("installed") ) {
```
ogoffart: So this is still there even when the version is not there?

ogoffart:
:+1:  Looks good

src/libsync/networkjobs.cpp
ckamm: Yes. The presence of the other attributes is currently discussed at https://github.com/owncloud/core/issues/27669

PR #5711, opened 2017-04-19
Fix a crash in ProxyAuthHandler

ckamm:
See
https://sentry.io/owncloud/desktop-win-and-mac/issues/243433178/activity/
https://sentry.io/owncloud/desktop-win-and-mac/issues/234182688/activity/

The problem was that an account's QNetworkAccessManager can be deleted
when reentering the event loop.


guruz:
There seems to be a crash still
https://sentry.io/owncloud/desktop-win-and-mac/issues/346938388/

PR #5725, opened 2017-04-26
SqlDatabase: Ask freeSpace for directory, fix for Windows

ckamm:
See owncloud/enterprise#1955


guruz:
👍 
did you check the other usages of freeDiskSpace if they possibly use a file?

ckamm:
@guruz Yes, the other uses are all fine.

Issue #5710, opened 2017-04-19
Port usage of src/3rdparty/qjson/json.cpp to Qt 5 JSON library


guruz:

ckamm:
Closing, see PR https://github.com/owncloud/client/issues/5727

guruz:
ReadyToTest, PR already merged
https://github.com/owncloud/client/pull/5727#issuecomment-297400809

PR #5727, opened 2017-04-26
Remove old 3rdparty QtJson code in favor of Qt5's QJsonDocument

ckamm:
For https://github.com/owncloud/client/issues/5710


guruz:
Looks ok to me 👍  I've tested, connecting, activity view 

Maybe @SamuAlfageme could do a smoke test if adding a shibboleth account in wizard still works and if the client will display the user name in the tool bar for shibboleth. And sharing :)

BTW, maybe you could have avoided to change the API by using ```toVariantMap```. But it's all good now if it works, code looks fine

ckamm:
Yep, affected areas are status.php, notifications (incoming remote shares etc.), sharing, shibboleth, activity view. I've tested all of them, but having another person look is certainly appreciated.

@guruz I thought about just using `toVariantMap` (I did that for capabilities, because that way it's easier to mock if necessary), but using the actual json should be more correct (for instance I'm not sure whether we can distinguish null and undefined otherwise).

src/gui/activitylistmodel.cpp
```diff
@@ -152,17 +154,19 @@ void ActivityListModel::slotActivitiesReceived(const QVariantMap& json, int stat
     _currentlyFetching.remove(ast);
 
     foreach( auto activ, activities ) {
-        auto json = activ.toMap();
+        auto json = activ.toObject();
+
+        qDebug() << "ZZZ" << json;
```
SamuAlfageme: @ckamm 💤? 😄

ckamm: Oops, will get rid of that.

Issue #5726, opened 2017-04-26
[OS X][Overlays] Finder extension does not work until the last page of the wizard is closed


SamuAlfageme:
### Expected behavior
Once the account is set up, finder extension should start working on the local sync folder. 

### Actual behavior
Neither overlay icons **inside the folder**, nor the "Share with ownCloud" context menu entry are available until the last page of the wizard is closed. 

![apr-26-2017 11-46-02](https://cloud.githubusercontent.com/assets/2644445/25428406/0eee6c4e-2a76-11e7-93f3-c8afadd7135e.gif)

### Steps to reproduce
1. Complete account setup but leave the "Everything set up!" page opened. 
2. Open the local sync folder on the finder.

guruz:
@SamuAlfageme didn't we say we'll just remove the last wizard page and pop up the config dialog?

guruz:
Please re-test with the last wizard page removed :-)

SamuAlfageme:
After closing https://github.com/owncloud/client/issues/5764 I think we've run every possible scenario affected by the last page removal. 

Continue the discussion about the page's dead code in the PR?

guruz:
@settermjd FYI https://doc.owncloud.org/desktop/2.2/installing.html the https://doc.owncloud.org/desktop/2.2/_images/client-4.png wizard page has been removed.

PR #5729, opened 2017-04-26
Wizard: Don't show not-so-useful result page #5726

guruz:
Better immediately show the settings dialog where people can see what's going on.
They can go to sync folder or webinterface from there.

### DIFF
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -333,24 +333,6 @@ void Application::slotownCloudWizardDone( int res )
     folderMan->setSyncEnabled(true);
 
     if( res == QDialog::Accepted ) {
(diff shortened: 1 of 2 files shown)


guruz:
On OS X this goes to sync protocol for me after the wizard.
On Windows/Linux I don't know.

ogoffart:
Code looks good.  Did you check it still works with the "skip folder configuration" button?

ogoffart:
But how about removing the page completely from git?  There is no need to have dead code.

guruz:
> But how about removing the page completely from git? There is no need to have dead code.

I wanted to wait with bigger changes until we get the wizard code from @danimo ;-)

Issue #5596, opened 2017-03-10
Allow context menu to open the local Folder/subfolder in selective sync list in main window


guruz:

ckamm:
@guruz Do you mean for the sync folder or for the individual selective-sync folders?

It seems we'll have to manually disambiguate single and double clicks for that.

guruz:
I meant individual folders.

guruz:
Apparantly double click detection is tricky for Qt/legacy reasons, let's have it in a right-mouse button menu.

guruz:
Needs a quick test on Windows

SamuAlfageme:
Lookin' neat: 

![opensubfolder](https://cloud.githubusercontent.com/assets/2644445/25800377/5c497e62-33e9-11e7-9dc0-1d51af303087.gif)

PR #5730, opened 2017-04-26
Selective Sync: Open sub folder context menu #5596

guruz:


src/gui/accountsettings.cpp
```diff
@@ -403,6 +426,16 @@ void AccountSettings::slotOpenCurrentFolder()
     }
 }
 
+void AccountSettings::slotOpenCurrentLocalSubFolder()
+{
+    QModelIndex selected = ui->_folderList->selectionModel()->currentIndex();
+    if( !selected.isValid() )
```
ckamm: Add a check for `_model->classify(indx) == FolderStatusModel::SubFolder`?

src/gui/accountsettings.cpp
```diff
@@ -403,6 +426,16 @@ void AccountSettings::slotOpenCurrentFolder()
     }
 }
 
+void AccountSettings::slotOpenCurrentLocalSubFolder()
+{
+    QModelIndex selected = ui->_folderList->selectionModel()->currentIndex();
```
ckamm: I'm slightly worried that the selected item could be different from the one that was clicked on. I couldn't make that happen in my tests though.

ckamm:
Looks good and works in my tests, added two minor comments.

ogoffart:
You should move your code in AccountSettings::slotCustomContextMenuRequested  rather than in AccountSettings::slotFolderListMousePressed.  That way, it also work with the "open menu" key on the keyboard. (Accessibility)

src/gui/accountsettings.cpp
ogoffart: I think that's fine.

Altough there is no need to go trough the selectionModel:  `ui->_folderList->currentIndex()` should work

src/gui/accountsettings.cpp
ogoffart: Normally, that function should only be called in this case. But it's true we could play it save.

src/gui/accountsettings.h
```diff
@@ -75,7 +75,8 @@ protected slots:
     void slotScheduleCurrentFolderForceRemoteDiscovery();
     void slotForceSyncCurrentFolder();
     void slotRemoveCurrentFolder();
-    void slotOpenCurrentFolder();
+    void slotOpenCurrentFolder(); // sync folder
+    void slotOpenCurrentLocalSubFolder(); // selected subfolder in sync folder
```
ogoffart: (Maybe the two function could be merged in one)

guruz:
@ckamm @ogoffart Please re-review.
I didn't see/know about a context menu signal even..

ogoffart:
:+1:

PR #5731, opened 2017-04-27
SSL error handling: Remove unnecessary confusing code

ckamm:
I'm confident this is unnecessary. The original bug in #3283 was
to call ignoreSslErrors() without an argument in the 'accept'
case, which meant ignoring *all* subsequent SSL errors.

With that fixed, explicitly aborting the reply and resetting QNAM
is not needed since not ignoring the error will lead to the SSL
handshake failing.

See also:
  75b38d1a2ffe57d0f1eb3ebb8c5f30b8b2a185e4 (workaround introduced)
  89376e14d6135a6f39a6df99d54fde253573575c (real fix)
  76ce5adbf02af052ab8d7596b8ded75c7d4d7fcb (cherry-pick of workaround)

### DIFF
--- a/src/libsync/account.cpp
+++ b/src/libsync/account.cpp
@@ -366,11 +366,8 @@ void Account::slotHandleSslErrors(QNetworkReply *reply , QList<QSslError> errors
                 _rejectedCertificates.append(error.certificate());
             }
         }
-        // if during normal operation, a new certificate was MITM'ed, and the user does not
-        // ACK it, the running request must be aborted and the QNAM must be reset, to not
-        // treat the new cert as granted. See bug #3283
-        reply->abort();
-        resetNetworkAccessManager();
+
+        // Not calling ignoreSslErrors will make the SSL handshake fail.
         return;
     }
 }


ogoffart:
Does anyone knows why was the workaround cherry-picked again?

ckamm:
@ogoffart No :/

My main source of confidence about this not being necessary is: If it was mandatory to abort and reset, then this would be a major security bug in Qt that everyone would run into and that would be fixed quickly.

The documentation on the sslErrors() signal literally says 
> To indicate that the errors are not fatal and that the connection should proceed, the QNetworkReply::ignoreSslErrors() function should be called from the slot connected to this signal. If it is not called, the SSL session will be torn down before any data is exchanged (including the URL)."

guruz:
👍 even though I'm a fan of viel-hilft-viel..

Discussion point: @danimo @dragotin @ogoffart @pmaier1 @michaelstingl In general, now in the time of letsencrypt & co maybe we should remove the whole cert dialogue?
If users want an unsigned cert, they could still either import it into their OS or set the cert hash in the config file or so...

PR #5668, opened 2017-03-30
Authentication with OAuth2

ogoffart:
When the OAuth2 app (https://github.com/owncloud/oauth2) is enabled,
We will open a browser and perform the OAuth2 authentication

Issue: #4798 and https://github.com/owncloud/platform/issues/17

PR #5732, opened 2017-04-28
Reformatting with clang-format

ckamm:
The style differences were getting on my nerves. Now it's pretty nice, and consecutive invocations of clang-format have no effect on the code.

### DIFF
--- a/src/cmd/cmd.h
+++ b/src/cmd/cmd.h
@@ -22,12 +22,17 @@
  * @brief Helper class for command line client
  * @ingroup cmd
  */
-class Cmd : public QObject {
+class Cmd : public QObject
+{
     Q_OBJECT
 public:
-    Cmd() : QObject() { }
+    Cmd()
+        : QObject()
+    {
+    }
 public slots:
-    void transmissionProgressSlot() {
+    void transmissionProgressSlot()
+    {
     }
 };
 
--- a/src/gui/accountmanager.cpp
+++ b/src/gui/accountmanager.cpp
@@ -50,18 +50,18 @@ bool AccountManager::restore()
     auto settings = Utility::settingsWithGroup(QLatin1String(accountsC));
     if (settings->status() != QSettings::NoError) {
         qCWarning(lcAccountManager) << "Could not read settings from" << settings->fileName()
-                 << settings->status();
+                                    << settings->status();
         return false;
     }
 
     // If there are no accounts, check the old format.
     if (settings->childGroups().isEmpty()
-            && !settings->contains(QLatin1String(versionC))) {
+        && !settings->contains(QLatin1String(versionC))) {
         restoreFromLegacySettings();
         return true;
     }
 
-    foreach (const auto& accountId, settings->childGroups()) {
+    foreach (const auto &accountId, settings->childGroups()) {
         settings->beginGroup(accountId);
         if (auto acc = loadAccountHelper(*settings)) {
             acc->_id = accountId;
@@ -78,40 +78,44 @@ bool AccountManager::restore()
 bool AccountManager::restoreFromLegacySettings()
 {
     qCInfo(lcAccountManager) << "Migrate: restoreFromLegacySettings, checking settings group"
-             << Theme::instance()->appName();
+                             << Theme::instance()->appName();
 
     // try to open the correctly themed settings
     auto settings = Utility::settingsWithGroup(Theme::instance()->appName());
 
     // if the settings file could not be opened, the childKeys list is empty
     // then try to load settings from a very old place
-    if( settings->childKeys().isEmpty() ) {
+    if (settings->childKeys().isEmpty()) {
         // Now try to open the original ownCloud settings to see if they exist.
-        QString oCCfgFile = QDir::fromNativeSeparators( settings->fileName() );
+        QString oCCfgFile = QDir::fromNativeSeparators(settings->fileName());
         // replace the last two segments with ownCloud/owncloud.cfg
-        oCCfgFile = oCCfgFile.left( oCCfgFile.lastIndexOf('/'));
-        oCCfgFile = oCCfgFile.left( oCCfgFile.lastIndexOf('/'));
+        oCCfgFile = oCCfgFile.left(oCCfgFile.lastIndexOf('/'));
+        oCCfgFile = oCCfgFile.left(oCCfgFile.lastIndexOf('/'));
         oCCfgFile += QLatin1String("/ownCloud/owncloud.cfg");
 
         qCInfo(lcAccountManager) << "Migrate: checking old config " << oCCfgFile;
 
-        QFileInfo fi( oCCfgFile );
-        if( fi.isReadable() ) {
+        QFileInfo fi(oCCfgFile);
+        if (fi.isReadable()) {
             std::unique_ptr<QSettings> oCSettings(new QSettings(oCCfgFile, QSettings::IniFormat));
             oCSettings->beginGroup(QLatin1String("ownCloud"));
 
             // Check the theme url to see if it is the same url that the oC config was for
             QString overrideUrl = Theme::instance()->overrideServerUrl();
-            if( !overrideUrl.isEmpty() ) {
-                if (overrideUrl.endsWith('/')) { overrideUrl.chop(1); }
+            if (!overrideUrl.isEmpty()) {
+                if (overrideUrl.endsWith('/')) {
+                    overrideUrl.chop(1);
+                }
                 QString oCUrl = oCSettings->value(QLatin1String(urlC)).toString();
-                if (oCUrl.endsWith('/')) { oCUrl.chop(1); }
+                if (oCUrl.endsWith('/')) {
+                    oCUrl.chop(1);
+                }
 
                 // in case the urls are equal reset the settings object to read from
                 // the ownCloud settings object
                 qCInfo(lcAccountManager) << "Migrate oC config if " << oCUrl << " == " << overrideUrl << ":"
-                         << (oCUrl == overrideUrl ? "Yes" : "No");
-                if( oCUrl == overrideUrl ) {
+                                         << (oCUrl == overrideUrl ? "Yes" : "No");
+                if (oCUrl == overrideUrl) {
                     settings = std::move(oCSettings);
                 }
             }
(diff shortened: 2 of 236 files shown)


guruz:
I don't like those kind of commits because they clutter the ```git blame```. But I guess one could use ```-w``` and then one would not see this showing up?

ckamm:
@guruz Yes, `-w` gets rid of the noise for whitespace-only changes. That reduces the number of blame-affected lines from 9000 to 4000.

So there's still a good amount of noise where the reformatting went beyond whitespace, but I think it's worth it since makes the whole codebase more consistent and appealing and permanently (with the hook) solves any style questions.

Also, while plain `git blame` doesn't support it, these is `git hyper-blame` (https://commondatastorage.googleapis.com/chrome-infra-docs/flat/depot_tools/docs/html/git-hyper-blame.html) which supports having a list of sha1s to skip for blame purposes. I'd add such a `.git-blame-ignore-revs` file once this is applied.

jturcotte:
I wouldn't mind a better style, but since yours is automated, could you make sure to wait that https://github.com/owncloud/client/pull/5671 lands before pushing this? I'm afraid this would make the rebase more horrible than necessary.

ckamm:
@jturcotte Sure, not a problem.

ckamm:
@guruz @jturcotte It's rebased on current master. Does it look good to you?

jturcotte:
@ckamm Is there a way to test the code for conformance to the new format? We could add a step in Jenkinsfile to fail the build if the format is wrong maybe?

ckamm:
@jturcotte Personally, I'm using git-clang-format along with this kind of pre-commit hook:
```
#!/usr/bin/env python

import subprocess
import os

exit(0) # Still disabled

if 'SKIP_FORMAT_CHECK' in os.environ:
    exit(0)

output = subprocess.check_output(["git", "clang-format", "--diff"])

if output not in ['no modified files to format\n', 'clang-format did not modify any files\n']:
    print "Run git clang-format, then commit.\n"
    exit(1)
else:
    exit(0)
```

But adding a check to jenkins would be very helpful for PRs. Who would I need to talk to?

jturcotte:
@ckamm Could we also set MaxEmptyLinesToKeep=2 and KeepEmptyLinesAtTheStartOfBlocks=false?

About the jenkinsfile check I could add it.

ckamm:
@jturcotte I agree about these, will adjust.

src/cmd/cmd.h
```diff
@@ -22,13 +22,16 @@
...
+{
     Q_OBJECT
 public:
-    Cmd() : QObject() { }
-public slots:
-    void transmissionProgressSlot() {
+    Cmd()
+        : QObject()
+    {
     }
```
jturcotte: Do you know what would cause this? It looked better before IMO.

src/gui/accountmanager.cpp
```diff
@@ -78,40 +77,45 @@ bool AccountManager::restore()
 bool AccountManager::restoreFromLegacySettings()
 {
     qCInfo(lcAccountManager) << "Migrate: restoreFromLegacySettings, checking settings group"
-             << Theme::instance()->appName();
+                             << Theme::instance()->appName();
```
jturcotte: I don't know what we should choose, but I find that aligning with the top token often breaks over time and prefer to just align it one block more than the wrapping brackets (could be done with `AlignAfterOpenBracket: DontAlign`). What do you think?

src/gui/accountmanager.cpp
```diff
@@ -78,40 +77,45 @@ bool AccountManager::restore()
...
+        oCCfgFile = oCCfgFile.left(oCCfgFile.lastIndexOf('/'));
+        oCCfgFile = oCCfgFile.left(oCCfgFile.lastIndexOf('/'));
         oCCfgFile += QLatin1String("/ownCloud/owncloud.cfg");
 
         qCInfo(lcAccountManager) << "Migrate: checking old config " << oCCfgFile;
 
-        QFileInfo fi( oCCfgFile );
-        if( fi.isReadable() ) {
+        QFileInfo fi(oCCfgFile);
+        if (fi.isReadable()) {
```
jturcotte: I'm so glad we can make/keep this stuff consistent.

src/gui/accountmanager.cpp
```diff
@@ -347,15 +355,13 @@ QString AccountManager::generateFreeAccountId() const
...
 }
 
-void AccountManager::addAccountState(AccountState* accountState)
+void AccountManager::addAccountState(AccountState *accountState)
 {
-    QObject::connect(accountState->account().data(),
-                     SIGNAL(wantsAccountSaved(Account*)),
-                     SLOT(saveAccount(Account*)));
+    QObject::connect(accountState->account().data(), SIGNAL(wantsAccountSaved(Account *)),
+                     SLOT(saveAccount(Account *)));
```
jturcotte: Humm that's a bit weird too, but I guess it's fine.

src/gui/accountmanager.h
```diff
@@ -55,14 +56,15 @@ class AccountManager : public QObject {
 
     /**
      * Return a list of all accounts.
-     * (this is a list of QSharedPointer for internal reasons, one should normally not keep a copy of them)
+     * (this is a list of QSharedPointer for internal reasons, one should normally not keep a copy
+     * of them)
```
jturcotte: What would you think about not setting `ColumnLimit`? Or maybe set it to 120, people have large screens nowadays.

src/gui/authenticationdialog.cpp
```diff
@@ -39,7 +41,8 @@ AuthenticationDialog::AuthenticationDialog(const QString &realm, const QString &
     lay->addLayout(form);
     _password->setEchoMode(QLineEdit::Password);
 
-    QDialogButtonBox *box = new QDialogButtonBox(QDialogButtonBox::Ok|QDialogButtonBox::Cancel, Qt::Horizontal);
+    QDialogButtonBox *box =
+        new QDialogButtonBox(QDialogButtonBox::Ok | QDialogButtonBox::Cancel, Qt::Horizontal);
```
jturcotte: This one for example would be better not line-broken as well.

src/gui/accountmanager.h
jturcotte: I'd personally vote for not setting it and let the people decide how to line-break, long lines never really annoyed me in the code.

src/gui/accountmanager.cpp
ckamm: @jturcotte I think token alignement looks nice and clean but breaks easily and is annoying to reformat. If I aligned this manually, I'd not do token alignment because `qCInfo(lcAccountManager)` is too long for that. 

What if we did `AlwaysBreak`? Gain neatness at the expense of shorter lines and more vertical space used.

src/gui/accountmanager.h
ckamm: I'd rather not (my editor window width is around 105 characters, for example)

src/cmd/cmd.h
ckamm: I don't think there's a `AllowCtorSingleLine`, so probably it's just the generic formatting for these.

src/gui/accountmanager.h
guruz: For me on laptop it ends at "copy" (col 99)

src/gui/accountmanager.cpp
jturcotte: @ckamm I think we mean the same thing, isn't what `DontAlign` would give us? The name is misleading, the documentation shows the following example:

```
someLongFunction(argument1,
    argument2);
```

The example for `AlwaysBreak` doesn't look nice I think:
```
someLongFunction(
    argument1, argument2);
```

src/gui/accountmanager.h
jturcotte: I prefer having to scroll horizontally in that case, but alright I'm fine keeping it at 100.

src/cmd/cmd.h
jturcotte: OK, in this specific example we could just remove the whole thing anyway...

src/gui/accountmanager.cpp
ckamm: @jturcotte DontAlign gives me lines like
```
    QObject::connect(
        _requestEtagJob, SIGNAL(etagRetreived(QString)), this, SLOT(etagRetreived(QString)));
    connect(
        _engine.data(), SIGNAL(finished(bool)), SLOT(slotSyncFinished(bool)), Qt::QueuedConnection);
```
I guess clang-format realizes this have to be two lines and then prefers not splitting the argument list. And yet, AlwaysBreak has a bunch of not-pretty cases. I could live with the current setting or with DontAlign -- @ogoffart, do you have opinions?

src/gui/accountmanager.h
ckamm: @jturcotte Not setting the column limit is good - I agree that's not been an annoyance so far. Also reduces the number of affected lines from 9k to 7k.

src/gui/accountmanager.cpp
jturcotte: @ckamm Humm OK, and is there a way to just tell it not to worry about that?

Also maybe worth trying `DontAlign ` together with `BinPackArguments: false`.

ckamm: @jturcotte Actually, clang-format seems to align `<<` no matter what that setting is. I recommend looking at the patchset for both cases. (alwaysbreak reformats everything to be on one line if columnlimit is disabled) I'm currently leaning towards DontAlign.

jturcotte: I tried a few things in your branch and it's a bit annoying that clang-format needs to apply the rules no matter what (like we can't seem to tell it to jut not touch the code already there for some criteria). But playing around I think this gives the results I like the most:

```
ColumnLimit: 100
AlignAfterOpenBracket: DontAlign
BinPackArguments: false
```

So I think it would be nice to add `BinPackArguments: false` and I prefer `AlignAfterOpenBracket: DontAlign` over `true` because of the issue we talked above (and some people will continue to indent it manually, I don't think everybody will install clang-format).

guruz:
When is the right moment to merge something like this?

We still have @ogoffart 's big OAuth2 patch pending https://github.com/owncloud/client/pull/5668

ckamm:
@guruz Since a large chunk of this patch is just applying clang-format, it's easy to rebase. Delaying until oauth2 is merged is fine with me - though that patch isn't huge and could be rebased if necessary.

Issue #1067, opened 2013-10-03
Directory tree depth limit


moscicki:
Above 50 levels of directories the sync breaks however the sync client does not report an error (RC=0).

If there is a policy on a maximum allowed tree depth then it should be handled somehow properly.

hodyroff:
Does this still exist as an issue?

ogoffart:
yes

hodyroff:
@pmaier1 @michaelstingl to define wanted behaviour and then lets implement it. In any case, when we reach the limit we should have a defined behaviour and not just break.

michaelstingl:
Limit 50 seems to work. We should display a error message in the UI.

michaelstingl:
@ogoffart Milestone?

michaelstingl:
@ogoffart Could me make this configurable with a environment variable?

pmaier1:
For now we think a limit of 50 is fine - at least until we hear something else. Anyway we need proper error handling and docs that explain the reason for such limit (symlink loop limiting). In case the directory depth exceeds the limit, the client should show an error message like "Could not be synced due to exceeded directory depth (50). See <link to docs>".

michaelstingl:
What needs to be tested? @SamuAlfageme ? Is there already an error message as @pmaier1 suggested?

ckamm:
I've tested the current behavior: The too-deep directory is not synced and is shown in the "not synced" list of files. It's a silent error, similar to other files being excluded due to hiddenness or by exclude pattern.

hodyroff:
Perfect IMO, if anybody thinks it should be more then 50, please raise your voice now! We can test more and up the limit ...

ckamm:
@hodyroff Note that 13705999f731042965bae3caa43b7245e7b369e0 increased the limit to 100 and I've updated the docs accordingly in https://github.com/owncloud/client/commit/096cd348f0d1616576bc84cc3b0e282dd26b671b

PR #5735, opened 2017-04-28
Discovery: Increase the MAX_DEPTH and show deep folder as ignored

ogoffart:
Before this patch, to deep folder would just be ignored, without any feedback.
This patch makes it so deep folder are properly shown as ignored in the UI.

Also increase the MAX_DEPTH

Issue: #1067


csync/src/csync.h
```diff
@@ -99,7 +99,8 @@ enum csync_status_codes_e {
     CSYNC_STATUS_INDIVIDUAL_EXCLUDE_HIDDEN,
     CSYNC_STATUS_INVALID_CHARACTERS,
     CSYNC_STATUS_INDIVIDUAL_STAT_FAILED,
-    CSYNC_STATUS_FORBIDDEN
+    CSYNC_STATUS_FORBIDDEN,
+    CSYNC_STATUS_INVIDUAL_TOO_DEEP
```
phil-davis: s/INVIDUAL/INDIVIDUAL
through the whole PR code

guruz:
FYI @settermjd for updating https://github.com/settermjd/client/commit/e7546903cc027a282f68a9844107a8abd0816ece

FYI @PVince81 @DeepDiver1975 if they have a comment regarding server side (yes, storage specific..)

PR #5741, opened 2017-05-02
Document the checksums addition in 10.0

settermjd:
This PR:

- documents the new checksums addition in 10.0

### Relates To 

owncloud/documentation/issues/2964


doc/architecture.rst
```diff
@@ -143,16 +143,135 @@ the remote file will be downloaded and saved as message.txt.
...
+  </d:error>
+
+The client retries the upload using exponential back-off. 
+On success (matching checksum) the computed checksums are stored by the server in ``oc_filecache`` alongside the file.
+
+Chunked Upload
+~~~~~~~~~~~~~~
+
+Mostly same as above. 
+The checksum of the full file is sent with every chunk, but the server compares the checksum of the last chunk.
```
phil-davis: s/of the last chunk/after receiving the last chunk/
or
"but the server compares the checksum received with the last chunk to the server-calculated checksum of the whole file"
I just feel like it needs some words to be explicit that the checksum in the last chunk is the checksum of the whole file.

settermjd: Will see what I can do. Thanks for the feedback @phil-davis.

settermjd: @phil-davis, on reading through the text along with your feedback, is the text either wrong or misleading? The reason I'm asking that is that, excuse my misunderstanding, the text states: 

> The checksum of the full file is sent with every chunk

but you've said: 

> the checksum in the last chunk is the checksum of the whole file

So, should the text be updated to say something like:

> A checksum of every chunk is sent along with it, but the checksum with the last chunk is the checksum of the whole file

phil-davis: I believe "The checksum of the full file is sent with every chunk" is correct.
It is just that when then reading further to "compares the checksum of the last chunk", it is ambiguous - because actually it is not "the checksum of the last chunk" but "the checksum sent with the last chunk".
Clear as mud???

phil-davis: @guruz mentions/queries this also "You mean compares AFTER the last chunk right?" in comment https://github.com/owncloud/documentation/issues/2964#issuecomment-291235760

phil-davis: The expected behavior is documented by tests like:
https://github.com/owncloud/core/blob/b0fa71d74a8ce0a7cc3cc18e9901b5ace1f251c0/tests/integration/features/checksums.feature#L118 - the same checksum (of the whole file) is sent with every chunk.
https://github.com/owncloud/core/blob/b0fa71d74a8ce0a7cc3cc18e9901b5ace1f251c0/tests/integration/features/checksums.feature#L145 - actually the server expects the same checksum to be provided with every chunk. e.g. if the client is to lazy to send the full-file checksum with every chunk, and just sends it with the last chunk, then even though the server can calculate that the checksum is correct, the server is going to reject the file.

settermjd: @phil-davis thanks for the clarification.

settermjd:
@phil-davis update made. Ready to merge?

PR #5747, opened 2017-05-04
Sharing: Optionally show warning that links are public

guruz:
For https://github.com/owncloud/enterprise/issues/1914

screen shot 2017-05-04 at 11 29 38


guruz:
@michaelstingl @pmaier1 Looks ok?

Now in retrospect I'm thinking maybe this message should be default?

michaelstingl:
> I'm thinking maybe this message should be default?

You are right. In this view it don't hurt too much.

michaelstingl:
Slightly changed plan:

We will get a capability for this on the server side. Then we don't need a ownBrander option.(see https://github.com/owncloud/enterprise/issues/1914#issuecomment-299157208)

Issue #5701, opened 2017-04-13
Remove .htaccess from ignored file list


wedi:
Remove `.htaccess` from ignored files list.

### Reasoning
It is not possible to completely backup a web project that requires a `.htaccess` file. There is no way to remove `.htaccess` from the ignored files list as it is included in [sync-exclude.lst](https://github.com/owncloud/client/blob/280906eab990eadb94eefc2e7514a3f5855d25ad/sync-exclude.lst#L25).
**`.htaccess` is a configuration file.** It seems to be the only configuration file in the list. All other entries care about OS cruft and tmp files and stuff.

### Example
For a university team project we are developing a web application. For mod_rewrite among others some server settings need to be changed using a `.htaccess` file. We don’t want to use `http.conf` here because we need to have these settings synced between the team for the application to run.

### Expected behavior
`.htaccess` should be synced when enabling _Synchronize hidden files_.

### Actual behaviour
`.htaccess` is _always_ ignored.

### Steps to reproduce
Add a `.htaccess` file to a folder, try to sync.

### Client configuration
Client version: Version 2.3.1 (build 1602)
`GIT-Revision 433ba0 on Mar 28 2017, 03:34:43 uses Qt 5.6.2, OpenSSL 1.0.2k 26 Jan 2017`

Operating system: Mac OS X 10.11.6

OS language: German

Qt version used by client package (Linux only, see also Settings dialog): Qt 5.6.2

Installation path of client: `/Application`

guruz:
FYI @pmaier1 @michaelstingl I remember some discussion

pmaier1:
> There is no way to remove .htaccess from the ignored files list

You can edit the local `sync-exclude.lst` if you want to change that.

The thing is that .htaccess files that are deeper in the hierarchy can override the .htaccess file ownCloud provides in it's main directory which _can_ cause security issues afaik. We need to check if this can be disabled in the main .htaccess or if there's another way to prevent it. Also the server should decide which files _really_ can't be synced.

In general all files that do not have a real justification for not being synced (like '.part' which is not accepted by the server for architecture reasons) should be removed from the list or not be excluded by default to avoid files not being synced without the user noticing.

I think we also agree on switching the default of 'Sync hidden files' to enabled as discussed [here](https://github.com/owncloud/client/issues/4719)

michaelstingl:
`.htaccess` is currently blacklisted on the server too:
[...]

PR #5748, opened 2017-05-05
Exclude list: remove .htaccess

ogoffart:
As per issue #5701, if the server does not support it, let the server
show return an error, but we should not blacklist it localy

### DIFF
--- a/sync-exclude.lst
+++ b/sync-exclude.lst
@@ -21,7 +21,6 @@ System Volume Information
 .fseventd
 .apdisk
 
-.htaccess
 .directory
 
 *.part


ogoffart:
When trying to upload .htaccess, I get a 507 insufficient storage exception from the server.

guruz:
> When trying to upload .htaccess, I get a 507 insufficient storage exception from the server.

Not necessarily a nice error, but good enough for here 👍 

FYI @PVince81

pmaier1:
> Not necessarily a nice error, but good enough for here

Without proper errors the user does not know what went wrong. We need something like for '.part' files:

'".htaccess" could not be uploaded ('.htaccess' is a forbidden file extension)'

guruz:
@pmaier1 Those messages are created on the server, e.g. the server sent the Insufficient Storage exception.

i meant "good enough" because the use case (.htaccess) is quite exotic and from client side the patch is enough. :)

pmaier1:
> @pmaier1 Those messages are created on the server, e.g. the server sent the Insufficient Storage exception.
i meant "good enough" because the use case (.htaccess) is quite exotic and from client side the patch is enough. :)

Yes, I thought so. Then let's declare it as 'good enough' for now. Added to the list for future error handling improvements.

Issue #4485, opened 2016-02-18
If you set to maintenance mode, on the Desktop It could say that the server is down for maintenance


Dianafg76:
### Expected behaviour

It could say that the server is down for maintenance

![screen shot 2016-02-18 at 10 34 31](https://cloud.githubusercontent.com/assets/12032058/13139209/b7e28402-d62b-11e5-88eb-9b82fa6945d4.png)
### Actual behaviour

You can see the Admin user with grey icon
![screen shot 2016-02-18 at 10 33 25](https://cloud.githubusercontent.com/assets/12032058/13139100/27ab2db2-d62b-11e5-900d-5b575c2aef43.png)
### Steps to reproduce
1. Go to the Desktop with Admin
2. Go to the Server web with Admin
3. Set to maintenance mode
4. Go to the Desktop with Admin
5. Go to Seeting
6. Select the Admin user
### Server configuration

Desktop v 2.2.0

SamuAlfageme:
Maintenance mode support is about to get introduced in the mobile apps; https://github.com/owncloud/ios/pull/821 and https://github.com/owncloud/android/issues/1722

iOS maintenance mode error message for uniformity: https://github.com/owncloud/ios/blob/1cbb881ba2e010e22fcaeda9a124a3fccb66c7ec/Owncloud%20iOs%20Client/SupportingFiles/Languages/en.lproj/Localizable.strings#L111

SamuAlfageme:
Also regarding maintenance mode operation on the client: https://github.com/owncloud/client/issues/5738

ckamm:
Pull request: https://github.com/owncloud/client/pull/5755

PR #5755, opened 2017-05-08
Detect maintenance mode #4485

ckamm:
When we first detect a 503 (probably from a PROPFIND) and enter the
ServiceUnavailable state, we new trigger a status.php query that will
switch the state to MaintenanceMode if necessary.

### DIFF
--- a/src/gui/accountstate.cpp
+++ b/src/gui/accountstate.cpp
@@ -94,6 +94,14 @@ void AccountState::setState(State state)
             _connectionStatus = ConnectionValidator::Undefined;
             _connectionErrors.clear();
         } else if (oldState == SignedOut && _state == Disconnected) {
+            // If we stop being voluntarily signed-out, try to connect and
+            // auth right now!
+            checkConnectivity();
+        } else if (_state == ServiceUnavailable) {
+            // Check if we are actually down for maintenance.
+            // To do this we must clear the connection validator that just
+            // produced the 503. It's finished anyway and will delete itself.
+            _connectionValidator.clear();
             checkConnectivity();
         }
         if (oldState == Connected || _state == Connected) {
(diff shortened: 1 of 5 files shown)


src/gui/accountstate.cpp
```diff
@@ -94,6 +94,14 @@ void AccountState::setState(State state)
             _connectionStatus = ConnectionValidator::Undefined;
             _connectionErrors.clear();
         } else if (oldState == SignedOut && _state == Disconnected) {
+            // If we stop being voluntarily signed-out, try to connect and
```
guruz: Is this related to the maintenance mode?

ckamm: Nah, I just sneaked in this comment because I felt it was missing. It's technically unrelated.

SamuAlfageme:
@ckamm I'm seeing `AccountState::ServiceUnavailable`'s Connection Label (`Server <server> is temporarily unavailable.`) on the account tab after setting the maintenance mode in oC 10.0.2. The value is correctly set in `status.php` but the client seems to be ignoring it 😕  

(I'm on [`2.4.0git 03aa7e`](https://github.com/owncloud/client/commit/03aa7e30449e6777fc23a96ff315020ed33da05b) revision)

ckamm:
@SamuAlfageme You're *not* seeing it, I guess? Will test again.

SamuAlfageme:
@ckamm oh, no. I mean by looking at 0ba33e0f399db04f6e5cc4a5fed99058e719a36f:

```diff
+ } else if (state == AccountState::MaintenanceMode) {
+            showConnectionLabel( tr("Server %1 is currently in maintenance mode.").arg(server) );
```

I thought `Server <server> is currently in maintenance mode.` is the label to expect there.

ckamm:
@SamuAlfageme Yep, I misread your comment. I can reproduce it and am surprised about that.

ckamm:
@SamuAlfageme Fixed, got broken when we switched to Qt json parser.

PR #5762, opened 2017-05-10
Update client docs version string to 2.3.2

settermjd:


ckamm:
@settermjd The master branch is actually targetting 2.4.0. Is there a way to make this automatically dependent on VERSION.cmake? That one will always contain the correct version information.

guruz:
You can have something like doc/conf.py**.in** and use a method like here: https://github.com/owncloud/client/blob/b26db062d23a54357695522680b196d0ad89558b/CMakeLists.txt#L181

But then you need to build the docs from the build directory (vs source directory) I don't know if that is supported

settermjd:
@ckamm @guruz I'm not sure. I'll have to do some research into Sphinx-Doc to find out. Thanks for raising the point.

settermjd:
I've had a bit of a look into it, but don't know that much about CMake (yet). I'll see what I can find out.

settermjd:
@ckamm, no, there's no way in Sphinx-Doc to automatically do this. It's a very manual process.

settermjd:
I'll make a change to this PR to point it to 2.4.0, instead of 2.3.2.

Issue #5023, opened 2016-06-30
Use "Local Links" on the client


dragotin:
### Expected behaviour

Since 9.1 ownCloud server will understand the "local links" as described in owncloud/core#11732

For a good platform experience, there are a couple of places in the client where we could make use of the local links. 

Maybe we should introduce a kind of "local link" symbol that copies the local link to the clipboard, similar to the "copy full SHA" button on github.

The symbol could appear here:
- [ ] In the sync protocol in the file column (or in a column next to it, to make it easier)
- [ ] In the activity files concerned by an activity could display a link copy button
- [x] In the file manager context menu, with an entry "Copy ownCloud Link".

I see the problem that users will not understand the difference between a public link that works without that the user has an account on the ownCloud that has the file and the _local link_ discussed here, that requires a valid user.   

Related: owncloud/core#25317

hodyroff:
Yes, important for 2.4. Include into the share menu, or make it a seperate right click for easiest experience.
User story: Just sitting in the train and could use this right now ;)
If share menu we could also include the /download link thing requested: https://github.com/owncloud/client/issues/5627

SamuAlfageme:
Implement server permalink feature (https://github.com/owncloud/core/pull/24434) via context menu shortcut.

### Scenario: 

There's a file an user wants to point at to a colleague he's already sharing the parent directory with. Instead of generating a new shared link for the individual file, he can use a permalink to point to any file in the shared folder hierarchy: e.g.

`https://<server>/index.php/f/29` will always point to the same file on both accounts no matter if it's displaced within any of the file trees. 

An entry in the file explorer's context menu could be used as shortcut to directly:

* **Copy the permalink to the clipboard**: ("`Copy permalink`") 
* Open file in the containing folder context (file's row highlighted) ("`Open <server> in browser`")
    * Ideally for this, the automatic login (#633) should also be implemented for a better UX.

The permalink is available for both shared and non-shared files so should be available for both use cases.

pmaier1:
Yup, https://github.com/owncloud/client/issues/5023#issuecomment-289399666 looks good to me except that I would prefer `Copy permanent link` and would change `Open <server> in browser` to `View in browser` (why would you mention `<server>` there?)

pmaier1:
After some discussions we decided to change the name of the permalink to "private link" as it reflects that you need access, doesn't necessarily exclude other users and fits with public links. Still need to promote to all platforms but FYI.

ckamm:
[...]

PR #5763, opened 2017-05-10
Private links in share dialog and shell integrations

ckamm:
This is an initial draft for comment. 

Mostly the patch is quite straightforward. The exception is that storing the fileid in addition to the id adds some complications, see https://github.com/owncloud/client/commit/875b9bbe9c405557448cdad8a68c2f504ab6811d . In particular
1. We need to initially-populate the new column from the full id column. That shouldn't cause problems though.
2. PUT/MKCOL only give us the full id, not the fileid. If we extract the fileid by doing `.left(8)`, we might as well not add the new column and just do that whereever needed. 

The other option would be to have a transition plan where the server also replies with a new `OC-FileIdLocal: <oc:fileid>` header. Then oc:id and oc:fileid would be truly decoupled.

Opinions?

The last thing is that of course the OSX and Windows integrations also need to be updated to include the new action. Note that while I only added one action to the context menu for now, we might desire more, see discussion in https://github.com/owncloud/client/issues/5023. 

(I wonder whether we should generalize and have the socket api give a list of context menu commands and menu titles to the integrations - similar to GET_STRINGS - then we could adjust context menus without touching the integrations in the future)

### DIFF
--- a/src/gui/socketapi.cpp
+++ b/src/gui/socketapi.cpp
@@ -526,8 +526,7 @@ void SocketApi::command_EMAIL_PRIVATE_LINK(const QString &localFile, SocketListe
 void SocketApi::command_GET_STRINGS(const QString &, SocketListener *listener)
 {
     static std::array<std::pair<const char *, QString>, 5> strings { {
-        { "SHARE_MENU_TITLE", tr("Share with %1...", "parameter is ownCloud").arg(Theme::instance()->appNameGUI()) },
-        { "APPNAME", Theme::instance()->appNameGUI() },
+        { "SHARE_MENU_TITLE", tr("Share...") },
         { "CONTEXT_MENU_TITLE", Theme::instance()->appNameGUI() },
         { "COPY_PRIVATE_LINK_MENU_TITLE", tr("Copy private link to clipboard") },
         { "EMAIL_PRIVATE_LINK_MENU_TITLE", tr("Send private link by email...") },


csync/src/csync_update.c
```diff
@@ -494,6 +494,7 @@ static int _csync_detect_update(CSYNC *ctx, const char *file,
       st->etag  = c_strdup(fs->etag);
   }
   csync_vio_set_file_id(st->file_id, fs->file_id);
+  csync_vio_set_file_id(st->file_id_local, fs->file_id_local);
```
guruz: Man that was confusing.. :)

src/libsync/syncjournaldb.cpp
```diff
@@ -665,8 +666,26 @@ bool SyncJournalDb::updateMetadataTableStructure()
...
+        query.prepare("ALTER TABLE metadata ADD COLUMN fileidlocal TEXT;");
+        if( !query.exec() ) {
+            sqlFail("updateMetadataTableStructure: Add column fileidlocal", query);
+            re = false;
+        }
+
+        // Populate the new column from fileid by using its structure:
+        // fileid is %08d of the numeric id concatenated with the instanceid,
+        // guaranteed for owncloud servers <= 10.0.0
+        query.prepare("UPDATE metadata SET fileidlocal = CAST(CAST(SUBSTR(fileid,1,8) as INTEGER) as TEXT);");
```
guruz: yay for doing this in SQL :)

guruz:
Would you have a way to autotest the migration step? Maybe in tx.pl or so?
(..basically make sure that if the old id is set, the new one should also be set).

guruz:
FYI @PVince81 @DeepDiver1975 to keep you in the loop

guruz:
Code wise it looks good.
But we should probably indeed have the server also return the ```fileid``` on the operations (like MKCOL) where it now returns the ID.

I just wonder how decoupled this can be. Could we merge this PR here and just schedule something for the server?

ckamm:
@guruz Will look at tests.

The idea for decoupling id/fileid would be that we check for the presence of the new fileid header, and use left(8) of the full id if the new header isn't present. That way it works for old servers and future servers can have completely different id/fileid.

src/libsync/owncloudpropagator.cpp
```diff
@@ -323,6 +323,27 @@ bool PropagateItemJob::checkForProblemsWithShared(int httpStatusCode, const QStr
     return false;
 }
 
+void PropagateItemJob::readFileIdHeaders(const QNetworkReply &reply)
```
guruz: Any special reason to use reference for a qobject which was on heap before? (i don't think we do that anywhere else in the client)

src/libsync/propagateremotemkdir.cpp
```diff
@@ -127,6 +126,10 @@ void PropagateRemoteMkdir::propfindResult(const QVariantMap &result)
     }
     if (result.contains("id")) {
         _item->_fileId = result["id"].toByteArray();
+        _item->_fileIdLocal = result["fileid"].toByteArray();
+        if (_item->_fileIdLocal.isEmpty()) {
+            _item->_fileIdLocal = _item->_fileId.left(8);
```
guruz: Should all the .left(8) magic (it's in other commits too right?) be moved to a function that has a fat comment to remove the whole stuff in a year or two?

src/libsync/propagateuploadng.cpp
```diff
@@ -488,20 +488,12 @@ void PropagateUploadFileNG::slotMoveJobFinished()
         return;
     }
 
-    QByteArray fid = job->reply()->rawHeader("OC-FileID");
-    if(fid.isEmpty()) {
-        qWarning() << "Server did not return a OC-FileID" << _item->_file;
+    if(job->reply()->rawHeader(fileIdHeaderC).isEmpty()) {
```
guruz: So we assume that (future) servers will continue to always emit both "local" and full IDs?
@PVince81

src/libsync/owncloudpropagator.cpp
ckamm: Ah, I did this automatically because it can't be 0. But yes, it's not the style we use and will adjust it.

src/libsync/propagateuploadng.cpp
ckamm: The client at least needs both for now.

src/libsync/propagateremotemkdir.cpp
ckamm: I had it in a function, but then removed it when there was only a single use left. Then I later found this corner case (PropFind is only sent for servers <= 7.0.0). Yep, will put it into a function again.

ckamm:
Rebased now that the logging PR was merged.

src/gui/owncloudgui.h
```diff
@@ -86,7 +86,16 @@ public slots:
     void slotOpenPath(const QString& path);
     void slotAccountStateChanged();
     void slotTrayMessageIfServerUnsupported(Account *account);
-    void slotShowShareDialog(const QString &sharePath, const QString &localPath, bool resharingAllowed);
+
+    /**
+     * Open a share dialog for a file or folder.
+     *
+     * sharePath is the full remote path to the item,
+     * localPath is the absolute local local path to it (so not relative
```
phil-davis: How local is the local path? "local local" ;)

ckamm:
@guruz This looks good to me now

ckamm:
I made some minor adjustments to allow the new plugins work better with older clients.

src/gui/guiutility.cpp
```diff
@@ -0,0 +1,35 @@
...
+#include "guiutility.h"
+
+#include <QClipboard>
+#include <QApplication>
+
+using namespace OCC;
+
+#ifdef Q_OS_MAC
+// Defined in clipboard.mm
+extern void copyToPasteboard(const QString &string);
```
ogoffart: (could be moved within the #ifdef block bellow,  at function scope)

src/gui/guiutility.h
```diff
@@ -0,0 +1,35 @@
...
+
+namespace OCC
+{
+namespace Utility
+{
+
+/** Copies a string to the clipboard.
+ *
+ * Includes a workaround for
+ * https://github.com/owncloud/client/issues/3300
```
ogoffart: The bug https://bugreports.qt.io/browse/QTBUG-48953 has been  fixed in Qt 5.6.1,  so do we still need it as we only use Qt 5.6 on mac anyway?

In that case, this function would not be needed, as this is just a wrapper for QApplication::clipboard()->setText(...);

src/libsync/account.cpp
```diff
@@ -160,6 +160,11 @@ QUrl Account::davUrl() const
     return Utility::concatUrlPath(url(), davPath());
 }
 
+QUrl Account::filePermalinkUrl(const QByteArray& fileIdLocal) const
+{
+    return Utility::concatUrlPath(url(), QLatin1String("/index.php/f/") + QString::fromLatin1(fileIdLocal));
```
ogoffart: this assume that fileIdLocal does not contains '/'.

src/gui/socketapi.cpp
```diff
@@ -524,6 +524,19 @@ void SocketApi::command_COPY_LOCAL_LINK(const QString &localFile, SocketListener
     }
 }
 
+void SocketApi::command_GET_STRINGS(const QString &, SocketListener *listener)
```
ogoffart: (ideally, this commit should be rebased before the changes in the plugins.)

src/gui/guiutility.h
ckamm: Yes, looks reasonable. I'll drop our workaround `copyToClipboard` (@guruz)

ckamm:
@ogoffart Thanks for the review. I have rebased and addressed the comments. Our discussion about not storing id and fileid in the db is not yet reflected in code.

ckamm:
@ogoffart @guruz I've now checked what happens with remote shares: They seem to be assigned instance-local ids. So their fileid column entry in our database does also use the local instance's identification string.

That means using the numeric-only file id instead of the full id would work just fine for rename/move detection even in the presence of federated shares.

My current opinion is: Don't change the database at all. Query the 'local' file id on demand. That would cover what we need without potentially introducing issues down the line. (why do two ids even exist if the instance identification string is always constant and thus redundant?)

PVince81:
just use the fileid and ignore federated shares

ckamm:
@PVince81 You're saying it's safe to use `oc:fileid` (only numeric) over `oc:id`? If so, we might - in a migration step to 2.4.0 - remove the `id` column from our database and only keep the `fileid`.

PVince81:
both are safe to use, it's the same value except that one has the instance id appended.

The part I dislike is that one of them is padded with zeroes which will overflow one day...

guruz:
btw migration steps mean no downgrade possible..

guruz:
```<oc:id>00000003ocr2n5bhxjuy</oc:id>```
Does it make sense to go up to first letter instead of first-8 chars?

**Comment by guruz** [comment]
Acording to @hodyroff in https://github.com/owncloud/enterprise/issues/2015#issuecomment-303302287 this is now called "private link". Maybe you need to adjust something in the UI? (code can stay?)

**Comment by davivel** [comment]
Thinking about all this for mobile implementation.

What's the advantage of storing both fields separately? Knowing that ```fileId``` can be derived from the "full id" in a consistent way, I don't see a reason to retrieve, store and maintain two values instead of one.

Also, parting from a situation where "full id" is already stored, ¿why changing it for ```fileId```? Even if all the information of the first one is not strictly needed in the client, is the benefit of getting rid of it really bigger than the cost of changing the current state? Is it worth the risk inherent to 
 migrate users database?

**Comment by ckamm** [comment]
@davivel I'd be very happy with just deriving the fileid from the id. The issue is that it's unsafe as long as the relationship between fileid and id is just an implementation detail. What if it changes?

@PVince81 Would you be up for codifying the fileid/id relationship in the server? Put it in the property spec, code comments, add a test etc? If so, the clients could indeed get by with less effort.

Also, if the private link urls would work with the full id, the clients wouldn't even need to care about getting the fileid. But since we want private links to work with server version 10 we can't completely ignore them.

**Comment by davivel** [comment]
>  The issue is that it's unsafe as long as the relationship between fileid and id is just an implementation detail.

Agree.

> Would you be up for codifying the fileid/id relationship in the server? Put it in the property spec, code comments, add a test etc?

Great point.

**Comment by ckamm** [comment]
@guruz @davivel To move this forward I've created this PR against core: https://github.com/owncloud/core/pull/28193 - ideally we can drop the discovery and storage of the short numeric ids.

**Comment by ckamm** [comment]
I've updated this PR to remove the fileid discovery and storage and instead generate the fileid from the id we store already!

File: src/gui/socketapi.cpp [hunk 11]
```diff
@@ -514,6 +504,41 @@ void SocketApi::command_SHARE_MENU_TITLE(const QString &, SocketListener *listen
...
+            url.toString(QUrl::FullyEncoded),
+            0);
+    }
+}
+
+void SocketApi::command_GET_STRINGS(const QString &, SocketListener *listener)
+{
+    static std::array<std::pair<const char *, QString>, 5> strings { {
+        { "SHARE_MENU_TITLE", tr("Share with %1...", "parameter is ownCloud").arg(Theme::instance()->appNameGUI()) },
+        { "APPNAME", Theme::instance()->appNameGUI() },
```
jturcotte: Is this string going to be used?

src/gui/socketapi.cpp
```diff
@@ -514,6 +508,39 @@ void SocketApi::command_SHARE_MENU_TITLE(const QString &, SocketListener *listen
...
+            tr("I shared something with you"),
+            url.toString(),
+            0);
+    }
+}
+
+void SocketApi::command_GET_STRINGS(const QString &, SocketListener *listener)
+{
+    static auto map = std::initializer_list<std::pair<const char *, QString>>{
+        { "SHARE_MENU_TITLE", tr("Share with %1...", "parameter is ownCloud").arg(Theme::instance()->appNameGUI()) },
```
jturcotte: Putting the appName in both the submenu title and the share entry looks a bit redundant now:
screen shot 2017-07-04 at 13 37 16

Maybe we could name it just "Share..."?

guruz:
@SamuAlfageme The contents should now also be in ```local_links``` branch in normal client repo..

src/gui/socketapi.cpp
ckamm: Yes

src/gui/socketapi.cpp
ckamm: Not right now, it seemed like a string we might need - can drop for now.

src/gui/socketapi.cpp
```diff
@@ -514,6 +508,39 @@ void SocketApi::command_SHARE_MENU_TITLE(const QString &, SocketListener *listen
...
+        { "APPNAME", Theme::instance()->appNameGUI() },
+        { "CONTEXT_MENU_TITLE", Theme::instance()->appNameGUI() },
+        { "COPY_PRIVATE_LINK_TITLE", tr("Copy private link to clipboard") },
+        { "EMAIL_PRIVATE_LINK_TITLE", tr("Send private link by email...") },
```
ckamm: Agreed!

ckamm:
@jturcotte I think I have addressed the remaining issues and will test some more (dolphin only tomorrow though). 

@SamuAlfageme I also updated the local_links branch in the main client repo.

ckamm:
Merged manually, squashing various review and fixup commits together.

PR #5772, opened 2017-05-15
Selective sync: Skip excluded folders when reading db

ckamm:
When a new folder becomes selective-sync excluded, we already mark it
and all its parent folders with _invalid_ etags to force rediscovery.

That's not enough however. Later calls to csync_statedb_get_below_path
could still pull data about the excluded files into the remote tree.

That lead to incorrect behavior, such as uploads happening for folders
that had been explicitly excluded from sync.

To fix the problem, statedb_get_below_path is adjusted to not read the
data about excluded folders from the database.

Currently we can't wipe this data from the database outright because we
need it to determine whether the files in the excluded folder can be
wiped away or not.

See owncloud/enterprise#1965

### DIFF
--- a/test/testsyncengine.cpp
+++ b/test/testsyncengine.cpp
@@ -213,6 +213,49 @@ private slots:
         }
     }
 
+    void testSelectiveSyncBug() {
+        // issue owncloud/enterprise#1965: files from selective-sync ignored
+        // folders are uploaded anyway is some circumstances.
+        FakeFolder fakeFolder{FileInfo{ QString(), {
+            FileInfo { QStringLiteral("parentFolder"), {
+                FileInfo{ QStringLiteral("subFolder"), {
+                    { QStringLiteral("fileA.txt"), 400 },
+                    { QStringLiteral("fileB.txt"), 400, 'o' }
+                }}
+            }}
+        }}};
+
+        QCOMPARE(fakeFolder.currentLocalState(), fakeFolder.currentRemoteState());
+        auto expectedServerState = fakeFolder.currentRemoteState();
+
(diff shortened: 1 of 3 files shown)


ckamm:
@ogoffart Yes, testing and documentation are still missing.

guruz:
@ckamm Could you put this in the 2.3 branch instead?
Then @michaelstingl can make a 2.3.3git build for the customer to test.
(I think it's ok to merge without test for now?)

ckamm:
Tests and some doc added, will retarget for 2.3

michaelstingl:
Will it be in tomorrow's nightlies?

ckamm:
@michaelstingl I've merged it now, should be in the nightlies. @guruz ?

test/testsyncengine.cpp
```diff
@@ -213,6 +213,49 @@ private slots:
...
+        for (int i = 0; i < 3; ++i) {
+            fakeFolder.syncOnce();
+
+            {
+                // Nothing changed on the server
+                QCOMPARE(fakeFolder.currentRemoteState(), expectedServerState);
+                // The local state should still have subFolderA
+                auto local = fakeFolder.currentLocalState();
+                QVERIFY(local.find("parentFolder/subFolder"));
+                QVERIFY(local.find("parentFolder/subFolder/fileA.txt"));
```
ckamm: @ogoffart I expected that was intentional. Currently it either deletes the whole folder -- or doesn't do anything about it. Was the originally intended behavior to remove all files that should be safe to remove?

ogoffart: --->  https://github.com/owncloud/client/issues/5783

PR #5781, opened 2017-05-19
Sharing layout

ckamm:
@SamuAlfageme: Could you test whether the layout problems you observed on Win and OSX are gone?

### DIFF
--- a/src/gui/sharedialog.cpp
+++ b/src/gui/sharedialog.cpp
@@ -107,6 +107,7 @@ ShareDialog::ShareDialog(QPointer<AccountState> accountState,
         return;
     }
 
+    _ui->label_icon->hide();
     if (QFileInfo(_localPath).isFile()) {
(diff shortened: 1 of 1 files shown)


ckamm:
@SamuAlfageme As far as I'm aware I didn't remove any icons that existed before. The "Folder" label is missing because ownCloud Manual.pdf is in the root folder. I don't know where that odd square comes from, unfortunately.

SamuAlfageme:
@ckamm 

> As far as I'm aware I didn't remove any icons that existed before. 

I think https://github.com/owncloud/client/pull/5781/commits/21fa9be560eac12542469fffed870ef2982a4999 is hiding the OS's default icon displayed on the sharing dialog:

<p align="center">
  
</p>

... unless the thumbnail from the server is fetched. 

About the label, you're right, I did just check on the root but It's still present on subfolders. And I can open a separate issue about the square-container-thingie on [OS X]

ckamm:
@SamuAlfageme Hm, I had never seen this OS default icon. Will investigate.

Issue #5728, opened 2017-04-26
Client asks for account password after certificate is rejected


ckamm:
When a SSL server certificate is presented to the user and rejected the client will not be able to communicate with that server. However, the client will still pop up a dialog asking for the user's password *after* the certificate was rejected. That is pointless and confusing.

Steps:
1. Have a server with a self-signed certificate (or mitmproxy one)
2. Set up an account, log out of it, quit the client.
3. Remove the certificate authorization from the config file. (or restart mitmproxy)
4. Start the client.
5. Reject the server's certificate.
6. Notice that the "Please enter your password" dialog pops up for the account.

@guruz

ckamm:
This happens because the SSL Error that happens during the GET status.php causes a CredentialsMissingOrWrong. That's the case because we might need to retrieve the SSL *client* certificates from the keychain. So the httpcreds are fetched, but not ready (no password), and we go to askFromUser().

guruz:
As discussed on IRC, should listen to ```authenticationRequired``` instead.
We discussed some more, need to summarize this still

ckamm:
@guruz The code for sending the Auth header all the time refers to issue #4326 - doesn't that mean we can't re-revert https://github.com/owncloud/client/commit/ef9483c82d111464e67291cd134beef99d0afcfc as long as we support servers 8.X?

ckamm:
PR https://github.com/owncloud/client/pull/5782

guruz:
-> https://github.com/owncloud/client/issues/5789

SamuAlfageme:
@ckamm @guruz somehow related: wouldn't it be useful that if the user mistakenly rejects the certificate and clicks on `Account > log-in`, the certificate will be displayed for acceptance again. Client restart is needed now for that

ckamm:
@SamuAlfageme That'd be possible and make some sense. We'd need to track which rejected cert belongs to which account, so we can clear the "previously rejected" list of the right certificates when a relogin is triggered. It's a different ticket though, I'll make one.

SamuAlfageme:
@ckamm closing here then 🎉

PR #5782, opened 2017-05-19
Credentials: Simplify credential flow #5728

ckamm:
And as a side effect: don't ask for user password when we can't
connect to the server in the first place.

I've gone through a lot of test cases, but still review carefully please.

### DIFF
--- a/src/libsync/connectionvalidator.cpp
+++ b/src/libsync/connectionvalidator.cpp
@@ -48,15 +48,17 @@ QString ConnectionValidator::statusString(Status stat)
     case Connected:
         return QLatin1String("Connected");
     case NotConfigured:
-        return QLatin1String("NotConfigured");
+        return QLatin1String("Not configured");
     case ServerVersionMismatch:
         return QLatin1String("Server Version Mismatch");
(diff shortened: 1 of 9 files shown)


src/libsync/connectionvalidator.cpp
```diff
@@ -143,22 +145,21 @@ void ConnectionValidator::slotStatusFound(const QUrl &url, const QJsonObject &in
...
 void ConnectionValidator::slotNoStatusFound(QNetworkReply *reply)
 {
     auto job = qobject_cast<CheckServerJob *>(sender());
     qCWarning(lcConnectionValidator) << reply->error() << job->errorString() << reply->peek(1024);
-    if (!_account->credentials()->ready()) {
-        // This could be needed for SSL client certificates
-        // We need to load them from keychain and try
-        reportResult(CredentialsMissingOrWrong);
-    } else if (!_account->credentials()->stillValid(reply)) {
+    if (reply->error() == QNetworkReply::SslHandshakeFailedError) {
```
ogoffart: I like the fact that this is more accurate than reporting wrong credentials for every error.

Issue #5752, opened 2017-05-08
When removing an account, the keychain entry for the password should be erased


SamuAlfageme:
Similar discussion to https://github.com/owncloud/client/issues/5342. 

### Actual behavior
* When you logout of an account in the client, the keychain entry for the password is removed. (Certificate and key are preserved though.)
* When the account is deleted, all the information on the keychain is preserved. 

### Expected behavior
Semantically, the keychain entry for the deleted account should be removed on the account's deletion (deletion should be seen as log-out)

### Steps to reproduce
1. Remove an account from the client
2. Check Keychain Access app (OS X) for `ownCloud` entries relative to the account that was removed

Seen it on OS X, but I believe it affects the rest of the OSS -> will reproduce there.

ckamm:
PR: https://github.com/owncloud/client/pull/5787

SamuAlfageme:
- [x]  OS X 👍 
- [ ] Windows ⚠️
- [ ] Linux ⚠️

SamuAlfageme:
👍

SamuAlfageme:
Note to self: verify if OAuth access tokens are being removed from the keychain when deleting the account. I'm getting way too many stale entries lately.

SamuAlfageme:
Some findings info after checking out the keychain integration:

- Non-SSL (plain http://) connections also create a `<username>_clientCertificatePEM` & `<username>_clientKeyPEM` empty entries in the keychain
- These are not erased together with the account (like the password is)

That's why I was getting the so many entries left. @ckamm would it be worth/easy to erase those as well?

ckamm:
@SamuAlfageme I'll check it out. Probably simple enough to get into 2.4

ckamm:
@SamuAlfageme PR is ready, see #6118 -- wiping the client cert keychain entries when the account is deleted won't make it into 2.4 though. But at least we'll not be creating these entries when there's not client cert anymore.

PR #5787, opened 2017-05-22
Account: forget credentials on delete #5752

ckamm:
@SamuAlfageme This should do it, I see the keychain-removal code being run on delete now. But I've had some trouble seeing the keychain entries. Testing would be appreciated.


ogoffart:
:+1:

PR #5793, opened 2017-05-26
Fix Debian 7 builds

dschmidt:


guruz:
👍 with me.
Is this only broken on master or also 2.2.x?

guruz:
(I guess so because of https://github.com/owncloud/client/pull/5759/files ?)

dschmidt:
Exactly - so it only affects 2.3.2

guruz:
@dschmidt Sorry I meant 2.3.x.. then your PR should go to 2.3 branch

dschmidt:
That didnt quiten work because of the merges. Will reopen later

PR #5795, opened 2017-05-29
Application: debug implies logdebug

guruz:


guruz:
Too late, i had already told @settermjd to document it ;-)
It's fine i think.

PR #5799, opened 2017-05-29
Download: Retrigger folder discovery on 404

ckamm:
See owncloud/enterprise#1966

If the server and the client's database go out of sync, there could be
persistent 404 errors. This change ensures that the problem corrects
itself eventually by triggering a remote discovery of the file's
parent folders.

It does not address the root cause that might have lead to the
divergence.


ckamm:
I don't think this is ideal: The issue shouldn't happen and hiding symptoms might make genuine bugs harder to track down, or get reported less.

SamuAlfageme:
@ckamm linking https://github.com/owncloud/core/pull/28598 to have a reference here.

PR #5800, opened 2017-05-29
Selective sync: Add folder removal warning to wizard ui

ckamm:
Reported in a comment of #2297

### DIFF
--- a/src/gui/selectivesyncdialog.cpp
+++ b/src/gui/selectivesyncdialog.cpp
@@ -99,6 +99,11 @@ SelectiveSyncWidget::SelectiveSyncWidget(AccountPtr account, QWidget *parent)
     _folderTree->header()->setStretchLastSection(true);
     _folderTree->headerItem()->setText(0, tr("Name"));
     _folderTree->headerItem()->setText(1, tr("Size"));
+
+    auto footer = new QLabel(this);
+    footer->setText(tr("Existing deselected folders will be <b>removed locally</b> if the server's copy is up to date."));
+    footer->setWordWrap(true);
+    layout->addWidget(footer);
 }
 
 QSize SelectiveSyncWidget::sizeHint() const


ogoffart:
Normally, in the wizard, there should not be existing sync.  Maybe we can detect if there is already a sync dtabase?

ckamm:
@ogoffart Am I wrong in thinking you could:
* Already have the server and local data synced up manually
* Create a new account, point it to the folder
* Use selective sync to deselect some subfolders
* Be surprised that some of your local data is deleted?

ogoffart:
@ckamm  in that scenario, the files would not be deleted, they would be ignored.  We would only delete if there was already a database and the files where in sync.

phil-davis:
Trying this sequence:
1) Create users t1 and t2, put a folder Crud in each user on the server with a file in it.
2) Add account to the client for t1, let it sync everything to some folder t1t2 (which now gets a subfolder "Crud")
3) Add account to the client for t2, select what to sync during the wizard, deselect "Crud".

Folder t1t2/Crud now shows with the "!" warning overlay icon. It is being synced to t1 on the server but is not being kept in sync with t2 on the server.

That seems reasonable - and the files on the client do not get deleted in this case.

phil-davis:
A different sequence:
1) Create users t1 and t2, put a folder Crud in each user on the server with a file in it.
2) Add account to the client for t1, let it sync everything to some folder t1t2 (which now gets a subfolder "Crud")
3) Add account to the client for t2, also let it sync everything to some folder t1t2. t1t2 is updated on the client with files from the t2 account on the server. These are then sync back to t1 on the server - good.
4) On the client, for account/sync connection for t1, deselect "Crud" from sync.

a) Crud remains on the server t1 account.
b) Crud is deleted on the client.
c) The removal of Crud on the client is propagated up to the t2 account, where Crud is deleted.

Should (b) and (c) happen? Or should Crud on the client just be left "disconnected" from Crud on t1 on the server?

The issue here is about the wizard, so actually this case is not directly relevant here. But I wonder if something "wrong" happens here? Or what the actual result of such a sequence of user actions should be? Is there a misbehavior to report? Or is this the expected outcome?

ckamm:
@phil-davis Nice catch about selective sync and syncing the same folder to multiple accounts! Could you move the second case to a separate ticket?

@ogoffart You are right, the warning is unnecessary.

PR #5821, opened 2017-06-06
FolderStatusModel: fix assert in Qt when the list of subfolder is empty

ogoffart:
Fix an assert that happens in beginInsertRows when opening a folder
and that folder is empty.
This can only be reproduced with a debug build of Qt.

### DIFF
--- a/src/gui/folderstatusmodel.cpp
+++ b/src/gui/folderstatusmodel.cpp
@@ -614,6 +614,8 @@ void FolderStatusModel::slotUpdateDirectories(const QStringList &list)
     if (!parentInfo) {
         return;
     }
+    ASSERT(parentInfo->_fetching); // we should only get a result if we were doing a fetch
+    ASSERT(parentInfo->_subs.isEmpty());
 
     if (parentInfo->hasLabel()) {
         beginRemoveRows(idx, 0, 0);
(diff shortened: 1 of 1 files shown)


src/gui/folderstatusmodel.cpp
```diff
@@ -712,9 +712,11 @@ void FolderStatusModel::slotUpdateDirectories(const QStringList &list)
...
     }
 
-    beginInsertRows(idx, 0, newSubs.size() - 1);
-    parentInfo->_subs = std::move(newSubs);
-    endInsertRows();
+    if (!newSubs.isEmpty()) {
+        beginInsertRows(idx, 0, newSubs.size() - 1);
+        parentInfo->_subs = std::move(newSubs);
+        endInsertRows();
+    }
```
ckamm: might still need to clear up parentInfo->_subs in the else case?

ogoffart: parentInfo->_subs should be empty.  I added an assert now to verify it.
Otherwise, the insertRows would also be wrong as it would be replacing rows.

ogoffart:
This was just reported ( https://github.com/owncloud/client/issues/5823 )  
Perhaps this should actually be cherry-picked to 2.3.3

guruz:
> Perhaps this should actually be cherry-picked to 2.3.3

Not needed if it can't be reproduced with our packages or when using a non-debug Qt build.

Issue #5783, opened 2017-05-19
Selective sync may not locally remove unneeded files


ckamm:
When a folder is removed from the sync by deselecting it in the selective sync ui, typically it is deleted locally. However, when there are new, conflicting or untracked files inside the folder, nothing at all is done with it. Discussions with @ogoffart suggest that removing all subitems that are safe to remove may be more desirable.

Steps to reproduce:
* Inside a synced folder, create a folder `foo/` containing `foo/uptodate`, `foo/subfolder/uptodate`.
* Fully sync it, then pause the sync
* Create a new file `foo/newfile`
* Open selective sync and deselect `foo/`
* Unpause the sync

Expected: `foo/uptodate`, `foo/subfolder/uptodate` and `foo/subfolder` should be removed
Actual: Nothing is removed

This behavior is the result of `checkForPermission()`: All subitems inside a selective-sync ignored folder get set to `IGNORED`. But even if they weren't: as long as the parent is `IGNORED` subjobs are not executed in `PropagateDirectory`.

ogoffart:
Test case in TestSyncEngine::testSelectiveSyncBug  :  parentFolder/subFolder/fileA.txt  should be deleted but it is not.

guruz:
What would be the milestonde for this @ogoffart @ckamm ?

ckamm:
I've tested it using the test case from the initial comment

PR #5822, opened 2017-06-06
SyncEngine: SelectiveSync: Remove local files of undelected folder de…

ogoffart:
…spite other modified files

Issue #5783

When the directry that should be removed by selective sync contains changes,
we ignore the whole sub tree instead of only ignoreing new files.
We cannot ignore the whole directory, we need to ignore only the directory
that do not have files to remove

### DIFF
--- a/src/libsync/syncengine.cpp
+++ b/src/libsync/syncengine.cpp
@@ -1159,6 +1159,7 @@ void SyncEngine::checkForPermission(SyncFileItemVector &syncItems)
     bool selectiveListOk;
     auto selectiveSyncBlackList = _journal->getSelectiveSyncList(SyncJournalDb::SelectiveSyncBlackList, &selectiveListOk);
     std::sort(selectiveSyncBlackList.begin(), selectiveSyncBlackList.end());
+    SyncFileItemPtr needle;
 
     for (SyncFileItemVector::iterator it = syncItems.begin(); it != syncItems.end(); ++it) {
         if ((*it)->_direction != SyncFileItem::Up) {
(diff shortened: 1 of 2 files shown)


src/libsync/syncengine.cpp
```diff
@@ -1179,6 +1180,34 @@ void SyncEngine::checkForPermission(SyncFileItemVector &syncItems)
             if ((*it)->_isDirectory) {
                 for (SyncFileItemVector::iterator it_next = it + 1; it_next != syncItems.end() && (*it_next)->_file.startsWith(path); ++it_next) {
                     it = it_next;
+                    if ((*it)->_direction != SyncFileItem::Up && (*it)->_instruction == CSYNC_INSTRUCTION_REMOVE) {
```
ckamm: Why `!= Up` and not `== Down`? I'd be more comfortable with this pointing specifically at local removes.

src/libsync/syncengine.cpp
```diff
@@ -1179,6 +1180,34 @@ void SyncEngine::checkForPermission(SyncFileItemVector &syncItems)
...
+                                break;
+                            }
+                            // Find the parent directory in the syncItems vector. Since the vector
+                            // is sorted we can use a lower_bound, but we need a fake
+                            // SyncFileItemPtr needle to compare against
+                            if (!needle)
+                                needle = SyncFileItemPtr::create();
+                            needle->_file = parentDir;
+                            auto parent_it = std::lower_bound(syncItems.begin(), it, needle);
+                            if (parent_it == syncItems.end() || (*parent_it)->destination() != parentDir) {
```
ckamm: Maybe at a `|| !(*parent_it)->_file.startsWith(path)` to ensure we're not traversing too far upwards?

src/libsync/syncengine.cpp
```diff
@@ -1179,6 +1180,34 @@ void SyncEngine::checkForPermission(SyncFileItemVector &syncItems)
...
+                                break;
+                            }
+                            ASSERT((*parent_it)->_isDirectory);
+                            if ((*parent_it)->_instruction != CSYNC_INSTRUCTION_IGNORE) {
+                                break; // already changed
+                            }
+                            (*parent_it)->_instruction = CSYNC_INSTRUCTION_UPDATE_METADATA;
+                            (*parent_it)->_status = SyncFileItem::NoStatus;
```
ckamm: Is it okay to keep `_errorString` set? I think some code considers finished items with a non-empty errorString an error.

src/libsync/syncengine.cpp
ckamm: Suggestion: Add a comment before the `if` along these lines:
```
We want to ignore almost all instructions for items inside selective-sync excluded folders. 
The exception are DOWN/REMOVE actions that remove local files and folders that are 
guaranteed to be up-to-date with their server copies.
```

src/libsync/syncengine.cpp
ogoffart: Good point.

I'm going to do the check slightly differently.  (because path ends with '/' but we also need to do it for path itself)

src/libsync/syncengine.cpp
ogoffart: I was just using the same pattern as above. Normally, conflicts should not happen there. But still I guess it's more clear with == Down, so i'll change it.

src/libsync/syncengine.cpp
ogoffart: Fixing

Issue #5813, opened 2017-06-02
[OAuth2] When opening the browser for authentication, the account wizard displays no content and allows forward navigation 


SamuAlfageme:
Relates to: https://github.com/owncloud/client/issues/5811

### Steps to reproduce
1. Use the URL of a server with [OAuth app](https://github.com/owncloud/oauth2) active in the account wizard. 
2. Clicking on the tray icon brings back the wizard with no content but the navigation buttons

<p align="center">
  <img src="https://cloud.githubusercontent.com/assets/2644445/26722198/3887a350-478f-11e7-9f5d-169d5eb20cee.png"/>
</p>

* If `next >` is selected, wizard will take you to next page where it fails to connect **but create the local sync folder**.
* If `Skip folders configuration` is selected next without authenticating in the browser, a _userless_ account will be created (i.e. `@<server.tld>` on the account tab, etc.)

<p align="center">
  <img src="https://cloud.githubusercontent.com/assets/2644445/26723316/43a44888-4794-11e7-9564-d21ed3692a77.png"/>
</p>

### Expected behavior
- [x] Display a spinner or some other component to inform the user that an action is required (maybe also a button to relaunch the browser in case it was closed by mistake)
- [ ] Still offer the "Application Password" option as a fallback from OAuth.

<p align="center">
  <img src="https://cloud.githubusercontent.com/assets/2644445/26723048/ed2c6eaa-4792-11e7-84f6-a15dbb458d32.png"/>
</p>

- [x] Change the behavior of the buttons:
    * `Cancel`: will close the wizard and discard the account creation (e.g. if the user forgot the password and wants to stop the action for good) 
    * `Back` will allow to go back to wizard's first page to change the URL in case of mistake.

cc/ @ogoffart

guruz:
I'm thinking this should be higher priority, maybe even mandatory for 2.4.0
What do you think @pmaier1 @ogoffart @tomneedham @lukebier @joneug  ?

joneug:
I agree!

Emil-G:
"Still offer the "Application Password" option as a fallback from OAuth." - Do you know if this will be server configurable to force OAuth and block Application Password?

SamuAlfageme:
Status with @ogoffart's tweaks: https://github.com/owncloud/client/pull/5825:

| Wizard waiting for browser | OAuth flow error |
|:-----------------------:|:----------------------------------------:|
[...]

Issue #5811, opened 2017-06-01
[OAuth2] Error recovery and timeouts, especially in wizard


guruz:
Follow up issue from https://github.com/owncloud/client/pull/5668#issuecomment-302044607

ogoffart:
This is actually a duplicate of https://github.com/owncloud/client/issues/5813

PR #5825, opened 2017-06-07
OAuth: Wizard page

ogoffart:
Show smething sensible if the user click on the owncloud icon, and handle errors

Issues: #5813 and #5811

### DIFF
--- a/src/gui/owncloudsetupwizard.cpp
+++ b/src/gui/owncloudsetupwizard.cpp
@@ -346,7 +346,7 @@ void OwncloudSetupWizard::slotAuthError()
     }
 
     _ocWizard->show();
-    if (_ocWizard->currentId() == WizardCommon::Page_ShibbolethCreds) {
+    if (_ocWizard->currentId() == WizardCommon::Page_ShibbolethCreds || _ocWizard->currentId() == WizardCommon::Page_OAuthCreds) {
         _ocWizard->back();
     }
     _ocWizard->displayError(errorMsg, _ocWizard->currentId() == WizardCommon::Page_ServerSetup && checkDowngradeAdvised(reply));
--- a/src/gui/wizard/owncloudoauthcredspage.h
+++ b/src/gui/wizard/owncloudoauthcredspage.h
@@ -24,6 +24,9 @@
 #include "accountfwd.h"
 #include "creds/oauth.h"
 
+#include "ui_owncloudoauthcredspage.h"
+
+
 namespace OCC {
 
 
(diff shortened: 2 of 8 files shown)


src/gui/owncloudsetupwizard.cpp
```diff
@@ -346,7 +346,7 @@ void OwncloudSetupWizard::slotAuthError()
     }
 
     _ocWizard->show();
-    if (_ocWizard->currentId() == WizardCommon::Page_ShibbolethCreds) {
+    if (_ocWizard->currentId() == WizardCommon::Page_ShibbolethCreds || _ocWizard->currentId() == WizardCommon::Page_HttpCreds) {
```
ckamm: Why? Isn't it okay to stay on the HttpCreds page if auth fails?

Also, isn't this unrelated to OAuth?

src/gui/wizard/owncloudoauthcredspage.h
```diff
@@ -36,25 +39,24 @@ class OwncloudOAuthCredsPage : public AbstractCredentialsWizardPage
     AbstractCredentials *getCredentials() const Q_DECL_OVERRIDE;
 
     void initializePage() Q_DECL_OVERRIDE;
+    void cleanupPage() override;
```
ckamm: We no longer use Q_DECL_OVERRIDE?

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -201,6 +201,11 @@ void OwncloudWizard::slotCurrentPageChanged(int id)
     }
 
     setOption(QWizard::HaveCustomButton1, id == WizardCommon::Page_AdvancedSetup);
+    if (id == WizardCommon::Page_AdvancedSetup && _credentialsPage == _browserCredsPage) {
+        // For OAuth, disable the back button in the Page_AdvancedSetup because we don't want
+        // to re-open the browser.
+        button(QWizard::BackButton)->setEnabled(false);
```
ckamm: Is there the need to re-enable the back button on other pages? In case we add further pages, or OAuth isn't supported and we fall back to http creds?

src/gui/owncloudsetupwizard.cpp
ogoffart: Oops, it was supposed to be the OAuth page of course.  Well spotted.

(This normally should not never happen because if we have a oauth connection, it should normaly work always. )

src/gui/wizard/owncloudoauthcredspage.h
ogoffart: AFAIK we dropped centos 6 whi had that old compiler. And even then we had upgraded to a more recent one.

src/gui/wizard/owncloudwizard.cpp
ogoffart: 1) The QWizard already reset the enabled property on the buttons each time the page change.
2) Since the next page is the final one and its not possible to go back, either, we are fine.

ogoffart:
@ckamm  In order to try it out, you simply need to checkout the oauth2 apps in your apps directory 
https://github.com/owncloud/oauth2/   and enable it from the apps setting page.

Normally, since https://github.com/owncloud/oauth2/pull/38 ,  the client app should automatically be registered, but this did not happen to me or @SamuAlfageme .  So I had to manually add the client in the server DB:  

```
INSERT INTO oc_oauth2_clients (identifier, secret, redirect_uri, name) VALUES('xdXOt13JKxym1B1QcEncf2XDkLAexMBFwiT9j6EfhhHFJhs2KM9jbjTmf8JBXE69', 'UBntmLjC2yYCeHwsyj73Uwo9TAaecAetRwMw0xYcvNL9yRdLSUi0hUAHfvCHFeFh', 'http://localhost:*', 'Desktop Client');
```

Alternatively, you can also register a new client from the admin interface, and change the values in theme.cpp

ckamm:
@ogoffart From testing:
* I also had to manually insert the "Desktop Client" oauth client. The table was empty initially.
* "Login Successfull" -> one too many "l"
* I had one http and one oauth account set up for the same user+server, that lead to problems. (same keychain entry, probably) We might want to use some unique identifier to disambiguate.

Issue #5809, opened 2017-05-31
Disable the external storages/big folders notice on account log-out


SamuAlfageme:
### Steps to reproduce
1. Add a new external storage/folder exceeding the size limit set on the client
2. Get notified in the account tab
3. Don't select any option and log-out

<p align="center">
  <img src="https://cloud.githubusercontent.com/assets/2644445/26641105/cef20a86-4629-11e7-8eb5-a131ef73d4bc.png"/>
</p>

### Expected behavior
The options should be gone after log-out, the same way the selective sync list is. 

<p align="center">
  <img src="https://cloud.githubusercontent.com/assets/2644445/26641230/394c86f4-462a-11e7-8059-72c24bd00795.gif"/>
</p>

ckamm:
PR https://github.com/owncloud/client/pull/5826

SamuAlfageme:
Still reproduces in `Version 2.5.0alpha1 (build 9885)` - my concerns from the comments on https://github.com/owncloud/client/pull/5826 are still valid - e.g.

1. Wait till the buttons appear
2. Log out / disable the network on the OS 
3. Click on `Synchronize all`
4. Log in / re-enable the connection on the host
5. **Client lost the selection** 

I'm still voting for the "_if we don't display the selective sync list at all, do not display these buttons until connection is established back either_"

ckamm:
@SamuAlfageme Checking it out. Possibly my change only covered the case of edits to the selective sync list, not the "big folder" edit state.

SamuAlfageme:
Now, this is great stuff:

![hide_external](https://user-images.githubusercontent.com/2644445/41593688-302089d4-73c1-11e8-8a17-6c8b7d00cf20.gif)

Closing here! 🤘

ckamm:
Finally! :)

PR #5826, opened 2017-06-08
Hide selective sync buttons while disconnected #5809

ckamm:
The selective sync selections get reset in this case, while the
big folder warnings will pop up again on reconnection.

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -688,14 +688,22 @@ void AccountSettings::slotAccountStateChanged()
     /* Allow to expand the item if the account is connected. */
     ui->_folderList->setItemsExpandable(state == AccountState::Connected);
 
-    /* check if there are expanded root items, if so, close them, if the state is different from being Connected. */
     if (state != AccountState::Connected) {
+        /* check if there are expanded root items, if so, close them */
         int i;
         for (i = 0; i < _model->rowCount(); ++i) {
             if (ui->_folderList->isExpanded(_model->index(i)))
                 ui->_folderList->setExpanded(_model->index(i), false);
         }
+    } else if (_model->isDirty()) {
+        // If we connect and have pending changes, show the list.
+        doExpand();
     }
+
+    // Disabling expansion of folders might require hiding the selective
+    // sync user interface buttons.
+    refreshSelectiveSyncStatus();
+
     /* set the correct label for the Account toolbox button */
     if (_accountState) {
         if (_accountState->isSignedOut()) {
@@ -748,7 +756,7 @@ AccountSettings::~AccountSettings()
 
 void AccountSettings::refreshSelectiveSyncStatus()
 {
-    bool shouldBeVisible = _model->isDirty();
+    bool shouldBeVisible = _model->isDirty() && _accountState->isConnected();
 
     QString msg;
     int cnt = 0;


src/gui/accountsettings.cpp
```diff
@@ -620,14 +620,24 @@ void AccountSettings::slotAccountStateChanged(int state)
...
-    /* check if there are expanded root items, if so, close them, if the state is different from being Connected. */
     if (state != AccountState::Connected) {
+        /* check if there are expanded root items, if so, close them */
         int i;
         for (i = 0; i < _model->rowCount(); ++i) {
             if (ui->_folderList->isExpanded(_model->index(i)))
                 ui->_folderList->setExpanded(_model->index(i), false);
         }
+
+        // Pending selective sync changes get dropped
```
ogoffart: That's sad :-(

The user did his selection and then the connection drops and the selection is gone?

guruz:
I haven't read the patch, but I think pausing the syncing to set up things (before wasting bandwidth/disk space) and then resuming the syncing is a usecase i did several times

SamuAlfageme:
@guruz but this is about disconnected accounts, not paused. 

In the same way the selective sync list is no longer displayed/no changes can be made to it, to maintain the coherence I'd also hide the 3 action buttons.

ckamm:
@ogoffart Yes, it's a bit sad. The disconnect already takes away the ability to edit the selective sync list and I agree with @SamuAlfageme that keeping the buttons without being able to see what you're doing is awkward. 

We could keep the folder list open on disconnect and just forbid further expanding of directories... but given the rarity of the situation I don't think it's worth it to spend the required engineering/testing/maintenance time on it.

Maybe it *is* best to keep it as-is. It's a bit odd, but rare and nothing bad can happen.

@SamuAlfageme Do you have strong feelings on the matter?

SamuAlfageme:
Invoking @pmaier1 and @michaelstingl for their input:

| Without this PR | With this PR |
|:---------------:|:------------:|
| log_out_present | log_out_hide |

guruz:
I vote for closing this PR.

michaelstingl:
Okay, I've got @ogoffart 's and @ckamm 's point, and it's valid. Keep it as-is, but thanks for testing it. This was definitely valuable work!

SamuAlfageme:
> The user did his selection and then the connection drops and the selection is gone?

https://github.com/owncloud/client/issues/5809 never was about loosing the selection, it was about keeping a consistent UI when the connection drops. Instead of discarding the changes the user did in the sync-list, why not keep them in memory (hide when connection drops out) and re-apply once connection is back?

ckamm:
To move this forward I've created a version that retains the selective sync selections but hides the surprising buttons on disconnection.

src/gui/accountsettings.cpp
```diff
@@ -688,14 +688,22 @@ void AccountSettings::slotAccountStateChanged()
...
     if (state != AccountState::Connected) {
+        /* check if there are expanded root items, if so, close them */
         int i;
         for (i = 0; i < _model->rowCount(); ++i) {
             if (ui->_folderList->isExpanded(_model->index(i)))
                 ui->_folderList->setExpanded(_model->index(i), false);
         }
+    } else if (_model->isDirty()) {
+        // If we connect and have pending changes, show the list.
+        doExpand();
```
ogoffart: however, does that show the "changes" if they are deep in the list?  doExpand only expand one item.  Maybe it stays expanded as before because the model was not reset?

(But even if it does not, I think it's still better than not showing it.)

SamuAlfageme:
@ckamm 

> hides the surprising buttons on disconnection.

 I'm still getting the old behavior with 2.5.0 (https://github.com/owncloud/client/commit/02e69b62a990949a9206732c8df17f1876915fd8) 😕 Both on connection refused (server down) and account log-out.

ckamm:
@SamuAlfageme This PR is not yet merged! :)

PR #5833, opened 2017-06-13
Don't link dirrectly against openssl

ogoffart:
All our crypto code is handled by qt nodaways.
No need to carry this dependency.

Especially since it causes warnings on system where there are
twp openssl version installed:
/usr/bin/ld: warning: libcrypto.so.1.0.0, needed by /usr/lib/libQt5Network.so.5.9.0, may conflict with libcrypto.so.1.1


dschmidt:
Awesome, I like!

Issue #5819, opened 2017-06-06
Triggering log in should re-ask about previously rejected certificates


ckamm:
Suggested by @SamuAlfageme  in https://github.com/owncloud/client/issues/5728#issuecomment-305225953:

Currently rejecting a certificate is permanent for the duration the client is running. So the client will never ask about a certificate again if it was rejected once unless the client is restarted.

What if the client asked again when the user explicitly triggers the "log in" action?

ogoffart:
Patch: https://github.com/owncloud/client/pull/5835

ckamm:
I've tested it.

PR #5835, opened 2017-06-13
AccountSettings: Triggering log in re-ask about previously rejected vertificates #5819

ogoffart:
Issue #5819

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -202,6 +202,7 @@ void AccountSettings::slotOpenAccountWizard()
 void AccountSettings::slotToggleSignInState()
 {
     if (_accountState->isSignedOut()) {
+        _accountState->account()->resetRejectedCertificates();
         _accountState->signIn();
     } else {
         _accountState->signOutByUi();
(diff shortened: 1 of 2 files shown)


src/gui/accountsettings.cpp
```diff
@@ -202,6 +202,7 @@ void AccountSettings::slotOpenAccountWizard()
 void AccountSettings::slotToggleSignInState()
 {
     if (_accountState->isSignedOut()) {
+        _accountState->account()->resetRejectedCertificates();
```
ckamm: I think it'd be better to have this in Account::signIn, otherwise it doesn't happen when you click login on the tray menu!

ogoffart: I made it like that on purpose because i was thinking about the scenrario in which the user has two accounts: one with a valid account, and one for which the certificate is invalid for an unknown reason (attacker eavesdropping?) for several days. After having rejected the certificate, we want to keep this account offline.  But when the user wants to go online with the other account by clicking on the system tray, we don't want to show the dialog again for the broken server.  Or do we?

ckamm: @ogoffart  I'm not sure I understand. Do you mean to say that "Log into all accounts" should not un-reject certificates? If so, then I agree. 

But clicking "log in" for one single account should be consistent no matter whether from the tray menu or the accountsettings. Therefore I suggest adding `resetRejectedCertificates` to the single-account case in `ownCloudGui::slotLogin`.

ogoffart: ah ok, you're right, i'll fix the patch.

PR #5838, opened 2017-06-14
Compare hashes for files where mtime and size match

ckamm:
Addresses https://github.com/owncloud/client/issues/5589

@moscicki

### DIFF
--- a/csync/src/csync_reconcile.c
+++ b/csync/src/csync_reconcile.c
@@ -272,11 +272,22 @@ static int _csync_merge_algorithm_visitor(void *obj, void *data) {
                     // Folders of the same path are always considered equals
                     is_conflict = false;
                 } else {
+                    // If the size or mtime is different, it's definitely a conflict.
                     is_conflict = ((other->size != cur->size) || (other->modtime != cur->modtime));
-                    // FIXME: do a binary comparision of the file here because of the following
-                    // edge case:
-                    // The files could still have different content, even though the mtime
-                    // and size are the same.
+
+                    // It could be a conflict even if size and mtime match!
+                    // When we have the remote checksum available we can detect
+                    // it without downloading the file.
+                    is_conflict |= (ctx->current == REMOTE_REPLICA ? cur->checksumHeader : other->checksumHeader) != 0;
+
+                    // NOTE: If there is no checksum, we can have !is_conflict here
+                    // even though the files have different content! This is an
+                    // intentional tradeoff. Downloading and comparing files would
+                    // be technically correct in this situation but leads to too
+                    // much waste.
+                    // In particular this kind of NEW/NEW situation with identical
+                    // sizes and mtimes pops up when the local database is lost for
+                    // whatever reason.
                 }
                 if (ctx->current == REMOTE_REPLICA) {
                     // If the files are considered equal, only update the DB with the etag from remote
--- a/csync/src/csync_reconcile.c
+++ b/csync/src/csync_reconcile.c
@@ -65,6 +65,19 @@ static c_rbnode_t *_csync_check_ignored(c_rbtree_t *tree, const char *path, int 
     }
 }
 
+/* Returns true if we're reasonably certain that hash equality
+ * for the header means content equality.
+ *
+ * Cryptographic safety is not required - this is mainly
+ * intended to rule out checksums like Adler32 that are not intended for
+ * hashing and have high likelihood of collision with particular inputs.
+ */
+static bool _csync_is_collision_safe_hash(const char *checksum_header)
+{
+    return strncmp(checksum_header, "SHA1:", 5) == 0
+        || strncmp(checksum_header, "MD5:", 4) == 0;
+}
+
 /**
  * The main function in the reconcile pass.
  *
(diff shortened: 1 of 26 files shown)


test/syncenginetestutils.h
```diff
@@ -706,17 +706,26 @@ class FakeErrorReply : public QNetworkReply
 
 class FakeQNAM : public QNetworkAccessManager
 {
+public:
+    using Observer = std::function<void(QByteArray verb, QString fileName)>;
```
ogoffart: Would it be not more generic to have the whole QNetworkRequest as a parameter? 
So we could also test "Does the call contains the header FooBar"

ogoffart: And in fact, it could return a QNetworkReply *,  which would be returned if it is not nullptr

So some tests could throw custom error

src/libsync/checksums.cpp
```diff
@@ -120,6 +132,12 @@ QByteArray contentChecksumType()
     return type;
 }
 
+QList<QByteArray> collisionSafeHashes()
+{
+    static QList<QByteArray> list = { "SHA1" };
```
ogoffart: Well well well....  haven't you heard the news about sha1?
But yeah, since we don't support anything better anyway...

csync/src/csync_reconcile.c
```diff
@@ -272,11 +272,22 @@ static int _csync_merge_algorithm_visitor(void *obj, void *data) {
...
                     is_conflict = ((other->size != cur->size) || (other->modtime != cur->modtime));
-                    // FIXME: do a binary comparision of the file here because of the following
-                    // edge case:
-                    // The files could still have different content, even though the mtime
-                    // and size are the same.
+
+                    // It could be a conflict even if size and mtime match!
+                    // When we have the remote checksum available we can detect
+                    // it without downloading the file.
+                    is_conflict |= (ctx->current == REMOTE_REPLICA ? cur->checksumHeader : other->checksumHeader) != 0;
```
ogoffart: This logic is duplicated (and different, in particular the check for collisionSafeHashes) from the logic in PropagateDownloadFile::start
Have you considered always creating the Download job  (in any cases)   and have the logic only there.

(But it's true that if it's is_conflict is strictly more "true" than the logic in PropagateDownloadFile::start, that's fine if they are different. as long as  PropagateDownloadFile does not actually fetch the file)

src/libsync/propagatedownload.cpp
```diff
@@ -344,6 +344,42 @@ void PropagateDownloadFile::start()
...
     }
 
+    // If we have a conflict where size and mtime are identical,
+    // compare the remote checksum to the local one.
+    // Maybe it's not a real conflict and no download is necessary!
+    QByteArray checksumType = parseChecksumHeaderType(_item->_checksumHeader);
+    if (_item->_instruction == CSYNC_INSTRUCTION_CONFLICT
+        && _item->_size == _item->log._other_size
+        && _item->_modtime == _item->log._other_modtime
+        && collisionSafeHashes().contains(checksumType)) {
```
ogoffart: How about doing it even if it's not a collisionsafe hash, it would still be better than before where we were not comparing anything.

test/syncenginetestutils.h
ckamm: Makes sense, will do.

src/libsync/checksums.cpp
ckamm: SHA1 may not be safe against attacks, but collisions are still overwhelmingly unlikely. Even MD5 should be okay.

csync/src/csync_reconcile.c
ckamm: Yes, this is an "early out" optimization to avoid creating unnecessary jobs. However, it won't gain us much in practice when servers generally have hashes for all files. It may be worth it to remove this logic and just keep the `is_conflict=true` case to simplify the code. Do you agree?

src/libsync/propagatedownload.cpp
ckamm: I'd rather not allow it for Adler32 - it's not designed as a hash function, so collisions are even more likely than its short 32 bit output would suggest, in particular for short strings.

ogoffart: So you accept collision without a hash (from the reconsile phase),   but not if adler32 is used?

ckamm:
@ogoffart I've fixed up the network observer addition, but left my other changes in a separate commit so they're easier to review. Let me merge when you're done, so I can wipe away that fixup commit.

csync/src/csync_reconcile.c
ogoffart: Yes, I agree that's what i had in mind.

> This will lead to a strong behavior change though: Previously, in the absence of hashes, we'd treat NEW/NEW with same mtime/size as no conflict - now we would.

I was thinking the Download job logic would be changed not to download the file if mtime and size are the same, even if there is no hash.

ogoffart:
:+1:

ckamm:
merged manually

src/libsync/propagatedownload.cpp
guruz: Adler32 is not good enough

ogoffart: You're missing the point, it goes the other way.
We were considering the file to be equal if they have the same size and mtime.
When there is a checksum, we now add a security check by comparing the checksum.

Issue #5837, opened 2017-06-14
[Sharing Dialog] Include "Show file listing" checkbox on public link section on folders


SamuAlfageme:
See: https://github.com/owncloud/core/pull/27548 

### API logic changes:
* `permissions` is set to `4` when `Show file listing` is disabled (~~shouldn't it be `6`: `CREATE` & `UPDATE` instead? cc/ @PVince81~~: it's creating new versions with `(SEQUENCE_NUMBER)` but it's not reflected in the https://github.com/owncloud/core/pull/27548#issuecomment-300905510 view -> https://github.com/owncloud/core/issues/28132)

- [x] Show file listing checkbox: https://github.com/owncloud/core/pull/27548#issuecomment-299473915
- [x] Do not display the "Direct download" option on non-listed directories: https://github.com/owncloud/core/pull/27548#issuecomment-300431656

ckamm:
It looks like this should be gated by the "supports_upload_only" capability.

ckamm:
PR https://github.com/owncloud/client/pull/5843

ckamm:
Further tracked in PR

PR #5843, opened 2017-06-15
Link shares: Add "show file listing" option #5837

ckamm:


ckamm:
@SamuAlfageme Hm, the spacing between these options doesn't look uniform on OSX. It makes a bit of sense though, since the two options are related, but wasn't intentional.

Issue #5633, opened 2017-03-20
[Regression] [Windows] Sync folder on UNC-symlinked network storage is not working anymore


bOOT-CoDe:
### Expected behaviour
Synchromize sync folder.

### Actual behaviour
An error about not accessible folder is repeated. 

### Steps to reproduce
1. set the sync folder to network drive
2. or set the sync folder to symlink on local drive, symlink's target is on the network drive (this worked in the client release 2.2.4.6408)
3. try to synchronize

### Client configuration
Client version:
2.3.0.6780
Operating system:
Windows Vista Home 32-bit
Windos 7 Enterprise 64-bit

SamuAlfageme:
Hey @bOOT-CoDe! thanks for reporting! 

I've just done a few tests with a similar setup (syncing my `Z:\\` drive, a network-mounted drive) with client version 2.3 and everything seemed to work properly: 

![smb](https://cloud.githubusercontent.com/assets/2644445/24094701/164bfda0-0d5a-11e7-9ad2-ebbef838797f.png)

Do you have any client logs or more info (e.g. which kind of filesystem runs, if the disk is encryped, etc.) to provide that could help us reproduce the issue you are running into? Thanks!

bOOT-CoDe:
Hi @SamuAlfageme,
my setup is little bit different, I have so many drive letters used therefore I am using UNC paths too. Because I was unabl to settup direct UNC path as the sync folder in ownCloud client I use symlink to folder on network drive, see the screenshot. There is another advantage, I can move sync folder to different path (e.g. another network storage) and then just update symlink to point to a new target.
![image](https://cloud.githubusercontent.com/assets/23643083/24712578/7aed026a-1a23-11e7-9ae8-61ad5e64e06c.png)
Target network disk is not encypted, it is some Synology NAS accessed over SAMBA, local and remote file systems are both NTFS. This setup is working with older ownCloud clients.

SamuAlfageme:
@bOOT-CoDe I see, that's a bit more complex scenario, I will try to reproduce. 

> An error about not accessible folder is repeated.

Can you provide some [client logs](https://doc.owncloud.org/desktop/2.3/troubleshooting.html#log-files) and the error string the client is displaying? 

Thanks a lot!

NM78:
I have excactly the same issue: https://central.owncloud.org/t/network-folder-wont-work-in-v2-3-0-v2-3-2-but-in-v2-2-4-win-desktop-client/7627/2

[...]

PR #5844, opened 2017-06-16
Journal: Don't use a ._ path if it won't work #5633

ckamm:
When synchronizing a folder on a samba share, creating files that begin
with ._ is often forbidden. This prevented the client from creating
its ._sync_abcdef.db file.

Now, it'll check whether the preferred filename is creatable, and if
it isn't it'll use .sync_abcdef.db instead.

The disadvantage is that this alternative path won't be ignored by
older clients - that was the reason for the ._ prefix.

### DIFF
--- a/src/gui/ignorelisteditor.cpp
+++ b/src/gui/ignorelisteditor.cpp
@@ -48,6 +48,9 @@ IgnoreListEditor::IgnoreListEditor(QWidget *parent) :
                          "and cannot be modified in this view.")
             .arg(QDir::toNativeSeparators(cfgFile.excludeFile(ConfigFile::SystemScope)));
 
+    addPattern(".csync_journal.db*", /*deletable=*/false, /*readonly=*/true);
+    addPattern("._sync_*.db*", /*deletable=*/false, /*readonly=*/true);
+    addPattern(".sync_*.db*", /*deletable=*/false, /*readonly=*/true);
     readIgnoreFile(cfgFile.excludeFile(ConfigFile::SystemScope), true);
     readIgnoreFile(cfgFile.excludeFile(ConfigFile::UserScope), false);
 
--- a/src/libsync/syncjournaldb.cpp
+++ b/src/libsync/syncjournaldb.cpp
@@ -17,6 +17,7 @@
 #include <QDebug>
 #include <QElapsedTimer>
 #include <QUrl>
+#include <QDir>
 
 #include "ownsql.h"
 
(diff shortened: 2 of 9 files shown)


ckamm:
@SamuAlfageme I haven't actually tested this yet!

src/libsync/syncjournaldb.cpp
```diff
@@ -56,6 +58,42 @@ QString SyncJournalDb::makeDbName(const QUrl& remoteUrl,
...
+    if (file2.open(QIODevice::ReadWrite)) {
+        // The alternative worked, use it
+        qDebug() << "Using alternate database path" << alternateJournalPath;
+        file2.close();
+        file2.remove();
+        return alternateJournalPath;
+    }
+
+    // Neither worked, just keep the original and throw errors later
+    qDebug() << "Could not find a writable database path" << file.fileName();
```
ogoffart: qCWarning  please

src/libsync/syncjournaldb.cpp
```diff
@@ -56,6 +58,42 @@ QString SyncJournalDb::makeDbName(const QUrl& remoteUrl,
...
+        file.remove();
+        return journalPath;
+    }
+
+    // Can we create it if we drop the underscore?
+    QString alternateJournalPath = journalPath.mid(2).prepend(".");
+    QFile file2(QDir(localPath).filePath(alternateJournalPath));
+    if (file2.open(QIODevice::ReadWrite)) {
+        // The alternative worked, use it
+        qDebug() << "Using alternate database path" << alternateJournalPath;
```
ogoffart: qCInfo

ckamm:
@ogoffart This patch is against 2.3, the logging change is in master only.

@guruz I agree with @ogoffart about not adding it to the exclude pattern list. I'll make it show up in the ui though.

ckamm:
@SamuAlfageme Tested now. To me it looks like
* Old and new sync folders use the old path if ._ file creation is allowed
* When I create a new sync folder on a samba folder with ._* forbidden, the new journal name is used
* When I switch on ._* veto on a samba share with an existing sync folder with a ._sync journal on it, the client creates a new journal with the .sync name (correct, because the ._ file becomes unavailable)

What I didn't test yet is the migration from .csync_journal.db to .sync_

ckamm:
@SamuAlfageme I'm glad your tests worked out. 

Doing the check and migration of journal on loading of settings seems sufficent to me - cases where ._ files become suddenly unavailable while the client is running sound vanishingly rare.

src/gui/ignorelisteditor.cpp
```diff
@@ -48,6 +48,9 @@ IgnoreListEditor::IgnoreListEditor(QWidget *parent) :
                          "and cannot be modified in this view.")
             .arg(QDir::toNativeSeparators(cfgFile.excludeFile(ConfigFile::SystemScope)));
 
+    addPattern(".csync_journal.db*", /*deletable=*/false, /*readonly=*/true);
+    addPattern("._sync_*.db*", /*deletable=*/false, /*readonly=*/true);
+    addPattern(".sync_*.db*", /*deletable=*/false, /*readonly=*/true);
```
guruz: nice!

PR #5845, opened 2017-06-19
Minor translation fixes

Cheet4h:
changes "Order" to "Ordner" in two cases.


phil-davis:
I believe that translations need to be submiited in Transifex as mention at the bottom of:
https://github.com/owncloud/client/blob/master/CONTRIBUTING.md
Then the output of the translation effort will be updated in ownCloud repos.

Issue #5850, opened 2017-06-20
chunked uploads: Don't mix upload V1 and NG


guruz:
probably because of dynamic job creation

Placeholder issue, @mrow4a and me are aware

mrow4a:
Pr which should solve the issue - https://github.com/owncloud/client/pull/5852 - it is not dynamic jobs creaton, it was dynamic chunking. 

Just also merged smashbox PR which will test new and old endpoints https://github.com/owncloud/smashbox/pull/162, in chunked/non-chunked/conflict/share/reshared/etc scenarios

guruz:
Grrrreat!

PR #5852, opened 2017-06-20
Classify chunked items correctly. Issue #5850

mrow4a:
This should solve https://github.com/owncloud/client/issues/5850. 

Please note: we cannot set chunk size of PUT V1 to _chunkSize which is dynamic because PUT V1 will not adjust dynamic chunk size. If we want it, we need to make code adjusting dynamic chunk size both for V1 and NG.

### DIFF
--- a/src/libsync/discoveryphase.h
+++ b/src/libsync/discoveryphase.h
@@ -53,9 +53,10 @@ struct SyncOptions
     /** If a confirmation should be asked for external storages */
     bool _confirmExternalStorage;
 
-    /** The initial un-adjusted chunk size in bytes for chunked uploads
+    /** The initial un-adjusted chunk size in bytes for chunked uploads, both
+     * for old and new chunking algorithm, which classifies the item to be chunked
      *
-     * When dynamic chunk size adjustments are done, this is the
+     * In chunkingNG, when dynamic chunk size adjustments are done, this is the
      * starting value and is then gradually adjusted within the
      * minChunkSize / maxChunkSize bounds.
      */
--- a/src/libsync/owncloudpropagator.cpp
+++ b/src/libsync/owncloudpropagator.cpp
@@ -379,7 +379,8 @@ PropagateItemJob *OwncloudPropagator::createJob(const SyncFileItemPtr &item)
             return job;
         } else {
             PropagateUploadFileCommon *job = 0;
-            if (item->_size > _chunkSize && account()->capabilities().chunkingNg()) {
+            if (item->_size > syncOptions()._initialChunkSize && account()->capabilities().chunkingNg()) {
+                // Item is above _initialChunkSize, thus will be classified as to be chunked
                 job = new PropagateUploadFileNG(this, item);
             } else {
                 job = new PropagateUploadFileV1(this, item);
--- a/src/libsync/propagateuploadng.cpp
+++ b/src/libsync/propagateuploadng.cpp
@@ -400,6 +400,7 @@ void PropagateUploadFileNG::slotPutFinished()
         // the chunk sizes a bit.
         quint64 targetSize = (propagator()->_chunkSize + predictedGoodSize) / 2;
 
+        // Adjust the dynamic chunk size _chunkSize used for sizing of the item's chunks to be send
         propagator()->_chunkSize = qBound(
             propagator()->syncOptions()._minChunkSize,
             targetSize,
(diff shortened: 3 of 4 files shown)


mrow4a:
selection_029

src/libsync/owncloudpropagator.cpp
```diff
@@ -379,7 +379,8 @@ PropagateItemJob *OwncloudPropagator::createJob(const SyncFileItemPtr &item)
             return job;
         } else {
             PropagateUploadFileCommon *job = 0;
-            if (item->_size > _chunkSize && account()->capabilities().chunkingNg()) {
```
mrow4a: Because of that, after dynamic chunk has been readjusted to 100MB, this file has been classified as V1, being 50MB in size. With _initialChunkSize being 10MB, it is classified as to be chunked and having high dynamic chunk, being send in one shot.

src/libsync/propagateuploadng.cpp
```diff
@@ -400,6 +400,7 @@ void PropagateUploadFileNG::slotPutFinished()
         // the chunk sizes a bit.
         quint64 targetSize = (propagator()->_chunkSize + predictedGoodSize) / 2;
 
+        // Adjust chunk size for the item being classified to be chunked.
```
ogoffart: this comment is wrong.  This has nothing to do with the classification, quite the contrary.  It is the actual size that will be used by the chunking-ng algorithm

src/libsync/owncloudpropagator.cpp
ogoffart: yes, that's correct.  Thanks for spotting and fixing this.

src/libsync/propagateuploadng.cpp
mrow4a: Hmm, maybe I need to rewrite this. The intiention was to say that this is the dynamic chunk size of the item, which has been classified to be chunked.

src/libsync/propagateuploadv1.cpp
```diff
@@ -79,6 +79,7 @@ void PropagateUploadFileV1::startNextChunk()
     qint64 chunkStart = 0;
     qint64 currentChunkSize = fileSize;
     bool isFinalChunk = false;
+
```
guruz: If you anyway change the commit, remove this one too :)

mrow4a:
Done

src/libsync/discoveryphase.h
```diff
@@ -53,9 +53,10 @@ struct SyncOptions
     /** If a confirmation should be asked for external storages */
     bool _confirmExternalStorage;
 
-    /** The initial un-adjusted chunk size in bytes for chunked uploads
+    /** The initial un-adjusted chunk size in bytes for chunked uploads, both
+     * for old and new chunking algorithm, which classifies the item to be chunked
```
ckamm: When you say "which classifies the item to be chunked" (also below) that's confusing because it's ambiguous. "which is used to decide whether an item should be chunked" would be clearer.

mrow4a:
@guruz why the test is not passing (it is also not for master), can we merge this PR?

ogoffart:
@mrow4a  the jenkins error seems to be unrelated to this change:

```
6: QWARN  : TestSyncJournalDB::testFileRecord() sync.database.sql: Sqlite exec statement error: 5 "database is locked" in "INSERT OR IGNORE INTO checksumtype (name) VALUES (?1)"
6: QWARN  : TestSyncJournalDB::testFileRecord() sync.database.sql: Sqlite exec statement error: 5 "database is locked" in "INSERT OR REPLACE INTO metadata (phash, pathlen, path, inode, uid, gid, mode, modtime, type, md5, fileid, remotePerm, filesize, ignoredChildrenRemote, contentChecksum, contentChecksumTypeId) VALUES (?1 , ?2, ?3 , ?4 , ?5 , ?6 , ?7,  ?8 , ?9 , ?10, ?11, ?12, ?13, ?14, ?15, ?16);"
6: FAIL!  : TestSyncJournalDB::testFileRecord() 'storedRecord == record' returned FALSE. ()
6:    Loc: [/var/lib/jenkins/workspace/cloud-client_client_PR-5852-5C3JT3EY7EHDKWSDVRI76HKJJMTNLBZ76KGHLTTGG25CKR2ASZYA/test/testsyncjournaldb.cpp(78)]
```

I'd say we can merge this change. yes

Issue #5516, opened 2017-02-05
Errors appear briefly and then are gone


dbolton:
### Expected behaviour
If there is an error that needs user attention, then the user should get a chance to read the error at their own pace (or copy the error to the clipboard if needed)

### Actual behaviour
The error message appears briefly as a Windows notification, but the notification/error message disappears before I get a chance to find the file location, or if I've stepped away from the computer when the sync happens, I never see the message. If there are multiple errors, the windows notification says the first error and then says something like "and 2 more errors" without details. 

If I open the error log with F12 it only shows logs starting from the moment I open the log window, not the errors in the last 10 minutes or so. 

Examples of these errors include:
 * Conflicting files (edited by two users at once)
 * File name conflicts (sometimes happens if I rename a file and only change a letter from upper to lowercase)

### Client configuration
Client version: Version 2.2.4 (build 6408). 

Operating system: Windows 10

OS language: US English

Qt version used by client package (Linux only, see also Settings dialog):

Client package (From ownCloud or distro) (Linux only):

Installation path of client:

SamuAlfageme:
@dbolton but you would still have info on the non-synced errors on the `Activity > Not Synced` tab, these include both error scenarios you pointed out, they also show in the account tab, right next to the folder sync connection that presents problems. 

True that the windows notifications could be linked to the first view or be persistent (not be closed automatically, until the user does). What's your opinion on this? 

Thanks for reporting! 😉

dbolton:
The Activity > Not Synced tab shows a list of ignored files (like Microsoft Word hidden temporary files), but it does not list file conflicts or file name conflicts. (I just created a file conflict to double check)

hodyroff:
Would not a 4th tab for errors make most sense? If I like to check for conflict files I would not want to go through all the ignored once and not go through all my activities, rather have them in one view: "conflict and errors" maybe? Lets collect/discuss ideas.

dbolton:
The ignore files list isn't very long on my computer so I'm fine with them there or in a separate fourth tab.  My main concern is that the errors are listed for longer than a moment so that the are available for reference when troubleshooting.

pmaier1:
So a 4th tab for errors definitely makes sense, some ideas:
- errors should stay permanently in a list until they are resolved
- would be nice to also provide some kind of error handling (e.g. ignore the file or similar)
- Error notification should pop-up once and need to be confirmed (not with every 'sync run' again and again)
[...]

PR #5861, opened 2017-06-27
Add a more functional error view #5516

ckamm:
* Allow filtering of ignores and warnings to see only important bits.
* Navigate from the folder view to the error view by clicking on the
  error list with the red background.
* Move the error list into its own ui file to allow easier extension.
* Fix issue around tab id handling in ActivitySettings.
* Rename "Action" column to "Issue".

Not done:
* Protocol and Issue view probably want to share the same underlying
  model.
* Blacklisting of errors currently moves them into the 'ignored file'
  category; that should change because then they don't appear as errors
  and are easily filtered away.
* How should non-file errors be displayed? (e.g. "not enough disk space
  to start the sync!")

### DIFF
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -21,6 +21,7 @@ set(client_UI
     ignorelisteditor.ui
     networksettings.ui
     protocolwidget.ui
+    issueswidget.ui
     activitywidget.ui
     synclogdialog.ui
     settingsdialog.ui
@@ -64,6 +65,7 @@ set(client_SRCS
     owncloudgui.cpp
     owncloudsetupwizard.cpp
     protocolwidget.cpp
+    issueswidget.cpp
     activitydata.cpp
     activitylistmodel.cpp
     activitywidget.cpp
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -132,6 +132,8 @@ AccountSettings::AccountSettings(AccountState *accountState, QWidget *parent)
     auto mouseCursorChanger = new MouseCursorChanger(this);
     mouseCursorChanger->folderList = ui->_folderList;
     mouseCursorChanger->model = _model;
+    ui->_folderList->setMouseTracking(true);
+    ui->_folderList->setAttribute(Qt::WA_Hover, true);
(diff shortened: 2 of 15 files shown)


ckamm:
Can now navigate from the folder settings page
shot1

to the errors list, and it can filter away ignores and warnings:

shot2

SamuAlfageme:
@ckamm this looks great!! 😍

guruz:
> Protocol and Issue view probably want to share the same underlying
model.

Is that necessary? If it works already now it's fine?

> Blacklisting of errors currently moves them into the 'ignored file'
category; that should change because then they don't appear as errors
and are easily filtered away.

Done https://github.com/owncloud/client/pull/5865 ?

> How should non-file errors be displayed? (e.g. "not enough disk space
to start the sync!")

Maybe it's enough to keep them on account tab?

guruz:
Could you make the margin(?) higher on both bottom and top? The 'p' is cut off..

errordisplay

Would it be possible to remove the "Click for error list"? It looks a bit out of place and not like something clickable. Maybe just have the whole read area clickable? (possiblye with hand-cursor?)

guruz:
* [ ] Make ```ActivitySettings::slotShowIssuesTab``` work on macOS (where we use different tab logic..) @guruz

guruz:
Except for my comments the code looks good :)

I'm not super happy about the UX but we can iterate on that

ckamm:
@guruz 
> Unified model

Yes, not strictly necessary. There's a bit of code duplication/long range interaction between the issues and the protocol view, but it's not horrible.

> Blacklisting

Yes #5865 addresses that.

> "Click for error list"

I added that to make it discoverable that the red box is clickable (it's the whole box). Maybe a cursor and tooltip message would also do it. We should definitely say something whenever we elide errors from that view, in my opinion.

The obvious thing would be a button, but I couldn't make that look good. Further iterations on making this pretty are definitely appreciated.

My main goal here was to have a dedicated per-account error view.

dragotin:
Nice! However, in case a user only has one account, could you hide the Account-Combo? And similar for the folders: If someone does not have folders defined, or only one, it would be good to hide that combo. 

That will unclutter the UI for inexperienced users :-)

ckamm:
@dragotin Nice idea, I'll do that!

Issue #5862, opened 2017-06-27
Link to central.owncloud.org leads to 404 in client


mghansen256:
In the OwnCloud client under "General" (Allgemein in German) is a link to central.owncloud.org, leading to https://central.owncloud.org/c/help/desktop-file-sync, which gives an error 404 upon following the link.

### Client configuration
Client version: 2.3.2 build 6928
Operating system: Windows 7
OS language: German

SamuAlfageme:
@mghansen256 indeed! central.owncloud.com categories were recently redone. I'm updating the link everywhere it appears in the repo. Thanks for the heads up! 

PR in https://github.com/owncloud/client/pull/5869

PR #5869, opened 2017-06-28
Update central.owncloud.org link (#5862)

SamuAlfageme:
Desktop client category was renamed, old link was 404 - see: https://central.owncloud.org/t/new-central-categories-you-decide

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -2,7 +2,7 @@
 Please try to only report a bug if it happens with the latest version
 The latest version can be seen by checking the ChangeLog: https://owncloud.org/changelog/desktop/
 
-For support try: https://central.owncloud.org/c/help/desktop-file-sync
+For support try: https://central.owncloud.org/c/desktop-client
 --->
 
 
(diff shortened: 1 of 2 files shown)


guruz:
FYI @michaelstingl  @lefherz

guruz:
Slightly modified for 2.3 and pushed there too aeea27b57b8496e7dc375b341f1bd0b22c8297a9

lefherz:
actually, we need this link to redirect to https://central.owncloud.org/c/desktop-client/help instead. it is a central subcategory specifically dedicated to Support - this is where we want that stuff to happen, not in the main category.

If it is too late for such changes, then just drop them though. it is not that important, we can also just do more support stuff in the https://central.owncloud.org/c/desktop-client. but directing this to https://central.owncloud.org/c/desktop-client/help is the right way to do it.

guruz:
@mrlinux1 Can you add a redirect/rewrite from https://central.owncloud.org/c/help/desktop-file-sync to https://central.owncloud.org/c/desktop-client/help ? thx

Issue #5719, opened 2017-04-21
[Linux] Support non-utf8 locale for filename


guruz:
FYI @ogoffart @DeepDiver1975 

Related https://github.com/owncloud/client/issues/5676 ```[Feature] Filename encoding with gbk standard should be supported```

```
<ogoffart> There is no concept of fs encoding on linux
<ogoffart> Everything has to be encoded with the locale
<jturcotte> You'll also have to deal with chars that you can't convert to his encoding, and you want to make sure that if the user modifies that file that it's sent back to the server with the right name.
<jturcotte> And if the user sees blah???.png weird characters in the file, maybe he'll feel like he has to rename it too, so it shouldn't look too weird.
```

Challenges: How to handle invalid encodings?
https://github.com/owncloud/client/issues/4811#issuecomment-218076160

ogoffart:
(I changed the title because there is no need to involve the server, it's purely a client issue)

How to handle invalid encoding?:  The way we currently do it in SyncEngine::treewalkFile, for example.
https://github.com/owncloud/client/blob/e86499d9907f3bab5eac6e371eef772f8befd2e5/src/libsync/syncengine.cpp#L364
(however this has to happen at the time of conversion from the system locale, which has to happen in csync, before the reconcile phase.)

guruz:
Moving to 2.5 as https://github.com/owncloud/client/issues/5676 was moved too

ogoffart:
Duplicate of #5676

PR #5715, opened 2017-04-19
macOS: Emoji support WiP #5661

guruz:

Issue #5676, opened 2017-04-03
[Feature] Filename encoding with gbk standard should be supported


ericlin1001:
### Steps to reproduce
1. In linux, use owncloud client(owncloud version 2.3.0 Using Qt 4.8.6)
2. In the owncloud syncing directory.
  1. `touch 乱码.txt`
  2. `convmv -f utf-8 -t gbk 乱码.txt`
  3. owncloud complained: *???.txt  could not be synced due to errors. *
3. In log, I found: *File ignored because of invalid utf-8 sequence: ???.txt*

### Expected behaviour
Tell us what should happen
The file with name encoding in gbk, should be synced. Because I'm converting file name encoding of a large amount of files in linux, and these files are to be used in windows(win7).

### Actual behaviour
Tell us what happens instead
1. owncloud complained: *???.txt  could not be synced due to errors. *
2. In log, I found: *File ignored because of invalid utf-8 sequence: ???.txt*

### Server configuration
**Operating system**:
LSB Version:	core-2.0-amd64:core-2.0-noarch:core-3.0-amd64:core-3.0-noarch:core-3.1-amd64:core-3.1-noarch:core-3.2-amd64:core-3.2-noarch:core-4.0-amd64:core-4.0-noarch:core-4.1-amd64:core-4.1-noarch:security-4.0-amd64:security-4.0-noarch:security-4.1-amd64:security-4.1-noarch
Distributor ID:	Ubuntu
Description:	Ubuntu 14.04.5 LTS
Release:	14.04
Codename:	trusty

**Web server:**

**Database:**

**PHP version:**

**ownCloud version:** (see ownCloud admin page)

**Updated from an older ownCloud or fresh install:**

**Where did you install ownCloud from:**

**Signing status (ownCloud 9.0 and above):**

```
Login as admin user into your ownCloud and access 
http://example.com/index.php/settings/integrity/failed 
paste the results here.
```

**List of activated apps:**

```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your ownCloud installation folder
```

**The content of config/config.php:**

```
If you have access to your command line run e.g.:
sudo -u www-data php occ config:list system
from within your ownCloud installation folder

or 

Insert your config.php content here
(Without the database password, passwordsalt and secret)
```

**Are you using external storage, if yes which one:** local/smb/sftp/...

**Are you using encryption:** yes/no

**Are you using an external user-backend, if yes which one:** LDAP/ActiveDirectory/Webdav/...

[...]

Issue #5661, opened 2017-03-28
[OC 10.0][OS X] libsync is able to download files with 4 byte-unicode names but upload etc


SamuAlfageme:
4 byte-unicode (a.k.a. emoji) filenames are supported in ownCloud 10.0 (https://github.com/owncloud/core/issues/4513)

### Expected behavior
The client should be able to handle this filenames as long as the underlying filesystem supports them as well. 

### Actual behavior
- [x] Sync a file/folder with an emoji filename **from the server into the desktop client** (`v2.3.1`)
- [x] Upload a file/folder to the server with a 4byte-unicode name prompts [this error](https://github.com/owncloud/client/blob/master/src/libsync/syncengine.cpp#L678-L681)
- [x] Delete/rename the file/folder on the server, the client doesn't keep up with these changes

<p align="center">
  <img src="https://cloud.githubusercontent.com/assets/2644445/24409389/5a760080-13d0-11e7-8bc7-e724df86305a.gif"/>
</p>

Solving this will be enough to close some old issues: 
- https://github.com/owncloud/client/issues/4811
- https://github.com/owncloud/client/issues/5231

guruz:
Note that this might be macOS specific because we have to do file name mangling on macOS https://github.com/owncloud/client/search?utf8=%E2%9C%93&q=normalization&type=  . 

How does it look on other OS?

@mstingl this would mandate an earlier 2.3.2

SamuAlfageme:
@guruz you're right, I assumed the name-conversion error was platform-independent but just tried with Win10 and no flags were raised there. My bad 😅

michaelstingl:
> @michaelstingl this would mandate an earlier 2.3.2

Works for me

ogoffart:
I bet that's because iconv can't convert the string (cf. https://github.com/owncloud/client/issues/2777 )

guruz:
@ogoffart Any proposal for your Qt code?
Alternatively 2: Instead of using POSIX/csync _treaddir we could use something from Qt for reading directories.

guruz:
maybe https://doc.qt.io/qt-5/qfile.html#decodeName-1 ..

guruz:
This is messy. I'm trying a few things but I think this is too invasive for a patch release.

guruz:
@michaelstingl @pmaier1 Moving to 2.4, the needed patch is too invasive

SamuAlfageme:
👍

PR #5875, opened 2017-07-04
csync: Use Qt for encoding conversion instead of iconv

ogoffart:
Replaces #5715

Should fix issues:

- #5661   On mac, iconv did not support all of unicode and some files with emoji in the filename could not be uploaded
    
- #5719 , #5676  On linux, we will now support non utf-8 locale


ckamm:
This looks good, in particular the dependency removal and being able to use Qt in csync.

I worry about testing a bit - if this introduces issues we'll only notice by seeing bad filenames on users' machines. We should run the encoding tests on all platforms (I assume you covered linux?), and maybe check how thorough they are.

csync/src/std/c_utf8.cc
```diff
@@ -64,26 +60,33 @@ char* c_utf8_from_locale(const mbchar_t *wstr)
     WideCharToMultiByte(CP_UTF8, 0, wstr, len, mdst, size_needed, NULL, NULL);
     dst = mdst;
   }
+  return dst;
 #else
-  dst = c_strdup(wstr);
+    QTextDecoder dec(QTextCodec::codecForLocale());
+    QString s = dec.toUnicode(wstr, qstrlen(wstr));
```
dragotin: `const QString s = ....` ?

Tested on Linux and works, but might be an issue on MacOSX

ogoffart: 's' can't be const for two reason:  First, on mac, we update it with `s = s.normalized(QString::NormalizationForm_C);`  
Second, we std::move it before calling toUtf8. This saves an allocation as there is an overload of toUtf8 which does the conversion in-place. (but does not work if s was const)

ogoffart:
Regarding the  check_encoding.c, it is true that the #ifdef HAVE_ICONV will not be compiled and could be cleaned up.  I just kept it because i did not want to change the test. It should anyway pass with or without iconv as the behavior should be the same. (except for the codepoints that were not understood by macos's iconv)


The CI is failiong for unrelated reason (it ran out of disk)

Any objection for merging this?

ogoffart:
@jturcotte quickly tested this patch on windows and did not notice regressions.

Since there are no objections, i'm merging now.

guruz:
Some troubles with APFS see here: https://github.com/owncloud/client/issues/5650
(But they appear unrelated to this PR..)

PR #5890, opened 2017-07-11
Error handling improvements

ckamm:
Various improvements to the issues tab and to local and remote insufficient storage errors.


src/libsync/propagateupload.cpp
```diff
@@ -522,6 +522,29 @@ void PropagateUploadFileCommon::checkResettingErrors()
     }
 }
 
+void PropagateUploadFileCommon::commonErrorHandling(AbstractNetworkJob *job)
```
ogoffart: Great to reduce some duplication :+1:

PR #5892, opened 2017-07-12
PropagateUpload: Model of remote quota, avoid some uploads #5537

ckamm:
When we see a 507 error, assume that quota is < uploaded size.

### DIFF
--- a/src/libsync/propagateupload.cpp
+++ b/src/libsync/propagateupload.cpp
@@ -176,6 +176,17 @@ void PropagateUploadFileCommon::start()
         return;
     }
 
+    // Check if we believe that the upload will fail due to remote quota limits
+    const quint64 quotaGuess = propagator()->_folderQuota.value(
+        QFileInfo(_item->_file).path(), std::numeric_limits<quint64>::max());
+    if (_item->_size > quotaGuess) {
+        // Necessary for blacklisting logic
+        _item->_httpErrorCode = 507;
+        emit propagator()->insufficientRemoteStorage();
+        done(SyncFileItem::DetailError, tr("Upload of %1 exceeds the quota for the folder").arg(Utility::octetsToString(_item->_size)));
+        return;
+    }
+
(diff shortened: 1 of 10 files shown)


ckamm:
@ogoffart I'll add a test if you think doing this makes sense -- writing the test will probably take as long as writing the feature.

src/libsync/propagateupload.cpp
```diff
@@ -197,6 +197,18 @@ void PropagateUploadFileCommon::start()
...
+    const quint64 quotaGuess = propagator()->_folderQuota.value(
+        QFileInfo(_item->_file).path(), std::numeric_limits<quint64>::max());
+    if (_item->_size > quotaGuess) {
+        // Necessary for blacklisting logic
+        _item->_httpErrorCode = 507;
+        _item->_errorMayBeBlacklisted = true;
+        emit propagator()->insufficientRemoteStorage();
+        done(SyncFileItem::BlacklistedError, tr("Upload of %1 exceeds the quota for the folder").arg(Utility::octetsToString(_item->_size)));
+        return;
+    }
```
ogoffart:
```
else  {
propagator()->_folderQuota[QFileInfo(_item->_file).path()] -= _item->_size;
}
```


(you probably want to use an  iterator to avoid multiple lookups, and only do it if there is a value there)

ogoffart:
the last commit conflicts with the HTTP/2 change.  I'm afraid we can't put the logic to compute the max number of job there with HTTP/2 as we con't have a pointer to the account.
Maybe what should be configured is anyway not these two number, but only a boolean telling if we should do parallel or not.

ckamm:
@ogoffart Is the HTTP/2 change already merged? I think the suggestion to configure only parallelism-or-not makes sense and I'd adapt it accordingly.

ogoffart:
Yes, HTTP/2 is merged now.

ckamm:
@ogoffart I updated the PR accordingly, please rereview!

Issue #5273, opened 2016-10-25
read-only share: locally modified file mutliple times should create multiple conflict files (?)


moscicki:
ownCloud 2.2.4 on MacOSX.

In read-only share I modified a file b.txt which was fully synchronized initially. I had to add the write-bit because it was read-only. The system tried to revert the file but since it found a modification it created a conflict file. The sync was green at the end. That's all expected.

```
%ls -l
   -r--r--r--  1 moscicki  staff  23 Aug 23 16:11 b.txt
   -rw-r--r--  1 moscicki  staff   2 Oct 25 13:23 b_conflict-20160823-161149.txt
```

![image](https://cloud.githubusercontent.com/assets/621965/19684312/d0ca99a0-9ab6-11e6-8e05-515781dfcdc8.png)

When I then again modified file b.txt the system did not create the second conflict file as it should. Instead it left the b.txt modified locally. The resulting log was this:

![image](https://cloud.githubusercontent.com/assets/621965/19684239/734ecad0-9ab6-11e6-846d-d5c1b548c7f3.png)

When I subsequently removed the conflict file b.txt was restored from the server and a new conflict file created. 

Should it create multiple conflict files in this cases to have consistent behaviour?

ckamm:
@moscicki Yes, I agree there should be multiple conflict files.

ckamm:
@dragotin @ogoffart @guruz What's the reason we use the server's timestamp during conflict file name creation? Wouldn't the local mtime work just as well? I'm wondering whether I should change this generally, and not just for restorations.

ogoffart:
Yes, i suppose using the local time is actually better.

ckamm:
Pull request including change and test: https://github.com/owncloud/client/pull/5300

SamuAlfageme:
@ckamm Yup, now the conflict files are created with the local timestamp (`<file>_conflict-yyyyMMdd-HHmmss.<ext>`) and the original file is restored, there are some overlay-related issues though:

* The ones in the conflict files are initially in 🔁 Syncing status until you close the folder and reopen it.
* The original file appears with the ⚠️ overlay as it's not being correctly synced (instead of the original shared green one). This one is kept even when you reopen the folder.

![conflict_overlay](https://cloud.githubusercontent.com/assets/2644445/20751951/f02db69c-b6fe-11e6-8ef9-3f2e8a271bd2.gif)

[...]

PR #5894, opened 2017-07-12
Conflict view

ckamm:
Shows new and old conflicts in the issue tab:

spectacle n29588
spectacle x29588


ogoffart:
Looks good to me.

Just wondering if the user will understand how to resolve the conflict

ckamm:
@ogoffart Indeed. Maybe we want to give a link to a page that explains conflicts and ways of resolving them nearby? I thought about whether there should be any buttons on this item, but having "quick resolution" buttons would most likely be dangerous.

Also: I think I'll update the message to include a "Click for details." sentence, to make it clearer that one can click the box to go to the issues tab.

SamuAlfageme:
Once again, really amazing feature to ship in 2.4 ☺️

I have a couple of doubts: 

1. What will happen with non-writable folder conflicts*? Could this be extended to support those too? - see related https://github.com/owncloud/client/issues/5273 
2. Maybe we could adapt something similar as the webUI for conflict resolution?

<p align="center">
  
</p>

(*) -> not really _"conflicts"_ 😁

ckamm:
@SamuAlfageme Can you be more specific about 1. #5273? I don't see how it relates. All existing conflict files will now lead to this message and will be listed in the issues tab.

About 2.: Indeed, we could now provide a better conflict resolution user interface. It needs discussion and design though. The webui like dialog wouldn't be particularly useful in most cases though: Usually users will need to look at both files and create a merge manually.

I think there should be actions for
* "Open folder" (similar to clicking the conflict list entry now)
* "Open synchronized server version"
* "Open unsynchronized local conflict copy"
* "The conflict is resolved, delete the conflict file"
* "The conflict is resolved, overwrite the file with the conflict file"
* "The conflict is resolved, keep both files renaming the conflict file"

That would allow opening, merging and resolving in a convenient way. I'll make a new ticket for that.

SamuAlfageme:
@ckamm - about 1 - I was just wondering if it might be useful to offer the user the chance to see all the "_conflict_" files (e.g. each save he did on the file) that were created (in this same tab) and maybe the count/list of local "_conflicts_" (versions). As I write it, I realize that could be a bit overkill though. 

I can imagine a scenario when the user does not care about having a local file unsynced on a read-only folder and making edits leads him to many of these files. 

Cool ideas in #5896!

ckamm:
Merged

Issue #5853, opened 2017-06-22
Send a unique X-Request-ID header to allow tracing requests in apache and owncloud logs.


butonic:
https://github.com/owncloud/core/pull/28144 has been merged. It will use the value of a `X-Request-ID` header as the request id in log files to allow linking events on the client to events in the apache and owncloud logs.

- [x] generate a string to use as unique id.
  - 20-200 chars
  - only `[a-zA-Z0-9-+/_=.:]` allowed
- [x] use it in the client log
- [x] send it as `X-Request-ID` header on every request

ckamm:
@butonic @ogoffart Should this id be saved in the config file? Identifying clients across restarts would be handy for debugging sometimes, but the down side would be that you could identify clients across restarts. (and some people copy config files around, leading to clients with the same request ids)

ckamm:
PR https://github.com/owncloud/client/pull/5897

ckamm:
Closed because I tested with mitmproxy and I don't think it needs an extra verification step.

guruz:
Re-opened as there was a misunderstanding between us..

PR #5897, opened 2017-07-13
Account: Send a X-Request-ID header #5853

ckamm:


src/cmd/cmd.cpp
```diff
@@ -311,7 +311,7 @@ int main(int argc, char **argv)
     qputenv("OPENSSL_CONF", opensslConf.toLocal8Bit());
 #endif
 
-    qsrand(QTime::currentTime().msec() * QCoreApplication::applicationPid());
+    qsrand(QDateTime::currentMSecsSinceEpoch());
```
guruz: Why did you remove this? I remember this was be added for smashbox doing parallel testing which had before this lead to duplicate transferids used in chunking

ckamm: Oh, makes sense to leave in the pid then. My complaint was that msec() just has values between 0-999.

ckamm:
@ogoffart QUuid has curly braces, which can't appear. And not being limited to hex digits allows for shorter ids. Will update.

guruz:
People on the internet seem to be using this header with smaller values separated by dashes.. (looks UUIDish..)

but no braces..

src/cmd/cmd.cpp
```diff
@@ -311,7 +311,7 @@ int main(int argc, char **argv)
     qputenv("OPENSSL_CONF", opensslConf.toLocal8Bit());
 #endif
 
-    qsrand(QTime::currentTime().msec() * QCoreApplication::applicationPid());
+    qsrand(QDateTime::currentMSecsSinceEpoch() * QCoreApplication::applicationPid());
```
ogoffart: could this overflow? both calls return (signed) qint64  
  Epoch should be ~40 bits and i don't know how many bits is PID on windows for example.
could be worth casting to unsiged.

I would suggest   qHash(QDateTime::currentMSecsSinceEpoch()) ^ qhash(QCoreApplication::applicationPid())

butonic:
Uuid is perfect as request id.

butonic:
We could allow curly braces on the server side. Discuss with @PVince81 or @tomneedham while I am on vacation ;-)

butonic:
Oh and yes, the request id must be uniqe for every request so it can be logged and traced through the whole server side stack.

Issue #5893, opened 2017-07-12
[OAuth 2] Extend the "reopen browser to proceed" and error handling logic to accounts already created


SamuAlfageme:
When the access to an account has been revoked (access/refresh token combination deleted/invalidated) there's no recovery option if the browser was closed but to restart the client.

### Expected behavior
The account tab should implement the same behavior the wizard does when waiting for the redirect URL (i.e. the chance to relaunch the browser tab in case it was lost and timeout/error handling)

![wizard](https://user-images.githubusercontent.com/2644445/27023405-aeeff5c2-4f52-11e7-8876-b1fac7b728b3.png)

### Actual behavior
The client does launch a new browser tab/window to authenticate with the application. If this window is closed, the account is left in a "configuration error" state with no chance to relaunch it again until the client itself is restarted (and the browser respawned). Transitions to "logged in" should not be possible.

![blank diagram - page 1 2](https://cloud.githubusercontent.com/assets/2644445/21677670/6b6714a4-d33b-11e6-80b6-468a8940cc3b.png)

### Steps to reproduce
1. Add an account via OAuth authentication 
2. Remove the access & refresh token from the server tables (= revoke)
3. Relaunch the client / force sync. 
4. Close the browser.

ogoffart:
Pull request: https://github.com/owncloud/client/pull/5898

SamuAlfageme:
@ogoffart can we disable the "log out" option from the "Account" dropdown & tray menu while "Obtaining authorization from the browser"? It's causing the AccountState to go into the `"Disconnected" <-> "Signed out"` loop:

![no_connection](https://user-images.githubusercontent.com/2644445/28457115-2ec21048-6e05-11e7-82f0-0a40f1706a59.gif)

Either that, or trigger the "Obtaining authorization from the browser" on "Log in" selection.

SamuAlfageme:
👍

PR #5898, opened 2017-07-13
OAuth2: Have a link to the browser in the owncloud UI

ogoffart:
When the browser is open, ad a link in the ui to re-open
the browser.

Issue #5893


ogoffart:
I tested pure HTTP and it works fine. (I did not tested shib)

I'm confident that this should not disturb the AccountState machine as i grepped for all usages of "ConfigurationError".  I could have avoided adding AskingCredentials as a different state and just do that logic in the AccountSettings for ConfigurationError, but i think it is more clean this way.

Issue #5895, opened 2017-07-12
[OAuth 2][Multiaccount] Determine how to login into multiple logged out accounts


SamuAlfageme:
### Steps to reproduce

1 . Login to one OAuth server using the client
2. Open server in the web browser & log out
3. Repeat 1 **with a different account** on the client
4. One of the following:
 - Log out of both accounts on the client
 - Remove both authorization & refresh tokens from the server DB 
 - Delete the cookies & keychain entries from the client
5. Restart the client

Just one login page is triggered from the client, the rest of the accounts will stay in the same state as the one in https://github.com/owncloud/client/issues/5893 ("Configuration Error") until all the accounts complete the log-in process from the account tab / the client is restarted again as many times as accounts logged-out. 

### Actual behavior
Depending on which account is the one launching the auth. flow from the client, just one (the first) is able to login. If using the wrong credentials a message is displayed:

<p align="center">
  <img src="https://user-images.githubusercontent.com/2644445/28127430-d2c75ba6-672c-11e7-844c-4ee063953b55.png"/>
</p>

...but up until this point, the user does not know on which account is he trying to login to. Also, since the session is kept in the browser after each authorization, it's quite tricky to recover and re-login to all accounts.

### Expected behavior
Don't really know how the best way to deal with this scenario. Might as well do some research on how other multi-account applications that use OAuth2 w/ the browser do it. 

Some ideas: 

* In SAML auth flow we spawned as many independent web-views as accounts created in the client
    - Maybe we could force the browser to launch in "incognito mode" (e.g. flag to `xdg-open`) - but I don't really think this is even possible / an optimal cross-platform solution.
* Queue the logins combined with https://github.com/owncloud/oauth2/issues/48
    - cc/ @ogoffart should we escalate this to the oauth2 repo to force the application logins/authorizations not to leave a valid session open behind on the browser? 
* **Do not trigger** any login attempt and leave the user do it manually. 

cc/ @DeepDiver1975 @butonic

ogoffart:
What could be done better already in the client is that we could show the error in the browser instead of redirecting to the login page.

I believe otherwise the best way is that we tell the oauth app a hint about the user, as i wrote in owncloud/oauth2#48  then there would be different tab/window for each, and the user will have to authenticate on each tab.  

>  Maybe we could force the browser to launch in "incognito mode" 

One of the advantage of the oauth process is that it does not require entering the password if one is logged in the browser already. So doing that would be unfortunate.
[...]

PR #5901, opened 2017-07-13
OAuth: Better message when loggin in with the wrong username

ogoffart:
Since the user is already in the browser, put the error message in the
browser with a message to log out and then log in as the right user.

Issue #5895


ogoffart:
Ideally not required if  owncloud/oauth2#48  gets implemented

Issue #5692, opened 2017-04-10
owncloudcmd:  Don't start the sync when the capability call fails.


ogoffart:
Reported here:  https://github.com/owncloud/client/issues/3936#issuecomment-292921020

 owncloudcmd still starts the sync even tough the capabilities call fails. (For example because there is no connection or that the login/password are wrong)

kaustubh-welankar:
I'll fix this bug this week. You could assign it to me

ckamm:
I've tested this by trying to use owncloudcmd with an invalid password and checking the log output.

PR #5905, opened 2017-07-14
Fix for issue number 5692

kaustubh-welankar:
https://github.com/owncloud/client/issues/5692

### DIFF
--- a/src/cmd/cmd.cpp
+++ b/src/cmd/cmd.cpp
@@ -442,6 +442,11 @@ int main(int argc, char **argv)
     loop.exec();
 #endif
 
+    if (job->reply()->error() != QNetworkReply::NoError){
+        std::cout<<"Authentication Error\n";
+        return EXIT_FAILURE;
+    }
+
     // much lower age than the default since this utility is usually made to be run right after a change in the tests
     SyncEngine::minimumFileAgeForUpload = 0;
 
--- a/src/cmd/cmd.cpp
+++ b/src/cmd/cmd.cpp
@@ -425,7 +425,6 @@ int main(int argc, char **argv)
     account->setCredentials(cred);
     account->setSslErrorHandler(sslErrorHandler);
 
-#if QT_VERSION >= QT_VERSION_CHECK(5, 0, 0)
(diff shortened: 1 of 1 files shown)


kaustubh-welankar:
@SamuAlfageme Please take a look at this. Thanks !

src/cmd/cmd.cpp
```diff
@@ -442,6 +442,11 @@ int main(int argc, char **argv)
     loop.exec();
 #endif
```
ogoffart: Looks good.  Just a detail: you need to put the code before the #endif  even if it does not really matter since Qt 4 is no longer supported.  (But then we should remove the #if

src/cmd/cmd.cpp
```diff
@@ -442,6 +442,11 @@ int main(int argc, char **argv)
...
 
+    if (job->reply()->error() != QNetworkReply::NoError){
+        std::cout<<"Authentication Error\n";
```
ogoffart: Actually, that error message is not really accurate. The problem could be anything such as network down or server under maintainnence.

kaustubh-welankar: Should I change it to "Problem connecting to server"? Even if there is any other error , the next steps (database creation etc) shouldn't happen.

src/cmd/cmd.cpp
kaustubh-welankar: I've removed the Qt version check condition.

ogoffart:
Thanks for the patch, i merged it.
An improvement could be to show  the actual error string  (`job->reply()->errorString()`

kaustubh-welankar:
@ogoffart 
I see 49 occurences of QT_VERSION_CHECK(5, 0, 0) and around 100 of QT_VERSION_CHECK . I'll make a new pull request where I'll remove all the unnecessary version checks. I need some clarity on which Qt versions are supported for building the client. I cant find which Qt versions are supported in the building instructions. I'll also add the improvement you suggested.

ogoffart:
We require 5.6 minimum.

Issue #5848, opened 2017-06-20
[OAuth2] - Fallback options


SamuAlfageme:
Placeholder issue to discuss when to offer fallback options (if applicable) for the authentication step in case the OAuth2 workflow/app is not properly working. 

Continuing the discussion on https://github.com/owncloud/client/issues/5813

ogoffart:
So how do we proceed?
@michaelstingl  @pmaier1 

Currently, if the server has the oauth2 app enabled, oauth2 is the only authentication method allowed by the client.  The question is if we should still allow login/password to be entered?  This may be usefull if the oauth2 auth is not working for a reason or another. (Misconfigured browser, client_id/client_secret not valid, ...)

So questions:
 - Should we have a login/password fallback?
 - Should the fallback be always visible, or only if the authentication did not work after a timeout?
 - Should there be branding options to 
      - Disable oauth completely?
      - Disable the fallback?


If we decide to have a password fallback, here is my suggestion on how to do it:
In the UI, there are two workflow to log-in:

1. In the wizard when adding an account:

Currently, we hide the wizard and and show the browser.  If the user click on on the tray icon before logging in, the wizard is shown again and show a text telling to switch to the browser, with a button to re-open it.

![spectacle tj8774](https://user-images.githubusercontent.com/959326/28453027-99669ece-6df5-11e7-90ac-c9c34b528f88.png)


**Suggestion:**  Add a login/password  field on this screen so the user can connect with a password.

2. When we were logged out, and the user click login from the Account menu

Currently, we show a text in the header with "Obtaining authorization from the browser. Click here to re-open the browser"

![spectacle pw8455](https://user-images.githubusercontent.com/959326/28452817-b21b460a-6df4-11e7-98c6-21bf305ac50c.png)

**Suggestion:**  continue the sentence with ..." or click here to enter your credentials directly"

michaelstingl:
From my understanding, no fallback is needed. If there is something wrong with OAuth, the admin can always disable it on the server.

SamuAlfageme:
@ogoffart then we'll need a "application uninstall/disabling" path to switch back `http_oauth` accounts if we detect the server is no longer sending`WWW-Authenticate: Bearer realm="ownCloud"` headers.

ogoffart:
@SamuAlfageme that's done.  If the server disables oauth, the fallback is automatic.

SamuAlfageme:
[...]

PR #5912, opened 2017-07-18
OAuth: Fix refresh of token after expiration

ogoffart:
Before commit d3b00532b1dd9f44cc606e6738b53345c37582cf,
fetchFromKeychain was called everytime we detect that the creds are
invalid (in AccountState::slotInvalidCredentials)
But since that commit, AccountState was calling askFromUser directly,
breaking the refresh of the token.

So I made sure AccountState::slotInvalidCredentials still calls
refreshAccessToken.

Another change that was made was too be sure to clear the cookies
in HttpCredentials::invalidateToken even when we are only clearing the
access_token. That's because the session with a cookie may stay valid
longer than the access_token

### DIFF
--- a/src/gui/accountstate.cpp
+++ b/src/gui/accountstate.cpp
@@ -16,6 +16,7 @@
 #include "accountmanager.h"
 #include "account.h"
 #include "creds/abstractcredentials.h"
+#include "creds/httpcredentials.h"
 #include "logger.h"
 #include "configfile.h"
 
@@ -305,12 +306,17 @@ void AccountState::slotInvalidCredentials()
     qCInfo(lcAccountState) << "Invalid credentials for" << _account->url().toString()
                            << "asking user";
 
-    if (account()->credentials()->ready())
-        account()->credentials()->invalidateToken();
-    account()->credentials()->askFromUser();
-
-    setState(AskingCredentials);
     _waitingForNewCredentials = true;
+    setState(AskingCredentials);
+
+    if (account()->credentials()->ready()) {
+        account()->credentials()->invalidateToken();
+        if (auto creds = qobject_cast<HttpCredentials *>(account()->credentials())) {
+            if (creds->refreshAccessToken())
+                return;
+        }
+    }
+    account()->credentials()->askFromUser();
 }
 
 void AccountState::slotCredentialsFetched(AbstractCredentials *)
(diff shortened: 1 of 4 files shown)


ogoffart:
(I wish we had tests for this, but the AccountState is not tested :-( )

SamuAlfageme:
> That's because the session with a cookie may stay valid longer than the access_token 

Should not happen after https://github.com/owncloud/core/issues/28336 -> https://github.com/owncloud/oauth2/pull/53 but good point

ogoffart:
Updated this merge request with a commit to fix https://github.com/owncloud/client/issues/5848#issuecomment-317353049

src/gui/accountstate.cpp
```diff
@@ -305,12 +306,17 @@ void AccountState::slotInvalidCredentials()
...
 
-    if (account()->credentials()->ready())
+    _waitingForNewCredentials = true;
+    setState(AskingCredentials);
+
+    if (account()->credentials()->ready()) {
         account()->credentials()->invalidateToken();
+        if (auto creds = qobject_cast<HttpCredentials *>(account()->credentials())) {
+            if (creds->refreshAccessToken())
+                return;
```
jturcotte: This is a bit weird, it breaks encapsulation.
I know how much HttpCredential is spaghetti, but did you have any other solution in mind that could avoid this?

Maybe something like moving `refreshAccessToken` down to `AbstractCredentials`, which would fall-back to call `askFromUser()` itself.

src/libsync/creds/httpcredentials.cpp
```diff
@@ -397,6 +397,7 @@ void HttpCredentials::persist()
 
     _account->setCredentialSetting(QLatin1String(userC), _user);
     _account->setCredentialSetting(QLatin1String(isOAuthC), isUsingOAuth());
+    _account->wantsAccountSaved(_account);
```
jturcotte: Maybe `ShibbolethCredentials::persist` would need the same line?

src/gui/accountstate.cpp
ogoffart: Yes, i know, it breaks the abstraction layer. 
I tought about putting refreshAccessToken in AbstractCredentials.  But i think in the long term we could get rid of it and only have HttpCredentials.
If you insist i can do it.

(Note: there are other places where i broke this abstraction layer)

jturcotte: Not sure how long that will be, since all customer servers will need to move to 10.1 before we can get rid of shibboleth, but up to you.

PR #5916, opened 2017-07-20
Splitt off Abstract Folder class from Folder

dragotin:
Please consider this (large) patch. I splits off an abstract class for folders, providing the integration with the application by managing the interfaces with GUI, socketapi, scheduling and such. The folder class now only still contains things specific to this (the old, common) type  of sync folder. 

However, with the new structure we will be able to develop other folder types, such as folders that are actually containers containing a file folder structure, which require special handling in the client code.

A long term goal is that folders can even have different sync engines.

I am looking forward to hearing your comments.

### DIFF
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -48,6 +48,8 @@ set(client_SRCS
     accountsettings.cpp
     application.cpp
     folder.cpp
+    abstractfolder.cpp
+    folderdefinition.cpp
     folderman.cpp
     folderstatusmodel.cpp
     folderstatusdelegate.cpp
--- a/src/gui/abstractfolder.cpp
+++ b/src/gui/abstractfolder.cpp
@@ -1,7 +1,6 @@
 /*
- * Copyright (C) by Duncan Mac-Vicar P. <duncan@kde.org>
  * Copyright (C) by Daniel Molkentin <danimo@owncloud.com>
- * Copyright (C) by Klaas Freitag <freitag@owncloud.com>
+ * Copyright (C) by Klaas Freitag <freitag@owncloud.org>
  *
  * This program is free software; you can redistribute it and/or modify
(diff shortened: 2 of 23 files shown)


dragotin:
Hmm, I forgot to rebase the WIP commit away. 😠

dragotin:
Rebased and commits cleaned, as much as doable. Please let me know what you think.

ogoffart:
Hi @dragotin,

Can you elaborate on the use case for this change?

I'm just worried that adding an abstraction layer that we don't need makes things more complicated.

guruz:
We have discussions here internally in Berlin about if this is really the way to go for your user case :-/

guruz:
(Depends also on how the rest of your PR looks like.. because this one is still OK, but how much would the rest influence the code base?)

dragotin:
I am not really seeing how this particular PR makes things more complicated. I think it is already now a better encapsulation of functions in their classes.

To implement a nice integration of things like Veracrypt, I think it will be neccessary to have a specialized sync engine, and for that, the current SyncEngine class would also have to be split to a SyncEngineBase and the current implementation.

I think that is not bad, because the "old" sync engine code stays nearly untouched while implementing the new functionality, and thus breakage risk is smaller.

Obviously I am happy to explain and evaluate my idea more in detail, and hope to meet you guys on the oCC.

ogoffart:
Currently this abstraction serves no use. And in that case the abstraction add complexity because it force to cross the abstraction layer when we want to add things. For example, you still in your patch need to do many cast to Folder, showing that the abstraction layer is leaky.

You don't need a new Folder class to change the SyncEngine.

I rather not add an abstraction layer we don't need. Otherwise it gets in the way of future changes.

guruz:
We've discussed this at the oC conference and I think we all agree that we go another route now
(e.g. complaining in wizard if user accidently picks a dir inside container instead of container itself).
Can we close this @dragotin ? :)

Issue #4556, opened 2016-03-09
File Versions on the Client: Use the Server API


dragotin:
owncloud/core#10571 talks about an server API that provides information about the previous versions of a file.  

The client should provide a client native dialog that shows the previous versions and allows the user to revert the file back to one of the previous verisons.

The versions dialog 
- should be invoked by a context menu entry in the file manager
- should show a list of available previous versions
- should show a thumbnail with the previous versions 
- allow the user to revert to a previous one and manage the sync down of the file.

@MTRichards @cmonteroluque something to address with server version 9.1?

MTRichards:
Very similar to something I have been looking at - and that was requested for the future. The idea that, on the desktop, I could right click on a file and get a previous version list, and roll back to that version from the desktop.

It is not common as an ask, but I do think NetAPP, for example, has a feature like this on the network drive mounts for desktop users. Something to consider longer term.

butonic:
NetApp seems to implement a Volume Copy Service Provider: https://msdn.microsoft.com/en-us/library/windows/desktop/aa381601(v=vs.85).aspx

The result is a native integration with the windows file property dialog:

![image001 1](https://cloud.githubusercontent.com/assets/956847/13914877/936ec74a-ef51-11e5-97a3-822c04275840.png)

![image002](https://cloud.githubusercontent.com/assets/956847/13914884/99a1d42c-ef51-11e5-8f9a-5f41a01710b7.png)

@dragotin I have not looked at the Windows Interfaces. Do you think this is possible?

00004796

dragotin:
@butonic everything is possible.

butonic:
@dragotin good, how much effort do you think it is?

dragotin:
lol @butonic lets talk once we have a server api.

dragotin:
@butonic ah, you probably mean how much effort it is to add that tab to the properties dialog?! No idea, honestly, needs investigation. But stuff like that is usually doable as you find examples in the net. An issue is to build that in our infra, as we use cross compiling, and these things often require native building.

butonic:
@MTRichards schedule for 9.2?

MTRichards:
Love the idea, we need to add to the versions API to make it happen, then possible. Not going to make 9.1 right now, that is largely set with L1 and L2 and this didn't make it high enough on the priority list.
https://github.com/owncloud/enterprise/issues/1205

SamuAlfageme:
[...]

Issue #3096, opened 2015-04-10
Reset Client Data to Server State


dragotin:
### Usecase:

A user should have the opportunity to reset his sync client data to exactly the same state the server is in. Obviously that should be the  fact anyway but reality shows that users sometimes do not feel confident that their sync data is *really* in sync and want to go back to the servers state. Mainly this is about clearing all errors that users might see. Not needed downloads should be avoided.

### Implementation Options:
1. Remove the local sync journal. That results in recreation of the journal with the superset of the files. It would not remove files that exist on the client but not on the server.
2. Change the reconcile phase to have a "Server wins" option. That would basically create a propagator list that does not propagate changes from client to server. 

@ogoffart what is your opinion?

@msrex @MTRichards please discuss scheduling of this feature.

ogoffart:
> 1. Remove the local sync journal. 

That would not help at all, files with error would still be there and the client would still try to upload them on next sync.

> 2. Change the reconcile phase to have a "Server wins" option

Yes. That would be a totally new reconcile logic that would remove the files that do not exist on the server, and always download the files that are different.

Question is how do we show that to the user? This is also a destructive operation that can lead to data losses. 
This is also not a trivial change to do in the sync engine.

guruz:
Yep reminds me what I mentioned on IRC some days ago, I accidently triggered a big file upload of an existing file. I wanted to pause the sync and then 'revert to server'

PR #5921, opened 2017-07-25
view and restore version of a file

heavymanto:
Hi, i like your work and love your project. I was working on a file for a long time, unfortunately at some point, it was unusable. I would have been very helpful in having copies of earlier versions but unfortunately this functionality is missing. Can you consider the idea of developing a file update history feature? Of course there are many issues open, disk quotas will be affected by the space occupied by the various versions, but I think a user would like to be able to configure whether to activate the history and how many copies of the file to keep in order to best manage their space / security coefficient .

Greetings
Francis


guruz:
-> https://github.com/owncloud/client/issues/4556 https://github.com/owncloud/client/issues/3096  :-)

Issue #296, opened 2013-01-30
[OS X] Add link to sync folder with ownCloud icon in Finder Sidebar


rgr101:
Just a nice-to-have with no priority: 

Have an Owncloud icon instead of the ordinary folder icon in Finder's Sidebar
(just like Dropbox does).

jturcotte:
I looked at how to do this in relation with https://github.com/owncloud/client/issues/2340, there is an official way documented within the FinderSync doc in the "Adding a Sidebar icon":
https://developer.apple.com/library/mac/documentation/General/Conceptual/ExtensibilityPG/Finder.html#//apple_ref/doc/uid/TP40014214-CH15

This seems to rely on the FinderSync watched folder list (I didn't try yet) but fetches the icon from the wrapping application's own CFBundleIconFile .icns file. We'd need to replace the manual kdelibs logic in cmake/modules/AddAppIconMacro.cmake with an .iconset folder (containing extra sidebar_123x123.png files) and generate the .icns using iconutils instead.

The main problem I see with this though is that I'd have to change the structure of all the branding repository which fetches those files or create this .iconset directory on the fly in the build directory. I'm not sure yet if that amount of work is worth showing a fancier icon in the Finder sidebar, what do you think?

rgr101:
I strongly believe this is an important part of visibility and brand generation, and so it'll  be worth the efforts by sure...

jorgelbg:
This little features creates a lot of user value (from my point of
view) mainly due the fact that  increases the integration between the
application and the user environment.

guruz:
@MTRichards @jancborchardt Please also discuss with regards to multi-account or multiple sync folders.
And see @jturcotte 's comment above.

jancborchardt:
For multi-account, I’d say every local folder should have an entry in the sidebar. For multiple sync folders, not sure.

MTRichards:
>For multi-account, I’d say every local folder should have an entry in the sidebar.

Yes, each root folder should get a shortcut on the finder menu, and in windows in explorer.

For multi account, the rule remains that root sync folders get left menu icons - but only one. So if a root folder is synced in two accounts, only 1 sidebar as it is the same folder.
If the root and a sub folder are synced to two different accounts, then that is like having 2 root folders, so two icons on the left.

@jancborchardt what do you think?

jancborchardt:
@MTRichards sounds reasonable, yes.

MTRichards:
[...]

Issue #455, opened 2013-03-18
File Managers: ownCloud folder icon


jancborchardt:
(Similar to #296 but not the same)

The icon of the ownCloud folder (currently clientsync) is just a regular folder. Would be cool to have that be the ownCloud logo to be able to quickly identify it.

MattiJoh:
The example of icon: http://sc20k1.deviantart.com/art/ownCloud-Icon-492615985
![owncloud](https://cloud.githubusercontent.com/assets/8496953/5812785/74378f9a-a026-11e4-9433-00d86ab8b380.png)

bezin:
MattiJoh

danimo:
@jturcotte Is there anything you need from the art department in order to get this forward?

jancborchardt:
@jturcotte if there is, best shoot a mail to Achim. :)

g4b0:
+1 It would be nice to have a custom icon instead of the flat one

bitbonk:
Please note that under OSX we'd also need an icon for the sidebar in finder.
![osx finder sidebar](https://cloud.githubusercontent.com/assets/418030/14474518/d1487d64-00fd-11e6-8844-9cde072c82f8.png)

antoinehenrich:
+1

guruz:
Just leaving this link here: Windows - Integrate a Cloud Storage Provider

https://msdn.microsoft.com/en-us/en-en/library/windows/desktop/dn889934(v=vs.85).aspx

MegaV0lt:
Propably also related the Nextcloud Icon i made for Windows 7. See there.
https://help.nextcloud.com/t/folder-icon-for-desktop-client-windows7/6188

guruz:
(At least currently the code to create a sidebar favorite seems not to be called at all because it is in a not possible anymore ```else``` branch.. check the wizard for ```Utility::setupFavLink( localFolder )```)

PR #5928, opened 2017-07-31
macos: Add a sidebar icon

jturcotte:
In the process, use an iconset to generate the icns using iconutil.
Also add some missing icon resolutions according to the guidelines.

Issue #296


SamuAlfageme:
@jturcotte this is dope! 😎 

Will try it later, can you post a screenshot to see how does it look for reference?

jturcotte:
I just picked the same icon we have and put it there, and this is the result:

screen shot 2017-07-31 at 11 22 01

It would be better if some designer would do it but I think it's not too bad already.

src/gui/CMakeLists.txt
```diff
@@ -232,7 +232,7 @@ if (NOT DEFINED APPLICATION_ICON_NAME)
     set(APPLICATION_ICON_NAME ${APPLICATION_SHORTNAME})
 endif()
 
-kde4_add_app_icon( ownCloud "${theme_dir}/colored/${APPLICATION_ICON_NAME}-icon*.png")
+kde4_add_app_icon( ownCloud "${theme_dir}/colored/${APPLICATION_ICON_NAME}-*.png")
```
guruz: The files in theme/colored still have the "-icon" in their name, any reason to change this ```kde4_add_app_icon``` thingy?

SamuAlfageme:
@guruz @jturcotte can any of you guys change the PR's target so I can build the installers with rotor (remember the limitations to build fork's PR with it 😕) while keeping master clean 'till I do some good ol' testin'?

guruz:
ACK, @jturcotte please push to this repo, not your own

jturcotte:
I asked for **code review**, the CI reported "All checks have passed", this is not a whole QA process. We do not ask this from all contributors.

@guruz You don't even ask for code reviews and push your stuff directly to master yourself, what is your point?

jturcotte:
I've been under the impression that QA is happening in master, and I think that it's not worth doing custom rotor builds on feature branches to test once, and then test again for any possible merge issue in master, but this is for the QA team to decide. It seems to be the consensus for that reason that the team should make pull request from the main repo only and I'll do it if you really think that this kind of consistency is worth the [proliferating amount of stale branches](https://github.com/owncloud/client/branches/stale), the main reason why I've been avoiding it personally.

I can't retarget the pull request, so for this time I think it should be fine.

I tested the build on my machine and I'm pretty confident it's going to pass on rotor as well. Even if I does break the build, that I add a commit here or on master afterwards won't change anything as long as I push the fix shortly after the merge.

guruz:
@jturcotte i've deleted like 10-15 branches now from that list

guruz:
@SamuAlfageme Build created https://download.owncloud.com/desktop/daily/ownCloud-2.4.0.8114-nightly20170731.pkg

src/gui/CMakeLists.txt
guruz: ^^

guruz:
@jturcotte great works, it works for me on 10.12.5
However I need to move the Folder manually to sidebar, I'll work on a patch to change this.
https://github.com/owncloud/client/issues/455#issuecomment-291885831

src/gui/CMakeLists.txt
jturcotte: Yes that's to be able to also get files like `owncloud-sidebar-36.png` in the glob. `-icon` is for the application icon.

guruz: Aaah it's because of ```macro (KDE4_ADD_APP_ICON appsources pattern)``` and has nothing to do with kd4...

SamuAlfageme:
@jturcotte is there a way to override the icon to be the default's (i.e. the old one) in case of a branding client with no alternative choice?

guruz:
I think the default won't copy the icon into the bundle if it is not found in the theme. Then the OS will behave as currently

jturcotte:
@michaelstingl Is there any theming documentation I could update to list needed and optional icon sizes with their name?

SamuAlfageme:
@jturcotte we can probably ease @jvillafanez's job updating ownBrander by posting it on https://github.com/owncloud/ownbrander/issues/748 - I think there's no theming docs available

SamuAlfageme:
@jturcotte also, would it be hard to consider a migration path for existing local sync folders to be "updated" in the finder sidebar with the new icon when upgrading the client?

guruz:
> a migration path for existing local sync folders

I'm working on that, there is a bug -> https://github.com/owncloud/client/pull/5928#issuecomment-319307968

jancborchardt:
So cool, nice work @jturcotte!

guruz:
See the linked PR which is related for depending on which wizard you use.

There is no need for a migration path, the Finder will automatically pick up the sidebar icon :)

PR #5929, opened 2017-07-31
Clean up the client docs removing old, outdated, content

settermjd:
This makes changes as suggested by @michaelstingl to clean out old content in the docs, in
https://github.com/owncloud/documentation/issues/3240.


hefee:
why tare you deleting the manpages doc/owncloud.1.rst and doc/owncloudcmd.1.rst?

guruz:
> why are you deleting the manpages doc/owncloud.1.rst and doc/owncloudcmd.1.rst?

Because we don't want to maintain duplicate information. there is already ```--help```

hefee:
But for Debian we need to provide a man page.

guruz:
> But for Debian we need to provide a man page.

Are you saying this is 100% mandatory?

And why is the URL in https://twitter.com/debiannew/status/906561649733193729 404 ? :-(

(CC @jnweiger )

hefee:
> And why is the URL in https://twitter.com/debiannew/status/906561649733193729 404 ? :-(

well the NEW queue is only a temporaly location. In meanwhile it entered unstable. The best way to see the overview is the tracker:
https://tracker.debian.org/pkg/owncloud-client

hefee:
> Are you saying this is 100% mandatory?
no not mandatory, but a man page is more than just the "--help" in owncloudcmd there are also examples etc. But if they are outdated and no one caring about it...

michaelstingl:
`client/doc/owncloudcmd.1.rst` hasn't been touched since January 2016:
https://github.com/owncloud/client/commits/51a70cee3d08d477d9a724bd93ccdfd37fb874f6/doc/owncloudcmd.1.rst

`client/doc/owncloudcmd.rst` gets maintained every few months:
https://github.com/owncloud/client/commits/master/doc/owncloudcmd.rst

Sorry, I wasn't aware about the meaning of this file. Name `owncloudcmd.1.rst` sounds too generic for me, also there is no comment in the top of the file that says what is the purpose. I would definitely prefer a way to not maintain duplicate files as @jnweiger suggested in https://github.com/owncloud/client/issues/6011#issuecomment-328906597 .  I also don't like the fact, that content of `owncloudcmd.1.rst`shows up in the search results: https://doc.owncloud.org/desktop/latest/search.html?q=owncloudurl

`client/doc/owncloud.1.rst` hasn't been touched since December 2014:
https://github.com/owncloud/client/commits/2ee287685582680f5eb174203959c4275dad0912/doc/owncloud.1.rst

Issue #5922, opened 2017-07-25
[Notifications] Make close/dismiss button in the Activity tab match server's logic


SamuAlfageme:
I just checked how's the implementation to close notifications is in the client. While doing this on server:

<p align="center">
  <img src="https://user-images.githubusercontent.com/2644445/28580635-63faf2c6-7160-11e7-916e-3509c8000dc5.gif"/>
</p>

... is the equivalent of removing the notification / dismiss **forever**:

```
curl 'https://<server>/ocs/v2.php/apps/notifications/api/v1/notifications/5?format=json' \
  -X DELETE \
  [...]
  -H 'OCS-APIREQUEST: true' \
```

Closing it on the client's Activity tab:

<p align="center">
  <img src="https://user-images.githubusercontent.com/2644445/28580935-30355840-7161-11e7-9693-73bf2e676fbe.gif"/>
</p>

**Does not cause any effects on the server** (see https://github.com/owncloud/client/blob/master/src/gui/notificationwidget.cpp#L100-L108) ; meaning on next client startup, those notifications will be displayed again.

I think these should be adjusted to match the ❌ (close) logic in the server.

ref. https://github.com/owncloud/enterprise/issues/2178

guruz:
They even re-appear while the client is running sometimes.

CC @dragotin

guruz:
The code was easy but it does not work :-(

We expect:
```// the ocs API returns stat code 100 if it succeeded.```

But calling ```DELETE``` on the notification as in returns a HTTP 200
https://github.com/owncloud/notifications/blob/master/docs/ocs-endpoint-v1.md#deleting-a-notification-for-a-user

I'm not sure I understand this, why is 100 expected?
```#define OCS_SUCCESS_STATUS_CODE 100```
HTTP 100 is "Continue", it's not a confirmation code.

So this is a server bug that we could work around by just accepting 200, but...

SamuAlfageme:
👍

PR #5936, opened 2017-08-03
Notifications: Propagate "Dismiss" as DELETE to server #5922

guruz:


guruz:
Thanks @DeepDiver1975 for clarifying that code 200 is also OK

src/gui/ocsjob.h
```diff
@@ -24,6 +24,8 @@
 #include <QUrl>
 
 #define OCS_SUCCESS_STATUS_CODE 100
+// Apparantly the v2.php URLs can return that
```
ogoffart: That comment sounds a bit unsure.

dragotin:
You guys are soo brave changing the return code to ..... something standard 😀

PR #5937, opened 2017-08-03
Sync Folder: Add file manager favorite also in folder wizard #455

guruz:
Not only in the account wizard.


guruz:
As per the (wrongly worded) comment in https://github.com/owncloud/client/issues/455#issuecomment-291885831

jturcotte:
How does this behave on different platforms if the favorite is already there?

guruz:
@jturcotte This is anyway run only in the add sync folder wizard. I don't know.

guruz:
Just verified, it doesnt duplicate them

Issue #5949, opened 2017-08-10
Resuming a paused sync that was uploading a bunch of files and were moved in the client is duplicating one file in both locations


SamuAlfageme:
### Expected behavior
The reconcile phase should match the server and local contents properly and move the last file the client uploaded to the server.

### Actual behavior
That last file is being downloaded from the server:

![screenshot_2017-08-10_10_14_56](https://user-images.githubusercontent.com/2644445/29161398-5b21160a-7db6-11e7-8c32-eb6c6a58d887.png)

I think it has to do with pausing the sync either:
- **...before receiving the upload ACK reply from the server with the ETAG** (we should wait for it in this case)
- **...and not storing the last ETAG properly on the sync journal** 

### Steps to reproduce
1. Create a bunch of large files (I did `generate_n_files.sh 130 file_ 15M`: 130 files , 15MB size with the `file_` prefix)
2. Pause the sync; you'll be prompted with the 'sync will be terminated' confirmation dialog 
3. Move all the files to a new location (`mkdir filezz && mv file_* filezz`)
4. Resume the sync

![screenshot_2017-08-10_10_22_07](https://user-images.githubusercontent.com/2644445/29161438-7e315024-7db6-11e7-9e2e-b53c3cbb9ca2.png)

### Server configuration
* **Operating system**: ubuntu 
* **Web server**: Apache/2.4.18 (Ubuntu)
* **Database**: 5.7.17-0ubuntu0.16.04.1 - (Ubuntu)
* **PHP version**: 7.0.15-0ubuntu0.16.04.4
* **ownCloud version**: 10.0.2

### Client configuration
Client version: **2.3.3rc1**

### Logs

1. Client logfile: I pasted just the related entries (i.e. `grep file_114` in this case) - https://gist.github.com/SamuAlfageme/b9e6a34e7fcd3963c7bb6a353995668f

I also have the recorded mitmproxy session recorded, will upload in a minute (after removing sensitive info)

cc/ @ogoffart @guruz @mrow4a

ogoffart:
Here is what happens:
The PUT is cancelled as we already sent data, but we did not receive the answer yet with the new etag, so we don't write it in the DB.

The solution could be to change what "pause sync" is doing. We should not blindly abort all requests. We should wait for the PUT to finish (possibly with a lower timeout)

mrow4a:
@ogoffart :+1

guruz:
[...]

PR #5953, opened 2017-08-10
Fix paused sync file move issue #5949

mrow4a:
@SamuAlfageme can you verify that this fixes issue for ChunkingNG for you? I checked on my local machine and this fixed it.

@guruz I will also add similar thing, but for V1, however there I will check if the file bytes been sent or not. 

Issue #5949

### DIFF
--- a/src/libsync/owncloudpropagator.cpp
+++ b/src/libsync/owncloudpropagator.cpp
@@ -742,6 +742,16 @@ PropagatorJob::JobParallelism PropagatorCompositeJob::parallelism()
     return FullParallelism;
 }
 
+void PropagatorCompositeJob::slotSubJobAbortFinished()
+{
+    // Count that job has been finished
+    _abortsCount--;
+
+    // Emit abort if last job has been aborted
+    if (_abortsCount == 0) {
+        emit abortFinished();
+    }
+}
 
 bool PropagatorCompositeJob::scheduleSelfOrChild()
 {
@@ -900,7 +910,8 @@ void PropagateDirectory::slotFirstJobFinished(SyncFileItem::Status status)
 
     if (status != SyncFileItem::Success && status != SyncFileItem::Restoration) {
         if (_state != Finished) {
-            abort();
+            // Synchronously abort
+            abort(AbortType::Synchronous);
             _state = Finished;
             emit finished(status);
         }
--- a/src/libsync/owncloudpropagator.h
+++ b/src/libsync/owncloudpropagator.h
@@ -65,6 +65,11 @@ class PropagatorJob : public QObject
 public:
     explicit PropagatorJob(OwncloudPropagator *propagator);
 
+    enum AbortType {
+        Synchronous,
+        Asynchronous
+    };
+
     enum JobState {
         NotYetStarted,
         Running,
@@ -98,7 +103,14 @@ public:
     virtual qint64 committedDiskSpace() const { return 0; }
 
 public slots:
-    virtual void abort() {}
+    /*
+     * Asynchronous abort requires emit of abortFinished() signal,
+     * while synchronous is expected to abort immedietaly.
+    */
+    virtual void abort(PropagatorJob::AbortType abortType) {
+        if (abortType == AbortType::Asynchronous)
+            emit abortFinished();
+    }
 
     /** Starts this job, or a new subjob
      * returns true if a job was started.
@@ -110,10 +122,13 @@ signals:
      */
     void finished(SyncFileItem::Status);
 
+    /**
+     * Emitted when the abort is fully finished
+     */
+    void abortFinished(SyncFileItem::Status status = SyncFileItem::NormalError);
 protected:
     OwncloudPropagator *propagator() const;
 };
-
 
 /*
  * Abstract class to propagate a single item
@@ -185,10 +200,11 @@ public:
     SyncFileItemVector _tasksToDo;
     QVector<PropagatorJob *> _runningJobs;
     SyncFileItem::Status _hasError; // NoStatus,  or NormalError / SoftError if there was an error
+    quint64 _abortsCount;
 
     explicit PropagatorCompositeJob(OwncloudPropagator *propagator)
         : PropagatorJob(propagator)
-        , _hasError(SyncFileItem::NoStatus)
+        , _hasError(SyncFileItem::NoStatus), _abortsCount(0)
     {
     }
 
@@ -209,15 +225,32 @@ public:
 
     virtual bool scheduleSelfOrChild() Q_DECL_OVERRIDE;
     virtual JobParallelism parallelism() Q_DECL_OVERRIDE;
-    virtual void abort() Q_DECL_OVERRIDE
+
+    /*
+     * Abort synchronously or asynchronously - some jobs
+     * require to be finished without immediete abort (abort on job might
+     * cause conflicts/duplicated files - owncloud/client/issues/5949)
+     */
+    virtual void abort(PropagatorJob::AbortType abortType) Q_DECL_OVERRIDE
     {
-        foreach (PropagatorJob *j, _runningJobs)
-            j->abort();
+        if (!_runningJobs.empty()) {
+            _abortsCount = _runningJobs.size();
+            foreach (PropagatorJob *j, _runningJobs) {
+                if (abortType == AbortType::Asynchronous) {
+                    connect(j, &PropagatorJob::abortFinished,
+                            this, &PropagatorCompositeJob::slotSubJobAbortFinished);
+                }
+                j->abort(abortType);
+            }
+        } else if (abortType == AbortType::Asynchronous){
+            emit abortFinished();
+        }
     }
 
     qint64 committedDiskSpace() const Q_DECL_OVERRIDE;
 
 private slots:
+    void slotSubJobAbortFinished();
     bool possiblyRunNextJob(PropagatorJob *next)
     {
         if (next->_state == NotYetStarted) {
@@ -258,11 +291,17 @@ public:
 
     virtual bool scheduleSelfOrChild() Q_DECL_OVERRIDE;
     virtual JobParallelism parallelism() Q_DECL_OVERRIDE;
-    virtual void abort() Q_DECL_OVERRIDE
+    virtual void abort(PropagatorJob::AbortType abortType) Q_DECL_OVERRIDE
     {
         if (_firstJob)
-            _firstJob->abort();
-        _subJobs.abort();
+            // Force first job to abort synchronously
+            // even if caller allows async abort (asyncAbort)
+            _firstJob->abort(AbortType::Synchronous);
+
+        if (abortType == AbortType::Asynchronous){
+            connect(&_subJobs, &PropagatorCompositeJob::abortFinished, this, &PropagateDirectory::abortFinished);
+        }
+        _subJobs.abort(abortType);
     }
 
     void increaseAffectedCount()
@@ -280,6 +319,7 @@ private slots:
 
     void slotFirstJobFinished(SyncFileItem::Status status);
     void slotSubJobsFinished(SyncFileItem::Status status);
+
 };
 
 
@@ -324,6 +364,7 @@ public:
         , _chunkSize(10 * 1000 * 1000) // 10 MB, overridden in setSyncOptions
         , _account(account)
     {
+        qRegisterMetaType<PropagatorJob::AbortType>("PropagatorJob::AbortType");
     }
 
     ~OwncloudPropagator();
@@ -406,11 +447,19 @@ public:
     {
         _abortRequested.fetchAndStoreOrdered(true);
         if (_rootJob) {
-            // We're possibly already in an item's finished stack
-            QMetaObject::invokeMethod(_rootJob.data(), "abort", Qt::QueuedConnection);
+            // Connect to abortFinished  which signals that abort has been asynchronously finished
+            connect(_rootJob.data(), &PropagateDirectory::abortFinished, this, &OwncloudPropagator::emitFinished);
+
+            // Use Queued Connection because we're possibly already in an item's finished stack
+            QMetaObject::invokeMethod(_rootJob.data(), "abort", Qt::QueuedConnection,
+                                      Q_ARG(PropagatorJob::AbortType, PropagatorJob::AbortType::Asynchronous));
+
+            // Give asynchronous abort 5000 msec to finish on its own
+            QTimer::singleShot(5000, this, SLOT(abortTimeout()));
+        } else {
+            // No root job, call emitFinished
+            emitFinished(SyncFileItem::NormalError);
         }
-        // abort() of all jobs will likely have already resulted in finished being emitted, but just in case.
-        QMetaObject::invokeMethod(this, "emitFinished", Qt::QueuedConnection, Q_ARG(SyncFileItem::Status, SyncFileItem::NormalError));
     }
 
     // timeout in seconds
@@ -430,6 +479,13 @@ public:
     DiskSpaceResult diskSpaceCheck() const;
 
 private slots:
+
+    void abortTimeout()
+    {
+        // Abort synchronously and finish
+        _rootJob.data()->abort(PropagatorJob::AbortType::Synchronous);
+        emitFinished(SyncFileItem::NormalError);
+    }
 
     /** Emit the finished signal and make sure it is only emitted once */
     void emitFinished(SyncFileItem::Status status)
--- a/src/libsync/owncloudpropagator.h
+++ b/src/libsync/owncloudpropagator.h
@@ -445,7 +445,9 @@ public:
 
     void abort()
     {
-        _abortRequested.fetchAndStoreOrdered(true);
+        bool alreadyAborting = _abortRequested.fetchAndStoreOrdered(true);
+        if (alreadyAborting)
+            return;
         if (_rootJob) {
             // Connect to abortFinished  which signals that abort has been asynchronously finished
             connect(_rootJob.data(), &PropagateDirectory::abortFinished, this, &OwncloudPropagator::emitFinished);
--- a/src/libsync/propagateupload.h
+++ b/src/libsync/propagateupload.h
@@ -301,15 +301,15 @@ private:
      * In the non-resuming case it is 0.
      * If we are resuming, this is the first chunk we need to send
      */
-    int _startChunk;
+    int _startChunk = 0;
     /**
      * This is the next chunk that we need to send. Starting from 0 even if _startChunk != 0
      * (In other words,  _startChunk + _currentChunk is really the number of the chunk we need to send next)
      * (In other words, _currentChunk is the number of the chunk that we already sent or started sending)
      */
-    int _currentChunk;
-    int _chunkCount; /// Total number of chunks for this file
-    int _transferId; /// transfer id (part of the url)
+    int _currentChunk = 0;
+    int _chunkCount = 0; /// Total number of chunks for this file
+    int _transferId = 0; /// transfer id (part of the url)
 
     quint64 chunkSize() const {
         // Old chunking does not use dynamic chunking algorithm, and does not adjusts the chunk size respectively,
@@ -342,11 +342,11 @@ class PropagateUploadFileNG : public PropagateUploadFileCommon
 {
     Q_OBJECT
 private:
-    quint64 _sent; /// amount of data (bytes) that was already sent
-    uint _transferId; /// transfer id (part of the url)
-    int _currentChunk; /// Id of the next chunk that will be sent
-    quint64 _currentChunkSize; /// current chunk size
-    bool _removeJobError; /// If not null, there was an error removing the job
+    quint64 _sent = 0; /// amount of data (bytes) that was already sent
+    uint _transferId = 0; /// transfer id (part of the url)
+    int _currentChunk = 0; /// Id of the next chunk that will be sent
+    quint64 _currentChunkSize = 0; /// current chunk size
+    bool _removeJobError = false; /// If not null, there was an error removing the job
 
     // Map chunk number with its size  from the PROPFIND on resume.
     // (Only used from slotPropfindIterate/slotPropfindFinished because the LsColJob use signals to report data.)
@@ -366,7 +366,6 @@ private:
 public:
     PropagateUploadFileNG(OwncloudPropagator *propagator, const SyncFileItemPtr &item)
         : PropagateUploadFileCommon(propagator, item)
-        , _currentChunkSize(0)
     {
     }
 
--- a/src/libsync/propagateupload.h
+++ b/src/libsync/propagateupload.h
@@ -129,6 +129,11 @@ public:
         return true;
     }
 
+    QIODevice *device()
+    {
+        return _device;
+    }
+
     QString errorString()
     {
         return _errorString.isEmpty() ? AbstractNetworkJob::errorString() : _errorString;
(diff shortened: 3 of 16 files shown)


src/libsync/owncloudpropagator.h
```diff
@@ -214,6 +214,16 @@ class PropagatorCompositeJob : public PropagatorJob
...
             j->abort();
+
+        // Give running jobs max 5 seconds to finish on their own, otherwise
+        // continue (that might result in conflicts owncloud/client/issues/5949)
+        int retriesNo = 5;
+        for(int i = 0; i < retriesNo ; i++) {
+            if (_runningJobs.size() == 0) {
+                break;
+            }
+            Utility::sleep(1);
```
mrow4a: This is required, because if composite job gets aborted, but one of the job isnt, it will not finalize adding file to journal

src/libsync/propagateuploadng.cpp
```diff
@@ -502,4 +502,16 @@ void PropagateUploadFileNG::slotUploadProgress(qint64 sent, qint64 total)
...
     propagator()->reportProgress(*_item, _sent + sent - total);
 }
+
+
+void PropagateUploadFileNG::abort()
+{
+    foreach (AbstractNetworkJob *job, _jobs) {
+        // Abort only PUT and MKDIR jobs, since abording MoveJob
+        // might result in conflict.
+        if (job->reply() && !job->inherits("OCC::MoveJob")) {
```
mrow4a: if it is not MoveJob, just exit. (Wont it result in tmp new chunking directory be left without deletion, is server deleting periodicaly these tmp directories? @DeepDiver1975 )

src/libsync/owncloudpropagator.h
ogoffart: but that's not acceptable. You can't let the main thread sleep like this.
You need to find a better way to wait that the job is actually finished (by listening to the finished signal or so)

src/libsync/propagateuploadng.cpp
ogoffart: nitpick: use qobject_cast   (more efficient and also compile-time checked)

Also, we shold probzbly decrease the timeout.  For example, if the abort happens because the connection  was detected to be dropped, we should not wait a full HTTP timeout again, IMHO.

ogoffart: And to answer your question about stale directory, the next sync will resume the upload, so it won't be stale.

guruz:
> Also the chunkingv1 need to be changed not to abort the last job. This is also what is used when the file is too small to be split in chunks.

I think it's better to hook into how UploadDevice is used and from there control an abort. Then you for sure know you didn't send all bytes and can safely abort.

src/libsync/owncloudpropagator.h
mrow4a: Ohh, true, just tested and it stops the main thread, yep will do that with signals

src/libsync/propagateuploadng.cpp
mrow4a: Changed to qobject_cast.

mrow4a:
@ogoffart can you have a look on this approach? One disadvantage is that now after paused, MOVE job continues (1st stage):
selection_162

after it is finished, view switches to (2nd stage):
selection_163

One thing is that the button resume sync is still available also in "1st stage", not sure what happens if it still processes MOVE, but someone clicks resume in that time. 

@SamuAlfageme I did manual testing and for ChunkingNG problem is solved.

src/libsync/owncloudpropagator.h
```diff
@@ -227,8 +240,20 @@ private slots:
         return next->scheduleSelfOrChild();
     }
 
+    void terminateJobs() {
+        // Send abort to the subjobs and emit abort finished signal
+        abortJobs();
```
ogoffart: That's not needed, abort was already sent

So you don't need to put abortJobs in a different function

src/libsync/owncloudpropagator.cpp
```diff
@@ -939,6 +940,11 @@ void PropagateDirectory::slotSubJobsFinished(SyncFileItem::Status status)
     emit finished(status);
 }
 
+void PropagateDirectory::slotAbortFinished()
+{
+    emit abortFinished();
```
ogoffart: You need to count the jobs and only emit the signal when ALL subjobs have finished their abort.

src/libsync/owncloudpropagator.h
mrow4a: The idea here was to force the abort. Thus if after 5 seconds they did not finish, you kill the job just by sending abort jobs again, please have a look on chunkingng implementation

src/libsync/owncloudpropagator.cpp
ogoffart: Ah no.  i got confused with CompositeJob.

Well, you don't need a slot for this single line, just connect to SIGNAL(abortFinished())

mrow4a: abortFinished signal is being send after 5 seconds timeout regardless if jobs terminated in that time or not. I give jobs max 5 seconds to finish what they started.

mrow4a: maybe I should change namings, since I see your confusion.

mrow4a: @ogoffart so I can connect signal to signal?

ogoffart:
Looks mostly good.  But don't forget the put in the V1,  this one is even more important as it happens more often.

mrow4a:
Yep, I will do now V1, just wanted to verify that aborting implementation will work for NG (if it will work there, it will probably work everywhere just by implementing abort() in the job)

src/libsync/owncloudpropagator.h
mrow4a: Hmm, maybe in fact I dont need it also... it will roll back to what was before this PR for the jobs which did not terminate within 5sec

mrow4a:
Guys, I tried also the version where CompositeJob waits for aborts from its child, but for some reason with "not aborted MOVE", it was not working properly (looks like a mix of finished and abortFinished signals or even their lack messes things). It took me quite long to debug everything. 

Should I try once again Timeout in the job itself, or keep with version of timeout in  composite job (which will finish sync run no matter what since it does not care if subjob finished abort or not) ?

mrow4a:
@ogoffart @guruz @SamuAlfageme PR ready, fixes both V1 and NG

src/libsync/owncloudpropagator.cpp
```diff
@@ -734,6 +740,20 @@ PropagatorJob::JobParallelism PropagatorCompositeJob::parallelism()
     return FullParallelism;
 }
 
+void PropagatorCompositeJob::abort()
+{
+    foreach (PropagatorJob *j, _runningJobs)
+        j->abort();
```
guruz: please use { and }

src/libsync/owncloudpropagator.h
```diff
@@ -210,11 +210,7 @@ class PropagatorCompositeJob : public PropagatorJob
 
     virtual bool scheduleSelfOrChild() Q_DECL_OVERRIDE;
     virtual JobParallelism parallelism() Q_DECL_OVERRIDE;
-    virtual void abort() Q_DECL_OVERRIDE
-    {
-        foreach (PropagatorJob *j, _runningJobs)
-            j->abort();
-    }
+    virtual void abort() Q_DECL_OVERRIDE;
```
guruz: Please also document here that the abort() might not be synchronous/immediate

mrow4a:
> I'm missing a bit of info what happens in a situation where an immediate abort does not work. What will > happen after those 5000msec? It will go to emitFinished but what happens with the running stuff?

Good question, in my observation when job finishes the sync run is already gone, thus whatever finishes after that time is ignored.

I think I will try to implement "forced abort".

guruz:
I'm not sure, maybe the QNetworkReplys are parented to the Jobs which get deleted when SyncEngine is deleted.
(Which is the force abort, basically)

mrow4a:
@guruz how to check it? @jturcotte

jturcotte:
Maybe just call `abort` in the code where you think the issue could appear.

For example in tests we have `FakeFolder::execUntilItemCompleted` that allows you to something right after an `itemCompleted` signal. But you could also just test this manually in the code with a QTimer or something like that.

src/libsync/owncloudpropagator.h
ogoffart: Actually, the PropagatorJob::abort documentation should state that this function must emit abortFinished

src/libsync/owncloudpropagator.h
```diff
@@ -261,6 +261,9 @@ class OWNCLOUDSYNC_EXPORT PropagateDirectory : public PropagatorJob
     virtual JobParallelism parallelism() Q_DECL_OVERRIDE;
     virtual void abort() Q_DECL_OVERRIDE
     {
+        // Abort first job and sub jobs
+        // Finished abort will be announced via abortFinished()
+        // (look contructor implementation)
```
ogoffart: (Maybe you can do the connection here.)

Here as well you need to have a counter to make sure that both jobs sent their abortFinished.

src/libsync/propagateuploadng.cpp
```diff
@@ -502,4 +502,20 @@ void PropagateUploadFileNG::slotUploadProgress(qint64 sent, qint64 total)
...
 }
+
+
+void PropagateUploadFileNG::abort()
+{
+    foreach (AbstractNetworkJob *job, _jobs) {
+        // Dont abort MoveJob since it might result in conflict,
+        // only PUT and MKDIR jobs can be safely aborted
+        if (qobject_cast<MoveJob *>(job)){
+            continue;
```
ogoffart: I miss the point where abortFinished is sent when that jobs finished

src/libsync/propagateuploadv1.cpp
```diff
@@ -360,4 +358,21 @@ void PropagateUploadFileV1::slotUploadProgress(qint64 sent, qint64 total)
...
 }
+
+void PropagateUploadFileV1::abort()
+{
+    foreach (AbstractNetworkJob *job, _jobs) {
+        // Dont abort final PUT which uploaded its data,
+        // since this might result in conflicts
+        if (PUTFileJob *putJob = qobject_cast<PUTFileJob *>(job)){
+            if (isFinalChunk() && putJob->device()->atEnd()) {
+                continue;
```
ogoffart: Same thing, We need to emit abortFinished when the job finishes.

mrow4a:
@ogoffart I did that, and for some reason it was not working properly e.g. finished signal never went back after aborting, and other issues... I can give a try once again on that approach.. 

I just discovered that I also have to be careful not to break abort mechanism on network error with move, since it will finish in infinite aborts loop (it will retry all the time).

src/libsync/owncloudpropagator.h
```diff
@@ -394,11 +399,13 @@ class OwncloudPropagator : public QObject
     {
         _abortRequested.fetchAndStoreOrdered(true);
         if (_rootJob) {
-            // We're possibly already in an item's finished stack
```
ogoffart: keep this comment, it explain why we do a QueuedConnection

src/libsync/owncloudpropagator.h
```diff
@@ -394,11 +399,13 @@ class OwncloudPropagator : public QObject
...
+            // Invoke abort and wait for abort finish to finalize
+            connect(_rootJob.data(), SIGNAL(abortFinished()), this, SLOT(slotAbortFinished()));
```
ogoffart: maybe here we could add a timeout in case the jobs takes too long to finish .

src/libsync/owncloudpropagator.h
```diff
@@ -395,10 +442,15 @@ class OwncloudPropagator : public QObject
         _abortRequested.fetchAndStoreOrdered(true);
         if (_rootJob) {
             // We're possibly already in an item's finished stack
-            QMetaObject::invokeMethod(_rootJob.data(), "abort", Qt::QueuedConnection);
+            // Invoke abort and wait for abort finish
+            // to finalize asynchronously with abortFinished
+            connect(_rootJob.data(), SIGNAL(abortFinished()), this, SLOT(slotAbortFinished()));
+            QTimer::singleShot(jobTimeoutDuration(), this, SLOT(abortTimeout()));
```
mrow4a: I give the jobs 5sec to finish on their own asychronously, otherwise I force them to abort synchronously and invoke emitFinished

mrow4a:
@ogoffart @guruz ready for review

mrow4a:
Any input here @ogoffart @guruz @jturcotte ? Should I split it to smaller PRs?

src/libsync/owncloudpropagator.cpp
```diff
@@ -410,6 +410,12 @@ quint64 OwncloudPropagator::smallFileSize()
     return smallFileSize;
 }
 
+quint64 OwncloudPropagator::jobTimeoutDuration()
```
ogoffart: You don't need this function. And that's not even a good name. (you don't mention it is for abort, you don't mention what unit it is)

Since it is only used once, just have the constant in OwncloudPropagator::abort

(Note: in new code, the time constant should probably be in Ideally, in new code, the type of such constant should be `std::chrono::milliseconds` so the unit is in the type )

src/libsync/owncloudpropagator.h
```diff
@@ -419,6 +461,17 @@ class OwncloudPropagator : public QObject
 
 private slots:
 
+    void abortTimeout()
+    {
+        // Abort synchronously (asyncAbort set to false) and finish
+        QMetaObject::invokeMethod(_rootJob.data(), "abort", Qt::QueuedConnection, Q_ARG(bool, false));
```
ogoffart: Is there a reason you use a QueuedConnection here?
the abort(false) will be called after the slotAbortFinished.
If it is required, add a comment why we don't just call abort(false),   if it is not required, just make a call to abort(false)



(also that's a short function, it might be cleaner to use a lambda for it)

src/libsync/owncloudpropagator.h
```diff
@@ -419,6 +461,17 @@ class OwncloudPropagator : public QObject
...
+        slotAbortFinished();
+    }
+
+    void slotAbortFinished() {
+        QMetaObject::invokeMethod(this, "emitFinished", Qt::QueuedConnection, Q_ARG(SyncFileItem::Status, SyncFileItem::NormalError));
```
ogoffart: please add a comment on why we need a queued connection

ogoffart:
Also please rebase in a single commit.

mrow4a:
Ok, did what requested @ogoffart :
- [x] Used QT5 connect style
- [x] Updated/Deleted Queued Connection where required
- [x] Test New Chunking MOVE abort 
- [x] Test New Chunking MOVE abort with timeout of 5s exceeded -> it starts downloading file after forced abort as in the synchronous scenario
- [x] Test OLD Chunking FINAL PUT abort 
- [x] Test New Chunking FINAL PUT abort with timeout of 5s exceeded -> it starts downloading file after forced abort as in the synchronous scenario

mrow4a:
Why this PR started failing continuous-integration test for ChunkingNG? I dont get the error and it makes no sense

jturcotte:
@mrow4a A few lines lower:
```
16: 09-05 14:45:07:141 [ info sync.engine ]:	There are 487604224 bytes available at "/tmp/ChunkingNgTest-ULWWGt/" and at least 50000000 are required
```
@jnweiger said there is only 1GB in /tmp on that machine. Maybe we could lower that threshold for auto tests so that we don't have to do manual cleanup that often.

src/libsync/owncloudpropagator.h
```diff
@@ -98,7 +98,14 @@ class PropagatorJob : public QObject
     virtual qint64 committedDiskSpace() const { return 0; }
 
 public slots:
-    virtual void abort() {}
+    /*
+     * Asynchronous abort requires emit of abortFinished() signal,
+     * while synchronous is expected to abort immedietaly.
+    */
+    virtual void abort(const bool asyncAbort) {
```
ogoffart: parameters passed by value should not be const.   It should simply be `bool asyncAbort`


(Ideally, even, this should not be a bool, but an enum { Synchronously, Asynchronously } )

mrow4a: omg, I overlooked it, of course it makes no sense, thanks! about enum, I am not sure, but definitely that would improve readability.

src/libsync/owncloudpropagator.cpp
```diff
@@ -891,7 +901,8 @@ void PropagateDirectory::slotFirstJobFinished(SyncFileItem::Status status)
 
     if (status != SyncFileItem::Success && status != SyncFileItem::Restoration) {
         if (_state != Finished) {
-            abort();
+            // Synchronously abort
+            abort(AbortType::Synchronous);
```
mrow4a: Now, instead of boolean, there is enum

src/libsync/owncloudpropagator.h
```diff
@@ -98,7 +103,14 @@ class PropagatorJob : public QObject
     virtual qint64 committedDiskSpace() const { return 0; }
 
 public slots:
-    virtual void abort() {}
+    /*
+     * Asynchronous abort requires emit of abortFinished() signal,
+     * while synchronous is expected to abort immedietaly.
+    */
+    virtual void abort(PropagatorJob::AbortType abortType) {
```
mrow4a: function requires "full definition" due to the fact that we call it using invokeMethod. Without it it thrown ```QMetaObject::invokeMethod: No such method OCC::PropagateDirectory::abort(PropagatorJob::AbortType)
Candidates are:
    abort(AbortType)```

src/libsync/owncloudpropagator.h
```diff
@@ -324,6 +364,7 @@ class OwncloudPropagator : public QObject
         , _chunkSize(10 * 1000 * 1000) // 10 MB, overridden in setSyncOptions
         , _account(account)
     {
+        qRegisterMetaType<PropagatorJob::AbortType>("PropagatorJob::AbortType");
```
mrow4a: It needs to register metatype, othewise ```09-06 16:41:33:780 [ info sync.engine ]:	Aborting sync
09-06 16:41:33:780 [ warning default ]:	QMetaMethod::invoke: Unable to handle unregistered datatype 'PropagatorJob::AbortType'
09-06 16:41:33:780 [ fatal default ]:	ASSERT: "invoked" in file /home/mrow4a/Projects/client/src/libsync/owncloudpropagator.h, line 444```

ckamm:
@jturcotte About
```
16: 09-05 14:45:07:141 [ info sync.engine ]:	There are 487604224 bytes available at "/tmp/ChunkingNgTest-ULWWGt/" and at least 50000000 are required
```
Note that this says: 487MB are available and 50MB are required. I've also stumbled over this and will adjust the message to make it more clear whether it's informative or an error.

ckamm:
I've rebased to master and will review the current state to help move this forward.

ckamm:
Looks good. My remaining worry is about what happens when abort(Async) is called twice. Do we guard against this somewhere?

mrow4a:
I mean, if you call it again, it will basically let unfinished job run (as with previous call of abort(Async)), the other jobs are synchronously aborted. The only difference is, that now you might have 2 timers (timeout timers) running in parallel, and you add another connects on signals. @ogoffart is that a problem?

ckamm:
@mrow4a I added a minor guard against it that seems sufficient for me.

ckamm:
@mrow4a @guruz @ogoffart Are there any remaining concerns (also about the code I added)? Can we merge this?

src/libsync/propagateupload.h
```diff
@@ -320,6 +320,10 @@ class PropagateUploadFileV1 : public PropagateUploadFileCommon
 public:
     PropagateUploadFileV1(OwncloudPropagator *propagator, const SyncFileItemPtr &item)
         : PropagateUploadFileCommon(propagator, item)
+        , _startChunk(0)
+        , _currentChunk(0)
```
ogoffart: Note: this could be initialized in the header file.

src/libsync/propagateuploadv1.cpp
```diff
@@ -363,8 +363,10 @@ void PropagateUploadFileV1::abort(PropagatorJob::AbortType abortType)
             // dont abort final PUT which uploaded its data,
             // since this might result in conflicts
             if (PUTFileJob *putJob = qobject_cast<PUTFileJob *>(job)){
-                if (abortType == AbortType::Asynchronous && (((_currentChunk + _startChunk) % _chunkCount) == 0)
-                        && putJob->device()->atEnd()) {
+                if (abortType == AbortType::Asynchronous
+                    && _chunkCount > 0
```
ogoffart: I wonder how chunk count can be 0

src/libsync/propagateuploadv1.cpp
```diff
@@ -350,4 +350,31 @@ void PropagateUploadFileV1::slotUploadProgress(qint64 sent, qint64 total)
...
+
+    // Abort all jobs (if there are any left), except final PUT
+    foreach (AbstractNetworkJob *job, _jobs) {
+        if (job->reply()) {
+            // If asynchronous abort allowed,
+            // dont abort final PUT which uploaded its data,
+            // since this might result in conflicts
+            if (PUTFileJob *putJob = qobject_cast<PUTFileJob *>(job)){
+                if (abortType == AbortType::Asynchronous
+                    && _chunkCount > 0
```
mrow4a: Can you elaborate more here?

mrow4a:
Any progress here?

src/libsync/propagateupload.h
mrow4a: Should I step in and fix it, or @ckamm ?

src/libsync/propagateuploadv1.cpp
mrow4a: I guess this is safety check right?

src/libsync/propagateupload.h
ckamm: Sorry, I missed the notifications for this. Looking now.

src/libsync/propagateuploadv1.cpp
ckamm: `chunkCount` gets set by `doStartUpload` and the job's `abort` can be called before we get there.

src/libsync/propagateuploadv1.cpp
ckamm: See above. I was getting crashes because of zero chunkCount when I aborted jobs quickly.

src/libsync/propagateupload.h
ckamm: @mrow4a (in general, feel free to just fix issues if I'm unresponsive!)

ckamm:
@mrow4a I think you can merge this now

src/libsync/propagateuploadv1.cpp
mrow4a: Ohh that would make sense did not try indeed this corner case.

mrow4a:
What is happening with Jenkins build? Did you run tests locally?

ckamm:
No, I hadn't. The test that fails to compile on jenkins works locally. But I'll check what's going on there.

ckamm:
Jenkins should pass now, let's wait and see.

ckamm:
@DeepDiver1975 I think we might want this in 2.4, but jenkins failed in something after running the tests. Is this a transient issue where retriggering jenkins would help? Could you do that?

ogoffart:
We anyway need to merge this to the 2.4 branch.

ckamm:
Targeted to 2.4 and merged!

PR #5955, opened 2017-08-11
Windows: Use the application icon for the sidebar

### LINKED ISSUE #2446: Proposal: Add icon (from owncloud.exe) to explorer favorites shortcut in Windows 7-8.1 created via installer
**Proposal:**
Add icon (from owncloud.exe) to explorer favorites shortcut in Windows 7-8.1 created via installer.

_Currently it is only the default folder icon._

jturcotte:
By setting the icon in Desktop.ini of the root folder, this adds the icon
both when browsing the folder directly and to the sidebar shortcut.

To avoid overwriting any user setting that could exist in Desktop.ini,
only do this if the file doesn't exist. Editing .ini files on Windows
isn't trivial and isn't worth it given that this file won't exist most
of the time.

Fixes #2446

### DIFF
--- a/src/libsync/utility_win.cpp
+++ b/src/libsync/utility_win.cpp
@@ -28,8 +28,20 @@ namespace OCC {
 
 static void setupFavLink_private(const QString &folder)
 {
-    // Windows Explorer: Place under "Favorites" (Links)
+    // First create a Desktop.ini so that the folder and favorite link show our application's icon.
+    QFile desktopIni(folder + QLatin1String("/Desktop.ini"));
+    if (desktopIni.exists()) {
+        qCWarning(lcUtility) << desktopIni.fileName() << "already exists, not overwriting it to set the folder icon.";
+    } else {
+        qCInfo(lcUtility) << "Creating" << desktopIni.fileName() << "to set a folder icon in Explorer.";
+        desktopIni.open(QFile::WriteOnly);
+        desktopIni.write("[.ShellClassInfo]\r\nIconResource=");
+        desktopIni.write(QDir::toNativeSeparators(qApp->applicationFilePath()).toUtf8());
+        desktopIni.write(",0\r\n");
+        desktopIni.close();
+    }
 
+    // Windows Explorer: Place under "Favorites" (Links)
     QString linkName;
     QDir folderDir(QDir::fromNativeSeparators(folder));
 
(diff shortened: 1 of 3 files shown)


jturcotte:
One thing to think about: do we allow the default ignore list to be edited in order to allow syncing `Desktop.ini` files?
This would mean that every new device starting to sync a folder would probably propagate its new Desktop.ini to other machines (possibly with an icon path that won't work for them).

src/libsync/utility_win.cpp
```diff
@@ -28,8 +28,20 @@ namespace OCC {
...
-    // Windows Explorer: Place under "Favorites" (Links)
+    // First create a Desktop.ini so that the folder and favorite link show our application's icon.
+    QFile desktopIni(folder + QLatin1String("/Desktop.ini"));
+    if (desktopIni.exists()) {
+        qCWarning(lcUtility) << desktopIni.fileName() << "already exists, not overwriting it to set the folder icon.";
+    } else {
+        qCInfo(lcUtility) << "Creating" << desktopIni.fileName() << "to set a folder icon in Explorer.";
+        desktopIni.open(QFile::WriteOnly);
+        desktopIni.write("[.ShellClassInfo]\r\nIconResource=");
+        desktopIni.write(qApp->applicationFilePath().toUtf8().replace('/', '\\'));
```
ogoffart: should any other character be escaped?

jturcotte: It's not escaped (besides the C++ literal), it converts slashes to backslashes.

ogoffart: Ah right.

Then maybe QDir::toNativeSeparators would be better

ogoffart: Are you sure UTF-8  is the right encoding for desktop.ini?

jturcotte: Ah yes, forgot about toNativeSeparators again, I'll fix it.
I thought I saw a not on some documentation that this should be UTF-8, but I can't find it anymore. If there is non-latin1 characters in the path, it won't work if I use toLatin1 either, so I'd keep it that way.

jturcotte: done

jturcotte:
I also added b5a1270 to hardcode the `desktop.ini` exclusion, could you quickly review?

guruz:
@jturcotte Could you harmonize Desktop.ini vs desktop.ini to whatever Windows normally does?
I know it doesn't really play a role on Windows, but still :/

(btw, does csync_fnmatch watch case?)

jturcotte:
Thanks, I was waiting for the CI to pass, the previous build didn't pass.

PR #5979, opened 2017-08-22
csync: Merge csync_vio_file_stat_t, csync_file_stat_t and TREE_WALK_FILE data structures

### LINKED ISSUE #1817: Consider unifying csync_file_stat_s and csync_vio_file_stat_s
Even though we mostly got rid of the vio abstraction, this is a layer that is still there.
The conversion is handled in `_csync_detect_update` in `csync_update.c`

Related to #1813 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/2358904-consider-unifying-csync_file_stat_s-and-csync_vio_file_stat_s?utm_campaign=plugin&utm_content=tracker%2F216457&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F216457&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jturcotte:
Since we don't need a public/private layer in csync anymore, those three data structures achieving the same just get in the way.

This will use csync_file_stat_t for everything, and keep the same instances created during local and remote discovery all the way up to the SyncEngine (database instances are thrown away). This already starts using QByteArray for everything in those data structures, keeping UTF-8 in csync as it was before.

The Utility class was moved to a new static src/common library and is not dynamically exported (which means its binary gets duplicated in ocsync, libsync and gui) and should prevent having to add C++ stuff needed by csync from now on to have to be put in csync directly.

Fixes #1817


jturcotte:
I'm creating the pull request already for context, but we should only merge this to master once 2.4 has been branched.

src/csync/std/c_utf8.cpp
```diff
@@ -34,31 +35,26 @@
...
   len = wcslen(wstr);
   /* Call once to get the required size. */
   size_needed = WideCharToMultiByte(CP_UTF8, 0, wstr, len, NULL, 0, NULL, NULL);
   if (size_needed > 0) {
-    mdst = (char*)c_malloc(size_needed + 1);
-
-    memset(mdst, 0, size_needed + 1);
-    WideCharToMultiByte(CP_UTF8, 0, wstr, len, mdst, size_needed, NULL, NULL);
-    dst = mdst;
+    dst.resize(size_needed);
```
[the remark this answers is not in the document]
jturcotte: Yes

src/csync/vio/csync_vio_local_unix.cpp
```diff
@@ -122,31 +118,22 @@ csync_vio_file_stat_t *csync_vio_local_readdir(csync_vio_handle_t *dhandle) {
...
-        file_stat->type = CSYNC_VIO_FILE_TYPE_DIRECTORY;
+        file_stat->type = CSYNC_FTW_TYPE_DIR;
       } else {
-        file_stat->type = CSYNC_VIO_FILE_TYPE_REGULAR;
+        file_stat->type = CSYNC_FTW_TYPE_FILE;
       }
       break;
-    case DT_UNKNOWN:
-      file_stat->fields |= CSYNC_VIO_FILE_STAT_FIELDS_TYPE;
-      file_stat->type = CSYNC_VIO_FILE_TYPE_UNKNOWN;
```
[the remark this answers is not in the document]
jturcotte: `file_stat->type` is initialized to `CSYNC_FTW_TYPE_SKIP` in the new `csync_file_stat_s` constructor, which has the same outcome as this flag did.

src/csync/vio/csync_vio_local_win.cpp
```diff
@@ -134,94 +134,60 @@ static time_t FileTimeToUnixTime(FILETIME *filetime, DWORD *remainder)
...
   } else {
       if( FindNextFile(handle->hFind, &(handle->ffd)) == 0 ) {
-          // might be error, check!
-          int dwError = GetLastError();
-          if (dwError != ERROR_NO_MORE_FILES) {
-              errno = EACCES; // no more files is fine. Otherwise EACCESS
-          }
-          goto err;
+          return nullptr;
       }
```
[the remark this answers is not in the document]
jturcotte: `errno` was just a local variable that's never read...
`goto err;` used to release memory and return null, now the smart pointer will take care of releasing the memory so I can just return null here.

src/libsync/discoveryphase.h
```diff
@@ -138,20 +106,22 @@ class DiscoverySingleDirectoryJob : public QObject
     void setIsRootPath() { _isRootPath = true; }
     void start();
     void abort();
+    std::vector<std::unique_ptr<csync_file_stat_t>> &&takeResults() { return std::move(_results); }
+
```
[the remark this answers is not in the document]
jturcotte: Although the API is more familiar, the issue with Qt containers is that they don't support rvalue-ref emplacing, which means I'd have to use a `QSharedPointer` instead of a `unique_ptr` (since `unique_ptr` isn't copyable). And we don't need the shared pointer, I like the very clear ownership transfers of `unique_ptr` through `std::move`.

src/libsync/syncengine.cpp
```diff
@@ -1307,10 +1304,14 @@ void SyncEngine::checkForPermission(SyncFileItemVector &syncItems)
...
-                (*it)->_fileId = (*it)->log._other_fileId;
-                (*it)->_etag = (*it)->log._other_etag;
+                ASSERT(_csync_ctx->status != CSYNC_STATUS_INIT);
+                auto csyncIt = _csync_ctx->remote.files.find((*it)->_file.toUtf8());
+                if (csyncIt != _csync_ctx->remote.files.end()) {
+                    (*it)->_modtime = csyncIt->second->modtime;
+                    (*it)->_size = csyncIt->second->size;
+                    (*it)->_fileId = csyncIt->second->file_id;
+                    (*it)->_etag = csyncIt->second->etag;
+                }
```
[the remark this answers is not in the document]
jturcotte: The comment already there explains it, but added another line to explain the lookup.

src/libsync/discoveryphase.cpp
```diff
@@ -704,16 +677,14 @@ csync_vio_handle_t *DiscoveryJob::remote_vio_opendir_hook(const char *url,
...
     void *userdata)
 {
     DiscoveryJob *discoveryJob = static_cast<DiscoveryJob *>(userdata);
     if (discoveryJob) {
         DiscoveryDirectoryResult *directoryResult = static_cast<DiscoveryDirectoryResult *>(dhandle);
         if (directoryResult->listIndex < directoryResult->list.size()) {
-            csync_vio_file_stat_t *file_stat = directoryResult->list.at(directoryResult->listIndex++).data();
-            // Make a copy, csync_update will delete the copy
-            return csync_vio_file_stat_copy(file_stat);
+            return std::move(directoryResult->list.at(directoryResult->listIndex++));
```
ogoffart: You are moving from an item in the list?  leaving a moved-from item in the list?
That's OK. But maybe having a queue data structure would be better.

src/csync/vio/csync_vio_local_unix.cpp
```diff
@@ -83,31 +84,26 @@ int csync_vio_local_closedir(csync_vio_handle_t *dhandle) {
...
+
+std::unique_ptr<csync_file_stat_t> csync_vio_local_readdir(csync_vio_handle_t *dhandle) {
 
   dhandle_t *handle = NULL;
-  csync_vio_file_stat_t *file_stat = NULL;
 
   handle = (dhandle_t *) dhandle;
   struct _tdirent *dirent = NULL;
-
-  errno = 0;
```
ogoffart: This must stay i think...   or?

src/csync/vio/csync_vio_local_win.cpp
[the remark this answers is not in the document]
jturcotte: Ah yes you're both right then, for some reason I sloppily imagined `int errno = 0` instead of `errno = 0`.

src/libsync/discoveryphase.cpp
jturcotte: Humm but then the index won't make sense. Since the caller just calls this repetitively without providing the index, this should be OK, no?

jturcotte: Although yeah I could just remove that weird index that's true.

src/csync/csync_update.cpp
```diff
@@ -481,63 +467,56 @@ static int _csync_detect_update(CSYNC *ctx, const char *file,
...
+      if (c_rbtree_insert(ctx->remote.tree, (void *) st.release()) < 0) {
         ctx->status_code = CSYNC_STATUS_TREE_ERROR;
         return -1;
       }
       break;
     default:
       break;
   }
-  CSYNC_LOG(CSYNC_LOG_PRIORITY_INFO, "file: %s, instruction: %s <<=", st->path,
-      csync_instruction_str(st->instruction));
```
ogoffart: did you just remove the instructions log?

src/libsync/syncjournaldb.cpp
```diff
@@ -900,21 +897,17 @@ bool SyncJournalDb::setFileRecord(const SyncJournalFileRecord &_record)
         int contentChecksumTypeId = mapChecksumType(checksumType);
         _setFileRecordQuery->reset_and_clear_bindings();
         _setFileRecordQuery->bindValue(1, QString::number(phash));
-        _setFileRecordQuery->bindValue(2, plen);
```
ogoffart: Not storing the plen will break downgrading the client, would it not?

src/csync/csync_update.cpp
```diff
@@ -96,6 +76,14 @@ static bool _last_db_return_error(CSYNC* ctx) {
     return ctx->statedb.lastReturnValue != SQLITE_OK && ctx->statedb.lastReturnValue != SQLITE_DONE && ctx->statedb.lastReturnValue != SQLITE_ROW;
 }
 
+static QByteArray _rel_to_abs(CSYNC* ctx, const QByteArray &relativePath) {
+    QByteArray abs = QByteArray::fromRawData(ctx->local.uri, strlen(ctx->local.uri));
```
ogoffart: this still allocates.  and then re-allocates.

It should simply be:

```
return QByteArray() % ctx->local.uri % '/' % relativePath;
```

Let QStringBuilder do the job and avoid allocation.

src/csync/csync_update.cpp
```diff
@@ -124,38 +112,21 @@ static bool _csync_mtime_equal(time_t a, time_t b)
...
  * See doc/dev/sync-algorithm.md for an overview.
  */
-static int _csync_detect_update(CSYNC *ctx, const char *file,
-    const csync_file_stat_t *fs, enum csync_ftw_type_e type) {
-  uint64_t h = 0;
-  size_t len = 0;
-  size_t size = 0;
-  const char *path = NULL;
-  std::unique_ptr<csync_file_stat_t> st;
+static int _csync_detect_update(CSYNC *ctx, std::unique_ptr<csync_file_stat_t> fs) {
```
ogoffart: Can we keep the absolute file in the arguments? ...

src/csync/csync_update.cpp
```diff
@@ -245,57 +208,57 @@ static int _csync_detect_update(CSYNC *ctx, const char *file,
...
             // Checksum comparison at this stage is only enabled for .eml files,
             // check #4754 #4755
-            bool isEmlFile = csync_fnmatch("*.eml", file, FNM_CASEFOLD) == 0;
+            bool isEmlFile = csync_fnmatch("*.eml", fs->path, FNM_CASEFOLD) == 0;
             if (isEmlFile && fs->size == tmp->size && !tmp->checksumHeader.isEmpty()) {
                 if (ctx->callbacks.checksum_hook) {
-                    st->checksumHeader = ctx->callbacks.checksum_hook(
-                        file, tmp->checksumHeader,
+                    fs->checksumHeader = ctx->callbacks.checksum_hook(
+                        _rel_to_abs(ctx, fs->path), tmp->checksumHeader,
```
ogoffart: ... so we don't need to allocate memory here for the path.

src/csync/csync_update.cpp
```diff
@@ -387,18 +387,13 @@ static int _csync_detect_update(CSYNC *ctx, std::unique_ptr<csync_file_stat_t> f
...
       csync_instruction_str(fs->instruction));
 
+  QByteArray path = fs->path;
   switch (ctx->current) {
     case LOCAL_REPLICA:
-      if (c_rbtree_insert(ctx->local.tree, (void *) fs.release()) < 0) {
-        ctx->status_code = CSYNC_STATUS_TREE_ERROR;
-        return -1;
-      }
+      ctx->local.files.emplace(path, std::move(fs));
```
ogoffart: s/emplace/insert/

src/csync/csync_private.h
```diff
@@ -103,14 +104,15 @@ struct csync_s {
     int lastReturnValue;
   } statedb;
 
+  typedef std::map<QByteArray, std::unique_ptr<csync_file_stat_t>> FileMap;
```
ogoffart: could also be std::unordered_map

src/csync/csync.cpp
```diff
@@ -525,7 +429,7 @@ int csync_destroy(CSYNC *ctx) {
   SAFE_FREE(ctx->local.uri);
```
ogoffart: can we put all of this in the destructor of CSYNC where it belongs?

src/csync/csync.cpp
```diff
@@ -51,42 +51,11 @@
...
-
-  return 0;
-}
-
 void csync_create(CSYNC **csync, const char *local) {
   CSYNC *ctx;
   size_t len = 0;
 
-  ctx = (CSYNC*)c_malloc(sizeof(CSYNC));
+  ctx = new CSYNC{};
```
ogoffart: c_malloc was zeroing some fields. 

 I guess we should put some of the code bellow in the constuctor or in the the struct itself, in the form of  `=0` and so on.

src/libsync/syncengine.cpp
```diff
@@ -1473,7 +1465,18 @@ QByteArray SyncEngine::getPermissions(const QString &file) const
...
     }
-    return _remotePerms.value(file);
+
+    // Fetch from the csync context while we still have it.
+    ASSERT(_csync_ctx->status != CSYNC_STATUS_INIT);
+
+    if (file == QLatin1String(""))
+        return _csync_ctx->remote.root_perms;
+
+    auto it = _csync_ctx->remote.files.find(file.toUtf8());
```
ogoffart: too bad we now have to allocate only to convert to a QByteArray,  but i guess this is compensated by the fact we did not insert it into the map in the first place.

src/csync/vio/csync_vio_local_unix.cpp
jturcotte: Looks like this came from https://github.com/owncloud/client/commit/c281a720b030fc52c22036f10fa2d2f3180e4359 and can probably go since we don't rely on errno to detect errors anymore.

src/csync/vio/csync_vio_local_win.cpp
```diff
@@ -134,94 +134,60 @@ static time_t FileTimeToUnixTime(FILETIME *filetime, DWORD *remainder)
...
   }
-  file_stat->name = c_utf8_from_locale(handle->ffd.cFileName);
+  file_stat.reset(new csync_file_stat_t);
+  file_stat->path = c_utf8_from_locale(handle->ffd.cFileName);
```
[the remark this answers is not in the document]
jturcotte: Yeah it's a mess, trying no to touch more than I should and keep things in some readable state, but passing clang-format could be worth it.

src/csync/csync_update.cpp
jturcotte: I moved it before the tree insert, where I release the `unique_pointer`.

src/libsync/syncjournaldb.cpp
jturcotte: Good point, I'll just remove that whole commit I guess.

src/csync/csync_update.cpp
jturcotte: This is using `fromRawData` and I don't see why it would detach, why would it allocate?

src/csync/csync_update.cpp
jturcotte: This is currently only for `.eml` files, and I don't like having a mix of absolute and relative paths here and there. I think always using absolute makes things easier to follow.

We're already doing tons of allocations everywhere, I think that having one here is fine, and I'm already saving quite a few by passing CoW QByteArrays instead of c_strdups everywhere.
Let me know what you think.

src/csync/csync.cpp
jturcotte: Where is that zeroing happening?

src/csync/csync.cpp
jturcotte: I tried to move this stuff into constructor/destructor but the whole csync_init, csync_commit and csync_destroy stuff wouldn't match very well to that concept and I thought I'd just leave things like they were and just do the `new` to get my std::maps constructed.

This could maybe help with that clang crash though, could be worth a try.

src/csync/csync_private.h
jturcotte: I had a hard time making that call since we iterate those trees, but I realize now that the previous tree was ordered by phash so I guess the order didn't really matter.

Would there be any advantage to use `std::map`?

jturcotte: Meh I just tried and it complains about `error: implicit instantiation of undefined template 'std::__1::hash<QByteArray>'`.
I'd have to bridge qHash on top of std hash or something like that I guess, not sure if this is worth it.

src/csync/csync_update.cpp
jturcotte: It would need to be `insert(std::make_pair(path, std::move(fs))`, is this better? I'm changing the vector to push_back, but emplace would make sense here I think (although can't really understand why would anybody use insert unless they already have the key/value pair).

src/csync/csync.cpp
jturcotte: Ahh, `calloc` does. The `CSYNC{}` syntax did zero fields since it doesn't have an explicit default constructor, but I'll add normal constructors/destructors.

src/csync/csync_update.cpp
ogoffart: fromRawData still allowates memory for the QByteArrayData.  That makes the use of fromRawData quite useless, actually.

src/csync/vio/csync_vio_local_win.cpp
```diff
@@ -134,94 +134,60 @@ static time_t FileTimeToUnixTime(FILETIME *filetime, DWORD *remainder)
...
+std::unique_ptr<csync_file_stat_t> csync_vio_local_readdir(csync_vio_handle_t *dhandle) {
 
   dhandle_t *handle = NULL;
-  csync_vio_file_stat_t *file_stat = NULL;
+  std::unique_ptr<csync_file_stat_t> file_stat;
   DWORD rem;
 
   handle = (dhandle_t *) dhandle;
 
-  errno = 0;
```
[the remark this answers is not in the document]
jturcotte: done

src/libsync/discoveryphase.cpp
```diff
@@ -423,25 +397,24 @@ void DiscoverySingleDirectoryJob::directoryListingIteratedSlot(QString file, con
...
+            file_stat->remotePerm.replace('M', 'm');
         }
 
         QStringRef fileRef(&file);
         int slashPos = file.lastIndexOf(QLatin1Char('/'));
         if (slashPos > -1) {
             fileRef = file.midRef(slashPos + 1);
         }
-        _results.append(file_stat);
+        _results.emplace_back(std::move(file_stat));
```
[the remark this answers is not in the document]
jturcotte: done

src/libsync/discoveryphase.cpp
jturcotte: Done, used an `std::deque`

src/libsync/syncjournaldb.cpp
jturcotte: Removed

src/csync/csync_update.cpp
jturcotte: Done

src/csync/csync_update.cpp
jturcotte: Done

src/csync/csync.cpp
jturcotte: Added a constructor/destructor in 5fb559c

jturcotte:
@ogoffart Was proposing to push parts of this branch to 2.4 in order to be able to already use the src/common separation without having to implement a feature twice if we want it in 2.4 csync.

I kept the less risky mechanical type changes and skipped the "Replace the custom rbtree by a std::map" and "csync: Keep the csync_file_stat_t instances created during discovery" commits. I'll do a new pull request for those that we'll keep for 2.5 only with some additional stuff (replacing csync_statedb with SyncJournalDB).

src/csync/csync_private.h
```diff
@@ -64,136 +65,93 @@ enum csync_replica_e {
...
+  enum csync_status_codes_e status_code = CSYNC_STATUS_OK;
 
-  char *error_string;
+  char *error_string = nullptr;
 
-  int status;
-  volatile int abort;
-  void *rename_info;
+  int status = CSYNC_STATUS_INIT;
+  volatile bool abort = false;
```
ogoffart: (unrelated:  this should be std::atomic i guess)

src/csync/csync.h
```diff
@@ -149,134 +157,51 @@ enum csync_ftw_type_e {
...
-
-  int flags;
-
-  char *original_name; // only set if locale conversion fails
-
-  // For remote file stats: the highest quality checksum the server provided
-  // in the "SHA1:324315da2143" form.
-  char *checksumHeader;
+  enum csync_ftw_type_e type BITFIELD(4);
+  bool child_modified BITFIELD(1);
```
ogoffart: Ony the enum needs the BITFIELD macro.   The bool can stay 1.
But that's not su important because it is only for old compiler anyway.

jturcotte:
Merged manually after fixup-rebase in 57c812b

Issue #5257, opened 2016-10-17
[macOS] Overlay icon missed in folder after renaming to other existing folder but not synced


davitol:
1. In the OC server, create a folder named 'ama' and other named 'Ama' and add info inside them
2. In the desktop client, select sync selective only one folder e.g. ama
3. Rename in the desktop folder 'ama' to 'Ama'

Current behaviour

The folder has no longer an overlay icon and the files inside have a conflic related to 'what to sync' blacklist

![screen shot 2016-10-17 at 14 01 57](https://cloud.githubusercontent.com/assets/10596560/19436751/4b0e1ef8-9472-11e6-9b68-7cc4702ae3a9.png)

![screen shot 2016-10-17 at 14 00 07](https://cloud.githubusercontent.com/assets/10596560/19436737/3b3e0dbc-9472-11e6-8452-f510529b9377.png)

![screen shot 2016-10-17 at 14 00 34](https://cloud.githubusercontent.com/assets/10596560/19436721/29cc28f2-9472-11e6-9060-29c89015c2d0.png)

![synchange](https://cloud.githubusercontent.com/assets/10596560/19436604/6d0067c4-9471-11e6-8669-c566d1799bc0.gif)

Desktop Client OSX Version 2.3.0-nightly20160929 (build 3715).

ogoffart:
It is expected that the folder cannot be sync'ed because it is on the blacklist.
This is a bug that you don't see overlay icon, but can you reproduce the problem with the 2.3 client?
I have tried on linux and i can clearly see the error icon on both the folder and the files.

guruz:
@davitol @SamuAlfageme Please try with 2.3.2

SamuAlfageme:
@guruz @ogoffart @davitol Yup, still happening, the icon is missing until you force a sync/one is scheduled and turns ✅. Then, if the folder is opened, it goes missing again.

<p align="center">
  <img src="https://cloud.githubusercontent.com/assets/2644445/25841676/57c56834-34a1-11e7-8c5a-4ae24fb78e73.gif"/>
</p>

> It is expected that the folder cannot be sync'ed because it is on the blacklist.

However the original folder appears checked in the selective sync list but it's not re-downloaded (it can't because of the case-unsensitive FS). For consistency I'd uncheck it there.

SamuAlfageme:
**Overlays seem to be fixed for this case**; however case-resolution handling is rather edgy and there's many stuff happening under the hood; let's recap:

[...]

PR #5981, opened 2017-08-28
StatusTracker: Fix different case paths not matching

jturcotte:
Use a custom std::map comparator functor to do all comparisons
on contained QStrings using Qt::CaseInsensitive on macOS and Windows.

Issue #5257

### DIFF
--- a/src/libsync/syncfilestatustracker.cpp
+++ b/src/libsync/syncfilestatustracker.cpp
@@ -25,20 +25,49 @@ namespace OCC {
 
 Q_LOGGING_CATEGORY(lcStatusTracker, "sync.statustracker", QtInfoMsg)
 
-static SyncFileStatus::SyncFileStatusTag lookupProblem(const QString &pathToMatch, const std::map<QString, SyncFileStatus::SyncFileStatusTag> &problemMap)
+static int pathCompare( const QString& lhs, const QString& rhs )
+{
+    // Should match Utility::fsCasePreserving, we want don't want to pay for the runtime check on every comparison.
+    return lhs.compare(rhs,
+#if defined(Q_OS_WIN) || defined(Q_OS_MAC)
+        Qt::CaseInsensitive
+#else
+        Qt::CaseSensitive
+#endif
+        );
+}
+
+static bool pathStartsWith( const QString& lhs, const QString& rhs )
+{
+    return lhs.startsWith(rhs,
+#if defined(Q_OS_WIN) || defined(Q_OS_MAC)
+        Qt::CaseInsensitive
+#else
+        Qt::CaseSensitive
+#endif
+        );
+}
+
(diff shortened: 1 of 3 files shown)


src/libsync/syncfilestatustracker.cpp
```diff
@@ -25,20 +25,49 @@ namespace OCC {
 
 Q_LOGGING_CATEGORY(lcStatusTracker, "sync.statustracker", QtInfoMsg)
 
-static SyncFileStatus::SyncFileStatusTag lookupProblem(const QString &pathToMatch, const std::map<QString, SyncFileStatus::SyncFileStatusTag> &problemMap)
+static int pathCompare( const QString& lhs, const QString& rhs )
+{
+    // Should match Utility::fsCasePreserving, we want don't want to pay for the runtime check on every comparison.
```
ogoffart: And yet, it does not becuase it does not check the env variable.
I suggest you still use that, is the runtime cost so high? 
If you really want to save the runtime cost on every call, i suggest you pass the case sensitivity as a parametter of the comparator.

test/testsyncfilestatustracker.cpp
```diff
@@ -260,6 +260,24 @@ private slots:
...
+        fakeFolder.serverErrorPaths().append("A/a1");
+        fakeFolder.localModifier().appendByte("A/a1");
+
+        fakeFolder.syncOnce();
+        QCOMPARE(fakeFolder.syncEngine().syncFileStatusTracker().fileStatus(""), SyncFileStatus(SyncFileStatus::StatusWarning));
+        QCOMPARE(fakeFolder.syncEngine().syncFileStatusTracker().fileStatus("A"), SyncFileStatus(SyncFileStatus::StatusWarning));
+        QCOMPARE(fakeFolder.syncEngine().syncFileStatusTracker().fileStatus("A/a1"), SyncFileStatus(SyncFileStatus::StatusError));
+        QCOMPARE(fakeFolder.syncEngine().syncFileStatusTracker().fileStatus("B"), SyncFileStatus(SyncFileStatus::StatusWarning));
+
+        // Should still get the status for different casing on macOS and Windows.
```
ogoffart: But do we even run the tests on different platform?

Well, at least we could simulate it with env variable.

src/libsync/syncfilestatustracker.cpp
jturcotte: Arf just saw "approved" in the email and didn't see the comment.
The env variable is just for auto tests anyway, no?

test/testsyncfilestatustracker.cpp
jturcotte: I run them on macOS, and the CI ran them on Linux, so at least both cases were checked once.

PR #5995, opened 2017-09-04
Enable DEP & ASLR on Windows

rockihack:
Enable **Data Execution Prevention** (DEP) & **Address space layout randomization** (ASLR) on Windows as exploit mitigation.

See:
https://www.microsoft.com/security/sir/strategy/default.aspx#!section_3_3
https://blogs.technet.microsoft.com/srd/2010/12/08/on-the-effectiveness-of-dep-and-aslr/

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -8,6 +8,12 @@ if(NOT TOKEN_AUTH_ONLY)
     find_package(Qt5Keychain REQUIRED)
 endif()
 
+if(WIN32)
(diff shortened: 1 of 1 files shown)


dschmidt:
Is this flag only supported by GCC or also by MSVC? We should only add it for compilers that support it.

rockihack:
@dschmidt These flags are supported by MSVC and GCC.

guruz:
good to have..

dschmidt:
LGTM then.

PR #5996, opened 2017-09-04
Add 2.3.3 System Requirements to client repo

michaelstingl:
I would like to maintain the System Requirements in the client repo. (Switch planned for 2.4.0 release).

[documentation/admin_manual/installation/system_requirements.rst](https://github.com/owncloud/documentation/blob/master/admin_manual/installation/system_requirements.rst) > [client/doc/installing.rst](https://github.com/owncloud/client/blob/master/doc/installing.rst)

@guruz @settermjd @SamuAlfageme What do you think?

### DIFF
--- a/doc/installing.rst
+++ b/doc/installing.rst
@@ -25,6 +25,20 @@ KWallet, so that the sync client can login automatically.
 You will also find links to source code archives and older versions on the 
 download page.
 
+System Requirements (Version 2.3.3)
+----------------------------------
+
+- Windows 7+
+- Mac OS X 10.7+ (**64-bit only**)
+- CentOS 6 & 7 (64-bit only)
+- Debian 7.0 & 8.0 & 9.0
+- Fedora 24 & 25 & 26
+- Ubuntu 16.04 & 16.10 & 17.04
+- openSUSE Leap 42.1 & 42.2 & 42.3
+
+.. note::
+   For Linux distributions, we support, if technically feasible, the latest 2 versions per platform and the previous `LTS`_.
+
 Installation Wizard
 -------------------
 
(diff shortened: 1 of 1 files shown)


doc/installing.rst
```diff
@@ -25,6 +25,20 @@ KWallet, so that the sync client can login automatically.
 You will also find links to source code archives and older versions on the 
 download page.
 
+System Requirements (Version 2.3.3)
```
ogoffart: Maybe "Supported Platforms"  instead of "system requirements"?
Because it is still possible to compile from source for many other distro.

Also, if this is about 2.3.3, it should go to the 2.3 branch.  master contains is the documentation for 2.4 now.  but if the platform are the same maybe  adding 2.4 is ok

michaelstingl: > it is still possible to compile from source for many other distro.

good point.

Wasn't aware of the process. I other repos we first merge to master, before backport to other branches.

guruz: > Wasn't aware of the process. I other repos we first merge to master, before backport to other branches.

I think @PVince81 @DeepDiver1975 wanted to change that, but never got around it..

guruz: > (Version 2.3.3)

Please don't mention the version. It should be only in VERSION.cmake, everything else is implicit from the branch/tgz/whatever that contains the README :)

michaelstingl: @settermjd Can I Haz Variables in Docs? (from VERSION.cmake for example?)

guruz: @settermjd discussed before in https://github.com/owncloud/client/pull/5762#issuecomment-300486352

guruz: @michaelstingl Can you just remove the 2.3.3 above then we can merge this?

michaelstingl: done

PR #6001, opened 2017-09-05
Update VERSION.cmake

jnweiger:
We still do nightlies in 23 branch. Their versions should be higher than 2.3.3, right?


guruz:
You can also set the release date in ChangeLog :)

jnweiger:
@guruz Ah, sure. Done :-)

Issue #5972, opened 2017-08-19
Save the new URL when hitting a permanent redirection 301


tflidd:
### Expected behaviour
When you server URL is updated and your are redirected permanently (301) to a new URL, update the URL in the config.

Here: https://mycloud.example.org/owncloud -> https://mycloud.example.org/

Thunderbird lightning detects the redirect and asks if it should update the URL.

### Actual behaviour
Follows the redirect but it means you have to keep the redirection forever

OC Client  version 2.3.2

guruz:
(I thought we did this, but maybe only in wizard.. FYI @ogoffart )

ckamm:
Since I'm set up for redirect testing anyway from #6003 I'll take care of this. I'll add redirection checking in the request for `status.php`. If we see a 301/308 there, it'll update the account with the new url.

guruz:
thanks a ton @ckamm

SamuAlfageme:
👍 yay!

PR #6007, opened 2017-09-08
Use DetermineAuthTypeJob in HttpCredentials

ckamm:
* Move it to networkjobs
* Minor adjustments to its logic
* Fixes redirect handling for oauth/basic http auth check #6003

PR #6009, opened 2017-09-08
Update server url in case of permanent redirection #5972

### LINKED ISSUE #5954: Wizard: 301 'Moved Permanently' redirection is not being followed
### Actual behavior
The client appends `/status.php` and `/owncloud/status.php` to the shortened entrypoint URL, which results on a 404 since that resource was moved permanently to the URL on the `Location` header. Error is displayed. 

<p align="center">
  
</p>

### Expected behavior
The client should be able to follow a redirection chain.  
 
### Steps to reproduce
1. Create a short URL to an ownCloud instance using e.g. https://goo.gl
2. Try to connect using the short URL

ckamm:
This is the first time the account url may update outside of
account setup.

Summary of redirection handling:
1. During account setup (wizard) - fixes #5954
   - status.php gets permanently redirected -> adjust url 
   - authed PROPFIND gets *any* redirection -> adjust url
2. During connectivity ping (ConnectionValidator)
   - status.php gets permanently redirected -> adjust url (new!)

All other redirections should be followed transparently and
don't update the account url in the settings.

See #5972

### DIFF
--- a/src/gui/creds/httpcredentialsgui.cpp
+++ b/src/gui/creds/httpcredentialsgui.cpp
@@ -22,6 +22,7 @@
 #include "creds/httpcredentialsgui.h"
 #include "theme.h"
 #include "account.h"
+#include "networkjobs.h"
 #include <QMessageBox>
 #include "asserts.h"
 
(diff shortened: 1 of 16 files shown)


ckamm:
Depends on #6007

ogoffart:
:+1:   but dependent commit needs fixes for the build

PR #6016, opened 2017-09-11
Fix redirection for accounts that use oauth credentials

ckamm:


ckamm:
Depends on #6009

guruz:
@ckamm Can you re-base so Jenkins runs through?

ckamm:
@guruz Was already done. `TestUtility::testLaunchOnStartup()` failed, but is unrelated.

src/gui/creds/oauth.cpp
```diff
@@ -79,10 +79,11 @@ void OAuth::start()
                     + QLatin1String("/index.php/apps/oauth2/api/v1/token?grant_type=authorization_code&code=")
                     + code
                     + QLatin1String("&redirect_uri=http://localhost:") + QString::number(_server.serverPort()));
-                requestToken.setUserName(Theme::instance()->oauthClientId());
-                requestToken.setPassword(Theme::instance()->oauthClientSecret());
                 QNetworkRequest req;
                 req.setHeader(QNetworkRequest::ContentTypeHeader, "application/x-www-form-urlencoded");
+                QString basicAuth = QString("%1:%2").arg(
+                    Theme::instance()->oauthClientId(), Theme::instance()->oauthClientSecret());
+                req.setRawHeader("Authorization", "Basic " + basicAuth.toUtf8().toBase64());
```
ogoffart: Too bad for the code duplication with the one in httpcredentials,  but so be it, it's not so much code anyway.

src/libsync/abstractnetworkjob.cpp
```diff
@@ -175,6 +175,13 @@ void AbstractNetworkJob::slotFinished()
     if (_followRedirects && !redirectUrl.isEmpty()) {
         _redirectCount++;
 
+        // For POSTs where the query args are sent in the body, we need to
+        // manually forward these to the target url - the server can only
+        // return the query-arg-less redirected url in that case.
+        if (reply()->operation() == QNetworkAccessManager::PostOperation && !_requestBody) {
```
ogoffart: yes.
That's kind of a hack, i fgound myself surpsie that Qt does change the query url to the post data. 
So that's two hacks combinated.

As discussed on IRC, my original intention was that the query URL stay in the URL, as the README of https://github.com/owncloud/oauth2 mention, i did not expect it to be moved to the POST data. So I think the best would be to put them directly in the data and not in the query URL. 

Maybe there should be an extra check here to check that redirectUrl does not already have an encoded query?

ckamm: Good point about not overwriting any query args from the forward. And I might also change the existing POST call to just have a body instead.

ckamm:
@ogoffart Please check again (different body handling in oauth.cpp and httpcreds.cpp, just a warning in ANjob)

PR #6017, opened 2017-09-11
Wizard: Handle url-shortener redirects

ckamm:
Grab any permanent redirects from the base url the user entered before attempting to connect to a modified url (with status.php added).

Solves #5954 
Depends on #6016 (for SimpleNetworkJob), only the last commit is new

### DIFF
--- a/src/gui/owncloudsetupwizard.cpp
+++ b/src/gui/owncloudsetupwizard.cpp
@@ -180,6 +180,10 @@ void OwncloudSetupWizard::slotContinueDetermineAuth()
     // redirect subpaths.
     auto redirectCheckJob = account->sendRequest("GET", account->url());
 
+    // Use a significantly reduced timeout for this redirect check:
+    // the 5-minute default is inappropriate.
+    redirectCheckJob->setTimeout(qMin(2000ll, redirectCheckJob->timeoutMsec()));
+
     // Grab the chain of permanent redirects and adjust the account url
     // accordingly
     auto permanentRedirects = std::make_shared<int>(0);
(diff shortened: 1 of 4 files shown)


guruz:
@ckamm Can you re-base so Jenkins runs through?

ckamm:
I even had to rebase on top of master since jenkins had a merge conflict

src/gui/owncloudsetupwizard.cpp
```diff
@@ -173,13 +173,37 @@ void OwncloudSetupWizard::slotContinueDetermineAuth()
...
+    // Grab the chain of permanent redirects and adjust the account url
+    // accordingly
+    int permanentRedirects = 0;
+    connect(redirectCheckJob, &AbstractNetworkJob::redirected, this,
+        [&, account](QNetworkReply *reply, const QUrl &targetUrl, int count) {
+            int httpCode = reply->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
+            if (count == permanentRedirects && (httpCode == 301 || httpCode == 308)) {
+                qCInfo(lcWizard) << account->url() << " was redirected to" << targetUrl;
+                account->setUrl(targetUrl);
+                permanentRedirects += 1;
```
ogoffart: Are you capturing a reference to a local variable?  does not look right

ckamm:
Aww, fixed. Ironically, I did think about capturing `account`, but then entirely missed that part.

This works with the back button. Similar redirect behavior is already in the other jobs (authed and non-authed). It should work with client certs unless the server issues a permanent redirect for queries that don't have client certs.

ckamm:
@ogoffart I've shortened the timeout (also in other places) and fixed relative redirects (technically a separate issue, but probably a bug that's mostly relevant for these top-level redirects)

PR #6018, opened 2017-09-11
Windows: Make sure that the folder and Desktop.ini have the right attributes

jturcotte:
When created from scratch, the owncloud icon won't be picked-up by
explorer unless the right attributes are set on both the sync root and the
Desktop.ini file.

Issue #2446


SamuAlfageme:
Yesterday, while debugging an error related to a mounted storage (not using NTFS) I noticed neither the `Desktop.ini` metadata gets applied as folder properties (icon and so on) nor the file appears hidden. Weird.

jturcotte:
Maybe the filesystem can't handle attributes.

Issue #5723, opened 2017-04-25
[Linux] syncing Host-shared directory from Guest VM: "Create table metadata" "disk I/O error"


polrus:
### Expected behaviour
Start client and start sync

### Actual behaviour
Crash

### Steps to reproduce
1. Share host directory (win7) to guest Vritualbox VM (LInux)
2. Start client in guest to sync this shared directory
3. client crashes - deleting .db files does not help

### Server configuration
* Operating system: Opensuse
* Web server:  apache
* Database: Mysql
* ownCloud version: latest

### Client configuration
* Client version: 2.3.1 and 2.3.2 rc1 also
* Operating system: Linux Opensuse Leap 42.2
* OS language: En
* Qt version used by client package (Linux only, see also Settings dialog):
* Client package (From ownCloud or distro) (Linux only): Distro OS Leap 42.2

### Logs

```
04-25 12:50:13:190 2901 OCC::FolderMan::addFolderInternal: Adding folder to Folder Map  OCC::Folder(0x1c1b0b0) "1"
04-25 12:50:13:190 2901 OCC::FolderWatcherPrivate::slotAddFolderRecursive: (+) Watcher: "/media/sf_C_DRIVE1/Users/sebastian.turzanski/Documents/"
04-25 12:50:15:432 2901 OCC::FolderWatcherPrivate::slotAddFolderRecursive:     `-> and 346 subdirectories
04-25 12:50:15:432 2901 OCC::FolderMan::scheduleFolder: Schedule folder  "1"  to sync!
04-25 12:50:15:432 2901 OCC::FolderMan::scheduleFolder: Folder is not ready to sync, not scheduled!
04-25 12:50:15:432 2901 OCC::ClientProxy::setupQtProxyFromConfig: Set proxy configuration to use system configuration
04-25 12:50:15:453 2901 OCC::ActivitySettings::setNotificationRefreshInterval: Starting Notification refresh timer with  300  sec interval
04-25 12:50:15:475 2901 OCC::Updater::getSystemInfo: Sys Info size:  113
04-25 12:50:15:580 2901 OCC::SqlQuery::prepare: Sqlite prepare statement error: "disk I/O error" in "PRAGMA quick_check;"
04-25 12:50:15:580 2901 OCC::SqlDatabase::checkDb: Error running quick_check on database
[...]

PR #6023, opened 2017-09-12
Journal: Fall back to DELETE on IO error #5723

ckamm:
Also add logging of extended error codes for this IO error, maybe we can
become more specific about which situations should trigger a journal
mode switch.

See #5723


ckamm:
@guruz In my tests it did seem to switch the db to DELETE mode (the -wal temporary files were gone)

ckamm:
I agree that this is experimental - but won't affect users that don't see the disk I/O error appear.

PR #6026, opened 2017-09-13
Journal: Second attempt at journal mode fallback #5723

ckamm:


ogoffart:
Can you elaborate  (maybe in the commit message)  on what does this fix exactly?

ckamm:
@ogoffart Done. Note that this is an experimental commit since I can't reproduce the problem locally.

Issue #5830, opened 2017-06-12
Different accounts may use same keychain entry


ckamm:
Currently the client saves and looks up keychain entries by user + server url. This leads to problems when there are several accounts with the same identifier.

Specifics:
* Deleting one of the accounts will wipe the keychain entries, making the second one unusable.
* If the two accounts use different auth mechanisms (oauth2 vs http), the keychain entries get overwritten with different data.

An account should generate an additional unique identifier to disambiguate the different accounts. Care needs to be taken to make it backwards compatible.

ogoffart:
> If the two accounts use different auth mechanisms (oauth2 vs http), the keychain entries get overwritten with different data.

That's not possible,  the client uses oauth2 always if the server supports it. Even if the account was configured with basic http auth before, it will then switch to oauth2 on the next connection.

SamuAlfageme:
Related behavior for the cookies on log out/account removal: https://github.com/owncloud/client/pull/5383#issuecomment-267327203

ckamm:
@ogoffart Then there's some other bug lurking there, because I set up a new oauth2 account. Started the client again and one of them was signed out for no apparent reason.

guruz:
> An account should generate an additional unique identifier to disambiguate the different accounts. Care needs to be taken to make it backwards compatible.

Remember that this might be a quite exotic setup, so don't introduce too much complexity for this.
Maybe WONTFIX? Or whatever @ogoffart says?

SamuAlfageme:
> An account should generate an additional unique identifier to disambiguate the different accounts. 

Couldn't we simply use the id from the `.cfg` file also to reference the accounts on the keychain, etc...  (i.e. the N in `N\config_key=config_value`)

```
[Accounts]
0\Folders\1\ignoreHiddenFiles=true
0\Folders\1\journalPath=._sync_fd41dc4a0d48.db
0\Folders\1\localPath=/Users/ownCloud/
0\Folders\1\paused=false
0\Folders\1\targetPath=/
0\authType=http
0\http_oauth=false
0\http_user=admin
0\serverVersion=10.0.2.1
0\url=https://demo.owncloud.com
0\user=admin
1\Folders\2\ignoreHiddenFiles=true
1\Folders\2\journalPath=._sync_8da9d25ef3ff.db
1\Folders\2\localPath=/Users/ownCloud2/
1\Folders\2\paused=false
1\Folders\2\targetPath=/
1\authType=http
1\http_oauth=false
[...]

PR #6027, opened 2017-09-13
Credentials: Use per-account keychain entries #5830

ckamm:
This requires a lot of migration code: the old entries need to be read,
saved to the new locations and then deleted.

See #5830. It's potentially breaking, so please review carefully.

Known side effects of this patch:
* If you go back from 2.4 to 2.3, you'll have to enter the keychain data for your accounts again.
* If you have several accounts pointing to the same url+user, one of them might not have credentials set up on the first start of 2.4.

While working on this I had an idea about an alternate approach: What if we added a "what's the keychain base id" entry to the settings? Then we could add the new account identifier to new accounts and old accounts would work without losing keychain data even when switching between 2.3 and 2.4. The drawback is that we'd never migrate old accounts.

### DIFF
--- a/src/libsync/creds/httpcredentials.cpp
+++ b/src/libsync/creds/httpcredentials.cpp
@@ -103,6 +103,7 @@ static void addSettingsToJob(Account *account, QKeychain::Job *job)
 
 HttpCredentials::HttpCredentials()
     : _ready(false)
+    , _keychainMigration(false)
 {
 }
 
@@ -113,6 +114,7 @@ HttpCredentials::HttpCredentials(const QString &user, const QString &password, c
     , _ready(true)
     , _clientSslKey(key)
     , _clientSslCertificate(certificate)
+    , _keychainMigration(false)
 {
 }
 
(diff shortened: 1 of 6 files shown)


src/libsync/creds/httpcredentials.cpp
```diff
@@ -174,21 +176,43 @@ void HttpCredentials::fetchFromKeychain()
...
+    job->setKey(kck);
+    connect(job, SIGNAL(finished(QKeychain::Job *)), SLOT(slotReadClientCertPEMJobDone(QKeychain::Job *)));
+    job->start();
+}
+
+void HttpCredentials::deleteOldKeychainEntries()
+{
+    auto startDeleteJob = [this](QString user) {
+        DeletePasswordJob *job = new DeletePasswordJob(Theme::instance()->appName());
+        addSettingsToJob(_account, job);
```
ogoffart: In HttpCredentials::invalidateToken, there is a job2 to remove the password from a "deprecated location" which does not set the settings.

Should we not do the same here?   (in case of upgrading from a 1.7 client?  maybe this is anyway no longer supported)

But in any case, we can remove this job2 from HttpCredentials::invalidateToken

ckamm:
@ogoffart Done

ogoffart:
I guess the jenkins failure is unrelated

Issue #3645, opened 2015-08-18
[2.0RC1] Greyed "Add folder to Synchronize" button?


tflidd:
![screenshot](https://cloud.githubusercontent.com/assets/8111200/9342281/259bf0d0-45fb-11e5-8121-dbf5fef705e3.png)
- The "Add Folder to Synchronize" button is never active. Is it not used any more? Before I was able to sync a folder from my account to a different place on my filesystem. I can still add my account twice.

phil-davis:
AFAIK when the whole root of the user's server folder tree is already being synced to the client then there is nothing more that can be synced, so "Add Folder to Synchronize" is disabled.
If you remove the root folder sync then you should be able to add individual sync of various folders to different place on the client.

phil-davis:
I guess it is valid to support adding the same account twice to a multi-account client install. Then you can sync some combination of your files to client location A, and then in the 2nd (copy of same) account sync some different (and possibly overlapping) set of folders to a different place on the client.
I am struggling to think of a real use case for this - but I tried it just now and it works, so I also can't think of a reason to add code to prevent it.

tflidd:
> AFAIK when the whole root of the user's server folder tree is already being synced to the client then there is nothing more that can be synced, so "Add Folder to Synchronize" is disabled.
> If you remove the root folder sync then you should be able to add individual sync of various folders to different place on the client.

That what was it like in the previous version. But even if I only choose a sub-folder to sync upon account creation:
![screenshot](https://cloud.githubusercontent.com/assets/8111200/9349221/02b0d1a6-4643-11e5-8ef1-e08efd7a2734.png)
but the subfolder-structure is always shown and all single files on the root-directory will be synced (the owncloud manual):
![screenshot2](https://cloud.githubusercontent.com/assets/8111200/9349253/6c86295a-4643-11e5-9f86-6cae359702fd.png)
So it seems that it is not possible any more, to sync directly the sub-folder. Or do I need to change something during the creation of the account?

phil-davis:
See discussion of this at https://github.com/owncloud/client/issues/3632

tflidd:
Thanks for pointing me to the other discussion. But with this behavior, it is impossible to create a case where the "Add Folder to Synchronize"-button will be active. Perhaps on upgrade from a previous version.

phil-davis:
Here is an example. 
1. Create user "User2", with the default Documents and Photos folders.
[...]

PR #6032, opened 2017-09-14
Gui: Allow to add multiple sync folder connection of the same folder

ogoffart:
... or child folders

There is also no real reason to forbid the user from syncing the same
folder to multiple location on its hardrive.
A real use case is when the user uncheck a big directory using "choose
what to sync", but would still like to sync a folder within this disabled
tree. The user can now do this with the "add folder" feature

Since 2.3, we even support syncing the same local folder to multiple
remote folder, so why not allow syncing the same remote folder several
times?

Relates to issue #3645

### DIFF
--- a/src/gui/folderwizard.cpp
+++ b/src/gui/folderwizard.cpp
@@ -445,16 +445,10 @@ bool FolderWizardRemotePath::isComplete() const
         } else if (dir.startsWith(curDir + QLatin1Char('/'))) {
             warnStrings.append(tr("You are already syncing <i>%1</i>, which is a parent folder of <i>%2</i>.").arg(Utility::escape(curDir), Utility::escape(dir)));
         }
-
-        if (curDir == QLatin1String("/")) {
-            warnStrings.append(tr("You are already syncing all your files. Syncing another folder is <b>not</b> supported. "
-                                  "If you want to sync multiple folders, please remove the currently configured "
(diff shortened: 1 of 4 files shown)


src/gui/folderwizard.cpp
```diff
@@ -431,30 +431,6 @@ bool FolderWizardRemotePath::isComplete() const
...
-            warnStrings.append(tr("This folder is already being synced."));
-        } else if (dir.startsWith(curDir + QLatin1Char('/'))) {
-            warnStrings.append(tr("You are already syncing <i>%1</i>, which is a parent folder of <i>%2</i>.").arg(Utility::escape(curDir), Utility::escape(dir)));
-        }
-
-        if (curDir == QLatin1String("/")) {
-            warnStrings.append(tr("You are already syncing all your files. Syncing another folder is <b>not</b> supported. "
-                                  "If you want to sync multiple folders, please remove the currently configured "
-                                  "root folder sync."));
-        }
```
ckamm: I think it'd be nice to warn the user about redundant syncs without prohibiting them from proceeding. Maybe isComplete could still return true when these warnings are shown? (with the third warning text changed, of course)

ogoffart: Updated:  I kept the warning but still allow the creation.

PR #6033, opened 2017-09-14
csync: refactoring and removal of c_rbtree

jturcotte:


src/csync/csync_update.cpp
```diff
@@ -172,14 +171,14 @@ static int _csync_detect_update(CSYNC *ctx, std::unique_ptr<csync_file_stat_t> f
...
-    tmp = csync_statedb_get_stat_by_hash(ctx, fs->phash);
+    tmp = csync_statedb_get_stat_by_path(ctx, fs->path);
 
     if(_last_db_return_error(ctx)) {
         ctx->status_code = CSYNC_STATUS_UNSUCCESSFUL;
         return -1;
     }
 
-    if(tmp && tmp->phash == fs->phash ) { /* there is an entry in the database */
+    if(tmp && tmp->path == fs->path ) { /* there is an entry in the database */
```
ogoffart: (minor:  tmp->path == fs->path should actually be an assert)

src/common/syncjournaldb.cpp
```diff
@@ -978,37 +978,42 @@ bool SyncJournalDb::deleteFileRecord(const QString &filename, bool recursively)
 }
 
 
-SyncJournalFileRecord SyncJournalDb::getFileRecord(const QString &filename)
+bool SyncJournalDb::getFileRecord(const QString &filename, SyncJournalFileRecord *rec)
```
ogoffart: I'm not a fan of this kind of API.
Maybe do something like std::optional<SyncJournalFileRecord> instead

src/csync/csync_update.cpp
jturcotte: This is changed further by b4b0f63e6d043ceeaec5e9ebf008b94898e4bb04

src/csync/csync_update.cpp
```diff
@@ -387,66 +351,47 @@ static int _csync_detect_update(CSYNC *ctx, const char *file,
...
-  st->etag = fs->etag;
-  st->file_id = fs->file_id;
-  st->directDownloadUrl = fs->directDownloadUrl;
-  st->directDownloadCookies = fs->directDownloadCookies;
-  st->remotePerm = fs->remotePerm;
-
-  // For the remote: propagate the discovered checksum
-  if (ctx->current == REMOTE_REPLICA) {
-      st->checksumHeader = fs->checksumHeader;
-  }
```
ckamm: Self-note: Is having the stat's checksum data also in the local replica going to cause issues?

jturcotte: Could maybe an issue if it's outdated, we check it in `PropagateUploadFileCommon::slotComputeContentChecksum`.

src/csync/csync_reconcile.cpp
```diff
@@ -187,20 +168,18 @@ static int _csync_merge_algorithm_visitor(void *obj, void *data) {
...
-                        node = c_rbtree_find(tree, &h);
-                        CSYNC_LOG(CSYNC_LOG_PRIORITY_TRACE, "PHash of temporary opposite (%s): %" PRIu64 " %s",
-                                tmp->path.constData() , h, node ? "found": "not found" );
-                        if (node) {
-                            other = (csync_file_stat_t*)node->data;
+                        other_file_it = other_tree->find(tmp->path);
+                        CSYNC_LOG(CSYNC_LOG_PRIORITY_TRACE, "Temporary opposite (%s) %s",
+                                tmp->path.constData() , other_file_it != other_tree->end() ? "found": "not found" );
+                        if (other_file_it != other_tree->end()) {
+                            other = other_file_it->second.get();
```
ckamm: It may be convenient to have a function
```
find(tree, path) -> csync_file_stat_t *
```
Then we could skip dealing with iterators in most places.

src/csync/csync_reconcile.cpp
```diff
@@ -395,25 +374,27 @@ static int _csync_merge_algorithm_visitor(void *obj, void *data) {
...
       break;
     default:
       break;
   }
 
-  rc = c_rbtree_walk(tree, (void *) ctx, _csync_merge_algorithm_visitor);
-  if( rc < 0 ) {
-    ctx->status_code = CSYNC_STATUS_RECONCILE_ERROR;
+  auto end = tree->end();
+  for (auto it = tree->begin(); it != end; ++it) {
```
ckamm: Minor: possibly range-based for

src/common/syncjournaldb.cpp
jturcotte: This is how the call site looks with this API:
```
SyncJournalFileRecord rec;
if (!journal->getFileRecord("path", &rec)
    emit error();
else if (rec.isValid())
    qDebug() << rec._path;
```

This is how it would like if we passed the success as the out argument (takes one more line):
```
bool ok = false;
SyncJournalFileRecord rec = journal->getFileRecord("path", &ok);
if (!ok)
    emit error();
else if (rec.isValid())
    qDebug() << rec._path;
```

And this is how it could be with some custom return data structure (what a more specific implementation of std::optional could look like):
```
JournalResult<SyncJournalFileRecord> result = journal->getFileRecord("path");
if (!result.isSuccess)
    emit error();
else if (result.value.isValid())
    qDebug() << result.value._path;
```
But this would ideally also mean I have to rewrite most of the DB access code for things to be consistent. So that's why I chose to go with bool return value for now. Let me know if you think otherwise.

ckamm:
Please also update `doc/dev/sync-algorithm.md` when you're done! :)

src/common/syncjournaldb.cpp
jturcotte: Or we could start using exceptions 😜

src/libsync/syncfileitem.cpp
```diff
@@ -0,0 +1,71 @@
...
+    rec._fileSize = _size;
+    rec._remotePerm = _remotePerm;
+    rec._serverHasIgnoredFiles = _serverHasIgnoredFiles;
+    rec._checksumHeader = _checksumHeader;
+
+    // Go through csync vio just to get the inode.
+    csync_file_stat_t fs;
+    if (csync_vio_local_stat(localFileName.toUtf8().constData(), &fs) == 0) {
+        rec._inode = fs.inode;
+        qCDebug(lcFileItem) << localFileName << "Retrieved inode " << _inode << "(previous item inode: " << _inode << ")";
```
ckamm: I was about to complain that you have to destroy the `csync_file_stat_t`, but it's now value-like! +1000

src/common/filesystembase.h
```diff
@@ -0,0 +1,189 @@
...
+     * string. That means that it prepends a \\?\ (unless already UNC) and converts
+     * all slashes to backslashes.
+     *
+     * Note the following:
+     *  - The string must be absolute.
+     *  - it needs to contain a drive character to be a valid UNC
+     *  - A conversion is only done if the path len is larger than 245. Otherwise
+     *    the windows API functions work with the normal "unixoid" representation too.
+     *
+     *  This function allocates memory that must be freed by the caller.
```
ckamm: This is comment seems no longer correct.

Minor: I wonder if this could be two overloads that call a template defined in a cpp file.

src/common/syncjournaldb.cpp
```diff
@@ -1028,6 +1050,114 @@ bool SyncJournalDb::getFileRecord(const QString &filename, SyncJournalFileRecord
...
+        rec->_path = _getFileRecordQueryByFileId->baValue(0);
+        rec->_inode = _getFileRecordQueryByFileId->intValue(1);
+        rec->_modtime = _getFileRecordQueryByFileId->int64Value(2);
+        rec->_type = _getFileRecordQueryByFileId->intValue(3);
+        rec->_etag = _getFileRecordQueryByFileId->baValue(4);
+        rec->_fileId = _getFileRecordQueryByFileId->baValue(5);
+        rec->_remotePerm = _getFileRecordQueryByFileId->baValue(6);
+        rec->_fileSize = _getFileRecordQueryByFileId->int64Value(7);
+        rec->_serverHasIgnoredFiles = (_getFileRecordQueryByFileId->intValue(8) > 0);
+        rec->_checksumHeader = _getFileRecordQueryByFileId->baValue(9);
```
ckamm: Could this data extraction also be a common function, preferably defined close the the query string macro?

src/csync/csync_private.h
```diff
@@ -76,8 +77,8 @@ struct OCSYNC_EXPORT csync_s {
       void *update_callback_userdata = nullptr;
 
       /* hooks for checking the white list (uses the update_callback_userdata) */
-      int (*checkSelectiveSyncBlackListHook)(void*, const char*) = nullptr;
-      int (*checkSelectiveSyncNewFolderHook)(void*, const char* /* path */, const char* /* remotePerm */) = nullptr;
+      int (*checkSelectiveSyncBlackListHook)(void*, const QByteArray &) = nullptr;
```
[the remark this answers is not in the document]
jturcotte: Slightly, it goes with the change of the return value of `csync_rename_adjust_path_source` to QByteArray, ultimately because of the folder_renamed_to/from data structure changes.

ckamm:
@jturcotte I wonder about QByteArray vs QString for the path value. It appears it's now a bytearray everywhere. But don't we assume utf8 encoding? What were your thoughts on that?

src/common/syncjournaldb.h
```diff
@@ -236,6 +236,7 @@ class OCSYNC_EXPORT SyncJournalDb : public QObject
     QString _dbFile;
     QMutex _mutex; // Public functions are protected with the mutex.
     int _transaction;
+    int _dbIsEmpty;
```
ckamm: Minor: more like _metadataTableIsEmpty?

jturcotte:
> @jturcotte I wonder about QByteArray vs QString for the path value. It appears it's now a bytearray everywhere. But don't we assume utf8 encoding? What were your thoughts on that?

This basically didn't change, as before all this we had UTF-8 `char*` everywhere in csync, and UTF-16 QString everywhere else. I just replaced many `char*` by `QByteArray`, both in UTF-8. I think we should either keep csync all UTF-8, or all QString, so that we know clearly when we have to convert to/from UTF-8. This is important since the default `QString(const QByteArray &a)` silently uses `fromAscii` and if we don't have a clear layer where the conversion happens, we might lose non-ASCII data.

src/csync/csync.cpp
```diff
@@ -331,28 +274,25 @@ static int _csync_treewalk_visitor(void *obj, void *data) {
...
-    
     tw_ctx.userdata = ctx->callbacks.userdata;
     tw_ctx.user_visitor = visitor;
     tw_ctx.instruction_filter = filter;
 
     ctx->callbacks.userdata = &tw_ctx;
 
-    rc = c_rbtree_walk(tree, (void*) ctx, _csync_treewalk_visitor);
+    auto end = tree->end();
+    for (auto it = tree->begin(); it != end; ++it) {
```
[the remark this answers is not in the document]
jturcotte: done

src/csync/csync_reconcile.cpp
jturcotte: Done. Made FileMap a subclass and implemented it as `csync_file_stat_t *findFile(const QByteArray &key);`

src/csync/csync_reconcile.cpp
jturcotte: done

src/common/filesystembase.h
jturcotte: done

src/common/syncjournaldb.cpp
jturcotte: done, it indeed looks cleaner

src/common/syncjournaldb.h
jturcotte: done

jturcotte:
Rebased and removed the statedb part of the pull request in order to get this into 2.4 (again).

jturcotte:
After talking with @ogoffart, also pushed those changes manually to master:
* 5e332983e SyncJournalDB: Only reset_and_clear_bindings once
* 6e240a3e9 csync_update: Remove the check for csync_get_statedb_exists
* a1f1775d1 Move SyncJournalDB to src/common
* 5fbed0d1c Reverse the dependency between SyncJournalFileRecord and SyncFileItem
* bd2522504 Move non-csync-dependent parts of FileSystem to src/common
* 013eba0e7 Update sync-algorithm.md

Issue #6020, opened 2017-09-11
Adapt private link generation to core#28984


michaelstingl:
After https://github.com/owncloud/core/issues/28984 has been implemented on the server side, the private link generation in desktop client needs a small rework, so parameters in `config.php` like `overwritehost` are applied to the link…

---

**config/config.php**
```
  'trusted_domains' => 
  array (
    0 => 'mycloud.example.com',
    1 => '10.135.63.184',
    2 => 'overwritecloud.example.com',
  ),
  'datadirectory' => '/var/www/owncloud/data',
  'overwrite.cli.url' => 'https://overwritecloud.example.com',
  'overwritehost' => 'overwritecloud.example.com',
```

### Expected behaviour
Client should respect overwritehost parameter from config.php for private link generation, same as for public link generation. (Step 5.)

### Actual behaviour
Client ignores [overwritehost parameter from config.php](https://doc.owncloud.com/server/10.0/admin_manual/configuration/server/config_sample_php_parameters.html#proxy-configurations) for private link generation (Step 4.)

### Steps to reproduce
1. Install `2.4.0.*-nightly`
2. Restart `explorer.exe`
3. Right-click folder > ownCloud > Copy private link to clipboard  
  ==> `https://mycloud.example.com/index.php/f/00000171`
4. Right-click folder > ownCloud > Share…
5. Public Links > Create new > Copy link to clipboard  
  ==> `https://overwritecloud.example.com/index.php/s/8fJt9YgSZBOmY0o`

### Server configuration
Operating system:
Ubuntu 16.04.3 LTS

Web server: 
Apache/2.4.18 (Ubuntu)

Database:
10.0.31-MariaDB-0ubuntu0.16.04.

PHP version:
PHP 7.0.22-0ubuntu0.16.04.1

ownCloud version:
{"installed":"true","maintenance":"false","needsDbUpgrade":"false","version":"10.0.3.2","versionstring":"10.0.3 RC1","edition":"Enterprise","productname":"ownCloud"}

Storage backend (external storage):
no

### Client configuration
Client version:
[ownCloud-2.4.0.8310-nightly20170911-setup.exe](https://download.owncloud.com/desktop/daily/ownCloud-2.4.0.8310-nightly20170911-setup.exe)

Operating system:
Win 8.1, 32-bit

OS language:
EN

Qt version used by client package (Linux only, see also Settings dialog):
5.6.2

Client package (From ownCloud or distro) (Linux only):

Installation path of client:
default

[...]

PR #6037, opened 2017-09-18
Private links: Retrieve link through propfind property #6020

ckamm:
* The sharing ui does a propfind anyway: use that to query the new
property as well!
* For the socket api, asynchronously query the server for the right url
when an action that needs it is triggered.

The old, manually generated URL will be used as fallback in case the
server doesn't support the new property or the property can't be
retrieved for some reason.

Depends on owncloud/core#29021
For #6020

### DIFF
--- a/src/gui/socketapi.cpp
+++ b/src/gui/socketapi.cpp
@@ -491,23 +491,70 @@ void SocketApi::command_SHARE_MENU_TITLE(const QString &, SocketListener *listen
     listener->sendMessage(QLatin1String("SHARE_MENU_TITLE:") + tr("Share with %1", "parameter is ownCloud").arg(Theme::instance()->appNameGUI()));
 }
 
+// Fetches the private link url asynchronously and then calls the target slot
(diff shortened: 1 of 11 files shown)


src/gui/socketapi.cpp
```diff
@@ -491,23 +491,72 @@ void SocketApi::command_SHARE_MENU_TITLE(const QString &, SocketListener *listen
     listener->sendMessage(QLatin1String("SHARE_MENU_TITLE:") + tr("Share with %1", "parameter is ownCloud").arg(Theme::instance()->appNameGUI()));
 }
 
-void SocketApi::command_COPY_PRIVATE_LINK(const QString &localFile, SocketListener *)
+// Fetches the private link url asynchronously and then calls the target slot
+template <typename T>
+void fetchPrivateLinkUrl(const QString &localFile, typename QtPrivate::FunctionPointer<T>::Object *target,
```
ogoffart: Don't use QtPrivate,  this is not Qt public API and it might be changed.
Just use SocketApi* or another template type. 
(maybe no need to have a template:  `void fetchPrivateLinkUrl(const QString &localFile, SocketApi *target, void (SocketApi::*targetFunc)() ) `

PR #6039, opened 2017-09-19
RemotePermissions: Store in a class rather than in a QByteArray to save memory

ogoffart:
Create a specific type that parses the permissions so we can store
it in a short rather than in a QByteArray

(Note: since most files have the same permissions, i was also wondering if we could not try to "internate" the QByteArray  (storing it into a QSet so they can all be shared.)  But this is Good enough I think.
The other advantage is that we keep all these letters in a centralized place.

### DIFF
--- a/src/common/remotepermissions.cpp
+++ b/src/common/remotepermissions.cpp
@@ -1,0 +1,67 @@
+/*
+ * Copyright (C) by Olivier Goffart <ogoffart@woboq.com>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ */
+
+#include "remotepermissions.h"
+#include <cstring>
+
+namespace OCC {
+
+static const char letters[] = " WDNVCKRSMm";
+
+
+template <typename Char>
+void RemotePermissions::fromArray(const Char *p)
+{
+    _value = p ? notNullMask : 0;
+    if (!p)
+        return;
+    while (*p) {
+        if (auto res = std::strchr(letters, static_cast<char>(*p)))
+            _value |= (1 << (res - letters));
+        ++p;
+    }
+}
+
+RemotePermissions::RemotePermissions(const char *p)
+{
+    fromArray(p);
+}
+
+RemotePermissions::RemotePermissions(const QString &s)
+{
+    fromArray(s.isEmpty() ? nullptr : s.utf16());
+}
+
+QByteArray RemotePermissions::toString() const
+{
+    QByteArray result;
+    if (isNull())
+        return result;
+    result.reserve(PermissionsCount);
+    for (uint i = 1; i <= PermissionsCount; ++i) {
+        if (_value & (1 << i))
+            result.append(letters[i]);
+    }
+    if (result.isEmpty()) {
+        // Make sure it is not empty so we can differentiate null and empty permissions
+        result.append(' ');
+    }
+    return result;
+}
+
+} // namespace OCC
--- a/src/common/remotepermissions.h
+++ b/src/common/remotepermissions.h
@@ -1,0 +1,92 @@
+/*
+ * Copyright (C) by Olivier Goffart <ogoffart@woboq.com>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ */
+
+#pragma once
+
+#include <QString>
+#include <QMetaType>
+#include "ocsynclib.h"
+
+namespace OCC {
+
+/**
+ * Class that store in a memory efficient way the remote permission
+ */
+class OCSYNC_EXPORT RemotePermissions
+{
+private:
+    // The first bit tells if the value is set or not
+    // The remaining bits correspond to know if the value is set
+    quint16 _value = 0;
+    static constexpr int notNullMask = 0x1;
+
+    template <typename Char> // can be 'char' or 'ushort' if conversion from QString
+    void fromArray(const Char *p);
+
+public:
+    enum Permissions {
+        CanWrite = 1,             // W
+        CanDelete = 2,            // D
+        CanRename = 3,            // N
+        CanMove = 4,              // V
+        CanAddFile = 5,           // C
+        CanAddSubDirectories = 6, // K
+        CanReshare = 7,           // R
+        // Note: on the server, this means SharedWithMe, but in discoveryphase.cpp we also set
+        // this permission when the server reports the any "share-types"
+        IsShared = 8,             // S
+        IsMounted = 9,            // M
+        IsMountedSub = 10,        // m (internal: set if the parent dir has IsMounted)
+
+        // Note: when adding support for more permissions, we need to invalid the cache in the database.
+        // (by setting forceRemoteDiscovery in SyncJournalDb::checkConnect)
+        PermissionsCount = IsMountedSub
+    };
+    RemotePermissions() = default;
+    explicit RemotePermissions(const char *);
+    explicit RemotePermissions(const QString &);
+
+    QByteArray toString() const;
+    bool hasPermission(Permissions p) const
+    {
+        return _value & (1 << static_cast<int>(p));
+    }
+    void setPermission(Permissions p)
+    {
+        _value |= (1 << static_cast<int>(p)) | notNullMask;
+    }
+    void unsetPermission(Permissions p)
+    {
+        _value &= ~(1 << static_cast<int>(p));
+    }
+
+    bool isNull() const { return !(_value & notNullMask); }
+    friend bool operator==(RemotePermissions a, RemotePermissions b)
+    {
+        return a._value == b._value;
+    }
+    friend bool operator!=(RemotePermissions a, RemotePermissions b)
+    {
+        return !(a == b);
+    }
+};
+
+
+} // namespace OCC
+
+Q_DECLARE_METATYPE(OCC::RemotePermissions)
--- a/src/csync/csync_update.cpp
+++ b/src/csync/csync_update.cpp
@@ -183,10 +183,10 @@ static int _csync_detect_update(CSYNC *ctx, std::unique_ptr<csync_file_stat_t> f
       /* we have an update! */
       CSYNC_LOG(CSYNC_LOG_PRIORITY_INFO, "Database entry found, compare: %" PRId64 " <-> %" PRId64
                                           ", etag: %s <-> %s, inode: %" PRId64 " <-> %" PRId64
-                                          ", size: %" PRId64 " <-> %" PRId64 ", perms: %s <-> %s, ignore: %d",
+                                          ", size: %" PRId64 " <-> %" PRId64 ", perms: %x <-> %x, ignore: %d",
                 ((int64_t) fs->modtime), ((int64_t) tmp->modtime),
                 fs->etag.constData(), tmp->etag.constData(), (uint64_t) fs->inode, (uint64_t) tmp->inode,
-                (uint64_t) fs->size, (uint64_t) tmp->size, fs->remotePerm.constData(), tmp->remotePerm.constData(), tmp->has_ignored_files );
+                (uint64_t) fs->size, (uint64_t) tmp->size, *reinterpret_cast<short*>(&fs->remotePerm), *reinterpret_cast<short*>(&tmp->remotePerm), tmp->has_ignored_files );
       if (ctx->current == REMOTE_REPLICA && fs->etag != tmp->etag) {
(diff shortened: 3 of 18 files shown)


src/common/remotepermissions.h
```diff
@@ -0,0 +1,85 @@
...
+        CanReshare = 7,           // R
+        // Note: on the server, this means SharedWithMe, but in discoveryphase.cpp we also set
+        // this permission when the server reports the any "share-types"
+        IsShared = 8,             // S
+        IsMounted = 9,            // M
+        IsMountedSub = 10,        // m (internal: set if the parent dir has IsMounted)
+    };
+    RemotePermissions() = default;
+    RemotePermissions(const char *);
+    RemotePermissions(const QString &);
```
ckamm: explicit?

src/common/remotepermissions.h
```diff
@@ -0,0 +1,85 @@
...
+        CanRename = 3,            // N
+        CanMove = 4,              // V
+        CanAddFile = 5,           // C
+        CanAddSubDirectories = 6, // K
+        CanReshare = 7,           // R
+        // Note: on the server, this means SharedWithMe, but in discoveryphase.cpp we also set
+        // this permission when the server reports the any "share-types"
+        IsShared = 8,             // S
+        IsMounted = 9,            // M
+        IsMountedSub = 10,        // m (internal: set if the parent dir has IsMounted)
```
ckamm: Suggest to add a `PermissionsCount = IsMountedSub` here, so you can use that for `resize()` etc

src/common/remotepermissions.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+QByteArray RemotePermissions::toString()
+{
+    QByteArray result;
+    if (isNull())
+        return result;
+    result.reserve(IsMountedSub);
+    for (uint i = 0; i <= IsMountedSub; ++i) {
+        if (_value & (1 << (i + 1)))
+            result.append(letters[i]);
```
ckamm: The different ranges here are confusing. The `for` looks wrong, shouldn't it go to `< IsMountedSub` only?
I suggest you add a dummy letter to `letters` such that the letter index and enum index always line up.

src/common/remotepermissions.h
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Class that store in a memory efficient way the remote permission
+ */
+class OCSYNC_EXPORT RemotePermissions
+{
+private:
+    // The first bit tells if the value is set or not
+    // The remaining bits correspond to know if the value is set
+    short _value = 0;
```
ckamm: Minor: would prefer quint16 to make the size explicit.

src/common/remotepermissions.h
```diff
@@ -0,0 +1,85 @@
...
+    QByteArray toString();
+    bool hasPermission(Permissions p) const
+    {
+        if (isNull())
+            return p < IsShared; // when the permissions are unset, everything is allowed
+        return _value & (1 << static_cast<int>(p));
+    }
+    void setPermission(Permissions p)
+    {
+        _value |= (1 << static_cast<int>(p));
```
ckamm: Hm, so if you default construct `RemotePermissions` and do `setPermission`, `isNull` will still return true!

src/common/remotepermissions.h
ogoffart: you think?  "RDNVCKW" is indeed a valid representation of a permission.
But maybe explicit is indeed appropriate.

src/common/remotepermissions.h
ogoffart: Ok

src/common/remotepermissions.h
ogoffart: right

src/common/remotepermissions.h
ogoffart: yes., i'll fix that

src/common/remotepermissions.h
```diff
@@ -0,0 +1,85 @@
...
+    };
+    RemotePermissions() = default;
+    RemotePermissions(const char *);
+    RemotePermissions(const QString &);
+
+    QByteArray toString();
+    bool hasPermission(Permissions p) const
+    {
+        if (isNull())
+            return p < IsShared; // when the permissions are unset, everything is allowed
```
ckamm: Mmmh, this kind of logic looks wrong to me in a utility class like this. Maybe separate flag-checking from permission-granting?

ogoffart: I don't understand your objection.  It is just setting what flags are the defaults and what are not.

src/common/remotepermissions.h
```diff
@@ -0,0 +1,85 @@
...
+        return _value & (1 << static_cast<int>(p));
+    }
+    void setPermission(Permissions p)
+    {
+        _value |= (1 << static_cast<int>(p)) | 0x1;
+    }
+    bool isNull() const { return !(_value & 0x1); }
```
jturcotte: Nit: It took me a minute to understand what `0x1` means, although I'm to blame since I only saw your comment at the declaration afterward. Maybe there could be some kind of explicit `IsNotNull = 0` value in the enum, just an idea.

src/common/remotepermissions.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+namespace OCC {
+
+static const char letters[] = " WDNVCKRSMm";
+
+
+template <typename Char>
+void RemotePermissions::fromArray(const Char *p)
+{
+    _value = p ? 0x1 : 0;
```
jturcotte: Nit: This could go after the `if (!p) return;` as `_value = 0x1;`.

src/common/remotepermissions.h
```diff
@@ -0,0 +1,85 @@
...
+    friend bool operator==(RemotePermissions a, RemotePermissions b)
+    {
+        return a._value == b._value;
```
jturcotte: `0x1` is essentially equal to `0x0`, could that be an issue?

src/csync/csync_update.cpp
```diff
@@ -181,10 +181,10 @@ static int _csync_detect_update(CSYNC *ctx, std::unique_ptr<csync_file_stat_t> f
       /* we have an update! */
       CSYNC_LOG(CSYNC_LOG_PRIORITY_INFO, "Database entry found, compare: %" PRId64 " <-> %" PRId64
                                           ", etag: %s <-> %s, inode: %" PRId64 " <-> %" PRId64
-                                          ", size: %" PRId64 " <-> %" PRId64 ", perms: %s <-> %s, ignore: %d",
+                                          ", size: %" PRId64 " <-> %" PRId64 ", perms: %x <-> %x, ignore: %d",
                 ((int64_t) fs->modtime), ((int64_t) tmp->modtime),
                 fs->etag.constData(), tmp->etag.constData(), (uint64_t) fs->inode, (uint64_t) tmp->inode,
-                (uint64_t) fs->size, (uint64_t) tmp->size, fs->remotePerm.constData(), tmp->remotePerm.constData(), tmp->has_ignored_files );
+                (uint64_t) fs->size, (uint64_t) tmp->size, *reinterpret_cast<short*>(&fs->remotePerm), *reinterpret_cast<short*>(&tmp->remotePerm), tmp->has_ignored_files );
```
jturcotte: I think this could justify some kind of `internalValue()` getter...

src/libsync/discoveryphase.cpp
```diff
@@ -403,9 +401,8 @@ void DiscoverySingleDirectoryJob::directoryListingIteratedSlot(QString file, con
         }
         if (_isExternalStorage) {
             /* All the entries in a external storage have 'M' in their permission. However, for all
-               purposes in the desktop client, we only need to know about the mount points.
-               So replace the 'M' by a 'm' for every sub entries in an external storage */
-            file_stat->remotePerm.replace('M', 'm');
+               purposes in the desktop client, we only need to know about the mount points. */
+            file_stat->remotePerm.setPermission(RemotePermissions::IsMountedSub);
```
jturcotte: This means you will now write `'M'` + `'m'` in the database, is this better making them mutually exclusive?

src/common/remotepermissions.h
ckamm: Yes, it's the only thing that makes this not a pure storage object. That is what makes it stand out to me. The rest of the class is obvious: it stores a bunch of flags - and here you need to be aware of implicit behavior.

It makes `hasPermission/setPermission` somewhat asymmetric:
```
RemotePermissions p;
assert(p.hasPermission(CanRead));
assert(p.hasPermission(CanWrite));
p.setPermission(CanWrite);
assert(!p.hasPermission(CanRead)); // eh?
```

I expect having this default allows the most convenient use of this object, keep it! Adding explicit documentation would already make me happy.

jturcotte: In which case would a null permission field mean "everything is allowed"? In most of the current code it seems to be that `perms.isEmpty()` rather means "unset".

ogoffart: The thing is that it ssimplifies the code a lot as i don't have to add the isNull check.  (I don't have to do   (if perm.isNull() || perm.hasPermissions(...)

But if you all say this is so bad design, i can then change that.

jturcotte: You could possibly make the default constructor completely equivalent to `RemotePermissions("WDNVCKR")`, also returning `isNull() == true` by setting _value accordingly rather than adding logic in some of the methods.

You currently only call setPermission for stuff `>= IsShared`.

src/csync/csync_update.cpp
ogoffart: I don't really want to expose that in the API of that class.

src/libsync/discoveryphase.cpp
ogoffart: then i need to have a unsetPermission

src/csync/csync_update.cpp
jturcotte: You're basically already exposing it here, without giving the ability to the compiler, nor grep, to warn people they might be breaking this, but alright...

ogoffart:
Updated

src/common/remotepermissions.cpp
```diff
@@ -0,0 +1,67 @@
...
+    if (isNull())
+        return result;
+    result.reserve(PermissionsCount);
+    for (uint i = 1; i < PermissionsCount; ++i) {
+        if (_value & (1 << i))
+            result.append(letters[i]);
+    }
+    if (result.isEmpty()) {
+        // Make sure it is not empty so we can differentiate null and empty permissions
+        result.append(' ');
```
jturcotte: You could also remove that case and start the loop at `i = 0`, since you use that space in fromArray to read that non-empty value from the database anyway.

src/common/remotepermissions.h
ckamm: But now PermissionsCount is actually PermissionsCount+1?

ogoffart: yes,   but i do <  in the for.  Do you think it is confusing?

src/csync/csync_update.cpp
ogoffart: it is just loggin.

src/common/remotepermissions.cpp
ogoffart: No, because i don't want to add a space in the common case.

src/common/remotepermissions.h
ckamm: Somewhat, because iterating from 1 to < count sounds like you're skipping a permission when you're not.

src/common/remotepermissions.h
```diff
@@ -0,0 +1,84 @@
...
+        IsShared = 8,             // S
+        IsMounted = 9,            // M
+        IsMountedSub = 10,        // m (internal: set if the parent dir has IsMounted)
+        PermissionsCount
+    };
+    RemotePermissions() = default;
+    explicit RemotePermissions(const char *);
+    explicit RemotePermissions(const QString &);
+
+    QByteArray toString();
```
ckamm: minor: const

src/common/remotepermissions.h
```diff
@@ -0,0 +1,84 @@
...
+    bool hasPermission(Permissions p) const
+    {
+        return _value & (1 << static_cast<int>(p));
+    }
+    void setPermission(Permissions p)
+    {
+        _value |= (1 << static_cast<int>(p)) | notNullMask;
+    }
+    bool isNull() const { return !(_value & notNullMask); }
+    friend bool operator==(RemotePermissions a, RemotePermissions b)
```
ckamm: minor: I don't think `friend` does anything

src/gui/owncloudgui.cpp
```diff
@@ -1019,7 +1019,7 @@ void ownCloudGui::slotShowShareDialog(const QString &sharePath, const QString &l
     bool resharingAllowed = true; // lets assume the good
     if (fileRecord.isValid()) {
         // check the permission: Is resharing allowed?
-        if (!fileRecord._remotePerm.contains('R')) {
+        if (!fileRecord._remotePerm.isNull() && !fileRecord._remotePerm.hasPermission(RemotePermissions::CanReshare)) {
```
ckamm: Sorry to open that again, but this is a behavior change, right? Previously empty permissions would not allow resharing, and now they do?

src/libsync/discoveryphase.cpp
```diff
@@ -350,20 +349,19 @@ static std::unique_ptr<csync_file_stat_t> propertyMapToFileStat(const QMap<QStri
...
                 qWarning() << "Server returned a share type, but no permissions?";
             } else {
                 // S means shared with me.
                 // But for our purpose, we want to know if the file is shared. It does not matter
                 // if we are the owner or not.
-                // Piggy back on the persmission field 'S'
-                if (!file_stat->remotePerm.contains('S'))
-                    file_stat->remotePerm.append('S');
+                // Piggy back on the persmission field
+                file_stat->remotePerm.setPermission(RemotePermissions::IsShared);
```
ckamm: Oh, this only works because "permissions" < "share-types" (old stuff)

src/libsync/discoveryphase.cpp
ckamm: It doesn't seem to cause a problem now, but if we don't update old 'm' to also have 'M', we'll have both styles in databases now, potentially creating confusion down the line.

src/common/remotepermissions.h
ogoffart: Fixed

src/common/remotepermissions.h
ogoffart: friend operators are a common pattern to declare the operators in the class without them being member function.

src/gui/owncloudgui.cpp
ogoffart: Yes, it is a behavior change, but that is a fix.  If the server don't have permissions (old server) then we should allow resharing.

If the server give empty permissions, isNull will be false.

src/libsync/discoveryphase.cpp
ogoffart: yes, this is documented and enforced a few line earlier.

src/common/remotepermissions.h
ckamm: Wow, I had forgotten that that was how it worked. Thanks!

PR #6040, opened 2017-09-20
Linux Hardening

rockihack:
Enable gcc security features on linux/unix.
see: https://wiki.debian.org/Hardening#User_Space

Current situation:
```
$ hardening-check -v ./owncloud
./owncloud:
 Position Independent Executable: no, normal executable!
 Stack protected: no, not found!
 Fortify Source functions: no, only unprotected functions found!
    unprotected: memcpy
    unprotected: memset
    unprotected: read
 Read-only relocations: yes
 Immediate binding: no, not found!
```

With features:
```
$ hardening-check -v ./owncloud
./owncloud:
 Position Independent Executable: yes
 Stack protected: yes
 Fortify Source functions: no, only unprotected functions found!
    unprotected: memcpy
    unprotected: memset
    unprotected: read
 Read-only relocations: yes
 Immediate binding: yes
```


rockihack:
`-fstack-protector-strong` and `-D_FORTIFY_SOURCE=2` also work with mingw-gcc and apple-clang. I think both should be enabled on all platforms.

hefee:
Well on Debian, I use just 
```
export DEB_BUILD_MAINT_OPTIONS = hardening=+all
```
(https://wiki.debian.org/Hardening)

but this enables all things you are setting in your PR, so it should work. And it builds successfully for all archs:
https://buildd.debian.org/status/package.php?p=owncloud-client

rockihack:
@ogoffart What about enabling `-fstack-protector-strong` and `-D_FORTIFY_SOURCE=2` for all platforms i.e. mingw-gcc and apple-clang?

For example keepassx has both enabled for about 4 years now.
see: https://github.com/keepassx/keepassx/blob/master/CMakeLists.txt

guruz:
I can't comment on mingw-gcc and apple-clang but if it works why not? :)

ogoffart:
Now i'm getting lots of warnings:

```
clang-5.0: warning: argument unused during compilation: '-pie' [-Wunused-command-line-argument]
```

Do we really need pie when there is already pic ?

rockihack:
I thought gcc is the only officially supported compiler for owncloud on unix.
Clang wants `-pie` as a linker flag, so we need to differentiate between compilers.
This should work with clang:
```
set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-pie")
set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,-pie")
```

`pic` (libraries) and `pie` (executables) are closely related but not the same thing.
`pie` is needed to make use of **ASLR**.

ogoffart:
@rockihack  I fixed it in commit b8f7c6d  as a just  

```
set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -pie")
```

Is that not the same?

rockihack:
I don't know if that works. Normally you just pass `-pie` as a gcc compiler flag, but the only reason might be that gcc is compiler and linker in one...

PR #6041, opened 2017-09-21
Use the Qt5 connection syntax

ogoffart:
This is motivated by the fact that QMetaObject::noralizeSignature takes 7.35%
CPU of the LargeSyncBench. (Mostly from ABstractNetworkJob::setupConnections and
PropagateUploadFileV1::startNextChunk). It could be fixed by using normalized
signature in the connection statement, but i tought it was a good oportunity
to modernize the code.

One commit is done automatically,  then i started doing the the rest is manually, altough not everything has been converted.


ogoffart:
(was rebased to fix conflicts)

Issue #880, opened 2013-08-17
Investigate if csync processing local diretories can be optimized


guruz:
We're doing a ```opendir/readdir``` call but then also ```stat``` each of the contained files.

I'm guessing this is expensive.

We could optimize this. However the problem is that ```readdir``` does not return all needed information (e.g. mtime)

There is probably system calls that allow this though, for example ```FindFile``` on Windows (
http://msdn.microsoft.com/en-us/library/aa364418%28VS.85%29.aspx) or ```fts``` on Linux (http://linux.die.net/man/3/fts).

Should investigate if this is faster.

The relevant code is in ```csync_ftw``` in ```csync_update.c```

guruz:
@ckamm Are you interested in taking this over from me? For Windows using ```FindFirstFileW``` in ```csync_vio_local_opendir``` instead of the Windows POSIX stuff?

guruz:
Another idea instead of using ```FindFirstFileW``` directly In the same go this should fix https://github.com/owncloud/client/issues/60 via https://github.com/owncloud/client/issues/60#issuecomment-70245354

guruz:
Update: @dragotin has switched the Windows code to use ```FindFirstFileW``` for 2.0. I'll try to do some optimizations based on this.

dragotin:
This is done.

PR #6042, opened 2017-09-21
csync_vio_local_stat: Try to avoid expensive call to c_utf8_path_to_locale

ogoffart:
Keep the value stored in the original_path
Saves 15% of the CPU time for csync_update

### DIFF
--- a/src/csync/vio/csync_vio_local_unix.cpp
+++ b/src/csync/vio/csync_vio_local_unix.cpp
@@ -45,6 +45,8 @@ typedef struct dhandle_s {
   char *path;
 } dhandle_t;
 
+static int _csync_vio_local_stat_mb(const mbchar_t *wuri, csync_file_stat_t *buf);
+
 csync_vio_handle_t *csync_vio_local_opendir(const char *name) {
   dhandle_t *handle = NULL;
   mbchar_t *dirname = NULL;
@@ -84,7 +86,6 @@ int csync_vio_local_closedir(csync_vio_handle_t *dhandle) {
   return rc;
 }
 
-
 std::unique_ptr<csync_file_stat_t> csync_vio_local_readdir(csync_vio_handle_t *dhandle) {
 
   dhandle_t *handle = NULL;
@@ -93,17 +94,17 @@ std::unique_ptr<csync_file_stat_t> csync_vio_local_readdir(csync_vio_handle_t *d
   struct _tdirent *dirent = NULL;
   std::unique_ptr<csync_file_stat_t> file_stat;
 
-  dirent = _treaddir(handle->dh);
-  if (dirent == NULL) {
-      return {};
-  }
+  do {
+      dirent = _treaddir(handle->dh);
+      if (dirent == NULL)
+          return {};
+  } while (qstrcmp(dirent->d_name, ".") == 0 || qstrcmp(dirent->d_name, "..") == 0);
 
   file_stat.reset(new csync_file_stat_t);
   file_stat->path = c_utf8_from_locale(dirent->d_name);
+  QByteArray fullPath = QByteArray() % const_cast<const char *>(handle->path) % '/' % QByteArray() % const_cast<const char *>(dirent->d_name);
   if (file_stat->path.isNull()) {
-      file_stat->original_path = handle->path;
-      file_stat->original_path += '/';
-      file_stat->original_path += dirent->d_name;
+      file_stat->original_path = fullPath;
       CSYNC_LOG(CSYNC_LOG_PRIORITY_WARN, "Invalid characters in file/directory name, please rename: \"%s\" (%s)",
                 dirent->d_name, handle->path);
   }
@@ -129,23 +130,35 @@ std::unique_ptr<csync_file_stat_t> csync_vio_local_readdir(csync_vio_handle_t *d
   }
 #endif
 
+  if (file_stat->path.isNull())
+      return file_stat;
+
+  if (_csync_vio_local_stat_mb(fullPath.constData(), file_stat.get()) < 0) {
+      // Will get excluded by _csync_detect_update.
+      file_stat->type = CSYNC_FTW_TYPE_SKIP;
+  }
   return file_stat;
 }
 
 
-int csync_vio_local_stat(const char *uri, csync_file_stat_t *buf) {
-  csync_stat_t sb;
-
-  mbchar_t *wuri = c_utf8_path_to_locale( uri );
-
-  if( _tstat(wuri, &sb) < 0) {
+int csync_vio_local_stat(const char *uri, csync_file_stat_t *buf)
+{
+    mbchar_t *wuri = c_utf8_path_to_locale(uri);
+    *buf = csync_file_stat_t();
+    int rc = _csync_vio_local_stat_mb(wuri, buf);
     c_free_locale_string(wuri);
-    return -1;
-  }
+    return rc;
+}
 
-  *buf = csync_file_stat_t();
+static int _csync_vio_local_stat_mb(const mbchar_t *wuri, csync_file_stat_t *buf)
+{
+    csync_stat_t sb;
 
-  switch(sb.st_mode & S_IFMT) {
+    if (_tstat(wuri, &sb) < 0) {
+        return -1;
+    }
+
+    switch (sb.st_mode & S_IFMT) {
     case S_IFDIR:
       buf->type = CSYNC_FTW_TYPE_DIR;
       break;
@@ -170,7 +183,5 @@ int csync_vio_local_stat(const char *uri, csync_file_stat_t *buf) {
   buf->inode = sb.st_ino;
   buf->modtime = sb.st_mtime;
   buf->size = sb.st_size;
-
-  c_free_locale_string(wuri);
   return 0;
 }
--- a/src/csync/vio/csync_vio_local_win.cpp
+++ b/src/csync/vio/csync_vio_local_win.cpp
@@ -45,8 +45,10 @@ typedef struct dhandle_s {
   WIN32_FIND_DATA ffd;
   HANDLE hFind;
   int firstFind;
-  char *path;
+  mbchar_t *path; // Always ends with '\'
 } dhandle_t;
+
+static int _csync_vio_local_stat_mb(const mbchar_t *uri, csync_file_stat_t *buf);
 
 csync_vio_handle_t *csync_vio_local_opendir(const char *name) {
   dhandle_t *handle = NULL;
@@ -73,6 +75,7 @@ csync_vio_handle_t *csync_vio_local_opendir(const char *name) {
   }
 
   if (!dirname || handle->hFind == INVALID_HANDLE_VALUE) {
+      c_free_locale_string(dirname);
       int retcode = GetLastError();
       if( retcode == ERROR_FILE_NOT_FOUND ) {
           errno = ENOENT;
@@ -85,8 +88,8 @@ csync_vio_handle_t *csync_vio_local_opendir(const char *name) {
 
   handle->firstFind = 1; // Set a flag that there first fileinfo is available.
 
-  handle->path = c_strdup(name);
-  c_free_locale_string(dirname);
+  dirname[std::wcslen(dirname) - 1] = L'\0'; // remove the *
+  handle->path = dirname;
 
   return (csync_vio_handle_t *) handle;
 }
@@ -109,7 +112,7 @@ int csync_vio_local_closedir(csync_vio_handle_t *dhandle) {
       errno = EBADF;
   }
 
-  SAFE_FREE(handle->path);
+  c_free_locale_string(handle->path);
   SAFE_FREE(handle);
 
   return rc;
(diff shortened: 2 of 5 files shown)


src/csync/vio/csync_vio_local_unix.cpp
```diff
@@ -99,13 +99,13 @@ std::unique_ptr<csync_file_stat_t> csync_vio_local_readdir(csync_vio_handle_t *d
...
   file_stat.reset(new csync_file_stat_t);
-  file_stat->path = c_utf8_from_locale(dirent->d_name);
-  if (file_stat->path.isNull()) {
-      file_stat->original_path = handle->path;
-      file_stat->original_path += '/';
-      file_stat->original_path += dirent->d_name;
-      CSYNC_LOG(CSYNC_LOG_PRIORITY_WARN, "Invalid characters in file/directory name, please rename: \"%s\" (%s)",
+  if (*dirent->d_name) {
+    file_stat->path = c_utf8_from_locale(dirent->d_name);
+    file_stat->original_path = QByteArray() % const_cast<const char *>(handle->path) % '/' % const_cast<const char *>(dirent->d_name);
```
jturcotte: This keeps more memory around (increases the peak) and makes unix behave differently than Windows.

Do we need to separate `stat` from `readdir` for the local tree (and we don't stat at all for the remote tree). Could we merge them in one call instead (or make `readdir` call `stat` itself with the right `mbchar_t`)? This would make Windows also benefit from the improvement, and prevent reading from `buf` which was write-only until now.

ogoffart:
we could merge the readdir and the stat, yes

src/csync/vio/csync_vio_local_unix.cpp
guruz: > Do we need to separate stat from readdir for the local tree (since we don't stat at all for the remote tree)? Could we merge them in one call instead (or make readdir call stat itself with the right mbchar_t)? 


Related discussion: https://github.com/owncloud/client/issues/880

Would it make sense to switch directly to ```QDirIterator```?

ogoffart:
The updated commit is completely different.  It merges the readdir and the stat.

src/csync/vio/csync_vio_local_win.cpp
```diff
@@ -158,8 +161,12 @@ std::unique_ptr<csync_file_stat_t> csync_vio_local_readdir(csync_vio_handle_t *d
           return nullptr;
       }
   }
+  auto path = c_utf8_from_locale(handle->ffd.cFileName);
+  if (path == "." || path == "..")
```
jturcotte: You could change that check in `csync_ftw` to an assert now that it's taken care of here. Or will the server send us such file entries?

src/csync/vio/csync_vio_local_unix.cpp
```diff
@@ -129,22 +130,33 @@ std::unique_ptr<csync_file_stat_t> csync_vio_local_readdir(csync_vio_handle_t *d
   }
 #endif
 
-  return file_stat;
+    if (file_stat->path.isNull())
+        return file_stat;
```
ckamm: This can't happen, can it?

src/csync/vio/csync_vio_local_unix.cpp
```diff
@@ -93,17 +94,17 @@ std::unique_ptr<csync_file_stat_t> csync_vio_local_readdir(csync_vio_handle_t *d
   struct _tdirent *dirent = NULL;
   std::unique_ptr<csync_file_stat_t> file_stat;
 
-  dirent = _treaddir(handle->dh);
-  if (dirent == NULL) {
+  do {
+    dirent = _treaddir(handle->dh);
+    if (dirent == NULL)
```
jturcotte: Maybe we should re-indent everything to 4 spaces in csync one day...

src/csync/vio/csync_vio_local_unix.cpp
```diff
@@ -129,22 +130,33 @@ std::unique_ptr<csync_file_stat_t> csync_vio_local_readdir(csync_vio_handle_t *d
...
+    return file_stat;
 }
 
 
-int csync_vio_local_stat(const char *uri, csync_file_stat_t *buf) {
-  csync_stat_t sb;
+int csync_vio_local_stat(const char *uri, csync_file_stat_t *buf)
+{
+    mbchar_t *wuri = c_utf8_path_to_locale( uri );
+      *buf = csync_file_stat_t();
```
ckamm: odd indent

src/csync/vio/csync_vio_local_unix.cpp
```diff
@@ -129,22 +130,33 @@ std::unique_ptr<csync_file_stat_t> csync_vio_local_readdir(csync_vio_handle_t *d
...
+    int rc = _csync_vio_local_stat_mb(wuri, buf);
+    c_free_locale_string(wuri);
+    return rc;
+}
 
-  mbchar_t *wuri = c_utf8_path_to_locale( uri );
+static int _csync_vio_local_stat_mb(const mbchar_t *wuri, csync_file_stat_t *buf)
```
ckamm: The fact that this doesn't wipe buf but only augment it was surprising to me

src/csync/vio/csync_vio_local_unix.cpp
```diff
@@ -129,22 +130,33 @@ std::unique_ptr<csync_file_stat_t> csync_vio_local_readdir(csync_vio_handle_t *d
   }
 #endif
 
-  return file_stat;
+    if (file_stat->path.isNull())
+        return file_stat;
+
+    if(_csync_vio_local_stat_mb(fullPath.constData(), file_stat.get()) < 0) {
+        file_stat->type = CSYNC_FTW_TYPE_SKIP;
```
jturcotte: There was a log in `csync_vio_stat` in that case that could be useful, maybe it could be added  in `_csync_vio_local_stat_mb`.
Also the comment `// Will get excluded by _csync_detect_update.` from `csync_ftw ` could be preserved/rewritten here.

src/csync/vio/csync_vio_local_unix.cpp
jturcotte: Indentation

src/csync/vio/csync_vio_local_win.cpp
```diff
@@ -85,8 +88,8 @@ csync_vio_handle_t *csync_vio_local_opendir(const char *name) {
 
   handle->firstFind = 1; // Set a flag that there first fileinfo is available.
 
-  handle->path = c_strdup(name);
-  c_free_locale_string(dirname);
+  dirname[std::wcslen(dirname)-1] = L'\0'; // remove the *
```
ckamm: Previously the path didn't have a trailing /, it seems

src/csync/vio/csync_vio_local_win.cpp
```diff
@@ -189,12 +196,29 @@ std::unique_ptr<csync_file_stat_t> csync_vio_local_readdir(csync_vio_handle_t *d
 
     file_stat->size = (handle->ffd.nFileSizeHigh * ((int64_t)(MAXDWORD)+1)) + handle->ffd.nFileSizeLow;
     file_stat->modtime = FileTimeToUnixTime(&handle->ffd.ftLastWriteTime, &rem);
+
+    std::wstring fullPath;
+    fullPath.reserve(std::wcslen(handle->path) + std::wcslen(handle->ffd.cFileName));
+    fullPath += handle->path;
```
ckamm: Ok, but you depend on the trailing / here -- can you add this as a comment to the `path` member? (I haven't checked whether it is used elsewhere)

src/csync/vio/csync_vio_local_win.cpp
```diff
@@ -189,12 +196,29 @@ std::unique_ptr<csync_file_stat_t> csync_vio_local_readdir(csync_vio_handle_t *d
...
+    }
+
     return file_stat;
 }
 
 
+int csync_vio_local_stat(const char *uri, csync_file_stat_t *buf)
+{
+    mbchar_t *wuri = c_utf8_path_to_locale( uri );
+    int rc = _csync_vio_local_stat_mb(wuri, buf);
```
ckamm: The linux version does ` *buf = csync_file_stat_t();` beforehand

ckamm:
@ogoffart Some inline comments.

src/csync/vio/csync_vio_local_win.cpp
```diff
@@ -189,12 +196,29 @@ std::unique_ptr<csync_file_stat_t> csync_vio_local_readdir(csync_vio_handle_t *d
 
     file_stat->size = (handle->ffd.nFileSizeHigh * ((int64_t)(MAXDWORD)+1)) + handle->ffd.nFileSizeLow;
     file_stat->modtime = FileTimeToUnixTime(&handle->ffd.ftLastWriteTime, &rem);
+
+    std::wstring fullPath;
+    fullPath.reserve(std::wcslen(handle->path) + std::wcslen(handle->ffd.cFileName));
+    fullPath += handle->path;
+    fullPath += handle->ffd.cFileName;
```
jturcotte: A backslash will never be needed here?
Also wstring should be UTF-16, so I think you could also use `QString`.

src/csync/vio/csync_vio_local_win.cpp
```diff
@@ -158,8 +161,12 @@ std::unique_ptr<csync_file_stat_t> csync_vio_local_readdir(csync_vio_handle_t *d
           return nullptr;
       }
   }
+  auto path = c_utf8_from_locale(handle->ffd.cFileName);
```
jturcotte: I just thought this was a leak, maybe I should have renamed it to `c_utf8_bytearray_from_locale` when I changed the return type.

src/csync/vio/csync_vio_local_unix.cpp
ogoffart: It can, if the conversion from the locale failed (invalid utf-8 for example)

src/csync/vio/csync_vio_local_win.cpp
ogoffart: correct, but it was not used so i allowed myself to change the behavior there.

src/csync/vio/csync_vio_local_win.cpp
ogoffart: I don't think the server send .. or . ,  but maybe it does.

src/csync/vio/csync_vio_local_win.cpp
ogoffart: i'll add a comment

src/csync/vio/csync_vio_local_win.cpp
ogoffart: I could use QString but then i need to reinterpret_cast to wchar_t.

I don't understand your comment about backslash.  handle->path; always end with a slash, by contruction. I'll add a comment

src/csync/vio/csync_vio_local_win.cpp
ogoffart: I did not change the behaivor.

src/csync/vio/csync_vio_local_unix.cpp
jturcotte: Humm, this only seems to be possible on Windows if `WideCharToMultiByte` fails in `c_utf8_from_locale`. Maybe the `QTextDecoder::hasFailure()` code path on unix should also return a null QByteArray? Else the rest of the `original_path` codepaths probably won't be used properly, or maybe we should remove `original_path` also on Windows to follow the unix implementation comment:

> Conversion error: since we can't report error from this function, just return the original string.  We take care of invalid utf-8 in SyncEngine::treewalkFile

ogoffart: yes that's right, but that's outside of the scope of this patch.
I've tried to keep the old behavior in this respect.
Ideally, we don't need original_name at all if we keep the wrong encoding in the "path".

src/csync/vio/csync_vio_local_win.cpp
```diff
@@ -189,12 +196,30 @@ std::unique_ptr<csync_file_stat_t> csync_vio_local_readdir(csync_vio_handle_t *d
 
     file_stat->size = (handle->ffd.nFileSizeHigh * ((int64_t)(MAXDWORD)+1)) + handle->ffd.nFileSizeLow;
     file_stat->modtime = FileTimeToUnixTime(&handle->ffd.ftLastWriteTime, &rem);
+
+    std::wstring fullPath;
+    fullPath.reserve(std::wcslen(handle->path) + std::wcslen(handle->ffd.cFileName));
+    fullPath += handle->path; // path always ends with '/', by construction
```
jturcotte: That would be a `\` and not a `/`.

ogoffart: I was wondering the same.

Line 65 do `strncat(h, "/*", 2);`
So it is a `/`.   I tested and it works like this.
Should I change it to a `\`  when i remove the '*' earlier ?

jturcotte: From IRC discussion: No I just mean the comment, dirname went through c_utf8_path_to_locale, where / is converted to \\. So the comment lower should say // path always ends with '\\', by construction

src/csync/vio/csync_vio_local_win.cpp
ckamm: The interface being different between linux and windows variants of this function should be a bug. Either clearing `*buf` is pointless, or this function incorrectly doesn't do it. But yes, it's independent of your change.

ogoffart:
For reference, this is how the CPU is divided now that this is merged for csync_update with many local files:

spectacle wy9730

jturcotte:
How does IO plays into this chart? A lot of user time should normally be spent doing file stats and querying the DB.

ogoffart:
this is only measuring CPU usage in userspace.

PR #6043, opened 2017-09-21
csync_update/reconcile: replace CSYNC_LOG with qDebug/qInfo

ogoffart:
CSYNC_LOG has quite an overhead as it uses vsprintf and it might
be run even if the category is disabled


ogoffart:
updated

Issue #5957, opened 2017-08-14
master/edge: Linux builds would break without intrusive patching


guruz:
https://build.opensuse.org/public/build/isv:ownCloud:testpilot:edge/xUbuntu_16.10/i586/testpilotcloud-client-overlays/_log

I don't see any obvious changes in the client sources so it must have been from the OBS templates? Or the underlying OBS Qt or?

```
[  112s] CMake Error at cmake/modules/QtVersionAbstraction.cmake:8 (find_package):
[  112s]   By not providing "FindQt5Core.cmake" in CMAKE_MODULE_PATH this project has
[  112s]   asked CMake to find a package configuration file provided by "Qt5Core", but
[  112s]   CMake did not find one.
[  112s] 
[  112s]   Could not find a package configuration file provided by "Qt5Core" with any
[  112s]   of the following names:
[  112s] 
[  112s]     Qt5CoreConfig.cmake
[  112s]     qt5core-config.cmake
[  112s] 
[  112s]   Add the installation prefix of "Qt5Core" to CMAKE_PREFIX_PATH or set
[  112s]   "Qt5Core_DIR" to a directory containing one of the above files.  If
[  112s]   "Qt5Core" provides a separate development package or SDK, be sure it has
[  112s]   been installed.
[  112s] Call Stack (most recent call first):
```

Detected first as QtWebKit problem in https://github.com/owncloud/client/issues/5945#issuecomment-321766059

guruz:
@jnweiger Since when does this fail btw? Is it indirectly related to your NSIS changes? (even though it's Linux not Windows)

jnweiger:
I don't see any changes on my side. The relevant files debian.control, *.dsc, *.spec are identical in the 2.3.3 and 2.4.0 templates. 

Hard to tell since when it is broken. OBS does not archive any build logs. Manual bisecting needed to find out.

This is where the log between nightly and edge differs:
nightly:
```
[   93s] -- Build of crashreporter disabled.
[   93s] 5a9286a68ce09d80f603e552a528d57f3b853bef
[   93s] -- GIT_SHA1 5a9286a68ce09d80f603e552a528d57f3b853bef
[   93s] CMake Warning at shell_integration/CMakeLists.txt:16 (find_package):
[   93s]   By not providing "FindKF5.cmake" in CMAKE_MODULE_PATH this project has
```

edge:
```
[  130s] -- Build of crashreporter disabled.
[  130s] 20a979173f648a134d97497165a71130861c9066
[...]

PR #6045, opened 2017-09-22
Upstream packaging.diff

ckamm:
Please review and test. The admin removal looks dubious to me! See #5957


CMakeLists.txt
```diff
@@ -243,7 +267,7 @@ configure_file(version.h.in ${CMAKE_CURRENT_BINARY_DIR}/version.h)
 if(BUILD_OWNCLOUD_OSX_BUNDLE)
     install(FILES sync-exclude.lst DESTINATION ${OWNCLOUD_OSX_BUNDLE}/Contents/Resources/)
     configure_file(sync-exclude.lst bin/${OWNCLOUD_OSX_BUNDLE}/Contents/Resources/sync-exclude.lst COPYONLY)
-else()
+elseif(BUILD_CLIENT)
```
dschmidt: This is nonsense - don't know how it got into the diff

CMakeLists.txt
```diff
@@ -247,7 +247,6 @@ if(BUILD_CLIENT)
     if(NOT BUILD_LIBRARIES_ONLY)
         add_subdirectory(doc)
         add_subdirectory(doc/dev)
-        add_subdirectory(admin)
```
dschmidt: Was this really in the diff? Looks like an accident or something - it looks pretty wrong, don't even know what it would be good for in the linux packages

dschmidt:
I took the liberty of adding the RPATH change to your branch/PR, hope that was ok :trollface:

src/libsync/theme.cpp
```diff
@@ -505,5 +505,20 @@ QString Theme::oauthClientSecret() const
...
+    QString helpText;
+    QTextStream stream(&helpText);
+    stream << appName().toLatin1().constData()
+           << QLatin1String(" version ")
+           << version().toLatin1().constData() << endl;
+#ifdef GIT_SHA1
+    stream << "Git revision " << GIT_SHA1 << endl;
+#endif
+    stream << "Using Qt " << qVersion() << ", built against Qt " << QT_VERSION_STR << endl;
+    stream << "Using '" << QSslSocket::sslLibraryVersionString() << "'" << endl;
```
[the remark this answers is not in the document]
ogoffart: I checked,  it should not crash.

src/cmd/cmd.cpp
```diff
@@ -196,8 +196,7 @@ void help()
 
 void showVersion()
 {
-    const char *binaryName = APPLICATION_EXECUTABLE "cmd";
-    std::cout << binaryName << " version " << qPrintable(Theme::instance()->version()) << std::endl;
+    std::cout << qPrintable(Theme::instance()->versionSwitchOutput());
```
ogoffart: -1:   Encoding!  should be qPrintableUtf8
(the application executable might contains non-ascii)

cmake/modules/DefineCompilerFlags.cmake
```diff
@@ -16,7 +16,8 @@ if (${CMAKE_C_COMPILER_ID} MATCHES "(GNU|Clang)")
     # cannot be pedantic with sqlite3 directly linked
     # FIXME Can we somehow not use those flags for sqlite3.* but use them for the rest of csync?
     if (NOT USE_OUR_OWN_SQLITE3)
-        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu99 -pedantic -pedantic-errors")
+        # -pedantic-errors explodes in src/csync/csync_private.h:166
```
ogoffart: csync_private.h:166   is a comment for me.
So I guess this comment is already outdated and should be removed.

Maybe pedentic-errors would make sens on a CI, but not when building packages, so it is fine to remove it.

CMakeLists.txt
```diff
@@ -162,28 +177,32 @@ if(APPLE)
   set( SOCKETAPI_TEAM_IDENTIFIER_PREFIX "" CACHE STRING "SocketApi prefix (including a following dot) that must match the codesign key's TeamIdentifier/Organizational Unit" )
 endif()
 
-if(APPLE)
-  find_package(Sparkle)
-endif(APPLE)
+if(BUILD_CLIENT)
+    find_package(OpenSSL 1.0.0 REQUIRED)	# still needed?
```
ogoffart: -1:  Not needed

ckamm:
@dschmidt Sure :)

src/libsync/theme.cpp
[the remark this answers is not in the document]
dschmidt: What's the output then?

cmake/modules/DefineCompilerFlags.cmake
dschmidt: Ooh of course, it was early, when I wrote the comment ... 😄 
Yeah, let's just remove it.

CMakeLists.txt
ckamm: @dschmidt Why is it nonsense? Shouldn't the whole preparation for sync-exclude.lst be done for BUILD_CLIENT only?

ckamm:
@ogoffart @dschmidt I did the `qPrintableUtf8` change, removed the compile flags change, removed the unneeded `find_package(OpenSSL)`.

Open issue: Is the `admin` directory intentionally not traversed in cmakelists? @guruz maybe?

guruz:
> Open issue: Is the admin directory intentionally not traversed in cmakelists? @guruz maybe?

👎  

Indeed, because for macOS it contains important stuff like ```macdeployqt.py```. So it needs to be there actually.

The packaging.diff was applied on Linux only.

@jnweiger opinion?

guruz:
FYI @hefee @dragotin

guruz:
> But removing the admin folder via de8ab35 is probably not needed

ack

> But removing the admin folder via de8ab35 is probably not needed. I suggest change clean_tarball.sh to not remove the admin folder.

clean_tarball.sh was only intended for external linux packagers, not for us doing all kind of stuff

> I cannot currently run a test with the nightlies, looks like part of the changes are already in master..

Wait for this PR to get merged and then see that packaging.diff should not be needed anymore?

ckamm:
@jnweiger @guruz I've removed the commit that skipped the admin/ directory. I'll merge this now. Let's test and get this rolling :)

src/libsync/theme.cpp
```diff
@@ -505,5 +505,20 @@ QString Theme::oauthClientSecret() const
     return "UBntmLjC2yYCeHwsyj73Uwo9TAaecAetRwMw0xYcvNL9yRdLSUi0hUAHfvCHFeFh";
 }
 
+QString Theme::versionSwitchOutput() const
+{
+    QString helpText;
+    QTextStream stream(&helpText);
+    stream << appName().toLatin1().constData()
```
ogoffart: Ooops!  Encoding  (not new in this patch)

src/libsync/theme.cpp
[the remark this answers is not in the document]
ogoffart: nothing:  https://code.woboq.org/qt5/qtbase/src/network/ssl/qsslsocket_openssl11.cpp.html#180

arrguably, it is a bit strange to have an empty Using,  but this should not happen in practice.

PR #6046, opened 2017-09-22
Only use $(MAKE) in doc target if make is used as generator

dschmidt:
The old behavior broke builds with other CMake generators like Ninja as it obviously has no concept of $(MAKE)

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -39,7 +39,13 @@ if(SPHINX_FOUND)
 				     -D latex_logo=${LATEX_LOGO}
 				     ${CMAKE_CURRENT_SOURCE_DIR}
 				     ${SPHINX_PDF_DIR} )
-        add_custom_target(doc-pdf $(MAKE) -C ${SPHINX_PDF_DIR} all-pdf
+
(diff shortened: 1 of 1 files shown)


doc/CMakeLists.txt
```diff
@@ -39,7 +39,13 @@ if(SPHINX_FOUND)
 				     -D latex_logo=${LATEX_LOGO}
 				     ${CMAKE_CURRENT_SOURCE_DIR}
 				     ${SPHINX_PDF_DIR} )
-        add_custom_target(doc-pdf $(MAKE) -C ${SPHINX_PDF_DIR} all-pdf
+
+        set(MAKE "make" CACHE FILEPATH "make to be used for documentation generation if not using make as generator anyway")
+        if(CMAKE_GENERATOR STREQUAL "Unix Makefiles")
```
jturcotte: Maybe `CMAKE_GENERATOR MATCHES "Makefiles"`?
https://cmake.org/Wiki/CMake_Generator_Specific_Information#Makefile_generators

dschmidt: Sounds reasonable, yeah.

dschmidt:
@jturcotte Updated.

dschmidt:
cool, thanks :)

Issue #6044, opened 2017-09-22
"re-login" adds / to serverpath, making request token link wrong


Emil-G:
### Expected behaviour
When a user is for some reason logged out of ownCloud the login to existing account popup should use the correct server link. 

### Actual behaviour
For first time setup, adding the server url and then pressing "request token" takes the user to the correct link. But when the account is already added the request token link is wrong because the application adds an extra "/" to the path.

### Steps to reproduce
1. Login using token to server url : https://mysite.com/app/
2. Log out of client from "account -> log out" . 
3. Try to log back in on the already configured account using token. ("account -> log in" )
4. Pressing "request token" takes the user to : https://mysite.com/app//index.php/settings/personal?sectionid=security#apppasswords
5. User gets HTTP 500 error as the link now contains two // after app instead of one. 

### Server configuration
Operating system:
Red Hat
Web server: 
Apache
ownCloud version:
10.0.X

### Client configuration
Client version:
2.3.3 
Operating system:
Windows 7
OS language:
English

ogoffart:
Fix: https://github.com/owncloud/client/pull/6047

ogoffart:
btw, what is the exact version of the server?

Emil-G:
10.0.2

michaelstingl:
@Emil-G You could check with the [latest 2.4.0 nightlies](https://download.owncloud.com/desktop/daily/?C=M;O=D).

00008230

Emil-G:
@michaelstingl hi, downloaded _testpilotcloud-2.4.0.8408-nightly20170925-setup.exe_ to test. 

It resolves the issue, but it adds one you should be aware of. Our normal branded client is www.mysite.com/app/ <- see the "/" at the end in ownBrander. When using the testpilot version, it seems to add a / on its own during initial login, making the "correct url" fail, so that I had to input www.mysite.com/app <- no "/" at the end to connect to server. 

Cheers,
Emil

michaelstingl:
@ogoffart this is how @Emil-G 's `mysite-theme.h` looks like

```
    QString defaultServerFolder() const { return  QString::fromUtf8("/"); }
    QString quotaBaseFolder() const { return QString::fromUtf8("/"); }
    QString forceConfigAuthType() const { return QLatin1String("http"); }
      QString overrideServerUrl() const { return QLatin1String("https://mysite.com/app/"); }
[...]

PR #6047, opened 2017-09-22
Fix double slash in URL

ogoffart:
We need to use concatPath to avoid possible double '/' in the URLs if the
account url() ends with '/'.

This has become even more of a problem since commit
d1b8370a4ad21c741da507f64a5dbfe82a3fad05 which was resolving the url after
a redirect where most server actually add a '/' if the url is a folder

CC/ @DeepDiver1975

### DIFF
--- a/src/gui/creds/httpcredentialsgui.cpp
+++ b/src/gui/creds/httpcredentialsgui.cpp
@@ -148,7 +148,10 @@ QString HttpCredentialsGui::requestAppPasswordText(const Account *account)
         return QString();
     }
 
+    auto baseUrl = account->url().toString();
+    if (baseUrl.endsWith('/'))
+        baseUrl.chop(1);
     return tr("<a href=\"%1\">Click here</a> to request an app password from the web interface.")
-        .arg(account->url().toString() + path);
+        .arg(baseUrl + path);
 }
 } // namespace OCC
--- a/src/gui/creds/oauth.cpp
+++ b/src/gui/creds/oauth.cpp
@@ -76,7 +76,7 @@ void OAuth::start()
 
                 QString code = rx.cap(1); // The 'code' is the first capture of the regexp
 
-                QUrl requestToken(_account->url().toString() + QLatin1String("/index.php/apps/oauth2/api/v1/token"));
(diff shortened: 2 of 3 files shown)


ogoffart:
This is actually issue #6044

ogoffart:
Note  to @DeepDiver1975 : double slashes only causes 500 internal server errors since recently.   for example, owncloud 9.1.6 did not have the problem (proof: https://cloud.owncloud.com//index.php/login )  ,  while 10.0.2 has  the problem 
 (this is 500 https://hydro.woboq.im/owncloud//index.php/login )

src/gui/creds/httpcredentialsgui.cpp
```diff
@@ -148,7 +148,10 @@ QString HttpCredentialsGui::requestAppPasswordText(const Account *account)
         return QString();
     }
 
+    auto baseUrl = account->url().toString();
+    if (baseUrl.endsWith('/'))
+        baseUrl.chop(1);
     return tr("<a href=\"%1\">Click here</a> to request an app password from the web interface.")
-        .arg(account->url().toString() + path);
+        .arg(baseUrl + path);
```
ckamm: Why not concatUrlPath here?

ogoffart: because path contains a query and a anchor, that vary depending on the version, so that would make the code way more complicated.

DeepDiver1975:
> Note to @DeepDiver1975 : double slashes only causes 500 internal server errors since recently. for example, owncloud 9.1.6 did not have the problem (proof: https://cloud.owncloud.com//index.php/login ) , while 10.0.2 has the problem
> (this is 500 https://hydro.woboq.im/owncloud//index.php/login )

related to subfolder afaik ... will research

DeepDiver1975:
hmmm .... maybe this is correct behavior. A url with one / is not identical to a url with two //.
So if we throw back an error this is correct. Or am I missing anything?

ogoffart:
It is just that it used to work.  But If you say that it does not need to work, then it's fine by me.

PR #6048, opened 2017-09-23
Updater: Rudimentary support for beta channel

guruz:
So we can get people update from our alpha to the beta and rc


dschmidt:
Why don't you just add the actual version as parameter? And once you have the checkbox a second parameter with a boolean value. 
Then the server can do the actual logic of deciding what updates to deliver, might come in handy at some point.

guruz:
hmmm my SUFFIX is "git" but the URL indeed does not contain this
https://updates.owncloud.com/client/?version=2.4.0.0&platform=macos&oem=ownCloud&channel=beta

> Then the server can do the actual logic of deciding what updates to deliver, might come in handy at some point.

The user shall be in control via checkbox (or for now implicitely by having downloaded a alpha/beta/rc) of what level of update he's willing to accept.

For example Chrome download URL: https://www.google.com/chrome/browser/beta.html?platform=mac&extra=betachannel

guruz:
Updated commit leads to send 2.4.0git as stringversion.

guruz:
... or shall we just send the versionsuffix? hm

dschmidt:
Yes of course the user can decide - that's why I want to transmit the choice too. But I feel we should put as much logic on the server side as possible. It's way more flexible and allows changing it "dynamically" which could come in handy at some point.

ogoffart:
:+1:  looks good to me, as far as i'm concerned.

guruz:
The build number from our CI is part of the version e.g. 2.4.0.31337.
So that would be enough for detecting the update.
So I recommend instead of sending stringversion additionally we just send versionsuffix additionally e.g. "git" or "beta1".
OK?

guruz:
>  If so, I am missing a channel concept there....

Discussed in agenda-desktop-meeting-mar-2017 :-)

```
Have release channels "stable" and "beta" for the auto updater
- Selection should be disabled in branded clients (or branding option later)
- Update channels for the production client: stable, beta
- Update channels for the testpilots client: stable, beta, nightly
```

(with sending nothing to updater implying stable)

SamuAlfageme:
Transparent option for nightly update channel would make me so 😍

guruz:
> Transparent option for nightly update channel would make me so 😍

What do you mean by transparent? if you use linux, isn't it just apt-get update?

SamuAlfageme:
@guruz yeah. I meant something like seeing the updater/sparkle dialog on win/macOS every morning.

guruz:
I opted for the versionsuffix solution.
```https://updates.owncloud.com/client/?version=2.4.0.0&platform=macos&oem=ownCloud&versionsuffix=alpha1&channel=beta```

guruz:
@michaelstingl Is there anyone that volunteers to do the server-side code for this?
Then we can update 2.4 alpha1 users to 2.4 beta1.

guruz:
Started working on this https://github.com/owncloud/enterprise/tree/master/client_update_checker

guruz:
https://github.com/owncloud/enterprise/pull/2332

PR #6052, opened 2017-09-25
Optimize csync_reconcile.

ogoffart:
csync_reconcile takes 1/4 of the CPU  time after this patch


src/csync/csync_private.h
```diff
@@ -67,14 +67,47 @@ enum csync_replica_e {
...
 };
 
+
+/*
+ * This is a structurere similar to QStringRef
+ * The difference is that it keeps the QByteArray by value and not by pointer
+ * And it only implements a very small subset of the API that is required by csync, the API can be
+ * added as we need it.
+ */
+class ByteArrayRef
```
mrow4a: Did you just optimize QByteArray and it immedietaly gives 1/4 speed-up for csync_reconcile? nice :+1:

ogoffart: Instead of using QByteArray, i use a structure that is a slice of a QByteArray. Before we created lots of QByteArray for the parent directory, but this allocated and copied a lot of memory.  By just having a slice, we don't need to allocate.

src/csync/csync_private.h
```diff
@@ -67,14 +67,47 @@ enum csync_replica_e {
...
+        , _begin(begin)
+        , _size(qMin(arr.size() - begin, size < 0 ? arr.size() : size))
+    {
+    }
+    ByteArrayRef left(int l) const { return ByteArrayRef(_arr, _begin, l); };
+    char at(int x) const { return _arr.at(_begin + x); }
+    int size() const { return _size; }
+    int length() const { return _size; }
+    bool isEmpty() const { return _size == 0; }
+    /* Pinter to the beginning of the data. WARNING: not null terminated */
```
mrow4a: pointer?

ogoffart: yes

src/csync/csync_reconcile.cpp
```diff
@@ -36,9 +36,11 @@ Q_LOGGING_CATEGORY(lcReconcile, "sync.csync.reconciler", QtInfoMsg)
 #define __STDC_FORMAT_MACROS
 #include "inttypes.h"
 
+
```
jturcotte: Nit: empty line

src/csync/csync_reconcile.cpp
```diff
@@ -47,7 +49,7 @@ static csync_file_stat_t *_csync_check_ignored(csync_s::FileMap *tree, const QBy
     if (parentlen <= 0) {
         return nullptr;
     }
-    QByteArray parentPath = path.left(parentlen);
+    auto parentPath = path.left(parentlen);
```
jturcotte: I think this would be worth typing explicitly, the API is too similar and people should be able to be aware of the type for things like the no null-termination above.

src/csync/csync_private.h
```diff
@@ -67,14 +67,44 @@ enum csync_replica_e {
...
+        : _arr(arr)
+        , _begin(begin)
+        , _size(qMin(arr.size() - begin, size < 0 ? arr.size() : size))
+    {
+    }
+    ByteArrayRef left(int l) const { return ByteArrayRef(_arr, _begin, l); };
+    char at(int x) const { return _arr.at(_begin + x); }
+    int size() const { return _size; }
+    /* Pointer to the beginning of the data. WARNING: not null terminated */
+    const char *data() const { return _arr.constData() + _begin; }
```
jturcotte: Could you make it private so that we control the call sites and wrap them all with size-checking? Nothing in csync should ever call this function, especially if there is any way this could be mistaken for a real QByteArray.

src/csync/csync_private.h
```diff
@@ -67,14 +67,44 @@ enum csync_replica_e {
...
+{
+    QByteArray _arr;
+    int _begin = 0;
+    int _size = -1;
+
+public:
+    ByteArrayRef(const QByteArray &arr = {}, int begin = 0, int size = -1)
+        : _arr(arr)
+        , _begin(begin)
+        , _size(qMin(arr.size() - begin, size < 0 ? arr.size() : size))
```
jturcotte: Could this be an assert instead?

src/csync/csync_private.h
```diff
@@ -67,14 +67,44 @@ enum csync_replica_e {
...
+ * added as we need it.
+ */
+class ByteArrayRef
+{
+    QByteArray _arr;
+    int _begin = 0;
+    int _size = -1;
+
+public:
+    ByteArrayRef(const QByteArray &arr = {}, int begin = 0, int size = -1)
```
jturcotte: `arr` could be passed by value and moved into `_arr`. Would it make sense for a type like QByteArray?

src/csync/csync_private.h
ogoffart: I did not make it an assert after all,  this is also how QByteArray::left works, by being very tollerent in what we accept

PR #6054, opened 2017-09-25
Use QDateTime::currentDateTimeUtc instead of the non utc version

ogoffart:
QDateTime::currentDateTime is terribly slow.

### DIFF
--- a/src/libsync/propagatedownload.cpp
+++ b/src/libsync/propagatedownload.cpp
@@ -673,7 +673,7 @@ namespace { // Anonymous namespace for the recall feature
             dotLocation = recallFileName.size();
         }
 
-        QString timeString = QDateTime::currentDateTime().toString("yyyyMMdd-hhmmss");
+        QString timeString = QDateTime::currentDateTimeUtc().toString("yyyyMMdd-hhmmss");
         recallFileName.insert(dotLocation, "_.sys.admin#recall#-" + timeString);
 
         return recallFileName;
--- a/src/libsync/propagateupload.cpp
+++ b/src/libsync/propagateupload.cpp
@@ -53,7 +53,7 @@ Q_LOGGING_CATEGORY(lcPropagateUpload, "sync.propagator.upload", QtInfoMsg)
 static bool fileIsStillChanging(const SyncFileItem &item)
 {
     const QDateTime modtime = Utility::qDateTimeFromTime_t(item._modtime);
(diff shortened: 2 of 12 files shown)


jturcotte:
Could you explain why they are equivalent? Won't this offset all timestamps vs the database?

ogoffart:
The difference between the two is the Qt::TimeSpec.
But any computation on the QDateTime will result in the same results. In particular toMSecsSinceEpoch.

src/libsync/propagateupload.cpp
```diff
@@ -53,7 +53,7 @@ Q_LOGGING_CATEGORY(lcPropagateUpload, "sync.propagator.upload", QtInfoMsg)
 static bool fileIsStillChanging(const SyncFileItem &item)
 {
     const QDateTime modtime = Utility::qDateTimeFromTime_t(item._modtime);
-    const qint64 msSinceMod = modtime.msecsTo(QDateTime::currentDateTime());
+    const qint64 msSinceMod = modtime.msecsTo(QDateTime::currentDateTimeUtc());
```
guruz: What about
```pi._modtime = Utility::qDateTimeFromTime_t(_item->_modtime);```
Is there anything that need to be done?

File: src/libsync/syncresult.cpp [hunk 2]
```diff
@@ -85,7 +85,7 @@ QString SyncResult::statusString() const
 void SyncResult::setStatus(Status stat)
 {
     _status = stat;
-    _syncTime = QDateTime::currentDateTime();
+    _syncTime = QDateTime::currentDateTimeUtc();
```
jturcotte: This actually looks unused now.

src/libsync/propagatedownload.cpp
```diff
@@ -673,7 +673,7 @@ namespace { // Anonymous namespace for the recall feature
             dotLocation = recallFileName.size();
         }
 
-        QString timeString = QDateTime::currentDateTime().toString("yyyyMMdd-hhmmss");
+        QString timeString = QDateTime::currentDateTimeUtc().toString("yyyyMMdd-hhmmss");
         recallFileName.insert(dotLocation, "_.sys.admin#recall#-" + timeString);
```
jturcotte: This one could actually stay, but I guess users shouldn't really see this anyway (nor do I fully understand why we append a date there).

PR #6056, opened 2017-09-25
Csync remove statedb

ogoffart:


ogoffart:
( that's the patches from @jturcotte )
They are :+1:  from me

PR #6060, opened 2017-09-26
Excludes: During directory traversal, use QRegularExpression

guruz:
(Just for @ckamm for now, need to fix up more)

### DIFF
--- a/src/csync/csync_exclude.cpp
+++ b/src/csync/csync_exclude.cpp
@@ -39,6 +39,8 @@
 
 #include "common/utility.h"
 
+#include <QString>
+
 #ifdef _WIN32
 #include <io.h>
 #else
@@ -216,22 +218,11 @@ bool csync_is_windows_reserved_word(const char* filename) {
   return false;
 }
 
-static CSYNC_EXCLUDE_TYPE _csync_excluded_common(c_strlist_t *excludes, const char *path, int filetype, bool check_leading_dirs) {
-    size_t i = 0;
-    const char *bname = NULL;
+/* FIXME: This should use a big thread-local QRegularExpression */
+static CSYNC_EXCLUDE_TYPE _csync_excluded_common(const char *bname) {
     size_t blen = 0;
     int rc = -1;
     CSYNC_EXCLUDE_TYPE match = CSYNC_NOT_EXCLUDED;
-    CSYNC_EXCLUDE_TYPE type  = CSYNC_NOT_EXCLUDED;
-    c_strlist_t *path_components = NULL;
-
-    /* split up the path */
-    bname = strrchr(path, '/');
-    if (bname) {
-        bname += 1; // don't include the /
-    } else {
-        bname = path;
-    }
     blen = strlen(bname);
 
     rc = csync_fnmatch("._sync_*.db*", bname, 0);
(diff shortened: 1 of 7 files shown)


src/csync/csync_exclude.cpp
```diff
@@ -320,106 +322,249 @@ static CSYNC_EXCLUDE_TYPE _csync_excluded_common(c_strlist_t *excludes, const ch
         goto out;
```
ckamm: function doesn't need `excludes` anymore

src/csync/csync_exclude.cpp
```diff
@@ -320,106 +322,249 @@ static CSYNC_EXCLUDE_TYPE _csync_excluded_common(c_strlist_t *excludes, const ch
...
+    if (match != CSYNC_NOT_EXCLUDED) {
+        return match;
+    }
+    //qDebug() << path;
+
+    const char *bname = NULL;
+    size_t blen = 0;
+    /* split up the path */
+    bname = strrchr(path, '/');
+    if (bname) {
```
ckamm: probably pass bname into `excluded_common`

src/csync/csync_exclude.cpp
```diff
@@ -320,106 +322,249 @@ static CSYNC_EXCLUDE_TYPE _csync_excluded_common(c_strlist_t *excludes, const ch
...
+    blen = strlen(bname);
+    QString p = QString::fromUtf8(bname);
 
+    if (ctx->parsed_traversal_excludes.regexp_exclude.match(p).hasMatch()) {
+        //qDebug() << "WOULD EXCLUDE";
+        return CSYNC_FILE_EXCLUDE_LIST;
+    }
+    if (ctx->parsed_traversal_excludes.regexp_exclude_and_remove.match(p).hasMatch()) {
+        //qDebug() << "WOULD EXCLUDE AND REMOVE";
+        return CSYNC_FILE_EXCLUDE_AND_REMOVE;
```
ckamm: could we use one regexp by using captures? `(exclude|these)|(exclude|and|remove|these)` and then check which capture is nonempty? Unclear whether that'd be faster

src/csync/csync_exclude.cpp
```diff
@@ -320,106 +322,249 @@ static CSYNC_EXCLUDE_TYPE _csync_excluded_common(c_strlist_t *excludes, const ch
...
+    QString s = QRegularExpression::escape(exclude).replace("\\*", "[^/]*").replace("?", "[^/]");
+    // Anchoring to the beginning if pattern starts with a /
+    if (exclude.startsWith("/")) {
+        s.prepend("^");
+    } else {
+        // we don't want "foo/bar" to match "afoo/bar"
+        s.prepend("/");
+    }
+    if (!exclude.endsWith("/"))
+        s.append("/?");
```
ckamm: I'm pretty sure this should be `(/|$)`, otherwise the pattern "foo" will match "foobar"

src/csync/csync_exclude.cpp
```diff
@@ -320,106 +322,249 @@ static CSYNC_EXCLUDE_TYPE _csync_excluded_common(c_strlist_t *excludes, const ch
...
+      if (!pattern[0]) { /* empty pattern */
+          continue;
+      }
+      /* Excludes starting with ']' means it can be cleanup */
+      if (pattern[0] == ']') {
+          ++pattern;
+          if (filetype == CSYNC_FTW_TYPE_FILE) {
+              type = CSYNC_FILE_EXCLUDE_AND_REMOVE;
+          }
+      }
```
ckamm: Obviously, duplicating the logic like this is really bad.

src/csync/csync_exclude.cpp
```diff
@@ -320,106 +322,249 @@ static CSYNC_EXCLUDE_TYPE _csync_excluded_common(c_strlist_t *excludes, const ch
...
+    QString _exclude_traversel_regexp_exclude_and_remove;
+    for (unsigned int i = 0; i < ctx->excludes->count; i++) {
+        char *exclude = ctx->excludes->vector[i];
+        QString *builderToUse = & _exclude_traversel_regexp_exclude;
+        if (exclude[0] == '\n') continue; // empty line
+        if (exclude[0] == '\r') continue; // empty line
+
+        /* If an exclude entry contains a slash, we use the C-style codepath without QRegularEpression */
+        if (strchr(exclude, '/')) {
+            _csync_exclude_add(&ctx->parsed_traversal_excludes.list_patterns_with_slashes, exclude);
```
ckamm: I think special casing these at first to get the patch done quicker and with less room for mistakes is nice.

src/csync/csync_private.h
```diff
@@ -99,8 +101,17 @@ struct OCSYNC_EXPORT csync_s {
...
-  c_strlist_t *excludes = nullptr;
-  
+
+  c_strlist_t *excludes = nullptr; /* list of individual patterns collected from all exclude files */
+  struct {
+      QRegularExpression regexp_exclude;
+      QRegularExpression regexp_exclude_and_remove;
+      /* FIXME ^^ could we use one regexp by using captures? (exclude|these)|(exclude|and|remove|these)
+       * and then check which capture is nonempty? Unclear whether that'd be faster*/
+      c_strlist_t *list_patterns_with_slashes = nullptr;
```
ogoffart: I guess this should be a QVector or something

guruz:
Also adding the fixed patterns into a regexp is not worth it on macOS (performance regresses). https://gist.github.com/guruz/11f29b8e340d4755ccfa919d1d2bd971 
Might be worth it on Windows though with all those special Windows filenames..?

src/csync/csync_private.h
guruz: @ogoffart I copy pasted code from the previous implementation which is using c_strlist_t. I agreed with @ckamm that we keep the code un-cleaned up for now and try to get this PR in. Cleaning can be done in a follow up (assuming tests etc still pass)

src/csync/csync_exclude.cpp
guruz: done

guruz:
I consider this mergeable for the 2.4 alpha as soon as the last test is fixed.
```
The following tests FAILED:
	 14 - SyncFileStatusTrackerTest (OTHER_FAULT)
Errors while running CTest
```

need help for this (see IRC)

guruz:
Benchmark:
```

before on master
csync_excluded_traversal: 0.005948 ms per call
csync_excluded_traversal: 0.006038 ms per call
csync_excluded_traversal: 0.005742 ms per call

after:
csync_excluded_traversal: 0.002532 ms per call
csync_excluded_traversal: 0.002361 ms per call
csync_excluded_traversal: 0.002386 ms per call
csync_excluded_traversal: 0.002445 ms per call
```

ckamm:
Benchmark of f533bd95c9209dbd69d645fe3239d6982cb0bf7b on linux:

```
without patch
csync_excluded: 0.020413 ms per call
csync_excluded_traversal: 0.000756 ms per call

with patch
csync_excluded: 0.020131 ms per call
csync_excluded_traversal: 0.000650 ms per call
```

So it's not making it worse, like the old patch! :)

src/csync/csync_exclude.cpp
ckamm: @guruz Also, `list_patterns_with_slashes` needs to be wiped in this function to make it not grow longer and longer.

src/csync/csync_exclude.cpp
```diff
@@ -316,110 +307,265 @@ static CSYNC_EXCLUDE_TYPE _csync_excluded_common(c_strlist_t *excludes, const ch
...
+    if (!ctx->excludes) {
+        qDebug() << "FURUZ";
+        //ctx->parsed_traversal_excludes.list_patterns_with_slashes = c_strlist_new(0);
+        return;
+    }
 
-        type = CSYNC_FILE_EXCLUDE_LIST;
-        if (!pattern[0]) { /* empty pattern */
+    QString _exclude_traversel_regexp_exclude;
+    QString _exclude_traversel_regexp_exclude_and_remove;
```
ckamm: There's a bug here: an empty regex will match anything. Check my patchset for how to start out with a match-nothing regex!

ckamm:
I'm now much more worried about this patch: I'm not convinced our tests are sufficient to guarantee `_no_ctx` and `_traversal` both behave correctly when they use different codepaths.

guruz:
superseded by https://github.com/owncloud/client/pull/6063

PR #6063, opened 2017-09-27
Excludes: During directory traversal, use QRegularExpression

guruz:
On my Mac, this halves the time spent in csync_excluded_traversal
when using check_csync_excluded_performance.
On Linux the changes were not so much visible, maybe the fnmatch
there is much faster.

### DIFF
--- a/src/csync/csync_exclude.cpp
+++ b/src/csync/csync_exclude.cpp
@@ -39,6 +39,8 @@
 
 #include "common/utility.h"
 
+#include <QString>
+
 #ifdef _WIN32
 #include <io.h>
 #else
@@ -234,20 +236,28 @@ static CSYNC_EXCLUDE_TYPE _csync_excluded_common(c_strlist_t *excludes, const ch
     }
     blen = strlen(bname);
 
-    rc = csync_fnmatch("._sync_*.db*", bname, 0);
-    if (rc == 0) {
-        match = CSYNC_FILE_SILENTLY_EXCLUDED;
-        goto out;
-    }
-    rc = csync_fnmatch(".sync_*.db*", bname, 0);
-    if (rc == 0) {
-        match = CSYNC_FILE_SILENTLY_EXCLUDED;
-        goto out;
-    }
-    rc = csync_fnmatch(".csync_journal.db*", bname, 0);
-    if (rc == 0) {
-        match = CSYNC_FILE_SILENTLY_EXCLUDED;
-        goto out;
+    // 9 = strlen(".sync_.db")
+    if (blen >= 9 && bname[0] == '.') {
+        rc = csync_fnmatch("._sync_*.db*", bname, 0);
+        if (rc == 0) {
+            match = CSYNC_FILE_SILENTLY_EXCLUDED;
+            goto out;
+        }
+        rc = csync_fnmatch(".sync_*.db*", bname, 0);
+        if (rc == 0) {
+            match = CSYNC_FILE_SILENTLY_EXCLUDED;
+            goto out;
+        }
+        rc = csync_fnmatch(".csync_journal.db*", bname, 0);
+        if (rc == 0) {
+            match = CSYNC_FILE_SILENTLY_EXCLUDED;
+            goto out;
+        }
+        rc = csync_fnmatch(".owncloudsync.log*", bname, 0);
+        if (rc == 0) {
+            match = CSYNC_FILE_SILENTLY_EXCLUDED;
+            goto out;
+        }
     }
 
     // check the strlen and ignore the file if its name is longer than 254 chars.
@@ -298,12 +308,6 @@ static CSYNC_EXCLUDE_TYPE _csync_excluded_common(c_strlist_t *excludes, const ch
 
     /* We create a desktop.ini on Windows for the sidebar icon, make sure we don't sync them. */
     rc = csync_fnmatch("Desktop.ini", bname, 0);
-    if (rc == 0) {
-        match = CSYNC_FILE_SILENTLY_EXCLUDED;
-        goto out;
-    }
-
-    rc = csync_fnmatch(".owncloudsync.log*", bname, 0);
     if (rc == 0) {
         match = CSYNC_FILE_SILENTLY_EXCLUDED;
         goto out;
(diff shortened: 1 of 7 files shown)


guruz:
Done with a lot of help from @ckamm !
Kudos..

src/csync/csync_exclude.cpp
```diff
@@ -415,8 +438,90 @@ static CSYNC_EXCLUDE_TYPE _csync_excluded_common(c_strlist_t *excludes, const ch
...
+        if (builderToUse->size() > 0) {
+            builderToUse->append("|");
+        }
+        builderToUse->append(convertToBnameRegexpSyntax(exclude));
+    }
+
+    QString pattern = "^(" + exclude_only + ")$|^(" + exclude_and_remove + ")$";
+    ctx->parsed_traversal_excludes.regexp_exclude.setPattern(pattern);
+    ctx->parsed_traversal_excludes.regexp_exclude.setPatternOptions(QRegularExpression::OptimizeOnFirstUsageOption
+#if defined(Q_OS_WIN) || defined(Q_OS_MAC)
```
ogoffart: should be using the check for case sensitivity in utility

src/csync/csync_exclude.cpp
```diff
@@ -204,16 +206,35 @@ bool csync_is_windows_reserved_word(const char* filename) {
...
             return true;
         }
     }
-  }
-  return false;
+
+    return false;
 }
 
 static CSYNC_EXCLUDE_TYPE _csync_excluded_common(c_strlist_t *excludes, const char *path, int filetype, bool check_leading_dirs) {
```
ogoffart: My benchmark test shows that this is still significant when there are no exclude.  Can they not be put in the regexp?

src/csync/csync_exclude.cpp
```diff
@@ -204,16 +206,35 @@ bool csync_is_windows_reserved_word(const char* filename) {
             return true;
         }
     }
-    if (c_strncasecmp(filename, win_reserved_words[j], len_reserved_word) == 0) {
-        if (len_filename == len_reserved_word) {
-            return true;
+
+    if (len_filename == 3 || (len_filename > 3 && filename[3] == '.')) {
+        size_t n_words = sizeof(win_reserved_words_3) / sizeof(char *);
+        for (size_t j = 0; j < n_words; j++) {
```
ogoffart: could be written:
```
for (const char *word : win_reserved_words_3)
```

(less error prone)

src/csync/csync_exclude.cpp
```diff
@@ -204,16 +206,35 @@ bool csync_is_windows_reserved_word(const char* filename) {
...
+            if (c_strncasecmp(filename, win_reserved_words_3[j], 3) == 0) {
+                return true;
+            }
+        }
+    }
+
+    if (len_filename == 4 || (len_filename > 4 && filename[4] == '.')) {
+        size_t n_words = sizeof(win_reserved_words_4) / sizeof(char *);
+        for (size_t j = 0; j < n_words; j++) {
```
ogoffart: (same here)

src/csync/csync_exclude.cpp
```diff
@@ -204,16 +206,35 @@ bool csync_is_windows_reserved_word(const char* filename) {
...
+            if (c_strncasecmp(filename, win_reserved_words_4[j], 4) == 0) {
+                return true;
+            }
         }
-        if ((len_filename > len_reserved_word) && (filename[len_reserved_word] == '.')) {
+    }
+
+    size_t n_words = sizeof(win_reserved_words_n) / sizeof(char *);
+    for (size_t j = 0; j < n_words; j++) {
```
ogoffart: (and here)

src/csync/csync_exclude.cpp
```diff
@@ -415,8 +438,90 @@ static CSYNC_EXCLUDE_TYPE _csync_excluded_common(c_strlist_t *excludes, const ch
...
-  return _csync_excluded_common(excludes, path, filetype, false);
+/* Only for bnames (not paths) */
+static QString convertToBnameRegexpSyntax(QString exclude)
+{
+    QString s = QRegularExpression::escape(exclude).replace("\\*", ".*").replace("?", ".");
+    //qDebug() << "Converted pattern" << exclude << "to regex" << s;
+    return s;
+}
+
+void csync_exclude_traversal_prepare(CSYNC *ctx)
```
ogoffart: wondering if this should not be a member function of  the parsed_traversal_excludes class

src/csync/csync_exclude.cpp
```diff
@@ -415,8 +438,90 @@ static CSYNC_EXCLUDE_TYPE _csync_excluded_common(c_strlist_t *excludes, const ch
...
+    /* Now check with our optimized regexps */
+    const char *bname = NULL;
+    /* split up the path */
+    bname = strrchr(path, '/');
+    if (bname) {
+        bname += 1; // don't include the /
+    } else {
+        bname = path;
+    }
+    QString p = QString::fromUtf8(bname);
```
ogoffart: Please move under the if

src/csync/csync_exclude.cpp
ckamm: I don't understand what you mean? That excluded_common still shows up in benchmarks even when there are no excludes?

ckamm:
@ogoffart updated

ogoffart:
From jenkins:

```
CMakeFiles/owncloudsync.dir/objects.a(discoveryphase.cpp.obj):discoveryphase.cpp:(.text+0x573d): undefined reference to `csync_exclude_traversal_prepare(csync_s*)'

```

guruz:
@ogoffart Jenkins passing, review needed:)

src/csync/csync_private.h
```diff
@@ -102,10 +104,18 @@ struct OCSYNC_EXPORT csync_s {
...
-  
+
   OCC::SyncJournalDb *statedb;
 
+  c_strlist_t *excludes = nullptr; /* list of individual patterns collected from all exclude files */
+  struct TraversalExcludes {
+      void prepare(c_strlist_t *excludes);
+
+      QRegularExpression regexp_exclude;
+      c_strlist_t *list_patterns_with_slashes = nullptr;
```
ogoffart: Is this not leaking?  Where is it destroyed?

src/csync/csync_private.h
```diff
@@ -102,10 +104,18 @@ struct OCSYNC_EXPORT csync_s {
...
+      /* FIXME ^^ at a later point use QRegularExpression too if those become popular */
+  } parsed_traversal_excludes;
```
ogoffart: Maybe this should be a pointer to the ExcludedFiles object instead, and everything should be put in the ExcludeFiles class.  This would avoid the leak.   Curently , the ~ExcludeFiles take care about destroying the `excludes` pointer.
Also now i guess we could use QVector<QByteArray>  instead od c_strlist_t.

ogoffart:
Also it seems that this does not optimize csync_excluded_no_ctx which is called by ExcludedFiles::isExcluded,   this is important as well at startup as it is called for every directory while initiating the file system watcher...  (see also https://github.com/owncloud/client/pull/5245 )

guruz:
@ogoffart Pushed a fix for the leak. Looks good?

PS: I agree with optimizing the _no_ctx too. I don't agree with having the list in ```Excludes``` object, it feels weird to have the ownership there. The ownership of something inside ```CSYNC* ```should be from ```CSYNC*```

src/csync/csync_exclude.cpp
```diff
@@ -415,8 +434,93 @@ static CSYNC_EXCLUDE_TYPE _csync_excluded_common(c_strlist_t *excludes, const ch
     return match;
 }
 
-CSYNC_EXCLUDE_TYPE csync_excluded_traversal(c_strlist_t *excludes, const char *path, int filetype) {
-  return _csync_excluded_common(excludes, path, filetype, false);
+/* Only for bnames (not paths) */
+static QString convertToBnameRegexpSyntax(QString exclude)
+{
+    QString s = QRegularExpression::escape(exclude).replace("\\*", ".*").replace("?", ".");
```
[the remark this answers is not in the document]
ckamm: +1 on fixing "?"

I'm not convinced it's worth it to try to emulate `fnmatch` behavior completely. For example `PathMatchSpec` we use on windows also offers quite different matching options.

If we want to be 100% backwards compatible we'll need to introduce a new prefex character to go into the 'fast-regex' codepath and can never get rid of the `fnmatch` based codepath.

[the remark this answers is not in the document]
ogoffart: we would not need a pattern-prefix. We could just detect occurence of [ or { in the pattern and then put them in the slash pattern section. they should be rare anyway.

guruz:
@ogoffart fixup pushed

guruz:
thanks!

will rebase and squash my 3 commits and have @ckamm 's on top

src/csync/csync_exclude.cpp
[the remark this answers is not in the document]
ckamm: @ogoffart True. Minor: `{` does not seem to have special meaning to fnmatch, only ?, *, [ and escaping. (had to check glibc's fnmatch_loop.c)

PR #6066, opened 2017-09-27
support old sqlite before 3.7.7 e.g. CentOS-6

jnweiger:
quickfix: define the symbol for backwards compat
or FIXME: better solution #ifdef SQLITE_IOERR_SHMMAP where needed...

### DIFF
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -345,6 +345,12 @@ bool SyncJournalDb::checkConnect()
                         "PRIMARY KEY(phash)"
                         ");");
 
+#ifndef SQLITE_IOERR_SHMMAP
+// Requires sqlite >= 3.7.7 but old CentOS6 has sqlite-3.6.20
+// Definition taken from https://sqlite.org/c3ref/c_abort_rollback.html
+#define SQLITE_IOERR_SHMMAP            (SQLITE_IOERR | (21<<8))
+#endif
+
     if (!createQuery.exec()) {
         // In certain situations the io error can be avoided by switching
         // to the DELETE journal mode, see #5723


michaelstingl:
I don't understand the implications? @guruz Is it okay to support old sqlite? Would you like to drop support for old sqlite?

jnweiger:
@michaelstingl the implication is RHEL6 or not RHEL6. Their sqlite is and old 3.6.20 that lacks features.

michaelstingl:
So drop CentOS-6 (again)?

jnweiger:
I had investigated a little into FindSQLite3.cmake and found the code to be not very strict:
In find_package(SQLite3 3.8.0 REQUIRED) we happily returned a 3.6.20 and used that. 
Fallback to USE_OUR_OWN_SQLITE3 does not seem to happen automatically there.

guruz:
@jnweiger Meh. In my opinion we should always use the bundles sqlite3 because that's what sqlite3 recommends. (last time i checked)

ogoffart:
:-1:  This will compile but will fail at runtime.  The client will not use the index, which might result in horribly slow code at best, or dysfunctional discovery or simply crash.

jnweiger:
I still have not heared a clear advice what to do. I'd vote for always embed the exact version that we want to use, as sqlite is perfectly suited for embedding. But there is a scary comment in https://github.com/owncloud/client/pull/6066#issuecomment-332796317 saying that embedding would cause clashes with QtWebKit.

guruz:
@jnweiger wrong link? I don't see anything about clashing with QtWebKit?

@ogoffart Please don't edit comments if there is new information. I guess @jnweiger didn't get the notification that there is actually nothing urgent that needs to be done.
I can make some code for 2.5.2 that prefers the internal sqlite3 if necessary.

jnweiger:
Quote from https://github.com/owncloud/client/pull/6066#pullrequestreview-65930010: "... i think it's too risky ..."

guruz:
I'm trying to remember..
Was it some OS difference about symbol visibility / exports that would behave differently on macOS vs Linux?
Maybe @crrodriguez @jturcotte @ogoffart remember?

guruz:
Some related discussion here https://news.ycombinator.com/item?id=7756332 (and on the linked blog post).

guruz:
Maybe we can just postpone this until we remove QtWebkit :-/

PR #6070, opened 2017-10-02
use IS_DIRECTORY ${CMAKE_SOURCE_DIR}/admin

jnweiger:
We need to test presence of admin folder using ${CMAKE_SOURCE_DIR}, or it is always false. Bad for osx, (but okayish for everything else).

This should fix failing nightlies.


jnweiger:
The 2.3.4 nightly ran 1h before this merge, the 2.4.0 nightly ran 1h after this merge.
Perfect timing! All nightlies succeed. Thanks!

PR #6078, opened 2017-10-05
Use display-name from the ocs call in the settings dialog

DeepDiver1975:
refs https://github.com/owncloud/oauth2/issues/74#issuecomment-319914664

bildschirmfoto von 2017-10-05 21-13-07

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -638,7 +638,11 @@ void AccountSettings::slotAccountStateChanged()
                                  Utility::escape(safeUrl.toString()));
         QString serverWithUser = server;
         if (AbstractCredentials *cred = account->credentials()) {
-            serverWithUser = tr("%1 as <i>%2</i>").arg(server, Utility::escape(cred->user()));
+            QString user = account->davDisplayName();
+            if (user.isEmpty()) {
+                user = cred->user();
+            }
+            serverWithUser = tr("%1 as <i>%2</i>").arg(server, Utility::escape(user));
         }
 
         if (state == AccountState::Connected) {
--- a/src/gui/settingsdialog.cpp
+++ b/src/gui/settingsdialog.cpp
@@ -236,7 +236,7 @@ void SettingsDialog::accountAdded(AccountState *s)
 
     if (!brandingSingleAccount) {
         accountAction->setToolTip(s->account()->displayName());
-        accountAction->setIconText(s->shortDisplayNameForSettings(height * buttonSizeRatio));
+        accountAction->setIconText(shortDisplayNameForSettings(s->account().data(),  height * buttonSizeRatio));
     }
     _toolBar->insertAction(_toolBar->actions().at(0), accountAction);
     auto accountSettings = new AccountSettings(s, this);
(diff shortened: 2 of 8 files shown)


src/gui/settingsdialog.cpp
```diff
@@ -236,7 +236,7 @@ void SettingsDialog::accountAdded(AccountState *s)
 
     if (!brandingSingleAccount) {
         accountAction->setToolTip(s->account()->displayName());
-        accountAction->setIconText(s->shortDisplayNameForSettings(height * buttonSizeRatio));
+        accountAction->setIconText(shortDisplayNameForSettings(s->account().data(),  height * buttonSizeRatio));
```
mrow4a: You have two spaces here

DeepDiver1975: 😱

src/gui/accountsettings.cpp
```diff
@@ -638,7 +638,11 @@ void AccountSettings::slotAccountStateChanged()
                                  Utility::escape(safeUrl.toString()));
         QString serverWithUser = server;
         if (AbstractCredentials *cred = account->credentials()) {
-            serverWithUser = tr("%1 as <i>%2</i>").arg(server, Utility::escape(cred->user()));
+            QString user = account->davDisplayName();
```
mrow4a: You dont need to instantiate new string, just do ```if(account->davDisplayName().isEmpty())``` and then e.g. ```serverWithUser = tr("%1 as <i>%2</i>").arg(server, Utility::escape(account->davDisplayName()));```

mrow4a:
For me looks good, but lets @ogoffart have a look also

DeepDiver1975:
I just saw that the tray context menu needs to be updated as well?

mrow4a:
let me build it later and check manually how does it look like - but yes, if this is possible. 

BTW, I should give :+1 just for Alice Cooper and John Snow :laughing:

src/gui/accountsettings.cpp
mrow4a: But whatever seems clearer to you is good.

jturcotte: QString is implicitly shared. This will just takes stores the pointer on the stack inside that `QString` until the end of the block and won't impact memory usage at all.

Repeating `davDisplayName()` would actually be slightly worse since the function isn't inlined and you'd call it twice across DLLs, but performance-wise this also won't make any noticeable difference unless you call this 1M times either anyway.

ckamm:
Looks good to me! :+1:

guruz:
(might break compiling on macOS because of the SettingsDialogMac.. we'll see)

src/gui/accountsettings.cpp
mrow4a: Qt will not stop suprise me, good to know

guruz:
I'll fix the mac compilation

ogoffart:
Well, my previous review was sent as I reviewed an outdated patch it seems

SamuAlfageme:
@guruz @DeepDiver1975 it seems that different values are being applied both in the tab label and the account identifier; see: 

<p align="center">
  
</p>


I'm testing this on the [latest nightly](http://download.owncloud.com/desktop/daily/ownCloud-2.4.0.8403-nightly20171009.pkg) available for macOS (`20171009`)

DeepDiver1975:
> I'm testing this on the latest nightly available for macOS (20171009)

I did develop this on linux - afaik macos has it's own impl of settings (as I learned later)

@guruz you run a mac - right? can you take care please? THX

ckamm:
@guruz I assume the equivalent of `SettingsDialog::slotAccountDisplayNameChanged()` is missing on macos and the name is simply not updated once it's been retrieved from the server.

guruz:
-> https://github.com/owncloud/client/pull/6102

PR #6079, opened 2017-10-06
Context menu: More detailed status messages

ckamm:
Previously it could only display synchronization progress or "up to
date". Now it also communicates the same overall state that the icon
shows.

See owncloud/enterprise#2134

### DIFF
--- a/src/gui/owncloudgui.cpp
+++ b/src/gui/owncloudgui.cpp
@@ -207,13 +207,23 @@ void ownCloudGui::slotComputeOverallSyncStatus()
 {
     bool allSignedOut = true;
     bool allPaused = true;
+    bool allDisconnected = true;
     QVector<AccountStatePtr> problemAccounts;
+    auto setStatusText = [&](const QString &text) {
+        // Don't overwrite the status if we're currently syncing
+        if (FolderMan::instance()->currentSyncFolder())
+            return;
+        _actionStatus->setText(text);
+    };
+
     foreach (auto a, AccountManager::instance()->accounts()) {
         if (!a->isSignedOut()) {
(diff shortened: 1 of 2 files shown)


src/gui/owncloudgui.cpp
```diff
@@ -250,45 +265,53 @@ void ownCloudGui::slotComputeOverallSyncStatus()
...
+            setStatusText(tr("Up to date"));
+        } else if (overallResult == SyncResult::Paused) {
+            setStatusText(tr("Synchronization is paused"));
+        } else {
+            setStatusText(tr("Error during synchronization"));
+        }
     } else {
-        // undefined because there are no folders.
-        QIcon icon = Theme::instance()->syncStateIcon(SyncResult::Problem, true, contextMenuVisible());
+        if (overallResult == SyncResult::Undefined)
```
guruz: this new behaviour is on purpose right?

ogoffart:
Build failure:
 ```
AW3AQRNBRINOALDQH52TOARZHFRUNKQESJBPVLYZHCMYBK2AWUXQ/src/gui/owncloudgui.cpp:287:63: error: request for member 'status' in 'overallResult', which is of non-class type 'OCC::SyncResult::Status'
         trayMessage = folderMan->statusToString(overallResult.status(), false);
```

src/gui/owncloudgui.cpp
ckamm: Actually, the behavior is identical. There were two branches before that lead to `no folders` and now there's only one because I unified them.

ckamm:
Fixed windows compile (waiting for jenkins)

PR #6080, opened 2017-10-06
Settings Dialog: Fix display name compilation for macOS #6078

guruz:


guruz:
(Let's see if Jenkins linux compile goes through..)

ckamm:
Looks ok, though probably I'd prefer a `settingsdialogcommon.h` for consistency.

guruz:
@mrow4a On macOS we use a special implementation that feels more native to macOS.
Maybe this is not needed anymore with newer Qt versions though

CC @shadone

PR #6087, opened 2017-10-09
Speed up local discovery by not doing it

ckamm:
Use the data from the db instead and only check the file system when a file watcher event suggests it's necessary.

Since we're not confident that file watcher events are good enough, we still do a full local discovery every hour for now. This also doesn't get rid of the initial sync after client startup. But incremental syncs should be sped up significantly.

See #2297

### DIFF
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -547,11 +547,26 @@ bool SyncJournalDb::checkConnect()
         return sqlFail("prepare _getFileRecordQueryByFileId", *_getFileRecordQueryByFileId);
     }
 
+    // This query is used to skip discovery and fill the tree from the
+    // database instead
     _getFilesBelowPathQuery.reset(new SqlQuery(_db));
     if (_getFilesBelowPathQuery->prepare(
             GET_FILE_RECORD_QUERY
-            " WHERE path > (?1||'/') AND path < (?1||'0') ORDER BY path||'/' ASC")) {
+            " WHERE path > (?1||'/') AND path < (?1||'0')"
+            // We want to ensure that the contents of a directory are sorted
+            // directly behind the directory itself. Without this ORDER BY
+            // an ordering like foo, foo-2, foo/file would be returned.
+            // With the trailing /, we get foo-2, foo, foo/file. This property
+            // is used in fill_tree_from_db().
+            " ORDER BY path||'/' ASC")) {
         return sqlFail("prepare _getFilesBelowPathQuery", *_getFilesBelowPathQuery);
+    }
+
+    _getAllFilesQuery.reset(new SqlQuery(_db));
+    if (_getAllFilesQuery->prepare(
+            GET_FILE_RECORD_QUERY
+            " ORDER BY path||'/' ASC")) {
+        return sqlFail("prepare _getAllFilesQuery", *_getAllFilesQuery);
     }
 
     _setFileRecordQuery.reset(new SqlQuery(_db));
@@ -704,6 +719,7 @@ void SyncJournalDb::close()
     _getFileRecordQueryByInode.reset(0);
     _getFileRecordQueryByFileId.reset(0);
     _getFilesBelowPathQuery.reset(0);
+    _getAllFilesQuery.reset(0);
     _setFileRecordQuery.reset(0);
     _setFileRecordChecksumQuery.reset(0);
     _setFileRecordLocalMetadataQuery.reset(0);
@@ -1133,16 +1149,23 @@ bool SyncJournalDb::getFilesBelowPath(const QByteArray &path, const std::functio
     if (!checkConnect())
         return false;
 
-    _getFilesBelowPathQuery->reset_and_clear_bindings();
-    _getFilesBelowPathQuery->bindValue(1, path);
+    // Since the path column doesn't store the starting /, the getFilesBelowPathQuery
+    // can't be used for the root path "". It would scan for (path > '/' and path < '0')
+    // and find nothing. So, unfortunately, we have to use a different query for
+    // retrieving the whole tree.
+    auto &query = path.isEmpty() ? _getAllFilesQuery : _getFilesBelowPathQuery;
 
-    if (!_getFilesBelowPathQuery->exec()) {
+    query->reset_and_clear_bindings();
+    if (query == _getFilesBelowPathQuery)
+        query->bindValue(1, path);
+
+    if (!query->exec()) {
         return false;
     }
 
-    while (_getFilesBelowPathQuery->next()) {
+    while (query->next()) {
         SyncJournalFileRecord rec;
-        fillFileRecordFromGetQuery(rec, *_getFilesBelowPathQuery);
+        fillFileRecordFromGetQuery(rec, *query);
         rowCallback(rec);
     }
 
--- a/src/csync/csync_update.cpp
+++ b/src/csync/csync_update.cpp
@@ -435,28 +435,31 @@ static bool fill_tree_from_db(CSYNC *ctx, const char *uri)
 {
     int64_t count = 0;
     QByteArray skipbase;
-    auto rowCallback = [ctx, &count, &skipbase](const OCC::SyncJournalFileRecord &rec) {
-        /* When selective sync is used, the database may have subtrees with a parent
-         * whose etag (md5) is _invalid_. These are ignored and shall not appear in the
-         * remote tree.
-         * Sometimes folders that are not ignored by selective sync get marked as
-         * _invalid_, but that is not a problem as the next discovery will retrieve
-         * their correct etags again and we don't run into this case.
-         */
-        if( rec._etag == "_invalid_") {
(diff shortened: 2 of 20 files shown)


ckamm:
This seemed to work in simple tests but clearly needs a lot more testing.

ckamm:
The existing unittests work again now. Next steps: Add new unittests that verify it works.

src/gui/folder.cpp
```diff
@@ -448,12 +448,33 @@ int Folder::slotWipeErrorBlacklist()
 
 void Folder::slotWatchedPathChanged(const QString &path)
 {
+    if (!path.startsWith(this->path())) {
+        qCDebug(lcFolder) << "Changed path is not contained in folder, ignoring:" << path;
+        return;
+    }
+
+    auto relativePath = path.mid(this->path().size());
```
ogoffart: s/mid/midRef/

src/gui/folder.cpp
```diff
@@ -783,6 +824,21 @@ void Folder::slotSyncFinished(bool success)
         journalDb()->setSelectiveSyncList(SyncJournalDb::SelectiveSyncWhiteList, QStringList());
     }
 
+    // bug: This function uses many different criteria for "sync was successful" - investigate!
+    if ((_syncResult.status() == SyncResult::Success
+            || _syncResult.status() == SyncResult::Problem)
```
ogoffart: So it is problem even for soft errors?

src/gui/folder.cpp
```diff
@@ -783,6 +824,21 @@ void Folder::slotSyncFinished(bool success)
...
+        && success) {
+        if (_engine->lastLocalDiscoveryStyle() == LocalDiscoveryStyle::FilesystemOnly) {
+            _timeSinceLastFullLocalDiscovery.start();
+        } else if (_syncResult.status() == SyncResult::Success) {
+            // Optimization opportunity: Currently we wipe this only
+            // on 100% successful sync. ('Problem' syncs could have the touched
+            // file trigger a soft-error-and-retry scenario and wiping this list
+            // would mean the file isn't retried)
+            _locallyModifiedDirs.clear();
```
ogoffart: What if there were change hapenning during the sync!

Maybe we should clear it as we start the sync, and if the sync fails for any problem, we add them back.

src/gui/folder.h
```diff
@@ -365,6 +378,21 @@ private slots:
      * path.
      */
     bool _saveBackwardsCompatible;
+
+    /**
+     * Watches this folder's local directory for changes.
+     *
+     * Created by registerFolderWatcher(), triggers slotWatchedPathChanged()
+     */
+    FolderWatcher *_folderWatcher = 0;
```
ogoffart: ohoh!  you are putting the FolderWatcher back in the Folder itself :-)  cc/ @dragotin

src/gui/folder.cpp
ckamm: @ogoffart Yes, great catch! That means we really need to remove files from this list as the individual jobs report success even in the initial version. At least that also helps with the issue regarding SoftErrors.

src/common/syncjournaldb.cpp
```diff
@@ -550,7 +550,7 @@ bool SyncJournalDb::checkConnect()
     _getFilesBelowPathQuery.reset(new SqlQuery(_db));
     if (_getFilesBelowPathQuery->prepare(
             GET_FILE_RECORD_QUERY
-            " WHERE path > (?1||'/') AND path < (?1||'0') ORDER BY path||'/' ASC")) {
+            " WHERE ?1 == '' OR (path > (?1||'/') AND path < (?1||'0')) ORDER BY path||'/' ASC")) {
```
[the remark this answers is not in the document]
ckamm: Good catch, adding that term does indeed break index use (checked with `EXPLAIN QUERY PLAN`)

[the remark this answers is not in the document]
ckamm: The problem is that our database doesn't keep paths with a initial /, so `path > (''||'/') AND path < (''||'0'')` matches nothing. Something like `path > '' AND path < 'last-unicode-character'` would work, but I think we just need a different query for clarity.

ckamm:
@ogoffart Significantly updated!

src/common/syncjournaldb.cpp
```diff
@@ -554,6 +554,13 @@ bool SyncJournalDb::checkConnect()
         return sqlFail("prepare _getFilesBelowPathQuery", *_getFilesBelowPathQuery);
     }
 
+    _getAllFilesQuery.reset(new SqlQuery(_db));
+    if (_getAllFilesQuery->prepare(
+            GET_FILE_RECORD_QUERY
+            " ORDER BY path||'/' ASC")) {
```
ogoffart: Why do we concatenate '/' to the path while ordering?

src/csync/csync_update.cpp
```diff
@@ -522,6 +525,25 @@ int csync_ftw(CSYNC *ctx, const char *uri, csync_walker_fn fn,
...
+      if (*local_uri == '/')
+          ++local_uri;
+      db_uri = local_uri;
+      do_read_from_db = true;
+
+      // Minor bug: local_uri doesn't have a trailing /. Example: Assume it's "d/foo"
+      // and we want to check whether we should read from the db. Assume "d/foo a" is
+      // in locally_touched_dirs. Then this check will say no, don't read from the db!
+      // (because "d/foo" < "d/foo a" < "d/foo/bar")
+      auto it = ctx->locally_touched_dirs.lower_bound(QByteArray(local_uri));
```
ogoffart: The QByteArray allocation could be avoided in C++14, there is an overload of lower_bound that can work for types for which the comparator is transparent.

But that would imply a change to the makefile to enable C++14.
Maybe it is not needed as it should not happen so often anyway.

src/gui/folder.cpp
```diff
@@ -645,6 +656,37 @@ void Folder::startSync(const QStringList &pathList)
...
+    if (_folderWatcher && _folderWatcher->isReliable()
+        && _timeSinceLastFullLocalDiscovery.isValid()
+        && (fullLocalDiscoveryInterval < 0
+               || _timeSinceLastFullLocalDiscovery.elapsed() < fullLocalDiscoveryInterval)) {
+        qCInfo(lcFolder) << "Allowing local discovery to read from the database";
+        _engine->setLocalDiscoveryOptions(LocalDiscoveryStyle::DatabaseAndFilesystem, _localDiscoveryPaths);
+
+        if (lcFolder().isDebugEnabled()) {
+            QByteArrayList paths;
+            for (auto &path : _localDiscoveryPaths) {
```
ogoffart: (minor:  could be done with  std::copy)

src/gui/folder.cpp
```diff
@@ -645,6 +656,37 @@ void Folder::startSync(const QStringList &pathList)
...
+                paths.append(path);
+            }
+            qCDebug(lcFolder) << "local discovery paths: " << qUtf8Printable(paths.join(", "));
```
ogoffart: can you not just qCDebug << paths?   why do we need to convert it to a QByteArray first?
Also , qUtf8Priontable  is for QString.   When you have a QByteArray you can just use constData().

src/libsync/configfile.h
```diff
@@ -72,6 +72,9 @@ class OWNCLOUDSYNC_EXPORT ConfigFile
     /* Force sync interval, in milliseconds */
     quint64 forceSyncInterval(const QString &connection = QString()) const;
 
+    /** Interval within which full local discovery is required, -1 to disable */
+    qint64 fullLocalDiscoveryInterval() const;
```
ogoffart: What's the unit?
I know this is not yet your coding style.  But in the future it would be nice to use things like std::chrono::milliseconds as a return type to make it explicit.

src/csync/csync_update.cpp
ckamm: I'll add a note about this for C++14, but won't enable it in this PR.

src/libsync/configfile.h
ckamm: Agreed!

ckamm:
@ogoffart Resolved remaining issues

src/csync/csync_update.cpp
```diff
@@ -522,6 +525,26 @@ int csync_ftw(CSYNC *ctx, const char *uri, csync_walker_fn fn,
...
+      if (*local_uri == '/')
+          ++local_uri;
+      db_uri = local_uri;
+      do_read_from_db = true;
+
+      // Minor bug: local_uri doesn't have a trailing /. Example: Assume it's "d/foo"
+      // and we want to check whether we should read from the db. Assume "d/foo a" is
+      // in locally_touched_dirs. Then this check will say no, don't read from the db!
+      // (because "d/foo" < "d/foo a" < "d/foo/bar")
+      // C++17: Could skip the conversion to QByteArray here.
```
ogoffart: c++14

ckamm: yes...

Issue #5989, opened 2017-09-01
[Regression] In 2.4 "Credentials Wrong" state does not retrigger the password-asking form


SamuAlfageme:
### Expected behavior
After receiving the 401: Unauthorized from the server (`<s:message>No public access to this resource., Username or password was incorrect, Username or password was incorrect</s:message>`) client 2.3.3 displays again the password dialog. 

### Actual behavior
The status on the account tab stays `Connecting to <server> as <username>...` until it times out and switches to `Signed out...`: 

```
[ warning sync.connectionvalidator ]:	******** Password is wrong! QNetworkReply::NetworkError(OperationCanceledError) "Operation canceled"
[ info gui.account.state ]:	AccountState connection status change:  "Credentials not ready" -> "Credentials Wrong"
```

### Steps to reproduce
1. Log out from any client's account and try to log in with the wrong password

ckamm:
@ogoffart I've debugged this using mitmproxy, and what seems to be happening is this: When the password is rejected, we *do* call `HttpCredentialsGui::askFromUser` again, but the GET request (to check whether the server can do oauth) stalls. After 30s the `::abort` timer triggers.

Do you have any ideas about why the request would not be performed? (there's no entry in mitmproxy for it)

ogoffart:
@ckamm: I have no idea,  if askFromUser is called, then normally the GET should be done. Some debuging might be in order.

ckamm:
Yeah, I'll look at it more deeply today. It's confusing that the GET request doesn't even go out. Also, I think this might deserve code sharing with DetermineAuthType - currently it doesn't deal with redirections properly.

ckamm:
Debugging has yielded the following (thanks to @guruz, @jturcotte, @ogoffart for suggestions):
* When `HttpCredentialsGui::askFromUser`is called from user action, everything works, including the GET request to determine the auth type.
* When the user provides an incorrect password, the PROPFIND fails and `askFromUser` is called again. That is good.
* However, this time, the `sendRequest("GET", ...)` in there fails.
* This seems likely to be caused by initiating a new request directly in the call stack that's processing the PROPFIND's `QNetworkReply::finished`. Evidence: If `askFromUser` is delayed to the event loop by using `invokeMethod(QueuedConnection)`, the GET request is processed as expected.

Possibly a QNAM bug in Qt 5.7.1?

ckamm:
[...]

PR #6091, opened 2017-10-11
HttpCreds: Fix retry after wrong password

ckamm:
This is an ugly solution.

See  #5989

@ogoffart Do you know if timers are stable in this way? I think we'll have to look very carefully at how the credential / account machinery works together in this case.

### DIFF
--- a/src/gui/creds/httpcredentialsgui.cpp
+++ b/src/gui/creds/httpcredentialsgui.cpp
@@ -32,10 +32,11 @@ namespace OCC {
 
 void HttpCredentialsGui::askFromUser()
 {
-    // Unfortunately there's a bug that doesn't allow us to send the "is this
-    // OAuth2 or Basic auth?" GET request directly. Scheduling it for the event
-    // loop works though. See #5989.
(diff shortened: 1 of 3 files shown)


src/gui/creds/httpcredentialsgui.cpp
```diff
@@ -32,10 +32,11 @@ namespace OCC {
...
 {
-    // Unfortunately there's a bug that doesn't allow us to send the "is this
-    // OAuth2 or Basic auth?" GET request directly. Scheduling it for the event
-    // loop works though. See #5989.
-    QMetaObject::invokeMethod(this, "askFromUserAsync", Qt::QueuedConnection);
+    // This function can be called from AccountState::slotInvalidCredentials,
+    // which (indirectly, through HttpCredentials::invalidateToken) schedules
+    // a cache wipe of the qnam. We can only execute a network job again once
+    // the cache has been cleared, otherwise we'd interfere with the job.
+    QTimer::singleShot(100, this, SLOT(askFromUserAsync()));
```
ogoffart: you should use the functoin pointer syntax

ckamm: @ogoffart updated

Issue #6086, opened 2017-10-09
UI: allow sorting columns in Not Synced Activity Tab


moscicki:
Wouldn't it be useful to allow sorting of columns when clicked on the header? We have got such suggestions from our users.

![image](https://user-images.githubusercontent.com/621965/31338174-7b043d4a-acfe-11e7-9e40-463c8e6b3253.png)

hodyroff:
Yes! Not sure if we can still make this happen quickly for 2.4.0 @ckamm @michaelstingl the heavy users will like it a lot.

ckamm:
I see that sorting would be nice sometimes but there is a complication: We've recently started having a category of "overall" issues (like insufficent remote or local storage) that have no "File" entry and are always sorted to the top. Enabling column sorting (which is easy) would break that.

To avoid users accidentally breaking their error view I'd suggest we encode that "native" sorting into the "Time" row and switch back to that sorting on restart or on window-hide.

Note that newer versions already provide filters by account/folder as well as for warnings or ignored files that should help a bit with locating the relevant issues.

ckamm:
PR: https://github.com/owncloud/client/pull/6093

guruz:
@hodyroff @moscicki please test with tomorrow's 2.4 nightly and report back here to close issue.

guruz:
https://software.opensuse.org/download.html?project=isv%3AownCloud%3Acommunity%3Anightly&package=owncloud-client

michaelstingl:
@SamuAlfageme please check

ckamm:
@moscicki Yes, probably a regression due to https://github.com/owncloud/client/issues/5954 - a patch to fix it is currently in review.

ckamm:
The PR is merged now and the next nightly will contain the patch.

guruz:
@moscicki Works?

moscicki:
Sorting works OK. @giorgosalex was testing it. 

The status of connection wizard fix - I don't know. @giorgosalex could you re-check/comment on this?

guruz:
About the connection wizard: Please try with a nightly from TOMORROW, we have/had some things open with regards to redirects, auth detection and wizard.

Then, open a **new issue** and assign 2.4.0 milestone to it.

PR #6093, opened 2017-10-12
Activity: Allow sorting of issues and protocol #6086

ckamm:
The issues tab uses custom ordering where overall and summary sync
issues are displayed first. This ordering is preserved by creating
special sorting logic for the "time" column.

It needed special handling anyway, since sorting by time-string would
have yielded incorrect results.

### DIFF
--- a/src/gui/protocolwidget.h
+++ b/src/gui/protocolwidget.h
@@ -35,6 +35,21 @@ namespace Ui {
 class Application;
 
 /**
+ * A QTreeWidgetItem with special sorting.
+ *
+ * It allows items for global entries to be moved to the top if the
+ * sorting section is the "Time" column.
+ */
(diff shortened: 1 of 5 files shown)


src/gui/protocolwidget.h
```diff
@@ -35,6 +35,21 @@ namespace Ui {
...
+ * It allows items for global entries to be moved to the top if the
+ * sorting section is the "Time" column.
+ */
+class SortedTreeWidgetItem : public QTreeWidgetItem
+{
+public:
+    using QTreeWidgetItem::QTreeWidgetItem;
+
+private:
+    bool operator<(const QTreeWidgetItem &other) const;
```
ogoffart: Please add the override keyword.

I was wondering how can this could even work until i realized the QTreeWidgetItem operator was virtual.

PR #6101, opened 2017-10-17
owncloudcmd: Fix timestamps, Fix --logdebug

guruz:
We did not set a log handler so there were no timestamps.
The --debug didn't have an effect, let's use --logdebug like in GUI version.
(Command line always outputs some log)

Found in owncloud/documentation#3436


guruz:
@ogoffart Changed the code. for some reason github web still shows the old patch and not 4f619bf998130774760368a60c0c0d19cdf735c3

PR #6102, opened 2017-10-17
SettingsDialogMac: Fix account display name #6078

guruz:


guruz:
I didnt' want to invest more time in this as I think at some point we could refactor this beast :-/

PR #6104, opened 2017-10-18
Download: Apply metadata-only changes to locked files

ckamm:
The intention is to work around a server bug in 8.2.11 where the etag
and mtime would suddenly update spontaneously after an upload.

This introduces *incorrect* behavior. The only good thing is that
the surface area is small: it applies only to metadata-only changes
for locked files.

See owncloud/enterprise#2288

I'd consider this a proof-of-concept and we need to debate whether we really want to go down this path!

### DIFF
--- a/src/libsync/propagatedownload.cpp
+++ b/src/libsync/propagatedownload.cpp
@@ -724,9 +724,14 @@ void PropagateDownloadFile::downloadFinished()
 
     // In case of conflict, make a backup of the old file
     // Ignore conflicts where both files are binary equal
-    bool isConflict = _item->_instruction == CSYNC_INSTRUCTION_CONFLICT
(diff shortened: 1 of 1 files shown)


ogoffart:
:+1:   Looks good.

But do we know that owncloud 10 actually fixes this bug?

ckamm:
@ogoffart I don't know for sure that this bug is fixed. But since we haven't seen something like this reported in a while it seems unlikely that this would still be around. Also, I vaguely remember bugs like this.

guruz:
@tomneedham wrote in https://github.com/owncloud/enterprise/issues/2288
> It is probably easier to patch the server than update all the clients...

ckamm:
We're only going to do this if there's no other option. Closing for now.

PR #6106, opened 2017-10-19
Wizard: Resolve url/ redirects only if url/status.php not found

ckamm:
Unfortunately checking the base-url for redirects in all cases lead
to incorrect behavior in some SAML/OAuth2 edge cases.

This new iteration checks the base url for redirects only if the
standard CheckServerJob can't reach the server. That way the 2.3
behavior is only changed in cases that would have lead to errors.

See #5954

### DIFF
--- a/src/gui/owncloudsetupwizard.cpp
+++ b/src/gui/owncloudsetupwizard.cpp
@@ -45,7 +45,7 @@ OwncloudSetupWizard::OwncloudSetupWizard(QObject *parent)
     , _remoteFolder()
 {
     connect(_ocWizard, &OwncloudWizard::determineAuthType,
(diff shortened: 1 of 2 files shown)


SamuAlfageme:
I think:

https://github.com/owncloud/client/blob/224d9194256c91631ea63adb1ec7fd35974d1335/src/gui/owncloudsetupwizard.cpp#L153 

is causing:

```
10-20 10:16:01:080 [ warning default ]:	QMetaObject::invokeMethod: No such method OCC::OwncloudSetupWizard::slotContinueDetermineAuth()
```

On every wizard run.

ckamm:
Fixed (and double-checked the other renames)

Issue #5889, opened 2017-07-11
Client 2.3.3 release


guruz:
Some weeks before the release:
* [x] Check if we should update the bundled sqlite3 (https://github.com/owncloud/client/tree/master/src/3rdparty/sqlite3)
* [x] Check if we should update Sparkle on build machine (https://github.com/sparkle-project/Sparkle/releases)
* [x] Ensure NSIS is up to date on the build machine
* [x] Ensure up-to-date dependencies (e.g. [latest Qt version](http://qt-project.org/downloads#qt-lib) is installed on the machine and picked up (cmake output)
* [x] Ensure the crash reporter server is up
* [x] Check crash reporter for bad crashes
* [x] Ensure Windows Overlay DLLs are rebuilt
* [x] Check nightly builds are up and running, that is Jenkins jobs ownCloud-client-linux, ownCloud-client-osx and ownCloud-client-win32 all green.
* [x] Ensure Linux nightlies are built too for all distros https://build.opensuse.org/package/show/isv:ownCloud:community:nightly/owncloud-client
* [ ] Build branded clients through the scripting machine and smoke test one or two branded clients (especially with predefined url)
* [x] Upload a nightly build of the windows version to virustotal.com
  * Contact AV vendors whom's engine reports a virus
* [x] Documentation should be online before the release http://doc.owncloud.org/desktop/1.X/
* [x] QA goes over https://github.com/owncloud/mirall/wiki/Testing-Scenarios
* [x] Make sure to have `client/ChangeLog` updated
 * use `git log --format=oneline v<lastrelease>...master` if your memory fails you
* [x] check if enterprise issues are fixed

One week before the release:
* [x] Communicate the release schedule on mailinglist release-coordination@owncloud.com. Give a high level overview of the upcoming new features, changes etc.
* [ ] Ensure marketing is aware (marketing@owncloud.com) and prepared for the release (social, .com website, cust. communications)
* [x] Inform GCX knows the next version is about 1 week out (gcx@owncloud.com)

For all Betas and RCs:
* [x] Branch off a release branch called VERSION-rcX or VERSION-betaX  (without v, v is for tags)
* [x] Edit ```VERSION.cmake``` to set the suffix to beta1, beta2 etc. Commit the result to the release branch only
* [x] Create build for Windows using rotor job owncloud-client-win32 (uncheck the "nightly build" checkbox, check the "sign package" checkboxes) both themes 'ownCloud' and 'testpilotcould'
[...]

PR #6109, opened 2017-10-19
Adding post-release checks to template

jnweiger:
Copied 3 items from https://github.com/owncloud/client/issues/5889
(more items needed to copy??)


.github/release_template.md
```diff
@@ -105,4 +105,10 @@ On Release Day (for final release):
 * [ ] Days later: Update the updater script ```clientupdater.php``` (check the crash reporter if auto update is a good idea or we need a new release)
```
michaelstingl: redundancy?

Issue #6096, opened 2017-10-14
 Desktop Client doesn’t move files rather then download them 


exzel123:
### Expected behaviour
Files should be correctly moved by the sync client.

### Actual behaviour
In some cases the files appear again in the original folder

### Steps to reproduce
User 1 creates a folder in the root directory named AAA.
User 1 creates a subfolder in AAA named BBB.

User 1 shares AAA with user 2.
User 1 shares BBB with user 2 and 3.
User 3 creates a subfolder in BBB.
In folder BBB are some files.

If user 3 moves the files from folder BBB to the subfolder in BBB then the sync client makes a mistake with the syncronization.
The files in the folder BBB at user 2 are not moved. They are uploaded again at the original folder in the cloud by user 2. 

And after a while they appear at the folder BBB and the subfolder at user 3. 

### Server configuration
Operating system:

CloudLinux

Web server: 
Apache
Database:
mysql

PHP version:
?
ownCloud version:
10.0.3
Storage backend (external storage):

### Client configuration
Client version:
2.3.3 (build) 8250. Also tested with the 2.4 Alpha.

Operating system:
Windows 7 64bit

OS language:
German

hodyroff:
Can you reproduce it with logfiles from user 2 and user 3 as well as the server side?

exzel123:
I can sent the logfile from the server. But it's almost impossible for me to get the protokoll from the clients because the owncloud client freezes if I open the logfile windows.

exzel123:
But it works correctly in the WebClient. The SyncClient did cause this error.

PVince81:
@exzel123 you should be able to open the log file directly from disk instead of looking at the OC client.

Does any of the involved folders (source/target) look weird in the UI, like some appearing twice ? (in case of filecache breakages)

Check owncloud.log from the time of the MOVE for errors.

guruz:
What @PVince81 meant is https://doc.owncloud.org/desktop/2.3/troubleshooting.html#saving-files-directly for the client log files to be saved in a file instead of our log window.

(owncloud.log is server side)

exzel123:
I replaced the domain in the logfiles but the rest is the same.

Thank you.

[Clientlog.txt](https://github.com/owncloud/client/files/1394035/Clientlog.txt)
[Serverlog.log](https://github.com/owncloud/client/files/1394036/Serverlog.log)

exzel123:
And the file which was moved and appeard at the original directory is called "Neues Textdokument"

[...]

PR #6114, opened 2017-10-23
Reconcile: Rename handling fixes: duplicate file ids

ckamm:
When users share the same tree several times (say A/ and A/B/ are both
shared) the remote tree can have several entries that have the same
file id. This needs to be respected in rename detection.

Also adds several tests and fixes for issues noticed during testing.

See #6096


src/csync/csync_reconcile.cpp
```diff
@@ -310,7 +345,13 @@ static int _csync_merge_algorithm_visitor(csync_file_stat_t *cur, CSYNC * ctx) {
                 break;
             case CSYNC_INSTRUCTION_IGNORE:
                 cur->instruction = CSYNC_INSTRUCTION_IGNORE;
-            break;
+                break;
+            case CSYNC_INSTRUCTION_EVAL_RENAME:
+                // Local modification and remote rename:
+                // Typically a parent directory was renamed!
+                // The locally touched file should be uploaded.
+                cur->instruction = CSYNC_INSTRUCTION_NEW;
```
ogoffart: Are you sure about the comment?
Here is how i think it can happen:

on the client:
```
echo "foo" > foo  # new file
```
on the server:
```
mv bar foo
```

I suppose this should be a conflict.
Currently this should be handled as such when EVAL_RENAME is handled.

src/csync/csync_update.cpp
```diff
@@ -297,41 +297,60 @@ static int _csync_detect_update(CSYNC *ctx, std::unique_ptr<csync_file_stat_t> f
...
               if (fs->type == CSYNC_FTW_TYPE_DIR) {
-                  csync_rename_record(ctx, base._path, fs->path);
-              } else {
-                  if( base._etag != fs->etag ) {
-                      /* CSYNC_LOG(CSYNC_LOG_PRIORITY_DEBUG, "ETags are different!"); */
-                      /* File with different etag, don't do a rename, but download the file again */
-                      fs->instruction = CSYNC_INSTRUCTION_NEW;
+                  // If the same folder was already renamed by a different entry,
+                  // skip to the next candidate
+                  if (ctx->renames.folder_renamed_to.find(base._path) != ctx->renames.folder_renamed_to.end()) {
```
ogoffart: i believe there is a count(const K&) function that is more appropriate

test/testsyncmove.cpp
```diff
@@ -0,0 +1,414 @@
...
+        auto mtime = fakeFolder.remoteModifier().find("B/b2")->lastModified;
+        fakeFolder.localModifier().rename("B/b2", "B/b2m");
+        fakeFolder.localModifier().appendByte("B/b2m");
+        fakeFolder.localModifier().setModTime("B/b2m", mtime);
+        QVERIFY(fakeFolder.syncOnce());
+        QCOMPARE(fakeFolder.currentLocalState(), remoteInfo);
+        QCOMPARE(nPUT, 3);
+        QCOMPARE(nDELETE, 3);
+
+        // Move-and-change, content only -- c1 has no checksum, so we fail to detect this!
```
ogoffart: (already existing:  this should probably be a EXPECTED_FAIL,  rather than enforcing the "wrong" behavior)

test/testsyncmove.cpp
```diff
@@ -0,0 +1,414 @@
...
+        }
+
+        // Touch+Move on same side
+        local.rename("A/a2", "A/a2m");
+        local.setContents("A/a2m", 'A');
+        remote.rename("B/b2", "B/b2m");
+        remote.setContents("B/b2m", 'A');
+        QVERIFY(fakeFolder.syncOnce());
+        QCOMPARE(fakeFolder.currentLocalState(), fakeFolder.currentRemoteState());
+        QCOMPARE(nMOVE, 1);
```
ogoffart: maybe also compute the amount of GET/PUT/DELETE  to make sure that there as not too many.

src/csync/csync_reconcile.cpp
ckamm: I've changed things a bit. The problem is this case:
```
// local
echo "foo" >> A/f
// remote
mv A B
```
then A/f was left with EVAL since B/f is EVAL_RENAME. When we went to the remote tree, there's no other since B/f doesn't exist in local. Then the rename doesn't work and the touched A/f isn't uploaded.

I've added your test case and that just works. The reconcile doesn't care if the instruction is NEW or EVAL.

ckamm:
Updated with more tests and small amendments.

src/csync/csync_reconcile.cpp
```diff
@@ -310,10 +344,19 @@ static int _csync_merge_algorithm_visitor(csync_file_stat_t *cur, CSYNC * ctx) {
...
                 cur->instruction = CSYNC_INSTRUCTION_IGNORE;
-            break;
+                break;
             default:
                 break;
             }
+            // Ensure we're not leaving discovery-only instructions
+            // in place. This can happen, for instance, when other's
+            // instruction is EVAL_RENAME because the parent dir was renamed.
+            // NEW is safer than EVAL because it will end up with
```
ogoffart: I guess you meanst s/EVAL/SYNC/g  (also next line)

ckamm: No. I mean that changing discovery's EVAL to NEW is generally safe in reconcile, since all `switch` clauses typically check for EVAL/NEW. Switching to SYNC may have side effects that I haven't investigated. The effect for propagation is the same. (Why do we even distinguish NEW/SYNC? For reporting?)

Issue #6098, opened 2017-10-16
[Overlays] Use "Share" overlays 


SamuAlfageme:
While checking https://github.com/owncloud/client/issues/6094 I found out that the "Share" overlay icon is not being used in the shell integations neither in Linux ~~nor in Windows~~, while in macOS is correctly set right after sharing from the desktop/receiving a shared file. 

### Expected behavior
When a file is shared, the overlay icon should be [the right one](https://github.com/owncloud/client/blob/master/shell_integration/icons/nopadding/SVG/shared.svg)

### Actual behavior
The "OK" overlay is the one displayed on the file explorer. 

### Client configuration
Client version: 2.3.3/2.4.0 (current master)
Operating system: Ubuntu 16.04(Nautilus)/Windows 7

ogoffart:
I guess the problem is that when sharing, the database still has the old cache information weather the file is cached or not.  I do not believe sharing change the etag of the parent folder, which means we won't notice the change until there is an actual change in that folder.

guruz:
> I do not believe sharing change the etag of the parent folder, which means we won't notice the change until there is an actual change in that folder.

CC @DeepDiver1975 @PVince81

PVince81:
We already fixed etag propagation a while ago with OC 9.0. Every mount point has an etag and the etag of the parents is computed based on all child etags. So if anything changes within a received share, all receivers will have their etag changes.

If the etag did not change then it's a bug.

SamuAlfageme:
@ogoffart yeah, I think it differs since setting up folders with shared items from scratch still displays ✅ (while in macOS you get the shared one)

Otherwise, the behavior with the parent etags is erratic, agreed. Could we retrigger an individual PROPFIND after editing shared properties of a file?

ckamm:
@SamuAlfageme I'll check this out on linux/nautilus.

SamuAlfageme:
@ckamm sorry, I think I'm not being very clear. [This is what I see on Ubuntu 16.04 with Nautilus](https://i.imgur.com/sUuWSQp.gifv) - note the 2 files being shared by link on the server after recreated locally and getting the wrong overlay. Is it different to what happens on your side?

> When my user shares a file with someone, the icon doesn't change. (possibly because I'm testing against an older server right now, this only works with servers >=10)

[...]

PR #6115, opened 2017-10-23
Sharing: fix overlay icons directly after sharing a file

ogoffart:
See commit message for more information.
The two first commit are just code cleanup
The other commit fix the issue #6098

This include a workaround of the fact that the server don't update the etags.

### DIFF
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -841,6 +841,11 @@ void Folder::slotTransmissionProgress(const ProgressInfo &pi)
 // a item is completed: count the errors and forward to the ProgressDispatcher
 void Folder::slotItemCompleted(const SyncFileItemPtr &item)
 {
+    if (item->_instruction == CSYNC_INSTRUCTION_NONE || item->_instruction == CSYNC_INSTRUCTION_UPDATE_METADATA) {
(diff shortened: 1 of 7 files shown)


src/gui/folder.cpp
```diff
@@ -841,6 +841,11 @@ void Folder::slotTransmissionProgress(const ProgressInfo &pi)
 // a item is completed: count the errors and forward to the ProgressDispatcher
 void Folder::slotItemCompleted(const SyncFileItemPtr &item)
 {
+    if (item->_instruction == CSYNC_INSTRUCTION_NONE || item->_instruction == CSYNC_INSTRUCTION_UPDATE_METADATA) {
+        // We only care about the updates that deserve to be shown in the UI
+        return;
+    }
```
jturcotte: Including `CSYNC_INSTRUCTION_NONE` sounds wrong, would it be possible to check at the emit point?

ogoffart: CSYNC_INSTRUCTION_NONE should never happen. But i'm putting it there as it goes hand to hand with UPDATE_METADATA. (Should the sync engine change, and emit the signal with CSYNC_INSTRUCTION_NONE in the future, we probably want to filter it there anyway)
I could remove it from this check if you think this is confusing.

Issue #6125, opened 2017-10-26
Credential store keys are not namespaced


ckamm:
@guruz pointed out https://github.com/frankosterfeld/qtkeychain/issues/105 and I've verified it on Windows 10:

When we store credentials in the credential store on Windows through QtKeychain we choose the key similarly to other platforms. However, on Windows the library doesn't automatically add the application name or organization name.

Effects
* Keys like "user:http://localhost:0" show up in "Generic Credenticals" section of the Windows Credential Manager
* Different branded clients may share credentials if the same user/server is used. (should be exceedingly rare)

MSDN says the following about the key: 

> If the Type is CRED_TYPE_GENERIC, this member should identify the service that uses the credential in addition to the actual target. Microsoft suggests the name be prefixed by the name of the company implementing the service. Microsoft will use the prefix "Microsoft". Services written by Microsoft should append their service name, for example Microsoft_RAS_TargetName.

(https://msdn.microsoft.com/en-us/library/windows/desktop/aa374788(v=vs.85).aspx)

Suggestion: Change the key used on Windows to be "$AppName_$oldkey". This is a good opportunity since we're migrating http credential keys anyway.

SamuAlfageme:
It seems to be working just fine now (tested in W7/8.1). Nice catch. 

| 2.3.3 | 2.4.0 |
|---|---|
| ![screenshot 2017-11-02 13 42 30](https://user-images.githubusercontent.com/2644445/32327248-27f8c996-bfd6-11e7-9d2c-e48841613b1a.png) | ![screenshot 2017-11-02 13 34 51](https://user-images.githubusercontent.com/2644445/32327250-281b1028-bfd6-11e7-8d7d-9a83f4e28aa5.png) |

Some notes:
- After https://github.com/owncloud/client/issues/5830 we have to be extra-careful and check for the trailing slash to be present when creating the keychain/cred.store entry in all cases - otherwise, the account id might be taken by mistake as the port no.
- Also, in some upgrade scenario, the standard SSL port was included in the instance's URL everywhere - (might be unrelated - I'm trying to find out the conditions/steps again, this is merely aesthetic though)

<p align="center">
  <img src="https://user-images.githubusercontent.com/2644445/32328141-525f7cd6-bfd9-11e7-818b-44bf2e1313fa.png"/>
</p>

👍

Issue #1416, opened 2014-02-11
Investigate if we should have QtKeychain as a git submodule


guruz:
This allows way easier compilation for potential contributors.

Questions to solve:
- Packaging? My personal suggestion is that we compile it as `owncloudqtkeychain.dylib` and don't use any upstream package.

guruz:
Reopening as per discussion @jnweiger @michaelstingl @SamuAlfageme @ogoffart 

Related #2558

jnweiger:
Current status: We don't use any distro-packages of qtkeychain. 
We always provide our own, as we want to have the same version (currently 0.7.0) across all platforms.

jturcotte:
For the record, we can have the submodule without a fork.

We can point the submodule to https://github.com/frankosterfeld/qtkeychain.git, pin the SHA1 of the version that we need and do a pull request upstream if we ever need to change anything.

guruz:
I removed this from the milestone as @jnweiger @dschmidt anyway did the linux 5.6.2 without this change.

I still think it would be good to do this to have it easier for developers to set up a build environment

ogoffart:
Note that we could also use CMake's ExternalProject feature which is capabable fetch git repositories.

fmoc:
We use craft to manage almost all the dependencies including QtKeychain. Only for the Linux repos, we still have system packaging on OBS, but the long-term goal is to deploy AppImages only to Linux (or alternatively system packages based on the AppDirs contained in the AppImages, thus using craft for those as well).

PR #6126, opened 2017-10-26
Credentials: Namespace windows cred keys

ckamm:
The application name is prepended to the key. QtKeychain doesn't
do that automatically on the platform.

See #6125 

@SamuAlfageme I'd like to ask you to redo the 2.3 -> 2.4 upgrade tests after this patch hits. It's expected that 2.4alpha credentials won't be found or deleted.

### DIFF
--- a/src/libsync/creds/abstractcredentials.cpp
+++ b/src/libsync/creds/abstractcredentials.cpp
@@ -14,6 +14,7 @@
 
 #include <QLoggingCategory>
 #include <QString>
+#include <QCoreApplication>
 
 #include "common/asserts.h"
 #include "creds/abstractcredentials.h"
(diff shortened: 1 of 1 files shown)


guruz:
It looks good. We could merge it but let's think about this:

Should we have QtKeychain as git submodule https://github.com/owncloud/client/issues/1416 and fix it directly in the QtKeychain fork and then attempt to upstream this to QtKeychain?

Hm.. @ikushn-s mentioned though that this sucks a bit for QtKeychain users updating to a newer QtKeychain with suddenly different keynames..
CC @frankosterfeld
Opinions?

ckamm:
Fixing it upstream would be just as well. I agree that there'd need to be a flag to toggle between old and new behavior to make migration possible. We'd still need the migration code - we'd just save on composing the key ourselves. The main advantage would be alerting other users of the library and, possibly, changing the default behavior.

@guruz I'd merge this as-is. If there's an update to QtKeychain we can update to using that with minor changes.

Issue #6117, opened 2017-10-25
Systray status icon missing when client starts before systray


RalfJung:
### Expected behaviour
The systray status icon sometimes disappears even though ownCloud clients keeps running.

### Actual behaviour
The status icon should not disappear.

### Steps to reproduce
1. Start owncloud
2. Suspend and resume the system a couple of times and plug an external screen in and out

### Client configuration
Client version: 2.3.3

Operating system: Linux, Debian, Gnome, with Topicon plus extension

OS language: English

Qt version used by client package (Linux only, see also Settings dialog):
The settings dialog dosn't say anything, but I have Qt 5.9.1 installed.

Client package (From ownCloud or distro) (Linux only): distro (Debian)

Installation path of client: /usr (?)

### Logs

The problem is not reproducible, so I have not managed to collect logs yet.

ckamm:
The client never explicitly hides itself so this is likely a bug in Qt or the tray icon implementation. At best we could attempt to work around it by doing `hide(); show();` on the tray icon. (but not too often, that's caused performance problems with some tray implementations in the past)

RalfJung:
Closer observation showed that actually this only seems to happen when ownCloud autostarts on log-in. Probably owncloud starts so early that Gnome shell has not yet set up the systray, and hence it does not work.

I will edit the .desktop file to include a delay; maybe including that upstream would make sense? It's a crude work-around though. Ideally, once the notification icon dbus service appears, ownCloud would register there.

ckamm:
@RalfJung Thanks for the clarification! I think something like this has been reported before. I'd be okay with adding a workaround. We don't want the tray to pop up late when users start the client explicitly though.

Due to old upstream bugs we currently don't query `isSystemTrayAvailable` if `QT_QPA_PLATFORMTHEME` is `appmenu-qt5`. I'll check whether that's still relevant today. 

Can you tell me the value of your `QT_QPA_PLATFORMTHEME` environment variable?

RalfJung:
> Can you tell me the value of your QT_QPA_PLATFORMTHEME environment variable?

It is not set.

> I'd be okay with adding a workaround. We don't want the tray to pop up late when users start the client explicitly though.

What I did for now is, in the desktop file, add `X-GNOME-Autostart-Delay=10`. That file is generated by ownCloud when I set the "autostart" option in the settings (I assume), so I guess that work-around could be applied automatically? However, KDE may have similar problems, and I don't know if it has similar autostart delay.

[...]

PR #6127, opened 2017-10-26
Autostart: Add 10s delay for gnome

ckamm:
Sometimes the client managed to start before the system tray. Then
there'd be no owncloud client tray icon.

See #6117. Thanks to @RalfJung for proposing this patch.


ckamm:
Doesn't suffice / work, see comment in ticket.

Issue #6050, opened 2017-09-24
client crashes when storage unaccessable or read only mode


sentabi:
### Expected behaviour
The  apps tell the folder is unaccessable, so it paused the sync.

### Actual behaviour
Aborted (core dumped), without explanation

### Steps to reproduce
I test this on dual boot system Windows 10 / Fedora 26. 
1. run owncloud client on read only system
2. owncloud client show nothing about what happen
3. using `strace` to debug 

### Client configuration
Client version:
ownCloud version 2.3.3 (build 8338)

Operating system:
Fedora 26 64 Bit

OS language:
English

### Logs
strace output
```
....
fcntl(18, F_SETLK, {l_type=F_RDLCK, l_whence=SEEK_SET, l_start=1073741824, l_len=1}) = 0
fcntl(18, F_SETLK, {l_type=F_RDLCK, l_whence=SEEK_SET, l_start=1073741826, l_len=510}) = 0
fcntl(18, F_SETLK, {l_type=F_UNLCK, l_whence=SEEK_SET, l_start=1073741824, l_len=1}) = 0
stat("/media/DATA/ownCloud/._sync_960e5fa11295.db-journal", 0x7ffdbb623360) = -1 ENOENT (No such file or directory)
stat("/media/DATA/ownCloud/._sync_960e5fa11295.db-wal", 0x7ffdbb623360) = -1 ENOENT (No such file or directory)
fstat(18, {st_mode=S_IFREG|0755, st_size=10104832, ...}) = 0
pread64(18, "SQLite format 3\0\20\0\2\2\0@  \0\0\0\31\0\0\t\243"..., 4096, 0) = 4096
getpid()                                = 2755
stat("/media/DATA/ownCloud/._sync_960e5fa11295.db", {st_mode=S_IFREG|0755, st_size=10104832, ...}) = 0
open("/media/DATA/ownCloud/._sync_960e5fa11295.db-wal", O_RDWR|O_CREAT|O_CLOEXEC, 0755) = -1 EROFS (Read-only file system)
open("/media/DATA/ownCloud/._sync_960e5fa11295.db-wal", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
fcntl(18, F_SETLK, {l_type=F_UNLCK, l_whence=SEEK_SET, l_start=0, l_len=0}) = 0
rt_sigprocmask(SIG_UNBLOCK, [ABRT], NULL, 8) = 0
[...]

PR #6129, opened 2017-10-27
Journal: Don't crash if the db file is readonly

ckamm:
Surprisingly sqlite3_open_v2() returns ok on readonly files with
SQLITE_OPEN_READWRITE.

See #6050


ckamm:
Huh, jenkins reports errors like `[ warning sync.database.sql ]:	"/tmp/AllFilesDeletedTest-ouCWhY/._sync_test.db" is not writable` in tests. Maybe `isWritable` isn't reliable?

ckamm:
Updated to be more sophisticated -- let's watch jenkins.

PR #6133, opened 2017-10-30
[Proxy Settings] Validate for empty hostname when manual proxy is selected. (good first issue)

wiggiBe:


guruz:
Did you create the wrong PR?

Issue #6135, opened 2017-10-31
[OAuth2] Replace GET http verb for PROPFIND on DetermineAuthTypeJob


SamuAlfageme:
Note that this issue is highly dependent on the server configuration (i.e. firewalls, extra security layers...) and does not affect most of OAuth2 server setups.

On account creation wizard, to determine the authentication type, the client performs a dummy (`GET`) request to the webdav endpoint where it checks the `WWW-Authenticate` headers on the response. However It seems that some firewalls are picky on which verbs can reach that endpoint and only answers what we expect if it gets hit by a `PROPFIND`. - this seems to be the original problem behind the behavior I saw on https://github.com/owncloud/client/issues/5954#issuecomment-337634940

### Expected behavior

As an illustrative example, here's the cURL version of [the request the iOS client does](https://github.com/owncloud/ios-library/blob/0c8bf94fe1743573edb50e9dba36b39c2a5f355f/OCCommunicationLib/OCCommunicationLib/OCWebDavClient/OCWebDAVClient.m#L219) right after checking `status.php` (as said, it's a "dummy" req. since we expect the server to reply w/ a 401 and the body will be ignored; the important part would be the HTTP method)

```
curl \
	-H 'Host:<host>' \
	-H 'Content-Type:application/xml' \
	-H 'Depth:0' \
	-H 'Connection:keep-alive' \
	-H 'Accept:*/*' \
	-H 'User-Agent:Mozilla/5.0 (iOS) ownCloud-iOS/3.7.0' \
	-H 'Content-Length:383' \
	-H 'Accept-Language:en-us' \
	-H 'Accept-Encoding:gzip, deflate' \
	-X PROPFIND 'https://<host>/remote.php/webdav/' \
	--data-binary '<?xml version="1.0" encoding="UTF-8"?>
        <d:propfind xmlns:d="DAV:">
        <d:prop>
            <d:resourcetype />
            <d:getlastmodified />
            <size xmlns="http://owncloud.org/ns" />
            <d:creationdate />
            <id xmlns="http://owncloud.org/ns" />
            <d:getcontentlength />
            <d:displayname />
            <d:quota-available-bytes />
            <d:getetag />
            <permissions xmlns="http://owncloud.org/ns" />
            <d:quota-used-bytes />
            <d:getcontenttype />
        </d:prop>
        </d:propfind>'
```

### Actual behavior

```
curl \
[...]

PR #6136, opened 2017-11-01
DetermineAuthType: Adjustments for tight firewalls

ckamm:
With some firewalls we can't GET /remote.php/webdav/. Here we keep the
GET request to detect shibboleth through the redirect pattern but then
use PROPFIND to figure out the http auth method.

This also restores the fallback behavior of assuming basic auth
when no auth type can be determined.

See  #6135

@SamuAlfageme Could you test with the specific problematic server?

### DIFF
--- a/src/gui/creds/httpcredentialsgui.cpp
+++ b/src/gui/creds/httpcredentialsgui.cpp
@@ -30,6 +30,8 @@ using namespace QKeychain;
 
 namespace OCC {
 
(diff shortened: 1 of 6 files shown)


SamuAlfageme:
@ckamm seems to be working; yup. 👍 

I'll try to get a second account on a different enterprise-firewall setup to double check.

ckamm:
@SamuAlfageme Let me know how the test went and I'll merge it.

ckamm:
Merged after discussion with @SamuAlfageme . He'll do the additional tests when it becomes possible and we can fix any remaining problems for the rc.

Issue #5885, opened 2017-07-10
[Proxy Settings] Validate for empty hostname when manual proxy is selected


SamuAlfageme:
### Expected behavior
Settings should not be able to set themselves to a non-valid config (i.e. no hostname when manual proxy selected). Error focus could be applied to the field to bring attention to it.

### Actual behavior
No error is displayed in Network tab, and "Host not found" error propagating to the account(s) tab(s):

![screenshot 2017-07-10 14 07 08](https://user-images.githubusercontent.com/2644445/28017388-4f77ca2e-6579-11e7-91c6-2dfbb2d4b663.png)

### Steps to reproduce
1. Select "Specify proxy manually as" 
2. Do not input a hostname

SamuAlfageme:
@ckamm even when not the empty field has error-like focus:

![screen shot 2018-04-12 at 10 50 13](https://user-images.githubusercontent.com/2644445/38666507-600c9be0-3e3f-11e8-9f44-dba29ef6c123.png)

It's still possible to leave it empty - causing the "host not found" error as above.

ckamm:
@SamuAlfageme I don't think so. If we want modal behavior for the settings they shouldn't be a non-modal page in the account settings in the first place. (but we won't make that change for 2.5)

Maybe I can improve the sync error message for the proxy with an empty hostname.

SamuAlfageme:
@ckamm agreed - let's review the settings UI/arrangement on a future version, nothing to do for now. 

> Maybe I can improve the sync error message for the proxy with an empty hostname.

WFM - also make the error linked to the "Network" tab like we do with conflict warnings and "Activity > Not Synced" maybe?

ckamm:
@SamuAlfageme This is a minor issue and I don't want to add complex ui for it. Currently what happens is that an empty proxy host gets treated as "no proxy" which seems ok to me: I guess that users would only ever leave this field empty if they wanted to unset the proxy, deleted the hostname but forgot to switch the combo box. 

I'd just leave this as-is. The most I'd do is set the ui to "no proxy" if the user leaves the network settings page without having filled in the hostname.

SamuAlfageme:
@ckamm You're right - Sorry if I went down the over-engineering rabbit hole 😅 

> set the ui to "no proxy" if the user leaves the network settings page without having filled in the hostname.

Perfect!

ckamm:
Let's track this minor change in the PR

Issue #3233, opened 2015-05-13
Proxy change should trigger sync


guruz:
E.g. by resetting the Etag check timestamps and scheduling all ETag checks.. ?

dragotin:
Can you elaborate? I dont understand this enhancement. Thx.

MTRichards:
More information needed on this one please.

guruz:
If you change the proxy settings in the settings dialog, the client should trigger a sync and not wait 30 sec or more before something happens.

PR #6140, opened 2017-11-05
Hostname validation and sync start if proxy settings changes

wiggiBe:
(good first issue)
1. Issue: 
[Proxy Settings] Validate for empty hostname when manual proxy is selected #5885 

2. Issue:
Proxy change in Settings dialog should trigger sync #3233

### DIFF
--- a/src/gui/folderman.h
+++ b/src/gui/folderman.h
@@ -178,9 +178,6 @@ public:
     /** Queues all folders for syncing. */
     void scheduleAllFolders();
 
-    /** Will start a sync after a bit of delay. */
-    void startScheduledSyncSoon();
-
     void setDirtyProxy(bool value = true);
     void setDirtyNetworkLimits();
 
@@ -279,6 +276,9 @@ private:
 
     /* unloads a folder object, does not delete it */
     void unloadFolder(Folder *);
+
+    /** Will start a sync after a bit of delay. */
+    void startScheduledSyncSoon();
 
     // finds all folder configuration files
     // and create the folders
--- a/src/gui/folderman.h
+++ b/src/gui/folderman.h
@@ -178,6 +178,9 @@ public:
     /** Queues all folders for syncing. */
     void scheduleAllFolders();
 
+    /** Will start a sync after a bit of delay. */
+    void startScheduledSyncSoon();
+
     void setDirtyProxy(bool value = true);
     void setDirtyNetworkLimits();
 
@@ -276,9 +279,6 @@ private:
 
     /* unloads a folder object, does not delete it */
     void unloadFolder(Folder *);
-
-    /** Will start a sync after a bit of delay. */
-    void startScheduledSyncSoon();
 
     // finds all folder configuration files
     // and create the folders
--- a/src/gui/networksettings.cpp
+++ b/src/gui/networksettings.cpp
@@ -161,7 +161,7 @@ void NetworkSettings::saveProxySettings()
 
         if(hostLineEditValue.isEmpty())
         {
-            _ui->hostLineEdit->setStyleSheet("border: 1px solid red"); // Not sure if it is the best method
+            _ui->hostLineEdit->setStyleSheet("border: 1px solid red");
             return;
         }
         else
--- a/src/gui/networksettings.cpp
+++ b/src/gui/networksettings.cpp
@@ -20,9 +20,11 @@
 #include "application.h"
 #include "configfile.h"
 #include "folderman.h"
+#include "accountmanager.h"
 
 #include <QNetworkProxy>
 #include <QString>
+#include <QList>
 
 namespace OCC {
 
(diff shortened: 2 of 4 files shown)


ckamm:
@wiggiBe This looks good!

1) Did you sign the contribution agreement already? Can I help with this?

2) I'm not sure `startScheduledSyncSoon` is the right choice here. Wouldn't you want to schedule all folders? But that happens automatically when an account becomes connected. So maybe the actual thing you want to trigger is an account-connectivity-check? (so you don't have to wait for 30s)

`AccountState::checkConnectivity()` would do that.

wiggiBe:
Sorry, I didn't. I will sign it right away.

Oh, I see. I thought it would trigger a almost instant sync in all folders which are scheduled.
Do AccountState::checkConnectivity() really trigger a sync for a signed account?

ckamm:
@wiggiBe In the use case of "The account is disconnected and the user enters the proxy information" it would do it - the account will connect and all folders will be scheduled. 

You also want to schedule all folders in cases where the account was already connected? (say, switching between two working proxys)

In that case you'd need to set the AccountState to Disconnected first. Then checkConnectivity would attempt using the new proxy and, if successful, set the state to Connected again and schedule all folders.

wiggiBe:
That wasn't my intention but it makes sense to disconnect first and connect the account again.
I will do that.

src/gui/folderman.h
```diff
@@ -277,9 +280,6 @@ private slots:
     /* unloads a folder object, does not delete it */
     void unloadFolder(Folder *);
 
-    /** Will start a sync after a bit of delay. */
-    void startScheduledSyncSoon();
```
guruz: Why did you move this? :-)
Use ```git show``` before pushing to see the commit if it makes sense.
You can use ```git commit --amend``` to change your existing commit.

ckamm: It was private and needed to be public. Should be undone when the Account reconnection is used instead.

guruz: @ckamm Meh, github didn't display the context properly. I thought he changed it from "private **slots**" to not being a slot.

wiggiBe: @ckamm Indeed I will undo that.

@guruz Thank you for you help. :)
Unfortunately git commit --amend dont work because the last commit I deleted some comments but I allready changed the postion.

wiggiBe:
@ckamm Do I have to do a connectivity check for every saved account? (So I have to get a list of all accounts from accountmanager)

ckamm:
@wiggiBe Yes, I'd iterate through the accounts, disconnect them (there's no function for that yet, you need to add one) and trigger the connectivity check.

src/gui/networksettings.cpp
```diff
@@ -155,12 +156,25 @@ void NetworkSettings::saveProxySettings()
...
+        QString hostLineEditValue = _ui->hostLineEdit->text();
+
+        if(hostLineEditValue.isEmpty())
+        {
+            _ui->hostLineEdit->setStyleSheet("border: 1px solid red");
+            return;
+        }
+        else
+        {
+            _ui->hostLineEdit->setStyleSheet("border: 1px solid black"); // or blue
```
ogoffart: did you mean  "border: native" ?
or just `_ui->hostLineEdit->setStyleSheet(QString())`

wiggiBe: Yes, thank you. "border: native" is very good.

ogoffart:
The change does not build:

Jenkins error:  

```
/var/lib/jenkins/workspace/cloud-client_client_PR-6140-VJ2XIDBENLXHDCMPYD3F44RPONFPGAIGK6ABLT3J66BPSU4OUQQQ/src/gui/networksettings.cpp: In member function ‘void OCC::NetworkSettings::saveProxySettings()’:
/var/lib/jenkins/workspace/cloud-client_client_PR-6140-VJ2XIDBENLXHDCMPYD3F44RPONFPGAIGK6ABLT3J66BPSU4OUQQQ/src/gui/networksettings.cpp:192:33: error: base operand of ‘->’ has non-pointer type ‘QList<QExplicitlySharedDataPointer<OCC::AccountState> >’
     for(int i=0; i < accountList->count(); i++)
                                 ^
/var/lib/jenkins/workspace/cloud-client_client_PR-6140-VJ2XIDBENLXHDCMPYD3F44RPONFPGAIGK6ABLT3J66BPSU4OUQQQ/src/gui/networksettings.cpp:196:13: error: ‘acouuntList’ was not declared in this scope
             acouuntList[i]->disconnectByProxySettings();
             ^

```

src/gui/networksettings.cpp
```diff
@@ -185,8 +187,20 @@ void NetworkSettings::saveProxySettings()
     // start the sync.
     FolderMan::instance()->setDirtyProxy(true);
 
-    // start sync after a change
-    FolderMan::instance()->startScheduledSyncSoon();
+    QList<AccountStatePtr> accountList = AccountManager::instance()->accounts();
+
+    for(int i=0; i < accountList->count(); i++)
```
ckamm: `for (auto account : accountList) {`

ogoffart:
@wiggiBe  are you still working on this?

ckamm:
I've reworked and merged it to master!

Issue #6121, opened 2017-10-25
Add context menu action to "open file in browser" to sync protocol and server activity


ckamm:
Follow up from #5023.

What hasn't yet been implemented is to use private links to navigate from the sync protocol or server activity pane of the client to the web user interface.

Users should be able to right-click file entries in these views to get access to a "Open in browser" action.

There were also other tickets that suggested adding a context menu to the sync protocol view: #5982

SamuAlfageme:
@ckamm we need to update https://github.com/owncloud/owncloud-client-binary right?

ckamm:
Not yet, we'll need to do that when we're done with the shell integration additions of the same function for #5903

guruz:
Works for me in protocol.
I guess it can't work in server activity as this is a text-only activity log with no real references to files..

SamuAlfageme:
It's about time to close this one as well 👍 this works great now.

jnweiger:
@guruz Improvement idea:
In the sync protocol view, double click on a File  brings up the file browser. A right click opens a menu with one entry. 'Open in browser'.
But for some files no clicking whatsoever works. It is not apparent when and why.

I suggest to make both 'open in local file browser' and 'open in web browser' part of the right click context menu. Would be much clearer which is which.
Files that were deleted currently do not react to neither double click nor right click. The context menu should also open for them, with grayed out entries, so that the user knows the right click itself is reliable.
Files in the 'not synced' tab could have the same context menu for consistency, but with either the local or remote entry grayed out as appropriate.

guruz:
@jnweiger I agree with you about the graying out etc. Maybe create a new issue?
Which operation though would the double click be if there is no operation? Just no operation like currently?

ckamm:
@jnweiger That makes a lot of sense, I'll move it to an issue!

Issue #6146, opened 2017-11-08
Fix "legacy" private link generation feature for old servers, rely on DAV API for new servers


hodyroff:
2.4.0 Alpha 1 with openSUSE Leap 42.2 against ownCloud Server 10.0.3

### Expected behaviour
Right click, I can generate a private link which works or if the server does not support it - I can't and its greyed out.

### Actual behaviour
Link gets generates, but its slightly off
Generated link: https://my.mydomain.com/index.php/f/03922059515
Real private link: https://my.mydomain.com/index.php/f/3922059

### Steps to reproduce
1. Generate private link
2. open in web browser
3. access forbidden, ask owner for access

ckamm:
@hodyroff What's going on here is that 10.0.3 does not contain the 'privateLinks' capability so the link url can't be retrieved by webdav. In this case the client falls back on generating the url itself -- which is more brittle but is supposed to work.

My best guess is that this server's instance id starts with "515"? Could you check what it is?

I thought that was forbidden by reading core's `getInstanceId` which only generates ids starting with "oc". If that assumption is wrong we'll need to revisit the fallback private link generation!

michaelstingl:
Please don't use a fallback.

Sorry wasn't clear in https://github.com/owncloud/client/issues/6020 .

We have this funny issue that @hodyroff described, and additionally oC admins expect parameters in `config.php` like `overwritehost` are applied to the link. (with the 3.6.2 iOS app, I even created private links connected to a 8.2.x server)

ckamm:
@michaelstingl So how does the iOS app do it? They must be using some kind of fallback generation then?

But yes, we can drop the fallback and make this feature available only for servers that announce their support in the capabilities. Unfortunately it requires minor changes in all shell integrations and is thus not trivial.

michaelstingl:
@ckamm iOS app and Android app also implemented their own link generation, even on a server that don't have this feature. They also ignore `overwritehost` etc. 

I somebody wants you to implement something that feels like it better should have a proper server-side implementation, instead of implementing it based on assumptions 3x on all client platforms, please say NO the next time! (maybe it's me that wants silly stuff)

guruz:
The link @hodyroff mentioned above works for me (adjusting the hostname).
If this is so much as effort as @ckamm mentioned then we maybe just should keep the fallback in.
I would guess ```overwritehost``` should be applied as we use the Account's URL?

guruz:
Sorry, I have picked the wrong link :-/
[...]

PR #6149, opened 2017-11-08
Protocol: Introduce context menu with "open in browser" action

ckamm:
To do this conveniently a bunch of functionality that's common to
IssueWidget and ProtocolWidget is moved to ProtocolItem.

Also the convenience function to asynchronously retrieve the private
link url is moved from the socket api to the network jobs.

See #6121

### DIFF
--- a/src/gui/protocolwidget.cpp
+++ b/src/gui/protocolwidget.cpp
@@ -25,6 +25,8 @@
 #include "folder.h"
 #include "openfilemanager.h"
 #include "activityitemdelegate.h"
+#include "guiutility.h"
+#include "accountstate.h"
 
 #include "ui_protocolwidget.h"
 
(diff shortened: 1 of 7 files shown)


src/gui/protocolwidget.cpp
```diff
@@ -25,14 +25,115 @@
...
+    twitem->setIcon(0, icon);
+    twitem->setToolTip(0, longTimeStr);
+    twitem->setToolTip(1, item._file);
+    twitem->setData(2, Qt::UserRole, folder);
+    twitem->setToolTip(3, message);
+    twitem->setData(3, Qt::UserRole, item._status);
+    return twitem;
+}
+
+SyncJournalFileRecord ProtocolItem::syncJournalRecord(QTreeWidgetItem *item)
```
guruz: Do you really want to polute the namespace for a function only used once? It's basically a oneliner since the callee also tries to get the folder.

src/libsync/networkjobs.cpp
```diff
@@ -911,4 +911,36 @@ bool SimpleNetworkJob::finished()
...
 
+void fetchPrivateLinkUrl(AccountPtr account, const QString &remotePath,
+    const QByteArray &numericFileId, QObject *target,
+    std::function<void(const QString &url)> targetFun)
+{
+    const QString oldUrl = account->deprecatedPrivateLinkUrl(numericFileId).toString(QUrl::FullyEncoded);
+
+    // If the server doesn't have the property, use the old url directly.
+    if (!account->capabilities().privateLinkPropertyAvailable()) {
+        QTimer::singleShot(0, target, [=]() {
```
guruz: @ogoffart Maybe Qt should have an overload for https://doc.qt.io/qt-5/qmetaobject.html taking a lambda with QueuedConnection?

guruz:
Merge on hold until it's clear no changes are needed due to https://github.com/owncloud/client/issues/6146#issuecomment-343167920

ckamm:
This'll need some significant updates given the other private link fixes.

src/libsync/networkjobs.cpp
ogoffart: It has in Qt 5.10

src/gui/protocolwidget.cpp
```diff
@@ -25,14 +25,115 @@
...
+    QObject::connect(&openInBrowser, &QAction::triggered, parent, [&]() {
+        fetchPrivateLinkUrl(f->accountState()->account(), rec._path,
+            rec.numericFileId(), parent, [parent](const QString &url) {
+                Utility::openBrowser(url, parent);
+            });
+    });
+
+    QMenu menu;
+    if (rec.isValid())
+        menu.addAction(&openInBrowser);
```
ogoffart: There is a QMenu::addAction overload that takes a labmda in Qt 5.6,  maybe it could be used instead of building the action on the stack.

src/gui/protocolwidget.cpp
```diff
@@ -25,14 +25,115 @@
...
+
+    menu.exec(QCursor::pos());
```
ogoffart: Any reason to use QMenu::exec instead of the safer QMenu::popup?
If you capture by value in the lambda, that should work.

Aslo, QCursor::pos() might be wrong, one should use the position given in the slot argument, that's because the menu can be opened with the keyboard shortcut.

src/gui/protocolwidget.cpp
```diff
@@ -25,14 +25,115 @@
...
+    QAction openInBrowser(ProtocolWidget::tr("Open in browser"), noParent);
+    QObject::connect(&openInBrowser, &QAction::triggered, parent, [&]() {
+        fetchPrivateLinkUrl(f->accountState()->account(), rec._path,
+            rec.numericFileId(), parent, [parent](const QString &url) {
+                Utility::openBrowser(url, parent);
+            });
+    });
+
+    QMenu menu;
+    if (rec.isValid())
```
ogoffart: So if not, we have a menu with 0 elements?  Why not make an early return in this case?

ckamm: No, an empty context menu has no effect. But I'll add an early return to make that more explicit.

src/gui/protocolwidget.cpp
ckamm: Using the slot's argument actually placed the menu in a slightly offset location, so I looked at the code we use for popups in other places.

Being asynchronous would be fine (have to take care with `parent` though) and I'll go for that.

ckamm:
@guruz @ogoffart I rebased on top of master to adjust for the private link changes. I've also switched to using `popup` over `exec`.

src/gui/protocolwidget.cpp
ckamm: I think it'll be convenient in other places too and makes the code a bit more readable, in my opinon.

src/gui/protocolwidget.cpp
```diff
@@ -25,14 +25,124 @@
...
+
+    // More actions will be conditionally added to the context menu here later
+
+    if (menu->actions().isEmpty()) {
+        delete menu;
+        return;
+    }
+
+    QObject::connect(menu, &QMenu::aboutToHide, menu, &QMenu::deleteLater);
+    menu->popup(QCursor::pos());
```
ogoffart: you should pass the position from  IssuesWidget::slotItemContextMenu
(You can pass  mapToGlobal(pos) )
Because the menu might be opened via keyboard shortcut.

src/gui/protocolwidget.cpp
```diff
@@ -25,14 +25,124 @@
...
+    }
+
+    // More actions will be conditionally added to the context menu here later
+
+    if (menu->actions().isEmpty()) {
+        delete menu;
+        return;
+    }
+
+    QObject::connect(menu, &QMenu::aboutToHide, menu, &QMenu::deleteLater);
```
ogoffart: We do that elsewhere with     
    menu->setAttribute(Qt::WA_DeleteOnClose);

Is it not the same?

src/gui/protocolwidget.cpp
ckamm: We spoke about this previously: I did that initially and it didn't work correctly: The menu was offset awkwardly by a small distance. I can try again though. (we also use QCursor::pos in other places, presumably for this reason?)

guruz: Apparantly this is the way: https://github.com/owncloud/client/pull/6173/files#diff-29ad999698e087a443d0cd76499838cbR277

ckamm: What made it look odd in my tests was that the QPoint passed to the context menu slot uses viewport coordinates for QAbstractScrollArea. Thanks to @ogoffart for reading http://doc.qt.io/qt-5/qwidget.html#customContextMenuRequested and pointing me to that. It's adjusted in the new commit.

PR #6152, opened 2017-11-10
Private link fixes

ckamm:
For #6146

### DIFF
--- a/src/libsync/connectionvalidator.cpp
+++ b/src/libsync/connectionvalidator.cpp
@@ -18,6 +18,7 @@
 #include <QNetworkReply>
 #include <QNetworkProxyFactory>
 #include <QPixmap>
+#include <QXmlStreamReader>
 
 #include "connectionvalidator.h"
 #include "account.h"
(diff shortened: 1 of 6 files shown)


src/libsync/connectionvalidator.cpp
```diff
@@ -247,10 +248,22 @@ void ConnectionValidator::slotAuthSuccess()
...
 void ConnectionValidator::checkServerCapabilities()
 {
+    // The main flow now needs the capabilities
     JsonApiJob *job = new JsonApiJob(_account, QLatin1String("ocs/v1.php/cloud/capabilities"), this);
     job->setTimeout(timeoutToUseMsec);
     QObject::connect(job, &JsonApiJob::jsonReceived, this, &ConnectionValidator::slotCapabilitiesRecieved);
     job->start();
+
+    // And we'll retrieve the ocs config in parallel
+    auto ocsConfigUrl = Utility::concatUrlPath(_account->url(), QLatin1String("ocs/v1.php/config"));
```
ogoffart: I think you could use JsonApiJob for that

src/libsync/connectionvalidator.cpp
```diff
@@ -247,10 +248,22 @@ void ConnectionValidator::slotAuthSuccess()
...
+    auto configJob = _account->sendRequest("GET", ocsConfigUrl);
+    configJob->setTimeout(timeoutToUseMsec);
+    // note that 'this' might be destroyed before the job finishes
+    auto account = _account;
+    QObject::connect(configJob, &SimpleNetworkJob::finishedSignal, _account.data(),
+        [=](QNetworkReply *reply) {
+            ocsConfigReceived(reply, account);
```
ogoffart: also be carefull that this might actually be deleted here.

ckamm: I know, that's exactly why I'm capturing `account` by value and have `ocsConfigReceived` as a static method; will improve comments stating that.

Issue #6049, opened 2017-09-23
[Regression] Client crashes as soon as sync dir becomes unavailable or when sync dir is not available on startup


progmem64:
### Expected behaviour
If storage becomes unavailable, the client should notice and pause sync.

### Actual behaviour
The client crashes.

### Steps to reproduce
1. Sync to folder /x 
2. Unmount /x
3. Observe ownCloud crashing

### Client configuration
Client version: 2.3.3 (8242)
Operating system: macOS Sierra 10.12.6
OS language: English
Installation path of client: Default

guruz:
@michaelstingl @SamuAlfageme If enough people report this, it might mandate a 2.3.4.. let's see.

SamuAlfageme:
Reproduced & confirmed the regression. 

I think there's some kind of delay on the crash-reporter (last update I see was 18h ago)

valentijnscholten:
I am seeing similar behaviour on Windows 10 Pro x64 (more similar to #6050 actually). 
When the user runs the owncloud client and doesn't have write access to the owncloud folder, the client won't start. 

Actually it will crash before generating any logging. Or it will just hang. And then crash silently.

Also the crash reporter doesn't work in this case, it will crash as well. **So no crash reports**

Workaround is to run as administrator. Or to assign write permissions to the stack folder.

It makes sense the client won't work without write permissions, but it should inform the user about the missing permissions.

progmem64:
@valentijnscholten This is a little different from my problem. I don't have problem with permissions, but with non existing paths (e.g. external hard drive not mounted) causing the client to crash.

valentijnscholten:
@progmem64 yes, my issue is more similar to 6050 (folder being readonly), but that is closed and points to this issue for follow up.

ckamm:
@SamuAlfageme @guruz I can't reproduce on linux. You can't unmount the filesystem with the sync folder while the client is open because it keeps the journal file openend. If I put it on an USB stick and plug it out without unmounting, nothing bad happens. I get `Error: "unable to open database file" for "/media/kamm/E9B9-AE9E/test/._sync_dba920715f43.db"` and the sync fails with something like `Unable to find sync dir`

SamuAlfageme:
@ckamm hmmm.. will try to reproduce on leenux (report in https://github.com/owncloud/client/issues/6050#issue-260077761 was over readonly-FS in Fedora and I assumed the same reason was behind it).

jturcotte:
[...]

PR #5518, opened 2017-02-07
Improve usage of asserts

ckamm:

Issue #5429, opened 2017-01-05
Change several Q_ASSERT to if() with qFatal


guruz:
We have several places where it is much better to crash than to assert only in debug mode.

* [ ] Change https://github.com/owncloud/client/commit/143320341e1b01a53e69755cbd2badc61cf4440c to be only be for rcs and betas
* [ ] Check all asserts and see if it is better to change them to qFatal (e.g. https://github.com/owncloud/client/commit/adbe5ecf5507f5f751d869b02a2033fb895aaa14 )

FYI @ckamm

dschmidt:
Additional consideration to take into account:

Asserts might also hide improper error handling, developers might never see how the application behaves after an assertion didn't "fire" - i.e. there might be crashes in completely unexpected parts of the code - and that only in release builds, not even in beta or rc builds.
I'd recommend to at least make the behavior of beta, rc and release builds identical - so basically reverting https://github.com/owncloud/client/commit/143320341e1b01a53e69755cbd2badc61cf4440c.

And yes, I agree, if we can't recover from situations where assertions failed, we should use `qFatal()`

jturcotte:
We need to at least revert 1433203 in 2.3.

ckamm:
Pull request: https://github.com/owncloud/client/pull/5518

guruz:
(Closing in favour of PR)

PR #6161, opened 2017-11-14
Sql: db close() failing isn't fatal

ckamm:
The assert was made fatal when looking at asserts for #5429 #5518
without having a particular problem in mind. Recent reports weakly
suggest that this might lead to occasional crashes for users when
sqlite_close fails in ways that look ignorable.

### DIFF
--- a/src/common/ownsql.cpp
+++ b/src/common/ownsql.cpp
@@ -184,8 +184,8 @@ void SqlDatabase::close()
 {
     if (_db) {
         SQLITE_DO(sqlite3_close(_db));
-        // Fatal because reopening an unclosed db might be problematic.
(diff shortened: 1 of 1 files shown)


guruz:
@ckamm Do you think this fixes https://github.com/owncloud/client/issues/6049#issuecomment-343893501 ?

guruz:
...or maybe not, the ENFORCE is after the sqlite3 function

ckamm:
@guruz I had hoped so, but the sentry backtrace linked from https://github.com/owncloud/client/issues/6049#issuecomment-343893501 suggests that it actually crashes in `sqlite_close`

Issue #6156, opened 2017-11-13
Wizard font size issues


PVince81:
![screenshot_20171113_110519](https://user-images.githubusercontent.com/277525/32720006-8a02c1c0-c862-11e7-90df-83b5a1596129.png)

### Setup
owncloud-client-2.4.0~alpha1-8649.1.x86_64
plasma5-workspace-5.11.2-1.2.x86_64
openSUSE Tumbleweed

I'm on a HighDPI screen and I might have tweaked some fonts in Plasma 5:

![screenshot_20171113_110625](https://user-images.githubusercontent.com/277525/32720058-b18f1e64-c862-11e7-9830-10c566c545f4.png)

jnweiger:
This is not only the header: https://github.com/owncloud/client/issues/6158

ckamm:
Adding @jnweiger's image from #6158 here:
![](https://user-images.githubusercontent.com/1108546/32732961-5d6c2378-c88e-11e7-9745-cf26d1db0211.png)

ogoffart:
Can be reproduced on linux on a normal screen by setting an env variable, for example:
`QT_FONT_DPI=180`

ogoffart:
I think the problem is that QWizard has some sizes hardcoded, and it cannot be resized dynamically

ogoffart:
It turns out that we make the wizard a fixed size by setting a banner to it.
The size of the banner is hardcoded in Theme::wizardHeaderBanner.  Maybe we should scale it by the DPI

ogoffart:
Suggested fix: https://github.com/owncloud/client/pull/6164

jnweiger:
I believe the fix from @ogoffart should help in Vincent's case, -- but probably not in my case.

ogoffart:
@jnweiger  yes, it fixes both,  the width of the wizard is fixed and the same as the width of the header banner

SamuAlfageme:
@PVince81 care to give latest nightly: https://software.opensuse.org//download.html?project=isv:ownCloud:community:nightly&package=owncloud-client a try?

SamuAlfageme:
@ogoffart I think the "Add Folder Sync Connection" button also has a prefixed width; the label is not being wrapped as for 2.4.0nightly-20171121:

<p align="center">
  <img src="https://user-images.githubusercontent.com/2644445/33167597-4f66512e-d03e-11e7-9adf-5dafe3362a61.png"/>
</p>

ogoffart:
[...]

PR #6164, opened 2017-11-14
Theme: Update the wizard size on high dpi screen

ogoffart:
Issue #6156

### DIFF
--- a/src/libsync/theme.cpp
+++ b/src/libsync/theme.cpp
@@ -428,7 +428,15 @@ QPixmap Theme::wizardHeaderBanner() const
     if (!c.isValid())
         return QPixmap();
 
-    QPixmap pix(QSize(750, 78));
+    QSize size(750, 78);
+    if (auto screen = qApp->primaryScreen()) {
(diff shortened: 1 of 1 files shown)


src/libsync/theme.cpp
```diff
@@ -428,7 +428,15 @@ QPixmap Theme::wizardHeaderBanner() const
     if (!c.isValid())
         return QPixmap();
 
-    QPixmap pix(QSize(750, 78));
+    QSize size(750, 78);
+    if (auto screen = qApp->primaryScreen()) {
+        // Adjust the the size if there is a different DPI. (Issue #6156)
+        // Indeed, this size need to be big enough to fir the banner height, and the wizard's width
```
ckamm: for

ckamm:
@ogoffart This looks fine for when the default `wizardHeaderBanner` is used but won't branded clients run into the same problem?

ogoffart:
Yes, branded client that override with a pixmap will have the same problem.

Issue #6163, opened 2017-11-14
Identify the button and ask confirmation on public link deletion


hodyroff:
Linux - the trashbin was not obvious to me, when I hover over it nothing tells me that is is a delete. When I press on it, link is gone without warning.

Should we not have a warning and Ok before Link is deleted?

SamuAlfageme:
Hm, I'd say tooltip on hoover is a good idea. A confused user will wait for it to click on the button  while the UI can be kept simple and compact. 

Confirmation dialog seems rather a bit too much - true that the webUI has it though. 

cc/ @pmaier1 @ckamm opinions?

ckamm:
I have added a tooltip to the delete button. Another approach here would be to remove the icon based delete button outright and replace it with an explicit "Delete" action inside the "..." menu - that would be very explicit.

SamuAlfageme:
@hodyroff @pmaier1 screenshot for ref:

<p align="center">
  <img src="https://user-images.githubusercontent.com/2644445/32829339-165a3960-c9f2-11e7-9d1e-711189730f07.png"/>
</p>

hodyroff:
Delete here is a non-reversable action therefor we should have a confirmation dialog, like we do in the WebUI = consistent. The delete should also show up in the ... menu as an option. Space is enough, maybe have a link to clipboard icon as well? and link to mail which are the 3 obvious functions and then have all in the "..." menu again. What do you think? @felixheidecke

michaelstingl:
Hover in general is a problem because more and more people use touch input on the desktop nowadays. Maybe there is another way…

guruz:
@hodyroff It's too late for so many new things, we want to release the beta1 ;-) But you can create an issue for 2.5 to have both UIs more aligned.

Having it in the "..." submenu **instead of the icon button** would be more explicit since one would see the text and not just a garbage icon.

SamuAlfageme:
@hodyroff @pmaier1 @michaelstingl reference behavior after #6166 

<p align="center">
  <img src="https://user-images.githubusercontent.com/2644445/32884771-ba967c36-cabb-11e7-8919-d621f51ca58d.gif"/>
</p>

hodyroff:
Looks great, but delete should be the last point, not the first ... or is it stricly alphabetically sorted?

ckamm:
@hodyroff I've moved it to the bottom of the menu.

hodyroff:
Perfect, can be closed for me ...

PR #6166, opened 2017-11-15
Share links: Confirm deletion explicitly

ckamm:
Also add the "Delete" action to the "..." menu.

See  #6163


ckamm:
@guruz What about this update?

guruz:
@SamuAlfageme Sorry, can follow this up for after the beta..
See also https://github.com/owncloud/client/issues/6163#issuecomment-344726202

Issue #6153, opened 2017-11-10
[Enhancement] Check file hash before starting network transfer


stefanochiappini:
Following [this discussion](https://central.owncloud.org/t/first-sync-of-big-folders-when-most-of-the-files-are-the-same/10302) with @dragotin, it would be very useful to implement file hash check in the client, in order to avoid unnecessary data transfer, saving time and bandwitdh when syncing big folder with most of the files  #equal.

### Expected behaviour
1. Whenever the client is started or there's a change in the synchronized folder, for every local file, the client should check if a file with the same name and size exists on the server;
2. if it already exists, calculate the hash of the file locally and compare that to the one on the server;
3. if both are equal, the client knows that both files are the same, and it can fill the entry in the local sync journal from the values from the server, while not downloading the file.

### Actual behaviour
Equal files are unnecessarly downloaded, wasting a lot of time and bandwidth.

### Steps to reproduce
1. Install client on PC-1 a fill the synchronized "owncloud" folder with files.
2. Data on PC-1 are copied in the server.
3. Install client on PC-2 a fill the synchronized "owncloud" folder with the same files in PC-1.
4. PC-2 downloads data from the server even if the files are exactly the same

### Server configuration
Operating system: CentOS 7.2
Web server: Apache 2.4.6
Database: MariaDB 5.5.56
PHP version:  7.1
ownCloud version:  10.0.3

### Client configuration
Client version:  2.3.3 (build 8250)
Operating system: Windows 7
OS language:  Italian

### Logs
no error log

ckamm:
This feature exists in 2.4, see #5838. If the size, mtime and content hash match, no data will be downloaded.

guruz:
I'm re-opening this. When having a good checksum (md5 or sha1) then there is no need for mtime check. The tuple (size,checksum/hash) is good enough. Please remove this mtime check if possible :-)

@ckamm Not sure if this is too late for 2.4.0. Maybe move to 2.4.1 or 2.5.0

ckamm:
If we don't discover significant issues during review I think this PR will be fine for 2.4

stefanochiappini:
I agree with **guruz**, the mtime check might be the reason why it didn't work for me during the first sync (a big folder was unnecessarily downloaded), maybe moving file across different OS (Win, Linux, OSx) mtime can be touched somewhere...

SamuAlfageme:
[...]

PR #6168, opened 2017-11-16
PropagateDownload: Skip identical files more

ckamm:
Previously we required matching mtimes but that's actually
unnecessary when the question is about whether to skip the
download. We will still update the downloaded file's metadata.

See #6153

### DIFF
--- a/src/libsync/propagatedownload.cpp
+++ b/src/libsync/propagatedownload.cpp
@@ -357,13 +357,16 @@ void PropagateDownloadFile::start()
         }
     }
 
-    // If we have a conflict where size and mtime are identical,
+    // If we have a conflict where size of the file is unchanged,
     // compare the remote checksum to the local one.
     // Maybe it's not a real conflict and no download is necessary!
+    // If the hashes are collision safe and identical, we assume the content is too.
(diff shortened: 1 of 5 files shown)


src/libsync/propagatedownload.cpp
```diff
@@ -357,12 +357,11 @@ void PropagateDownloadFile::start()
         }
     }
 
-    // If we have a conflict where size and mtime are identical,
+    // If we have a conflict where size of the file is unchanged,
     // compare the remote checksum to the local one.
     // Maybe it's not a real conflict and no download is necessary!
     if (_item->_instruction == CSYNC_INSTRUCTION_CONFLICT
         && _item->_size == _item->_previousSize
-        && _item->_modtime == _item->_previousModtime
```
ogoffart: I'd change that to

`_item->_modtime == _item->_previousModtime  || isChecksumSufficientlyGood`

ogoffart:
Use case to check the mtime here, is that if somebody deletes the database and do a sync again,  not everything should be conflictsn even if there is no hashes or that the hashes are week.
But if you remove the mtime check, i don't think it is good enough if the hashes are week

PR #6177, opened 2017-11-21
csync_update: ignore remote files that cannot be encoded

ogoffart:
This is mainly for linux, whose local is not UTF-8.
For example, in latin1, it is not possible to encode emoji or chinese character.
If there are such character in the filename, Qt would just save the file using
the replacement character ('?'). Then, on the next sync, client would rename
the files using this replacement character.

Avoid this by ignoring the files which cannot be downloaded because the
filename cannot be represented with the user's locale

Relates to issue #5676 and #5719


ogoffart:
I wonder if this should go in 2.4 or 2.5.
2.4.0 is doing some change in the encoding already, and i'm afraid this can be otherwise seen as a regression by some people.

Issue #4280, opened 2015-12-14
`add folder synchronization` dialog box too short


martin-rueegg:
(client: Version 2.1.0 (build 5683). Windows.)

when adding a new folder synchronization and there is a warning message displayed, the rest of the dialog gets shrunken and unreadable:
<img width="261" alt="new_folder_sync shrunken" src="https://cloud.githubusercontent.com/assets/6409516/11783213/d569f060-a26d-11e5-8021-a34f2fa76d1c.png">

when increasing hight, it's ok:
<img width="261" alt="new_folder_sync enlarged" src="https://cloud.githubusercontent.com/assets/6409516/11783230/fcfe2c7c-a26d-11e5-81c5-f0a97ee7e89c.png">

hight should be increased automatically by the hight of the notification message, or it should be increased in general.

martin-rueegg:
possibly related #3414

guruz:
@michaelstingl @martin-rueegg @jturcotte Is this fixed with the HiDPi improvements in 2.1.1?

danimo:
@guruz No idea, I don't have any HiDPI screen. Let's wait for @martin-rueegg to confirm. It should though.

martin-rueegg:
thanks for asking. unfortunately the OC Client 2.1.0 did not prompt for an update, even after restarting the client it showed to be the "latest version" ...

anyways, have updated to 2.1.1 (build 5837). however, the issue remains:
![oc_add_folder](https://cloud.githubusercontent.com/assets/6409516/13215324/59aaf792-d94c-11e5-90bf-9b7b701b1cd6.JPG)

thank you for your attention!

danimo:
@martin-rueegg We haven't yet enabled the auto-updated. I think you'll need to wait until 2.2.

martin-rueegg:
@danimo, no worries. i just wondered that the information on Allgemein/Updates says that i'm running the latest version (even though i wasn't).
if an information is given, i would expect it to be accurate, regardless of whether auto-update itself is performed. however, if no check is performed, i would expect no information is given in the first place ... ;-)
but that's an other issue. thanks for your comment anyway!

jturcotte:
Let's see if the move to Qt 5.6 helps.

guruz:
https://owncloud.org/changelog/desktop/ 2.3 is out which might fix this

martin-rueegg:
Yes partially it is fixed, apart from this dialog:

[...]

PR #6179, opened 2017-11-21
FolderWizard: Fix minimum size

ogoffart:
There are several bugs in QWizard that needs to be worked around for the minimum
size of the widget to take effect.

Issue #4280

### DIFF
--- a/src/gui/folderwizard.cpp
+++ b/src/gui/folderwizard.cpp
@@ -35,6 +35,7 @@
 #include <QWizardPage>
 #include <QTreeWidget>
 #include <QVBoxLayout>
+#include <QEvent>
 
 #include <stdlib.h>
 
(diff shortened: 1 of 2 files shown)


ogoffart:
@guruz   your call if this goes in 2.4 or 2.5

ogoffart:
Note: I only tested linux

src/gui/folderwizard.cpp
```diff
@@ -551,5 +554,26 @@ FolderWizard::~FolderWizard()
...
+    return QWizard::eventFilter(watched, event);
+}
+
+void FolderWizard::resizeEvent(QResizeEvent *event)
+{
+    QWizard::resizeEvent(event);
+
+    // workaround for QTBUG-22819: when the error label word wrap, the minimum height would be too small
+    int hfw = currentPage()->heightForWidth(currentPage()->width());
+    qDebug() << hfw << currentPage()->height();
```
ckamm: stray qDebug

ogoffart: Removed.

guruz:
@SamuAlfageme @ogoffart  Does this also relate to the discussion here? 
https://github.com/owncloud/client/issues/6156#issuecomment-346576328
I guess not since that screenshot is from the normal UI not the wizard?

ogoffart:
@guruz: no, that is unrelated.

PR #6184, opened 2017-11-23
libsync: Fix compilation with TOKEN_AUTH_ONLY

ogoffart:
This means we cannot use QtGui in libsync.
So this mostly disable the avatar from the account and the avatarjob

Note that there is one logic change: in ConnectionValidator::slotUserFetched
we do the avatar job even if the user is empty. Otherwise we would end up in
a invalid state. This restore the 2.3.x behavior that was broken in
commit e05d6bfcdc39b551d0a29c763afcabcadb5d05a6

### DIFF
--- a/src/libsync/account.cpp
+++ b/src/libsync/account.cpp
@@ -93,6 +93,7 @@ void Account::setDavUser(const QString &newDavUser)
     _davUser = newDavUser;
 }
 
+#ifndef TOKEN_AUTH_ONLY
 QImage Account::avatar() const
 {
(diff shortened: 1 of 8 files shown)


guruz:
Looks Ok but i'd like @ckamm to review this because of the reportResult change

ogoffart:
@guruz the reportResult is also what was before commit e05d6bf

ckamm:
Thanks for fixing! I didn't know TOKEN_AUTH required not using QtGui in libsync.

PR #6186, opened 2017-11-23
AccountSettings: Fix the size of the "Add Sync Folder Connection" button

ogoffart:
The problem here is that the QPainter is created on the viewport with is a
QWidget, but QAbstractItemView can have a different font, and the
QStyleOptionViewItem::font is this font. QStyleOptionViewItem::font
was used to compute the sizeHint, and the default font from the QPainter
was used to draw the text, so they could be not in sync.

Fix it by always using the font for QPushButton

Reported in
https://github.com/owncloud/client/issues/6156#issuecomment-346576328

### DIFF
--- a/src/gui/folderstatusdelegate.cpp
+++ b/src/gui/folderstatusdelegate.cpp
@@ -59,7 +59,7 @@ QSize FolderStatusDelegate::sizeHint(const QStyleOptionViewItem &option,
     auto classif = static_cast<const FolderStatusModel *>(index.model())->classify(index);
     if (classif == FolderStatusModel::AddButton) {
         const int margins = aliasFm.height(); // same as 2*aliasMargin of paint
-        QFontMetrics fm(option.font);
+        QFontMetrics fm(qApp->font("QPushButton"));
         QStyleOptionButton opt;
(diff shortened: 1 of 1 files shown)


ogoffart:
(I think this can go in 2.4 unless @guruz  has an objection)

src/gui/folderstatusdelegate.cpp
```diff
@@ -59,7 +59,7 @@ QSize FolderStatusDelegate::sizeHint(const QStyleOptionViewItem &option,
     auto classif = static_cast<const FolderStatusModel *>(index.model())->classify(index);
     if (classif == FolderStatusModel::AddButton) {
         const int margins = aliasFm.height(); // same as 2*aliasMargin of paint
-        QFontMetrics fm(option.font);
+        QFontMetrics fm(qApp->font("QPushButton"));
```
guruz: Am I blind? Where is that documented?
https://doc.qt.io/qt-5/qguiapplication.html#font

ogoffart: Documented there:  https://doc.qt.io/qt-5/qapplication.html#font-2

Issue #6183, opened 2017-11-22
Beta1: Manpage missing


dragotin:
The doc files for the manpages for owncloud and owncloudcmd were removed from the source tree, thus building man pages is broken.

Hint: Try `make doc-man` in the source build tree. Breaks because owncloud.1.rst and owncloudcmd.1.rst were removed, see https://github.com/owncloud/documentation/issues/3240 for details and the commit that removed it.

SamuAlfageme:
manpages were removed as per https://github.com/owncloud/client/issues/2153#issuecomment-286037172 in 2.3.2 - but maybe we should keep actively maintaining those if distros since it seems @dragotin is not the only one missing them (ref. https://github.com/owncloud/client/issues/6011#issuecomment-328476462) 

cc/ @jnweiger @hodyroff

jnweiger:
All releases have separate packaging templates. We never want changes to be retroactove.

Currently there is no config for 2.4.1 -- we support 2.4.0 from branch 2.4 and 2.5.0 from branch master.

guruz:
PR was merged.
@jnweiger need to make sure that branded packages don't ship with owncloud* man pages.

guruz:
Source file seems to have it again
```
drwxr-xr-x  0 jenkins jenkins       0 Dec  4 16:48 owncloudclient-2.4.0-rc1/man/
-rw-r--r--  0 jenkins jenkins     871 Dec  4 16:48 owncloudclient-2.4.0-rc1/man/owncloud.1.rst
-rw-r--r--  0 jenkins jenkins     520 Dec  4 16:48 owncloudclient-2.4.0-rc1/man/CMakeLists.txt
-rw-r--r--  0 jenkins jenkins    2664 Dec  4 16:48 owncloudclient-2.4.0-rc1/man/owncloudcmd.1.rst
-rw-r--r--  0 jenkins jenkins       1 Dec  4 16:48 owncloudclient-2.4.0-rc1/man/index.rst
```

guruz:
https://owncloud.org/changelog/desktop/

SamuAlfageme:
They continue non-linked (i.e. not present in the `manpath`, `/usr/share/man/man1`); so `man owncloud` will output "No manual entry for owncloud" 

@jnweiger where the post-install scripts adapted to do https://github.com/owncloud/client/issues/2153#issuecomment-269344867?

jnweiger:
[...]

PR #6193, opened 2017-11-27
move man rst files to separate dir

michaelstingl:
Proposal for https://github.com/owncloud/client/issues/6183

Idea:
- manpage content is no longer merged in [official docs](https://doc.owncloud.org/desktop/latest/)

To check:
- @dragotin Does this for you?
- @jnweiger Please disable manpages for branded client builds (as long as manpages aren‘t branded)


hefee:
I have moved the doc-man command to the new subdir. At least it compiles the manpages. Because the tests are broken, i can't test if it installs correctly. We really should fix this before 2.4.0 release.

guruz:
Changed base to 2.4

PR #6195, opened 2017-11-28
Using gwamp to connect to crossbar

DeepDiver1975:
# ToDo
- [x] query capabilities for the wamp router endpoint
- [ ] define proper session name - no idea what this should be
- [ ] define a common realm which is to be used (involves server & router setup)
- [ ] define topic names (involves server)
- [ ] test with multiple accounts being connected to the same router - I had some issues ...
- [ ] clarify authentication and topic access @DeepDiver1975

💟 I really love lambdas in C++ - nice to see the language evolving 💟

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,3 +7,10 @@
 [submodule "src/3rdparty/libcrashreporter-qt"]
 	path = src/3rdparty/libcrashreporter-qt
 	url = git://github.com/dschmidt/libcrashreporter-qt.git
+[submodule "src/3rdparty/qmsgpack"]
+	path = src/3rdparty/qmsgpack
+	url = https://github.com/romixlab/qmsgpack.git
+[submodule "src/3rdparty/qwamp"]
+	path = src/3rdparty/qwamp
+	url = https://github.com/DeepDiver1975/qwamp.git
+
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -218,6 +218,10 @@ add_definitions( -D_WIN32_WINNT=0x0600)
 add_definitions( -DWINVER=0x0600)
 endif( WIN32 )
 
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
+
 include(QtVersionAbstraction)
 setup_qt()
 if (${Qt5Core_VERSION_MAJOR} EQUAL "5")
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -28,6 +28,7 @@ endif()
 
 add_subdirectory(csync)
 add_subdirectory(libsync)
+
 if (NOT BUILD_LIBRARIES_ONLY)
     add_subdirectory(gui)
     add_subdirectory(cmd)
@@ -37,6 +38,8 @@ if (NOT BUILD_LIBRARIES_ONLY)
         add_subdirectory(crashreporter)
     endif()
 endif(NOT BUILD_LIBRARIES_ONLY)
+
+add_subdirectory(3rdparty/qmsgpack)
 
 find_program(KRAZY2_EXECUTABLE krazy2)
 if(KRAZY2_EXECUTABLE)
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -157,6 +157,13 @@ set(3rdparty_SRC
     ../3rdparty/qtsingleapplication/qtlocalpeer.cpp
     ../3rdparty/qtsingleapplication/qtsingleapplication.cpp
     ../3rdparty/qtsingleapplication/qtsinglecoreapplication.cpp
+    ../3rdparty/qmsgpack/src/msgpack.cpp
+    ../3rdparty/qmsgpack/src/private/pack_p.cpp
+    ../3rdparty/qmsgpack/src/private/unpack_p.cpp
+    ../3rdparty/qmsgpack/src/private/qt_types_p.cpp
+    ../3rdparty/qmsgpack/src/stream/time.cpp
+    ../3rdparty/qmsgpack/src/msgpackstream.cpp
+    ../3rdparty/qwamp/qwamp.cpp
    )
 
 if (APPLE)
@@ -178,6 +185,8 @@ set(3rdparty_INC
     ${CMAKE_SOURCE_DIR}/src/3rdparty/qtlockedfile
     ${CMAKE_SOURCE_DIR}/src/3rdparty/qtmacgoodies/src
     ${CMAKE_SOURCE_DIR}/src/3rdparty/qtsingleapplication
+    ${CMAKE_SOURCE_DIR}/src/3rdparty/qwamp
+    ${CMAKE_SOURCE_DIR}/src/3rdparty/qmsgpack/src
     )
 
 include_directories(${3rdparty_INC})
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -34,8 +34,6 @@
 
 #include "creds/abstractcredentials.h"
 
-#include "qwamp.h"
-
 #include <QTimer>
 #include <QUrl>
 #include <QDir>
@@ -324,40 +322,39 @@ void Folder::slotAccountCapabilitiesChanged()
         return;
     }
 
+    if (_webSocket) {
+        return;
+    }
     //initialize wamp connection
-    QWamp::Session *session;
     _webSocket.reset(new QTcpSocket());
     QUrl webSocketUrl(wsUrl);
     QObject::connect(_webSocket.data(), &QTcpSocket::connected, [&]() {
 
         QString sessionName("sessionName");
-        session = new QWamp::Session(sessionName, *_webSocket.data(), QWamp::Session::MessageFormat::Msgpack, true);
+        _session.reset(new QWamp::Session(sessionName, *_webSocket.data(), QWamp::Session::MessageFormat::Msgpack, true));
 
-        QObject::connect(session, &QWamp::Session::joined, [&](qint64 s) {
+        QObject::connect(_session.data(), &QWamp::Session::joined, [&](qint64 s) {
             qDebug() << "Session joined to realm1 with session ID " << s;
 
-            session->subscribe("etag-changed-channel", [&](const QVariantList& args, const QVariantMap& options) {
+            // TODO: add current userid
+            _session->subscribe("files.root-etag-changed.admin", [&](const QVariantList& args, const QVariantMap& options) {
                 qDebug() << "Event received";
-                //    bool success;
-                //    QVariantMap json = QtJson::parse(message, success).toMap();
-                //    if (success) {
-                //        QString notificationEtag = json["etag"].toString();
-                //        QString notificationUser = json["user"].toString();
-                //        QString user = _accountState->account()->credentials()->user();
-                //        if (user == notificationUser && _lastEtag != notificationEtag) {
-                //            qDebug() << "* [WebSocket] Compare etag with previous etag: last:" << _lastEtag << ", received:" << notificationEtag << "-> CHANGED";
-                //            _lastEtag = notificationEtag;
-                //            slotScheduleThisFolder();
-                //            _accountState->tagLastSuccessfullETagRequest();
-                //        }
-                //    }
+                QString notificationEtag = args.at(0).toString();
+                QString notificationUser = args.at(1).toString();
+                QString user = _accountState->account()->credentials()->user();
+                if (user == notificationUser && _lastEtag != notificationEtag) {
+                    qDebug() << "* [WebSocket] Compare etag with previous etag: last:" << _lastEtag << ", received:" << notificationEtag << "-> CHANGED";
+                    _lastEtag = notificationEtag;
+                    slotScheduleThisFolder();
+                    _accountState->tagLastSuccessfullETagRequest();
+                }
             });
         });
 
-        QObject::connect(session, &QWamp::Session::started, [&]() {
-            session->join("realm1");
+        QObject::connect(_session.data(), &QWamp::Session::started, [&]() {
+            _session->join("realm1");
         });
-        session->start();
+        _session->start();
     });
 
     QObject::connect(_webSocket.data(), static_cast<void (QAbstractSocket::*)(QAbstractSocket::SocketError err)>(&QAbstractSocket::error), [&](QAbstractSocket::SocketError err) {
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -85,6 +85,7 @@ Folder::Folder(const FolderDefinition &definition,
         qCWarning(lcFolder, "Could not read system exclude file");
 
     connect(_accountState.data(), &AccountState::isConnectedChanged, this, &Folder::canSyncChanged);
+    connect(_accountState->account().data(), &Account::accountChangedCapabilities, this, &Folder::slotAccountCapabilitiesChanged);
     connect(_engine.data(), &SyncEngine::rootEtag, this, &Folder::etagRetreivedFromSyncEngine);
 
     connect(_engine.data(), &SyncEngine::started, this, &Folder::slotSyncStarted, Qt::QueuedConnection);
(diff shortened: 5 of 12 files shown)


cmake/modules/QtVersionAbstraction.cmake
```diff
@@ -28,6 +28,8 @@ if(NOT TOKEN_AUTH_ONLY)
     endif()
 endif()
 
+find_package(Qt5WebSockets)
```
DeepDiver1975: not needed

cmake/modules/QtVersionAbstraction.cmake
```diff
@@ -47,6 +49,13 @@ if(APPLE AND NOT TOKEN_AUTH_ONLY)
     set (QT_LIBRARIES ${QT_LIBRARIES} ${Qt5MacExtras_LIBRARIES})
 endif()
 
+if(Qt5WebSockets_FOUND)
```
DeepDiver1975: not needed

src/CMakeLists.txt
```diff
@@ -26,8 +26,11 @@ elseif(UNIX AND NOT APPLE)
   set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,-z,relro -Wl,-z,now")
 endif()
 
+find_package(Qt5Qml REQUIRED)
```
DeepDiver1975: not needed

src/gui/CMakeLists.txt
```diff
@@ -292,7 +301,7 @@ endif()
 
 add_library(updater STATIC ${updater_SRCS} ${updaterMoc})
 target_link_libraries(updater ${synclib_NAME})
-qt5_use_modules(updater Widgets Network Xml)
+qt5_use_modules(updater Widgets Network Xml Qml)
```
DeepDiver1975: not needed

src/gui/CMakeLists.txt
```diff
@@ -157,6 +157,13 @@ set(3rdparty_SRC
     ../3rdparty/qtsingleapplication/qtlocalpeer.cpp
     ../3rdparty/qtsingleapplication/qtsingleapplication.cpp
     ../3rdparty/qtsingleapplication/qtsinglecoreapplication.cpp
+    ../3rdparty/qmsgpack/src/msgpack.cpp
```
DeepDiver1975: I'm sure there is a nicer way ...

ogoffart: I guess it's fine.

The ideal way would be to have that in a separate .cmake file included in the qmsgpack repository, but since it's not the case, then so be it.

src/gui/folder.cpp
```diff
@@ -108,15 +110,62 @@ Folder::Folder(const FolderDefinition &definition,
...
+          });
+        });
+
+        QObject::connect(session, &QWamp::Session::started, [&]() {
+          session->join("realm1");
+        });
+        session->start();
+      });
+
+    QObject::connect(_webSocket.data(), QOverload<QAbstractSocket::SocketError>::of(&QAbstractSocket::error), [&](QAbstractSocket::SocketError err){
```
ogoffart: QOverload requires Qt 5.7    (before that, we'd have to use static_cast)

(Actually, i've been thinking about copying QOverload within some header so we can use it while still being compatible with Qt 5.6)

CMakeLists.txt
```diff
@@ -218,6 +218,10 @@ add_definitions( -D_WIN32_WINNT=0x0600)
 add_definitions( -DWINVER=0x0600)
 endif( WIN32 )
 
+set(CMAKE_CXX_STANDARD 14)
```
ogoffart: Yay, \o/  finally i'd be able to use auto in lamda :-)

I guess we can do that because we stopped supporting these old debian and centOS that did not even have GCC 4.8 
although this is not strictly required for your patch.

DeepDiver1975: no - not required - but much more fun coding ;-)

src/gui/folder.cpp
DeepDiver1975: well - what ever works - tell me whyt to do

guruz:
CI fails

```
[ 62%] Building CXX object test/CMakeFiles/FolderManTest.dir/__/src/gui/socketapi.cpp.o
/var/lib/jenkins/workspace/cloud-client_client_PR-6195-BV7HAL2B65CGYGSIKPS4QYBZTIWUU3BEFWV3AREJE2WQUV77E2EA/src/gui/folder.cpp:37:19: fatal error: qwamp.h: No such file or directory
 #include "qwamp.h"
                   ^
compilation terminated.
test/CMakeFiles/FolderManTest.dir/build.make:100: recipe for target 'test/CMakeFiles/FolderManTest.dir/__/src/gui/folder.cpp.o' failed
```

@DeepDiver1975 you can build with tests to reproduce

.gitmodules
```diff
@@ -7,3 +7,9 @@
 [submodule "src/3rdparty/libcrashreporter-qt"]
 	path = src/3rdparty/libcrashreporter-qt
 	url = git://github.com/dschmidt/libcrashreporter-qt.git
+[submodule "src/3rdparty/qwamp"]
+	path = src/3rdparty/qwamp
+	url = git@github.com:unitednetworks/qwamp.git
+[submodule "src/3rdparty/qmsgpack"]
+	path = src/3rdparty/qmsgpack
+	url = git@github.com:romixlab/qmsgpack.git
```
guruz: Heads-up to @hodyroff regarding licenses.

CMakeLists.txt
guruz: FYI @jnweiger @dschmidt compilers

src/gui/folder.cpp
```diff
@@ -108,15 +110,62 @@ Folder::Folder(const FolderDefinition &definition,
...
+    QObject::connect(_webSocket.data(), &QTcpSocket::connected, [&]() {
+
+        QString sessionName("sessionName");
+        session = new QWamp::Session(sessionName, *_webSocket.data(), QWamp::Session::MessageFormat::Msgpack, true);
+
+        QObject::connect(session, &QWamp::Session::joined, [&](qint64 s) {
+          qDebug() << "Session joined to realm1 with session ID " << s;
+
+          session->subscribe("etag-changed-channel", [&](const QVariantList& args, const QVariantMap& options){
+              qDebug() << "Event received";
```
guruz: While in general I think the lambdas are cool, we should not become a nodejs with 3x nested callbacks :-)

src/gui/folder.h
```diff
@@ -353,6 +353,7 @@ private slots:
     bool _csyncUnavail;
     bool _proxyDirty;
     QPointer<RequestEtagJob> _requestEtagJob;
+    QScopedPointer<QTcpSocket> _webSocket;
```
guruz: QSslSocket?

guruz:
https://github.com/romixlab/qmsgpack
> By default these Qt types are supported: `Int, UInt, LongLong, ULongLong, Double, QByteArray, Bool, QString, QStringList, QVariantList, QVariantMap`.

Is QString serialized to the same kind of string as you will emit from the server side? So it's interoperable with non-Qt code using WAMP?

> Crossbar.io supports all the favors of WAMP-over-WebSocket, including different serialization formats (JSON and MsgPack) as well as listening transports.

DeepDiver1975:
> Is QString serialized to the same kind of string as you will emit from the server side? So it's interoperable with non-Qt code using WAMP?

In my non-representative local test the string as sent by the server popped up in the client the same way.
But it only contained ascii ...... need to have a second look.

.gitmodules
hodyroff: https://github.com/unitednetworks/qwamp
says Apache 2.0 correct? Thats perfect and I'll add it to the list.
https://github.com/romixlab/qmsgpack/blob/master/LICENSE
is MIT right? Also perfect ... I do need to add both, right?

DeepDiver1975: yes both

src/gui/folder.h
DeepDiver1975: Guide me - can QSslSocket also establish a non-ssl connection?

guruz: https://code.woboq.org/qt5/qtbase/src/network/ssl/qsslsocket.cpp.html#_ZN10QSslSocket22connectToHostEncryptedERK7QStringt6QFlagsIN9QIODevice12OpenModeFlagEEN15QAbstractSocket20NetworkLayerProtocolE

guruz: https://code.woboq.org/qt5/qtbase/src/network/ssl/qsslsocket.cpp.html#_ZN10QSslSocket13connectToHostERK7QStringt6QFlagsIN9QIODevice12OpenModeFlagEEN15QAbstractSocket20NetworkLayerProtocolE

guruz: You can use one of those depending on the account URL http vs https or how would this work? I wonder if the websocket URL should come from capabilities though so the server admin can scale it differently..

Things to keep in mind is the sslErrors signal. We have a ```AbstractSslErrorHandler``` that you can use.

The question is if websockets could return different ssl errors or not. Maybe as first implementation you assume they return the same error as the normal account WebDAV calls ..

DeepDiver1975: The server would expose the websocket url via a capability. The Url will either be ws:// or wss://. 

So depending on using ssl or not I either call connectToHost or connectToHostEncrypted ?

cmake/modules/QtVersionAbstraction.cmake
DeepDiver1975: qwamp needs it - hmmm

DeepDiver1975: maybe we patch qwamp to work without the qwebsocket dependency

DeepDiver1975:
the windows build hates me - qt version? compiler used?

src/gui/folder.h
DeepDiver1975: I'll postpone the ssl topic - websockets with self signed certificates are not working.
Hard to test at the moment.

DeepDiver1975:
> the windows build hates me - qt version? compiler used?

location of error: https://github.com/DeepDiver1975/qwamp/blob/782e65859617e52fa0d92d56551fc05da62c1f8f/qwamp.h#L415

compiler error:

````
[ 69%] Building CXX object src/gui/CMakeFiles/owncloud.dir/folderstatusdelegate.cpp.obj
In file included from /usr/i686-w64-mingw32/sys-root/mingw/include/windows.h:71:0,
                 from /var/lib/jenkins/workspace/cloud-client_client_PR-6195-BV7HAL2B65CGYGSIKPS4QYBZTIWUU3BEFWV3AREJE2WQUV77E2EA/src/common/utility.h:35,
                 from /var/lib/jenkins/workspace/cloud-client_client_PR-6195-BV7HAL2B65CGYGSIKPS4QYBZTIWUU3BEFWV3AREJE2WQUV77E2EA/src/gui/../libsync/account.h:31,
                 from /var/lib/jenkins/workspace/cloud-client_client_PR-6195-BV7HAL2B65CGYGSIKPS4QYBZTIWUU3BEFWV3AREJE2WQUV77E2EA/src/gui/folder.cpp:18:
/var/lib/jenkins/workspace/cloud-client_client_PR-6195-BV7HAL2B65CGYGSIKPS4QYBZTIWUU3BEFWV3AREJE2WQUV77E2EA/src/3rdparty/qwamp/qwamp.h:415:9: error: expected identifier before numeric constant
         ERROR = 8,
         ^
/var/lib/jenkins/workspace/cloud-client_client_PR-6195-BV7HAL2B65CGYGSIKPS4QYBZTIWUU3BEFWV3AREJE2WQUV77E2EA/src/3rdparty/qwamp/qwamp.h:415:9: error: expected '}' before numeric constant
/var/lib/jenkins/workspace/cloud-client_client_PR-6195-BV7HAL2B65CGYGSIKPS4QYBZTIWUU3BEFWV3AREJE2WQUV77E2EA/src/3rdparty/qwamp/qwamp.h:415:9: error: expected unqualified-id before numeric constant
In file included from /var/lib/jenkins/workspace/cloud-client_client_PR-6195-BV7HAL2B65CGYGSIKPS4QYBZTIWUU3BEFWV3AREJE2WQUV77E2EA/src/gui/folder.cpp:37:0:
/var/lib/jenkins/workspace/cloud-client_client_PR-6195-BV7HAL2B65CGYGSIKPS4QYBZTIWUU3BEFWV3AREJE2WQUV77E2EA/src/3rdparty/qwamp/qwamp.h:434:2: warning: extra ';' [-Wpedantic]
 };
  ^
/var/lib/jenkins/workspace/cloud-client_client_PR-6195-BV7HAL2B65CGYGSIKPS4QYBZTIWUU3BEFWV3AREJE2WQUV77E2EA/src/3rdparty/qwamp/qwamp.h:450:1: error: expected declaration before '}' token
 }
 ^
cc1plus: warning: unrecognized command line option '-Wno-gnu-zero-variadic-macro-arguments'
src/gui/CMakeFiles/owncloud.dir/build.make:573: recipe for target 'src/gui/CMakeFiles/owncloud.dir/folder.cpp.obj' failed
make[2]: *** [src/gui/CMakeFiles/owncloud.dir/folder.cpp.obj] Error 1

File: src/libsync/capabilities.cpp [hunk 11]
```diff
@@ -150,4 +150,15 @@ QString Capabilities::invalidFilenameRegex() const
 {
     return _capabilities["dav"].toMap()["invalidFilenameRegex"].toString();
 }
+
+QString Capabilities::getWebSocketUrl() const
+{
+    if (!_capabilities.contains("websocket")) {
+        return "";
```
ogoffart: `return QString()`

(Otherwise you make a useless conversion from char*)

Also later.

src/gui/folder.cpp
```diff
@@ -314,6 +314,62 @@ void Folder::etagRetreivedFromSyncEngine(const QString &etag)
...
+    //initialize wamp connection
+    _webSocket.reset(new QTcpSocket());
+    QUrl webSocketUrl(wsUrl);
+    QObject::connect(_webSocket.data(), &QTcpSocket::connected, [&]() {
+
+        QString sessionName("sessionName");
+        _session.reset(new QWamp::Session(sessionName, *_webSocket.data(), QWamp::Session::MessageFormat::Msgpack, true));
+
+        QObject::connect(_session.data(), &QWamp::Session::joined, [&](qint64 s) {
+            qDebug() << "Session joined to realm1 with session ID " << s;
```
ogoffart: I know this is only temporary and you are going to clean up. In the final PR there should be no qDebug without categories.

src/gui/folder.cpp
```diff
@@ -314,6 +314,62 @@ void Folder::etagRetreivedFromSyncEngine(const QString &etag)
...
+
+            // TODO: add current userid
+            _session->subscribe("files.root-etag-changed.admin", [&](const QVariantList& args, const QVariantMap& options) {
+                qDebug() << "Event received";
+                QString notificationEtag = args.at(0).toString();
```
ogoffart: `QList::at`  crashes if the list don't have enough argument, which can happen, i guess, if the server is sending garbage.
So you can use `value()` which will return an empty variant if out of bounds.  or you should check if there is enough element in the list before.

src/gui/folder.cpp
```diff
@@ -314,6 +314,62 @@ void Folder::etagRetreivedFromSyncEngine(const QString &etag)
...
+    if (wsUrl.isEmpty()) {
+        return;
+    }
+
+    if (_webSocket) {
+        return;
+    }
+    //initialize wamp connection
+    _webSocket.reset(new QTcpSocket());
+    QUrl webSocketUrl(wsUrl);
```
ogoffart: You also probably need to handle the disconnect case, and re try to connect.  Also handle error such as timeout and other kind.
So you will need to move this connection code somewhere else so you can call it when you want to re-connect.

src/gui/folder.cpp
```diff
@@ -314,6 +314,62 @@ void Folder::etagRetreivedFromSyncEngine(const QString &etag)
...
+        });
+
+        QObject::connect(_session.data(), &QWamp::Session::started, [&]() {
+            _session->join("realm1");
+        });
+        _session->start();
+    });
+
+    QObject::connect(_webSocket.data(), static_cast<void (QAbstractSocket::*)(QAbstractSocket::SocketError err)>(&QAbstractSocket::error), [&](QAbstractSocket::SocketError err) {
+        qInfo() << "tcp error: " << err;
```
ogoffart: This probably deserve to be a qCWarning. Don't forget to add a category.  Also keep it in one line.

src/gui/folder.cpp
```diff
@@ -314,6 +314,62 @@ void Folder::etagRetreivedFromSyncEngine(const QString &etag)
...
+        qInfo() << _webSocket->errorString();
+    });
+    qInfo() << "Connecting to websocket " << wsUrl;
+    _webSocket->abort();
```
ogoffart: abort?

src/gui/folder.cpp
```diff
@@ -314,6 +314,62 @@ void Folder::etagRetreivedFromSyncEngine(const QString &etag)
...
+    // TODO: start QSslSocket
+    if (webSocketUrl.scheme() == "ws") {
+        _webSocket->connectToHost(webSocketUrl.host(), webSocketUrl.port());
+    } else {
+        qCCritical(lcFolder) << "Invalid web socket scheme: " << webSocketUrl;
```
ogoffart: qCWarning...   critical is a bit too high.

PR #6200, opened 2017-11-29
Exclude regular expression improvements

ckamm:
A collection of refactor and improvement of the regex-based codepath. Calls to `fnmatch` are entirely gone.


src/csync/csync_exclude.cpp
```diff
@@ -520,7 +548,62 @@ CSYNC_EXCLUDE_TYPE csync_excluded_traversal(CSYNC *ctx, const char *path, int fi
...
+        builderToUse->append(convertToBnameRegexpSyntax(QString::fromUtf8(exclude)));
+    }
+
+    QString pattern = "^(" + exclude_only + ")$|^(" + exclude_and_remove + ")$";
+    _regex.setPattern(pattern);
+    QRegularExpression::PatternOptions patternOptions = QRegularExpression::OptimizeOnFirstUsageOption;
+    if (OCC::Utility::fsCasePreserving())
+        patternOptions |= QRegularExpression::CaseInsensitiveOption;
+    _regex.setPatternOptions(patternOptions);
+    _regex.optimize();
```
ogoffart: Why pass the OptimizeOnFirstUsageOption flag if you optimize now?

src/csync/csync_exclude.cpp
```diff
@@ -520,7 +548,62 @@ CSYNC_EXCLUDE_TYPE csync_excluded_traversal(CSYNC *ctx, const char *path, int fi
...
+    QString s = QRegularExpression::escape(exclude).replace("\\*", ".*").replace("\\?", ".");
+    return s;
+}
+
+void ExcludedFiles::prepare()
+{
+    _nonRegexExcludes.clear();
+
+    // Start out with regexes that would match nothing
+    QString exclude_only = "a^";
```
ogoffart: Maybe this can be done at the end   (`if exclude_only.isEmpty()) exclude_only = "a^";`)

src/csync/csync_exclude.cpp
```diff
@@ -573,39 +579,59 @@ void ExcludedFiles::prepare()
     _nonRegexExcludes.clear();
 
     // Start out with regexes that would match nothing
-    QString exclude_only = "a^";
-    QString exclude_and_remove = "a^";
+    QString patternFileDirKeep = "a^";
+    QString patternFileDirRemove = "a^";
+    QString patternDirKeep = "a^";
+    QString patternDirRemove = "a^";
+    auto regexAppend = [](QString &pattern, QString v) {
```
ogoffart: `const QString &v`

src/csync/csync_exclude.cpp
```diff
@@ -322,10 +309,12 @@ static CSYNC_EXCLUDE_TYPE _csync_excluded_common(c_strlist_t *excludes, const ch
 #endif
 
     /* We create a desktop.ini on Windows for the sidebar icon, make sure we don't sync them. */
-    rc = csync_fnmatch("Desktop.ini", bname, 0);
-    if (rc == 0) {
-        match = CSYNC_FILE_SILENTLY_EXCLUDED;
-        goto out;
+    if (blen == 11) {
+        rc = csync_fnmatch("Desktop.ini", bname, 0);
```
ogoffart: why not bname == "Desktop.ini"   (with Qt::CaseInsensitive for file system that needs it) ?

But also, why should we not put this and all other builtin in the regexp?

src/csync/csync_exclude.cpp
ckamm: Can do.

src/csync/csync_exclude.cpp
ckamm: That was kept from the initial version. I agree that it can be removed.

src/csync/csync_exclude.h
```diff
@@ -37,63 +44,127 @@ enum csync_exclude_type_e {
...
+    auto csyncTraversalMatchFun() const
+        -> std::function<CSYNC_EXCLUDE_TYPE(const char *path, int filetype)>;
+
+public slots:
+    /**
+     * Reloads the exclude patterns from the registered paths.
+     */
+    bool reloadExcludeFiles();
+
+#ifdef CSYNC_TEST
```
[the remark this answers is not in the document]
ckamm: Done

src/csync/csync_private.h
```diff
@@ -148,17 +147,13 @@ struct OCSYNC_EXPORT csync_s {
...
-      c_strlist_t *list_patterns_fnmatch = nullptr;
-
-  } parsed_traversal_excludes;
+  /**
+   * Function used to determine whether an item is excluded
+   * during the update phase.
+   *
+   * See ExcludedFiles in csync_exclude.
+   */
+  std::function<CSYNC_EXCLUDE_TYPE(const char *path, int filetype)> exclude_traversal_fn;
```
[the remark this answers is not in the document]
ckamm: I prefer tying ExcludedFiles and csync together through this function because it's more composable/decoupled. Now csync only ever cares about the traversal exclude match function which makes it easy to hook in a different exclude function for testing.

src/libsync/syncengine.cpp
```diff
@@ -26,6 +26,7 @@
 #include "propagateremotedelete.h"
 #include "propagatedownload.h"
 #include "common/asserts.h"
+#include "configfile.h"
```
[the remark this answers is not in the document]
ckamm: Yes, absolutely.

src/csync/csync_exclude.cpp
ckamm: Yes, that's definitely worthy of further simplification. Some of the patterns above have similar opportunities. Adding to the regex doesn't work easily since these want to be SILENTLY_EXCLUDED.

ckamm:
@ogoffart Addressed comments (separate commits for easy checking, will be rebased away before merge)

ckamm:
@guruz Current benchmarks on my machine.

```
Numbers are 
  2.3 / 2.4 / master
  in ms per call

traversal
  0.00098 / 0.00047 / 0.00046
  reduced to 47% of original

full
  0.0266 / 0.0251 / 0.0019
  reduced to 7% of original
```

The `traversal` number is the primarily important one - it speeds up local discovery for which exclude computations had been bottleneck. The `full` value is primarily used by the file watcher so the huge improvement should be easily noticable if one copies thousands of files into a synced folder.

Note that these values are for a high-end linux machine. Speedups were even larger when we profiled windows/osx during the initial investigation.

guruz:
Impressive :-)

FYI @michaelstingl

PR #6205, opened 2017-12-01
Use a different struct to handle uploading

tcanabrava:
As discussed with @ogoffart this is something that's required if we wanna support file-modificcations during upload in the future.

### DIFF
--- a/src/libsync/propagateupload.cpp
+++ b/src/libsync/propagateupload.cpp
@@ -163,27 +163,36 @@ void PropagateUploadFileCommon::setDeleteExisting(bool enabled)
     _deleteExisting = enabled;
 }
 
-
 void PropagateUploadFileCommon::start()
 {
+    /* Currently the File to upload has the same information as the SyncFileItem,
+     * but the idea is to make a separation between the local file and the file that's
+     * being uploaded, in our current case they are the same, but perhaps we can apply
+     * some filters in the future like 'resize' for pictures and so on.
+     *
+     * this by no means is a finished job, but a first step.
+     */
+    _fileToUpload._file = _item->_file;
+    _fileToUpload._size = _item->_size;
+
     if (propagator()->_abortRequested.fetchAndAddRelaxed(0)) {
         return;
     }
 
     // Check if the specific file can be accessed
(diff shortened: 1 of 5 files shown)


src/libsync/propagateupload.cpp
```diff
@@ -208,10 +218,12 @@ void PropagateUploadFileCommon::slotComputeContentChecksum()
         return;
     }
 
-    const QString filePath = propagator()->getFilePath(_item->_file);
+    const QString filePath = propagator()->getFilePath(_fileToUpload._file);
```
ogoffart: That's `_fileToUpload._path`
Please use that everywhere instead of re-computing it.

src/libsync/propagateupload.h
```diff
@@ -212,6 +212,20 @@ class PropagateUploadFileCommon : public PropagateItemJob
...
 
+    /* This is a minified version of the SyncFileItem,
+     * that holds only the specifics about the file that's
+     * being uploaded.
+     *
+     * This is needed if we wanna apply changes on the file
+     * that's being uploaded while keeping the original on disk.
+     */
+    struct UploadFileInfo {
+      QString _file; /// I'm still unsure if I should use a SyncFilePtr here.
```
ogoffart: I think you don't need this.
You can still use _item->_file in this condition.

src/libsync/propagateupload.cpp
```diff
@@ -163,27 +163,36 @@ void PropagateUploadFileCommon::setDeleteExisting(bool enabled)
...
 
-
 void PropagateUploadFileCommon::start()
 {
+    /* Currently the File to upload has the same information as the SyncFileItem,
+     * but the idea is to make a separation between the local file and the file that's
+     * being uploaded, in our current case they are the same, but perhaps we can apply
+     * some filters in the future like 'resize' for pictures and so on.
+     *
+     * this by no means is a finished job, but a first step.
```
ogoffart: Comments That explains what or why the code do something are great.
But this one comment should better be in the commit message as it does not really explain the code but rather the commit.

ogoffart:
@tcanabrava any progress?

guruz:
@tcanabrava ping?

PR #6206, opened 2017-12-01
Only enable -fstack-protector-strong compiler flag for gcc > 4.9

dragotin:
On older/other compilers this flag is not understood.

Fixes commit ecaa37efbeaf4057825d4bf6a72d10ce4f94e20b

@rockihack any input?

@guruz please pick to 2.4, thanks!


rockihack:
Instead of disabling stack protection completely you can use the non-strong version for older gcc.

```
-fstack-protector --param=ssp-buffer-size=4
```

rockihack:
@dragotin Regarding other compilers: stack-protector-strong also works with clang

guruz:
Changed base to 2.4

PR #6216, opened 2017-12-05
Make rename detection consistent for duplicate file ids

ckamm:
See #6212 and the commit message of the second commit.

### DIFF
--- a/src/csync/csync.cpp
+++ b/src/csync/csync.cpp
@@ -211,14 +211,14 @@ static int _csync_treewalk_visitor(csync_file_stat_t *cur, CSYNC * ctx) {
 
     if (other_file_it == other_tree->cend()) {
         /* Check the renamed path as well. */
-        QByteArray renamed_path = csync_rename_adjust_path(ctx, cur->path);
+        QByteArray renamed_path = csync_rename_adjust_parent_path(ctx, cur->path);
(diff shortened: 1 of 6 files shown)


src/csync/csync_private.h
```diff
@@ -102,6 +102,8 @@ class ByteArrayRef
 
     friend bool operator==(const ByteArrayRef &a, const ByteArrayRef &b)
     { return a.size() == b.size() && qstrncmp(a.data(), b.data(), a.size()) == 0; }
+
+    QByteArray toByteArray() const { return QByteArray(data()); }
```
ogoffart: :-1:  You need to pass the _size as well.

You don't use this, so just remove this

src/csync/csync.cpp
```diff
@@ -314,6 +314,9 @@ int csync_s::reinitialize() {
   local.files.clear();
   remote.files.clear();
 
+  renames.folder_renamed_from.clear();
+  renames.folder_renamed_to.clear();
```
ogoffart: :+1:

ogoffart:
Also, can more test be added? For both commit.  Is the fact that we don't clear not causing bugs when the files are renamed back ?

ckamm:
@ogoffart Updated and added a test for the second commit.

PR #6218, opened 2017-12-06
Set current version number to 2.3.4

settermjd:
This PR addresses https://github.com/owncloud/documentation/issues/3610, by setting the version number to the current stable version (2.3.4).


ogoffart:
:-1:  This file is in the master branch,  if anything, it should be changed to 2.5

The documentation needs to be generated from the right branch.

settermjd:
@ogoffart why would it be changed to 2.5, when the next release is 2.4?

settermjd:
Fair point. You're right, @ogoffart.

Issue #6213, opened 2017-12-04
Refactoring: Move the ConfigFile class out of libsync


ogoffart:
ConfigFile is currently in the libsync folder.
However, the sync engine should not read directly from the config file but should have all its parameter passed in the SyncOptions class.

Rationale: We don't want the tests, owncloudcmd, or other applications to share or mess with the same config file.

guruz:
Only place missing:
```
src/libsync/creds/httpcredentials.cpp:    auto settings = ConfigFile::settingsWithGroup(Theme::instance()->appName());
```

ckamm:
So it looks like we need to move the parts of `HttpCredentials` that interact with the keychain into gui and we're done. I don't think it's critical for 2.6, moving.

ogoffart:
Right, we will refactor the credentials once we can get rid of shibboleth

Issue #2245, opened 2014-09-24
Config directory should be derived Theme::appName(), and should respect XDG settings on linux


danimo:
Currently it's derived from `appNameGui()`, however the idea was that `appNameGui()` can be chosen freely. Now the character set is limited by what any of the filesystems can represent.

Also we should put the config in ~/.config instead of ~/.local

We need to write a migration task that moves the config _before_ anything that accesses the configuration is being initialized.

guruz:
@dragotin @ogoffart @danimo Is this something that should be done together with the other config migration tasks?
Multi-account config migration https://github.com/owncloud/client/issues/3084
Moving Linux config #1409

dragotin:
absolutely, good catch @guruz

Issue #1601, opened 2014-03-21
Incorrect usage of XDG base directories


hefee:
On linux the std location for the configfile is in ~/.local/share (i.e. XDG_DATA_HOME) and not ~/.config/ (i.e. XDG_CONFIG_HOME). This is according to [basedir-spec](http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html) wrong.

See also [debian bug report #741954 "Incorrect usage of XDG base directories"](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=741954)

ogoffart:
Superseded by #3017

hefee:
@ogoffart: This issue hasn't to do anything with #3017. 

3017 is about packaing stuff and the files that are shipped outside the user folder.

This one is about the configfiles that are created for the user aka owncloud.cfg etc.

hodyroff:
@crrodriguez please advise

crrodriguez:
@hodyroff not a packaging issue.. files in stored there are never packaged. distributions buildsystems will reject them if they were anyway.. this is application specific, so it is a client bug.

hodyroff:
@guruz is this something we can/will fix in the client or are we running into troubles with updates then?

guruz:
Not worth to change, that's why we have it P4. Feel free to close if this does/did not come up again.

ogoffart:
Part of https://github.com/owncloud/client/issues/2245

Issue #684, opened 2013-06-12
Windows Roaming Profiles: Make it easier to allow roamed client configuration


danimo:
Current issues:
- [ ] Option required to store ownCloud sync dir outside of roaming area (Microsoft mandates that apps are supposed to write their data into `%USERPROFILE%` only, but that conflicts with roaming profiles, so sysadmins need to be able to specify an alternate, user-writeable default location.)
- [ ] Check if we need to distinguish between user and machine specific settings when storing the configuration
- [ ] Ensure graceful behavior if the client gets started on a second machine

This needs investigation. Also, we need to decide how to provide sensible defaults.

guruz:
Related: https://github.com/owncloud/client/issues/3673

Helios07:
I just saw that there is some progress with this issue. From the comments I see here I am not completely sure if the problem is more regarded to implementation or design. 
We have several users in some of our institutes using roaming profiles. If it would help you, I could ask some of them if they could tell me how they want to use the client the roaming profiles.

danimo:
@Helios07 Please do.

MTRichards:
YES @Helios07 that would be GREAT!

Helios07:
Hi,
I spoke with some of our local administrators and gathered the following:
A typical scenario is that the admins are responsible for a larger amount of computers which are free to use by a certain group of users. The users might log in at any of them. The roaming profiles are used to make it unnecessary for the users to configure their profiles every time they use another computer. This becomes problematic, if huge amounts of data are stored in this profiles because they need to be copied to the computer when logging in. Therefore the admins want to keep the amount of data that is stored small.
If now users are able to choose the local sync path of their ownCloud clients freely, they might choose the profiles and increase the workload at the log in. This is also explained here:
https://anchorhelp.zendesk.com/entries/27754623-Get-Default-Data-Folder-Out-of-Roaming-Profile
As far as I understand, it is possible at the moment to set a certain directory for the sync client and exclude this from the profile replication: https://support.microsoft.com/en-us/kb/814592 Unfortunately, the user is still able to change the default path.
To conclude this, it would be most desirable, if the ownCloud client could be bound to a certain local directory via GPO (group policies).
As an example how this should look like you could have a look at chrome: https://support.google.com/chrome/a/answer/188446?hl=en

This request does not only exist for the Windows client, but also for Linux and MacOS X. 

I hope, this helps :)

MTRichards:
Excellent - thank you @Helios07 !
[...]

PR #6219, opened 2017-12-06
Move config and remove most dependencies form the sync engine.

ogoffart:
See individual commit.

Relates to issues #6213 #2245 #1601 and #684

One of the change is that it moves the config directory from
~/.local/share/data/ownCloud/ to ~/.config/ownCloud/
It migrates by moving the folder to the new location and creating a symlink.

### DIFF
--- a/doc/conffile.rst
+++ b/doc/conffile.rst
@@ -1,13 +1,13 @@
 The ownCloud Client reads a configuration file.  You can locate this configuration file as follows:
 
 On Linux distributions:
-        ``$HOME/.local/share/data/ownCloud/owncloud.cfg``
+        ``$HOME/.config/ownCloud/owncloud.cfg``
 
 On Microsoft Windows systems:
-        ``%LOCALAPPDATA%\ownCloud\owncloud.cfg``
+        ``%APPDATA%\ownCloud\owncloud.cfg``
 
 On MAC OS X systems:
-        ``$HOME/Library/Application Support/ownCloud/owncloud.cfg``
+        ``$HOME/Library/Preferences/ownCloud/owncloud.cfg``
 
 
 The configuration file contains settings using the Microsoft Windows .ini file
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -51,6 +51,7 @@
 #include <QTranslator>
 #include <QMenu>
 #include <QMessageBox>
+#include <QDesktopServices>
 
 class QSocket;
 
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -139,6 +139,11 @@ Application::Application(int &argc, char **argv)
     setupLogging();
     setupTranslations();
 
+    // The timeout is initialized with an environment variable, if not, override with the value from the config
+    ConfigFile cfg;
+    if (!AbstractNetworkJob::httpTimeout)
+        AbstractNetworkJob::httpTimeout = cfg.timeout();
+
     _folderManager.reset(new FolderMan);
 
     connect(this, &SharedTools::QtSingleApplication::messageReceived, this, &Application::slotParseMessage);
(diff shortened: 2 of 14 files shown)


src/gui/application.cpp
```diff
@@ -120,10 +121,27 @@ Application::Application(int &argc, char **argv)
...
     //    setOrganizationName(QLatin1String(APPLICATION_VENDOR));
     setOrganizationDomain(QLatin1String(APPLICATION_REV_DOMAIN));
-    setApplicationName(_theme->appNameGUI());
+    setApplicationName(_theme->appName());
     setWindowIcon(_theme->applicationIcon());
     setAttribute(Qt::AA_UseHighDpiPixmaps, true);
 
+    auto confDir = ConfigFile().configPath();
+    if (!QFileInfo(confDir).exists()) {
+        // Migrate from version <= 2.5
```
ogoffart: Should be < 2.5

src/libsync/abstractnetworkjob.h
```diff
@@ -88,6 +88,9 @@ class OWNCLOUDSYNC_EXPORT AbstractNetworkJob : public QObject
      */
     QString errorStringParsingBody(QByteArray *body = 0);
 
+    /** static variable the HTTP timeout (in seconds)  */
```
ckamm: The docs could state that setting it to 0 means using the default value of 5min. would it make sense to initialize it to 300 and then check the env variable in application.cpp? The current way with 0-means-default is a bit convoluted.

ogoffart: I know. I did it like this because I wanted to still read from the env variable for owncloudcmd.

ogoffart:
Updated so we use %APPDATA% on windows.  
On windows, i also do not create a symbolic link.  (I tried to make a junction, but i spend a long time without success)
I have not tried mac. Can @guruz  or @SamuAlfageme  try on mac?

ogoffart:
Merged anyway, if this cause problem, it can be fixed later

src/libsync/abstractnetworkjob.h
```diff
@@ -88,6 +88,10 @@ class OWNCLOUDSYNC_EXPORT AbstractNetworkJob : public QObject
      */
     QString errorStringParsingBody(QByteArray *body = 0);
 
+    /** static variable the HTTP timeout (in seconds). If set to 0, the default will be used
+     */
+    static int httpTimeout;
```
guruz: It would be nice to name variables always with the unit, e.g. httpTimeoutSeconds

src/libsync/configfile.cpp
```diff
@@ -252,26 +252,18 @@ QVariant ConfigFile::getPolicySetting(const QString &setting, const QVariant &de
 
 QString ConfigFile::configPath() const
 {
-#ifndef TOKEN_AUTH_ONLY
     if (_confDir.isEmpty()) {
```
guruz: Can you make a comment on https://github.com/owncloud/client/issues/684 and explain what you change for 2.5? Also does the milestone need to be set for that issue?

guruz:
Seems to work on macOS.
```
gui.application: Migrating old config from "/Users/guruz/Library/Application Support/ownCloud" to "/Users/guruz/Library/Preferences/ownCloud/"
```
...
```
lrwxr-xr-x  1 guruz  staff    42B Dec 12 12:20 /Users/guruz/Library/Application Support/ownCloud -> /Users/guruz/Library/Preferences/ownCloud/
```

doc/conffile.rst
```diff
@@ -1,13 +1,13 @@
 The ownCloud Client reads a configuration file.  You can locate this configuration file as follows:
```
guruz: FYI @michaelstingl @cdamken etc
For 2.5.0

src/libsync/configfile.cpp
guruz: ^^ FYI

PR #6220, opened 2017-12-06
Attempt to fix Jenkins

ogoffart:
Just to find out what's the problem on the CI

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,7 +16,7 @@ node('CLIENT') {
         sh '''rm -rf build
 		mkdir build
 		cd build
-		cmake -DCMAKE_BUILD_TYPE="Debug" -DUNIT_TESTING=1 -DWITH_TESTING=1 -DBUILD_WITH_QT4=OFF ..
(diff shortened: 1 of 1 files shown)


ogoffart:
The error is

```
../libsync/libowncloudsync.so.2.4.0: undefined reference to `QKeychain::Job::setKey(QString const&)'
```

But /usr/lib/x86_64-linux-gnu/libqt5keychain.so contains

```
0000000000006f10 g    DF .text  000000000000000d  Base        QKeychain::WritePasswordJob::setKey(QString const&)
00000000000071b0 g    DF .text  000000000000000d  Base        QKeychain::DeletePasswordJob::setKey(QString const&)
0000000000006dd0 g    DF .text  000000000000000d  Base        QKeychain::ReadPasswordJob::setKey(QString const&)
```

Maybe the headers and the binary don't match.

ogoffart:
The symbol was changed in the commit e6e828c33b09f84a954f9885d0b4efd706bcd439 in qtkeychain.

ogoffart:
So I added -DCMAKE_PREFIX_PATH=/var/lib/jenkins/qt/5.6.2 in the Jenkins file so it picks the right Qt5Keychain,  otherwise it tries to link against the system qtkeychain which is too old.

I am not sure why the build worked with Qt 5.5's qtkeychain. Maybe because it was older and that it could still link with the system's qtkeychain without error.

PR #6224, opened 2017-12-08
Remove use of deprecated Qt API

ogoffart:
Now that we require a recent Qt, we can get rid of most calls of functions deprecated since Qt 5.0


ckamm:
Nice!

PR #6228, opened 2017-12-11
Journal: Store and read inodes as 64 bit unsigned

ckamm:


ckamm:
Please play with it a bit. Somewhat surprisingly the change intValue -> int64Value didn't seem to be the crucial bit, but adding the case for ULongLong. I wonder if that means inodes are stored oddly in the db!

ckamm:
Updated: There were two bugs. uint64 values that didn't convert without loss to int64 values were not stored correctly in the database!

guruz:
macOS: ```100% tests passed, 0 tests failed out of 34```

Issue #6211, opened 2017-12-04
Can't build 2.4.0 beta1 on alpha&hppa platform


hefee:
for 2.3.3 i could build owncloud-client on alpha and hppa plafroms [owncloud-client(2.3.3+dfsg-2)/alpha](https://buildd.debian.org/status/fetch.php?pkg=owncloud-client&arch=alpha&ver=2.3.3%2Bdfsg-2&stamp=1505065649&raw=0), [owncloud-client(2.3.3+dfsg-2)/hppa](https://buildd.debian.org/status/fetch.php?pkg=owncloud-client&arch=hppa&ver=2.3.3%2Bdfsg-2&stamp=1505066516&raw=0)

With current 2.4.0 beta1 it fails with:
* [owncloud-client(2.4.0~beta1+dfsg-2)/alpha](https://buildd.debian.org/status/fetch.php?pkg=owncloud-client&arch=alpha&ver=2.4.0%7Ebeta1%2Bdfsg-1&stamp=1512370048&raw=0)
* [owncloud-client(2.4.0~beta1+dfsg-2)/hppa](https://buildd.debian.org/status/fetch.php?pkg=owncloud-client&arch=hppa&ver=2.4.0%7Ebeta1%2Bdfsg-1&stamp=1512380458&raw=0)

```
[  4%] Linking CXX shared module ownclouddolphinactionplugin.so
cd "/<<BUILDDIR>>/owncloud-client-2.4.0~beta1+dfsg/obj-alpha-linux-gnu/shell_integration/dolphin" && /usr/bin/cmake -E cmake_link_script CMakeFiles/ownclouddolphinactionplugin.dir/link.txt --verbose=1
[...]

PR #6229, opened 2017-12-11
Enable stack protector on windows / macos and disable for hppa.

rockihack:
This pull request enables stack-protector for windows / macos
and disables it for hppa builds (currently not supported).

I tested the windows and macos builds for a couple of weeks and it works just fine.
Hppa build is untested!

Follow up of: https://github.com/owncloud/client/pull/6040
Fixes: https://github.com/owncloud/client/issues/6211


src/CMakeLists.txt
```diff
@@ -8,30 +8,29 @@ if(NOT TOKEN_AUTH_ONLY)
...
-  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,--nxcompat -Wl,--dynamicbase")
-  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--nxcompat -Wl,--dynamicbase")
-elseif(UNIX AND NOT APPLE)
-  if (CMAKE_CXX_COMPILER_ID MATCHES "GNU")
-    if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 4.9)
-      set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fstack-protector-strong")
-      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fstack-protector-strong")
-    else()
+if(NOT MSVC)
+  if(NOT (CMAKE_SYSTEM_PROCESSOR MATCHES "^(parisc|hppa)"))
```
ckamm: This if used to have also a `AND NOT APPLE`. @guruz Will setting these compiler flags be ok on osx?

guruz: Yep should work, default compiler on macOS is clang

hefee: for sure you need also an |alpha cause also alpha platform is affective.

guruz:
We reverted this partly for 2.4.0 in 298f1ab57000afd7acb8c0ddbddca8a1fc6cd92a because our packages don't ship libssp-0.dll which is suddenly needed when enabling the compile flag.

Issue #5106, opened 2016-08-05
Conflict file creation on bad network connections


jancborchardt:
I just had a few days of bad connection and as a results a few conflict files of notes popped up. (During the time I basically exclusively edited stuff locally on my laptop so I wonder why there are conflict files, but I might be wrong.)

Anyway – I noticed a pattern that in basically all cases the `_conflict` files are the newer and correct ones. It seems that the algorithm assumes the server has the »right« version.

Instead I would propose that the latest / last modified file should get the »clean« name and the older one should get the `_conflict` suffix.

What do you think @dragotin @danimo @rullzer?

dragotin:
No. I know its debatable, but we always had the policy *server wins*. I am not seeing that we are going to change that now.

jturcotte:
Let's say you have userA and userB:
- userA modifies fileA (version 1) and sync, fileA(v1) is on the server
- userB modifies fileA meanwhile (version 2), and then syncs, and get a conflict. fileA(v1) is renamed to fileA_conflict and fileA(v2) is uploaded
- userA sync, and his file gets overwritten automatically on his local station. fileA(v1) is lost without anybody doing any conflict resolution.

On the other hand, if the local file (the last one trying to reach the server) is the one renamed to _conflict like we currently do, this means that a human is forced to decide which file gets chosen as the right one, possibly integrating both changes in the result. I also think that it's better to let the server "win" to force this process to happen without losing any data.

jturcotte:
The weird thing I think is that you get conflict files and that you don't care about what's on the server. If anybody has been updating those files, you shouldn't just throw away their modifications, both file would be "correct".

If there wasn't any modification, maybe the solution to your problem would be to make sure that we create conflict files only if there is conflicting data.

jancborchardt:
> If there wasn't any modification, maybe the solution to your problem would be to make sure that we create conflict files only if there is conflicting data.

Yeah, I guess that’s the main issue. Unfortunately I don’t know why there were conflict files to begin with. It was with 5 notes text files which I definitely only edited on my laptop. And the internet connection was very bad. So not sure how to help debug this further.

guruz:
> If there wasn't any modification, maybe the solution to your problem would be to make sure that we create conflict files only if there is conflicting data.

We check the file byte by byte.

@jancborchardt Would be good to see a diff of the files to reconstruct what happened.

jturcotte:
[...]

PR #6232, opened 2017-12-11
SyncEngine: Recover when the PUT reply (or chunkin's MOVE) is lost

ogoffart:
This can happen if the upload of a file is finished, but we just got
disconnected right before recieving the reply containing the etag.
So nothing was save din the DB, and we are not sure if the server
recieved the file properly or not. Further local update of the file
will cause a conflict.
   
In order to fix this, store the checksum of the uploading file in
the uploadinfo table of the local db (even if there is no chunking
involved).  And when we have a conflict, check that it is not because
of this situation by checking the entry in the uploadinfo table.
   
Issue #5106

This goes to master because it is an old issue and the fix is quite intrusive as it even add a row in the database.

### DIFF
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -389,6 +389,7 @@ bool SyncJournalDb::checkConnect()
                         "errorcount INTEGER,"
                         "size INTEGER(8),"
                         "modtime INTEGER(8),"
+                        "contentChecksum TEXT,"
                         "PRIMARY KEY(path)"
                         ");");
 
@@ -611,15 +612,15 @@ bool SyncJournalDb::checkConnect()
     }
 
     _getUploadInfoQuery.reset(new SqlQuery(_db));
-    if (_getUploadInfoQuery->prepare("SELECT chunk, transferid, errorcount, size, modtime FROM "
+    if (_getUploadInfoQuery->prepare("SELECT chunk, transferid, errorcount, size, modtime, contentChecksum FROM "
                                      "uploadinfo WHERE path=?1")) {
         return sqlFail("prepare _getUploadInfoQuery", *_getUploadInfoQuery);
     }
 
     _setUploadInfoQuery.reset(new SqlQuery(_db));
     if (_setUploadInfoQuery->prepare("INSERT OR REPLACE INTO uploadinfo "
-                                     "(path, chunk, transferid, errorcount, size, modtime) "
-                                     "VALUES ( ?1 , ?2, ?3 , ?4 ,  ?5, ?6 )")) {
+                                     "(path, chunk, transferid, errorcount, size, modtime, contentChecksum) "
+                                     "VALUES ( ?1 , ?2, ?3 , ?4 ,  ?5, ?6 , ?7 )")) {
         return sqlFail("prepare _setUploadInfoQuery", *_setUploadInfoQuery);
     }
 
@@ -847,6 +848,16 @@ bool SyncJournalDb::updateMetadataTableStructure()
             re = false;
         }
         commitInternal("update database structure: add contentChecksumTypeId col");
+    }
+
+    if (!tableColumns("uploadinfo").contains("contentChecksum")) {
+        SqlQuery query(_db);
+        query.prepare("ALTER TABLE uploadinfo ADD COLUMN contentChecksum TEXT;");
+        if (!query.exec()) {
+            sqlFail("updateMetadataTableStructure: add contentChecksum column", query);
+            re = false;
+        }
+        commitInternal("update database structure: add contentChecksum col for uploadinfo");
     }
 
 
@@ -1472,6 +1483,7 @@ SyncJournalDb::UploadInfo SyncJournalDb::getUploadInfo(const QString &file)
             res._errorCount = _getUploadInfoQuery->intValue(2);
             res._size = _getUploadInfoQuery->int64Value(3);
             res._modtime = _getUploadInfoQuery->int64Value(4);
+            res._contentChecksum = _getUploadInfoQuery->baValue(5);
             res._valid = ok;
         }
     }
@@ -1494,6 +1506,7 @@ void SyncJournalDb::setUploadInfo(const QString &file, const SyncJournalDb::Uplo
         _setUploadInfoQuery->bindValue(4, i._errorCount);
         _setUploadInfoQuery->bindValue(5, i._size);
         _setUploadInfoQuery->bindValue(6, i._modtime);
+        _setUploadInfoQuery->bindValue(7, i._contentChecksum);
 
         if (!_setUploadInfoQuery->exec()) {
             return;
@@ -2001,7 +2014,8 @@ bool operator==(const SyncJournalDb::UploadInfo &lhs,
         && lhs._modtime == rhs._modtime
         && lhs._valid == rhs._valid
         && lhs._size == rhs._size
-        && lhs._transferid == rhs._transferid;
+        && lhs._transferid == rhs._transferid
+        && lhs._contentChecksum == rhs._contentChecksum;
 }
 
 } // namespace OCC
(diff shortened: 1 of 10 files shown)


src/common/syncjournaldb.cpp
```diff
@@ -1494,6 +1506,7 @@ void SyncJournalDb::setUploadInfo(const QString &file, const SyncJournalDb::Uplo
         _setUploadInfoQuery->bindValue(4, i._errorCount);
         _setUploadInfoQuery->bindValue(5, i._size);
         _setUploadInfoQuery->bindValue(6, i._modtime);
+        _setUploadInfoQuery->bindValue(7, i._contentChecksum);
```
ckamm: There is a `operator==` for UploadInfo that you probably want to adjust too (or remove entirely!)

src/csync/csync_reconcile.cpp
```diff
@@ -316,6 +317,33 @@ static int _csync_merge_algorithm_visitor(csync_file_stat_t *cur, CSYNC * ctx) {
...
+                        // we recieved the etag (Issue #5106)
+                        auto up = ctx->statedb->getUploadInfo(cur->path);
+                        if (up._valid && up._contentChecksum == remoteChecksumHeader) {
+                            // Solve the conflict into an upload, or nothing
+                            auto remoteNode = ctx->current == REMOTE_REPLICA ? cur : other;
+                            auto localNode = ctx->current == REMOTE_REPLICA ? other : cur;
+                            remoteNode->instruction = CSYNC_INSTRUCTION_NONE;
+                            localNode->instruction = up._modtime == localNode->modtime ? CSYNC_INSTRUCTION_UPDATE_METADATA : CSYNC_INSTRUCTION_SYNC;
+
+                            // Update the etag and other server metadata in the journal already
```
ckamm: That's not safe to do in the SYNC case, is it?

src/common/syncjournaldb.cpp
```diff
@@ -389,6 +389,7 @@ bool SyncJournalDb::checkConnect()
                         "errorcount INTEGER,"
                         "size INTEGER(8),"
                         "modtime INTEGER(8),"
+                        "contentChecksum TEXT,"
```
guruz: Can this contain multiple checksums like the other column called like that? if yes, please add a plural "s" so in future we know for sure

src/common/syncjournaldb.h
```diff
@@ -112,6 +112,7 @@ class OCSYNC_EXPORT SyncJournalDb : public QObject
         qint64 _modtime;
         int _errorCount;
         bool _valid;
+        QByteArray _contentChecksum;
```
guruz: plural s?

src/csync/csync_reconcile.cpp
```diff
@@ -316,6 +317,33 @@ static int _csync_merge_algorithm_visitor(csync_file_stat_t *cur, CSYNC * ctx) {
...
+                            if (ctx->statedb->getFileRecord(remoteNode->path, &rec)) {
+                                rec._path = remoteNode->path;
+                                rec._etag = remoteNode->etag;
+                                rec._fileId = remoteNode->file_id;
+                                rec._modtime = remoteNode->modtime;
+                                rec._type = remoteNode->type;
+                                rec._fileSize = remoteNode->size;
+                                rec._remotePerm = remoteNode->remotePerm;
+                                rec._checksumHeader = remoteNode->checksumHeader;
+                                ctx->statedb->setFileRecordMetadata(rec);
```
guruz: For code reading clarity one could set is_conflict to false here?

src/libsync/propagateuploadv1.cpp
```diff
@@ -43,10 +43,22 @@ void PropagateUploadFileV1::doStartUpload()
...
+        // Write the checksum in the database, so we can protect ourselfs from issue #5106
+        SyncJournalDb::UploadInfo pi;
+        pi._valid = true;
+        pi._chunk = 0;
+        pi._transferid = _transferId;
+        pi._modtime = _item->_modtime;
+        pi._errorCount = 0;
+        pi._contentChecksum = _item->_checksumHeader;
+        propagator()->_journal->setUploadInfo(_item->_file, pi);
+        propagator()->_journal->commit("Upload info");
```
ckamm: This is done to prevent issues in cases when there's a single chunk upload that succeeds but we never hear back? Maybe add a one-liner explaining that? How about always setting upload info here, regardless of whether checksumHeader is empty?

test/syncenginetestutils.h
```diff
@@ -744,10 +755,32 @@ class FakeHangingReply : public QNetworkReply
...
+    {
+    }
+    quint64 _delayMs;
+
+    void respond() override
+    {
+        QTimer::singleShot(_delayMs, static_cast<OriginalReply *>(this), [this] {
+            // Explicit call to bases's respond();
+            this->OriginalReply::respond();
+        });
```
ckamm: Why doesn't this mean the OriginalReply's 0 timer *and* the delayed timer fire?

src/common/syncjournaldb.cpp
guruz: (maybe i mix this up and multiple checksums was only possible for the server or for the OC-checksum header?)

ckamm: Yes, the db only has one checksum

guruz:
```
FAIL!  : TestChunkingNG::testLateAbortHard() 'a0' returned FALSE. ()
   Loc: [/var/lib/jenkins/workspace/cloud-client_client_PR-6232-LCFBVM4WSGOES5SFCXHVF6PYM6TCD6LOL2BDQZOVL46DYHJVYUJQ/test/testchunkingng.cpp(172)]
```

test/syncenginetestutils.h
ogoffart: I don't understand the question.

src/csync/csync_reconcile.cpp
ogoffart: I think it is.  Why wouldn't it?

ogoffart:
Commit updated.

@ckamm:  please check  the part of your previous test which i have removed.  The thing is that we "resolve" the conflict in reconcile, so the hook don't see the item anymore.  I hope this does not void your test. Should another test be made?

test/syncenginetestutils.h
ckamm: The constructor of the underlying class sets up a single shot timer already. Won't this just start a second timer?

ckamm:
@ogoffart Since you improved the behavior for the tested case I think changing that test is exactly correct. It would be nice to see the UPDATE_METADATA, but should be fine as-is.

src/csync/csync_reconcile.cpp
ckamm: So, assume we end up in the sync case. That means we have an upload info with the same checksum as the remote, but the mtime differs. We set a SYNC instruction to trigger the download but also immediately adjust the database entry to have the new remote metadata.

Say the sync is aborted before the propagation of that SYNC actually happens. Would the following sync run still do the right thing given that the metadata in the db has already been updated? I expect it would not trigger the SYNC again?

src/common/syncjournaldb.h
ogoffart: it's just one

test/syncenginetestutils.h
ogoffart: Harder to override in a generic way.

src/csync/csync_reconcile.cpp
ckamm: Thanks for the explanation!

ogoffart:
CI error is
```
12-12 14:46:24:528 [ warning sync.propagator ]:	Could not complete propagation of "A/a0" by OCC::PropagateDownloadFile(0x148ef80) with status 8 and error: "The download would reduce free local disk space below the limit"
```

As it tries to download a 150MB file.

PR #6235, opened 2017-12-12
Upload conflict files

ckamm:
If the server has the 'uploadConflictFiles' capability conflict
files will be uploaded instead of ignored.

Uploaded conflict files have the following headers set during upload
  OC-Conflict: 1
  OC-ConflictBaseFileId: 172489174instanceid
  OC-ConflictBaseMtime: 1235789213
  OC-ConflictBaseEtag: myetag
when the data is available. Downloads accept the same headers in return
when downloading a conflict file.

In the absence of server support clients will identify conflict files
through the file name pattern and attempt to deduce the base fileid.
Base etag and mtime can't be deduced though.

The upload job for a new conflict file will be triggered directly from
the job that created the conflict file now. No second sync run is
necessary anymore.

This commit does not yet introduce a 'username' like identifier that
automatically gets added to conflict file filenames (to name the files
foo_conflict-Fred-1345.txt instead of just foo_conflict-1345.txt).

See #4557
@PVince81


ckamm:
@PVince81 @ogoffart I'd be happy if you could think through what the server and client might do with the conflict information and whether this is sending enough info. I think fileid + etag should be enough to clearly identify the base version. The mtime is just convenience.

src/libsync/capabilities.cpp
```diff
@@ -148,4 +149,14 @@ QString Capabilities::invalidFilenameRegex() const
...
 }
+
+bool Capabilities::uploadConflictFiles() const
+{
+    static auto envIsSet = !qEnvironmentVariableIsEmpty("OWNCLOUD_UPLOAD_CONFLICT_FILES");
+    static int envValue = qEnvironmentVariableIntValue("OWNCLOUD_UPLOAD_CONFLICT_FILES");
+    if (envIsSet)
+        return envValue != 0;
+
+    return _capabilities["uploadConflictFiles"].toBool();
```
ogoffart: Does that really make sense in the capabilities?
We were talking about a config endpoint. @PVince81, what do you say?

src/libsync/owncloudpropagator.cpp
```diff
@@ -739,6 +739,15 @@ void PropagatorCompositeJob::slotSubJobAbortFinished()
     }
 }
 
+void PropagatorCompositeJob::appendJob(PropagatorJob *job)
+{
+    // Some jobs need to know about their parent
+    if (auto download = qobject_cast<PropagateDownloadFile *>(job))
+        download->setCompositeParent(this);
```
ogoffart: Maybe this could be added in PropagatorJob itself so all the jobs would know about their parrent

src/libsync/propagatedownload.cpp
```diff
@@ -639,6 +644,27 @@ void PropagateDownloadFile::slotGetFinished()
         return;
     }
 
+    // Did the file come with conflict headers? If so, store them now!
+    // If we download conflict files but the server doesn't send conflict
+    // headers, the record will be established by SyncEngine::conflictRecordMaintenance.
+    // (we can't reliably determine the file id of the base file here,
+    // it might still be downloaded in a parallel job and not exist in
+    // the database yet!)
+    if (propagator()->account()->capabilities().uploadConflictFiles()
```
ogoffart: We should store in the database even if the capability is not set, i'd say

src/libsync/propagatedownload.cpp
```diff
@@ -820,6 +848,37 @@ void PropagateDownloadFile::downloadFinished()
...
+        ConflictRecord conflictRecord;
+        conflictRecord.path = conflictFileName.toUtf8();
+        conflictRecord.baseModtime = _item->_previousModtime;
+
+        SyncJournalFileRecord baseRecord;
+        if (propagator()->_journal->getFileRecord(_item->_originalFile, &baseRecord)) {
+            conflictRecord.baseEtag = baseRecord._etag;
+            conflictRecord.baseFileId = baseRecord._fileId;
+        } else {
+            // We might very well end up with no fileid/etag for new/new conflicts
```
ogoffart: getFileRecord returns false only if there is a database error!
If the entry is not in the database, baseRecord.isValid() == false

src/libsync/propagatedownload.cpp
```diff
@@ -820,6 +848,37 @@ void PropagateDownloadFile::downloadFinished()
...
+            conflictItem->_file = conflictFileName;
+            conflictItem->_type = SyncFileItem::File;
+            conflictItem->_direction = SyncFileItem::Up;
+            conflictItem->_instruction = CSYNC_INSTRUCTION_NEW;
+            conflictItem->_modtime = conflictModTime;
+            conflictItem->_size = _item->_previousSize;
+            if (_compositeParent) {
+                _compositeParent->appendTask(conflictItem);
+            } else {
+                qWarning() << "ZZZ null composite parent!";
```
ogoffart: ZZZ

src/libsync/propagateupload.cpp
```diff
@@ -605,6 +605,22 @@ QMap<QByteArray, QByteArray> PropagateUploadFileCommon::headers()
         //  csync_owncloud.c's owncloud_file_id always strips the quotes.
         headers["If-Match"] = '"' + _item->_etag + '"';
     }
+
+    // Set up a conflict file header pointing to the original file
+    if (propagator()->account()->capabilities().uploadConflictFiles()
```
ogoffart: Why do we need to check the capabilities here?

src/libsync/syncengine.cpp
```diff
@@ -329,6 +329,51 @@ void SyncEngine::deleteStaleErrorBlacklistEntries(const SyncFileItemVector &sync
...
 }
 
+void SyncEngine::conflictRecordMaintenance()
+{
+    QByteArrayList conflictRecordPaths;
+
+    // Remove stale conflict entries from the database
+    if (account()->capabilities().uploadConflictFiles()) {
+        // If conflicts are uploaded, we just wipe all entries that no
+        // longer have a corresponding metadata entry.
```
ogoffart: That sounds wrong to me,  what if the file failed to upload? It would not have a metadata, but the file would still exist on the disk.

src/libsync/syncengine.cpp
```diff
@@ -329,6 +329,51 @@ void SyncEngine::deleteStaleErrorBlacklistEntries(const SyncFileItemVector &sync
...
+        }
+    }
+
+    // Did the sync see any conflict files that don't yet have records?
+    // If so, add them now.
+    //
+    // This happens when the conflicts table is new or when conflict files
+    // are downlaoded but the server doesn't send conflict headers.
+    for (const auto &path : _seenFiles) {
+        auto bapath = path.toUtf8();
```
ogoffart: (Would be nice: have a Utility::isConflictFile that works with QString and avoid this allocation, _seenFiles can be big)

src/common/syncjournalfilerecord.h
```diff
@@ -110,6 +110,31 @@ class OCSYNC_EXPORT SyncJournalErrorBlacklistRecord
 
     bool isValid() const;
 };
+
+/** Represents a conflict in the conflicts table */
+class OCSYNC_EXPORT ConflictRecord
+{
+public:
+    /// Path to the conflict file (folder-relative)
```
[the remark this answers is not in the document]
ckamm: Agreed, will expand on documentation here.

src/common/utility.cpp
```diff
@@ -551,32 +566,27 @@ bool Utility::isConflictFile(const char *name)
         bname = name;
     }
 
-    if (std::strstr(bname, "_conflict-"))
-        return true;
-
-    if (shouldUploadConflictFiles()) {
-        // For uploads, we want to consider files with any kind of username tag
-        // as conflict files. (pattern *_conflict_*-)
-        const char *startOfMarker = std::strstr(bname, "_conflict_");
```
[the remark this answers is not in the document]
ckamm: The `_conflict_` pattern only appeared with the `CSYNC_CONFLICT_FILE_USERNAME` option which was internal. Do you think we need to worry about these? (I did initially, but after our discussion on IRC removed the code)

src/common/utility.cpp
```diff
@@ -551,32 +566,27 @@ bool Utility::isConflictFile(const char *name)
...
-            return true;
-        }
-    }
-
-    return false;
+    return std::strstr(bname, "_conflict-");
 }
 
-bool Utility::shouldUploadConflictFiles()
+QByteArray Utility::conflictFileBaseName(const QByteArray &conflictName)
```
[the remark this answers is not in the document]
ckamm: That's not what the function is used for. If these's an entry in the conflicts table, we can already find the base file name by getting the filerecord for the base file id.

This function is currently used when we don't have an entry in the conflicts table but want to create one. Particularly:
* When downloading a conflict file when server doesn't support the OC-Conflict headers. This function gives us the probable base filename and thereby the fileid.
* When the client is updated initially and the conflicts table is entirely empty, we're in a similar situation: We can see that a file looks like a conflict file but we don't know the base.

src/libsync/capabilities.cpp
ckamm: It's in capabilities for convenience. It's not a capability and I'm fine with having it available through a config endpoint. This isn't a new issue, many things in capabilities are actually options.

src/libsync/owncloudpropagator.cpp
ckamm: Sure. I didn't want to needlessly store the parent in all jobs. But since we don't create all job instances upfront anymore that may not be an issue. Will do.

src/libsync/propagatedownload.cpp
ckamm: Why would a server without the uploadConflictFiles capability send us a file that includes conflict headers? It'd be odd: we'd download the file and store it in the db, but due to the exclude pattern I don't think we'd upload any local changes.

src/libsync/propagateupload.cpp
ckamm: True, unnecessary

src/libsync/syncengine.cpp
ckamm: Very good point. Looks like I do need to check whether the actual files still exist here.

src/common/utility.cpp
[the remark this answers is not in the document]
ogoffart: ah ok.  Yes, we can remove it.

src/libsync/propagatedownload.cpp
ogoffart: Because we change the option locally with the env variable.
Or the files were uploaded before the server changed its config.

ckamm:
@ogoffart Updated.

Now that I thought about things again: I chose to only store the fileid of the base conflict file to ensure we wouldn't need to add a bunch of maintenance code around the base file being renamed.

What happens if the conflict file itself is renamed though? Should it still be considered a conflict file then? Maybe the conflict record should store the conflict file as a fileid only as well? (queries by path would still work through metadata joins)

Also, currently the code sometimes assumes only files that have `isConflictFile()` return true can have conflict records (optimization only).

ogoffart:
I think if the conflict file is renamed, then it is not longer a conflict file as the user "resolved" the conflict by renaming it.

But yes, another client implementation (or version) could use another pattern to name conflict file, so we should not rely on isConflictFile

src/libsync/propagatedownload.cpp
```diff
@@ -820,6 +842,34 @@ void PropagateDownloadFile::downloadFinished()
...
+        if (propagator()->account()->capabilities().uploadConflictFiles()) {
+            SyncFileItemPtr conflictItem = SyncFileItemPtr(new SyncFileItem);
+            conflictItem->_file = conflictFileName;
+            conflictItem->_type = SyncFileItem::File;
+            conflictItem->_direction = SyncFileItem::Up;
+            conflictItem->_instruction = CSYNC_INSTRUCTION_NEW;
+            conflictItem->_modtime = conflictModTime;
+            conflictItem->_size = _item->_previousSize;
+            ASSERT(_compositeParent);
+            _compositeParent->appendTask(conflictItem);
```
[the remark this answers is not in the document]
ckamm: Done now, had forgotten

ogoffart:
```
7FLBZ2TD5QGUIIS4JGAS4FYDQ4XWOQ3HYN4LKO2IA6VEBZSYZAQA/test/testsyncconflict.cpp:105:10: error: no matching function for call to ‘FakeFolder::setServerOverride(TestSyncConflict::testUploadAfterDownload()::<lambda(QNetworkAccessManager::Operation, const QNetworkRequest&)>)’
         });
          ^
```

ogoffart:
Ah yes, the signature of setServerOverride has changed recently to add the QIODevice*  you may want to rebase.

ckamm:
Rebased and squashed

PR #6239, opened 2017-12-13
Config: Look for exclude file in a relative path.

dragotin:
If the application binary is not installed in /usr/bin the client
with this patch considers to check the relative location
`../../etc/owncloud-client/` to find the system exclude.

This is an important bit for AppImage based packages of the client,
as this runs from a temporar mountpoint and the system file can not
be found under /etc.


ckamm:
(applied clang-format and fixed a minor spelling issue)

Issue #6236, opened 2017-12-12
[Notifications] Open "link" in browser 


SamuAlfageme:
This is quite similar to the FR in https://github.com/owncloud/client/issues/5922, this time about the "`link`" property of the notifications we get from the server. This got my eye on Marketplace notifications. 

Tagging as discussion to see if this is something that would bring value to users. e.g. an admin clicking on a desktop notification to open an app's page in marketplace to update the version: 

```json
{
   "ocs":{
      "meta":{
         "status":"ok",
         "statuscode":200,
         "message":null,
         "totalitems":"",
         "itemsperpage":""
      },
      "data":[
         {
            "notification_id":1,
            "app":"market",
            "user":"admin",
            "datetime":"2017-12-12T07:54:04+00:00",
            "object_type":"market",
            "object_id":"0.2.3",
            "subject":"Update for Market to version 0.2.3 is available.",
            "message":"",
            "link":"<server>/apps/market/#/app/market",
            "actions":[],
            "icon":""
         }
      ]
   }
}
```

This one gets rendered as: 

![screenshot 2017-12-12 16 21 00](https://user-images.githubusercontent.com/2644445/33892190-78567c70-df58-11e7-8054-2956e9dd2dd4.png)

We could simply add a new button to open it in browser or make the whole thing clickable.

guruz:
Indeed there is no link option from command line at least. I'll hack the source or so.

```
notifications:generate [-u|--user USER] [-g|--group GROUP] [--] <subject> [<message>]
..
sudo -u http ./occ notifications:generate -u guruz hello_subject "hello this is guruz speaking"
```

FYI @tomneedham

guruz:
> Indeed there is no link option from command line at least. 

https://github.com/owncloud/notifications/pull/172

guruz:
-> https://github.com/owncloud/client/pull/6453

ckamm:
Tested manually on a oc-master instance by using the `occ notifications:generate` command.

PR #6240, opened 2017-12-13
Notifications: Display the link in the subject line.

dragotin:
Added a new data called linkText. If the link in the activity JSON has a space in it, the part before that is considered the link text. If it does not contain a space, the word [link] is used.

Note that the idea to separate the link and the link text  with a space needed because the original specification does not contain a link text, just the link. I am not seeing a downside as long as spaces in the real link are encoded with %20.

This is may be a solution for #6236

### DIFF
--- a/src/gui/activitydata.h
+++ b/src/gui/activitydata.h
@@ -57,6 +57,7 @@ public:
     QString _message;
     QString _file;
     QUrl _link;
+    QString _linkText;
     QDateTime _dateTime;
     QString _accName;
 
--- a/src/gui/activitydata.h
+++ b/src/gui/activitydata.h
@@ -70,6 +70,23 @@ public:
 
 
     Identifier ident() const;
+
+    void extractLink( const QString& s )
+    {
+        if (!s.isEmpty()) {
+            // if there is a real whitespace in the link, the part before the space
+            // is rendered as a link text
+            int charPos = s.lastIndexOf(QChar(' '));
+            if( charPos > -1 ) {
+                const QString link = s.mid(charPos+1); // all chars from the char to back
+                _link = QUrl::fromEncoded(link.toUtf8());
+                _linkText = s.mid(0, charPos).toHtmlEscaped();
+            } else {
+                _link = QUrl(s);
+                _linkText.clear();
+            }
+        }
+    }
 };
 
 bool operator==(const Activity &rhs, const Activity &lhs);
--- a/src/gui/activitydata.h
+++ b/src/gui/activitydata.h
@@ -56,12 +56,13 @@ public:
     QString _subject;
     QString _message;
     QString _file;
-    QUrl _link;
-    QString _linkText;
+    QUrl _link;        /* a link that might be passed as part of the activity */
+    QString _linkText; /* unescaped label of the link above */
     QDateTime _dateTime;
-    QString _accName;
+    QString _accName;  /* display name of the account involved */
 
-    QVector<ActivityLink> _links;
+    QVector<ActivityLink> _links; /* These links are transformed into buttons that
+                                   * call links as reactions on the activity */
     /**
      * @brief Sort operator to sort the list youngest first.
      * @param val
@@ -77,10 +78,10 @@ public:
             // if there is a real whitespace in the link, the part before the space
             // is rendered as a link text
             int charPos = s.lastIndexOf(QChar(' '));
-            if( charPos > -1 ) {
+            if (charPos > -1) {
                 const QString link = s.mid(charPos+1); // all chars from the char to back
                 _link = QUrl::fromEncoded(link.toUtf8());
-                _linkText = s.mid(0, charPos).toHtmlEscaped();
+                _linkText = s.mid(0, charPos);
             } else {
                 _link = QUrl(s);
                 _linkText.clear();
--- a/src/gui/notificationwidget.cpp
+++ b/src/gui/notificationwidget.cpp
@@ -44,6 +44,16 @@ void NotificationWidget::setActivity(const Activity &activity)
     _ui._subjectLabel->setVisible(!activity._subject.isEmpty());
     _ui._messageLabel->setVisible(!activity._message.isEmpty());
 
+    QString subject = activity._subject;
+    if( !activity._link.isEmpty() ) {
+        // append a link to the message, if that is empty, to subject
+        subject.append(QString("&nbsp;<a href=\"%1\">%2</a>")
+                       .arg(activity._link.toString(QUrl::FullyEncoded))
+                       .arg(activity._linkText));
+        _ui._subjectLabel->setTextFormat(Qt::RichText);
+        _ui._subjectLabel->setOpenExternalLinks(true);
+
+    }
     _ui._subjectLabel->setText(activity._subject);
     _ui._messageLabel->setText(activity._message);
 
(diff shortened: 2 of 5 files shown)


dragotin:
DISCLAIMER: I did not have a ownCloud available that can do notifications, so this is PR is untested. Thus label "DO NOT MERGE YET" with the kind request to review/comment and maybe test.

src/gui/servernotificationhandler.cpp
```diff
@@ -81,7 +81,16 @@ void ServerNotificationHandler::slotNotificationsReceived(const QJsonDocument &j
         a._message = json.value("message").toString();
         QString s = json.value("link").toString();
         if (!s.isEmpty()) {
-            a._link = QUrl(s);
+            // if there is a real whitespace in the link, the part before the space
+            // is rendered as a link text
+            if( s.contains(QChar(' '))) {
+                const QStringList li = s.split(QChar(' '));
+                a._link = QUrl::fromEncoded(li.at(1).toLocal8Bit());
```
ogoffart: ToLatin1 or toUtf8.

toLocal8bit is for path on the filesystem, or things that we want to expose in stderr

There can only be one space?

src/gui/notificationwidget.cpp
```diff
@@ -44,6 +44,16 @@ void NotificationWidget::setActivity(const Activity &activity)
     _ui._subjectLabel->setVisible(!activity._subject.isEmpty());
     _ui._messageLabel->setVisible(!activity._message.isEmpty());
 
+    QString subject = activity._subject;
+    if( !activity._link.isEmpty() ) {
+        // append a link to the message, if that is empty, to subject
+        subject.append(QString("&nbsp;<a href=\"%1\">%2</a>")
+                       .arg(activity._link.toString(QUrl::FullyEncoded))
```
ogoffart: use arg(... , ...)  with two arguments.   In general this is better because it works even if the first string contains a %1 or a %2.

Also, please use escapeXml  to the _linkText

src/gui/servernotificationhandler.cpp
```diff
@@ -81,7 +81,16 @@ void ServerNotificationHandler::slotNotificationsReceived(const QJsonDocument &j
...
+                a._linkText = li.at(0);
+            } else {
+                a._link = QUrl(s);
+                a._linkText = QString("[%1]").arg(tr("link"));
```
ogoffart: (nitpick:  i'd use tr("[link]"))

src/gui/notificationwidget.cpp
```diff
@@ -44,6 +44,16 @@ void NotificationWidget::setActivity(const Activity &activity)
     _ui._subjectLabel->setVisible(!activity._subject.isEmpty());
     _ui._messageLabel->setVisible(!activity._message.isEmpty());
 
+    QString subject = activity._subject;
```
ogoffart: needs to be xml escaped.

ogoffart:
Why not a new button next to "Close" to open the link?

guruz:
@dragotin nice PR! When do you want to continue on it?

> Why not a new button next to "Close" to open the link?

If there is a long link text, this might look odd. But if you just want to have a button "Open" then I think it's fine..

dragotin:
Fixed all the outstanding things (I hope) and added a unit test for splitting of the url from the link text. Thanks @ogoffart for review.

I do not like the idea of a button, because a link should not be a button, but a link. Also please note that it is already possible to add additional buttons that trigger a link via the notification specification, see here: https://github.com/owncloud/client/blob/master/src/gui/notificationwidget.cpp#L72

src/gui/activitydata.h
```diff
@@ -57,6 +57,7 @@ class Activity
     QString _message;
     QString _file;
     QUrl _link;
+    QString _linkText;
```
ogoffart: Maybe document which field are html escaped.

Personally, i'd keep all these string unescaped.  and escape them right before putting them in some html text.

src/gui/activitydata.h
```diff
@@ -69,6 +70,23 @@ class Activity
...
 
     Identifier ident() const;
+
+    void extractLink( const QString& s )
+    {
+        if (!s.isEmpty()) {
+            // if there is a real whitespace in the link, the part before the space
+            // is rendered as a link text
+            int charPos = s.lastIndexOf(QChar(' '));
+            if( charPos > -1 ) {
```
ogoffart: (style:  we don't put spaces inside the parentheses, the space is before.  You could run git-clang-format, to fix all style issues.)

src/gui/notificationwidget.cpp
```diff
@@ -44,6 +44,20 @@ void NotificationWidget::setActivity(const Activity &activity)
...
+    QString subject = activity._subject.toHtmlEscaped();
+    if( !activity._link.isEmpty() ) {
+        // append a link to the message, if that is empty, to subject
+        QString lText = activity._linkText;
+        if( lText.isEmpty() ) {
+            lText = tr("[link]");
+        }
+        subject.append( QString("&nbsp;<a href=\"%1\">%2</a>")
+                        .arg(activity._link.toString(QUrl::FullyEncoded),
+                             lText.toHtmlEscaped() ));
```
ogoffart: Am I wrong that you double html escape?

test/testactivitydata.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+class TestActivityData: public QObject
+{
+    Q_OBJECT
+
+private:
+  
+private slots:
+  
+    void testLinkSplitEmpty() {
```
ogoffart: this kind of test would be much shorter using a testLink_data()  thing

src/gui/notificationwidget.cpp
```diff
@@ -44,6 +44,20 @@ void NotificationWidget::setActivity(const Activity &activity)
...
+        _ui._subjectLabel->setTextFormat(Qt::RichText);
+        _ui._subjectLabel->setOpenExternalLinks(true);
+
+    }
     _ui._subjectLabel->setText(activity._subject);
```
ogoffart: you mean just "subject" here!

dragotin:
Thanks @ogoffart, I fixed stuff accordingly.

With the test I would need more advise ;-)

test/testactivitydata.cpp
ogoffart: _data tests, like for example TestFolder::testFolder or TestUtility::testSanitizeForFileName_data  or ...


And also QCOMPARE instead of QVERIFY

ogoffart:
I wanted to remove my "Request change" without "Approving" but apparently this is not possible.

dragotin:
Should be good now.

guruz:
Works for me, but let's wait if the server guys are OK with the format. I'll ping them.

screen shot 2018-04-09 at 17 28 27

We can make it more beautiful later. Maybe the link should be in a newline?

src/gui/activitydata.h
```diff
@@ -69,6 +71,23 @@ class Activity
 
 
     Identifier ident() const;
+
+    void extractLink( const QString& s )
+    {
+        if (!s.isEmpty()) {
+            // if there is a real whitespace in the link, the part before the space
+            // is rendered as a link text
+            int charPos = s.lastIndexOf(QChar(' '));
```
DeepDiver1975: That's fishy ....

guruz:
Superseded by https://github.com/owncloud/client/pull/6453 after discussion on server issue about link text. Sorry @dragotin

PR #6243, opened 2017-12-14
Unify item type enum

ckamm:
Previously, there was csync_ftw_type_e and SyncFileItem::Type. Having
two enums lead to a bug where Type::Unknown == Type::File that went
unnoticed for a good while.

This patch keeps only a single enum.


ogoffart:
The CI show some rror on windows:

```
GOZDZ72KPI2HPVOFEJODKJCIW4E3QXZUVNVL3VSE2HOIWUPWMXCA/src/csync/vio/csync_vio_local_win.cpp:175:29: error: 'CSYNC_FTW_TYPE_SLINK' was not declared in this scope
           file_stat->type = CSYNC_FTW_TYPE_SLINK;
                             ^~~~~~~~~~~~~~~~~~~~
```

ogoffart:
The CI failure is probably caused by the change in the order of the include.

Just add `#include<functional>` in csync_excludes.h

ogoffart:
:+1:

PR #6244, opened 2017-12-14
Modernize makefiles and remove dead code

ogoffart:
Simplify a the CMakeLists.txt, and remove some of unused part of csync
See individual commit messages for more info


ogoffart:
> +146 −2328 
:-)

ogoffart:
Meh, the CI system is still on cmake 3.0.2 which does not support imported targets :-(
@DeepDiver1975 Any chance we can get cmake 3.1 ?

src/csync/CMakeLists.txt
```diff
@@ -89,52 +59,48 @@ else()
...
+find_package(SQLite3 3.8.0 REQUIRED)
 if (USE_OUR_OWN_SQLITE3)
-    list(APPEND csync_SRCS ${SQLITE3_SOURCE})
-    if (WIN32)
-        # We want to export sqlite symbols from the ocsync DLL without
-        # having to patch both sqlite3.h and the amalgation sqlite3.c,
-        # so do the import/export magic manually through the build system.
-        remove_definitions(-DSQLITE_API=__declspec\(dllimport\))
-        add_definitions(-DSQLITE_API=__declspec\(dllexport\))
-    endif()
```
dschmidt: Why isn't this needed anymore?

dschmidt:
We (@jnweiger and I) already planned to update package building machines on all platforms very soon, because I want/need to get rid of the old AppIconMacro. So I support the request for newer CMake on CI :-)

src/csync/CMakeLists.txt
ogoffart: Because sqlite is statically built into csync.

Before it was also used from the syncjournaldb, but the dependence was inverted by some refactor before 2.4 and the syncjournaldb is now used by csync.

So now there is only csync that uses sqlite3 symbols, and we don't need this trick to get it work with libowncloudsync.  

This is not because of this patch

dschmidt: Okay, thanks for the explanation!

ogoffart:
Rebased since there was conflicts.

dschmidt:
Rebased so we can have the CI passing...

dschmidt:
retest this please

admin/linux/Dockerfile
```diff
@@ -0,0 +1,19 @@
+# Distro with Qt 5.6
```
ogoffart: There should be a README comment stating what this docker image is for.

(Also a license?)

admin/win/Toolchain-mingw32-openSUSE.cmake
```diff
@@ -27,8 +27,8 @@ SET(QT_MKSPECS_DIR          ${CMAKE_FIND_ROOT_PATH}/share/qt5/mkspecs)
...
 # qt tools
-SET(QT_QMAKE_EXECUTABLE     ${MINGW_PREFIX}-qmake )
-SET(QT_MOC_EXECUTABLE       ${MINGW_PREFIX}-moc)
-SET(QT_RCC_EXECUTABLE       ${MINGW_PREFIX}-rcc)
-SET(QT_UIC_EXECUTABLE       ${MINGW_PREFIX}-uic)
-SET(QT_LRELEASE_EXECUTABLE  ${MINGW_PREFIX}-lrelease)
+SET(QT_QMAKE_EXECUTABLE         ${MINGW_PREFIX}-qmake-qt5)
+SET(QT_MOC_EXECUTABLE           ${MINGW_PREFIX}-moc-qt5)
+SET(QT_RCC_EXECUTABLE           ${MINGW_PREFIX}-rcc-qt5)
+SET(Qt5Widgets_UIC_EXECUTABLE   ${MINGW_PREFIX}-uic-qt5)
```
ogoffart: I don't understand this change.
Why is it needed and why not for qmake, moc,rcc,lrelease ?

Issue #6160, opened 2017-11-14
Release 2.4.0


SamuAlfageme:
Since most of the milestone is ReadyToTest and the rest is packaging/requests for info; we can start thinking about the upcoming release. I'll also start working in with https://github.com/owncloud/client/wiki/Testing-Scenarios-2.4 asap

Some weeks before the release:
* [x] Check if we should update the bundled sqlite3 (https://github.com/owncloud/client/tree/master/src/3rdparty/sqlite3)
* [x] Check if we should update Sparkle on build machine (https://github.com/sparkle-project/Sparkle/releases)
* [x] Ensure NSIS is up to date on the build machine
* [x] Ensure up-to-date dependencies (e.g. [latest Qt version](http://qt-project.org/downloads#qt-lib) is installed on the machine and picked up (cmake output)
* [x] Ensure the crash reporter server is up
* [x] Check crash reporter for bad crashes
* [x] Ensure Windows Overlay DLLs are rebuilt
* [x] Check nightly builds are up and running, that is Jenkins jobs ownCloud-client-linux, ownCloud-client-osx and ownCloud-client-win32 all green.
* [x] Ensure Linux nightlies are built too for all distros https://build.opensuse.org/package/show/isv:ownCloud:community:nightly/owncloud-client
* [ ] Build branded clients through the scripting machine and smoke test one or two branded clients (especially with predefined url **and OAuth client ids/secrets**)
* [ ] Upload a nightly build of the windows version to virustotal.com
  * Contact AV vendors whom's engine reports a virus
* [x] Documentation should be online before the release http://doc.owncloud.org/desktop/1.X/
* [ ] QA goes over https://github.com/owncloud/mirall/wiki/Testing-Scenarios
* [x] Make sure to have `client/ChangeLog` updated
 * use `git log --format=oneline v<lastrelease>...master` if your memory fails you
* [ ] check if enterprise issues are fixed

One week before the release:
* [ ] Communicate the release schedule on mailinglist release-coordination@owncloud.com. Give a high level overview of the upcoming new features, changes etc.
* [ ] Ensure marketing is aware (marketing@owncloud.com) and prepared for the release (social, .com website, cust. communications)
* [ ] Inform GCX knows the next version is about 1 week out (gcx@owncloud.com)

beta1:
* [x] Branch off a release branch called VERSION-rcX or VERSION-betaX  (without v, v is for tags)
[...]

PR #6260, opened 2017-12-20
2.4.0 Website changelog html

guruz:
@SamuAlfageme Can you do a smoke test on Windows? My VM is broken...

As discussed with @jnweiger , for now we have this file with the clickable HTML and the download links in this repo. Then the new website can embed it.
@jnweiger Please review and especially check the download links
(you still need to move from testing to stable, also in OBS)
(Also backup the old stable OBS build 2.3.4 to a named OBS project)
See https://github.com/owncloud/client/issues/6160

### DIFF
--- a/admin/website/changelog-desktop.html
+++ b/admin/website/changelog-desktop.html
@@ -4,7 +4,7 @@
 
 <!-- When doing beta/rc vs release, always check the subdirectory name and the OBS repo name!! //-->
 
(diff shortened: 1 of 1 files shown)


guruz:
By @SamuAlfageme 's suggestion, we have decided to use https://github.com/owncloud/client/releases instead. :)

PR #6265, opened 2017-12-25
Move to trash

Mc-Pain:
With this patch client can move deleted files to trash on UNIX systems in case of deleting them by server. (sometimes server-side trash does not work and that causes data losing)

### DIFF
--- a/src/common/filesystembase.cpp
+++ b/src/common/filesystembase.cpp
@@ -459,7 +459,7 @@ bool FileSystem::moveToTrashFunc(const QString &fileName)
         while (file.exists(path+QString(suffix_number))){ //or to "filename.2" if "filename.1" exists, etc
             suffix_number++;
         }
-        if (!file.rename(f.absoluteFilePath(),path+QString(suffix_number))){  // rename(file old path, file trash path)
+        if (!file.rename(f.absoluteFilePath(),path+QString::number(suffix_number))){  // rename(file old path, file trash path)
             return false;
         }
     } else {
@@ -471,7 +471,7 @@ bool FileSystem::moveToTrashFunc(const QString &fileName)
     // create file format for trash info file----- START
     QFile infoFile;
     if (file.exists(trashInfoPath+f.fileName())) { //TrashInfo file already exists, create "filename.1.trashinfo"
-        QString filename = trashInfoPath+f.fileName()+"."+QString(suffix_number)+".trashinfo";
+        QString filename = trashInfoPath+f.fileName()+"."+QString::number(suffix_number)+".trashinfo";
         infoFile.setFileName(filename);     //filename+.trashinfo //  create file information file in /.local/share/Trash/info/ folder
     } else {
         QString filename = trashInfoPath+f.fileName()+".trashinfo";
--- a/src/common/filesystembase.cpp
+++ b/src/common/filesystembase.cpp
@@ -416,8 +416,6 @@ bool FileSystem::remove(const QString &fileName, QString *errorString, bool move
 #endif
 #ifdef Q_OS_UNIX
     if(moveToTrash){
-        Q_UNUSED(errorString);
-        qDebug() << "moving" << fileName << "to trash";
         if (!moveToTrashFunc(fileName)) {
             return false;
         }
--- a/src/common/filesystembase.cpp
+++ b/src/common/filesystembase.cpp
@@ -19,6 +19,8 @@
 #include "filesystembase.h"
 
 #include <QDateTime>
+#include <QDir>
+#include <QUrl>
 #include <QFile>
 #include <QCryptographicHash>
 
@@ -403,7 +405,7 @@ QByteArray FileSystem::calcAdler32(const QString &filename)
 }
 #endif
 
-bool FileSystem::remove(const QString &fileName, QString *errorString)
+bool FileSystem::remove(const QString &fileName, QString *errorString, bool _moveToTrash)
 {
 #ifdef Q_OS_WIN
     // You cannot delete a read-only file on windows, but we want to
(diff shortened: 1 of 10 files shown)


src/common/filesystembase.cpp
```diff
@@ -433,15 +435,75 @@ bool FileSystem::remove(const QString &fileName, QString *errorString)
...
     }
 #endif
-    QFile f(fileName);
-    if (!f.remove()) {
-        if (errorString) {
-            *errorString = f.errorString();
+#ifdef Q_OS_UNIX
+    if(_moveToTrash){
+        Q_UNUSED(errorString);
+        qDebug() << "moving" << fileName << "to trash";
```
ogoffart: Please add a category.

src/common/filesystembase.cpp
```diff
@@ -433,15 +435,75 @@ bool FileSystem::remove(const QString &fileName, QString *errorString)
...
+        if (!moveToTrash(fileName)) {
+            return false;
+        }
+    }
+    if(!_moveToTrash) {
```
ogoffart: Please use else, and put the brace outside the #if so that you don't need another #if for the closing brace

```
    } else
#endif
    {
```

src/libsync/propagatorjobs.cpp
```diff
@@ -13,6 +13,7 @@
  * for more details.
  */
 
+#include "configfile.h"
```
ogoffart: Please don't add a dependency between the config file and the sync engine!
Add the option in the `SyncOption`  struct

translations/client_ru.ts
```diff
@@ -1244,6 +1244,11 @@ Continuing the sync as normal will cause all your files to be overwritten by an
         <source>Ask for confirmation before synchronizing external storages</source>
         <translation>Спрашивать подтверждение перед синхронизацией внешних хранилищ</translation>
     </message>
+    <message>
```
ogoffart: (No need to add translation manually.  It should be done via transifex)

src/common/filesystembase.cpp
```diff
@@ -433,15 +435,75 @@ bool FileSystem::remove(const QString &fileName, QString *errorString)
...
+    info += '\n';
+
+    stream << info;
+
+    infoFile.close();
+
+    // create info file format of trash file----- END
+
+    QDir file;
+    if (!file.rename(f.absoluteFilePath(),trashFilePath+f.fileName())){  // rename(file old path, file trash path)
```
ogoffart: should you not check if there is already a file with this name in the trash and add some suffix?

ogoffart: Also if it failes, i suppose we should also remove the info file

src/common/filesystembase.h
```diff
@@ -107,8 +107,11 @@ namespace FileSystem {
      * Equivalent to QFile::remove(), except on Windows, where it will also
      * successfully remove read-only files.
      */
-    bool OCSYNC_EXPORT remove(const QString &fileName, QString *errorString = 0);
+    bool OCSYNC_EXPORT remove(const QString &fileName, QString *errorString = 0, bool _moveToTrash = false);
```
ogoffart: coding style:  '_' at the beginning is indicating a member. So there should not be an _ there.

src/gui/generalsettings.cpp
```diff
@@ -68,6 +68,7 @@ GeneralSettings::GeneralSettings(QWidget *parent)
     connect(_ui->newFolderLimitCheckBox, &QAbstractButton::toggled, this, &GeneralSettings::saveMiscSettings);
     connect(_ui->newFolderLimitSpinBox, static_cast<void (QSpinBox::*)(int)>(&QSpinBox::valueChanged), this, &GeneralSettings::saveMiscSettings);
     connect(_ui->newExternalStorage, &QAbstractButton::toggled, this, &GeneralSettings::saveMiscSettings);
+    connect(_ui->moveToTrashCheckBox, &QAbstractButton::toggled, this, &GeneralSettings::saveMiscSettings);
```
ogoffart: also, the chekbox should be hidden on non-unix is the move to trash function is not implemented

michaelstingl:
Where can I test-drive this? How does this look like? Screenshot? Addition for docs?

src/common/filesystembase.cpp
Mc-Pain: What do you mean?

src/common/filesystembase.cpp
Mc-Pain: Suffix like somefile.1, somefile.2, etc?

src/gui/generalsettings.cpp
SamuAlfageme: Isn't [`Q_OS_UNIX`](https://doc-snapshots.qt.io/qt5-dev/qtglobal.html#Q_OS_UNIX) defined for macOS as well? I mean macOS is also unix-based.

SamuAlfageme:
@Mc-Pain thanks a lot for your contribution! ☺️

@michaelstingl basically, this moves any deletion coming from the server to `$HOME/.local/share/Trash/files` if a new setting checkbox (on the "General > Advanced" section) is active. The feature/checkbox are only present in Linux (would be pretty straightforward to port it to other OSes I guess)

michaelstingl:
Just my 2ct:

- I personally wouldn't expect my trash is filling up because in some of my shares someone is deleting something
- I really don't want to pollute the GUI with very special advanced settings for specific use cases. @SamuAlfageme any idea how to deal with advanced-advanced settings? (environment variable or config file option would be okay for the first release)
- What about multiple accounts? Create subfolders in trash? (`user@cloud.mydomain.com_trash`)
- Linux-only also would be okay for the first release

Mc-Pain:
> I personally wouldn't expect my trash is filling up because in some of my shares someone is deleting something
t won't happen: moving to trash is disabled by default

Mc-Pain:
@ogoffart please take a look at the changes you requested

src/common/filesystembase.h
```diff
@@ -107,8 +107,11 @@ namespace FileSystem {
      * Equivalent to QFile::remove(), except on Windows, where it will also
      * successfully remove read-only files.
      */
-    bool OCSYNC_EXPORT remove(const QString &fileName, QString *errorString = 0);
+    bool OCSYNC_EXPORT remove(const QString &fileName, QString *errorString = 0, bool moveToTrash = false);
```
ogoffart: I don't think the moveToTrash feature belong in the FileSystem::remove function.
Since we anyway call it only once moveToTrash=true, we can as well call "moveToTrash" instead.

src/common/filesystembase.h
```diff
@@ -107,8 +107,11 @@ namespace FileSystem {
...
 
+#ifdef Q_OS_UNIX
+    bool moveToTrashFunc(const QString &filename);
```
ogoffart: export it and rename it `moveToTrash`   (what's the `Func` for?)

src/libsync/propagatorjobs.cpp
```diff
@@ -54,6 +55,10 @@ QByteArray localFileIdFromFullId(const QByteArray &id)
  */
 bool PropagateLocalRemove::removeRecursively(const QString &path)
 {
+#ifdef Q_OS_UNIX
+    _moveToTrash = propagator()->syncOptions()._moveFilesToTrash;
```
ogoffart: Since removeToTrash is already recursive, you should not call this function when moving to trash.
It is already complicated enough.  But all the complexity comes from the fact we might have individual faillure for individual files, which can't happen here.

Don't use this code path when moving to trash.

src/libsync/propagatorjobs.cpp
```diff
@@ -128,8 +150,15 @@ void PropagateLocalRemove::start()
         }
```
ogoffart: Should simply be:

```
if (_moveToTrash) {
    if (!FileSystem::moveToTrash(absolute)) {
        // handle error;
     }
} else {
   // normal remove. 
}
```

src/libsync/propagatorjobs.cpp
```diff
@@ -121,17 +126,38 @@ void PropagateLocalRemove::start()
...
+            }
         }
     } else {
-        QString removeError;
-        if (FileSystem::fileExists(filename)
-            && !FileSystem::remove(filename, &removeError)) {
-            done(SyncFileItem::NormalError, removeError);
-            return;
+#ifdef Q_OS_UNIX
+        if (_moveToTrash) {
```
ogoffart: Note: this code path is the same as for the directory one  (QDir(filename).exists()  can be replaced by FileSystem::fileExists.  So If you do the  `if (_moveToTrash) ` check  before the isDirectory check, you can greatly simplify the code

Mc-Pain:
I can't find any compilation errors in the logs

src/libsync/propagatorjobs.cpp
```diff
@@ -121,18 +126,29 @@ void PropagateLocalRemove::start()
...
-            return;
-        }
-    } else {
-        QString removeError;
-        if (FileSystem::fileExists(filename)
-            && !FileSystem::remove(filename, &removeError)) {
+    QString removeError;
+#ifdef Q_OS_UNIX
+    if (_moveToTrash) {
+        if ((QDir(filename).exists() || FileSystem::fileExists(filename))
```
ogoffart: I think FileSystem::fileExists(filename)  should be enough.

ogoffart:
@michaelstingl :  The code looks good. (CI failure unrelated)    Should we then just merge it?  Or should we remove the checkbox and make it an hidden setting?

michaelstingl:
Please start with a hidden setting for now. @SamuAlfageme has a plan how to deal with advanced-advanced settings in general…

ogoffart:
I squashed the changes, and pushed to master. Also adjusted a bit in commit de929eb

ogoffart:
Well, what's remaining to do is to documment this.

@SamuAlfageme  What's tour plan?

I guess it should be documented in the file `doc/conffile.rst` 

@Mc-Pain do you want to do a new pull request with the documentation?

Mc-Pain:
@ogoffart
I already did: #6336

jnweiger:
@Mc-Pain Nice optional feature! It catches remote deletes,that might be out if my control!

Testing with `moveToTrash=true` in my cfg on linux mint tara.

What is to happen, if the user deselects a huge folder in the client's selective sync settings?
The client warns and deletes the folder locally, that much is fine. 

But should this folder be added to my Trash? It was not a remote delete -- it was me intentionally telling the client I don't want to see this folder. 

I'd say It should not populate my Desktop trash can. Currently it does.

PR #6271, opened 2017-12-29
macOS postinstall: check if Finder is running

JayBrown:
Some users have disabled Finder on their systems; the old postinstall script would always reenable Finder; therefore the quick check for Finder status. The postinstall script will only run the osascript reveal command, if Finder is enabled/running.

### DIFF
--- a/admin/osx/post_install.sh.cmake
+++ b/admin/osx/post_install.sh.cmake
@@ -1,12 +1,16 @@
 #!/bin/sh
 
-osascript << EOF
+# Check if Finder is running (for systems with Finder disabled)
+finder_status=`ps aux | grep "/System/Library/CoreServices/Finder.app/Contents/MacOS/Finder" | grep -v "grep"`
+if ! [ "$finder_status" == "" ] ; then # Finder is running
+   osascript << EOF
 tell application "Finder"
    activate
    select the last Finder window
 	reveal POSIX file "/Applications/@APPLICATION_EXECUTABLE@.app"
 end tell
 EOF
+fi
 
 # Always enable the new 10.10 finder plugin if available
 if [ -x "$(command -v pluginkit)" ]; then


guruz:
Interesting, I'll test and merge this.
Could you explain how/why would one disable Finder?

JayBrown:
Yes, **please test it first**. It's the first time I ever did something in `sh`. (Only used to `bash` & `zsh`.) On my end it worked, but only directly in the shell, not as a postinstall script running from a pkg.

As for Finder, people think it's "the system" (macOS), but it's basically just an application that you can disable or quit in favor of (arguably better) file managers like Nimble Commander and fman. (My experiences with Finder are subpar, to be diplomatic.) There are three methods:

(1) **Disable Finder completely:** disable SIP first, then unload `/System/Library/LaunchAgents/com.apple.Finder.plist` with `sudo launchctl unload -w /System/Library/LaunchAgents/com.apple.Finder.plist` and reboot. (I haven't tested this method, but it's preferred by the real Finder haters out there.)

(2) **Finder switcheroo:** `defaults write com.apple.loginwindow Finder /Applications/AlternateApplication.app` > apparently Finder then doesn't launch, but registers as just a normal application, i.e. it can be launched & quit like e.g. TextEdit. (I haven't tested this method.)

(3) **Adding a "Quit" menu command:** related method to method 2; just run once: `defaults write com.apple.finder QuitMenuItem -bool YES && killall Finder`. Then you have Finder with "Quit" in the "File" menu, and you can quit it manually, or you can (as I do) run a login script that launches my alternate file manager and upon success quits Finder with `osascript -e 'tell application "Finder" to quit'`

Method 3 is my preferred method, because it's the least intrusive, and it shows me, when other applications activate the Finder. Some of them actually seem to need Finder for certain things, so it's probably not a good idea to disable Finder complete (method 1).

PR #6284, opened 2018-01-04
Some more CMake fixes/improvements/enhancements

dschmidt:
This escalated a little - I'm sorry about that....

If you want me to split this into several pull requests, let me know.. otherwise it's probably a good idea to review this commit per commit.


dschmidt:
https://github.com/owncloud/client/pull/6284/commits/5d0d72417cd27de0b216816578fdcf3979c52a50
certainly too :) 

I'd still like very much if we could get all of our PRs into master soon.

Issue #6245, opened 2017-12-14
Regex excludes cause existing patterns to behave differently


ckamm:
From @sagamusix #5017 
> for example my pattern */.svn/pristine/ no longer seems to work

Summary from investigation:

The client used `fnmatch` with `FNM_PATHNAME` on unix and `PathMatchSpec` on Windows. These had different behavior when expanding `*` patterns: `fnmatch` never expands it to `/`, while that is allowed on windows. That means `*/foo/bar` would match `a/b/foo/bar` on windows, but not on other oses.

The new regex exclude matching code implements the `fnmatch` behavior on all platforms thereby changing the meaning of existing patterns on windows.

ckamm:
@sagamusix So it's now syncing some folder foo/.svn/pristine and its contents?

sagamusix:
Exactly, and previously these files were ignored as intended.

ckamm:
@sagamusix It works for me. The pattern gets translated to `[^/]*/\.svn/pristine)(?:$|/)`. Can you let me know what path exactly was excluded before and isn't anymore?

sagamusix:
Two paths were matched before I realized and aborted the sync; one was `Uni\powerwall-trunk\.svn\pristine` and the other one was similar in structure. It seems like if I move the folder one level up, it works as intended, which coincides with your idea, I think.

sagamusix:
I have changed the pattern to just `.svn/pristine` (including all other patterns in my list that were structured similarly) and it seems like the initial discovery is much, much slower now - it's now running for almost an hour for a folder that previously took a few minutes. In fact, it's still discovering folders so I cannot see yet if it helps fixing the problem.

ckamm:
@sagamusix changing too `.svn/pristine/` still wouldn't trigger the right behavior. See https://github.com/owncloud/client/issues/5275 for discussion of the unclear behavior.

ckamm:
Yes, unfortunately. But I don't think it's acceptable to break patterns windows users had added. I think we'll need to put in extra work to make sure that migrates nicely.

And that performance issue you're seeing also definitely needs to be investigated.

ckamm:
I'm looking at allowing `*` to match `.*` in the regex: it complicates the bname-as-activation optimization though since `foo/bar*daa` could very well be activated by a bname `xyzdaa`. It may be ok to be just mostly correct here.

ckamm:
[...]

PR #6286, opened 2018-01-05
Exclude regex: Restore old matching on Windows

ckamm:
Unfortunately matching behaved differently on Windows. This patch
restores the previous matching behavior but still uses the new regular
expression based matching.

Further work will hopefully unify the behavior between platforms without
breaking backwards compatibility.

See  #6245

### DIFF
--- a/src/csync/csync_exclude.cpp
+++ b/src/csync/csync_exclude.cpp
@@ -241,6 +241,8 @@ using namespace OCC;
 
 ExcludedFiles::ExcludedFiles()
 {
+    // Windows used to use PathMatchSpec which allows *foo to match abc/deffoo.
+    _wildcardsMatchSlash = Utility::isWindows();
 }
 
 ExcludedFiles::~ExcludedFiles()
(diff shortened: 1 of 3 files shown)


src/csync/csync_exclude.cpp
```diff
@@ -409,7 +417,12 @@ auto ExcludedFiles::csyncTraversalMatchFun() const
     return [this](const char *path, ItemType filetype) { return this->traversalPatternMatch(path, filetype); };
 }
 
-static QString convertToRegexpSyntax(QString exclude)
+/**
+ * On linux we used to use fnmatch with FNM_PATHNAME, but the windows function we used
+ * didn't have that behavior. wildcardsMatchSlash can be used to control which behavior
+ * the resulting regex shall use.
+ */
+static QString convertToRegexpSyntax(QString exclude, bool wildcardsMatchSlash)
```
ogoffart: Not related to this patch: `const QString &`

src/csync/csync_exclude.cpp
```diff
@@ -491,6 +512,42 @@ static QString convertToRegexpSyntax(QString exclude)
     return regex;
 }
 
+static QString extractBnameTrigger(const QString &exclude, bool wildcardsMatchSlash)
+{
+    // We can definitely drop everything to the left of a / - that will never match
+    // any bname.
+    QString pattern = exclude.mid(exclude.lastIndexOf('/') + 1);
```
ogoffart: I know it might be premature optimisations....  but:

`midRef`    (since you do another mid later,  and you can do   `return ("*" + pattern);` at the end so you'd save some allocations.

Issue #4341, opened 2016-01-08
Virtual Filesystem (VFS)


MTRichards:
This is a research task, to look into and define what it would take to make this happen. Basic user story looks like this:

As a user, I want to see all of my files on my desktop, regardless of if they have synced or not so that I can access my files at any time, from anywhere.

Acceptance Criteria:
- A threshold can be set for the "cache" that is allowed on the desktop
- Favorites are synced automatically with the client.
- The most recently accessed files are synced until the cache is full, then files are removed from the local device.
- If favorites are bigger than the cache, only the recently accessed favorites are kept

And much more.

@karlitschek @dragotin @danimo for your continued discussion and additional requirements.

jturcotte:
Since the source problem motivating this solution is that managing very large shares is cumbersome with the client, here are some specific problems/solution that could be incrementally improving the quality of the sync client to also solve the issue instead of a risky and costly file system solution:

### I'm subscribed to a large shared folder, but from time to time huge files that I don't need appear on my disk
Do shared folders need to be synced by default? Your own files, or files that you favorited in the web interface should be included in the selective sync by default. Other files could simply not be synced until you ask for them. The folder was shared with you, congratulations you have access, but in most cases you might not want them. A notification when you gain access could already ask you if you want to sync the new shared folder in totality.

### I want to work on my own files in the company shared folder, but activating it in the selective sync now means that I get every new files included and I need to keep deselecting them
The selective sync UI currently only allows excluding files. If you have the structure a/b/c, and you want the directory a/b/c/d1, but not a/b/c/d2, the UI will allow you to deselect d2, but if d3 appears it will be synced by default. There is no way to say that you *just* want a/b/c/d1, you can only say I don't want all other files, and you need to update this list every time a new folder appear.

We could provide a way to include and exclude folders, so that you can exclude a/b/c, but override this only for its a/b/c/d1 child, thus also excluding d3 by default when it appears. The control panel for the indexing service on Windows is an example of a UI that allows this. It's a bit more complicated, but allows to have both an include and exclude hierarchy.

For example:
![Windows indexing settings](http://cdn3.tekrevue.com/wp-content/uploads/2014/05/windows-search-index-locations.png)

### The selective sync UI is slow and it shows arrows to expand folders even if they don't have children
We currently PROPFIND the server each time you click in the UI, we could keep the whole file structure and store it in the database, updating it on each sync so that the UI is fast and complete.

[...]

PR #6290, opened 2018-01-09
On-demand downloading: Placeholder-file based prototype

ckamm:
This is an initial prototype for allowing the sync to skip downloading files. It's PRed for ease of initial feedback. We want to investigate whether this is a useful direction for #4341 

TODOs
* [x] register the placeholder extension for double click on windows
* [ ] register the placeholder extension for double click on osx
* [x] migrate OSX to the new shell extension api (#6328)
* [x] make placeholder extension configurable
* [x] add shell integration action for downloading placeholders
* [x] make option that enables the functionality available (config file editing only for starters?)
* [x] carefully test migration to older client versions
* [x] hide enabling of feature behind a config file switch


src/csync/csync_update.cpp
```diff
@@ -668,6 +709,22 @@ int csync_ftw(CSYNC *ctx, const char *uri, csync_walker_fn fn,
         fullpath = QByteArray() % uri % '/' % filename;
     }
 
+    // When encountering placeholder files, read the relevant
+    // entry from the db instead.
+    if (ctx->current == LOCAL_REPLICA
+            && dirent->type == ItemTypeFile
+            && filename.endsWith(".owncloud")) {
```
ogoffart: We will want to make this name configurable (in the theme)

test/testsyncplaceholders.cpp
```diff
@@ -0,0 +1,222 @@
...
+        QVERIFY(fakeFolder.currentLocalState().find("A/a1.owncloud"));
+        QVERIFY(fakeFolder.currentRemoteState().find("A/a1"));
+        QVERIFY(completeSpy.isEmpty());
+        cleanup();
+
+        // Neither does a remote change
+        fakeFolder.remoteModifier().appendByte("A/a1");
+        QVERIFY(fakeFolder.syncOnce());
+        QVERIFY(!fakeFolder.currentLocalState().find("A/a1"));
+        QVERIFY(fakeFolder.currentLocalState().find("A/a1.owncloud"));
```
ogoffart: maybe also check that it is empty?
And how about the mtime of the placeholder, is it still sync'ed?

test/testsyncplaceholders.cpp
```diff
@@ -0,0 +1,222 @@
...
+            fakeFolder.syncEngine().setLocalDiscoveryOptions(LocalDiscoveryStyle::DatabaseAndFilesystem, { "A" });
+        fakeFolder.localModifier().remove("A/a1.owncloud");
+        QVERIFY(fakeFolder.syncOnce());
+        QVERIFY(!fakeFolder.currentLocalState().find("A/a1"));
+        QVERIFY(fakeFolder.currentLocalState().find("A/a1.owncloud"));
+        QVERIFY(fakeFolder.currentRemoteState().find("A/a1"));
+        QVERIFY(itemInstruction(completeSpy, "A/a1", CSYNC_INSTRUCTION_NEW));
+        QCOMPARE(dbRecord(fakeFolder, "A/a1")._type, ItemTypePlaceholder);
+        QCOMPARE(dbRecord(fakeFolder, "A/a1")._fileSize, 65);
+        cleanup();
```
ogoffart: What if a directory fill of placeholder gets removed?

test/testsyncplaceholders.cpp
```diff
@@ -0,0 +1,222 @@
...
+        fakeFolder.remoteModifier().insert("A/a3");
+        fakeFolder.remoteModifier().insert("A/a4");
+        QVERIFY(fakeFolder.syncOnce());
+        QVERIFY(fakeFolder.currentLocalState().find("A/a1.owncloud"));
+        QVERIFY(fakeFolder.currentLocalState().find("A/a2.owncloud"));
+        QVERIFY(fakeFolder.currentLocalState().find("A/a3.owncloud"));
+        QVERIFY(fakeFolder.currentLocalState().find("A/a4.owncloud"));
+        cleanup();
+
+        // Download by changing the db entry
```
ogoffart: nice test! :+1:

ogoffart:
Overall, I like this prototype, and this is what i had in mind.
There is maybe lots of drawback for this approach, but it is at least quite simple, as this patch shows (even if it is not complete yet)

We should finish the UI and get feedback from some user to see if this is something that's good to have or not.

ogoffart:
I added a few commit, they add a checkbox in the folder wizard, and on linux, install a mime type so the `.owncloud` placeholder files are openned with owncloud.

So now, it can already be tested by creating folder from the "add folder connection" button and clicking that checkbox.  On linux this should work fine from the explorer.  (On windows/mac you still need to manually select to open the file with owncloud)

ckamm:
@ogoffart I resolved and added tests for inserting "foo" locally when there's a placeholder file "foo.owncloud" already.

src/gui/application.cpp
```diff
@@ -448,6 +448,9 @@ void Application::parseOptions(const QStringList &options)
             _debugMode = true;
         } else if (option == QLatin1String("--version")) {
             _versionOnly = true;
+        } else if (option.endsWith(".owncloud")) {
```
ckamm: I'd suggest to put this behind a `--downloadplaceholder` switch

src/gui/application.cpp
```diff
@@ -613,5 +616,29 @@ void Application::showSettingsDialog()
...
+    if (!folder) {
+        qWarning(lcApplication) << "Can't find sync folder for" << filename;
+        // TODO: show a QMessageBox for errors
+        return;
+    }
+    QString relativePath = QDir::cleanPath(normalName).mid(folder->cleanPath().length() + 1);
+    folder->downloadPlaceholder(relativePath);
+    auto con = QSharedPointer<QMetaObject::Connection>::create();
+    *con = QObject::connect(folder, &Folder::syncFinished, [con, normalName] {
+        QObject::disconnect(*con);
```
ckamm: Very cute... but understandable.

src/gui/application.h
```diff
@@ -72,6 +72,10 @@ public slots:
     // TODO: this should not be public
     void slotownCloudWizardDone(int);
     void slotCrash();
+    /**
+     * Will download a placeholder file, and open the result.
```
ckamm: Suggest to add whether this takes a path to a placeholder or a path to the underlying file.

src/csync/csync_update.cpp
dschmidt: Suggestion: simply use the shortname (maybe add a more semantic getter to the theme but please do not introduce a whole new configurable var, we have too many of those anyway)

src/gui/application.cpp
ogoffart: Then it is harder to just associate the .owncloud extension with the program.

src/gui/application.cpp
ogoffart: Yeah, this emulate a "singleshot" connection.

However this needs some work because the file might not be successfully downloaded on the first sync, so i would need to show some feedback in that case.

ckamm:
@ogoffart I've now testing downgrading a placeholder-sync 2.5 to 2.4 and while nothing bad happens anymore it *does* upload the placeholder files. I would have expected them to be deleted locally. So this will need some more investigation.

src/gui/socketapi.cpp
```diff
@@ -573,6 +605,19 @@ void SocketApi::command_GET_MENU_ITEMS(const QString &argument, OCC::SocketListe
             listener->sendMessage(QLatin1String("MENU_ITEM:EMAIL_PRIVATE_LINK::") + tr("Send private link by email..."));
         }
     }
+
+    // Placeholder download action
+    if (syncFolder) {
+        auto placeholderSuffix = QStringLiteral(OWNCLOUD_PLACEHOLDER_SUFFIX);
+        bool hasPlaceholderFile = false;
+        for (const auto &file : files) {
```
ogoffart: i'd use  std::all_of
You want to show this menu if all files are placeholder, right?
This code will work if only one of the files is a placeholder.

ckamm: I thought it'd be nice to show this as long as any one file in the selection was a placeholder. That way you can select a bunch of folder contents and click - don't have to make sure to skip the odd files that aren't placeholders. We could even have this option say "Download 37 files that are currently placeholders".

src/csync/csync_exclude.cpp
```diff
@@ -301,6 +337,32 @@ bool ExcludedFiles::reloadExcludeFiles()
...
+        return true;
+    QByteArrayList args = directive.split(' ');
+    if (args.size() != 3)
+        return true;
+    QByteArray op = args[1];
+    QByteArrayList argVersions = args[2].split('.');
+    if (argVersions.size() != 3)
+        return true;
+
+    auto argVersion = std::forward_as_tuple(argVersions[0].toInt(), argVersions[1].toInt(), argVersions[2].toInt());
```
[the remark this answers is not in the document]
ckamm: Yes, definitely.

src/libsync/syncoptions.h
```diff
@@ -38,7 +38,7 @@ struct SyncOptions
 
     /** Create a placeholder for new files instead of downloading */
     bool _newFilesArePlaceholders = false;
-    QString _placeholderSuffix;
+    QString _placeholderSuffix = ".owncloud";
```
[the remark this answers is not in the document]
ckamm: I like this being an opiton and using the define as little as possible to allow for easier testing if we end up wanting that. I'll add an abort to catch the case of `_placeholderSuffix` being empty when `_newFilesArePlaceholders` is enabled.

Do you think it's essential that this default uses the define? I wonder whether I'll just make a helper function that creates a nice `SyncOptions` for tests and leave it empty in the default constructor.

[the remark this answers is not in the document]
ogoffart: > Do you think it's essential that this default uses the define?

No it is fine if ".owncloud" is used.

ckamm:
@ogoffart What are your thoughts about merging this in disabled state?

The remaining tasks: Having a double-click file extension for windows and osx aren't crucial since one can use the shell integration to download placeholder files.

ogoffart:
I think having the double click is kind of the necessary feature.
It should be easy to do on windows.

But on osx, there is no shell integration yet because https://github.com/owncloud/client/issues/6328 is not done yet

guruz:
Just saw the demo at @ogoffart 's machine, quite nice!

What about a kind-of-modal-on-top progress dialog?

ckamm:
Known issue: If you enable placeholders, then switch back to an older version of the client, the `.owncloud` entries in the db will be ignored - but they will show up in the "ignored files" list as remote ignores, since the file exists locally and in the db.

@ogoffart Think we should add a startup db cleanup step that wipes all metadata entries with unknown type?

guruz:
Why not completely different metadata tables depending on if the placeholder checkbox was checked or not?

> I've now testing downgrading a placeholder-sync 2.5 to 2.4 and while nothing bad happens anymore it does upload the placeholder files. I would have expected them to be deleted locally. So this will need some more investigation.

There is no way to avoid this right? :-(
(Maybe adding it it in the user exclude list?)

ckamm:
@guruz That problem has been solved by adding the placeholder extension to the user exclude list when the newer version is started. See https://github.com/owncloud/client/pull/6290/commits/ac1cd1942cad341f7b6ea7304893af0a9eadee27

ckamm:
@dschmidt For this PR I added registering of a file association to the NSIS install script.

ckamm:
@ogoffart @SamuAlfageme I added the placeholder option to the account wizard (I had already added it to the folder wizard) and added a warning message box as well. I'd gladly take comments and suggestion improvements on the text for the warning message box. See https://github.com/owncloud/client/pull/6290/commits/8a4d8e333101e3b0e909e0920fcfae4dcb35f500 and particularly https://github.com/owncloud/client/commit/8a4d8e333101e3b0e909e0920fcfae4dcb35f500#diff-97f9501b9b855f5d1b16374f3b0d684fR249

src/gui/wizard/owncloudadvancedsetuppage.ui
```diff
@@ -377,6 +380,46 @@
...
+            <property name="text">
+             <string/>
+            </property>
+            <property name="checkable">
+             <bool>false</bool>
+            </property>
+           </widget>
+          </item>
+          <item>
+           <widget class="QPushButton" name="bPlaceholderSync">
```
ogoffart: Why is it a QPushButton?

The "choose what to sync" option is a button because clicking on it allow to choose what to sync.
But here there is no reason to push the button.

I'd say we should keep a normal checkbox.

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -245,4 +246,22 @@ AbstractCredentials *OwncloudWizard::getCredentials() const
     return 0;
 }
 
+bool OwncloudWizard::askExperimentalPlaceholderFeature()
```
ogoffart: Can we do it without re-entering the eventloop? Have a callback in a std::function<void(bool)>.

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -245,4 +246,22 @@ AbstractCredentials *OwncloudWizard::getCredentials() const
...
+    QMessageBox msgBox(
+        QMessageBox::Warning,
+        tr("Enable experimental feature?"),
+        tr("When the \"synchronize placeholders\" mode is enabled no files will be downloaded initially. "
+           "Instead, a tiny \".owncloud\" file will be created for each file on the server. "
+           "The contents can be downloaded by running these files or by using their context menu."
+           "\n\n"
+           "This is a new, experimental mode. If you decide to use it, please report any "
+           "issues that come up."));
+    QPushButton *yesBtn = msgBox.addButton(tr("Enable experimental mode"), QMessageBox::DestructiveRole);
```
ogoffart: Interrestingly, pressing "Enter" still enable the box.  However, pressing Esc has no effect.

I would not think that the experimental mode is destructive, as one can still uncheck the box.
I would say it is more intuitive if "esc" is the same as "stay save" and "Enter"  the same as enable

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -245,4 +246,22 @@ AbstractCredentials *OwncloudWizard::getCredentials() const
     return 0;
 }
 
+bool OwncloudWizard::askExperimentalPlaceholderFeature()
+{
+    QMessageBox msgBox(
+        QMessageBox::Warning,
+        tr("Enable experimental feature?"),
+        tr("When the \"synchronize placeholders\" mode is enabled no files will be downloaded initially. "
+           "Instead, a tiny \".owncloud\" file will be created for each file on the server. "
```
ogoffart: Branding!

src/gui/wizard/owncloudadvancedsetuppage.cpp
```diff
@@ -326,17 +326,28 @@ void OwncloudAdvancedSetupPage::slotSelectiveSyncClicked()
...
+void OwncloudAdvancedSetupPage::slotPlaceholderSyncClicked()
+{
+    if (!OwncloudWizard::askExperimentalPlaceholderFeature())
+        return;
+
+    _ui.lSelectiveSyncSizeLabel->setText(QString());
+    _selectiveSyncBlacklist.clear();
+
+    setRadioChecked(_ui.rPlaceholderSync);
+}
```
ogoffart: It is missing the code in OwncloudSetupWizard::slotAssistantFinished  to make the placeholder active

SamuAlfageme:
Maybe writing and linking some docs as in https://github.com/owncloud/client/issues/6396 would be cool for this feature as well.

ckamm:
This is now merged! Further work should go into separate tickets. (I agree about adding docs @SamuAlfageme)

guruz:
> @guruz maybe we can take a look at registering the .owncloud extension on macOS soon to publish experimental builds for all 3 platforms together with that.

WiP

Issue #6287, opened 2018-01-09
Linux: Encoding problems with non-uft8 locale


fmkaiser:
### Expected behaviour
Files with special characters in the name are synchronized correctly.

### Actual behaviour
With the 2.4.0 Linux ownCloud client and locale set to LANG=C, 
objects on the server with special characters such as umlauts in the name
get downloaded with the special character replaced by a question mark (?) 
and then renamed on the server on the next sync as well.

That's especially bad if the files are in a folder shared with you by someone else...

### Steps to reproduce
1. Start ownCloud client v2.4 under Linux with
> LANG=C owncloud --logwindow
2. Create a folder with some special characters on the server
3. Watch the sync log

### Server configuration
Doesn't seem to matter much, tested with ownCloud 9.0.8 and 10.0.3.

### Client configuration
Client version: 2.4.0 - Problem did not occur with 2.3.3

Operating system: openSUSE Leap 42.3

OS language: LANG=C

Qt version used by client package (Linux only, see also Settings dialog): Qt 5.6.2

Client package: owncloud-client-2.4.0-8911.1.x86_64 (ownCloud)

### Logs
Client log - the interesting parts:
> \# filename is "täßt.txt" - apologies to all German speakers for the spelling
> 01-09 10:50:43:896 [ info sync.accessmanager ]:	6 "PROPFIND" "https://staging.sis.rzg.mpg.de/remote.php/dav/files/test-fek/umlauttest" has X-Request-ID "2576f085-78e2-49b1-a59e-bb6aa77a3063"
01-09 10:50:44:031 [ info sync.csync.updater ]:	file: umlauttest/t??t.txt, instruction: INSTRUCTION_NEW <<=
01-09 10:50:44:202 [ info sync.database ]:	Updating file record for path: "umlauttest/t\xC3\xA4\xC3\x9Ft.txt" inode: 107843249
01-09 10:50:46:381 [ info sync.propagator ]:	Starting INSTRUCTION_RENAME propagation of "umlauttest/t??t.txt" by OCC::PropagateRemoteMove(0x56329b1622f0)
01-09 10:50:46:382 [ info sync.accessmanager ]:	6 "MOVE" "https://staging.sis.rzg.mpg.de/remote.php/dav/files/test-fek/umlauttest/t??t.txt" has X-Request-ID "e6c9de7a-b8ce-4e73-8bb5-a7c0ceeb55c7"

[...]

PR #6293, opened 2018-01-10
Ignore files that can't be encoded for the filesystem

ckamm:
There's an upstream bug where QTextCodec::canEncode returns true even
though it should be false. This works around that issue and adds a test.

The original work was done in 72809ef5b1aeb578976e4360ed267ac1c4d71ea6

See #6287, #5676, #5719
See https://bugreports.qt.io/browse/QTBUG-6925

### DIFF
--- a/test/testsyncengine.cpp
+++ b/test/testsyncengine.cpp
@@ -586,6 +586,56 @@ private slots:
         QVERIFY(localFileExists("A/.hidden"));
         QVERIFY(fakeFolder.currentRemoteState().find("B/.hidden"));
     }
+
+    void testNoLocalEncoding()
+    {
+        auto utf8Locale = QTextCodec::codecForLocale();
+        if (utf8Locale->mibEnum() != 106) {
+            QSKIP("Test only works for UTF8 locale");
+        }
+
+        FakeFolder fakeFolder{ FileInfo::A12_B12_C12_S12() };
+        QVERIFY(fakeFolder.syncOnce());
+        QCOMPARE(fakeFolder.currentLocalState(), fakeFolder.currentRemoteState());
+
+        // Utf8 locale can sync both
+        fakeFolder.remoteModifier().insert("A/tößt");
+        fakeFolder.remoteModifier().insert("A/t𠜎t");
+        QVERIFY(fakeFolder.syncOnce());
+        QVERIFY(fakeFolder.currentLocalState().find("A/tößt"));
+        QVERIFY(fakeFolder.currentLocalState().find("A/t𠜎t"));
+
+        // Try again with a locale that can represent ö but not 𠜎 (4-byte utf8).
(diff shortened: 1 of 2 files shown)


test/testsyncengine.cpp
```diff
@@ -586,6 +586,34 @@ private slots:
         QVERIFY(localFileExists("A/.hidden"));
         QVERIFY(fakeFolder.currentRemoteState().find("B/.hidden"));
     }
+
+    void testNoLocalEncoding()
+    {
+        auto utf8Locale = QTextCodec::codecForLocale();
+        if (utf8Locale->mibEnum() != 106) {
+            qWarning() << "Test only works for UTF8 locale";
```
ogoffart: There is a QSKIP macro for that

test/testsyncengine.cpp
```diff
@@ -586,6 +586,34 @@ private slots:
...
+        QVERIFY(fakeFolder.currentLocalState().find("A/tößt"));
+
+        QTextCodec::setCodecForLocale(QTextCodec::codecForName("ASCII"));
+        QVERIFY(QTextCodec::codecForLocale()->mibEnum() == 3);
+
+        fakeFolder.remoteModifier().insert("B/tößt");
+        QVERIFY(fakeFolder.syncOnce());
+        QVERIFY(!fakeFolder.currentLocalState().find("B/tößt"));
+        QVERIFY(!fakeFolder.currentLocalState().find("B/t??t"));
+        QVERIFY(!fakeFolder.currentLocalState().find("B/t????t"));
```
ogoffart: sync once more and check that the state on the server did not change. (no new files, no deleted files)

test/testsyncengine.cpp
```diff
@@ -586,6 +586,34 @@ private slots:
...
+
+        FakeFolder fakeFolder{ FileInfo::A12_B12_C12_S12() };
+        QVERIFY(fakeFolder.syncOnce());
+        QCOMPARE(fakeFolder.currentLocalState(), fakeFolder.currentRemoteState());
+
+        fakeFolder.remoteModifier().insert("A/tößt");
+        QVERIFY(fakeFolder.syncOnce());
+        QVERIFY(fakeFolder.currentLocalState().find("A/tößt"));
+
+        QTextCodec::setCodecForLocale(QTextCodec::codecForName("ASCII"));
```
ogoffart: changing the locale after the sync means that the file was already synchronzed.
I would change the locale at the beginning of the test.


Also, while we are at it, it could be better to have a ISO-8859-15 locale,  and check that tößt is proberly synced,  but  tœβt or  🙈 is not

ckamm: That's intentional, I wanted to also verify that with an utf8 locale the file is synchronized as expected.

The intermediate locale sounds like a great idea.

Issue #179, opened 2012-12-14
Sync only the file change, not entire file [$1,755]


wesleyhuang:
In Dropbox, there notes that only the file changes are sync-ed, not the entire file. https://www.dropbox.com/help/8/en.  It is great if ownCloud can do the same.  Especially for large files, sync-ing the entire files make a lot of bandwidth and time unnessararily wasted.

In my testing with latest 4.5.4 and sync client in Ubuntu 12.04, I prepared a 1GB text file, append a few characters to the end, and monitor the traffic and the file in the server, I see the entire 1GB file is transffered to the server and the server is actually creating a new file. 

A reply from the forum indicates that librsync has this feature http://librsync.sourceforge.net/, maybe the csync can be switched to the libsync.

setovi:
Is there an estimated date for this feature? Issue was opened two years ago... TIA.

LukasReschke:
Locking for contributors only because the notifications of all those "+1" are going to disturb me too much.

If you want to implement this feature please create a new issue or even better a Pull Request where we can discuss this. I think this issue left enough hints and more "+1" are not going to change anything.

enoch85:
Anyone working on this?

jospoortvliet:
@gadLinux wants to work on it but needs input from the client developers on what direction to take with this.

dragotin:
I added some brainstorming: https://github.com/owncloud/core/issues/16162#issuecomment-104616147

Note that probably the big thing is in the ownCloud server component.

enoch85:
$1755 folks! Where's the like button on that? :)

PVince81:
Ref, PRs in progress:
- server-side: https://github.com/owncloud/core/pull/29404
- client-side: https://github.com/owncloud/client/pull/6131

PVince81:
Does anyone has an idea what file types delta sync will actually make sense ?

@ahmedammar does delta sync / zsync also detect shifts in a file ? For example if I take a wav file (no compression) and delete a chunk of audio in the middle, will it still detect this or will it resync the whole file ?

ahmedammar:
Technically zsync could detect moves, but my code doesn’t support that, all bytes moved will be resent, this was to make it easier to get something working quickly. I mentioned this as a future work area.

DeepDiver1975:
> Technically zsync could detect moves, but my code doesn’t support that, all bytes moved will be resent, this was to make it easier to get something working quickly. I mentioned this as a future work area.

[...]

PR #6131, opened 2017-10-27
 Implementation of delta-sync support on client-side.

ahmedammar:
This commit adds client-side support for delta-sync, this adds a new 3rdparty submodule `gh:ahmedammar/zsync`. This zsync tree is a modified version of upstream, adding some needed support for the upload path and other requirements.

If the server does not announce the required `zsync` capability then a full upload/download is fallen back to.

On both upload and download paths, a check is made for the existance of a zsync metadata file on the server for a given path. If it doesn't exist the code reverts back to a complete upload or download, i.e. previous implementations. In the case of upload, a new zsync metadata file will be uploaded as part of the chunked upload and future synchronizations will be delta-sync capable.

Chunked uploads no longer use sequential file names for each chunk, instead, they are named as the byte offset into the remote file, this is a minimally intrusive modification to allow fo delta-sync and legacy code paths to run seamlessly. A new http header `OC-Total-File-Length` is sent, which informs the server of the final expected size of the file not just the total transmitted bytes as reported by `OC-Total-Length`.

The seeding of the zsync metadata file is done in a seperate thread since this is a cpu intensive task, ensuring main thread is not blocked.

Added configuration and command-line options to gui and cmd applications. Delta-sync is disabled by default, this can be changed once more testing is done.

PR #6297, opened 2018-01-11
Implementation of delta-sync support on client-side.

ahmedammar:
This commit adds client-side support for delta-sync, this adds a new
3rdparty submodule gh:ahmedammar/zsync. This zsync tree is a modified
version of upstream, adding some needed support for the upload path and
other requirements.
    
If the server does not announce the required zsync capability then a
full upload/download is fallen back to. Delta synchronization can be
enabled/disabled using command line, config, or gui options.
    
On both upload and download paths, a check is made for the existance of
a zsync metadata file on the server for a given path. This is provided
by a dav property called `zsync`, found during discovery phase. If it
doesn't exist the code reverts back to a complete upload or download,
i.e. previous implementations. In the case of upload, a new zsync
metadata file will be uploaded as part of the chunked upload and future
synchronizations will be delta-sync capable.
    
Chunked uploads no longer use sequential file names for each chunk id,
instead, they are named as the byte offset into the remote file, this is
a minimally intrusive modification to allow fo delta-sync and legacy
code paths to run seamlessly. A new http header OC-Total-File-Length is
sent, which informs the server of the final expected size of the file
not just the total transmitted bytes as reported by OC-Total-Length.
    
The seeding and generation of the zsync metadata file is done in a
separate thread since this is a cpu intensive task, ensuring main thread
is not blocked.
    
This commit closes owncloud/client#179.

### DIFF
--- a/src/gui/generalsettings.ui
+++ b/src/gui/generalsettings.ui
@@ -48,6 +48,53 @@
     </widget>
    </item>
    <item row="2" column="0">
+    <widget class="QGroupBox" name="advancedGroupBox">
+     <property name="title">
+      <string>Experimental</string>
+     </property>
+     <layout class="QVBoxLayout" name="verticalLayout_2">
+      <item>
+       <layout class="QHBoxLayout" name="horizontalLayout_2">
+        <item>
+         <widget class="QCheckBox" name="deltaSyncCheckBox">
+          <property name="text">
+           <string>Enable Delta-Synchronization for files larger than</string>
+          </property>
+         </widget>
+        </item>
+        <item>
+         <widget class="QSpinBox" name="deltaSyncSpinBox">
+          <property name="maximum">
+           <number>999999</number>
+          </property>
+         </widget>
+        </item>
+        <item>
+         <widget class="QLabel" name="label_2">
+          <property name="text">
+           <string>MB</string>
+          </property>
+         </widget>
+        </item>
+        <item>
+         <spacer name="horizontalSpacer_2">
+          <property name="orientation">
+           <enum>Qt::Horizontal</enum>
+          </property>
+          <property name="sizeHint" stdset="0">
+           <size>
+            <width>40</width>
+            <height>20</height>
+           </size>
+          </property>
+         </spacer>
+        </item>
+       </layout>
+      </item>
+     </layout>
+    </widget>
+   </item>
+   <item row="3" column="0">
     <widget class="QGroupBox" name="aboutGroupBox">
      <property name="title">
       <string>About</string>
@@ -69,7 +116,7 @@
      </layout>
     </widget>
    </item>
-   <item row="3" column="0">
+   <item row="4" column="0">
     <widget class="QGroupBox" name="updatesGroupBox">
      <property name="title">
       <string>Updates</string>
@@ -120,7 +167,7 @@
      </layout>
     </widget>
    </item>
-   <item row="4" column="0">
+   <item row="5" column="0">
     <spacer name="verticalSpacer">
      <property name="orientation">
       <enum>Qt::Vertical</enum>
--- a/src/libsync/CMakeLists.txt
+++ b/src/libsync/CMakeLists.txt
@@ -43,7 +43,9 @@ set(libsync_SRCS
     owncloudtheme.cpp
     progressdispatcher.cpp
     propagatorjobs.cpp
+    propagatecommonzsync.cpp
     propagatedownload.cpp
+    propagatedownloadzsync.cpp
     propagateupload.cpp
     propagateuploadv1.cpp
     propagateuploadng.cpp
@@ -66,6 +68,37 @@ if(TOKEN_AUTH_ONLY)
 else()
     set (libsync_SRCS ${libsync_SRCS} creds/httpcredentials.cpp)
 endif()
+
+## begin zsync
+
+include_directories( ${CMAKE_SOURCE_DIR}/src/3rdparty/zsync/c )
+
+set( libsync_SRCS
+    ${libsync_SRCS}
+    ../3rdparty/zsync/c/librcksum/hash.c
+    ../3rdparty/zsync/c/librcksum/md4.c
+    ../3rdparty/zsync/c/librcksum/range.c
+    ../3rdparty/zsync/c/librcksum/rsum.c
+    ../3rdparty/zsync/c/librcksum/state.c
+    ../3rdparty/zsync/c/libzsync/sha1.c
+    ../3rdparty/zsync/c/libzsync/zsync.c
+    ../3rdparty/zsync/c/progress.c
+)
+
+if ( WIN32 )
+    # ntohs
+    list(APPEND OS_SPECIFIC_LINK_LIBRARIES
+        ws2_32
+    )
+    # ensure size_t is 64 bits
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_FILE_OFFSET_BITS=64")
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -D_FILE_OFFSET_BITS=64")
+endif()
+
+set_source_files_properties( ../3rdparty/zsync/c/libzsync/zsync.c
+    PROPERTIES COMPILE_FLAGS -DVERSION=\\"0.6.3\\" )
+
+## end zsync
 
 # These headers are installed for libowncloudsync to be used by 3rd party apps
 set(owncloudsync_HEADERS
--- a/src/libsync/progressdispatcher.cpp
+++ b/src/libsync/progressdispatcher.cpp
@@ -195,6 +195,22 @@ void ProgressInfo::adjustTotalsForFile(const SyncFileItem &item)
     }
 }
 
+void ProgressInfo::updateTotalsForFile(const SyncFileItem &item, quint64 newSize)
+{
+    if (!shouldCountProgress(item)) {
+        return;
+    }
+
+    if (!_currentItems.contains(item._file)) {
+        _sizeProgress._total += newSize - item._size;
+    } else {
+        _sizeProgress._total += newSize - _currentItems[item._file]._progress._total;
+    }
+
+    setProgressItem(item, 0);
+    _currentItems[item._file]._progress._total = newSize;
+}
+
 quint64 ProgressInfo::totalFiles() const
 {
     return _fileProgress._total;
@@ -229,7 +245,7 @@ void ProgressInfo::setProgressComplete(const SyncFileItem &item)
     _currentItems.remove(item._file);
     _fileProgress.setCompleted(_fileProgress._completed + item._affectedItems);
     if (ProgressInfo::isSizeDependent(item)) {
-        _totalSizeOfCompletedJobs += item._size;
+        _totalSizeOfCompletedJobs += _currentItems[item._file]._progress._total;
     }
     recomputeCompletedSize();
     _lastCompletedItem = item;
--- a/src/libsync/progressdispatcher.h
+++ b/src/libsync/progressdispatcher.h
@@ -91,6 +91,19 @@ public:
      */
     void adjustTotalsForFile(const SyncFileItem &item);
 
+    /**
+     * Update totals for item.
+     * adjustTotalsForFile is called during the treewalk phase to collect
+     * the initial total size and file count.
+     * updateTotalsForFile is called at most once per item during propagation
+     * to adjust them when new information has become available.
+     *
+     * Example: With delta-sync, the actual size of the download will only
+     * be known during propagation - this function adjusts the total size
+     * to account for it.
+     */
+    void updateTotalsForFile(const SyncFileItem &item, quint64 newSize);
+
     quint64 totalFiles() const;
     quint64 completedFiles() const;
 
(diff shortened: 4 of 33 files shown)


ahmedammar:
webarchive of #6131 (since it unicorns a lot).

https://www.dropbox.com/s/5i0tdccn3dzbjlh/Implementation%20of%20delta-sync%20support%20on%20client-side.%20by%20ahmedammar%20%C2%B7%20Pull%20Request%20%236131%20%C2%B7%20owncloud%3Ac.webarchive?dl=1

src/libsync/CMakeLists.txt
```diff
@@ -15,6 +15,14 @@ if ( APPLE )
     )
 endif()
 
+if ( WIN32 )
+    list(APPEND OS_SPECIFIC_LINK_LIBRARIES
+        ws2_32
+    )
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_FILE_OFFSET_BITS=64")
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -D_FILE_OFFSET_BITS=64")
```
ogoffart: Why is this required?

It would be nice also if everything that is related to zsync lib was together so we could later split it in its own file.

src/libsync/propagatecommonzsync.cpp
```diff
@@ -0,0 +1,362 @@
...
+    // Create a temporary file to use with zsync_begin()
+    QTemporaryFile zsynctf, zsyncmeta;
+    zsyncmeta.open();
+    zsyncmeta.setAutoRemove(false);
+    zsynctf.open();
+
+    int metaHandle = zsyncmeta.handle();
+    int tfHandle = zsynctf.handle();
+
+    zsync_unique_ptr<FILE> meta(fdopen(dup(metaHandle), "w"), [](FILE *f) {
```
ogoffart: Why are you not using QFile here, and everywhere you use raw file operation?
QFile has a constructor that opens a file descriptor as well.
I'd think this would lead to more readable code.

src/libsync/propagatecommonzsync.h
```diff
@@ -0,0 +1,121 @@
...
+
+#include <QLoggingCategory>
+#include <QTemporaryFile>
+#include <QRunnable>
+#include <QThreadPool>
+
+extern "C" {
+#include "librcksum/rcksum.h"
+#include "libzsync/zmap.h"
+#include "libzsync/sha1.h"
```
ogoffart: btw, Qt also has a way to compute the sha1.

src/gui/generalsettings.ui
```diff
@@ -48,6 +48,53 @@
     </widget>
    </item>
    <item row="2" column="0">
+    <widget class="QGroupBox" name="advancedGroupBox">
```
ogoffart: I still think there should be no UI addition.

src/libsync/propagatedownload.h
```diff
@@ -15,33 +15,124 @@
...
 #include "owncloudpropagator.h"
 #include "networkjobs.h"
+#include "propagatecommonzsync.h"
 
 #include <QBuffer>
 #include <QFile>
 
 namespace OCC {
 
+class GETJob : public AbstractNetworkJob
```
ogoffart: If you've got time, you could try tomake a separate commit that refactor GETJob and GETFileJob.
That'd make reviewer slightly easier.

src/libsync/propagatecommonzsync.cpp
ahmedammar: These functions were copied from zsync source as-is since they are not part of the exported api. Needed when generating metadata for a file. My main task was to integrate zsync, not re-write it.

src/libsync/propagatecommonzsync.h
ahmedammar: Again, see previous comment.

src/libsync/propagatecommonzsync.cpp
ahmedammar: Also, please see the archive of all the comments I linked, none of this is *new*.

src/gui/generalsettings.ui
ahmedammar: Then you can feel free to remove it before the merge.

src/libsync/CMakeLists.txt
ahmedammar: Because of size_t usage and assumptions by zsync. See webarchive comments.

src/libsync/propagatedownload.h
ahmedammar: Based on the amount of times and nit-picking happening, my guess it oC don't really want to merge this. That's fine, I can withdraw my PRs and just see if nextCloud want to take them for free.

ahmedammar:
@guruz @ogoffart @dragotin @ckamm @DeepDiver1975 @PVince81
Based on the amount of time that has elapsed, and nit-picking happening, my guess is oC don't really want to merge this. That's fine, I can withdraw my PRs and just see if nextCloud want to take them as-is for free.

src/libsync/CMakeLists.txt
ahmedammar: Regarding putting all the changes together in one place, I can do that, no problem.

src/libsync/propagatedownload.h
ahmedammar: My time is all up on this work now, it's taken way to long to get this far ...

src/libsync/CMakeLists.txt
ahmedammar: Done.

ahmedammar:
Note: Jenkins failures were resolved after rebase on latest master ...

mrow4a:
Is this PR ready from the client side perspective? @guruz @ogoffart

ckamm:
@mrow4a Not ready in my opinion. See below.

@ahmedammar I understand your frustration. I genuinely think you've done an amazing job responding and fixing things based on the hundreds of review comments so far. The old PR ended up being too much for github because you were extremely willing to discuss, adjust and fix all of the issues and nitpicks that came up. :+1: 

Unfortunately this is a large change to a critical component of the sync client and thus everyone seems a bit extra picky and there are worries about reliability. In addition it seems that, with you busy and probably not interested in continuing with this patch after it lands, the existing maintainers will have to fix whatever issues pop up. That makes it necessary for at least two people to understand what's going on well enough to be able to work on it efficiently.

My main issues at this point are:
### reliability on Windows

With this patch zsync will be responsible for a lot of the filesystem operations. One wouldn't expect it, but basic stuff for dealing with files can be surprisingly hard to get right (I'm thinking of mtime timezone issues, locked files, shared access permissions, etc). And I'm convinced we'll have several regressions in this area. 

### reliability generally

While there are now two basic tests (yay!), there will be ample of bugs left to find.

### zsync integration generally

The current PR copies some of zsync's code into the client, other code we link against. In any case, the version we link against is a custom modified version of zsync. Essentially zsync becomes part of the client project, but resides in a different repository while being mutually dependent. If there's no strong reason to keep these separate I'd just fold zsync into the client repository to keep things easy.

### review age

My review was months ago and there have been many changes since (which is great and has improved things!). But still, I'll need to go over things again to get an up-to-date impression of what's going on.


## Path forward?

I'd suggest something like this:
1) Decide whether we're willing to make the investment of adopting this PR. It's pretty cool, but there are costs.
2) Review again with a focus on "does this addition break non-delta-sync codepaths?". Make any changes necessary to reduce risk of accidentally breaking default syncing.
3) Merge in a state where the feature is disabled by default.
4) Have 1-2 people clean up the remaining loose ends.

And then we need to stabilize. That means adding lots of tests, enabling it on dev machines, pinging interested users whether they're up for testing in low-risk scenarios. Eventually enable it by default.

ahmedammar:
First, thanks for the detailed response and sharing your views so succinctly.

All the points made by @ckamm are valid but I just wanted to respond to a few things.

1. I want to reaffirm that I _am_ willing to continue to maintain the code after the code lands, I might be busy with work but that doesn't mean I can't continue to maintain this in my spare time.
2. Regarding folding the zsync code in, I don't think that's a good idea since you'd lose the git commit history. As a compromise, if oC would fork my zsync tree, we could use that as the submodule reference instead.
3. Merging in a default disabled state is actually how it's implemented today and also the reason I made the GUI addition (to allow users to enable/test it).
4. I have no problem in helping clean-up any loose ends or adding more tests but I just want to see some positive movement from oC side before investing more time.

tflidd:
From a user's perspective, this would be really a cool new feature. We've never got so close to this feature. There has been such a demand that there will be some people ready to test it. It will certainly require some development costs as well but what better feature to invest in for a sync solution?

SamuAlfageme:
@tflidd it is indeed; we're thinking about generating some testpilotcloud installers as well as instructions for the server patch to get community to start trying this out as soon as possible to get feedback. We'll announce soon in https://central.owncloud.org.

PVince81:
I don't think a server patch will be needed. We'll merge the server PR and release it in the upcoming 10.1 pre-alpha/alpha so people can use that for testing

ogoffart:
@ahmedammar  Have you had any contact with the zsync maintainer?  Why did you not create pull requests there?
(Sorry if you already said that earlier, but i don't have easy access to previous comment)

ahmedammar:
I reached out to him, and got no response.

ckamm:
@ahmedammar There was a meeting where we talked about this PR. We want to integrate it - that means the "decision" part is done. Upcoming steps are:
1. Final review round with an eye on reducing risks of breaking unrelated behavior when this feature is disabled. Me and @ogoffart will do that.
2. Merge into a branch where the three of us can tie up loose ends. We'll also build experimental binaries for people to play with.
3. Merge into master.

At the same time there'll be another attempt to reach out to the author of zsync. Depending on how this goes we might upstream patches or create a fork in the owncloud organization.

@SamuAlfageme @ogoffart @guruz

src/libsync/progressdispatcher.cpp
```diff
@@ -229,7 +245,7 @@ void ProgressInfo::setProgressComplete(const SyncFileItem &item)
     _currentItems.remove(item._file);
```
ckamm: needs to move below the `if` or `_totalSizeOfCompletedJobs` won't update correctly

src/libsync/progressdispatcher.h
```diff
@@ -91,6 +91,19 @@ class OWNCLOUDSYNC_EXPORT ProgressInfo : public QObject
...
+     * Update totals for item.
+     * adjustTotalsForFile is called during the treewalk phase to collect
+     * the initial total size and file count.
+     * updateTotalsForFile is called at most once per item during propagation
+     * to adjust them when new information has become available.
+     *
+     * Example: With delta-sync, the actual size of the download will only
+     * be known during propagation - this function adjusts the total size
+     * to account for it.
+     */
```
ckamm: for later: This might need even more documentation: item._size must be the old size when this is called, different from `newSize` which isn't obvious to me.

src/libsync/propagateuploadng.cpp
```diff
@@ -131,13 +310,18 @@ void PropagateUploadFileNG::slotPropfindFinished()
...
+    for (auto &chunkId : _serverChunks.keys()) {
+        if (updateRanges(chunkId, _serverChunks[chunkId].size)) {
+            _sent += _serverChunks[chunkId].size;
+            _serverChunks.remove(chunkId);
+        }
     }
 
-    if (_sent > _item->_size) {
+    if (!_rangesToUpload.isEmpty())
+        _currentChunk = _rangesToUpload.first().start;
```
ckamm: later: Potentially redundant, `_currentChunk` is also set in `startNextChunk()` - only used in logging, it seems.

src/libsync/propagateuploadng.cpp
```diff
@@ -30,58 +30,213 @@
...
+        |
+        +-> MOVE +-----> moveJobFinished() +--> finalize()
  */
 
 void PropagateUploadFileNG::doStartUpload()
 {
     propagator()->_activeJobList.append(this);
 
+    _zsyncSupported = isZsyncPropagationEnabled(propagator(), _item);
+    if (_zsyncSupported && _item->_remotePerm.hasPermission(RemotePermissions::HasZSyncMetadata)) {
```
ckamm: later: In the case where this is false, one can still have `_zsyncSupported` be true and use the corresponding code paths below. I haven't looked in detail whether that's correct.

ckamm:
I've merged this PR manually into the `delta-sync` branch (there was a trivial conflict).

ahmedammar:
@ckamm thanks for the review and pulling into a branch but how should I address the things mentioned above after you've already merged into a branch?

ckamm:
@ahmedammar Let's continue working on it collaboratively in the branch by making small PRs against it. I've committed a couple of minor fixes directly but my larger changes will also go through pull requests for review.

orzeech:
@ahmedammar @ckamm Hi, I've been testing this code a bit. I've set up two testing environments and I've found some bigger and smaller problems. From progress bar not being updated correctly to total crash of synchronization after synchronizing many times a group of files from a delta-sync enabled and regular client one after another. There was also a bug when delta-sync was not triggered when adding just the last byte to a file. How do I go on with reporting those? Otherwise it's bringing great speed improvements in some cases.

ahmedammar:
@orzeech are you testing the delta-sync branch that @ckamm is now working in? Since some of these things should be fixed there. Also about reporting issues, reproducibility of the issue is quite important for me to duplicate the situation. Thanks for testing!

orzeech:
I will check on this new branch and get back to you soon

guruz:
This feature will be in the upcoming 2.6 alpha release.
Meanwhile you can try it inside the daily builds 2.6.x https://download.owncloud.com/desktop/daily/

PR #6299, opened 2018-01-11
Use docker for Linux CI

dschmidt:
I don't want to keep the image in my personal dockerhub repo, but I wanna test the image before we move it to the owncloud repo.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -21,8 +21,10 @@ node('CLIENT') {
         checkout scm
         sh '''git submodule update --init'''
 
-    stage 'Linux - GCC'
+    stage 'Linux - Pull Docker Image'
         linux.pull()
+
+    stage 'Linux - GCC'
         linux.inside {
             sh '''
             export HOME="$(pwd)/home"
@@ -36,7 +38,6 @@ node('CLIENT') {
         }
 
     stage 'Linux - Clang'
-        linux.pull()
         linux.inside {
             sh '''
             export HOME="$(pwd)/home"
@@ -49,8 +50,10 @@ node('CLIENT') {
             '''
         }
 
+    stage 'Win32 - Pull Docker Image'
+        win32.pull()
+
     stage 'Win32'
-        win32.pull() // make sure we have the latest available from Docker Hub
         win32.inside {
             sh '''
             rm -rf build-win32


Jenkinsfile
```diff
@@ -12,22 +12,18 @@ node('CLIENT') {
...
+        linux.inside {
+            sh '''
 		mkdir build
 		cd build
-		cmake -DCMAKE_BUILD_TYPE="Debug" -DUNIT_TESTING=1 -DWITH_TESTING=1 -DCMAKE_PREFIX_PATH=/var/lib/jenkins/qt/5.6.2 ..
+		cmake -DCMAKE_BUILD_TYPE="Debug" -DUNIT_TESTING=1 ..
 		make -j4
-		ctest -V --output-on-failure'''
-
-    stage 'Qt5 - clang'
```
DeepDiver1975: you are dropping the clang compilation ... intended?

Jenkinsfile
```diff
@@ -12,22 +12,18 @@ node('CLIENT') {
         checkout scm
         sh '''git submodule update --init'''
 
-    stage 'Qt5'
-        sh '''rm -rf build
+    stage 'Linux'
+        def linux = docker.image('dominikschmidt/owncloud-client-ci-image:latest')
```
DeepDiver1975: ping @tboerger to get this image into https://github.com/owncloud-ci - the windows build image as well

dschmidt: Obviously ... as stated in the PR text ;-)

Jenkinsfile
dschmidt: I took over a diff @ogoffart gave me, was about to discuss/readd this once the normal GCC compilation works.

dschmidt: We can have the discussion here actually:
@ogoffart was that intended or just for quicker testing? You installed clang in the docker image after all.

DeepDiver1975: in the past we had to compile with gcc and clang just because in some situations one of them was failing and the code needed to be adjusted

Jenkinsfile
```diff
@@ -6,46 +6,63 @@
...
+            mkdir build
+            cd build
+            cmake -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ -DCMAKE_BUILD_TYPE="Debug" -DUNIT_TESTING=1 ..
+            make -j4
+            LC_ALL=C.UTF-8 ctest -V --output-on-failure
+            '''
+        }
 
+    stage 'Linux - Clang'
+        linux.pull()
```
ogoffart: should the pull only be done once?

Jenkinsfile
```diff
@@ -6,46 +6,63 @@
...
+        linux.inside {
+            sh '''
+            export HOME="$(pwd)/home"
+            rm -rf build home
+            mkdir build
+            cd build
+            cmake -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ -DCMAKE_BUILD_TYPE="Debug" -DUNIT_TESTING=1 ..
+            make -j4
+            LC_ALL=C.UTF-8 ctest -V --output-on-failure
```
ogoffart: Shouldn't the docker image be configured to use UTF-8 by default?

Jenkinsfile
dschmidt: `.pull()` should be more or less a noop usually, unless the tag is changed in between...

My intention was to make the stages completely independent, so one can be commited out (like you did with the clang stage when working on this thing initially). I realize it might not be the best way to do it - also having the pull in the compile stage makes the times less comparable. I think I would like to move the pulls into their own stages .. sounds good?

Jenkinsfile
dschmidt: What do you mean? It is not configured like that by default. Do you mean I should rather export `LC_ALL` at the top or what is your point?

ogoffart:
The current CI error is:

> 13: 01-11 13:44:38:854 [ warning sync.propagator ]:	Could not complete propagation of "A/.hidden" by OCC::PropagateDownloadFile(0x12d6360) with status 8 and error: "The download would reduce free local disk space below the limit"

Does jenkins cleans the unused docker images?

DeepDiver1975:
> Does jenkins cleans the unused docker images?

after the job execution containers die and get killed - so we should have reproducible results

is this what you are asking for?

Jenkinsfile
dschmidt: This keeps being the long term goal. We had a lengthly discussion with @tboerger , @jnweiger and @ogoffart. 
Basically: If we want to move this to owncloud-ci, we need to base it on Ubuntu LTS. That's problematic because the current Ubuntu LTS has a too old Qt Version and the next one will only be released in April. We want to continue on this, but for now we want to stick to my image - so we can actually get a bit of work done now.

dschmidt:
retest this please

Jenkinsfile
dschmidt: https://stackoverflow.com/questions/28405902/how-to-set-the-locale-inside-a-docker-container

I tried a bunch of these proposed solutions but could none work ... I'm keeping it like I did it initially so we can track the bug *cough*

Issue #6298, opened 2018-01-11
Auto-generated public links don't work if they are longer than 64 characters


brozkeff:
### Expected behaviour
When using Windows Explorer context-menu "Share with ownCloud", tab Public links, and Create new link, the default name of this public link should always be working. If the filename is longer than the 64char limit ownCloud has for public link names, the desktop client should automatically shorten this name so user does not have to manually change anything.

### Actual behaviour
If the filename is long, the owncloud client 2.4.0 (build 8894) on WIndows concatenates (filename) + link to the box left to the "Create new" button, but when user clicks on this "Create new" the client replies on the bottom with a red text error saying "Share name cannot be more than 64 characters".

### Steps to reproduce
1. create file with filename longer than 64 char (or 60 char)
2. try to create public link without changing the default name "(filename) + link"
3. link is not created and owncloud client writes the error message

### Server configuration
Operating system: Linux Ubuntu Server 14.04.5

Web server: Apache 2.4.7

Database: MySQL 5.5.58

PHP version: 7.0.26-2+ubuntu14.04.1+deb.sury.org+2

ownCloud version: 10.0.4 _(correction from 10.0.2, noticed the server was already upgraded)_

### Client configuration
Client version: 2.4.0 build 8894

Operating system: Windows 10 Pro 32bit

OS language: Czech

Installation path of client: default

SamuAlfageme:
@brozkeff yup, indeed it's a server known limitation. There's a discussion ongoing in https://github.com/owncloud/core/issues/29913 

Also see related: https://github.com/owncloud/ios/issues/981 - Thanks for your report!

brozkeff:
@SamuAlfageme Server limitation 64char may be OK but the issue is in the client that even does not know this known server limitation and proposes INVALID public link share name. Since now the share link name is proposed by CLIENT it should be CLIENT proposing a valid less-than-64-char-link and automagically shorten it if it is longer than 64char.  Yes if this default link name generation would be done server-side and client would just display the default proposal from the server than it would be server issue as it proposed invalid name that the server itself cannot accept few seconds later.

ckamm:
The PR implements eliding to 64 characters until a different solution comes along.

SamuAlfageme:
Closing as workaround until https://github.com/owncloud/core/issues/29913 gets solved somehow and we can either recreate the server behavior or drink some sort of default share name from the ocs/share API.

PR #6300, opened 2018-01-12
Link shares: Elide default name to make it <64 chars

ckamm:
Eventually there might be server API for default share name generation.

See owncloud/core#29913, #6298

### DIFF
--- a/src/gui/sharelinkwidget.cpp
+++ b/src/gui/sharelinkwidget.cpp
@@ -58,7 +58,9 @@ ShareLinkWidget::ShareLinkWidget(AccountPtr account,
     //Is this a file or folder?
     QFileInfo fi(localPath);
     _isFile = fi.isFile();
-    _ui->nameLineEdit->setText(tr("%1 link").arg(fi.fileName()));
+
+    // Note: the share name cannot be longer than 64 characters
+    _ui->nameLineEdit->setText(tr("Public link"));
 
     // the following progress indicator widgets are added to layouts which makes them
     // automatically deleted once the dialog dies.


src/gui/sharelinkwidget.cpp
```diff
@@ -58,7 +58,14 @@ ShareLinkWidget::ShareLinkWidget(AccountPtr account,
     //Is this a file or folder?
     QFileInfo fi(localPath);
     _isFile = fi.isFile();
-    _ui->nameLineEdit->setText(tr("%1 link").arg(fi.fileName()));
+
+    // Set the default link share name: it cannot be >64 characters
```
ogoffart: What is the limit? 64 unicode codepoint? Or 64 utf-8 bytes, or 64 utf-16 character ?

This code assume the count is in utf-16 character.   Even if it is the case, we should still not cut in the middle of a high and low surrogate.

ckamm:
@ogoffart You're right. I don't actually know what the exact server limitation is, I assume bytes. 

Building correct eliding for this workaround seems like a waste. Alternate suggestion: Just change the default back to "public link" - no elision necessary, works everywhere, and it's not essential that the item's name is repeated in the link identifier anyway. @SamuAlfageme

SamuAlfageme:
@ckamm back to "Public link" WFM, no over-engineering required on the client side. 

> Server sets the name of the file as default name of the public link. 

in https://github.com/owncloud/client/pull/5695#pullrequestreview-32304429 was just to align server and client behaviors; but if server is limiting and no fix is scheduled in core; let's patch. 👍

Issue #5017, opened 2016-06-28
Excludes: Further optimize also patterns with slash and add trailing slash to some of the items in existing sync-exclude.lst


sagamusix:
### Actual behaviour

Even after fixing #4967, ownCloud keeps eating an entire CPU core for me at times, which I think it already did in 2.1.1 from time to time, so it might be a problem not caught by the fix for #4967.
I'm running ownCloud on a dual-core CPU, so it becomes obvious that something is wrong as soon as another application maxes out the other core. ownCloud's scan thread (which I assume is responsible for this high CPU usage) should run with a rather low priority as it is essentially a background service. This way, "normal" threads should be able to take over the CPU when they need it, meaning that the applications that I actually intend to use would be more responsive.

As the question already came up in #4967, Explorer overlays are disabled.
### Client configuration

Client version: 2.2.2-6192

Operating system: Windows 7 x64

OS language: German

Installation path of client: Default

guruz:
The solutions are https://github.com/owncloud/client/issues/2297 and https://github.com/owncloud/client/pull/4999

Although I'm wondering if you're right about that we shut throttle the discovery a bit to not make the CPU go up so much and the CPU fan spin. Opinions @owncloud/desktop-developers ?

sagamusix:
I do think it would be a sensible modification. Naturally optimizing the code itself should be one important part of fixing performance-related problems. But on the other hand (unless there is any artificial throttling) it is just natural that ownCloud will consume an entire CPU core when scanning folders. This is exactly what you want (it is generally known that "race to idle" is a good power-saving strategy) when you are alone on the system, but when there are other threads, you are competing for resources.
Other threads will most likely include other end-user software that the user actually intends to interact with, while ownCloud is just a background thread. Hence, if a lower priority is assigned to this thread, it will run at full performance when there is nothing to do (and race to idle), but if the user actually works with their system, the ownCloud thread will get fewer time slices and the remaining software will work more smoothly.
I have verified this to work as intended by setting the entire ownCloud process' priority to "lower" in the Windows task manager - when doing so, my desktop becomes more usable again.

ckamm:
@sagamusix I've changed the discovery thread's priority to 'low' for 2.3.0.

sagamusix:
Thanks, looking forward to test it. :)

guruz:
@sagamusix http://download.owncloud.com/desktop/daily/ownCloud-2.3.0.6228-nightly20160705-setup.exe :)

[...]

PR #6301, opened 2018-01-12
Exclude matching: Speedup the full-path traversal case

ckamm:
Previously we'd use the full regex when the bname triggered a full-path
matching to take place. Now we have a simplified full-traversal regex
for this case that can be significantly faster to apply.

Triggered by #5017 but doesn't actually solve it.

### DIFF
--- a/src/csync/csync_exclude.cpp
+++ b/src/csync/csync_exclude.cpp
@@ -361,29 +361,34 @@ CSYNC_EXCLUDE_TYPE ExcludedFiles::traversalPatternMatch(const char *path, ItemTy
 
     QRegularExpressionMatch m;
     if (filetype == ItemTypeDirectory) {
-        m = _bnameActivationRegexDir.match(bnameStr);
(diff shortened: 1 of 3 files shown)


src/csync/csync_exclude.cpp
```diff
@@ -361,29 +361,34 @@ CSYNC_EXCLUDE_TYPE ExcludedFiles::traversalPatternMatch(const char *path, ItemTy
...
     } else {
-        m = _bnameActivationRegexFile.match(bnameStr);
+        m = _bnameTraversalRegexFile.match(bnameStr);
     }
     if (!m.hasMatch())
-        return match;
-
-    // Now run the full match
+        return CSYNC_NOT_EXCLUDED;
+    if (!m.captured(1).isEmpty()) {
```
ogoffart: Just thinking:  Wouldn't it make the code more readable if we used named captures?  Or would that be slower?

ckamm: Good point, I'll test that!

PR #6310, opened 2018-01-17
Enable appveyor builds

dschmidt:


ogoffart:
looks good, the build is passing :-)

dschmidt:
I can probably add a status image to the readme :)

ogoffart:
I'm currious... how does this work?  I don't see the call to cmake. Is craft doing that? how does it know how to run cmake?

dschmidt:
@ogoffart https://github.com/owncloud/craft-blueprints-owncloud/blob/master/owncloud/owncloud-client/owncloud-client.py#L25

dschmidt:
We talked about it. Yes, that's indeed a (minor) problem because UNIT_TESTING enables additional codepaths that are undesirable in release builds. @ogoffart will take care of getting rid of it.

Issue #6312, opened 2018-01-18
Conflict between file and folder is not handled


ckamm:
Basic case:
* Add a file "foo" locally
* Add a folder "foo" remotely, with some files inside
* Start the sync

The sync will not complete successfully. The same happens if the folder is local and the file is remote. Or if what used to be a file becomes a directory on one end while the file changes on the other end.

What should happen is the usual conflict behavior: The local item gets renamed and the remote item gets downloaded.

I have already started to work on this issue.

SamuAlfageme:
Hmm... while I'm testing this /me wonders if this would be a nice way to solve cash-clashes on the client (things like https://github.com/owncloud/client/issues/5870#issuecomment-347564603 should get better with this). 

Using conflicts could be a way (not sure if good or not) to solve part of the neverending story: https://github.com/owncloud/client/issues/1348. What'cha think @ckamm?

ckamm:
@SamuAlfageme Making case clashes conflicts creates the following problem:

* Linux user: uploads foo.txt and Foo.txt. Syncs and updates them, all is good.
* Windows user starts syncing, gets foo.txt and Foo_case_clash_with_foo.txt. That doesn't make sense, foo and Foo are entirely unrelated.

SamuAlfageme:
Fixed 🎉  - great catch @ckamm!

PR #6314, opened 2018-01-18
Sync: Deal with file/folder conflicts

ckamm:
Previously conflicts with a different type on both ends lead to sync
errors. Now they are handled in the expected way: the local item gets
renamed and the remote item gets propagated downwards.

This also adds a unittest for the TYPE_CHANGE case. That one looks like
parts of it might be unified with CONFLICT cases.

See  #6312

### DIFF
--- a/src/libsync/owncloudpropagator.cpp
+++ b/src/libsync/owncloudpropagator.cpp
@@ -22,6 +22,7 @@
 #include "propagateremotemove.h"
 #include "propagateremotemkdir.h"
 #include "propagatorjobs.h"
+#include "filesystem.h"
 #include "common/utility.h"
 #include "account.h"
 #include "common/asserts.h"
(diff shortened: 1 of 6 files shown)


src/libsync/owncloudpropagator.cpp
```diff
@@ -368,8 +369,10 @@ PropagateItemJob *OwncloudPropagator::createJob(const SyncFileItemPtr &item)
             return new PropagateRemoteDelete(this, item);
     case CSYNC_INSTRUCTION_NEW:
     case CSYNC_INSTRUCTION_TYPE_CHANGE:
+    case CSYNC_INSTRUCTION_CONFLICT:
         if (item->isDirectory()) {
-            if (item->_direction == SyncFileItem::Down) {
+            // CONFLICT has _direction == None
+            if (item->_direction != SyncFileItem::Up) {
                 auto job = new PropagateLocalMkdir(this, item);
```
ogoffart: And this works because conflict only happen on type change for directories?

test/testsyncconflict.cpp
```diff
@@ -345,6 +345,211 @@ private slots:
...
+            QDir(fakeFolder.localPath() + conflict).removeRecursively();
+        QCOMPARE(fakeFolder.currentLocalState(), fakeFolder.currentRemoteState());
+
+        // Currently a1 and b1 don't get moved, but redownloaded
+    }
+
+    void testTypeChange()
+    {
+        Logger::instance()->setLogDebug(true);
+        Logger::instance()->setLogFile("-");
```
ogoffart: is that some leftover from some debugging?

(Btw, I'm always compiling the client with `-DNO_MSG_HANDLER=1` , keeping the default handler help while developping usually)

Issue #6292, opened 2018-01-10
[Meta] "ownCloud" context menu options Compilation


SamuAlfageme:
- [x] Add "Open file/folder in browser" to shell extension context menu - https://github.com/owncloud/client/issues/5903

## Dynamically disable/grey-out the context menu dropdown
- [x] When folder is not ready yet: https://github.com/owncloud/client/issues/4608
- [x] When sharing is disabled in server (globally) or file (`oc:permissions`): https://github.com/owncloud/client/issues/4205
- [ ] When the client is offline/server unreacheable (just the "Share" entry, i.e. mechanism to generate Private link based on DB contents is offline-friendly already)

## Additional tickets
- [ ] [Discussion] Multi-account locally shared folders (https://github.com/owncloud/client/issues/3764) - Add an intermediate "account" dropdown. https://github.com/owncloud/client/issues/5462
- [ ] [Discussion] Enable a `Move to ownCloud > account` outside the local sync folder? - Was disabled in: https://github.com/owncloud/client/issues/2732 
- [ ] Different actions based on the file/folder permissions: https://github.com/owncloud/client/issues/6252
- [ ] "Public link" share shortcut: https://github.com/owncloud/client/issues/6356

mmattel:
Maybe to add under additional tickets: https://github.com/owncloud/client/issues/6202
This is something where core is involved, but it affects the client

Issue #6252, opened 2017-12-17
[Feature Request] Define actions on read only mounts


mmattel:
Referencing: https://github.com/owncloud/core/issues/29873 ([Feature Request] Mark mount as read only in the mount settings)

If you try to write into a read only marked mount, you can do so because the filesystem allows it but the change gets never synced and you do not know what the original data was.

1. make the issue aware to the user
2. propose actions to the user like 
  -- "copy the data to a different place" 
  -- "restore original data"
  -- combinations of
  -- ...

@SamuAlfageme @PVince81

michaelstingl:
From client perspective, mounted storages with read only permission should behave the same way as received shares with read only permission? Currently nothing happens, but files/folders get error overlay icon and the client UI shows there is a problem.

@mmattel What do you mean with „restore original data“? If I delete files locally, they immediately download again.

mmattel:
> From client perspective, mounted storages with read only permission should behave the same way as received shares with read only permission?

Yes

> but files/folders get error overlay icon and the client UI shows there is a problem

But you do not know the cause. Right-click on that file and because the client knows the mount is read only, actions should be proposed

> What do you mean with „restore original data“? If I delete files locally, they immediately download again.

This is great, but if you change the file and save it, the changed version persists in status "problem". Therefore an action should be proposed to the user like move and as you say the original will be downloaded again. The same if you write a fresh file into a read only mount.

mmattel:
Just installed 2.4.0
It would be an idea, at least part of, to add actions to the item list (see below), but only in case the underlaying mount is read only...

![image](https://user-images.githubusercontent.com/3321281/34412715-13268172-ebdf-11e7-8580-c3a334fc5e2e.png)

mmattel:
From https://github.com/owncloud/core/issues/30070#issuecomment-356531035
> "oc:permissions" is fetched and used by the client in general, not only for mounts.

This means that the client does not need to wait for a core implementation as it is already present 😄 

A client implementation could start immediateley, because ``oc:permissions`` has the read-only attribute on read-only shares (=mount from the reciever side) already set.

If https://github.com/owncloud/core/issues/29873 ([Feature Request] Mark mount as read only in the mount settings) gets realized, no change would be needed in the client side!

(Sorry if some things are now double present, but I just wanted to sort out and document properly)

mmattel:
There is another good point for implementing a solution.
[...]

Issue #5903, opened 2017-07-14
Add "Open file/folder in browser" to shell extension context menu


ckamm:
See https://github.com/owncloud/client/issues/5023#issuecomment-315329650 my @michaelstingl 

There are options to copy or email the private link, but not one to open it in a browser.

guruz:
Works for me, tested on macOS.

jnweiger:
### mint linux tara
tested with nautilus browser.

OK works fine.

PR #6315, opened 2018-01-18
SocketAPI: Make it easier to add or remove item in the action menu

ogoffart:
By making it dynamic.
So far only the nautilus and dolphin shell extension have been ported

### DIFF
--- a/shell_integration/windows/OCContextMenu/OCClientInterface.cpp
+++ b/shell_integration/windows/OCContextMenu/OCClientInterface.cpp
@@ -34,7 +34,7 @@ using namespace std;
 #define PIPE_TIMEOUT  5*1000 //ms
 #define SOCK_BUFFER 4096
 
-OCClientInterface::ContextMenuInfo OCClientInterface::FetchInfo()
+OCClientInterface::ContextMenuInfo OCClientInterface::FetchInfo(const std::wstring &files)
 {
     auto pipename = CommunicationSocket::DefaultPipePath();
 
@@ -45,7 +45,8 @@ OCClientInterface::ContextMenuInfo OCClientInterface::FetchInfo()
     if (!socket.Connect(pipename)) {
         return {};
     }
-    socket.SendMsg(L"GET_STRINGS\n");
+    socket.SendMsg(L"GET_STRINGS:CONTEXT_MENU_TITLE\n");
+    socket.SendMsg((L"GET_MENU_ITEMS:" + files + L"\n").data());
 
     ContextMenuInfo info;
     std::wstring response;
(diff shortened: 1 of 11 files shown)


michaelstingl:
sounds 🆒  😎

ogoffart:
The reason to do that is that the new placeholder change might want to add entries in this menu, and so it would be much easier if we can change easily the entries without having to modify every shell extension.

So far I only did the change to nautilus and dolphin. Waiting for feedback before doing the rest.

The filter parametter is a regexp to on the whole filename.   And the filename is made to ends with '/' for directories.
So we can have special entries for directories, or .owncloud files.
I realized when porting nautilus that it only add entries on files that have the OK status, this might be a problem and something we might want to encode in the filter instead.

Also i just tought perhaps there should be a way to set the icon

SamuAlfageme:
Some relevant tickets that'd benefit from this: https://github.com/owncloud/client/issues/5903 (if I understood correctly) and, on a deeper level: https://github.com/owncloud/client/issues/6292

ogoffart:
@SamuAlfageme  Yes, this would help for all these issue.

There are some issues about graying out the menu for files that can be shared and so on. So I need to brainstorm with @ckamm on smarter filter.
An idea i had was that we could add one more string in the "STATUS:" replies or update and filter on that instead of on the name. That would work fine for Nautilus in which all extension are the same. That would somehow work for windows because we do synchronous query anyway we can as well query the status.
But for Dolphin that might be harder because the action plugin don't have access to the status right now.

We also need to encode in the filter if we should disable or remove the options.
Also some options might be valid if there is a multiple selection.

ckamm:
@SamuAlfageme @ogoffart To make it easier to think about this, here's a list of behaviors we want:
* Placeholder files should have a "Download this file" menu action, not others.
* Files (not directories) might have a "Remove local copy, keep remote" menu action to convert back to placeholder only.
* If the server has sharing disabled the "Share..." action should be hidden.
* For files that can't be shared the "Share..." action should be disabled. (Problem: User doesn't understand why)
* If the client isn't connected the "Share..." action should be disabled. (Problem: User doesn't understand why)
* It should be easy to add new simple actions that just trigger some behavior in the client.
* Maybe multiaccount issues - but it's so far out of usual user behavior that I'd not worry overmuch

I did not fully get whether #4608 and #6252 are related too.

shell_integration/windows/OCContextMenu/OCClientInterface.cpp
```diff
@@ -45,7 +45,8 @@ OCClientInterface::ContextMenuInfo OCClientInterface::FetchInfo()
     if (!socket.Connect(pipename)) {
         return {};
     }
-    socket.SendMsg(L"GET_STRINGS\n");
+    socket.SendMsg(L"GET_STRINGS:CONTEXT_MENU_TITLE\n");
```
ckamm: I don't think you can request a particular string only right now.

shell_integration/windows/OCUtil/StringUtil.h
```diff
@@ -61,6 +61,30 @@ class __declspec(dllexport) StringUtil {
...
+        auto statusBegin = source.find(L':', 0);
+        assert(statusBegin != std::wstring::npos);
+
+        auto statusEnd = source.find(L':', statusBegin + 1);
+        if (statusEnd == std::wstring::npos) {
+            // the command do not contains two colon?
+            return false;
+        }
+
+        auto thirdCollon = source.find(L':', statusEnd + 1);
```
ckamm: thirdColon

ogoffart: will fix

src/gui/socketapi.cpp
```diff
@@ -545,11 +545,37 @@ void SocketApi::command_GET_STRINGS(const QString &, SocketListener *listener)
     } };
     listener->sendMessage(QString("GET_STRINGS:BEGIN"));
     for (auto key_value : strings) {
-        listener->sendMessage(QString("STRING:%1:%2").arg(key_value.first, key_value.second));
+        if (argument.isEmpty() || argument == QLatin1String(key_value.first)) {
```
ogoffart: @ckamm:  This change allow to request a special string.

shell_integration/windows/OCContextMenu/OCClientInterface.cpp
ogoffart: See the change in SocketApi::command_GET_STRINGS

src/gui/socketapi.cpp
```diff
@@ -545,11 +545,37 @@ void SocketApi::command_GET_STRINGS(const QString &, SocketListener *listener)
...
+            listener->sendMessage(QString("STRING:%1:%2").arg(key_value.first, key_value.second));
+        }
     }
     listener->sendMessage(QString("GET_STRINGS:END"));
 }
 
+void SocketApi::command_GET_MENU_ITEMS(const QString &argument, OCC::SocketListener *listener)
+{
+    listener->sendMessage(QString("GET_MENU_ITEMS:BEGIN"));
+    bool hasSeveralFiles = argument.contains(QLatin1Char('\x1e')); // Record Separator
```
guruz: So while the extensions do support selecting and then querying multiple files, the client does not yet do I understand this correctly? @ogoffart @ckamm

ogoffart: The placeholder branch does.

ckamm: Yep, you can ask to download multiple placeholders at once with this functionality :)

PR #6319, opened 2018-01-19
Nautilus shell integration: Port to Python 3

ogoffart:
This is still WiP, as I don't know if it will work with python 2.7

Any suggestion on how to make it compatible with both?

### DIFF
--- a/shell_integration/nautilus/syncstate.py
+++ b/shell_integration/nautilus/syncstate.py
@@ -15,9 +15,13 @@
 # or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
 # for more details.
 
+import sys
+python3 = sys.version_info[0] >= 3
+
 import os
 import urllib
-import urllib.parse
+if python3:
+    import urllib.parse
 import socket
 import tempfile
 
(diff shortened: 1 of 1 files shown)


ckamm:
Huh, can we tell which python version nautilus is using to process the shell integrations? (I was also wondering whether I can use `time.monotonic()` which is in >=3.3 only)

ckamm:
FWIW my nautilus on Ubuntu 17.10 uses Python 2.7.14 (sys.version).

ogoffart:
On my archlinux it is using python 3

I think this is because i don't have python2-nautilus installed, and maybe installing it would have fixed my problem without porting to python3

Possibly, we can force python 3 by adding `#!/usr/bin/env python3`   as the first line

ckamm:
@ogoffart No, the file is not being `exec()`ed but loaded into some custom python context that probably comes with a specific version. (I tried)

ckamm:
And these changes don't work with the 2.7 version. It seems we'll have to do version detection in the scipt.

ckamm:
The updated version works with my python2 based nautilus. Please double check against python3.

ogoffart:
Merged manually. 
(There was still a few encode missing in code path that i had not ran before)

Issue #6318, opened 2018-01-18
UNIT_TESTING enables additional code paths in application


dschmidt:
This is kind of a code smell and makes it harder for us to build release builds and test builds with craft.

ogoffart:
The stopwatch is gone in #6320,  there is still one annoying one with OWNCLOUD_TEST_CASE_PRESERVING.  We don't want to do a qgetenv too often in release mode.

PR #6320, opened 2018-01-19
Get rid of the StopWatch in the PropagateUpload job

ogoffart:
For issue #6318

The StopWatch is using memory, and we are not really using it.


ogoffart:
(appveyor failure was unrelated)

Issue #6316, opened 2018-01-18
misleading "conflict file downloaded" activity message 


moscicki:
Sync client (2.3.3 OSX) reported conflict files but it did not create them:

![image](https://user-images.githubusercontent.com/621965/35109029-7d0f02c0-fc75-11e7-9f9d-2dabe0d20a29.png)

Why is that? AFAIK there is a content check of the temporarily downloaded conflict file and if it is the same no action is taken -- is this correct? In this case the message should not be there because it is confusing.

Or is this some other abnormal condition?

moscicki:
Actually looking into server logs do not see any GET for these files. So this looks like another wierd error condition.

ogoffart:
Version 2.3 should always have downloaded.
Version 2.4 does more check and avoid downloading if the file hashes are the same. But still, we should not see this message in that case.

Do you have any ways to reproduce the problem?

ckamm:
I can reproduce this. All items with a `CONFLICT` instruction get that message, even if they end up not downloading the remote file, or downloading it and realizing both files are identical.

guruz:
Maybe @moscicki can test with a daily build?

SamuAlfageme:
All my tests came up green after 2.4.1. All conflicts solved on hash comparison are no longer displayed on the issues tab. Very nice & clean code:

https://github.com/owncloud/client/blob/5f54a601083d2941ba01d81d26bb541f602bbbf2/src/libsync/syncfileitem.h#L190-L206

Great work @ckamm! 🙌

PR #6321, opened 2018-01-19
Activity: Avoid misleading message for INSTRUCTION_CONFLICT

ckamm:
The instruction just indicates a potential conflict. If we can determine
that there's no actual conflict, the activity item shouldn't announce
that a conflict file was created.

The "Downloaded" message is not necessarily technically correct, but
seems less confusing than "It looked like there might be a conflict, but
there didn't end up being one after all".

See #6316

### DIFF
--- a/src/gui/issueswidget.cpp
+++ b/src/gui/issueswidget.cpp
@@ -216,7 +216,7 @@ void IssuesWidget::slotProgressInfo(const QString &folder, const ProgressInfo &p
 
 void IssuesWidget::slotItemCompleted(const QString &folder, const SyncFileItemPtr &item)
 {
-    if (!item->hasErrorStatus())
+    if (!item->showInIssuesTab())
         return;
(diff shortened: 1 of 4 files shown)


src/libsync/progressdispatcher.cpp
```diff
@@ -34,7 +34,11 @@ QString Progress::asResultString(const SyncFileItem &item)
             return QCoreApplication::translate("progress", "Uploaded");
         }
     case CSYNC_INSTRUCTION_CONFLICT:
-        return QCoreApplication::translate("progress", "Server version downloaded, copied changed local file into conflict file");
+        if (item._status == SyncFileItem::Conflict) {
+            return QCoreApplication::translate("progress", "Server version downloaded, copied changed local file into conflict file");
+        } else {
+            return QCoreApplication::translate("progress", "Downloaded");
```
ogoffart: Downloaded? Would it not be better to remove the progress info completely?

Issue #4205, opened 2015-11-26
'Share with owncloud' option should not be shown if sharing is not available


davitol:
If  sharing is not allowed, it would be better to hide 'Share with owncloud' when right-clicking on a file instead of let select this option and disable the content of sharing window.

![screen shot 2015-11-26 at 18 47 42](https://cloud.githubusercontent.com/assets/10596560/11428740/0a9046b6-946f-11e5-856a-2be7ebf4b8b3.png)

Dianafg76:
'Share with owncloud' option should not be shown if sharing is not available, and when you have this option enable
![screen shot 2016-02-12 at 10 47 36](https://cloud.githubusercontent.com/assets/12032058/13003495/27b9ba12-d176-11e5-8abc-d458ac550f3b.png)

Desktop v ownCloud-2.2.0.3112-nightly20160211.pkg
Server v 8.2.2

SamuAlfageme:
Closed with https://github.com/owncloud/client/pull/6330 - ref:

<p align="center">
  <img src="https://user-images.githubusercontent.com/2644445/38661151-2d3b0a6c-3e30-11e8-8a8b-0bf4e71f6c5c.png"/>
</p>

Issue #6328, opened 2018-01-23
Port the mac shell extension to the new protocol


ogoffart:
The mac shell extension should be ported to the new protocol from
https://github.com/owncloud/client/pull/6315

SamuAlfageme:
@guruz is it feasible to ship this one on 2.5? (just to have all platforms on the same page).

guruz:
```
04-09 19:32:07:900 [ info gui.socketapi ]:	Received SocketAPI message <-- "GET_STRINGS:" from SocketApiSocket(0x7fa5918697b0)
04-09 19:32:07:900 [ info gui.socketapi ]:	Sending SocketAPI message --> "GET_STRINGS:BEGIN" to SocketApiSocket(0x7fa5918697b0)
04-09 19:32:07:901 [ info gui.socketapi ]:	Sending SocketAPI message --> "STRING:SHARE_MENU_TITLE:Share..." to SocketApiSocket(0x7fa5918697b0)
04-09 19:32:07:901 [ info gui.socketapi ]:	Sending SocketAPI message --> "STRING:CONTEXT_MENU_TITLE:ownCloud" to SocketApiSocket(0x7fa5918697b0)
04-09 19:32:07:901 [ info gui.socketapi ]:	Sending SocketAPI message --> "STRING:COPY_PRIVATE_LINK_MENU_TITLE:Copy private link to clipboard" to SocketApiSocket(0x7fa5918697b0)
04-09 19:32:07:902 [ info gui.socketapi ]:	Sending SocketAPI message --> "STRING:EMAIL_PRIVATE_LINK_MENU_TITLE:Send private link by email..." to SocketApiSocket(0x7fa5918697b0)
04-09 19:32:07:902 [ info gui.socketapi ]:	Sending SocketAPI message --> "STRING::" to SocketApiSocket(0x7fa5918697b0)
04-09 19:32:07:903 [ info gui.socketapi ]:	Sending SocketAPI message --> "GET_STRINGS:END" to SocketApiSocket(0x7fa5918697b0)
```

@ogoffart is it normal to get an empty STRING:: ?

ogoffart:
The empty STRING:: is a minor bug that's been there for ever i think.
You can use  `GET_STRINGS:CONTEXT_MENU_TITLE`  command, and you only get the one string you need.

guruz:
Yay..

[...]

PR #6330, opened 2018-01-23
SocketAPI: "Open in browser" and disable Share entries when sharing is disabled

ogoffart:
This adds "Open in browser" entry in the menu (Issue #5903)

Also mark the share entries as disabled when sharing is not allowed
or not available for a given file.
If sharing is disabled globaly in the branding or in the server,
the share entry will not be present.
(Issues #4205 and #4608)

Meta issue #6292

### DIFF
--- a/src/gui/socketapi.cpp
+++ b/src/gui/socketapi.cpp
@@ -522,6 +522,11 @@ void SocketApi::command_EMAIL_PRIVATE_LINK(const QString &localFile, SocketListe
     fetchPrivateLinkUrlHelper(localFile, this, &SocketApi::emailPrivateLink);
 }
 
+void SocketApi::command_OPEN_PRIVATE_LINK(const QString &localFile, SocketListener *)
+{
+    fetchPrivateLinkUrlHelper(localFile, this, &SocketApi::openPrivateLink);
+}
+
 void SocketApi::copyPrivateLinkToClipboard(const QString &link) const
 {
     QApplication::clipboard()->setText(link);
@@ -533,6 +538,11 @@ void SocketApi::emailPrivateLink(const QString &link) const
         tr("I shared something with you"),
         link,
         0);
+}
+
+void OCC::SocketApi::openPrivateLink(const QString &link) const
+{
+    Utility::openBrowser(link, nullptr);
 }
 
 void SocketApi::command_GET_STRINGS(const QString &argument, SocketListener *listener)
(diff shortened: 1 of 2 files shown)


ogoffart:
Should work on All platform except OSX. (because of https://github.com/owncloud/client/issues/6328 )

src/gui/socketapi.cpp
```diff
@@ -555,22 +565,36 @@ void SocketApi::command_GET_STRINGS(const QString &argument, SocketListener *lis
...
             systemPath.truncate(systemPath.length() - 1);
         }
-        QString relativePath = systemPath.mid(syncFolder->cleanPath().length() + 1);
 
         SyncJournalFileRecord rec;
-        if (syncFolder->accountState()->isConnected() && syncFolder->journalDb()->getFileRecord(relativePath, &rec) && rec.isValid()) {
+        QString relativePath = systemPath.mid(syncFolder->cleanPath().length() + 1);
+        bool isOnTheServer = syncFolder->journalDb()->getFileRecord(relativePath, &rec) && rec.isValid();
+
+        listener->sendMessage(QLatin1String("MENU_ITEM:OPEN_PRIVATE_LINK") + flagString(isOnTheServer) + tr("Open in browser"));
```
ckamm: Minor: I think it would look nicer if this was the last option in the menu, because it belongs with the other "private link" actions.

ckamm:
@SamuAlfageme This is what was requested in the ticket but I still think disabling sharing options for files that can't be reshared is confusing. The user doesn't know why the options are suddenly disabled! 

How about we put a single disabled "This item cannot be reshared" action in the menu instead?

src/gui/socketapi.cpp
ogoffart: I put it first because I thought it would be the most commonly used action.  But I can put it last.

ogoffart:
> How about we put a single disabled "This item cannot be reshared" action in the menu instead?

I like that idea.

So if the file is not on the server, it can be the normal menu, disabled (then normally the overlay icon should show it is syncing)
But if the permissions are wrong, change the menu to a disabled "this item cannot be reshared"   Do the email/copy private link menu actions still make sense in this case?

ckamm:
@ogoffart I'm not certain about email/copy private link. Why would you want to do that if you can't share it with anyone. That's why I thought disabling them when resharing is disabled is fine.

But maybe your private link works with the person who originally shared the item with you? Then the actions could be useful. I'll test that now.

ogoffart:
I updated the request.  The "open in browser" is now the last item.

If sharing is not allowed, there is a new disabled entry that says it is not allowed.
But the email and copy private link are still enabled.

src/gui/socketapi.cpp
```diff
@@ -557,21 +567,33 @@ void SocketApi::command_GET_MENU_ITEMS(const QString &argument, OCC::SocketListe
...
+        bool isOnTheServer = syncFolder->journalDb()->getFileRecord(relativePath, &rec) && rec.isValid();
+        auto flagString = isOnTheServer ? QLatin1String("::") : QLatin1String(":d:");
+
+        auto capabilities = syncFolder->accountState()->account()->capabilities();
+        auto theme = Theme::instance();
+        if (capabilities.shareAPI() && (theme->userGroupSharing() || (theme->linkSharing() && capabilities.sharePublicLink()))) {
+            // If sharing is globally disabled, do not show any sharing entries.
+            // If there is no permission to share for this file, add a disabled entry saying so
+            if (isOnTheServer && !rec._remotePerm.isNull() && !rec._remotePerm.hasPermission(RemotePermissions::CanReshare)) {
+                listener->sendMessage(QLatin1String("MENU_ITEM:DISABLED:d:") + tr("Re-sharing this file is not allowed"));
```
ckamm: Very minor: it's called "resharing" without dash on the server

SamuAlfageme:
@ogoffart 

> If sharing is not allowed, there is a new disabled entry that says it is not allowed.
> But the email and copy private link are still enabled.

That's working like a clock when sharing is disabled for a file ("reshare"):

resharing_not_allowed

But **when sharing is disabled in capabilities** we are not displaying the "private link entries": 

<p align="center">
  
</p>

(The icon overlay also stays in "shared" but this happens in the webUI as well - IIRC there was no "remove sharing from all `oc:permissions`" event when disabling this switch)

ckamm:
@SamuAlfageme You're saying: Please show the private link options when sharing capabilities are disabled? Does that make sense?

I thought the primary usecase for private links was for pointing to files inside a directory you had shared with someone - why would one care about private links when sharing is impossible globally? (I understand it makes sense when resharing is forbidden - file could already be shared)

SamuAlfageme:
@ckamm you're absolutely right on your explanation - I got myself influenced by what it happens on the webUI:

no_sharing_link

(I'll poke PMs to see if they should disable that switch in there as well)

michaelstingl:
@pmaier1 FYI ^

Thought about it. Private links maybe could stay, even when sharing is disabled. They are useful for the user itself. With "open in browser" feature in client, demand is probably smaller.

PR #6333, opened 2018-01-24
Use std::chrono::millisecond to represent milliseconds

ogoffart:


src/libsync/propagateuploadng.cpp
```diff
@@ -364,12 +364,10 @@ void PropagateUploadFileNG::slotPutFinished()
...
-    double targetDuration = propagator()->syncOptions()._targetChunkUploadDuration;
-    if (targetDuration > 0) {
-        double uploadTime = job->msSinceStart() + 1; // add one to avoid div-by-zero
-
-        auto predictedGoodSize = static_cast<quint64>(
-            _currentChunkSize / uploadTime * targetDuration);
+    auto targetDuration = propagator()->syncOptions()._targetChunkUploadDuration;
+    if (targetDuration.count() > 0) {
+        auto uploadTime = ++job->msSinceStart(); // add one to avoid div-by-zero
+        auto predictedGoodSize = _currentChunkSize * targetDuration / uploadTime;
```
ckamm: This particular part was hard to read. So `targetDuration` and `uploadTime` are `std::chrono::duration` and `_currentChunkSize` is an integer. If I read things right, the division of two `duration` is a plain integer (makes sense). But previously this was using floating point operations.

So while this is probably okay since `_currentChunkSize * targetDuration >> uploadTime`, that's not immediately obvious. I'd suggest making this more obvious, or explicitly doing floating point operations here by casting `_currentChunkSize`.

ogoffart: You want predictedGoodSize to be a an integer.
(actually, it should be an amount of bytes, i was thinking even to add a new class that represent the byte unit, later)

maybe i could use parentheses around the `*` to make it explicit that the multiplication is done first.  And i could use quint64 instead of auto.
Would that help ?

ckamm: Yes, I think parenthesis and quint64 would make it clear to the reader that this is probably not a bug. Or add a comment.

PR #6334, opened 2018-01-24
owncloudcmd: Do not do the capability call when --nonshib is passed

ogoffart:
https://github.com/owncloud/enterprise/issues/2391

### DIFF
--- a/src/cmd/cmd.cpp
+++ b/src/cmd/cmd.cpp
@@ -455,24 +455,30 @@ int main(int argc, char **argv)
     account->setCredentials(cred);
     account->setSslErrorHandler(sslErrorHandler);
 
-    //obtain capabilities using event loop
-    QEventLoop loop;
+    // Perform a call to get the capabilities.
+    if (!options.nonShib) {
+        // Do not do it if '--nonshib' was passed. This mean we should only connect to the 'nonshib'
+        // dav endpoint. Since we do not get the capabilities, in that case, this has the additional
+        // side effect that chunking-ng will be disabled. (because otherwise it would use the new
+        // 'dav' endpoint instead of the nonshib one (which still use the old chunking)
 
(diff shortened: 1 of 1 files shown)


ogoffart:
This means that chunking-ng and all other new features advertised from the server will be disabled.
It is good that we disable chunking-ng because the nonshib endpoints does not support it.
The other feature that we will not enable is checksum.

Considering that shibboleth will be phased down in the long term, i think it is ok if new features are disabled for owncloudcmd

SamuAlfageme:
Working flawlessly 🎉 

> Considering that shibboleth will be phased down in the long term, I think it is ok if new features are disabled for owncloudcmd 

Agreed; also considering:

- How residual this option is and 
- Decommissioning the old WebDAV endpoint in the server will take way longer than the support drop for shibboleth (hopefully 🤞 )

Using `/remote.php/webdav` as default makes complete sense.

ogoffart:
> Working flawlessly ￼

@SamuAlfageme  You mean that with this patch fix the issue that it would go to the wrong endpoint? And that without the patch it breaks?  
(I'm asking because i realised that if the shibboleth server filter the capability call anyway, we'd not enable chunking-ng)

But maybe some instances don't do a redirect on the capability call?

SamuAlfageme:
@ogoffart I meant it takes the conservative approach by assuming there's no chunking-ng support (I don't think the [`nonshib-webdav`](https://doc.owncloud.com/server/10.0/admin_manual/enterprise/user_management/user_auth_shibboleth.html#webdav-support) ever got support for it)

Whether the client syncs or not wo/ the patch depends on **how the server was set up** in terms of what's returned when querying the capabilities - empty response vs. redirection, etc. (see https://github.com/owncloud/enterprise/issues/2391#issuecomment-360731292)

PR #6336, opened 2018-01-25
Move to trash: documentation

Mc-Pain:


ogoffart:
You should mention it only works on linux.

ckamm:
Thanks!

PR #6347, opened 2018-02-05
Themable startup option

DeepDiver1975:


SamuAlfageme:
@DeepDiver1975 looks like https://github.com/owncloud/client/pull/6347/commits/aecf42a46af701853a5eb31524c51cb07576d930 messed with tabs/spacing; could you check & adjust?

DeepDiver1975:
> @DeepDiver1975 looks like aecf42a messed with tabs/spacing; could you check & adjust?

god damn - looks like my qtcreator is not properly setup? grrrr

src/gui/application.cpp
```diff
@@ -336,7 +344,7 @@ void Application::slotownCloudWizardDone(int res)
 #ifdef Q_OS_MAC
         // Don't auto start when not being 'installed'
         shouldSetAutoStart = shouldSetAutoStart
-            && QCoreApplication::applicationDirPath().startsWith("/Applications/");
+                             && QCoreApplication::applicationDirPath().startsWith("/Applications/");
 #endif
         Utility::setLaunchOnStartup(_theme->appName(), _theme->appNameGUI(), shouldSetAutoStart);
```
guruz: Did you see this invocation?

DeepDiver1975: ❓ ❔ 😕

DeepDiver1975: obviously not ..... crawling back into my cave

ogoffart: Note that this DISABLE the auto start when setting up a second account.

DeepDiver1975: So if I get this right: as soon as one account is setup auto launch is active.
why is it disabled with the second account ? THX

ogoffart: maybe that's a bug.  @guruz  why do you disable auto-launch with multi account

ckamm: It looks like a mistake: the commit message of fc7aaf792b26c22adaef6997bb18c58fe2c18f94 suggests it was intended to not adjust the autostart flag when further accounts are added.

ckamm:
@DeepDiver1975 I've removed the whitespace changes and made other adjustments. Is the new `Theme` option still desired for 2.4.1? I've changed its default to `true`...

DeepDiver1975:
@ckamm THX for taking care! The original idea to make this themable was based on my assumption that it is desired default behavior to NOT enables startup.

I was looking into an option to have influence on this default - the theme was my idea.

With #6367 I have no need for this themable change anymore. THX

ckamm:
Great, then let's not add the option.

Issue #6322, opened 2018-01-20
Client just deleted local files, even though they also exist on the server


sagamusix:
### Expected behaviour
The client should never delete files for no good reason.

### Actual behaviour
I had a tough time synchronizing big ownCloud folders (which is already being discussed e.g. in #5017), often resulting in 503s and timeouts during the sync and thus restarting the sync process.
After some days the folder was mostly synchornized at last, as I could observe on the web interface.
However, all of a sudden the client decided that it should **delete** the local folder, even though it was still present on the server as well! I got a message telling me that there is a big folder that won't be synchronized by default, I chose to synchronize everything but by that time the local files were already gone...

### Steps to reproduce
I'm not really sure.

### Client configuration
Client version: Version 2.5.0-nightly20180110 (build 8967).

Operating system: Windows 7 x64

OS language: German

Installation path of client: Default

ogoffart:
Start the client with  `--logdebug --logfile C:\log.txt`  To start the client will full debug output into a file.

Select the folder in the settings dialog and press the secret "Ctrl+F6" shortcut which will redo a sync with full remote discovery.  Does the deleted file re-appear?

Go in the log and search for filename of files that are deleted and send us the lines around that.

My suspission is that the server did not report the files for some reason, doing CTRL+F6 should force the client to read everything again from the server.
Either we will see the file reaparing, or we will see in the log why the client thinks the files are not there.

sagamusix:
After copying the files back to their original location, the bug didn't happen anymore, even after a enforcing a full remote discovery.

guruz:
@sagamusix Which exact server version are you using?
Can you grep your server access log for the respective path?

sagamusix:
10.0.4, but probably 10.0.3 at that time. I have pruned my logs since they grew to more than 6GB, so I'm afraid they are gone now.

guruz:
@sagamusix But you don't know if in this special case the server sent something unexpected, right?
The 503s you spoke about were on previous occasions?

sagamusix:
As far as I can tell, yes, there was no direct connection between the two.

sagamusix:
Coincidentally, this just happened **again** today. This time, it was even worse, though, because it deleted a **hardlink** (which it would normally not touch at all), and apparently, as a consequence, even deleted the contents of the linked-to folder. I cannot imagine how that is even possible.

sagamusix:
[...]

PR #6349, opened 2018-02-07
Don't delete contents behind directory junctions

ckamm:
QFileInfo::isSymLink() does detect reparse points that are symlinks but
returns false for junctions. The new function FileSystem::isJunction()
can detect those and is used to not recursively delete files inside
directories that are junctions.

See  #6322

### DIFF
--- a/src/common/filesystembase.cpp
+++ b/src/common/filesystembase.cpp
@@ -478,4 +478,22 @@ bool FileSystem::isLnkFile(const QString &filename)
     return filename.endsWith(".lnk");
 }
 
+bool FileSystem::isJunction(const QString &filename)
+{
+#ifdef Q_OS_WIN
+    WIN32_FIND_DATA findData;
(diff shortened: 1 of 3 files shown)


src/common/filesystembase.cpp
```diff
@@ -478,4 +478,20 @@ bool FileSystem::isLnkFile(const QString &filename)
     return filename.endsWith(".lnk");
 }
 
+bool FileSystem::isJunction(const QString &filename)
+{
+#ifdef Q_OS_WIN
+    WIN32_FIND_DATA findData;
+    HANDLE hFind = FindFirstFileEx((const wchar_t *)filename.utf16(), FindExInfoBasic, &findData, FindExSearchNameMatch, NULL, 0);
+    if (hFind != INVALID_HANDLE_VALUE)
+        FindClose(hFind);
```
ogoffart: and return false?  
I'm just afreak the findData is left uninitialized. (UB)

ckamm: Yes, seems safer that way.

PR #6350, opened 2018-02-07
SyncJournal: Don't use LIKE for invalidating etags

ckamm:
We essentially did "?1 LIKE path" which meant that underscores in path
would be used as wildpath characters in the comparison.

Maybe related to #6322

### DIFF
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -942,9 +942,10 @@ bool SyncJournalDb::setFileRecord(const SyncJournalFileRecord &_record)
 
     if (!_avoidReadFromDbOnNextSyncFilter.isEmpty()) {
         // If we are a directory that should not be read from db next time, don't write the etag
+        QByteArray prefix = record._path + "/";
         foreach (const QByteArray &it, _avoidReadFromDbOnNextSyncFilter) {
-            if (it.startsWith(record._path)) {
-                qCInfo(lcDb) << "Filtered writing the etag of" << record._path << "because it is a prefix of" << it;
+            if (it.startsWith(prefix)) {
+                qCInfo(lcDb) << "Filtered writing the etag of" << prefix << "because it is a prefix of" << it;
                 record._etag = "_invalid_";
                 break;
             }
@@ -1815,8 +1816,10 @@ void SyncJournalDb::avoidReadFromDbOnNextSync(const QByteArray &fileName)
     query.bindValue(1, argument);
     query.exec();
 
-    // Prevent future overwrite of the etag for this sync
-    _avoidReadFromDbOnNextSyncFilter.append(fileName);
+    // Prevent future overwrite of the etags of this folder and all
+    // parent folders for this sync
+    argument.append('/');
+    _avoidReadFromDbOnNextSyncFilter.append(argument);
 }
 
 void SyncJournalDb::forceRemoteDiscoveryNextSync()
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -32,6 +32,13 @@
 
 #include "common/c_jhash.h"
 
+// SQL expression to check whether path.startswith(prefix + '/')
+// Note: '/' + 1 == '0'
+#define IS_PREFIX_PATH_OF(prefix, path) \
+    "(" path " > (" prefix "||'/') AND " path " < (" prefix "||'0'))"
+#define IS_PREFIX_PATH_OR_EQUAL(prefix, path) \
+    "(" path " == " prefix " OR " IS_PREFIX_PATH_OF(prefix, path) ")"
+
 namespace OCC {
 
 Q_LOGGING_CATEGORY(lcDb, "sync.database", QtInfoMsg)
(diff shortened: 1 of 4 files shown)


test/testsyncjournaldb.cpp
```diff
@@ -183,6 +183,58 @@ private slots:
...
+
+        makeEntry("foodir", 2);
+        makeEntry("otherdir", 2);
+        makeEntry("foodir/file", 0);
+        makeEntry("foodir/subdir", 2);
+        makeEntry("foodir/subdir/file", 0);
+        makeEntry("foodir/otherdir", 2);
+        makeEntry("foodir/subdir/subsubdir", 2);
+        makeEntry("foodir/subdir/subsubdir/file", 0);
+        makeEntry("foodir/subdir/otherdir", 2);
```
ogoffart: Was this test failling without the patch?

The test could folders file with _ and %

src/common/syncjournaldb.cpp
```diff
@@ -1805,7 +1805,8 @@ void SyncJournalDb::avoidReadFromDbOnNextSync(const QByteArray &fileName)
     SqlQuery query(_db);
     // This query will match entries for which the path is a prefix of fileName
     // Note: CSYNC_FTW_TYPE_DIR == 2
-    query.prepare("UPDATE metadata SET md5='_invalid_' WHERE ?1 LIKE(path||'/%') AND type == 2;");
+    // Note: '/' + 1 == '0'
+    query.prepare("UPDATE metadata SET md5='_invalid_' WHERE ?1 > (path||'/') AND ?1 < (path||'0') AND type == 2;");
```
ogoffart: since this WHERE clause is used twice in the file, i would consider using a macro for it like the the SELECT clause

src/common/syncjournaldb.cpp
```diff
@@ -1780,9 +1788,9 @@ void SyncJournalDb::avoidRenamesOnNextSync(const QByteArray &path)
     }
 
     SqlQuery query(_db);
-    query.prepare("UPDATE metadata SET fileid = '', inode = '0' WHERE path == ?1 OR path LIKE(?2||'/%')");
+    query.prepare("UPDATE metadata SET fileid = '', inode = '0' WHERE path == ?1 OR path LIKE(?2||'/%') ESCAPE '\\'");
```
ogoffart: Why not use the same comparison way like the other ones?

guruz:
```
13: Totals: 25 passed, 1 failed, 0 skipped, 0 blacklisted
13: ********* Finished testing of TestSyncEngine *********
13/22 Test #13: SyncEngineTest ...................***Failed    1.56 sec
```

ckamm:
Improved now.

src/common/syncjournaldb.cpp
```diff
@@ -935,10 +942,9 @@ bool SyncJournalDb::setFileRecord(const SyncJournalFileRecord &_record)
 
     if (!_avoidReadFromDbOnNextSyncFilter.isEmpty()) {
         // If we are a directory that should not be read from db next time, don't write the etag
-        QByteArray prefix = record._path + "/";
         foreach (const QByteArray &it, _avoidReadFromDbOnNextSyncFilter) {
-            if (it.startsWith(prefix)) {
-                qCInfo(lcDb) << "Filtered writing the etag of" << prefix << "because it is a prefix of" << it;
+            if (it.startsWith(record._path)) {
```
ogoffart: This change looks supicious.

So, if `foo/bar/something` is in _avoidReadFromDbOnNextSyncFilter,  that means that we would store the  invalid etag for `foo/b`.

I guess this problem would stay asymptomatic because `foo` normally should have an invalid etag. However, i'd rather not rely on that and have it behave properly as this could otherwise bite us later when the code evolve.

test/testsyncengine.cpp
```diff
@@ -172,6 +172,14 @@ private slots:
...
                                                                 {"parentFolder/subFolderA/"});
         fakeFolder.syncEngine().journal()->avoidReadFromDbOnNextSync(QByteArrayLiteral("parentFolder/subFolderA/"));
+        auto getEtag = [&](const QByteArray &file) {
+            SyncJournalFileRecord rec;
+            fakeFolder.syncJournal().getFileRecord(file, &rec);
+            return rec._etag;
+        };
+        QVERIFY(getEtag("parentFolder") == "_invalid_");
+        QVERIFY(getEtag("parentFolder/subFolderA") == "_invalid_");
+        QVERIFY(getEtag("parentFolder/subFolderA/subsubFolder") != "_invalid_");
```
ogoffart: could we create a `parentFolder/sub` and check the etag is not invalid?
(this will not fail in this case, but it is good to have such tests)

src/common/syncjournaldb.cpp
ckamm: Indeed and that was not my intention at all. Instead, I should have made sure `_avoidReadFromDbOnNextSyncFilter` entries always have a trailing slash.

test/testsyncengine.cpp
ckamm: Agree, will add to the syncjournaldb tests.

ckamm:
Squashed and merged manually.

PR #6353, opened 2018-02-09
HTTP2: Only from Qt 5.9.4

guruz:


src/libsync/connectionvalidator.cpp
```diff
@@ -315,7 +315,7 @@ bool ConnectionValidator::setAndCheckServerVersion(const QString &version)
     // We attempt to work with servers >= 5.0.0 but warn users.
     // Check usages of Account::serverVersionUnsupported() for details.
 
-#if QT_VERSION >= QT_VERSION_CHECK(5, 9, 0)
+#if QT_VERSION >= QT_VERSION_CHECK(5, 10, 0)
```
ogoffart: this don't need to change.

src/libsync/accessmanager.cpp
```diff
@@ -93,7 +93,7 @@ QNetworkReply *AccessManager::createRequest(QNetworkAccessManager::Operation op,
     qInfo(lcAccessManager) << op << verb << newRequest.url().toString() << "has X-Request-ID" << requestId;
     newRequest.setRawHeader("X-Request-ID", requestId);
 
-#if QT_VERSION >= QT_VERSION_CHECK(5, 9, 0)
+#if QT_VERSION >= QT_VERSION_CHECK(5, 10, 0)
```
ogoffart: The comment next line needs an update.
Also point out which bugs exactly we are working around?

src/libsync/connectionvalidator.cpp
guruz: Hm yes because we'd not request HTTP2. But shouldnt we align both ifdefs for consitency?

guruz:
@ogoffart Commit updated

src/libsync/accessmanager.cpp
```diff
@@ -93,8 +93,8 @@ QNetworkReply *AccessManager::createRequest(QNetworkAccessManager::Operation op,
     qInfo(lcAccessManager) << op << verb << newRequest.url().toString() << "has X-Request-ID" << requestId;
     newRequest.setRawHeader("X-Request-ID", requestId);
 
-#if QT_VERSION >= QT_VERSION_CHECK(5, 9, 0)
-    // only enable HTTP2 with Qt 5.9 because Qt 5.8.0 has too many bugs
+#if QT_VERSION >= QT_VERSION_CHECK(5, 9, 4)
+    // only enable HTTP2 with Qt 5.9.4 because old Qt have too many bugs (e.g. QTBUG-64359 is fixed in >= Qt 5.9.4)
```
ogoffart: Remove the next line which is only about 5.8

ChangeLog
```diff
@@ -11,6 +11,7 @@ version 2.4.1 (2017-02-xx)
 * Avatars: Use old location for servers <10 (#6279)
 * Link shares: Change default share name (#6298)
 * Nautilus integration: Work with python2 and python3
+* HTTP2: Only allow with Qt 5.10.0 (#6285)
```
ogoffart: 5.9.4

src/libsync/connectionvalidator.cpp
```diff
@@ -315,8 +315,9 @@ bool ConnectionValidator::setAndCheckServerVersion(const QString &version)
     // We attempt to work with servers >= 5.0.0 but warn users.
     // Check usages of Account::serverVersionUnsupported() for details.
 
-#if QT_VERSION >= QT_VERSION_CHECK(5, 9, 0)
+#if QT_VERSION >= QT_VERSION_CHECK(5, 9, 4)
     // Record that the server supports HTTP/2
+    // QTBUG-64359 is fixed in >= Qt 5.9.4
```
ogoffart: This comment does not make sense here.
This code snippet is only there to enable the propagator to have more parallelism if the connection is HTTP/2,   but it does not enable or disable HTTP/2 or workaround any qtbug.

The #if is there only because so it compiles because  QNetworkRequest::HTTP2WasUsedAttribute is a 5.9 feature.  (and therefore the #if check should stay 5.9.0 regardless of any other checks that enable HTTP/2).

guruz:
@ogoffart updated

PR #6362, opened 2018-02-19
Fix appveyor

dschmidt:
Resubmitting the app veyor fixes without the qt 5.10 version bump as we don't have the right binaries in craft cache yet and building it ourselves takes too long for appveyor.


ogoffart:
yay \o/   please merge

PR #6363, opened 2018-02-19
SyncEngine: Use separate state for two unicode conversions

ckamm:
It's unlikely that there were interaction effects but still seems safer
to use different conversion state for the path and rename_path.


ckamm:
It was a good opportunity to make the thing easier to read too. :)

Issue #6325, opened 2018-01-22
Conflict file names should include user name if uploaded


ckamm:
In master the client can optionally upload conflict files. In this case it's easy for there to be several conflict files uploaded to the server by different users. In this situation it is very useful if it's immediately obvious which user created which conflict file. This should become part of the conflict file name:
```
foo-conflict_20180115-1412.txt
foo-conflict_20180112-1100.txt
vs
foo-conflict_frank_20180115-1412.txt
foo-conflict_greta_20180112-1100.txt
```

ckamm:
@PVince81 @DeepDiver1975 I'm planning to use the dav username, that is the string used in `/remote.php/dav/files/<user>/` inside generated conflict files. Would there be a better choice?

SamuAlfageme:
@ckamm I think that'd result in a UUID on LDAP setups (for the same reason we still have troubles with things like https://github.com/owncloud/oauth2/issues/109#issuecomment-357663934) 😕

ckamm:
@SamuAlfageme Great catch, so we need to be able to determine the "display user" first!

SamuAlfageme:
@ckamm since it's gonna be used for a filename (whitespace-safe) you can get the current logged in display name from the [User Provisioning API](https://doc.owncloud.org/server/10.0/admin_manual/configuration/user/user_provisioning_api.html) i.e.`/ocs/v1.php/cloud/user`

ckamm:
@SamuAlfageme I had edited my comment a bit. The display name can be long and have spaces though...

hodyroff:
Hm, really? This lands in the versions, right? The versions view should have the ability to list the user ... not sure changing the filename is a great idea ... think about long file names which will go out of bound quickly ...

ckamm:
@hodyroff No, as yet this does not interact with the server's versions functionality. See https://github.com/owncloud/platform/issues/159 for a detailed description of the conflict-uploading feature.

PVince81:
> The versions view should have the ability to list the user 

The server doesn't store the user id of who modified a file. Would require a new column in oc_filecache or better: move the versions out of oc_filecache into a separate table where we can store more information. See https://github.com/owncloud/core/issues/2263

guruz:
> Great catch, so we need to determine the "display user" first! 

[...]

PR #6364, opened 2018-02-19
Conflicts: Add user name to conflict file name

ckamm:
For the case of uploading conflict files only.

See #6325. Possible issue is too-long-path on windows.

### DIFF
--- a/src/common/utility.cpp
+++ b/src/common/utility.cpp
@@ -542,7 +542,8 @@ QUrl Utility::concatUrlPath(const QUrl &url, const QString &concatPath,
     return tmpUrl;
 }
 
-QString Utility::makeConflictFileName(const QString &fn, const QDateTime &dt)
+QString Utility::makeConflictFileName(
+    const QString &fn, const QDateTime &dt, const QString &user)
 {
     QString conflictFileName(fn);
     // Add _conflict-XXXX  before the extension.
@@ -551,9 +552,15 @@ QString Utility::makeConflictFileName(const QString &fn, const QDateTime &dt)
     if (dotLocation <= conflictFileName.lastIndexOf('/') + 1) {
         dotLocation = conflictFileName.size();
     }
-    QString timeString = dt.toString("yyyyMMdd-hhmmss");
 
-    conflictFileName.insert(dotLocation, "_conflict-" + timeString);
+    QString conflictMarker = QStringLiteral("_conflict-");
+    if (!user.isEmpty()) {
+        conflictMarker.append(sanitizeForFileName(user));
+        conflictMarker.append('-');
+    }
+    conflictMarker.append(dt.toString("yyyyMMdd-hhmmss"));
+
+    conflictFileName.insert(dotLocation, conflictMarker);
     return conflictFileName;
 }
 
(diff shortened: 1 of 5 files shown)


ogoffart:
Btw, Jan has been insisting in the past that the filename should not contains these wierd '_' and '-' that nobody understand.   Instead, it should maybe look like this more user friendly `" (Conflict UserName 2018-02-17 14.53.20)"`  name

Compare:

```
Readme  (Conflict OlivierGoffart 2018-02-17 14.53.20).txt
Readme_conflict-OlivierGoffart-20180217145320.txt
```

I guess now is time to do this change.

src/common/utility.cpp
```diff
@@ -595,4 +602,12 @@ QByteArray Utility::conflictFileBaseName(const QByteArray &conflictName)
     return "";
 }
 
+QString Utility::sanitizeForFileName(const QString &name)
+{
+    static QRegularExpression regex(QStringLiteral("[\x00-\x1f\x80-\x9f/?<>\\\\:*|\"]"));
```
ogoffart:
```
QString invalid("?<>\\:*|...") 
... std::remove_if(..., [&](auto c) { return !c.isPrint() || invalid.contains(c); }  ) ...
```

test/testutility.cpp
```diff
@@ -187,7 +187,28 @@ private slots:
         qunsetenv("OWNCLOUD_TEST_CASE_PRESERVING");
     }
 
+    void testSanitizeForFileName_data()
+    {
+        QTest::addColumn<QString>("input");
+        QTest::addColumn<QString>("output");
+
+        QTest::newRow("")
```
ogoffart: minor: maybe give title to the data

ckamm:
The more user-friendly conflict tags are also what dropbox does. Let's make a different ticket about changing our tags though.

src/common/utility.cpp
ckamm: That would also drop compositing unicode codepoints as far as I can tell.

src/common/utility.cpp
```diff
@@ -36,7 +36,7 @@
 #include <QStandardPaths>
 #include <QCollator>
 #include <QSysInfo>
-
+#include <QRegularExpression>
```
ogoffart: (leftover?)

src/common/utility.cpp
```diff
@@ -595,4 +602,19 @@ QByteArray Utility::conflictFileBaseName(const QByteArray &conflictName)
     return "";
 }
 
+QString Utility::sanitizeForFileName(const QString &name)
+{
+    const auto invalid = QStringLiteral("/?<>\\:*|\"");
+    QString result;
+    result.reserve(name.size());
+    for (const auto c : name) {
```
ogoffart: you did not like the use of std::remove_if?

ckamm: Yes, I didn't like it at all. I thought `remove_if` followed by `truncate(distance(begin, newend))` looked way worse than this

ogoffart:
Ah, this seems to have broken the tests:

15: FAIL!  : TestSyncConflict::testUploadAfterDownload() 'conflictMap[a1FileId].contains(QString("-%1-").arg(fakeFolder.syncEngine().account()->davUser()))' returned FALSE. ()

ckamm:
Noooo. Will fix now.

Issue #4424, opened 2016-02-02
Mac client does not detect changes in local files


josemaX:
I have a Mac (OSX: 10.11.3, Clients tested 2.1.0, 2.1.1-Beta1 and 2.1.1-RC1) that does not detect when a local file has changed to update it at server (OC 8.2.2 Linux, CentOS6 from RPM/YUM)

Other Macs works without problem and sees the changes to the same file inmediately. This only happens in a machine, but if i pause the sync and restart, then does it (also if i exit the client and start again).

I've restarted/updated the client/checked disk with first aid...

Note: Others in the same server, same network, same OS and same client, works as expected.

guruz:
Is there anything special about the local sync dir? (network drive? USB stick?)
Have you tried running the client with 
```
/Applications/owncloud.app/Contents/MacOS/owncloud --logfile - 
```
to dig out more info?

josemaX:
No, it is in the local hard drive a file in a folder in Desktop. I change it with an editor and save it. Then it not syncs.

In a few hours i'll try what you say and will update this.

josemaX:
At start it logs:

m_crashReporterChar:  /Applications/owncloud.app/Contents/MacOS/owncloud_crash_reporter
02-03 17:32:50:239 0x7fc348449510 OCC::Application::setupLogging: "################## ownCloud es_ES () 2.1.1rc1 (build 3069)"
02-03 17:32:50:239 0x7fc348449510 OCC::Application::setupTranslations: void OCC::Application::setupTranslations() Using "es" translation
02-03 17:32:50:241 0x7fc348449510 OCC::SocketApi::SocketApi: SocketApi:  server started, listening at  "9B5WD74GWJ.com.owncloud.desktopclient.socketApi"
02-03 17:32:50:242 0x7fc348449510 OCC::FolderMan::FolderMan: setting remote poll timer interval to 30000 msec
02-03 17:32:50:296 0x7fc348449510 unknown: Error receiving trust for a CA certificate
02-03 17:32:50:297 0x7fc348449510 unknown: Error receiving trust for a CA certificate
02-03 17:32:50:298 0x7fc348449510 unknown: Error receiving trust for a CA certificate
02-03 17:32:50:299 0x7fc348449510 unknown: Error receiving trust for a CA certificate
02-03 17:32:50:300 0x7fc348449510 unknown: Error receiving trust for a CA certificate
[...]

PR #6370, opened 2018-02-20
Folder: normalize the local path.

ogoffart:
We otherwise normalize all path in the C form, so we must have
the Folder's path normalized the same. Or all comparizon will fail
(such as knowing if a file from the SocketAPI or the FilesystemWatcher
are part of the folder)

Issue #4424

### DIFF
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -121,6 +121,10 @@ void Folder::checkLocalPath()
 {
     const QFileInfo fi(_definition.localPath);
     _canonicalLocalPath = fi.canonicalFilePath();
+#ifdef Q_OS_MAC
+    // Workaround QTBUG-55896  (Should be fixed in Qt 5.8)
+    _canonicalLocalPath = _canonicalLocalPath.normalized(QString::NormalizationForm_C);
+#endif
     if (_canonicalLocalPath.isEmpty()) {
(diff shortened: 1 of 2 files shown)


ogoffart:
I have not tested it, but that's the logical explaination.   
We also normalize it after cannonicalPath in OwncloudPropagator::localFileNameClash (because of QTBUG-39622 according to the comment)

ogoffart:
ping.  Anybody tested it?

ckamm:
Having read the discussion in #4424 this looks very sensible to me. Maybe the path we feed to Folder from the account or folder wizard isn't normalized correctly.

ogoffart:
I think it is just a bug in QFileInfo::canonicalFilePath Probably fixed by 49d3bb005889b9fb5a06d6aadac23ebd5b2f9f2d which is included in Qt 5.6.3 and Qt 5.8

If we upgrade Qt we don't need this patch.
But It might make sense to have it in 2.4.2

ogoffart:
I added the Qt bug in the comment

SamuAlfageme:
@ogoffart hm. 2.5.0 **must** be built with 5.10 - If we wanna keep support for older Qt (in case we want to have an alternative if something goes wild with the upgrade) I'm all in with merging here. Otherwise we should just go for 2.5 and re-unify the versions.

ogoffart:
Manually merged to 2.4 and 2.5

PR #6371, opened 2018-02-20
Utility::fsCasePreserving: remove UNIT_TESTING ifdef

ogoffart:
Since the release package will be build with unit test, we don't
want to query the env variable at every call to fsCasePreserving.
So only test the env variable at startup.
And the testutility can still change the value.

(The env variable is still used from t8.pl and maybe smashbox)

Issue #6318


ogoffart:
yeah, global initializer is not pretty, but the other option seemed worse.

Issue #6226, opened 2017-12-10
"Not synchronized" tab gets cleared when skipping local discovery


sagamusix:
### Expected behaviour
The "Not synchronized" tab contains critical information such as file conflicts. This information should be persistent until the application is closed or until it is superseded by other, similar information.

### Actual behaviour
It appears that the list of errors and warnings belonging to a specific folder is cleared the next time that folder is synchronized. While it may make sense for ignored files and such, I can no longer see previous messages regarding conflicts and other errors - despite the main page telling me that there was a conflict. If I click on that information, I just see an empty log.

### Steps to reproduce
1. Synchronize a folder.
2. Synchronize another folder.
3. Synchronize the folder from step 1 again.

### Client configuration
Client version: 2.4.0rc1

Operating system: Windows 7

OS language: German

Installation path of client: Default

ckamm:
The conflict information should be rediscovered by each the sync run. The idea was to use the data the sync run gathers to keep the conflict entries up to date. I'll attempt to reproduce this later.

ckamm:
I can reproduce the problem like this:
* produce a conflict
* notice the conflict entry on the first sync that created it
* restart the folder sync with F6
* the conflict entry is gone

ckamm:
The bug becomes visible when local discovery is skipped. There was a bug that made skipping of local discovery inconsistent. With that addressed I can also reproduce this reliably.

guruz:
@sagamusix  Do you want to try 2.5.0 alpha1? https://github.com/owncloud/client/releases

SamuAlfageme:
I see this solved now; tried both reproducers - @sagamusix care to confirm?

sagamusix:
I couldn't test it yet but will try to do so during the weekend!

guruz:
@sagamusix ping.. just if you had a chance.

sagamusix:
Sadly I haven't been running ownCloud a whole lot in the last weeks so I wasn't able to actively test this, but I remember that there was one conflict that was still being displayed after a while, so I'll assume for now that it works. Feel free to resolve.

PR #6373, opened 2018-02-21
Make sure ignored and conflict files show up in the issues tab even with partial local discovery

ckamm:
See #6226 

I still have to investigate some other errors that can be shown in the issues tab.

### DIFF
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -107,6 +107,9 @@ Folder::Folder(const FolderDefinition &definition,
     _scheduleSelfTimer.setInterval(SyncEngine::minimumFileAgeForUpload);
     connect(&_scheduleSelfTimer, &QTimer::timeout,
         this, &Folder::slotScheduleThisFolder);
+
+    connect(ProgressDispatcher::instance(), &ProgressDispatcher::folderConflicts,
+        this, &Folder::slotFolderConflicts);
 }
 
 Folder::~Folder()
@@ -963,6 +966,17 @@ void Folder::slotScheduleThisFolder()
 void Folder::slotNextSyncFullLocalDiscovery()
 {
     _timeSinceLastFullLocalDiscovery.invalidate();
+}
+
+void Folder::slotFolderConflicts(const QString &folder, const QStringList &conflictPaths)
+{
+    if (folder != _definition.alias)
+        return;
+    auto &r = _syncResult;
+
+    // If the number of conflicts is too low, adjust it upwards
+    if (conflictPaths.size() > r.numNewConflictItems() + r.numOldConflictItems())
+        r.setNumOldConflictItems(conflictPaths.size() - r.numNewConflictItems());
 }
 
 void Folder::scheduleThisFolderSoon()
--- a/src/gui/issueswidget.cpp
+++ b/src/gui/issueswidget.cpp
@@ -241,6 +241,20 @@ void IssuesWidget::slotProgressInfo(const QString &folder, const ProgressInfo &p
             return engine.shouldDiscoverLocally(path);
         });
     }
+    if (progress.status() == ProgressInfo::Done) {
+        // We keep track very well of pending conflicts.
+        // Inform other components about them.
+        QStringList conflicts;
+        auto tree = _ui->_treeWidget;
+        for (int i = 0; i < tree->topLevelItemCount(); ++i) {
+            auto item = tree->topLevelItem(i);
+            if (ProtocolItem::folderName(item) == folder
+                && ProtocolItem::status(item) == SyncFileItem::Conflict) {
+                conflicts.append(ProtocolItem::filePath(item));
+            }
+        }
+        emit ProgressDispatcher::instance()->folderConflicts(folder, conflicts);
+    }
 }
 
 void IssuesWidget::slotItemCompleted(const QString &folder, const SyncFileItemPtr &item)
--- a/src/gui/issueswidget.cpp
+++ b/src/gui/issueswidget.cpp
@@ -196,6 +196,17 @@ void IssuesWidget::addItem(QTreeWidgetItem *item)
         }
     }
 
+    // Wipe any existing message for the same folder and path
+    auto newData = ProtocolItem::extraData(item);
+    for (int i = 0; i < count; ++i) {
+        auto otherItem = _ui->_treeWidget->topLevelItem(i);
+        auto otherData = ProtocolItem::extraData(otherItem);
+        if (otherData.path == newData.path && otherData.folderName == newData.folderName) {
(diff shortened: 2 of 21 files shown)


ckamm:
The second commit is gratuitous, I didn't need that after all. I'd be ok with dropping it.

src/gui/issueswidget.cpp
```diff
@@ -209,9 +214,35 @@ void IssuesWidget::slotOpenFile(QTreeWidgetItem *item, int)
...
+        cleanItems([&](QTreeWidgetItem *item) {
+            if (ProtocolItem::folderName(item) != folder)
+                return false;
+            if (style == LocalDiscoveryStyle::FilesystemOnly)
+                return true;
+            if (!persistsUntilLocalDiscovery(item))
+                return true;
+
+            // Definitely wipe the entry if the file no longer exists
+            if (!QFileInfo(f->path() + ProtocolItem::filePath(item)).exists())
```
ogoffart: but maybe it only exists on the server!  (ignored on the server, or an error while downloading)

Anyway i don't think this is required.  If the file was deleted, this should trigger the filesystem watcher there anyway, wouldn't it?

src/gui/issueswidget.cpp
```diff
@@ -145,7 +146,13 @@ void IssuesWidget::hideEvent(QHideEvent *ev)
     QWidget::hideEvent(ev);
 }
 
-void IssuesWidget::cleanItems(const QString &folder)
+static bool persistsUntilLocalDiscovery(QTreeWidgetItem *item)
+{
+    const auto status = ProtocolItem::status(item);
+    return status == SyncFileItem::Conflict || status == SyncFileItem::FileIgnored;
```
ogoffart: what about blacklisted errors?

I guess everything should persist.

src/gui/folder.cpp
```diff
@@ -816,6 +819,7 @@ void Folder::slotSyncFinished(bool success)
     }
     _previousLocalDiscoveryPaths.clear();
 
+
```
ogoffart: !

src/gui/issueswidget.cpp
ckamm: Yes, this check should be superfluous, but I was concerned about bugs that would make conflict file items stick around forever and adding the "does the file exist" check seemed like a good sanity check. But yes, the ignore could be remote and we have no way to tell. I'll remove it.

src/gui/issueswidget.cpp
ckamm: The concern was about error items that could appear even without a local discovery. Like a download that failed. But now I know that the same thing might happen for `FileIgnored` meaning I need a new approach.

How about I make sure that items completing successfully also wipe issue items and that there can only be one issue item per file path? I think we need to think about issue item lifetime some more...

ckamm: After more testing it became clear that the current state is also broken. If you have a remote file `foo/remote-ignore.crdownload` then this will be listed as ignored on the initial sync. But the entry is wiped in a follow up sync and does not reappear unless the folder `foo` is rediscovered.

The only correct behavior that I see is to store per entry information about when it should be wiped. `FileIgnored` would depend either on local or remote discovery (could use the `direction` tag to propagate that information).

For all issues that aren't `FileIgnored` or `Conflict`, I'd stick to the behavior of invalidating them when a new sync propagation starts - an invalidation bug that makes scary looking errors stay around when they are no longer applicable would be bad.

ckamm:
@ogoffart What do you think about this now?

src/gui/issueswidget.cpp
```diff
@@ -196,6 +196,17 @@ void IssuesWidget::addItem(QTreeWidgetItem *item)
         }
     }
 
+    // Wipe any existing message for the same folder and path
+    auto newData = ProtocolItem::extraData(item);
+    for (int i = 0; i < count; ++i) {
```
ogoffart: hopefully this does not ruins the performence when adding many items.
This looks like O(n²) complexity. anything we can do about that?

Maybe storing a QSet of existing items so we only do that for duplicate items?

src/gui/issueswidget.cpp
```diff
@@ -209,9 +226,47 @@ void IssuesWidget::slotOpenFile(QTreeWidgetItem *item, int)
...
+        cleanItems([&](QTreeWidgetItem *item) {
+            if (ProtocolItem::extraData(item).folderName != folder)
+                return false;
+            if (style == LocalDiscoveryStyle::FilesystemOnly)
+                return true;
+            if (!persistsUntilLocalDiscovery(item))
+                return true;
+
+            // Definitely wipe the entry if the file no longer exists
+            if (!QFileInfo(f->path() + ProtocolItem::extraData(item).path).exists())
```
ogoffart: I still think this might be wrong.

src/gui/issueswidget.cpp
ckamm: Yes, we've had performance problems here before after all. I think this should be fine though since we're log n faster than last time and also limit n to 50k. 

Nevertheless, I had rejected adding a set because I thought it'd be too costly memory wise - but was just a mistake (as long as we don't copy any string data). Then the only remaining drawback is that adding a set means keeping the two in sync which shouldn't be an obstacle.

src/gui/issueswidget.cpp
ckamm: Can you elaborate again? Now that this covers only `FileIgnored` and `Conflict` statuses that are local this seemed like a correct safeguard to me.

ogoffart: So won't happen for errors  (eg, download error).
And file ignored on the server are already not in there?

Then i guess it is ok.

ckamm: Yes, exactly. This only applies to items that can stay around until a folder is locally rediscovered: currently only local ignores and conflicts.

src/gui/issueswidget.cpp
```diff
@@ -162,8 +167,10 @@ void IssuesWidget::cleanItems(const std::function<bool(QTreeWidgetItem *)> &shou
     int itemCnt = _ui->_treeWidget->topLevelItemCount();
     for (int cnt = itemCnt - 1; cnt >= 0; cnt--) {
         QTreeWidgetItem *item = _ui->_treeWidget->topLevelItem(cnt);
-        if (shouldDelete(item))
+        if (shouldDelete(item)) {
+            _pathsWithIssues.remove(pathsWithIssuesKey(ProtocolItem::extraData(item)));
             delete item;
+        }
```
ogoffart: thinking about it...  this also look like a O(n²) if "delete item" is O(n)

Issue #6222, opened 2017-12-07
Produce a notification when a new folder matching the selective sync list contents appears in the FS


SamuAlfageme:
This is by request of the mighty @cdamken 

### Steps to reproduce:

1. Uncheck a folder on your selective sync list
2. Create a folder matching that folder name
3. An entry about the ignored item because of the "choose what to sync" blacklist appears in the activity tab. 

### Expected Result 

A notification should be issued since the client is ignoring the user actions in that folder and might call for resolution in the selective sync list.

ckamm:
So the suggestion is that when the user creates "foo/" when that path was excluded through the selective sync functionality: emit a notification about it.

SamuAlfageme:
@ckamm exactly that, yeah. So the user is aware of any changes made locally that are currently not being synced because our internal logic.

Scenario: the user in the past didn't select the "Photos" to sync and he forgot. Turns out that he created now a new local "Photos" folder and it's not being synced, the user doesn't know why until he takes a look into the Activity > "Not Synced" tab.

Issue #5493, opened 2017-01-26
[Notifications] Group similar notifications received together instead of showing them in batch


SamuAlfageme:
Similarly as what happens when multiple files are erased at the same time and the client only displays one notification saying: "N files were deleted", when there's multiple equal events received together, they should be grouped into one notification. 

### Scenarios:
* When multiple folder's sync needs to be confirmed in the account tab (see https://github.com/owncloud/client/pull/5340#issuecomment-274878023)

![](https://cloud.githubusercontent.com/assets/2644445/22258599/82d0fd86-e262-11e6-8f98-9c65e451544c.png)

PR #6377, opened 2018-02-23
Notify if an explicitly excluded folder is created

ckamm:
See #6222 
@SamuAlfageme

### DIFF
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -488,6 +488,8 @@ void Folder::slotWatchedPathChanged(const QString &path)
         return; // probably a spurious notification
     }
 
+    warnOnNewExcludedItem(record, relativePath);
+
     emit watchedFileChangedExternally(path);
 
     // Also schedule this folder for a sync, but only after some delay:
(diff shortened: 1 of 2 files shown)


ogoffart:
Don't we get also this notification when subfolders or files within the folder are removed. In that case, it would be misleading.

We could check if it is in the DB or not maybe, to avoid such false warning.

src/gui/folder.cpp
```diff
@@ -965,6 +967,28 @@ void Folder::slotNextSyncFullLocalDiscovery()
     _timeSinceLastFullLocalDiscovery.invalidate();
 }
 
+void Folder::warnOnNewExcludedFolder(const QStringRef &path)
+{
+    // Note: This assumes we're getting file watcher notifications
+    // for folders only on creation and deletion.
+    QFileInfo fi(_canonicalLocalPath + path);
+    if (!fi.isDir())
```
ogoffart: A file would have the same problem.

ckamm:
@ogoffart At least on linux our removal notifications point to the full path that was removed, not the parent folder. I'm up for adding the db check as a safeguard anyway - possibly the win or osx watchers behave differently.

I used the `isDir` check to limit the number of cases in which we'd construct the selective sync exclude list from the database since I was worrying that doing this for each notification might create a new bottleneck. What's your intuition regarding performance?

ogoffart:
Looks good to me from the code point of view.
But this should be tested to find out if there are not too many unwanted warnings.

Is this warning really necessary when in any case it should show an error in the ui?

ckamm:
@ogoffart I agree this could use testing on osx and windows. @SamuAlfageme ?

The rationale provided for the warning makes some sense (https://github.com/owncloud/client/issues/6222#issuecomment-350246873) even if the client usually doesn't warn about everything that goes on that isn't synced. Making an exception for explicitly selective-sync excluded folders is ok with me.

ckamm:
@SamuAlfageme Rebased to allow for easy testing.

SamuAlfageme:
@ckamm awesome - I'll get to it

SamuAlfageme:
After some time using this patch, never managed to get unwanted notifications - just the ones @ckamm mentioned in https://github.com/owncloud/client/pull/6377#issuecomment-368885780 (i.e. https://github.com/owncloud/client/issues/6222) - I'll say it's safe to merge here.

As a rule of thumb I'd say we should notify (at least once) of everything that goes in the "Issues" tab. Maybe requires https://github.com/owncloud/client/issues/5493 tho.

Issue #6317, opened 2018-01-18
"All the local files deleted" -- wrong?


moscicki:
Hello,

We had a case today of a user who got this message: 

![image](https://user-images.githubusercontent.com/621965/35109870-cd5d9b68-fc77-11e7-9c23-7f48d90fd667.png)

We suspect that was triggered by a transient problem on our server. However, local files were all good and not removed. Also his files on the server were all OK. The user chose "Keep" and all was good.

The problem: this message is clearly wrong -- all local files were there. 

In order to understand how to fix this message: what exactly is the condition that triggers this message? https://github.com/owncloud/client/blob/master/src/libsync/syncengine.cpp#L1020

Is it absence of local files in the sync folder? 
Is it a PROPFIND response with xml body but no content in it? Is it a PROPFIND response with no body (0 content length)?

Any combination of the above?

ogoffart:
This happens if all the files were changed, and at least one of them is removed.
The detection of local vs. remote in that case might be bogus.

This may very well happen if the PROPFIND has no contents, yes.

ckamm:
@ogoffart We could at least have two versions of this dialog: all remote files seem to change vs all local files seem to change?

SamuAlfageme:
@ckamm there's already two versions:

https://github.com/owncloud/client/blob/ba4712c922c00681d19a48b42f3a0f248e053a07/src/gui/folder.cpp#L921-L952

What I was thinking about the other day is whether the dialog can close by itself (in case of server-deletion) when we (implement a way to and) detect it was indeed a bogus PROPFIND.

moscicki:
How can we detect bogus propfinds? As a general principle we should make sure that this message is triggered only after a positively verified reply from the server which cannot be generated by accident such as a proxy stripping down content from the response etc.

moscicki:
I can reproduce this situation with the client now. Do you want to check the client logs / tcpdump traffic to the proxy? I would need to communicate these out-of-bound as it contains real user filenames etc.

After first inspection I don't see anything particularly wrong with the propfinds...

ckamm:
@moscicki Yes please! A full log with --logdebug would probably be sufficient. mail at ckamm de.

moscicki:
I did more investigation on this. It is not that simple. I will provide you with the logs soon but first let me explain.

[...]

PR #6379, opened 2018-02-28
Discovery: Error if properties are missing

ckamm:
This is potentially too aggressive? What do you think @ogoffart?

See #6317

### DIFF
--- a/src/libsync/discoveryphase.cpp
+++ b/src/libsync/discoveryphase.cpp
@@ -300,10 +300,8 @@ void DiscoverySingleDirectoryJob::abort()
     }
 }
 
-static std::unique_ptr<csync_file_stat_t> propertyMapToFileStat(const QMap<QString, QString> &map)
+static void propertyMapToFileStat(const QMap<QString, QString> &map, csync_file_stat_t *file_stat)
 {
-    std::unique_ptr<csync_file_stat_t> file_stat(new csync_file_stat_t);
-
     for (auto it = map.constBegin(); it != map.constEnd(); ++it) {
         QString property = it.key();
         QString value = it.value();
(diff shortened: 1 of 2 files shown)


src/libsync/discoveryphase.cpp
```diff
@@ -374,12 +374,26 @@ void DiscoverySingleDirectoryJob::directoryListingIteratedSlot(QString file, con
             file = file.remove(0, 1);
         }
 
-
-        std::unique_ptr<csync_file_stat_t> file_stat(propertyMapToFileStat(map));
+        std::unique_ptr<csync_file_stat_t> file_stat(new csync_file_stat_t);
         file_stat->path = file.toUtf8();
-        if (file_stat->etag.isEmpty()) {
-            qCCritical(lcDiscovery) << "etag of" << file_stat->path << "is" << file_stat->etag << "This must not happen.";
+        file_stat->size = -1;
```
ogoffart: Should we give some default value in csync_file_stat_s

src/libsync/discoveryphase.cpp
```diff
@@ -374,12 +374,26 @@ void DiscoverySingleDirectoryJob::directoryListingIteratedSlot(QString file, con
...
+        if (file_stat->type == ItemTypeDirectory)
+            file_stat->size = 0;
+        if (file_stat->type == ItemTypeSkip
+            || file_stat->size == -1
+            || file_stat->modtime == -1
+            || file_stat->remotePerm.isNull()
+            || file_stat->etag.isEmpty()
+            || file_stat->file_id.isEmpty()) {
+            _error = tr("The server file discovery reply is missing data.");
+            qCCritical(lcDiscovery)
```
ogoffart: i think qCWarning should be enough compared to the other reason we use qcWarning

ogoffart:
The comment for `OwncloudPropagator::isInSharedDirectory`
```
// ownCloud server  < 7.0 did not had permissions
```

If we no longer support these server, it is a bunch of code we can remove.

ckamm:
Reading the fileid from PROPFIND was introduced in 4440acdb1de37ade4a8ef24cf97f8ab0915c1d33 (Oct 2013).

ckamm:
@ogoffart The current state as far as I am aware is that the client considers server versions before 7.0 as unsupported. The client shows a "Unsupported server version, proceed at your own risk" message with these. Requiring permissions support would break them. Let's consider whether to drop pre-7 support for 2.5?

ogoffart:
i agree, we probably need not to support these server. https://github.com/owncloud/client/pull/6380

PVince81:
was there a policy that says that clients must support EOL'd versions of OC, even very old ones ?

michaelstingl:
> was there a policy that says that clients must support EOL'd versions of OC, even very old ones ?

I'm not aware. Why should this be the case?

PVince81:
just to answer this: https://github.com/owncloud/client/pull/6379#issuecomment-369512277

if we don't need to support EOL'd versions of OC then these versions could be blocked completely by the client ?

ogoffart:
@PVince81  done in #6380

PR #6380, opened 2018-03-01
Increase the minimally supported server version to 7.0 and remove a bunch of code

ogoffart:


src/libsync/account.cpp
```diff
@@ -463,7 +463,7 @@ bool Account::serverVersionUnsupported() const
         // not detected yet, assume it is fine.
         return false;
     }
-    return serverVersionInt() < makeServerVersion(7, 0, 0);
+    return serverVersionInt() < makeServerVersion(9, 1, 0);
```
ckamm: I agree about raising this because we do in fact not test the client against these old server versions much.

How did you chose 9.1?

ckamm:
This should also go in the "important" section of the changelog.

src/libsync/account.cpp
ogoffart: As explained in the commit message:  because it's end of life of 9.0 is reached.

ogoffart:
> There's also a pre-7 check in propagateremotemove.cpp:95

Indeed. I missed that one.

PR #6384, opened 2018-03-06
CSync Errors cleanup

ogoffart:
Remove many unused csync error codes.
Reword the remaining errors a bit to avoid confusion.
See individual commits for more details.


src/csync/csync.cpp
```diff
@@ -203,8 +203,6 @@ static int _csync_walk_tree(CSYNC *ctx, csync_s::FileMap &tree, const csync_tree
 {
     for (auto &pair : tree) {
         if (_csync_treewalk_visitor(pair.second.get(), ctx, visitor) < 0) {
-            if( ctx->status_code == CSYNC_STATUS_OK )
-                ctx->status_code = csync_errno_to_status(errno, CSYNC_STATUS_TREE_ERROR);
```
ckamm: So this could be removed because `errno` is never set during treewalk?

ogoffart: Right.

(Anyway the treewalk is anyway never supposed to fail,  we don't look at the error_code after the treewalk)

ogoffart:
I need to rebase for there is an error with the latest commit.

PR #6388, opened 2018-03-14
Ownsql: Refactor and simplify

ogoffart:
The goal here is to simplify the code to put the SQL query near its used.
This way, when adding a new query,  we only need to modify 2 places instead of 4.
And when we need to modify the query for a reason or another, the change stay localised

@ckamm: what do you think?


src/common/ownsql.h
```diff
@@ -72,8 +78,15 @@ class OCSYNC_EXPORT SqlQuery
...
+    explicit SqlQuery() = default;
     explicit SqlQuery(SqlDatabase &db);
-    explicit SqlQuery(const QString &sql, SqlDatabase &db);
+    explicit SqlQuery(const QByteArray &sql, SqlDatabase &db);
+    /**
+     * Prepare the SQLQuery if it was not prepared yet.
+     * Otherwise, clear the results and the bindings.
+     * return false if there is an error
+     */
+    bool init(const QByteArray &sql, SqlDatabase &db);
```
ckamm: This function seems really convenient but the name is confusing. `prepareOnce`? `prepareOrClearResults`?

Now there are three ways of using this
```
SqlQuery q1;
q1.init();

SqlQuery q2(db);
q2.prepare();

SqlQuery q3(sql, db);
```
Can we get rid of one of them? Or document these three modes? Also move `prepare()` up here so the relation is clear visually.

src/common/ownsql.cpp
```diff
@@ -228,28 +236,29 @@ SqlQuery::~SqlQuery()
     if (_stmt) {
         finish();
     }
+    if (_sqldb) {
+        _sqldb->_queries.remove(this);
```
ckamm: `_sqldb` can be a pointer to a deleted value here. If you use the `SqlQuery(SqlDatabase&)` constructor and never `prepare()` the query can be deleted after the database. I suspect you always want to register queries with the database, not only when `prepare()` succeeds.

src/common/ownsql.cpp
```diff
@@ -49,6 +49,12 @@ SqlDatabase::SqlDatabase()
 {
 }
 
+SqlDatabase::~SqlDatabase()
+{
+    close();
```
ckamm: I have a vague memory of us removing close-on-destruction before? I didn't find it in the revision history though and might be wrong.

src/common/ownsql.cpp
```diff
@@ -262,18 +271,26 @@ int SqlQuery::prepare(const QString &sql, bool allow_failure)
...
             ENFORCE(allow_failure, "SQLITE Prepare error");
         }
+        _sqldb->_queries.insert(this);
     }
     return _errId;
 }
 
+// There is no overloads to QByteArray::startWith that takes Qt::CaseInsensitive
+template <int N>
+static bool startsWithInsensitive(const QByteArray &a, const char (&cal)[N])
```
ckamm: Mmh, is this complication for the sake of optimization really worth it?

ckamm:
Overall I really like that the SQL query code will be closer to the use and that adding new queries will be less error prone.

src/common/ownsql.cpp
ogoffart: We already close it on destruction in the ~SyncJounrnalDb  
so this close is in fact a no-op.  I just felt it made sense in SqlDatabase itself, but i can remove it.

src/common/ownsql.cpp
ogoffart: right. I need to fix this.

src/common/ownsql.cpp
ogoffart: Where is the complication? This function looks simple to me. How would you simplify it? Would you think it is simpler if it was taking a const char* and using strlen?

src/common/ownsql.cpp
ckamm: No, then I'm wrong about this

src/common/ownsql.h
ogoffart: Maybe `initOrReset`  ?

I was thinking removing the constructor taking just a database, and making the `prepare() ` call private, but then there is even more code that would need to be adjusted.  Do you think I should?

ckamm: Yes, I think that'd make it easier to see how the query is supposed to be used.

src/common/ownsql.cpp
ckamm: Maybe it's just me being unused to passing arrays this way. I think strlen use would have surprised me less. Taking a `QByteArray` and using `QByteArrayLiteral` might be the most straightforward path?

PR #6402, opened 2018-03-24
Dolphin plugin: fall back if $XDG_RUNTIME_DIR is empty

whitequark:
As per XDG Base Directories specification:
> If $XDG_RUNTIME_DIR is not set applications should fall back
> to a replacement directory [...]

In practice, the ownCloud client has the fallback, but the plugin
helper does not, and if $XDG_RUNTIME_DIR is not set, the plugin
mysteriously does not work.


ogoffart:
Right, that's the same code as in socketapi.cpp

Thanks for the patch!

PR #6404, opened 2018-03-26
Excludes: Expand doc, Desktop.ini only in root folder

ckamm:
The Desktop.ini exclude was added for 2.4 because we add such a file to
the synced folder on Windows. It doesn't need to cover subdirectories.

See https://github.com/owncloud/documentation/issues/2832


michaelstingl:
@voroyam FYI ^

Issue #6403, opened 2018-03-24
[Linux] Client crashes on sync at startup


Goodaan:
When starting up the client it scans the folders, starts syncing, then segfaults and dies. It is on a fresh Fedora 27 install on my home laptop, syncing on a workplace server that is terribly outdated. The first sync did work and download the contents of the server, but after restarting the machine I cannot get it to work again. I don't know much about the servers config where I work. But I do know that the server is outdated.

### Expected behaviour
The client should not segfault

### Actual behaviour
It does

### Steps to reproduce
1. Start the client

### Server configuration

ownCloud version: 5.0.25

### Client configuration
Client version: 2.4.1-9083

Operating system: Fedora 27

OS language: French

Qt version used by client package (Linux only, see also Settings dialog): Qt 5.6.2, built against Qt 5.6.2

Client package (From ownCloud or distro) (Linux only): From owncloud (https://download.opensuse.org/repositories/isv:ownCloud:desktop/Fedora_27/isv:ownCloud:desktop.repo)

Installation path of client: /usr/bin; /usr/lib64

### Logs
Client logfile:
<https://gist.github.com/Goodaan/d7916277d483048b7e448881573efa48>

gdb:
```
Starting program: /usr/bin/owncloud 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
[New Thread 0x7fffdf9e9700 (LWP 5254)]
[New Thread 0x7fffdebcd700 (LWP 5255)]
[New Thread 0x7fffde2a0700 (LWP 5256)]
Detaching after fork from child process 5257.
Detaching after fork from child process 5258.
[New Thread 0x7fffdcfbc700 (LWP 5261)]
[New Thread 0x7fffcffff700 (LWP 5262)]
[New Thread 0x7fffcf7fe700 (LWP 5263)]
Detaching after fork from child process 5264.
[New Thread 0x7fffceffd700 (LWP 5267)]
[Thread 0x7fffcffff700 (LWP 5262) exited]
[New Thread 0x7fffcffff700 (LWP 5278)]
[New Thread 0x7fffce7fc700 (LWP 5301)]

Thread 1 "owncloud" received signal SIGSEGV, Segmentation fault.
0x00007ffff5a186c1 in OCC::PropagatorCompositeJob::scheduleSelfOrChild() ()
   from /opt/ownCloud/ownCloud/lib64/libowncloudsync.so.0

(gdb) info threads 
[...]

PR #6407, opened 2018-03-27
Fix bug, and simplify future use of _finished upload flag

ckamm:
Addresses #6403 crash


ogoffart:
Ah, sorry, i just did the same commit.

ckamm:
Was merged

Issue #6411, opened 2018-03-27
Blacklisted items don't prevent parent directory etags from updating


ckamm:
Currently if an item is blacklisted it will not be reattempted. Instead it's set to `Ignored` and `BlacklistedError`. The intention is that we want to show the sync as successful and silently reattempt the file after some time. (the value of the whole feature is debatable)

The problem with this approach becomes clear in a specific scenario: Say the file `A/foo.txt` can't be downloaded and gets blacklisted after the first sync. On the second sync it doesn't get attempted and the sync as well as `PropagateDirectory` for `A` is successful, updating the etag of `A`. When the blacklist entry expires the client will know longer be aware that it needs to rediscover `A` and do something inside it.

ckamm:
PR #6412

ckamm:
Keeping this open because I still want to add blacklisting tests - we've had none of them for too long.

guruz:
This bug must exist since ages?

PR #6412, opened 2018-03-27
Blacklisting must prevent parent etag updates #6411

ckamm:


ckamm:
Merged and picked to 2.4

PR #6413, opened 2018-03-28
Add a unit test for blacklist behavior

ckamm:
For #6411 and as a follow up to #6412

### DIFF
--- a/test/testblacklist.cpp
+++ b/test/testblacklist.cpp
@@ -1,0 +1,187 @@
+/*
+ *    This software is in the public domain, furnished "as is", without technical
+ *    support, and with no warranty, express or implied, as to its usefulness for
+ *    any purpose.
+ *
+ */
+
+#include <QtTest>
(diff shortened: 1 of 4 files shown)


test/testblacklist.cpp
```diff
@@ -0,0 +1,173 @@
...
+        }
+        cleanup();
+
+        // Ignored during the second run - but soft errors are also errors
+        QVERIFY(!fakeFolder.syncOnce());
+        {
+            auto it = findItem(completeSpy, "A/new");
+            QVERIFY(it);
+            QCOMPARE(it->_status, SyncFileItem::BlacklistedError);
+            QCOMPARE(it->_instruction, CSYNC_INSTRUCTION_IGNORE); // no retry happened!
```
ogoffart: Maybe there could be a counter of the amount of PUT calls in addition to that.

ckamm: Added.

Issue #6365, opened 2018-02-19
Use more user friendly conflict tags in filenames


ckamm:
Currently conflict files are tagged with `-conflict-20180220-141213`. As suggested by @ogoffart in #6364 (and also implied by dropbox) we could be using `(Conflict 2018-02-20 14.12.13)` instead.

This kind of change would need to be backwards compatible and take into account the "user name in conflict tag" change #6364.

guruz:
Moving to 2.6.0

ogoffart:
I think we should try to take the oportunity to do it in 2.5 if possible so we don't change the format of conflict file too often.

ckamm:
@ogoffart @guruz What's the exact pattern we want to use? I'm going with " (Conflict Username yyyy-mm-dd hh:mm:ss)" for now. Dropbox seems to use " (Username's conflicted copy yyyy-mm-dd)".

ogoffart:
I'd say  " (Conflict Username yyyy-mm-dd hh-mm-ss)"
Upper case for "Conflict"  and you unfortunately cannot use ':' on windows, so better not use it. (maybe using ';' instead would be better?  or just nothing: hhmmss?)

SamuAlfageme:
We might also wanna think about a future where the server might be able to handle conflict files as such (not as plain, regular files but files with some metadata associated). I'm thinking on ways to feed e.g. https://github.com/owncloud/client/issues/5896

Will it make sense to send/store information on who created a conflict, when it was created, etc. without relying on the filename? e.g. to be sent on hypothetical `OC-Conflict-Creator`, `OC-Conflict-Mtime`... headers

ckamm:
@SamuAlfageme The client already sends this kind of data to the server: https://github.com/owncloud/platform/issues/159

SamuAlfageme:
@ckamm aw, you're a man ahead of your time! 😂

SamuAlfageme:
Same as in https://github.com/owncloud/client/issues/6325#issuecomment-389922348 - closing!

jnweiger:
With 2.5.0 beta1, the naming looks like this: 
`filename (conflicted copy jwtest 2018-08-08 201107).txt`

The username "jwtest" is only present, if the uploadConflictFiles feature is enabled. It should always be present for forward compatibility when the feature gets enabled later.

BAD
When the uploadConflictFiles feature is disabled, the client will silently ignore a file that I create via
[...]

PR #6417, opened 2018-03-28
Conflicts: Change tags to be more user friendly

ckamm:
From "_conflict-user-yyyymmdd-hhmmss"
to   " (Conflict user yyyy-mm-dd hhmmss)"

See  #6365

### DIFF
--- a/src/common/utility.cpp
+++ b/src/common/utility.cpp
@@ -548,19 +548,23 @@ QString Utility::makeConflictFileName(
     const QString &fn, const QDateTime &dt, const QString &user)
 {
     QString conflictFileName(fn);
-    // Add _conflict-XXXX  before the extension.
+    // Add conflict tag before the extension.
     int dotLocation = conflictFileName.lastIndexOf('.');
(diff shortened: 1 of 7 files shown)


src/common/utility.cpp
```diff
@@ -575,33 +579,58 @@ bool Utility::isConflictFile(const char *name)
         bname = name;
     }
 
-    return std::strstr(bname, "_conflict-");
+    // Old pattern
+    if (std::strstr(bname, "_conflict-"))
+        return true;
+
+    // New pattern
+    if (std::strstr(bname, "(Conflict"))
```
guruz: what about having a file called "Romeo and Juliet (Conflicts between families).pdf"?
"Git handbook (Conflict resolution chapter).txt"?

SamuAlfageme: Actually a good point

ckamm:
I've updated the PR to use "(conflicted copy". Following dropbox's way of conflict tagging would be no problem, but need additional work since the start of the tag isn't as obvious.

PR #6421, opened 2018-04-03
Deltasync: Add growth/shrink unittest for downloads

ckamm:
Similar to the one for uploads, to test these corner cases more.

@ahmedammar

### DIFF
--- a/test/testzsync.cpp
+++ b/test/testzsync.cpp
@@ -42,12 +42,17 @@ private slots:
         SyncOptions opt;
         opt._deltaSyncEnabled = true;
         opt._deltaSyncMinFileSize = 0;
+        // Reduce the chunk size, allowing us to work with smaller data
+        int chunkSize = 2 * 1000 * 1001;
+        opt._minChunkSize = opt._maxChunkSize = opt._initialChunkSize = chunkSize;
         fakeFolder.syncEngine().setSyncOptions(opt);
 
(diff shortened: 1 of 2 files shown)


test/testzsync.cpp
```diff
@@ -42,12 +42,17 @@ private slots:
         SyncOptions opt;
         opt._deltaSyncEnabled = true;
         opt._deltaSyncMinFileSize = 0;
+        // Reduce the chunk size, allowing us to work with smaller data
+        int chunkSize = 2 * 1000 * 1001;
```
ogoffart: Note: since we build with C++14,  we can also use the group separator notation: `2'002'000`  It might actually be more readable.

test/testzsync.cpp
```diff
@@ -42,12 +42,17 @@ private slots:
         SyncOptions opt;
         opt._deltaSyncEnabled = true;
         opt._deltaSyncMinFileSize = 0;
+        // Reduce the chunk size, allowing us to work with smaller data
```
ogoffart: Fine as long as there are other test testing with the actual chunk size.

test/testzsync.cpp
guruz: might also be nice in future to name variables with a unit.
E.g. chunkSizeBytes, timeOutMsec

PR #6426, opened 2018-04-04
Theme: unify ownCloudTheme and Theme classes

ogoffart:
The goal is to avoid confusion described in issue #6422 by removing
duplicates between the Theme and owncloudTheme.

 - Use the about from ownCloudTheme everywhere
 - Create default applicationIcons() and condifFileName() that should work
   everywhere
 - trayFolderIcon was removed as it is not used
 - the helpUrl from the default Theme now points to the owncloud client
   documentation. Before there was no help entry by default for branded
   client if the function was not overriden.
 - Do not merge functions that would otherwise break compatibility with
   theme that did not override them. For example colors or customMedia.

### DIFF
--- a/src/libsync/owncloudtheme.cpp
+++ b/src/libsync/owncloudtheme.cpp
@@ -19,8 +19,6 @@
 #ifndef TOKEN_AUTH_ONLY
 #include <QPixmap>
 #include <QIcon>
-#include <QStyle>
-#include <QApplication>
 #endif
 #include <QCoreApplication>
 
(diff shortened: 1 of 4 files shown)


src/libsync/owncloudtheme.cpp
```diff
@@ -35,63 +33,13 @@ ownCloudTheme::ownCloudTheme()
...
-        //                                   "see <a href=\"https://owncloud.com\">owncloud.com</a> for more info.",
-        //                                   "Top text in setup wizard. Keep short!");
-        return QVariant();
-    } else {
-        return QVariant();
-    }
+    return QVariant();
 }
-
 #endif
```
SamuAlfageme: There's a `#ifndef`/`#endif` based on the same `TOKEN_AUTH_ONLY` wrapping this block already - will it make sense to remove this and [L59](https://github.com/owncloud/client/pull/6426/commits/b615b8ce57f32f052a3fcb43cf95334ac0dd785e#diff-5c233b67dd98ce7e8ea3ffbe0a4b3547R59) macros? Or is it semantically relevant to keep them?

ogoffart: We should keep them, this eases when we sometimes build the cmd client without Qt Gui. (altough this is not officially supported, it works)

Issue #6420, opened 2018-04-03
Persist X-Request-ID with the events that originated them 


SamuAlfageme:
`X-Request-ID` was introduced in client v. 2.4: https://github.com/owncloud/client/issues/5853 for tracing back and cross-referencing requests with the server logs. This can be seen as an extension of that to help users locate the cause of their issues together with admins. 

Would it be possible to store this ID together with its originator event on the sync journal to allow some sort of "Copy X-Request-ID` when right-clicking an event from the "Activity > Not Synced" tab? (only in those cases that are strictly related to the server operation/network activity). This way is not strictly necessary to have the log window open or to be inspecting the network to get this ID. 

e.g. 

```
sqlite> select * from blacklist;
path        lastTryEtag  lastTryModtime  retrycount  errorstring                                                                                                                     X-Request-ID                                    lastTryTime  ignoreDuration  renameTarget  errorCategory
----------  -----------  --------------  ----------  --------------------------------------------------------------------------------------------------------------------            ----------------------------------------------  -----------  --------------  ------------  -------------
.htaccess                1522758480      2           Server replied "403 Forbidden" to "PUT https://localhost/remote.php/dav/files/admin/.htaccess" (Sabre\DAV\Exception\Forbidden)  4c5adb86-2a66-4450-9af9-f6f77b5bdf581522758593   125                           0
```

cc/ @michaelstingl

guruz:
@ogoffart What about adding a column for the in ```.owncloudsync.log ```?
And also modifying the server log templates (for apache, nginx) to show them in the server log file?

SamuAlfageme:
@guruz 

> And also modifying the server log templates (for apache, nginx) to show them in the server log file?

Makes complete sense, will open issues/poke people around to do so.

SamuAlfageme:
Link https://github.com/owncloud-docker/base/issues/22

SamuAlfageme:
WFM 👍 

--- 

For reference, since @michaelstingl asked me: log rotation period is 2MB - all blacklist errors are meant to re-appear in there, but old errors that went away are not granted to be preserved; an error-only log might be an interesting idea to explore, but a different topic. 

https://github.com/owncloud/client/blob/fabc706e43c54f8a9c7ee89807d16e63948fdb8b/src/gui/syncrunfilelog.cpp#L96-L110

I fear most of the log contain _not-so-useful_ info, but that's also a different topic: 

```
[...]

PR #6427, opened 2018-04-04
Blacklist: remember the X-Request-ID

ogoffart:
Issue #6420
Store the X-Request-ID in the SyncFileItem and also in the blacklist.
Note that for consistency reason, the X-Request-ID is also in the
SyncFileItem if the request succeeds.

Currently there is no UI to access it, but it can be queried with sql
commands

### DIFF
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -554,7 +554,7 @@ bool SyncJournalDb::checkConnect()
         return sqlFail("prepare _deleteUploadInfoQuery", _deleteUploadInfoQuery);
     }
 
-    QByteArray sql("SELECT lastTryEtag, lastTryModtime, retrycount, errorstring, lastTryTime, ignoreDuration, renameTarget, errorCategory "
+    QByteArray sql("SELECT lastTryEtag, lastTryModtime, retrycount, errorstring, lastTryTime, ignoreDuration, renameTarget, errorCategory, requestId "
                    "FROM blacklist WHERE path=?1");
     if (Utility::fsCasePreserving()) {
         // if the file system is case preserving we have to check the blacklist
@@ -748,6 +748,16 @@ bool SyncJournalDb::updateErrorBlacklistTableStructure()
         query.prepare("ALTER TABLE blacklist ADD COLUMN errorCategory INTEGER(8);");
         if (!query.exec()) {
             sqlFail("updateBlacklistTableStructure: Add errorCategory", query);
+            re = false;
+        }
+        commitInternal("update database structure: add errorCategory col");
+    }
+
+    if (columns.indexOf("requestId") == -1) {
+        SqlQuery query(_db);
+        query.prepare("ALTER TABLE blacklist ADD COLUMN requestId VARCHAR(36);");
+        if (!query.exec()) {
+            sqlFail("updateBlacklistTableStructure: Add requestId", query);
             re = false;
         }
         commitInternal("update database structure: add errorCategory col");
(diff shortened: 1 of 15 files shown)


ogoffart:
I'm a bit concerned about memory usage.  Should we only keep the request id in case of faillure?

SamuAlfageme:
@ogoffart yes, having a `QByteArray` allocated for each `SyncFileItem` might be a bit too much (specially after reading the considerations on https://github.com/owncloud/client/issues/5853#issue-237753091 - client currently uses ~42bytes) - I'd say keeping it on memory up until a request resolves and only store it on disk on failure will be a better idea. 

And I'd **really** like to have an option on the UI for inexperienced users (not savvy with sqlite commands) to fetch this info and send it to admins.

guruz:
> I'm a bit concerned about memory usage. 

Could avoid this by having the request ID be a crazy hash of the request, path, sync run iteration and PID and some other source. But maybe better not :-)

ckamm:
Personally I've never used `X-Request-ID` - putting it in the log I understand, as it doesn't cause other overhead. But do we really expect the ids to be useful enough to warrant storing them?

src/common/syncjournaldb.cpp
```diff
@@ -753,6 +753,16 @@ bool SyncJournalDb::updateErrorBlacklistTableStructure()
...
     }
 
+    if (columns.indexOf("requestId") == -1) {
+        SqlQuery query(_db);
+        query.prepare("ALTER TABLE blacklist ADD COLUMN requestId VARCHAR(36);");
+        if (!query.exec()) {
+            sqlFail("updateBlacklistTableStructure: Add requestId", query);
+            re = false;
+        }
+        commitInternal("update database structure: add errorCategory col");
```
ckamm: requestId

SamuAlfageme:
@ckamm about:

> putting it in the log I understand, as it doesn't cause other overhead. I don't currently see the ids as particularly useful 

Main concerns I've heard are related to non-persistent-logs setups (user ends up with issues on the "Not Synced" tab but was not using logging). Then he can use those to help the admin/support to correlate and determine which request on the server logs (usually persistent) was responsible for the issue. 

cc/ @michaelstingl

ckamm:
@SamuAlfageme Okay then. It's mostly that I've not personally been in a situation where correlating with server logs was useful and I wanted to understand where the need was coming from.

ogoffart:
@SamuAlfageme so where exactly do you want these ID to be stored? In which log?

guruz:
@ogoffart new column in ```.owncloudsync.log```?

SamuAlfageme:
@ogoffart on https://github.com/owncloud/client/issues/6420#issue-310817823 I thought on using sync journal's `blacklist` table.  As @guruz points out, `.owncloudsync.log` would also make sense (and will even keep **historical** data from previous sync runs, even when the error was solved afterwards):

```
# timestamp | duration | file | instruction | dir | modtime | etag | size | fileId | status | errorString | http result code | other size | other modtime | other etag | other fileId | other instruction | X-Request-ID |
[...]
||Screenshot 2018-04-19 11.52.43.png|INST_NEW|Down|1524131563|2ba3a0df3bd29a417512fe01fce5e522|17901|00000029ocdkxjqy1s03|2|Server replied "403 Forbidden" to "GET https://localhost/remote.php/dav/files/admin/Screenshot 2018-04-19 11.52.43.png"|403|0|0||||4c5adb86-2a66-4450-9af9-f6f77b5bdf581522758593|
[...]
```

guruz:
@ogoffart merge this and potentially do a follow up PR?

guruz:
ˆˆ @SamuAlfageme FYI this is in now

PR #6428, opened 2018-04-04
Uploads: Don't delete unexisting chunks

ogoffart:
Since commit 4dc49ff3, we store an entry in the upload info table even
for non chunked uploads. However, if this fails we don't want to remove
non-existant stale chunks if the upload fails.

Without this commit, we would send a DELETE command to clean non-existant
chunks in the dav/uploads/ namespace.


src/libsync/syncengine.cpp
```diff
@@ -303,6 +303,8 @@ void SyncEngine::deleteStaleUploadInfos(const SyncFileItemVector &syncItems)
     // Delete the stales chunk on the server.
     if (account()->capabilities().chunkingNg()) {
         foreach (uint transferId, ids) {
+            if (!transferId)
+                continue;
```
guruz: Add same comment here?
```// We set a null transfer id because it is not chunked.```

File: src/libsync/propagateuploadng.cpp [hunk 2]
```diff
@@ -83,7 +83,7 @@ void PropagateUploadFileNG::doStartUpload()
     propagator()->_activeJobList.append(this);
 
     const SyncJournalDb::UploadInfo progressInfo = propagator()->_journal->getUploadInfo(_item->_file);
-    if (progressInfo._valid && progressInfo._modtime == _item->_modtime) {
+    if (progressInfo._valid && progressInfo._modtime == _item->_modtime && progressInfo._transferid) {
```
guruz: Create a function instead?
```bool isChunked() { return progressInfo._transferid > 0 }```

ogoffart:
Commit updated

Issue #6277, opened 2018-01-02
Taskbar indicator is green even when conflicts appear


mueller-ma:
### Expected behaviour
When the client cannot sync because of a conflict, the task bar icon should be the orange exclamation mark.

### Actual behaviour
It has the green arrow.

### Client configuration
Client version: 2.4.0

Operating system: Windows 7 64 bit

OS language: DE

ckamm:
This used to be intentional. But I  agree that it makes sense to indicate that user action is required to resolve conflicts in the icon. @SamuAlfageme

SamuAlfageme:
@ckamm finally got to see this one working, nice! 

<p align="center">
  <img src="https://user-images.githubusercontent.com/2644445/38486917-b9c939d0-3bde-11e8-96dd-5c8f78f38e32.png"/>
</p>

Can we also extend the "info" status badge (i.e. https://github.com/owncloud/client/blob/master/theme/colored/state-information-256.png) to the folder sync connection for uniformity?

SamuAlfageme:
I was also discussing "state indicators" coherence (e.g. if the client encountered errors, don't display the green badge...) with @michaelstingl last week and I need to review some behaviors for the multi-account and >1 local sync folder scenarios. 

Bit offtopic: @ogoffart @guruz do you guys know any option for `QSystemTrayIcon`s to display a really simple (green, yellow, red) individual status badge on each "Managed Folder". e.g. how Docker reports status on macOS:  

<p align="center">
  <img src="https://user-images.githubusercontent.com/2644445/38487504-acbd36fe-3be0-11e8-80c1-88282cb654d9.png"/>
</p>

ckamm:
@SamuAlfageme The "green badge" that's simply an icon that was set for the menu item. We could definitely do that.

ckamm:
@SamuAlfageme PR is available for the icon at #6438 if you want to test quickly.

jnweiger:
OK: Taskbar icon has an exclamation mark, while there is a conflict:

![grafik](https://user-images.githubusercontent.com/1108546/43904432-eb6921ce-9bee-11e8-8be9-dbc1640a5d2e.png)

PR #6438, opened 2018-04-09
Folders: Use "Problem" icon for unresolved conflicts

ckamm:
For  #6277

@ogoffart Would you mind if I went in and refactored our icon usage to not depend on "SyncResult"? I think the icon enum should be separate from the sync result enum.


ogoffart:
> Would you mind if I went in and refactored our icon usage to not depend on "SyncResult"? I think the icon enum should be separate from the sync result enum.

Yes, please refactor as much as you like.

ckamm:
I'll merge this and refactor in a different PR

Issue #3819, opened 2015-09-12
[OS X] No overlay icons anymore - Yosemite -  client ownCloud-2.0.1.2694 


Claire75:
The folder I synchronise is /myname/owncloud , not a system or something "special" one .
I used to have ovelay icons until version 2.0.1 , I do not anymore 

I saw https://github.com/owncloud/client/issues/3680 but they say it's for ~/Library sub-folders 

FOr me, it's for my "standard" folder 

Thanks a lot ...

guruz:
Habe you tried to re-login yourself in OS X?

Claire75:
Yes, I did it more than once .

I fixed it on my mac by
- stopping owncloud client
- removing app 
- removing /Library/Application Support/Owncloud

then reboot
then clean RAM ( sudo purge )
then 
- install ownCloud-1.8.4.2531 : folder's small overlay is back :)
- stop owncloud
- install ownCloud-2.0.0.2666 : icons "dropbox like" is back ( see picture)
- stop owncloud
- install ownCloud-2.0.1.2694 : it's ok 

( I put this answer on the owncloud forum ) 
++ sorry for my english, I'm not a native ... :)

jturcotte:
My guess (copying that info from https://github.com/owncloud/client/issues/2340#issuecomment-135026172):

The extension loading is now handled by OSX directly instead of having to inject the bundle directly in Finder, so you can find the option in [System Preferences / Extensions / Finder] where you should be able to enable the "ownCloud Extensions"

There are some situations where the extension wouldn't be enabled properly on the first install, but normally just enabling it there should make it available again.

dragotin:
So can this be closed as nothing to be fixed? Maybe it's worth an entry in the FAQs.

jturcotte:
I feel quite bad about that behavior, already 2-3 users have reported the issue, and I think that many that aren't actively looking for their missing overlay icons might be suffering from it.

However:
- I'm not even sure if we can do better, I haven't found a way in the documentation. See http://apple.stackexchange.com/q/156023
- We already do the "pluginkit -e use ..." hack in the installer. This might fail if the extension wasn't discovered when we run the command though, but this should not happen anymore if the user upgrades from a version that already had the extension (like from 2.0.0 to 2.0.1 in @Claire75's example). So maybe the bug is only actually an issue when users are installing it for the first time or upgrading from <2.0.

[...]

PR #6444, opened 2018-04-10
macOS: Unload the Finder extension on exit #5382 #3819

### LINKED ISSUE #5382: [OS X > 10.10]  ownCloud Extensions is running even when owncloud is not
### Expected behaviour
When you close the sync client, the Extensions process helper should be closed too to avoid having resources consumed and files open. (owncloud should be the parent process)

### Actual behaviour
The `ownCloud Extensions`  process is created by `launchd(1)` and keeps running indefinitely:

img

### Steps to reproduce
1. Quit/Kill the ownCloud process 
2. Check the Activity Monitor and filter by `ownCloud Extensions`

### Client configuration
Client version: 2.2.4 
Operating system: Mac OS X > 10.10 (Finder Sync Extension)

guruz:
...and load it on startup.

Fixes #5382


guruz:
If people don't like the automatic enabling on start of the client, we'll need to provide a checkbox later in our client settings.

guruz:
Note that it takes like 10-15 seconds for macOS actually exitting the finder extension after the client had quit.

PR #6450, opened 2018-04-11
macOS: Implement new dynamic Finder menu items #6328

guruz:
@SamuAlfageme I've only tested this with my local building, but it should also be tested with a build from rotor. 

Sorry for the whitespace mess, XCode is doing that :-(

FYI @jturcotte

### DIFF
--- a/shell_integration/MacOSX/OwnCloudFinderSync/FinderSyncExt/FinderSync.m
+++ b/shell_integration/MacOSX/OwnCloudFinderSync/FinderSyncExt/FinderSync.m
@@ -21,7 +21,7 @@
 - (instancetype)init
 {
 	self = [super init];
-
+	
 	FIFinderSyncController *syncController = [FIFinderSyncController defaultController];
 	NSBundle *extBundle = [NSBundle bundleForClass:[self class]];
 	// This was added to the bundle's Info.plist to get it from the build system
@@ -43,7 +43,7 @@
 	[syncController setBadgeImage:sync label:@"Synchronizing" forBadgeIdentifier:@"NEW+SWM"];
 	[syncController setBadgeImage:warning label:@"Ignored" forBadgeIdentifier:@"IGNORE+SWM"];
 	[syncController setBadgeImage:error label:@"Error" forBadgeIdentifier:@"ERROR+SWM"];
-	
+
 	// The Mach port name needs to:
 	// - Be prefixed with the code signing Team ID
 	// - Then infixed with the sandbox App Group
@@ -55,12 +55,12 @@
 	// the sandboxed App Extension needs.
 	// https://developer.apple.com/library/mac/documentation/Security/Conceptual/AppSandboxDesignGuide/AppSandboxInDepth/AppSandboxInDepth.html#//apple_ref/doc/uid/TP40011183-CH3-SW24
 	NSString *serverName = [socketApiPrefix stringByAppendingString:@".socketApi"];
-	// NSLog(@"FinderSync serverName %@", serverName);
+	//NSLog(@"FinderSync serverName %@", serverName);
 
 	_syncClientProxy = [[SyncClientProxy alloc] initWithDelegate:self serverName:serverName];
 	_registeredDirectories = [[NSMutableSet alloc] init];
 	_strings = [[NSMutableDictionary alloc] init];
-	
+
 	[_syncClientProxy start];
 	return self;
 }
@@ -74,12 +74,26 @@
 		NSLog(@"ERROR: Could not determine file type of %@", [url path]);
 		isDir = NO;
 	}
-	
+
 	NSString* normalizedPath = [[url path] decomposedStringWithCanonicalMapping];
 	[_syncClientProxy askForIcon:normalizedPath isDirectory:isDir];
 }
 
 #pragma mark - Menu and toolbar item support
+
+- (NSString*) selectedPathsSeparatedByRecordSeparator
+{
+	FIFinderSyncController *syncController = [FIFinderSyncController defaultController];
+	NSMutableString *string = [[NSMutableString alloc] init];
+	[syncController.selectedItemURLs enumerateObjectsUsingBlock: ^(id obj, NSUInteger idx, BOOL *stop) {
+		if (string.length > 0) {
+			[string appendString:@"\x1e"]; // record separator
+		}
+		NSString* normalizedPath = [[obj path] decomposedStringWithCanonicalMapping];
+		[string appendString:normalizedPath];
+	}];
+	return string;
+}
 
 - (NSMenu *)menuForMenuKind:(FIMenuKind)whichMenu
 {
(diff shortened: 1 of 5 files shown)


SamuAlfageme:
😎 this is super cool 🎉 

<p align="center">
  
</p>

shell_integration/MacOSX/OwnCloudFinderSync/FinderSyncExt/FinderSync.m
```diff
@@ -101,54 +115,43 @@ - (NSMenu *)menuForMenuKind:(FIMenuKind)whichMenu
...
-	NSArray* items = [[FIFinderSyncController defaultController] selectedItemURLs];
-
-	[items enumerateObjectsUsingBlock: ^(id obj, NSUInteger idx, BOOL *stop) {
-		NSString* normalizedPath = [[obj path] decomposedStringWithCanonicalMapping];
-		[_syncClientProxy askOnSocket:normalizedPath query:@"EMAIL_PRIVATE_LINK"];
-	}];
+- (void)subMenuActionClicked:(id)sender {
+	long idx = [(NSMenuItem*)sender tag];
+	NSString *command = [[_menuItems objectAtIndex:idx] valueForKey:@"command"];
+	NSString *paths = [self selectedPathsSeparatedByRecordSeparator];
```
ckamm: minor: I hope this returns the same list of paths every time / users can't somehow change selected paths between the two calls

shell_integration/MacOSX/OwnCloudFinderSync/FinderSyncExt/SyncClientProxy.m
```diff
@@ -41,20 +41,20 @@ - (void)start
...
 	}
-	
+
 	[[NSNotificationCenter defaultCenter] addObserver:self
-		selector:@selector(connectionDidDie:)
-		name:NSConnectionDidDieNotification
-		object:conn];
+											 selector:@selector(connectionDidDie:)
+												 name:NSConnectionDidDieNotification
+											   object:conn];
```
ckamm: This spacing change is really weird

shell_integration/MacOSX/OwnCloudFinderSync/FinderSyncExt/FinderSync.m
```diff
@@ -101,54 +115,43 @@ - (NSMenu *)menuForMenuKind:(FIMenuKind)whichMenu
...
+		subMenuItem.submenu = subMenu;
+		subMenuItem.image = [[NSBundle mainBundle] imageForResource:@"app.icns"];
+
+		// There is an annoying bug in macOS (at least 10.13.3), it does not use/copy over the representedObject of a menu item
+		// So we have to use tag instead.
+		int idx = 0;
+		for (NSArray* item in _menuItems) {
+			NSMenuItem *actionItem = [subMenu addItemWithTitle:[item valueForKey:@"text"]
+														action:@selector(subMenuActionClicked:)
+												 keyEquivalent:@""];
```
ckamm: ridiculous indent

shell_integration/MacOSX/OwnCloudFinderSync/FinderSyncExt/FinderSync.m
```diff
@@ -101,54 +115,43 @@ - (NSMenu *)menuForMenuKind:(FIMenuKind)whichMenu
 		}
 	}];
 
+	NSString *paths = [self selectedPathsSeparatedByRecordSeparator];
+	// calling this IPC calls us back from client with several MENU_ITEM entries and then our askOnSocket returns again
+	[_syncClientProxy askOnSocket:paths query:@"GET_MENU_ITEMS"];
```
ckamm: I'm sorry, but I don't understand how it's guaranteed that the `END` has arrived before `askOnSocket` returns. I could imaging it waiting for a single message to be received, but how does it know to block until all these lines have arrived?

(I see that we have a custom `socketapisocket_mac.mm` and I don't know the details of how our OSX communication works - so if you're cetain - that's fine with me. To verify, maybe add a `sleep(5)` after sending the `BEGIN` in `GET_MENU_ITEMS` and check whether that breaks it?)

ogoffart: I don't know either.
I'm suspect the only way this can work now is because Qt uses the event loop to flush the "anwser" on the socket.
Would be nice if this was propertly understood though.

shell_integration/MacOSX/OwnCloudFinderSync/FinderSyncExt/FinderSync.m
guruz: screen shot 2018-04-12 at 12 24 49

shell_integration/MacOSX/OwnCloudFinderSync/FinderSyncExt/FinderSync.m
guruz: It's direct synchronous IPC. 
https://gist.github.com/guruz/a49cf1497844599bba58f445216c41cf#file-gistfile1-txt-L23 <- the Extension asks for the menu entries
https://gist.github.com/guruz/a49cf1497844599bba58f445216c41cf#file-gistfile1-txt-L2 <- the extension receives the 4 menu entries (and the BEGIN and END)

shell_integration/MacOSX/OwnCloudFinderSync/FinderSyncExt/SyncClientProxy.m
ckamm: Oh, tabs...

shell_integration/MacOSX/OwnCloudFinderSync/FinderSyncExt/FinderSync.m
ckamm: @guruz Okay, so the extension synchronously calls into the client and that synchronously calls the receiving function in the extension - and the `askOnSocket` only returns when the client function has returned? That would work.

guruz: (It's not a socket on the macOS implementation, even if we call it SocketApi.)

SamuAlfageme:
@guruz today I noticed this message: 

```
objc[13210]: Class FIFinderSyncExtensionHost is implemented in both /System/Library/PrivateFrameworks/FinderKit.framework/Versions/A/FinderKit (0x7fffa934bc90) and /System/Library/PrivateFrameworks/FileProvider.framework/OverrideBundles/FinderSyncCollaborationFileProviderOverride.bundle/Contents/MacOS/FinderSyncCollaborationFileProviderOverride (0x115a76cd8). One of the two will be used. Which one is undefined.
```

When running the client from the command line. Dunno if's a collateral of an OS update (I'm on 10.13.4) or this PR (it's the only change on macOS shell-ext. since 2.4). Any ideas? Google search throws https://github.com/jarvisteach/appJar/issues/285

guruz:
Maybe check https://developer.apple.com/bug-reporting/
:-/

michaelstingl:
@SamuAlfageme maybe something where @felix-schwarz could point us in the right direction? What is needed exactly?

Issue #6185, opened 2017-11-23
Opening the Sharing Dialog while Settings window is in the foreground replaces one with the other


SamuAlfageme:
### Steps to reproduce
1. Open the Settings window
2. Open the Sharing Dialog from the context menu

### Actual Behavior
The settings window is replaced by the sharing dialog. 

<p align="center">
  <img src="https://user-images.githubusercontent.com/2644445/33167120-d4f69b5c-d03c-11e7-97d1-7a95be5eb0aa.gif"/>
</p>

Note: on windows, with also means that the dialog is created without focus:

<p align="center">
  <img src="https://user-images.githubusercontent.com/2644445/33166963-514abbee-d03c-11e7-9049-3f973d2165e8.gif"/>
</p>

cc/ @michaelstingl this is the one we mentioned yesterday on the demo.

ogoffart:
I think this is intentional.  We do the same with the add account wizard.

ogoffart:
```
void ownCloudGui::slotShowShareDialog(const QString &sharePath, const QString &localPath)
{
  [...]
    // For https://github.com/owncloud/client/issues/3783
    _settingsDialog->hide();
```

ckamm:
So #3783 was exactly the "share dialog has no focus" issue on OSX and hiding the settings window was done to make it work there.

ckamm:
See also #3015. Any change here must be tested on Windows and OSX in particular. On linux hiding the settings window is unnecessary in my tests.

ckamm:
Win10 experiments:
* If the account dialog is open, clicking "Share..." will close it and show the share dialog *under* the explorer window
* But if the account dialog is closed, clicking "Share..." will open the share dialog above the explorer window

If hiding the account dialog is disabled, the share dialog still pops up under the explorer window.

ckamm:
@SamuAlfageme To summarize: I couldn't do anything about the "pops up under explorer without focus" problem. :(

guruz:
@ckamm I wonder if you can launch the client from the shell extension and windows would take this as a hint..

ckamm:
@guruz Possibly, but I'd guess not. Since the behavior is correct when the client window is closed I tried workarounds like "close the client, after 2s open the share dialog" but couldn't get it to pop up over the explorer window. :/

ckamm:
The closing of the settings dialog no longer happens on linux and windows. We keep it on osx because it presumably helps put the share dialog on top.

guruz:
[...]

Issue #3783, opened 2015-09-08
[10.10] [OS X] Desktop Client Share Feature opening in background when settings dialog is open and other Window covers settings dialog


ajpdmiv:
The Share with own cloud does not pop up a dialog box. I am currently using: iMac (21.5-inch, Late 2009), OS X 10.10.5

guruz:
@ajpdmiv But you have the menu item in the menu?
Do you see the green/yellow/red overlay icons over files though?

stonerl:
@guruz I observed something similar on OSX.

1. Open the ownCloud Settings Window.
2. Don't close the Window but switch to another application that covers the Settings Window completely.
3. Open Finder, select a file and click on "Share with ownCloud".
4. No Sharing Window appears.

In fact the sharing Windows gets opened, but in the Background. When on goes back to the ownCloud Settings Window and uses the App Exposé you can see the ownCloud Settings Window and the Sharing Window.

This doesn't happen when the Settings Window is closed. But there is a general problem with the sharing Window. As soon as you click another Window the Sharing Window vanishes in the background. One cannot CMD-TAB to get it back. The only way to get this window back is through Mission Control.

dragotin:
Seems to be a fail of `ownCloudGui::raiseDialog( QWidget *raiseWidget )` on MacOSX. @jturcotte ideas?

jturcotte:
I haven't investigated OS X throughly, but my guess is that the window manager protects the user from background applications taking the front without the user asking. And this is what happens in this case, we cheat by having Finder asking the ownCloud client to show a popup to the user through a socket rather than a normal input event handler of the process currently having the user's focus.

See https://github.com/owncloud/client/issues/3015 for a similar issue on Windows.

Still worth verifying on OS X though. http://stackoverflow.com/a/2767236/56315 mentions that we could also try [window setLevel:NSFloatingWindowLevel]; in addition to activateIgnoringOtherApps and makeKeyAndOrderFront, but that would make the dialog always floating and the user wouldn't be able to put anything in front of that dialog.

Dianafg76:
I can't reproduce.
@stonerl, @ajpdmiv you could try this version: http://download.owncloud.com/desktop/daily/ownCloud-2.0.2.2722-nightly20150918.pkg
Thanks

stonerl:
Downloaded the nightly. And the first part is fixed. The sharing window won't open in the background when the settings window is open.

[...]

PR #6452, opened 2018-04-12
Share dialog: Don't hide account settings before showing

ckamm:
At least on windows and linux. OSX would need tests first since
the addition was originally supposed to help with OSX problems,
see #3783.

For #6185

### DIFF
--- a/src/gui/owncloudgui.cpp
+++ b/src/gui/owncloudgui.cpp
@@ -1037,8 +1037,11 @@ void ownCloudGui::slotShowShareDialog(const QString &sharePath, const QString &l
         return;
     }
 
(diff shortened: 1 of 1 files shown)


SamuAlfageme:
@ckamm just tried with macOS 10.13.3 and one window still replaces the other; damn 😔

ckamm:
@SamuAlfageme I haven't touched it on OSX because the other issues suggested that the account settings hiding might actually be helpful there. I just disabled it for win and linux builds.

ckamm:
Jenkins failure is unrelated. @guruz @ogoffart I could use a review!

Issue #6396, opened 2018-03-20
Link to, and improve documentation on conflict resolution


ckamm:
Currently, when there are conflicts, the user gets tray notifications and sees a yellow box in the account status. These should redirect them to the "not synced" list where all pending conflicts are visible. However, currently there's no explanation on what the steps to resolve a conflict might be.

This page should have a link to our documentation and the documentation should be updated with an explanation and a step by step guide on resolving conflicts and common issues.

@guruz

SamuAlfageme:
👍

ckamm:
Not done, we want a link to the new documentation show up in the ui when conflicts are present

ckamm:
![image](https://user-images.githubusercontent.com/833306/38794380-305e460a-4155-11e8-98c6-52d13284ef97.png)

ckamm:
Note that we only want to merge the PR https://github.com/owncloud/client/pull/6456 once the documentation link becomes available.

ckamm:
Track in PR instead

jnweiger:
OK: 
clicking on the yellow badge, opens the 'not synced' tab, clicking on a conflicting file opens the file browser, 
clicking on the documentation link opens https://doc.owncloud.org/desktop/2.5/conflicts.html
which has a good description what to do.

Idea:
Add to the documentation, currently the strategy is 'server always wins", but give a warning that this strategy may change in  future.

PR #6456, opened 2018-04-16
Issues: Show link to conflicts doc if there are some

ckamm:
The link url is themed.

For #6396

Only merge once we can check that the link will actually work. Currently it's a 404: https://doc.owncloud.org/desktop/2.5/conflicts.html

### DIFF
--- a/src/gui/issueswidget.cpp
+++ b/src/gui/issueswidget.cpp
@@ -120,6 +120,11 @@ IssuesWidget::IssuesWidget(QWidget *parent)
     _ui->_tooManyIssuesWarning->hide();
     connect(this, &IssuesWidget::issueCountUpdated, this,
         [this](int count) { _ui->_tooManyIssuesWarning->setVisible(count >= maxIssueCount); });
+
+    _ui->_conflictHelp->hide();
+    _ui->_conflictHelp->setText(
+        tr("There were conflicts. <a href=\"%1\">Check the documentation on how to resolve them.</a>")
+            .arg(Theme::instance()->conflictHelpUrl()));
 }
 
 IssuesWidget::~IssuesWidget()
@@ -277,6 +282,8 @@ void IssuesWidget::slotProgressInfo(const QString &folder, const ProgressInfo &p
             }
         }
         emit ProgressDispatcher::instance()->folderConflicts(folder, conflicts);
+
+        _ui->_conflictHelp->setHidden(Theme::instance()->conflictHelpUrl().isEmpty() || conflicts.isEmpty());
     }
 }
 
--- a/src/libsync/theme.cpp
+++ b/src/libsync/theme.cpp
@@ -228,6 +228,16 @@ QString Theme::helpUrl() const
     return QString::fromLatin1("https://doc.owncloud.org/desktop/%1.%2/").arg(MIRALL_VERSION_MAJOR).arg(MIRALL_VERSION_MINOR);
 }
 
+QString Theme::conflictHelpUrl() const
+{
+    auto baseUrl = helpUrl();
+    if (baseUrl.isEmpty())
+        return QString();
+    if (!baseUrl.endsWith('/'))
+        baseUrl.append('/');
+    return baseUrl + QStringLiteral("conflicts.html");
+}
+
 QString Theme::overrideServerUrl() const
 {
     return QString();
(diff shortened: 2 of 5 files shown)


src/libsync/owncloudtheme.cpp
```diff
@@ -56,5 +56,11 @@ QPixmap ownCloudTheme::wizardHeaderLogo() const
     return QPixmap(hidpiFileName(":/client/theme/colored/wizard_logo.png"));
 }
 
+QString ownCloudTheme::conflictDocumentationLink() const
+{
+    return QString("https://doc.owncloud.org/desktop/%1.%2/conflicts.html")
```
ogoffart: Please Use Theme::helpUrl as a base,  and move the function to Theme  and remove the override in ownCloudTheme.

ckamm: Oh, I didn't see `helpUrl()`. Yes, I didn't want oc-branded urls to show up in themes that weren't adjusted for this change. But if `helpUrl` already exists I'll just use that.

src/gui/issueswidget.cpp
```diff
@@ -277,6 +282,8 @@ void IssuesWidget::slotProgressInfo(const QString &folder, const ProgressInfo &p
             }
         }
         emit ProgressDispatcher::instance()->folderConflicts(folder, conflicts);
+
+        _ui->_conflictHelp->setHidden(Theme::instance()->configFileName().isEmpty() || conflicts.isEmpty());
```
ckamm: oops

ckamm:
Ok, ready to merge once help url becomes valid.

guruz:
FYI @michaelstingl for themeing / white label stuff.

michaelstingl:
@guruz What will change for branded clients? Could you elaborate?

ckamm:
@michaelstingl See the screenshot in https://github.com/owncloud/client/issues/6396#issuecomment-381499838

The url that the help link points to is based on `Theme::helpUrl` by default and independently themable by overriding `Theme::conflictHelpUrl`. If a theme doesn't override either, the link will point to the ownCloud documentation, similar to the change done in 42b254785e8b91e5e2e4d8c9dfa7681d947f7f8f

michaelstingl:
Should be brandable, similar to Help URL:

> **Help URL**
> Select this option to show a help URL in your client. The help option will be shown on menu and clicking on it will open a browser viewing the selected URL.
> - [ ] Show help url?

> **Conflict Help URL**
> Select this option to show a help URL for conflict files in your client. The help option will be shown…
> - [ ] Show conflict help url?

If this isn't set, nothing should be shown. If it's set, a new input field should be shown to enter the URL.

src/libsync/theme.cpp
```diff
@@ -228,6 +228,11 @@ QString Theme::helpUrl() const
     return QString::fromLatin1("https://doc.owncloud.org/desktop/%1.%2/").arg(MIRALL_VERSION_MAJOR).arg(MIRALL_VERSION_MINOR);
 }
 
+QString Theme::conflictHelpUrl() const
+{
+    return helpUrl() + QStringLiteral("conflicts.html");
```
ogoffart: When helpUrl is empty, this should be empty too.

ckamm: updated

ckamm:
@michaelstingl Yes, this kind of option setup would work. If the options are not checked, the theme functions are overridden to return an empty string, and if options are checked they are overridden to return the specified url.

guruz:
https://doc.owncloud.org/desktop/2.5/conflicts.html is still 404. 
I think someone mentioned that we shall create a 2.5 branch then this can be done. So merge this (to master) so we can branch off 2.5 and then ping @tboerger @settermjd ?

ckamm:
@guruz that was also my understanding

ogoffart:
```
return QString::fromLatin1("https://doc.owncloud.org/desktop/%1.%2/").arg(MIRALL_VERSION_MAJOR).arg(MIRALL_VERSION_MINOR);
```

This will always be a 404 before the branching is done.  (i mean, once we change the master version to 2.6, the master dailly build will still have a 404.

ckamm:
Yes, let's just fix the link if it ends up incorrect.

Issue #6405, opened 2018-03-26
[2.5.x] Client recovery strategy for broken chunked uploads when server hd was full?


guruz:
Server replied "400 Bad Request" to "MOVE https://..../owncloud/remote.php/dav/uploads/guruz/45471869/.file"
["Chunks on server do not sum up to 39730872 but to 39688248"](https://github.com/owncloud/core/blob/6bc3941603a5fab30d85f89a75ae8ec53c158264/apps/dav/lib/Upload/ChunkingPlugin.php#L105)
(While uploading, I had also received "Connected Closed" but I guess that can be ignored as the server was bin byzzantine state.)

FYI @DeepDiver1975 @ogoffart

ogoffart:
Thanks for the but report.
Can you please fill the bug template? 
In particular, what are the step to reproduce and what would be the expected behavior?

If the server only reports Bad Request, that's what the client will show.

ckamm:
Fwiw, there already is a recovery strategy for server chunk brokenness - but only for http code 412 by default. See `PropagateUploadFileCommon::checkResettingErrors()`. The idea is that if you receive this error code three times you throw away the `UploadInfo` and start from scratch instead.

guruz:
@ogoffart I don't want to fill the server hard drive again to know how exactly to reproduce the bug, obviously.

@ckamm I unfortauntely don't know if it did a follow-up sync. Maybe it had the grace perioud and was waiting to do a follow up? I worked around the whole issue by renaming the file (after cleaning up the hd a bit)
So HTTP 412 is precondition failed? hmmmm.
I don't know if the server should return that. What does @DeepDiver1975 think?

ckamm:
@guruz I didn't mean that the server should reply 412, I was just saying that there was a mechanism that could be used for recovering from broken uploads if we can make it apply to this case (in case it doesn't work already, which it should!).

ckamm:
@guruz Why ReadyToTest, was there a change here?

ckamm:
This is closed by the PR #6460

PR #6460, opened 2018-04-17
Upload: Add more chunk resuming unit tests

ckamm:
For testing #6405

### DIFF
--- a/src/libsync/propagateuploadng.cpp
+++ b/src/libsync/propagateuploadng.cpp
@@ -143,6 +143,12 @@ void PropagateUploadFileNG::slotPropfindFinished()
         qCCritical(lcPropagateUpload) << "Inconsistency while resuming " << _item->_file
                                       << ": the size on the server (" << _sent << ") is bigger than the size of the file ("
                                       << _item->_size << ")";
+
+        // Wipe the old chunking data.
+        // Fire and forget. Any error will be ignored.
+        (new DeleteJob(propagator()->account(), chunkUrl(), this))->start();
+
+        propagator()->_activeJobList.append(this);
         startNewUpload();
         return;
     }
--- a/test/testchunkingng.cpp
+++ b/test/testchunkingng.cpp
@@ -64,8 +64,8 @@ private slots:
         QCOMPARE(fakeFolder.uploadState().children.count(), 2); // the transfer was done with chunking
     }
 
-
-    void testResume () {
+    // Test resuming when there's a confusing chunk added
+    void testResume1() {
         FakeFolder fakeFolder{FileInfo::A12_B12_C12_S12()};
         fakeFolder.syncEngine().account()->setCapabilities({ { "dav", QVariantMap{ {"chunking", "1.0"} } } });
         const int size = 300 * 1000 * 1000; // 300 MB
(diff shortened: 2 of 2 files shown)


src/libsync/propagateuploadng.cpp
```diff
@@ -219,7 +224,9 @@ void PropagateUploadFileNG::slotDeleteJobFinished()
 
 void PropagateUploadFileNG::startNewUpload()
 {
-    ASSERT(propagator()->_activeJobList.count(this) == 1);
+    if (propagator()->_activeJobList.count(this) == 0)
```
ogoffart: So you are changing the invariant?  I guess this is only hapenning when there were more data on the server.  But i'd argue that we should be consistant, and always add the activeJobList in the same place.

For example, if we start uploading chunks in parallel again, code like that can lead to inconsistency.

test/testchunkingng.cpp
```diff
@@ -96,6 +98,130 @@ private slots:
...
+                Q_ASSERT(request.rawHeader("OC-Chunk-Offset").toLongLong() >= firstChunk.size);
+            } else if (op == QNetworkAccessManager::DeleteOperation) {
+                deletedPaths.append(request.url().path());
+            }
+            return nullptr;
+        });
+
+        QVERIFY(fakeFolder.syncOnce());
+
+        for (const auto& toDelete : chunksToDelete) {
```
ogoffart: This block of code is not particularly efficient...  But that does not matter :-)

test/testchunkingng.cpp
```diff
@@ -96,6 +98,130 @@ private slots:
...
+        fakeFolder.syncEngine().account()->setCapabilities({ { "dav", QVariantMap{ {"chunking", "1.0"} } } });
+        const int size = 300 * 1000 * 1000; // 300 MB
+        partialUpload(fakeFolder, "A/a0", size);
+        QCOMPARE(fakeFolder.uploadState().children.count(), 1);
+        auto chunkingId = fakeFolder.uploadState().children.first().name;
+        const auto &chunkMap = fakeFolder.uploadState().children.first().children;
+        quint64 uploadedSize = std::accumulate(chunkMap.begin(), chunkMap.end(), 0LL, [](quint64 s, const FileInfo &f) { return s + f.size; });
+        QVERIFY(uploadedSize > 50 * 1000 * 1000); // at least 50 MB
+
+        // Add a chunk that makes the file completely uploaded
```
ogoffart: i'd would have made an error on the move...  but this is fine too.

test/testchunkingng.cpp
```diff
@@ -96,6 +98,130 @@ private slots:
...
+        fakeFolder.setServerOverride([&](QNetworkAccessManager::Operation op, const QNetworkRequest &, QIODevice *) -> QNetworkReply * {
+            if (op == QNetworkAccessManager::PutOperation) {
+                sawPut = true;
+            } else if (op == QNetworkAccessManager::DeleteOperation) {
+                sawDelete = true;
+            }
+            return nullptr;
+        });
+        QVERIFY(!sawPut);
+        QVERIFY(!sawDelete);
```
ogoffart: I guess you meant to put these two QVERIFY after the sync!

(Also maybe add a sawMove... why not)

ogoffart:
Jenkins is failing with:

```
FAIL!  : TestChunkingNG::connectionDroppedBeforeEtagRecieved(small file) '!fakeFolder.syncOnce()' returned FALSE. ()
   Loc: [/var/lib/jenkins/workspace/cloud-client_client_PR-6460-MGEVCM5XTFJQO7UC3CJR3U3LRVCP3QWVQGE5SP3OYDUCM2HY26JA/test/testchunkingng.cpp(553)]
```

I don't know what causes this error though

ckamm:
Updated. The ChunkingNg tests all pass for me locally.

PR #6467, opened 2018-04-18
Start to migrate to Drone in favor of Jenkins

tboerger:


tboerger:
I have dropped the `-V` flag now, otherwise the log output gets a hard cut after ~14k lines of log :(

DeepDiver1975:
Will you add the cross compiling as well?

tboerger:
> Will you add the cross compiling as well?

Talked to @michaelstingl, looks like we don't need that since Appveyor will do Windows builds.

.drone.yml
```diff
@@ -0,0 +1,81 @@
...
+
+  prepare-gcc:
+    image: owncloudci/client:latest
+    pull: true
+    environment:
+      - LC_ALL=C.UTF-8
+    commands:
+      - mkdir gcc-build
+      - cd gcc-build
+      - cmake -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ -DCMAKE_BUILD_TYPE="Debug" -DUNIT_TESTING=1 ..
```
guruz: Is there a reason to use ninja with clang but not gcc?

tboerger: Yes, same as for the Jenkinsfile:

```
// We are building "Linux - GCC" with "make" and "Linux - Clang" with ninja,
// the combinations are more or less arbitrarily chosen. We just want to
// check that both compilers and both CMake generators work. It's
// unlikely that a specific generator only breaks with a specific
// compiler.
```

.drone.yml
```diff
@@ -0,0 +1,85 @@
+workspace:
+  base: /drone
+  path: src/github.com/owncloud/client
+
+branches:
+  - master
+  - 2.4
+  - 2.3
```
ogoffart: I don't know what this "branches" entry is for exactly, but we don't need 2.3 as this branch is dead.
(But there will be a 2.5 soon)

.drone.yml
ogoffart: Exactly.

Maybe the comment could be put back in there too so future reader will not wonder the same.

ogoffart:
BTW, how easy is it to upgrade components on the machines?  (say, i want to upgrade Qt, or the compiler, or install some other third party library)
Can it be done per/branch.
Can we also have two different Qt build  (say one with Qt 5.6 and one with Qt 5.10)?  
Actually, what version of Qt is it currently using?

tboerger:
> BTW, how easy is it to upgrade components on the machines? (say, i want to upgrade Qt, or the compiler, or install some other third party library)
> Can it be done per/branch.
> Can we also have two different Qt build (say one with Qt 5.6 and one with Qt 5.10)?
> Actually, what version of Qt is it currently using?

Since everything is running within Docker container we just need to provide another docker image. Currently it's running on https://github.com/owncloud-ci/client which is based on QT shipped with Ubuntu 18.04 (5.9.5). If there is some need for different QT versions we will extend the mentioned repository with more branches which results in different tags of the Docker image.

.drone.yml
tboerger: I have added it to the top of the .drone.yml

.drone.yml
tboerger: It's similar to the `branches` attribute in Travis, it limits builds to these target branches, so it matches pushs to these branches as well as pull requests targetting these branches.

tboerger:
I have added commits to drop `Jenkinsfile` and `.travis.yml` as well.

dschmidt:
Yes, we are probably going to drop MinGW support - the problem is that maintaining the cross toolchain is a lot of work and we need native Windows for MSI anyway. 
What we could do without much effort (hopefully) is to enable native MinGW builds in AppVeyor if you want to. 

@DeepDiver1975 raised a good question here: Installers are a bit complicated right now, the old NSIS installer works only with the cross build (right now) and the MSI code is not public. That's why we currently can't build Qt 5.10 based installers for Windows in public...

ogoffart:
I think it is still important to build with Qt 5.6 in at least one configuration (gcc or clang). 
Because we still support Qt 5.6 and we tend to break that easily as we all use newer versions on our dev machines.

.drone.yml
```diff
@@ -0,0 +1,91 @@
...
+# with a specific compiler.
+#
+
+workspace:
+  base: /drone
+  path: src/github.com/owncloud/client
+
+branches:
+  - master
+  - 2.4
```
ogoffart: There is a 2.5 branch now so i guess it needs to be added there too.

tboerger: Done

tboerger:
I have reverted the deletion of the Jenkinsfile, that way we can keep Jenkins and Drone running side by side for some time. If we are fine with the results of Drone we can drop the Jenkinsfile later on with a followup PR.

Issue #3664, opened 2015-08-21
Improve the "skip folder configuration" and manual folder configuration workflow. (Was: Do not include 'root' by default in desktop client)


tzaeschke:
I'm new to OwnCloud. When I started the Client and wanted to add folders for syncing, it told me to 'unsync' the root first. However, unsyncing the root is not possible, instead I had to 'remove' the root and then add individual folders. I took me a while (and some help) to figure this out, and I don't think it is very intuitive.

An easy improvement would be to change the error/warning message in the 'Add folder...' menu to indicate that the 'root should be removed before new folders can be added'.

However I think the best option would be to not have the root by default in the client, instead, by default, either don't show any folders at all or add some default folders, such as, 'Music' and 'Document' as separate folders.
I think this is what most people need anyway, because often people may have more than a single folder/partition that they want to sync (especially in Windows).

This is related to https://github.com/owncloud/client/issues/3632 and https://github.com/owncloud/client/issues/3645. I previously posted this on the forum: https://forum.owncloud.org/viewtopic.php?f=14&t=29949&p=94583#p94583

phil-davis:
The 2.0 "Add an Account" ownCloud connection Wizard asks for the server address, then username and password, then the "Setup local folder options" dialog that asks details of what to sync from the server to where on the client. That allows you to sync whatever is on the server with a single location on the client, and, if you like, choose only a subset of folders on the server to sync with the client.

That has a button "Skip folders configuration".

If you want to sync various folders in different places on the client to individual folders on the server then you need to do "Skip folders configuration" in the wizard. After the account is saved, then add folders one-by-one.

Perhaps there can be some way to explain this on the "Setup local folder options" dialog, and that would help first-time users?

It might clutter the wizard if a lot of stuff was added to it to allow direct setup of multiple client-server folder pair sync within the wizard.

tzaeschke:
Maybe "Skip folders configuration" could also be renamed? As a new user I wouldn't expect to have to "skip folders configuration" if I want to configure multiple folders. Especially since the "folders configuration" only allows to configure a single folder... Or did I misunderstand that?

phil-davis:
You understand correctly. The wizard only allows for setting up the sync of the server root to a (single) place on the client, and (if you want) selecting a subet of folders from the server to sync inside that place on the client.
[...]

PR #6470, opened 2018-04-19
Wizard: Move the 'Skip folder configuration' in a radio button

ogoffart:
Issue #3664

The goal is to make it more visible than the button.
An alternative would have been to simply rename the button to "Manual folder configuration", but I feel a radio button works better.

Here is how it looks:
screenshot_20180419_141805

cc/ @michaelstingl

### DIFF
--- a/src/gui/owncloudsetupwizard.cpp
+++ b/src/gui/owncloudsetupwizard.cpp
@@ -57,7 +57,6 @@ OwncloudSetupWizard::OwncloudSetupWizard(QObject *parent)
     connect(_ocWizard, &OwncloudWizard::basicSetupFinished,
         this, &OwncloudSetupWizard::slotAssistantFinished, Qt::QueuedConnection);
     connect(_ocWizard, &QDialog::finished, this, &QObject::deleteLater);
-    connect(_ocWizard, &OwncloudWizard::skipFolderConfiguration, this, &OwncloudSetupWizard::slotSkipFolderConfiguration);
 }
 
 OwncloudSetupWizard::~OwncloudSetupWizard()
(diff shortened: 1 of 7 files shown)


guruz:
Can a native speaker say if "Manually create custom synchronization folders" is good wording?
@settermjd @phil-davis @konisolde ?
Maybe it should say "individual" instead of "custom"?

CC @michaelstingl @pmaier1

konisolde:
@guruz  "Manually create synchronization folders" - I think I'd leave out the "custom".

guruz:
@ogoffart also we use 'sync folder' everywhere not 'synchronization folder'.

ogoffart:
> also we use 'sync folder' everywhere not 'synchronization folder'.

Not really. The button in the UI is "Folder Sync Connection" which would be improved.
We probably should rename the other location "synchronization": It is not so good to use abbreviations.

phil-davis:
```
Synchronize everything from server (27 MB)
Choose what to sync
Create synchronization folders later
Create placeholders instead of downloading files (experimental)
```
Maybe this?
@settermjd

guruz:
So the choice is between
```Manually create sync folders``` 
and 
```Create sync folders later ```
(So far we use ```sync folder``` and not ```synchronization folder```everywhere and I'm sure there was a reason for this, so we keep it. Dropbox does the same)

Let's wait for @settermjd as tie breaker?

konisolde:
Vote for: Manually create sync folders

phil-davis:
I will not be offended - either is fine for me.

ogoffart:
So from the feedback I decided to name it `Manually create folder sync connections`  ("Folder Sync Connection" are the words we use everywhere and hence how it should be called.

I also added an explainatory tooltip which is shown then the mouse hover this item for some seconds.

screenshot_20180423_124102

guruz:
Sounds good..

jancborchardt:
Since it’s now a lot of options, it’s probably also good to add »(recommended)« behind the first option, in addition of having it as default. Otherwise it’s very overwhelming for less technical people. (Which was the reason for not having so many there in the first place. ;)

konisolde:
As a "less technical people" person, I like that suggestion!

ckamm:
@jancborchardt Good idea, PR here https://github.com/owncloud/client/pull/6534

PR #6477, opened 2018-04-20
[2.3] Start to migrate to Drone in favor of Jenkins

tboerger:


ogoffart:
The 2.3 branch is dead. (there won't be anymore 2.3 releases)
So I don't believe this is needed.

tboerger:
> The 2.3 branch is dead. (there won't be anymore 2.3 releases)
> So I don't believe this is needed.

Does this mean we can also disable Smashbox nightly tests for this branch?

ogoffart:
> Does this mean we can also disable Smashbox nightly tests for this branch?

Yes.

But that depends what are the smashbox test are testing.
If they are supposed to test the client, yes. It can be stopped.
If they are supposed to test that new server still work with older client, then, they need to be kept.

tboerger:
OK, than let's close this PR.

PR #6484, opened 2018-04-23
[2.5] Start to migrate to Drone in favor of Jenkins

tboerger:


ogoffart:
This commit should not have been merged as master and 2.5 have not been separated yet.
Not much of a problem, just a small glitch in the git history.

tboerger:
> This commit should not have been merged as master and 2.5 have not been separated yet.
> Not much of a problem, just a small glitch in the git history.

But at least the commits have been cherry-picked.

Issue #6119, opened 2017-10-25
Show a warning when folder watcher becomes unreliable


ckamm:
See #2297 and 66f0ce6616edcb6e44d235ae779170eb28c9c7c0. Suggested by @moscicki.

When a folder watcher becomes unreliable the client will revert to doing full local discovery. Possibly there should be a warning when that happens to let the user know what's happening and what steps they could take to make the problem go away.

Currently the only case is inotify file descriptor exhaustion.

moscicki:
Many thanks for opening this!

guruz:
Is the 100-sync-inotify.conf not working?
https://github.com/owncloud/client/issues/4107

guruz:
2.5.0 beta1 is out :-)
https://central.owncloud.org/t/desktop-sync-client-2-5-0-beta1-released/14667
Everyone, please comment here if we can close this issue. Thank you.

jnweiger:
### Mint Tara (aka Ubunut 18.04)

testpilotcloud version 2.5.0daily20180801 (build 9974)

sudo cat /proc/sys/fs/inotify/max_user_watches
524288

OK: setting from our sync-inotify.conf file is applied.

sudo sysctl fs.inotify.max_user_watches=280
sudo sysctl -p

populate ca 2000 files in ca 20 directories
find . | wc -l
2179
find . -type d | wc -l
24
![grafik](https://user-images.githubusercontent.com/1108546/43686661-e0e3a5da-98c9-11e8-815e-afe25ccd2b56.png)


OK: When the limit is too low, the warning appears. When the limit is set higher, now warning appears.
In this test, 270 was found too low, and 280 was found sufficient

BAD: The exact numbers are hard to explain. It should be either in the 2000 range (if files are watched) or well below 100 (if directories are watched).

PR #6485, opened 2018-04-24
Show a tray message when a folder watcher becomes unreliable 

ckamm:
See #6119 

It's only a tray message. I feel that if we made a message box, we'd need to include a "never show this again" box...


ckamm:
No, and some newlines got eaten too (kde) :/ I'm still looking at it

guruz:
(needs to be merged or cherry-picked into 2.5 too..)

Issue #5774, opened 2017-05-16
[Sharing Dialog] [OSX, Windows] Adjust the layout when a file cannot be shared


SamuAlfageme:
The layout of the new sharing dialog is not wrapping the error message when the target file/folder cannot be shared: 
![screenshot 2017-05-16 16 23 16](https://cloud.githubusercontent.com/assets/2644445/26111255/ea507ca8-3a54-11e7-85e3-d2574b68f2bb.png)

### Steps to reproduce
1. Disable sharing on an external storage mountpoint
2. Try to share a file from that storage in the client

ckamm:
Looks good on Linux, possibly OSX only?

SamuAlfageme:
Windows also: 

<p align="center">
  <img src="https://cloud.githubusercontent.com/assets/2644445/26150538/106dd8f4-3aff-11e7-9797-f4bdf91726bf.png"/>
</p>

SamuAlfageme:
Clarification, text is wrapping after: 8da7919a4375990c69569c56de5a17ff1464ded4 but the layout looks overlapping: 

<p align="center">
  <img src="https://cloud.githubusercontent.com/assets/2644445/26204046/d6857838-3bdd-11e7-8fd4-25dfeb4ed4e7.png"/>
</p>

ckamm:
PR #5781

SamuAlfageme:
Following the comments in https://github.com/owncloud/client/pull/5781, there's still a layout problem with the container widget box on macOS when there's no tabs to display: 

<p align="center">
  <img src="https://cloud.githubusercontent.com/assets/2644445/26244158/f1e34806-3c8e-11e7-9f0f-eccc08801fc8.png"/>
</p>

Also,

> I think https://github.com/owncloud/client/pull/5781/commits/21fa9be560eac12542469fffed870ef2982a4999 is hiding the OS's default icon displayed on the sharing dialog unless the thumbnail from the server is fetched. 

| pre- | post- |
|:---:|:--------------------:|
| ![screenshot 2017-05-22 12 29 09](https://cloud.githubusercontent.com/assets/2644445/26307370/6fe8785e-3ef4-11e7-9dcc-1d36d7d9eb58.png) | ![post](https://cloud.githubusercontent.com/assets/2644445/26307361/64c4e930-3ef4-11e7-9d01-3e8f9986c0fb.png) |

ckamm:
[...]

PR #6488, opened 2018-04-25
ShareDialog: macOS grey odditiy #5774

guruz:
Maybe @ogoffart knows why this would be needed at all?
(And only on macOS)

@ckamm how is it on Linux?
@SamuAlfageme might need to check Windows


ogoffart:
@guruz you need to mention the github issue number in the body or comments, so cross referencing works.
Like this:  #5774

guruz:
> you need to mention the github issue number in the body or comments, so cross referencing works.

Works for me:
screen shot 2018-04-25 at 13 54 14
Or in commits:
screen shot 2018-04-25 at 13 54 49

ogoffart:
@guruz i'm talking about the reference to the merge request (6488) inside the issue (5774).

guruz:
yes, i had edited my comment there to get a link back. meh. sounds like a bug in github

Issue #6406, opened 2018-03-26
Overlay icons problem on arch linux


lbayerlein:
### Expected behaviour
I am starting Nautilus, click on the folder and there are folders and files with a given state in it. This state would be shown by small icons extending the default icons.

### Actual behaviour
Nautilus does not show anything.

### Steps to reproduce
1. Install Nautilus and owncloud-client in the actual version. Nautilus via pacman, owncloud via aur (in my case yaourt)
2. Start owncloud client with a configured account
3. Start Nautilus from the system application (not from the terminal as user)

### Server configuration
Operating system: CentOS 7.15011

Web server: Apache2 (2.4)

Database: MariaDB

PHP version: 7.1

ownCloud version:

Storage backend (external storage):

### Client configuration
Client version: 2.4.1

Operating system: Arch Linux (Linux version 4.15.10-1-ARCH (builduser@heftig-18961) (gcc version 7.3.1 20180312 (GCC)) #1 SMP PREEMPT Thu Mar 15 12:24:34 UTC 2018)

OS language: DE

Qt version used by client package (Linux only, see also Settings dialog): 5.10.1

Client package (From ownCloud or distro) (Linux only): 

Installation path of client: default

guruz:
(See also discussion on #owncloud-client-dev on freenode:
```
<bigbalu> but what I have found: if I run nautilus from system tray - all icons are not shown. If I run it manually from terminal, everything works
<bigbalu> should I open a bug?
````
)

**Comment by ogoffart** [comment]
Any umlaut in the file paths?  I suspect this is a duplicate of  #6393 Fixed by #6397

**Comment by ckamm** [comment]
But why would the umlaut problem apply only to the nautilus that's started from the system tray? 

@lbayerlein Could you check if both are using the same nautilus binary (you're on arch, maybe one of them is using python3 and not having the encoding issues?)

```
ls -l /proc/`pgrep -n nautilus`/exe
```

Otherwise: Could you check what the system-tray started nautilus is saying?

```
strace -p`pgrep -n nautilus` -s9999 -e write
```

**Comment by lbayerlein** [comment]
Hi @ckamm and @ogoffart ,

```
ls -l /proc/`pgrep -n nautilus`/exe
```
shows me:
```
lrwxrwxrwx 1 ludwig ludwig 0 29. Mär 09:01 /proc/9001/exe -> /usr/bin/nautilus
```
And for the `strace` command, I get many lines with these informations:
```
[...]

PR #6490, opened 2018-04-25
Nautilus: Fix Python3 and remove many debug

ogoffart:
Use b'\n' in the call to rfind, as the _remainder is bytes, not a string.

Remove most of the debug message which happens during normal operation.
They are mostly spamming the nautilus console, and can also cause bug
as they may throw exception in case of wrong encoding.

Issue: #6406


ogoffart:
I think the b'\n'  should fix a problem that was introduced when merging 2.4 and master.

Removing the debug message should fix Issue: #6406

PR #6491, opened 2018-04-25
Proxy settings: Fix bad behavior with empty host

ckamm:
Fixes f6a075ef54a3434e758dcddc6a29b218c4f2a247 for #5885

When no host is given it shouldn't just *appear* that there is no
proxy, there actually shouldn't be a proxy set then.


src/gui/networksettings.cpp
```diff
@@ -164,6 +164,9 @@ void NetworkSettings::saveProxySettings()
         cfgFile.setProxyType(QNetworkProxy::DefaultProxy);
     } else if (_ui->manualProxyRadioButton->isChecked()) {
         int type = _ui->typeComboBox->itemData(_ui->typeComboBox->currentIndex()).toInt();
+        QString host = _ui->hostLineEdit->text();
+        if (host.isEmpty())
```
guruz: Didn't @SamuAlfageme somewhere mention to .trim() the hostname too?

ckamm: Hm, I didn't see that.

Issue #6259, opened 2017-12-20
Allow enabling/disabling the beta update channel


SamuAlfageme:
Recently I saw how some colleagues were offered to update to 2.4-RC2 (https://github.com/owncloud/enterprise/pull/2362) after we introduced support for beta channels in #6048

Would be cool to give the users the chance to only display update notices on stable releases on the Settings window. 

Related https://github.com/owncloud/enterprise/issues/2367

guruz:
@SamuAlfageme the logic is: If you use a alpha/beta/rc currently, then you get more alpha/beta/rc.
;-)

ckamm:
Edited to make this into a gui setting. For nightlies etc. it'd default to the beta channel, as right now, for stable releases it'd default to stable.

ckamm:
@guruz So for instance the testcloud client shouldn't show the channels, and neither should any other branding that has an `APPLICATION_UPDATER_URL`?

guruz:
You are right, testpilotcloud should show it too.
Other brandings - Hm, we can't know yet so better hide it?

guruz:
Although for simplicity maybe just show it as soon as there is a ```APPLICATION_UPDATER_URL```

michaelstingl:
Not many branding users have own `APPLICATION_UPDATER_URL` configured.

SamuAlfageme:
@michaelstingl but there's also not that many branding users building their own (branded) beta versions. In the future, we could ease beta generation/distribution for them provided they have a URL to server them.

SamuAlfageme:
Maybe @ChrisEdS is interested on this topic, since IIRC he's going to run beta distribution for the iOS client, right?

ChrisEdS:
Oh, yes. I am an automation fan! I am currently working on my thesis with complete automation of all iOS tasks related to releases. Beta releases are already automated.

guruz:
Closing.
The next releases will have a drop down to switch the channel.
If you downloaded a alpha/beta/rc yourself, it's your own fault :)

PR #6493, opened 2018-04-26
Settings: Add update channel combobox

ckamm:
Open questions / things this doesn't do:
- Should it ever be hidden when the "Updater" group is shown?
- As a user I'd expect that when I switch the release channel I'll be offered to download the newest version from that channel. I think this important and I'll look at it.

Intended for picking into 2.4 once merged.

For  #6259

### DIFF
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -139,7 +139,7 @@ IF( APPLE )
     if(SPARKLE_FOUND)
        # Define this, we need to check in updater.cpp
        add_definitions( -DHAVE_SPARKLE )
-       list(APPEND updater_SRCS updater/sparkleupdater_mac.mm)
+       list(APPEND updater_SRCS updater/sparkleupdater_mac.mm updater/sparkleupdater.h)
    endif()
 ENDIF()
 
--- a/src/gui/generalsettings.cpp
+++ b/src/gui/generalsettings.cpp
@@ -130,7 +130,7 @@ void GeneralSettings::loadMiscSettings()
 
 void GeneralSettings::slotUpdateInfo()
 {
-    if (ConfigFile().skipUpdateCheck()) {
+    if (ConfigFile().skipUpdateCheck() || !Updater::instance()) {
         // updater disabled on compile
         _ui->updatesGroupBox->setVisible(false);
         return;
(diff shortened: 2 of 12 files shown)


ogoffart:
One problem is that the users are already compleining that the settings dialog is too tall. And I have not seen the result, but I suspect only makes the settings dialog taller.

ckamm:
@ogoffart Indeed a problem. I ended up not adding another row to the "Updates" groupbox so the effect should be minor. We do definitely need to rework the settings ui to not make this a problem.

guruz:
This is how would look (if SparkleUpdater was an OCUpdater):
screen shot 2018-04-26 at 14 50 41

guruz:
> One problem is that the users are already compleining that the settings dialog is too tall.

@ogoffart The idea was to move the complete About-section into a new window that is accessible as "About" item from tray menu. Do you want to do this? (independant of the fact that @ckamm didnt increase the size)

guruz:
@ckamm Pushed showing the UI on macOS.
The state won't be shown, we can improve this at another commit (not for 2.4.2)

screen shot 2018-04-27 at 09 32 25

guruz:
^^ FYI @michaelstingl @SamuAlfageme

michaelstingl:
Where is this located?

@SamuAlfageme put in your planned "advanced" section?

SamuAlfageme:
@michaelstingl it's in the "About" section - last HBox in there is shown for the clients with the updater enabled when it has meaning (notify about updates - e.g. not with Sparkle).

And yes, it's one of those ideas to include on a future, leaner "Settings" redesign.

ckamm:
There was a bunch of discussion and obstacles around the question of what going from `beta` to `stable` would do.
- On OSX downgrading isn't possible
- Downgrading is potentially unsafe everywhere

A discussed behavior would be:
- Move from `stable` to `beta`: Show an extra message explaining the details, including not being able to return to stable immediately (as a link to documentation)
- When going from `beta` to `stable`: Explain that the change won't take effect immediately, but only when a newer `stable` version comes out that comes after the currently installed version.

Given these constraints, it doesn't seem to me like this is something regular users would want to do since there's a chance they'll be on a broken `beta` version without the ability to go back. @SamuAlfageme Could you explain the usecase to me again?

guruz:
@ckamm I think you are overthinking this. I think it's fine how it works now. Users will understand that they need to download a stable version manually if they want to go back.

Issue #6504, opened 2018-05-02
Ignore list editor breaks placeholder backwards/forwards compatibility


ckamm:
In 2.5. we write this to the *user's* ignore file:
```
#!version < 2.5.0
*.owncloud
```
to guard against using placeholders, downgrading to 2.4, and then uploading all these placeholder files.

However, if the user uses the ignore list editor at any time, this breaks because the `#!version` tag isn't preserved. And we can't teach the 2.4 editor to preserve it.

Particular failing case:
* use placeholders on 2.5
* downgrade to 2.4, all is good
* use ignore editor on 2.4, the version directive is wiped but all is good (unless the placeholder ignore entry is explicitly deleted)
* upgrade to 2.5 again - the version directive is gone and placeholder files are ignored

Discovered when thinking about #6498, which is related.

ckamm:
To solve this particular issue against the 2.4 installs in the wild is hard. I'd suggest storing placeholder-using folders in a way that's invisible to <2.5 clients?

For the future, each folder binding and account should have a `version` flag and by default older clients should warn, backup the config, and ignore entries that are too new for them. Then we can simply bump the version for features that aren't backwards-compatible.

ckamm:
Also related to #6506

PR #6506, opened 2018-05-02
Placeholder and general backwards compatibility

ckamm:
For #6504 

* Store placeholder folders differently, so 2.4 clients don't see them at all
* Add mechanism to do backwards compatibility in a smarter way next time
* Remove silly "stub" body of placeholders

Open question:
* Remove the placeholder ignore pattern that was added for backwards compatibility but doesn't really work?

@SamuAlfageme

### DIFF
--- a/src/gui/accountmanager.cpp
+++ b/src/gui/accountmanager.cpp
@@ -33,6 +33,10 @@ static const char caCertsKeyC[] = "CaCertificates";
 static const char accountsC[] = "Accounts";
 static const char versionC[] = "version";
 static const char serverVersionC[] = "serverVersion";
+
+// The maximum versions that this client can read
+static const int maxAccountsVersion = 2;
+static const int maxAccountVersion = 1;
 }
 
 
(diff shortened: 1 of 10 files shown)


ckamm:
The version flags I'd cherry-pick to 2.4.2

ogoffart:
Test is faling.

ogoffart:
Why do we wipe the "future" settings instead of just ignoring them and living them there?

ckamm:
@ogoffart On wiping settings: Yes, for individual accounts and folders in the list we could just ignore them; and that may be convenient. It wouldn't work for the version on the accounts/folders structure itself. 

I definitely want the warning to let users know what's happening. But if we also keep some of the too-new config in the file, we need to make sure to only show it once. Maybe the whole config file should have a client-version-id string and if that's the same as the current version we don't need to show a migration note?

guruz:
I think this was not an ideal way to do it.
We could have introduced special new $ACCOUNTNUMBER\placeHolderFolders\$FOLDERNUMBER\ instead.
Too late..

PR #6511, opened 2018-05-08
Support icotool in ECMAddAppIcon macro

dschmidt:


SamuAlfageme:
The overall aspect of the icons (application, desktop shortcut, folders on the cloud provider integration, virtual files association, etc.) - looks nice up until "`Extra large icons`" view option

virtualbox_msedge - win10_04_06_2018_11_13_07

The icon in the taskbar is the only one that still looks a bit pixelated on a normal display though.

guruz:
If you think it's worth it, you can create a followup bug

dschmidt:
FTR it's not related to the icon set for the application executable (Windows) or bundle (macOS) which this PR was about.

Issue #6516, opened 2018-05-11
Connection closed when unicode is present in filename


JetUni:
### Expected behaviour
The file should sync with the server, even if there is a Unicode character (specifically an emoji) in the filename
### Actual behaviour
There is an error "Connection closed" and the syncing stops entirely. It won't even move on to the next files. At the very least it should move on and try that file again before it finishes syncing, but instead it just stops everything.
### Steps to reproduce
1. Place a bunch of files in a sync directory and make sure that one file has a Unicode character (I haven't tested with anything other than emojis) in the filename
2. Let the client try to sync the directory
3. Notice that once the sync gets to the file with a Unicode character, it stops and doesn't sync any other files

### Server configuration
Operating system:
Ubuntu 16.04

Web server: 
Apache

Database:
MySQL

PHP version:
7.0.28

ownCloud version:
10.0.8.5

Storage backend (external storage):
Stored locally, Cached with Redis

### Client configuration
Client version:
2.4.1

Operating system:
Windows 10

OS language:
English

Installation path of client:
A:\C\Program Files (x86)\ownCloud

### Logs

1. Client logfile:
```
05-11 11:17:07:531 [ info gui.folder ]:	#### Propagation start ####################################################
05-11 11:17:07:532 [ info gui.application ]:	Sync state changed for folder  "https://my.jarrett.tk/remote.php/dav/files/jarrett/Docs" :  "Sync Running"
05-11 11:17:09:787 [ info sync.propagator ]:	Starting INSTRUCTION_NEW propagation of "Backups/Computers/BEAST-Laptop/Downloads/Video/Relive the best of Team USA at Rio ???? ??.MP4" by OCC::PropagateUploadFileV1(0x136f5f78)
05-11 11:17:09:787 [ info sync.checksums ]:	Computing "SHA1" checksum of "A:/Documents/Backups/Computers/BEAST-Laptop/Downloads/Video/Relive the best of Team USA at Rio ???? ??.MP4" in a thread
05-11 11:17:09:799 [ info sync.propagator.upload ]:	"/Docs/Backups/Computers/BEAST-Laptop/Downloads/Video/Relive the best of Team USA at Rio ???? ??.MP4" "SHA1:0bbde185b00fee7d8cc4b10736d8009a37bc12d4"
[...]

PR #6523, opened 2018-05-14
Cleanup upload abort, fix job counting

ckamm:
Relates to #6516 but doesn't fix the issue.

In general: Wouldn't it simplify things if we emitted `abortFinished()` unconditionally, possibly also had the signals connected unconditionally?


guruz:
Maybe there was a reason for this before?
@jturcotte @ogoffart

Issue #6522, opened 2018-05-14
Autostart on the client breaks transparent OAuth's token refresh 


SamuAlfageme:
Last Friday I was debugging an apparent OAuth2 issue with @Helios07 that turned out to be a problem with the auto-start feature of the client. 

When auto-start is enabled, the client seems to be **starting before the keychain/keyring service on the OS** and therefore, without access to the stored refresh token. This causes the client to start the authorize workflow from scratch. 

We detected this behavior on a couple of Linux distros (Ubuntu and Fedora) - but it might not limit to Linux only. 

### Actual behavior
Client pops up the "Authorize Desktop client..." browser window. 

### Expected behavior
Delay the client login up until the password-store service is (back) online. See also https://github.com/owncloud/client/issues/6440

### Steps to reproduce
1. Enable auto-start on the client
2. Log-in via OAuth2 to an account
3. Restart the OS

ogoffart:
I guess that's a duplicate of this one: 
https://github.com/owncloud/client/issues/4274

Not much we can do about it....  maybe add some seconds of timer on first attempt and do another attempt...

guruz:
Additionally yes. Viel hilft Viel ;-)

cleverer:
Not sure if this is the same issue or another one:
If I have no internet connection on my Mac and owncloud launches, everything is fine. But as soon as the internet comes back up, I have to reauthorize my OAuth2 accounts.
If I then quit the client and reopen it, it connects successfully again.

OS: macOS 10.13.5, Client: owncloud 2.4.1 (build 9367), Server: 10.0.8.5

SamuAlfageme:
@cleverer thanks for reporting! Yeah, it might be a different form of the same issue.

Just to be clear, you don't re-authorize the client in the browser right? Just reopening the client makes it work? That might point at https://github.com/owncloud/client/issues/6357 as real cause: the first unauthorized call races against the function that fetches the credentials from the keychain service. 

@ogoffart might be worth taking a look at this. Will try to reproduce

cleverer:
@SamuAlfageme Yes, I'm just closing the browser windows. Of course sync also continues normally when re-authorizing.
Its especially annoying, if you have a corporate WiFi which only connects when you are logged in to your User account. Often owncloud is already started by the point the WiFi connects.

I'm happy to open a new issue or contribute in any way I can, but I'm a bit unfamiliar with ownclouds workflows…

ogoffart:
[...]

Issue #4274, opened 2015-12-13
[KDE4] "You need to reconnect" on startup when kwallet starts slowly


PVince81:
### Order of events
1. Log in to KDE4
2. Sync client starts, shows message "You need to reconnect your account"
3. A few seconds later, kwallet prompt appears
4. Enter password in kwallet
5. Wait a long time / edit a file in sync folder to nudge sync client
### Expected behavior

Sync client tries again and succeeds
### Actual result

Sync client does not react, gave up already. The dialog still says I need to reconnect my session.
If I restart the sync client it will then properly use the existing kwallet session.
### Versions

openSUSE 13.2 x86_64
owncloud-client-2.1.0-1.1.x86_64 (from our official repos)
libqt5keychain0-0.5-4.1.x86_64
libqtkeychain0-0.5-8.1.x86_64
kwalletmanager-14.12.3-16.1
kdelibs4-4.14.9-24.2.x86_64

Please note that on my desktop machine I have a HDD that seems to be slow to start, so I hear it rattling for a while after logging in. This might cause the slow down and sometimes makes the kwallet password dialog appear too late after the sync client needed it.

guruz:
I wonder how does this behave with qtkeychain 0.6.1?

PVince81:
I haven't seen this in a while. But it could also be because I upgraded to Leap + Plasma 5 and it seems to boot up faster. It also means that I don't have the environment to reproduce this any more. :frowning:

PVince81:
@guruz unless you want to keep this open for further research, I'd suggest to simply close it.
If other people still report the issue, then reopen it ?

guoyunhe:
I can reproduce this in openSUSE Leap 42.1 with KDE Plasma 5. When I didn't type Kwallet password quickly enough, owncloud will popup a password dialog. Here are something like "Kwallet...DBus NoReply" in the dialog.

guoyunhe:
This is what I got:

![spectacle ej5199](https://cloud.githubusercontent.com/assets/5836790/15272650/9ebb320c-1a88-11e6-9622-6a68fedefcfb.png)

TheOneRing:
Should be fixed by https://github.com/owncloud/client/pull/6525

PR #6525, opened 2018-05-14
Credentials: Retry fetching from the keychain in case the keychain is still starting

ogoffart:
When owncloud is restored, at boot time, it might be started before the
crendential manager. So if we detect an error, wait 10 seconds and hopefully
it'd be loaded by then.

Issues: #4274, #6522


ogoffart:
CC @SamuAlfageme

ogoffart:
Note that this adds a 10 seconds delay at every start of owncloud for people who really do not have a password manager.

guruz:
cherry-picked into 2.4 as 084f522de16aa839cf0db12ba4ffbc754b8a0d7b
(The issue https://github.com/owncloud/client/issues/6522 was for 2.4)

PR #6530, opened 2018-05-16
SyncJournal: Check file existence even for open dbs #6049

ckamm:
With WAL mode sqlite seems to occasionally crash when the
underlying filesystem goes away.

Shall be picked to 2.4 afterwards.

### DIFF
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -272,6 +272,11 @@ bool SyncJournalDb::sqlFail(const QString &log, const SqlQuery &query)
 bool SyncJournalDb::checkConnect()
 {
     if (_db.isOpen()) {
+        if (!QFile::exists(_dbFile)) {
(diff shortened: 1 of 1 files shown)


guruz:
Tested, works

src/common/syncjournaldb.cpp
```diff
@@ -272,6 +272,11 @@ bool SyncJournalDb::sqlFail(const QString &log, const SqlQuery &query)
 bool SyncJournalDb::checkConnect()
 {
     if (_db.isOpen()) {
+        if (!QFile::exists(_dbFile)) {
+            qCWarning(lcDb) << "Database open, but file file" + _dbFile + " does not exist";
+            close();
+            return false;
+        }
```
dragotin: A comment explaining why this is here might make sense, because it looks pretty odd for somebody who does not know the history of this fix. This might easily look like a code left-over.

PR #6531, opened 2018-05-16
Rename "placeholder" feature

ckamm:
See code changes for concrete strings.


src/libsync/syncengine.cpp
```diff
@@ -854,7 +854,7 @@ void SyncEngine::startSync()
     _csync_ctx->new_files_are_placeholders = _syncOptions._newFilesArePlaceholders;
     _csync_ctx->placeholder_suffix = _syncOptions._placeholderSuffix.toUtf8();
     if (_csync_ctx->new_files_are_placeholders && _csync_ctx->placeholder_suffix.isEmpty()) {
-        csyncError(tr("Using placeholder files, but placeholder suffix is not set"));
+        csyncError(tr("Using virtual files but placeholder suffix is not set"));
```
SamuAlfageme: Missing a "placeholder" here

ckamm:
Renaming the code as well would make talking about the feature more consistent. But I'd only do it if it's really likely another name change is unlikely. In any case, the number of things that would need to be renamed is small. I might just do that.

PR #6534, opened 2018-05-16
Wizard: Add a "(recommended)" to the main option

ckamm:
@jancborchardt 

This will lead to a double parenthesis: "(recommended) (124 MB)". It'd be easy to make it say "(recommended, 124 MB)" if that seems better. (note that the size is not always available)

For #6470


jancborchardt:
@ckamm putting it in only one parentheses would be nice and simpler, yes! :) Also not sure if this is standard, but what might be nice in addition is to bold that default entry?

PR #6541, opened 2018-05-22
Add context menu actions for conflicts and new read-only dir files

ckamm:
See #6252

By default I'd target this for 2.6, but could also still go into 2.5 if desired. Similar (and more) actions should be available from the conflicts overview in the issues tab, but I didn't start that yet.


src/gui/socketapi.cpp
```diff
@@ -747,12 +814,58 @@ void SocketApi::command_GET_MENU_ITEMS(const QString &argument, OCC::SocketListe
...
+                        if (isOnTheServer) {
+                            // Uploaded conflict file in read-only directory
+                            listener->sendMessage(QLatin1String("MENU_ITEM:MOVE_ITEM::") + tr("Move and rename..."));
+                        } else {
+                            // Local-only conflict file in a read-only dir
+                            listener->sendMessage(QLatin1String("MENU_ITEM:MOVE_ITEM::") + tr("Move, rename and upload..."));
+                        }
+                    }
+                    listener->sendMessage(QLatin1String("MENU_ITEM:DELETE_ITEM::") + tr("Delete local changes"));
+                    // listener->sendMessage(QLatin1String("MENU_ITEM:DELETE_ITEM::") + tr("Promote to main file"));
```
ogoffart: I guess this comment can be removed.

ogoffart:
Also, I would expect an entry "Discard remote change, restore local change" when right clicking on a conflict file, which would rename the conflict file to the original name.  Not?

ckamm:
Thanks for the comments! I agree about the action and that's what I was going for with the "promote to main file" action. But that is for another PR I think.

guruz:
(Just phoned with @ogoffart .. let's not have this in the alpha until it's tested better by @SamuAlfageme .. or even move to 2.6)

ckamm:
@ogoffart ok for merge to `master` now that it's diverged from `2.5`?

ogoffart:
@ckamm  merge?

PR #6542, opened 2018-05-23
Update libcrashreporter-qt to build it with msvc

dschmidt:


dschmidt:
As this is a pretty trivial change and it block me right now, I'm merging without external review.

If it's causing any issues, it should be possible to revert the commit the submodule is pointing to without making the submodule point to guruz' repository again.

The issue is that craft has a bug that prevents checking out a submodule in a branch from a different remote than the one that is tracked in master - and I need this fix to continue...

SamuAlfageme:
Just tested it by reproducing https://github.com/owncloud/client/issues/6562 and worked flawlessly 🎉

Issue #6466, opened 2018-04-18
Bundled download of placeholders inside folder(s)


SamuAlfageme:
I see a common use-case where a user might want to "materialize" (i.e. download and sync) all the files inside a folder. 

It's too demanding in terms of user-interaction to go file-by-file fetching every placeholder from a directory. Also consider that in this case, we don't want the usual "download and forward the real file to the program responsible for its opening" - we'll easily end up with one "photo viewer" window per picture inside the `"Holidays 2017"` folder.  

Getting some sort of context menu entry/switch for folders (and groups of folders as well, maybe?) will ease up such workflow. 

cc/ opinions @ckamm @ogoffart

ckamm:
I agree that clicking on a folder and saying "download this and all contents, recursively" would be nice. I also thought about it, but it didn't make the "first version" cut. :)

SamuAlfageme:
This is so awesome 😍 

I can use my file explorer to do a funky selection of folders at different depths in the three and get their contents sync'ed on my disk. yay!

PR #6551, opened 2018-05-28
Virtual Files: Allow to download a folder recursively from the socketAPI

ogoffart:
Issue: #6466

### DIFF
--- a/src/gui/socketapi.cpp
+++ b/src/gui/socketapi.cpp
@@ -611,7 +611,7 @@ void SocketApi::command_DOWNLOAD_VIRTUAL_FILE(const QString &filesArg, SocketLis
     auto suffix = QStringLiteral(APPLICATION_DOTVIRTUALFILE_SUFFIX);
 
     for (const auto &file : files) {
-        if (!file.endsWith(suffix))
+        if (!file.endsWith(suffix) && !QFileInfo(file).isDir())
             continue;
         QString relativePath;
         auto folder = FolderMan::instance()->folderForPath(file, &relativePath);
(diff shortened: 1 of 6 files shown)


ogoffart:
(I think this could still go in 2.5...  otherwise we can just re-target)

ogoffart:
This allow to add the "Download files" menu entry for a folder in a sync directory with virtual files. 
This will download all the virtual files.
Note that future files added in that folder on the server will still be virtual files.

cc @hodyroff

src/gui/socketapi.cpp
```diff
@@ -757,11 +757,11 @@ void SocketApi::command_GET_MENU_ITEMS(const QString &argument, OCC::SocketListe
     }
 
     // Virtual file download action
-    if (folder) {
+    if (folder && folder->useVirtualFiles()) {
```
ckamm: Hm, we might want to rename/document this better. Actually, this setting just controls whether new files will start out as virtual or not. It's perfectly possible to switch this of (so all new files are no longer virtual) but still have heaps of virtual files in the sync directories.

Short: I'd show this option even for folders that don't have the option enabled.

ckamm:
I'm fine with this going into 2.5 since it just adds a new option to a feature that's experimental in the first place

src/gui/socketapi.cpp
ogoffart: Right, in the future, we could decouple the two.
The problem is that I don't really want to recursively go into the folder to see if there are placeholders.
(altough maybe a SELECT WHERE type=4 AND IS_PREFIX_PATH_OF might be simple enough)

ckamm: What about moving the check down to the files loop then and only removing the action from directories when the folder setting is disabled? Then people could at least still download files individually if the option is off.

Issue #6461, opened 2018-04-17
Placeholder-aware sharing from the context menu


SamuAlfageme:
When sharing a file from its placeholder counterpart, use the real name for across the sharing dialog  instead of the `.owncloud` file living on the real FS:

- as `path`for private sharing / public links 
- to query the `fileid` / `privatelink` via PROPFIND
- to fetch the thumbnail (in case there's one)

```
curl  \
    -H 'Ocs-APIREQUEST:true' \
    -H 'Content-Type:application/x-www-form-urlencoded' \
    -H 'User-Agent:Mozilla/5.0 (Windows) mirall/2.5.0placeholderprototype (build 9468)' \
    [...]
    -X POST 'https://demo.owncloud.com/ocs/v1.php/apps/files_sharing/api/v1/shares?format=json' \
    --data-binary 'path=%2FownCloud%20Manual.pdf.owncloud&shareType=3&name=Public%20link'
```

<p align="center">
  <img src="https://user-images.githubusercontent.com/2644445/38867493-830cb432-4244-11e8-941d-8fd5cde9e825.png"/>
</p>

A user might want to create a share link or copy its private link wo/ having to download the file.

SamuAlfageme:
Working like a charm! 🎉 

@ogoffart was also backported to 2.5 right?

ogoffart:
(yes, this is in 2.5)

PR #6552, opened 2018-05-28
Share dialog: use the original name and not the virtual file name

ogoffart:
When sharing a virtual file, we should actually use the original file name
not the virtual file name

Issue: #6461

### DIFF
--- a/src/gui/socketapi.cpp
+++ b/src/gui/socketapi.cpp
@@ -704,7 +704,10 @@ SocketApi::FileData SocketApi::FileData::get(const QString &localFile)
         return data;
 
     data.accountRelativePath = QDir(data.folder->remotePath()).filePath(data.folderRelativePath);
-
+    QString virtualFileExt = QStringLiteral(APPLICATION_DOTVIRTUALFILE_SUFFIX);
+    if (data.accountRelativePath.endsWith(virtualFileExt)) {
+        data.accountRelativePath.chop(virtualFileExt.size());
+    }
     return data;
 }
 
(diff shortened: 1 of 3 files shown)


src/gui/socketapi.cpp
```diff
@@ -704,7 +704,10 @@ SocketApi::FileData SocketApi::FileData::get(const QString &localFile)
         return data;
 
     data.accountRelativePath = QDir(data.folder->remotePath()).filePath(data.folderRelativePath);
-
+    QString virtualFileExt = QStringLiteral(APPLICATION_DOTVIRTUALFILE_SUFFIX);
+    if (data.accountRelativePath.endsWith(virtualFileExt)) {
+        data.accountRelativePath.chop(virtualFileExt.size());
+    }
```
ckamm: It seems confusing to me to do this operation to the account-path but not to the folder-path. 

It's indeed likely that the account-path will be used in server queries and the like while the folder path wouldn't be - but one of them refering to the virtual file while the other doesn't might cause hard to spot issues.

ogoffart: Right, in the future, we could decouple the two.
The problem is that I don't really want to recursively go into the folder to see if there are placeholders.

(altough maybe a SELECT WHERE type=4 AND IS_PREFIX_PATH_OF  might be simple enough)

ogoffart: (Note: I removed the previous comment because it was not applying here.)

I think it make sense to have the server path relate to the file on the server.
Maybe I should specify it in the code comments.

ckamm: If you rename it serverPath that's fine

Issue #6555, opened 2018-05-29
2.5.0alpha: `make doc-man` fails


dragotin:
tarball https://download.owncloud.com/desktop/testing/owncloudclient-2.5.0-alpha.tar.xz

When building, the the command `make doc-man` to create the man pages bails out and breaks the build.

The reason is that in the current 2.5.0 the directory `doc` does not contain a file `conf.py`, which was there in 2.4.1 still. It is actually correct that the file is NOT there, because it is supposed to be crated by cmake out of `conf.py.in` which is also there. 

However, for some reason cmake does not do that. Probably it also did not in 2.4.1 but that bug was shadowed because the conf.py was still there. 

Please fix that.

ogoffart:
Works for me.
But you do need sphinx installed.
Can you see this line when runing cmake?

```
-- Found Sphinx: /usr/bin/sphinx-build  
```

Maybe there is a problem in FindSphinx.cmake as it does not have good default location to find it.

ogoffart:
Maybe a shadow build problem then.  The conf.py is probably in the build directory.

dragotin:
please see what I wrote above: the tarball of 2.4.1 contained a conf.py. In the 2.5.0 tarball it is correctly removed. We did not see the problem in 2.4.1 because of that, but it was probably there.

Building with OBS (what I do)  guarantees clean build environments btw.

SamuAlfageme:
@dragotin I removed `conf.py` in https://github.com/owncloud/client/pull/6505/commits/2c53025b0aed7c1adffc8f66be92d0c80f92dc11 as it seemed to be a stale file from a build that was pushed to the repo. accidentally (e.g. there were hardcoded values that sphinx used instead of the right env.vars) as you mentioned.

Are you running CMake on the parent (i.e. client) directory or the local `client/doc`?  I probably missed some switch to enable generating `doc/Makefile` and `doc/conf.py`

ogoffart:
@dragotin  https://github.com/owncloud/client/pull/6556 should fix it. Can you try?

dragotin:
@ogoffart yes, #6556 fixes it. Thanks!

guruz:
@dragotin thank you for reporting

PR #6556, opened 2018-05-29
Fix make doc-man in shadow build

ogoffart:
Issue #6555


guruz:
@dragotin can you confirm this fixes your issue?

Issue #6462, opened 2018-04-17
Do not unfold "Sync Protocol" placeholder entries in deletion + real file download


SamuAlfageme:
![screen shot 2018-04-17 at 13 52 48](https://user-images.githubusercontent.com/2644445/38868251-b66f7790-4246-11e8-9f11-31baa85fa06c.png)

This will also avoid extra noise in terms of notifications (1 per "Sync Protocol" entry).

ckamm:
Sounds good!

SamuAlfageme:
The virtual files flow (i.e. delete `.owncloud` -> download original) is no longer reflected like this. Can be closed as no longer confuses the user with "Deleted" actions. 

However, there's still room for improvement as when virtual and real file are fetched on the same client run, they'll appear as "duplicated" downloads in there:

![placeholder_activity](https://user-images.githubusercontent.com/2644445/41219164-cde137be-6d5d-11e8-93e2-8cd8be40656f.png)

The desired entries in the Activity for a "_seamless VFS experience_" need to be properly defined by PM. cc/ @pmaier1 @hodyroff

PR #6558, opened 2018-05-30
Virtual files: Only remove virtual file once on download

ckamm:
With thanks to @ogoffart for spotting the problem.


ogoffart:
This should remove the spurious deleted notification and entry of the log which is mentioned in #6462

PR #6559, opened 2018-05-30
Virtual files: Wipe virtual after download completes, not before

ckamm:
Otherwise a interrupted or unsuccessful download would mean that the
download-intend was forgotten. The next sync would reestablish the
virtual file instead.

### DIFF
--- a/test/testsyncvirtualfiles.cpp
+++ b/test/testsyncvirtualfiles.cpp
@@ -354,6 +354,60 @@ private slots:
         QVERIFY(!dbRecord(fakeFolder, "A/a6.owncloud").isValid());
     }
 
+    void testVirtualFileDownloadResume()
+    {
+        FakeFolder fakeFolder{ FileInfo() };
+        SyncOptions syncOptions;
+        syncOptions._newFilesAreVirtual = true;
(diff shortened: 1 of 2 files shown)


test/testsyncvirtualfiles.cpp
```diff
@@ -354,6 +354,60 @@ private slots:
...
+        QCOMPARE(fakeFolder.currentLocalState(), fakeFolder.currentRemoteState());
+        QSignalSpy completeSpy(&fakeFolder.syncEngine(), SIGNAL(itemCompleted(const SyncFileItemPtr &)));
+
+        auto cleanup = [&]() {
+            completeSpy.clear();
+            fakeFolder.syncJournal().wipeErrorBlacklist();
+        };
+        cleanup();
+
+        auto triggerDownload = [&](const QByteArray &path) {
```
ogoffart: (nitpick:  since this is used in many tests, we should maybe have a `static void triggerDownload(FakeFolder&, const QByteArray&)` at the begining of that file.)

ogoffart:
I guess jenkins is failing because it depends on #6558

ckamm:
Jenkins: Indeed, all tests passed locally for me

Issue #6459, opened 2018-04-17
Use the "s:message" from erroneous requests on their "Issue" column


SamuAlfageme:
I was just testing the fix in https://github.com/owncloud/core/issues/29794#issuecomment-381880701 and realized we could rethink the way client/server errors get displayed:

![screen shot 2018-04-17 at 09 29 48](https://user-images.githubusercontent.com/2644445/38855393-a17dea92-4223-11e8-9cae-eb2eccc1e20f.png)

The "Issue" on  the "Not Synced" tab reads as follows: 

```
Server replied "403 Forbidden" to "GET https://alfageme.solidgear.prv/remote.php/dav/files/admin/mr-nobody-yify-english.srt"
```

Which does not contain the actual reasons for the issue. On the body of the response we have more info of the actual problem: 

```
<?xml version="1.0" encoding="utf-8"?>
<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
  <s:exception>Sabre\DAV\Exception\Forbidden</s:exception>
  <s:message>Encryption not ready: Private Key missing for user: please try to log-out and log-in again</s:message>
</d:error>
```

Replacing the full WebDAV path that failed (not that useful since the user already has this info in the "File" entry) by the `s:message` on the reply would be more useful and somehow user-friendly. 

Of course, if said response does not contain such `s:message`, we can use the old logic as fallback.

ogoffart:
Normally this was already supposed to be the case. AbstractNetworkJob::errorStringParsingBody  is supposed to extrace the Message.
So this is not working as expected.

SamuAlfageme:
@ogoffart can't see this working on latest nightly, might be missing something 😕 

With the steps from my original https://github.com/owncloud/client/issues/6459#issue-314945381 (the one for fetching files with oauth2 + user-key encryption enabled https://github.com/owncloud/oauth2/issues/105) I'm still seeing the long "issue" text.

SamuAlfageme:
Seen w/ [files firewall](https://marketplace.owncloud.com/apps/firewall) as well - in that case; there's a bogus timer as well:

```
Server replied "403 Forbidden" to "GET https://alfageme/remote.php/dav/files/admin/Documents/Example.odt" (skipped due to earlier error, trying again in 9 second(s))
```

i.e. does not update reactively.

guruz:
blue-ticket, moved into milestone

SamuAlfageme:
[...]

PR #6561, opened 2018-05-30
PropagateDownload: Don't discard the body of error message

ogoffart:
We want to keep the body so we can get the message from it
(Issue #6459)

TestDownload::testErrorMessage did not fail because the FakeErrorReply
did not emit readyRead and did not implement bytesAvailable.

### DIFF
--- a/src/libsync/propagatedownload.cpp
+++ b/src/libsync/propagatedownload.cpp
@@ -165,6 +165,9 @@ void GETFileJob::slotMetaDataChanged()
     // If the status code isn't 2xx, don't write the reply body to the file.
     // For any error: handle it when the job is finished, not here.
     if (httpStatus / 100 != 2) {
+        // Disable the buffer limit, as we don't limit the bandwidth for error messages.
+        // (We are only going to do a readAll() at the end.)
+        reply()->setReadBufferSize(0);
         _device->close();
         return;
     }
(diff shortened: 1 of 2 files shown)


guruz:
Sorry, I don't understand this fix.
1) Shouldnt the test need some changes too if it did not fail without your fix? 
2) Any side effects to _saveBodyToFile? Sounds dangerous and to me like you could end up writing the error message into the destination file? At least add some commenting then?

ogoffart:
> Sorry, I don't understand this fix.

The line
```
        qint64 r = reply()->read(buffer.data(), toRead);
```

Reads the data, but only saves it to a file if `_saveBodyToFile` which is only true for the 200 replies. Otherwise the content is discarded.
This is a problem because `AbstractNetworkJob::errorStringParsingBody` need the body to be there to extract the message (it does `reply()->readAll()`)

> Shouldnt the test need some changes too if it did not fail without your fix?

Yes, the test was changed!

> Any side effects to _saveBodyToFile? 

I hope not. But that's why we review. This code is all about the bandwidth limiting.
One problem I foresee is that if the body is bigger than all the internal buffer (which we try to limit to 16KiB) is that this would block the request completely, resulting in a timeout error instead. But normally, errors request should not be too big.

> Sounds dangerous and to me like you could end up writing the error message into the destination file? At least add some commenting then?

No that can't happen for sure. It is the other way around. Since it is in that function that we write in the file

src/libsync/propagatedownload.cpp
```diff
@@ -268,7 +268,7 @@ void GETFileJob::slotReadyRead()
     int bufferSize = qMin(1024 * 8ll, reply()->bytesAvailable());
     QByteArray buffer(bufferSize, Qt::Uninitialized);
 
-    while (reply()->bytesAvailable() > 0) {
+    while (reply()->bytesAvailable() > 0 && _saveBodyToFile) {
         if (_bandwidthChoked) {
             qCWarning(lcGetJob) << "Download choked";
             break;
```
ckamm: there is a `_saveBodyToFile` in a condition inside this `while` further down, which is now redundant.

ogoffart:
@ckamm  commit updated

Issue #6562, opened 2018-05-30
Segfault when clicking on unconfigured SFTP server


lefherz:
### Expected behaviour
The ownCloud settings view should display a proper error message for unconfigured SFTP folders.

### Actual behaviour
The client segfaults when trying to display the subfolders of the unconfigured SFTP folder.

### Steps to reproduce
1. unproperly configure SFTP external storage
2. open testpilot
3. in the account view, click on the triangle left to the SFTP folder
4. wait for the client to segfault

### Server configuration

demo.owncloud.org

### Client configuration
Client version:
```
testpilotcloud version 2.5.0alpha1 (build 9564)
Git revision 107baf2947959cd61f45951f7dbcef2f61bb9ab3
Using Qt 5.10.1, built against Qt 5.10.1
Using 'OpenSSL 1.0.2n  7 Dec 2017'
```

Operating system: Xubuntu 18.04

OS language: english

Client package (From ownCloud or distro) (Linux only): testpilotcloud-client

Installation path of client: /usr/bin/testpilotcloud

### Logs

Backtrace: https://gist.github.com/lefherz/7daf1b2bb6f5fc8eeb15f1ef79a0429e

lefherz:
Configuration of SFTP:

![screenshot_2018-05-30_18-47-08](https://user-images.githubusercontent.com/17278272/40735437-79c87924-643b-11e8-8d16-fd0b4aafb73a.png)

jnweiger:
does not crash when running under gdb. Heisenbug...

jnweiger:
@hodyroff maybe related to your SFTP issue?

SamuAlfageme:
Can reproduce. The moment the client detects the mounted folder I see in the logs: 

```
05-31 09:21:08:951 [ debug sync.discovery ]	[ OCC::DiscoveryMainThread::singleDirectoryJobFirstDirectoryPermissionsSlot ]:	Permissions for root dir: "DNVCKR"
05-31 09:21:08:952 [ warning sync.discovery ]:	Missing properties: "SFTP" 2 0 1527751269 "M" "" "-0000001oc0v33t8034l"
05-31 09:21:08:952 [ debug sync.discovery ]	[ OCC::DiscoveryMainThread::singleDirectoryJobFinishedWithErrorSlot ]:	10011 "The server file discovery reply is missing data."
[...]

Issue #5859, opened 2017-06-26
Never ever abort sync runs / check that the connection is indeed broken before aborting on a timeout


michaelstingl:
Some server side errors cause the client to abort the current sync run. If the ownCloud admin or user is not able to fix the error, EVERY sync run will abort at the same position, and following files that don't have an error will NEVER sync because of this.

@SamuAlfageme We discussed this a time ago. Please link existing issues here or open new ones that could could cause this behaviour. Please also elaborate on the server improvements.

guruz:
Please report only if this happens with latest client version as this is something which changed a lot back and forth :-)
(CC @ckamm @ogoffart )

SamuAlfageme:
**WIP**

## Sync aborted - on server/network error

- [...]

## Fixed 

- [X] Possibility to deselect a folder (from the selective sync list) with errors that cause sync aborts - https://github.com/owncloud/client/pull/5772

ckamm:
Incomplete list of potential sync run aborts during propagation stage, by looking at `FatalError` in the code
- can't write to database
- network error > `NoError` and <= `UnknownProxyError`
- network error 503
- timeout on download
- critical disk space error on download

ogoffart:
I believe the current behavior is the correct one:  We only abort the sync run if we experience fatal error  (network down, server on maintainnence, ...)

@michaelstingl  Do you have any concrete example of error that causes the sync to abort while it should not?

cdamken:
> I believe the current behavior is the correct one: We only abort the sync run if we experience fatal error (network down, the server on maintenance, ...)

Wenn a user has some shares and/or has some external storages and/or federated shares and/or some inconsistency in the oc_filecache  because the database has a problem (in some cases is normally fixed with `occ files:scan` to the user, the user that shares, etc)  the synclient just stops on the error and does not sync the new files and folders (even if they are in a different path)

IMO, the expected behavior, the sync-client should mark those files, and try to upload the new files (because those files are probably needed for someone else) and at the end send a summary the main errors.

Like: 
files x, y, z in folder w could not be downloaded because the server reported: connection timeout. ->Contact your administrator 
files a, b, c could not be uploaded because chunks weren't received correctly, we will try to upload later. -> if the problem persist please contact your administrator.
files m,n,o were successfully uploaded. -> Synced after the errors occurred.

ogoffart:
> files x, y, z in folder w could not be downloaded because the server reported: connection timeout. ->Contact your administrator

[...]

PR #6563, opened 2018-05-31
Crash fixes and valgrind error fix

ogoffart:
Fix the crash reported in #6562

Also cleanup another warning reported by valgrind.

See individual commit message for more info.

### DIFF
--- a/src/gui/folderstatusmodel.cpp
+++ b/src/gui/folderstatusmodel.cpp
@@ -740,10 +740,11 @@ void FolderStatusModel::slotLscolFinishedWithError(QNetworkReply *r)
     if (parentInfo) {
         qCDebug(lcFolderStatus) << r->errorString();
         parentInfo->_lastErrorString = r->errorString();
+        auto error = r->error();
 
         parentInfo->resetSubs(this, idx);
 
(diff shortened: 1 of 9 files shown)


SamuAlfageme:
Client no longer crashes with the path - couple of concerns tho: 

- Even though this is a corner case, UX is not the best: 

screenshot 2018-05-31 14 56 41

- **This "HTTP transmission error" aborts sync (https://github.com/owncloud/client/issues/5859) shouldn't be better to just blacklist/uncheck from selective sync those file/folders wo/ essential csync attrs. like e.g. ETAG?**

src/gui/folderstatusmodel.cpp
```diff
@@ -1226,7 +1227,10 @@ bool FolderStatusModel::SubFolderInfo::hasLabel() const
 void FolderStatusModel::SubFolderInfo::resetSubs(FolderStatusModel *model, QModelIndex index)
 {
     _fetched = false;
-    delete _fetchingJob;
+    if (_fetchingJob) {
```
ckamm: Does this guarantee that no signals from the stale job will be emitted? Maybe disconnect them?

ogoffart:
(Added the disconnection before merging.)

PR #6567, opened 2018-06-02
Fix build without Sparkle.framework

yan12125:
When Sparkle.framework is not installed, building the latest git master results in the following error:
```
[ 39%] Linking CXX executable ../../bin/owncloud.app/Contents/MacOS/owncloud
Undefined symbols for architecture x86_64:
  "OCC::SparkleUpdater::setUpdateUrl(QUrl const&)", referenced from:
      OCC::GeneralSettings::slotUpdateChannelChanged(QString const&)::$_0::operator()(int) const in generalsettings.cpp.o
  "OCC::SparkleUpdater::statusString()", referenced from:
      OCC::GeneralSettings::slotUpdateInfo() in generalsettings.cpp.o
  "OCC::SparkleUpdater::staticMetaObject", referenced from:
      OCC::SparkleUpdater* qobject_cast<OCC::SparkleUpdater*>(QObject*) in generalsettings.cpp.o
ld: symbol(s) not found for architecture x86_64
```
Adding two HAVE_SPARKLE guards fixes the issue.

### DIFF
--- a/src/gui/generalsettings.cpp
+++ b/src/gui/generalsettings.cpp
@@ -148,7 +148,7 @@ void GeneralSettings::slotUpdateInfo()
         _ui->restartButton->setVisible(ocupdater->downloadState() == OCUpdater::DownloadComplete);
 
     }
-#ifdef Q_OS_MAC
+#if defined(Q_OS_MAC) && defined(HAVE_SPARKLE)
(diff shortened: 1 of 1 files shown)


ogoffart:
Thanks for the patch.  Merged manually in 2.5  (9e26b0f)

Issue #6524, opened 2018-05-14
ownCloud-nemo not working anymore after installing testpilotcloud-nemo


lefherz:
### Expected behaviour
testpilotcloud-client + testpilotcloud-client-nemo can be installed next to owncloud-client + owncloud-client-nemo, and the file explorer overlays are usable for both separate sync folders.

### Actual behaviour
After I installed testpilotcloud, owncloud-client-nemo stopped working. you don't see the file explorer overlays anymore for the owncloud sync folder. they don't reappear after stopping the testpilotcloud process, or even rebooting.

### Steps to reproduce
1. sudo apt install owncloud-client owncloud-client-nemo
2. log in with owncloud-client and create a sync folder
3. sudo apt install testpilotcloud-client testpilotcloud-client-nemo
4. (log in with testpilotcloud-client and create a sync folder) <- maybe optional

### Server configuration

demo.owncloud.org

### Client configuration
Client version: 2.4.1

Testpilotcloud version: 2.5.0daily20180514

Operating system: Xubuntu 18.04

OS language: english

Qt version used by client package (Linux only, see also Settings dialog): 5.6.2

Qt version used by testpilotcloud package: 5.10.1

Client package (From ownCloud or distro) (Linux only): all packages from ownCloud repos

Installation path of client: both /usr/bin

michaelstingl:
related: https://github.com/owncloud/client/issues/5976

@jnweiger could you have a look?

lefherz:
I fixed it locally:

1. uninstalling testpilotcloud-client-nemo
2. restarting owncloud process
3. restarting nemo process

Now it works for my ownCloud directory again - but not for my testpilotcloud directory, of course.

guruz:
@ogoffart @ckamm Please triage

guruz:
@lefherz Can you retest with out 2.5 nightlies?

jnweiger:
without?

ogoffart:
> @lefherz Can you retest with out 2.5 nightlies?

> without?

I guess he meant "our"

guruz:
I did mean 'our'

guruz:
2.5.0 beta1 is out :-)
https://central.owncloud.org/t/desktop-sync-client-2-5-0-beta1-released/14667
@lefherz  please comment here if we can close this issue. Thank you.

lefherz:
it worked after I killed and restarted the nemo process :) thanks a lot!

lefherz:
[...]

PR #6568, opened 2018-06-04
nautilus shell integration: Fix when there are several branded client installed

ogoffart:
It appears that several extension can be loaded at the same time, but their
classname for the extension need to be different, otherwise only the last
loaded one would be active.

Issue #6524

### DIFF
--- a/shell_integration/nautilus/setappname.sh
+++ b/shell_integration/nautilus/setappname.sh
@@ -3,4 +3,6 @@
 # this script replaces the line
 #  appname = 'ownCloud'
 # with the correct branding name in the syncstate.py script
-sed -i.org -e 's/appname\s*=\s*'"'"'ownCloud'"'/appname = '$1'/" syncstate.py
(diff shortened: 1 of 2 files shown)


guruz:
CC @payload @Naereen

shell_integration/nautilus/setappname.sh
```diff
@@ -3,4 +3,4 @@
 # this script replaces the line
 #  appname = 'ownCloud'
 # with the correct branding name in the syncstate.py script
-sed -i.org -e 's/appname\s*=\s*'"'"'ownCloud'"'/appname = '$1'/" syncstate.py
+sed -i.org -e "s/ownCloud/$1/g" syncstate.py
```
ckamm: The comment is now outdated. Also, minor, I think replacing a different identifier "APP_NAME" or something) would make problems with this replace less likely.

ckamm:
+1

Issue #6546, opened 2018-05-24
text on the gui becomes unreadable when connect failed


jnweiger:
First screenshot below shows the next screen of the gui after entering login credentials.
If the credentiasl were invalid, and user presses the [Connect...] button, the screen changes to the second screenshot below. 

![40486212-ef587c14-5f60-11e8-8737-82adebd2e20f](https://user-images.githubusercontent.com/1108546/41785493-112ec9a8-7643-11e8-8af1-b6bd8db8a117.png)

Note the lines get squeezed to make room for the pink error message.  Here everything is still readable, but sometimes one can no longer decipher the texts.

Seen on Kubuntu18.04 with Plasma desktop.

ckamm:
@jnweiger I'll check it out and will be able to reproduce the label appearing -- but how did you get to this screen with invalid credentials? That shouldn't happen as far as I know!

michaelstingl:
Server issue/hack is needed to get stuck with this error.

guruz:
@jnweiger fix is in alpha, feel free to test and close

SamuAlfageme:
Still reproducible in 2.5: 

![screenshot 2018-06-04 12 25 53](https://user-images.githubusercontent.com/2644445/40912647-7d142c12-67f2-11e8-86c7-c3e04eb87842.png)

Hints to reproduce: 

- Set up mitmproxy on the client and set it to intercept all (`.*`) requests
- Get to the "Connect to ownCloud" page on the wizard and click connect
- Hit `a` to accept the request, then enter to inspect it
- Hit tab to switch to `Response` and `e` to edit, then `o` to tamper the HTTP code to e.g. 500
- Forward the response to the client (`a` to accept it)

ckamm:
@SamuAlfageme Yes, this still happens for a multi-line error message, I had only tested with single-line errors.

The underlying problem is that this layout has no vertical wriggle room. So when an error message becomes visible and needs more space than expected there's not enough space left over all and the layout becomes squished. I'm checking whether I can give the middle spacer some shrinking flexibility.

ckamm:
@SamuAlfageme There's a new PR to check out.

jnweiger:
@michaelstingl @ckamm no server hackery needed. This was simply a client with preconfigured server url

jnweiger:
reproduce attempt with https://github.com/owncloud/client/pull/6589 results in a nice little scrollbar next to the error mssage, no more squeezed widgets.
Thanks!
[...]

PR #6575, opened 2018-06-11
Wizard: Increase vertical size to fit error text

ckamm:
After some complex attempts of adding vertical scroll bars if the size got out of hand, I opted to just increase the window size if necessary.

For  #6546

### DIFF
--- a/src/gui/wizard/owncloudadvancedsetuppage.cpp
+++ b/src/gui/wizard/owncloudadvancedsetuppage.cpp
@@ -104,6 +104,7 @@ bool OwncloudAdvancedSetupPage::isComplete() const
 void OwncloudAdvancedSetupPage::initializePage()
 {
     WizardCommon::initErrorLabel(_ui.errorLabel);
+    _ui.syncModeLabel->setVisible(false);
 
     if (!ConfigFile().showExperimentalOptions()) {
         // If the layout were wrapped in a widget, the auto-grouping of the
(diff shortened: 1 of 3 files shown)


SamuAlfageme:
@ckamm everything looks right code-wise but can't see the patch working on windows - still reproducing with the steps on https://github.com/owncloud/client/issues/6546#issuecomment-394312492 and with the spinner standalone 😕

ckamm:
@SamuAlfageme That's frustrating, it works for me on linux now. Does the wizard window increase in size at all when the error label becomes visible?

ogoffart:
I tried the patch on linux (Kwin),  and the patch does not seem to have any effect.
The window is not resized.

Also i'm wondering if we should not prevent the window to make the size smaller, or change the horizontal size.

(Note: for your testing, are you using branding? because branded client size the wizard differently as there is no pixmap.  (I was not using branding) )

SamuAlfageme:
@ogoffart nope, no branding involved:

wizard_resize_issues

ckamm:
My test setup used a different code path to setting errors - indeed nothing is done for network errors, my oversight. But even if I make the resize happen, the checkboxes are still squished together in my windows tests. I think this does need something that scrolls.

guruz:
Superseded by https://github.com/owncloud/client/pull/6589 ?

ckamm:
Indeed.

Issue #6554, opened 2018-05-29
Windows: Our DLLs don't have the company name or version


guruz:
From Process explorer:

![screen shot 2018-05-29 at 12 03 34](https://user-images.githubusercontent.com/959327/40652482-a05abd3a-6338-11e8-98b3-b61c0ce6eb42.png)
![screen shot 2018-05-29 at 12 02 09](https://user-images.githubusercontent.com/959327/40652483-a0757580-6338-11e8-8379-8811cf62e396.png)

guruz:
Found in https://github.com/owncloud/client/issues/6553

dschmidt:
> I think @ogoffart or @ckamm should be assigned, this is about the DLL, not about MSI/NSIS.

In this case it's not related indeed. It could be though: The dlls for the NSIS are shipped in the `client-binaries` repo and built from MSVC solution files, the dlls for MSI are built as part of the client through cmake. The `.rc` would need to be added to solution files and `CMakeLists.txt` as long as we still support NSIS and MSI.

guruz:
Just noticed our normal DLLs don't have this either, the .exe has it. Something we want to fix? Hm.

![screen shot 2018-06-27 at 13 51 20](https://user-images.githubusercontent.com/959327/41972194-632fd338-7a11-11e8-99f1-a26b14cc730b.png)

guruz:
It works for the Explorer DLLs:
![screen shot 2018-06-27 at 14 00 30](https://user-images.githubusercontent.com/959327/41972554-7eb9024a-7a12-11e8-8038-343fe90fb7b0.png)

ogoffart:
I guess this would be the way to do it:
https://stackoverflow.com/questions/6693100/how-to-generate-windows-dll-versioning-information-with-cmake

Don't know if it is worth it.

dschmidt:
@TheOneRing something we want or rather not?

TheOneRing:
I don't really care.
@michaelstingl ?

michaelstingl:
Yeah, I think this would look better 👍

TheOneRing:
OK: let's do a cmake macro we can upstream to ecm

PR #6576, opened 2018-06-11
Windows shellext: Add rc for OCUtil 

ckamm:
For #6554

@ogoffart I haven't yet compiled the windows shell extension myself and just guessed at the values here. Could you see whether this works as intended?


dschmidt:
Yeah good catch. Maybe this is a good opportunity to update the year and to switch to ownCloud GmbH in all the rc files manually? Or maybe we can even use values defined in CMake somewhere already?

ckamm:
@dschmidt @ogoffart I've updated copyrights and company name, please double-check!

PR #6578, opened 2018-06-11
Add basic msi support to auto-updater

dschmidt:


guruz:
(Remember to backport this to 2.4.2)

guruz:
@ogoffart @ckamm FYI @dschmidt said that he's getting  https://stackoverflow.com/questions/45675936/executing-a-msi-file-from-a-qt-application .. also when using msiexec.exe it doesnt work. Hm.

@dschmidt To be honest I think your patch might be a bit overcomplicated. MSI is only for Windows and I thought it would be easier to just add "msiSupported=true" to the URL there. And when it's time to run the thing, just check the file extension is .msi and then use msiexec, otherwise use the old Windows code path.
No need for updateType.

dschmidt:
I messed a bit with paths and apparently it's working with msiexec, so I've got that going for me which is nice (tm). There is one last issue to resolve though: we need to restart the application after installation which is something that .msi does not support natively when run without a gui.

ckamm:
Did you mean to target 2.5 with this patch, or is 2.6 (master) intentional?

dschmidt:
@ckamm Changed, thanks!

@guruz Simplified. :beers:

src/gui/updater/ocupdater.cpp
```diff
@@ -303,7 +323,7 @@ void NSISUpdater::versionInfoArrived(const UpdateInfo &info)
             showDialog(info);
         }
         if (!url.isEmpty()) {
-            _targetFile = cfg.configPath() + url.mid(url.lastIndexOf('/'));
+            _targetFile = cfg.configPath() + url.mid(url.lastIndexOf('/')+1);
```
ogoffart: So this was broken all along...

src/gui/updater/ocupdater.cpp
```diff
@@ -192,8 +192,28 @@ void OCUpdater::slotStartInstaller()
...
+
+            return QDir::toNativeSeparators(path);
+        };
+
+        QString command = QString("&{msiexec /norestart /passive /i '%1'| Out-Null ; &'%2'}")
+            .arg(preparePathForPowershell(updateFile))
+            .arg(preparePathForPowershell(QCoreApplication::applicationFilePath()));
+
+        QProcess::startDetached("powershell.exe", QStringList{"-Command", command});
+    }
```
ogoffart: and else?

src/gui/updater/ocupdater.cpp
dschmidt: Yup, but no biggie. Windows usually handles doubled / just fine. Just msiexec is *very* picky about filenames.

src/gui/updater/ocupdater.cpp
dschmidt: I could just open the file with `openURL` but not sure if we really want that. 
Or I could show a QMessageBox ... but actually this should never happen, because there's no really reasonable way to handle *anything*

dschmidt:
Well, it should still start the client afterwards and let it do its "did the update fail detection"-magic. 
c.f. https://github.com/owncloud/client/pull/6578/files#diff-92391bca507954697996d2ddee514e25R211

guruz:
Please put https://github.com/owncloud/client/pull/6613 on top

dschmidt:
done

guruz:
@dschmidt Stash and merge if you're confident in it?

Issue #6581, opened 2018-06-13
Investigate white UI text


ckamm:
Also seen by @SamuAlfageme 

![image](https://user-images.githubusercontent.com/833306/41340205-1cef13a4-6ef7-11e8-89e7-119c940d711b.png)

Some normal ui text is white in this screenshot. Also something is up with the sync scheduling - for me it's stuck in "Preparing to sync" permanently.

ckamm:
This problem is odd. Using entirely new build directories each time I could initially reproduce it and gradually bisected through the recent commits. Then it was gone and is now also gone on the head of the 2.5 branch, I can not make it appear again.

I would guess at something timing / memory corruption related (why would the text go white?) but don't understand why it's now consistently gone for me.

@ogoffart The patches I thought might be related were 76bb76a59e6633b9a38c7a0e36d8e51efc7a4047 and 888f2aae2f31cb16ca8aa370ff44222d618bc146 - since not much else changed between 2.5.0-alpha1 and 2.5 - but I don't see anything wrong with them.

ckamm:
Another recompile without even changing code (just dependencies) and I reproducably get this on startup:
![image](https://user-images.githubusercontent.com/833306/41347127-fefb3db6-6f08-11e8-9653-ed53bedd5cb8.png)

ckamm:
The bug is in 13cb9ab1c5e645d1103254a8ef00273e87f67044 (download doesn't complete) and it was occasionally invisible when the file was blacklisted for a while.

ckamm:
@SamuAlfageme fixed in 2.5

guruz:
Did @SamuAlfageme manage to test this? I remember seeing something about this in an enterprise ticket

> Some normal ui text is white in this screenshot. 

Where was this fixed?

ckamm:
The odd white text wasn't fixed. It's possible that could still happen. What was fixed was the sync getting stuck - and with that no white text is visible anymore.

ogoffart:
White text was fixed in  36ca910d339fbb40eb9aeb225292af8b06258806

PR #6582, opened 2018-06-13
Ensure GETFileJob notices finishing 

ckamm:
It could happen that readyRead was emitted for incoming data while the
download was not yet finished. Then the network job could finish with
no more data arriving - so readyRead wasn't emitted again.

To fix this, the finished signal also gets connected to the readyRead
slot.

For #6581


ckamm:
@ogoffart I changed the tests to guarantee there's a test case that would fail if the connection was removed.

PR #6583, opened 2018-06-13
OAuth2: Try to refresh the token even if the credentials weren't ready.

ogoffart:
This can happen when the client is started and the internet connection
was not enabled. Then we would fetch the credentials, but we would
no do the refresh token step (because network is down).
So next time we try to connect, we would also not refresh the token
because the credentials are not marked as 'ready'

Reported in
https://github.com/owncloud/client/issues/6522#issuecomment-396845167


ogoffart:
@ckamm   no, master is not intentional.
I 'll move to 2.5

PR #6584, opened 2018-06-13
Add msi support to auto-updater  (2.4)

dschmidt:


src/gui/updater/ocupdater.cpp
```diff
@@ -195,9 +195,18 @@ void OCUpdater::slotStartInstaller()
...
+        // as they lack the user context. That is why we need to run the client
+        // manually here. We wrap the msiexec and client invocation in a powershell
+        // script because owncloud.exe will be shut down for installation.
+        // | Out-Null forces powershell to wait for msiexec to finish.
+        QString command = QString("&{msiexec /norestart /passive /i '%1' | Out-Null ; &'%2'}")
+            .arg(QDir::toNativeSeparators(updateFile))
+            .arg(QDir::toNativeSeparators(QCoreApplication::applicationFilePath()));
+
+        QProcess::startDetached("powershell.exe", QStringList{"-Command", command});
     }
```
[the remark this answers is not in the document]
SamuAlfageme: Hmmmm: http://doc.qt.io/qt-5/qstring.html#endsWith :

> If cs is `Qt::CaseSensitive` **(default)**, the search is case sensitive; otherwise the search is case insensitive.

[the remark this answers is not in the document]
dschmidt: Yeah, i'm not sure we ever want to ship `.EXE` or `.MSI` files or `.Exe` or `.mSi` but if it makes you happier I can certainly add `Qt::CaseInsensitive` ...

src/gui/updater/ocupdater.cpp
```diff
@@ -184,6 +184,81 @@ void OCUpdater::setDownloadState(DownloadState state)
...
 }
 
+namespace {
```
ckamm: minor: we usually just declare the functions `static` - here you do a mixture of both

src/gui/updater/ocupdater.cpp
```diff
@@ -184,6 +184,81 @@ void OCUpdater::setDownloadState(DownloadState state)
...
+            *pid = pinfo.dwProcessId;
+    // } else if (GetLastError() == errorElevationRequired) {
+    //     success = startDetachedUacPrompt(program, arguments, workingDir, pid);
+    // }
+
+    return success;
+}
+#else
+bool startDetached(const QString &, const QStringList &, const QString & = QString(), qint64 * = 0)
+{
```
ckamm: Would prefer
`return QProcess::startDetached(...)`
to avoid confusion about the method doing nothing on !Win32

src/gui/updater/ocupdater.cpp
dschmidt: A mixture of what?

src/gui/updater/ocupdater.cpp
dschmidt: Well it shouldn't be called at all on !Win32. But sure, can add it: easy enough and makes sense.

src/gui/updater/ocupdater.cpp
ckamm: Of an anonymous namespace and `static`. Both mean internal linkage for functions.

dschmidt: Ah, yes. I usually do the anonymous namespace and copied the static from the qt source.

guruz:
@SamuAlfageme let's try to test this together..

PR #6589, opened 2018-06-18
Wizard errors in a scroll area 

ckamm:
Another attempt at #6546

How about this?

### DIFF
--- a/src/gui/wizard/owncloudadvancedsetuppage.cpp
+++ b/src/gui/wizard/owncloudadvancedsetuppage.cpp
@@ -51,7 +51,7 @@ OwncloudAdvancedSetupPage::OwncloudAdvancedSetupPage()
 
     registerField(QLatin1String("OCSyncFromScratch"), _ui.cbSyncFromScratch);
 
-    _ui.resultLayout->addWidget(_progressIndi);
+    _ui.errorScrollContents->layout()->addWidget(_progressIndi);
     stopSpinner();
     setupCustomization();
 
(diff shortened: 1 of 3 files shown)


guruz:
(Code looks ok, havent tested it)

ogoffart:
I ran the code and it does not look so pretty when there is an error.
Maybe not related, but the spinner continues to spin, and that's what is causing the scrollbar to appear.

I'm wondering if it makes sense at all to have error there.
Network error should be shown in the UI anyway if they occurs. No need to check for network connectivity here.

Error while creating the local folder should stay there, but maybe they can be put right bellow the button.

ckamm:
I removed the leftover debug for starting a spinner when one of the radio buttons was clicked. I didn't touch the spinner show/hide much, so I'm surprised you see that. 

Agree that this is pretty painful and removing the error from this screen would simplify things.

guruz:
Show the error in a non-modal QMessageBox?
Something similar in https://github.com/owncloud/client/blob/46077f0de238298a8688f3ba8c4ab0c7fb3daf54/src/gui/owncloudsetupwizard.cpp#L286 ?

guruz:
tested by @jnweiger https://github.com/owncloud/client/issues/6546 .. merging

guruz:
CI failure is unrelated, CC @tboerger 
```java.io.IOException: Failed to rm container '3726526c32c7985378b69a8e57caf55743ef48001955f70382b8f3c96ecdc65d'.```

tboerger:
I think we can drop the jenkins pipeline, than we won't get these errors anymore.

guruz:
If it runs the same tests: sure. Maybe @DeepDiver1975 needs to remove it from the repo here?

https://drone.owncloud.com/owncloud/client/392/5
https://jenkins.owncloud.org/job/owncloud-client/job/client/job/PR-6589/5/consoleFull
Only difference seems to be logging level.

tboerger:
Yeah, Drone is not running in with full logging because it will cut off the output because it got some limitiation related to that. But beside this difference Drone seems to be doing a good job and we are another step forward to get rid of Jenkins.

Issue #6574, opened 2018-06-11
OAuth account wizard breaks when back button is used


ckamm:
Steps:
* Add a new oauth account
* In the advanced setup page hit "Back"
* Click "Re-open browser" and "Authorize" in the browser
* The browser will show "Login Error, Error returned from the server: invalid_request"

Note that the back button starts out as disabled but can be enabled by changing the folder sync type.

@ogoffart @SamuAlfageme

SamuAlfageme:
> Note that the back button starts out as disabled but can be enabled by changing the folder sync type.

Great catch @ckamm - disabling some navigation options was part of https://github.com/owncloud/client/issues/5813 but since the wizard rework happened this might have emerged

ckamm:
Huh. I cannot reproduce it anymore. :/

ogoffart:
I can reproduce!   It gets re-enabled when there is an error at the last step.

ogoffart:
When there is an error,  OwncloudAdvancedSetupPage::updateStatus and others call completeChanged() , which is connected to QWizardPrivate::_q_updateButtonStates which will re-enable the back button

We re-open the browser, but the account's credentials already have a oauth token set. So the call to the API to get a new token fails because we use the previous token instead of using the client's secret_id.

guruz:
2.5.0 beta1 is out :-)
https://central.owncloud.org/t/desktop-sync-client-2-5-0-beta1-released/14667
Everyone, please comment here if we can close this issue. Thank you.

jnweiger:
### linux mint tara

testpilotcloud version 2.5.0daily20180801 (build 9974)


    Add a new oauth account
    In the advanced setup page hit "Back"
    Click "Re-open browser" and "Authorize" in the browser
    The browser will show "Login Error, Error returned from the server: invalid_request"

The client window disappears when the browser is shown. Left-Click on the tray icon brings it back.

The first backbutton is always enabled
clicking there, visit the url again and press next, opens another browser window,

OK: login works.

Choose what to sync. Select one folder, visit the web interface, and rename that folder
change the sync folder to some folder that has contents, ignore the warnings, press cancel.
(some other error may be needed too, to enable the second back button. It worked only once for me)

Switch to 'Manually create folder sync connections" -> the second back button gets enabled.
 -> click re-open browser and auhtorize,

OK: connect successful.

PR #6591, opened 2018-06-18
Wizard + OAuth: Make opening a new browser after clicking back works again

ogoffart:
Issue #6574

When there is an error in the advanced page, OwncloudAdvancedSetupPage::updateStatus
(and others) call completeChanged(), which is connected to
QWizardPrivate::_q_updateButtonStates which will re-enable the back button from the
last page.

When the user click "back" and re-open the browser, the account's credentials
already have a oauth token set. So the call to the API to get a new token fails
because we use the previous token instead of using the client's secret_id.
Fix this with the HttpCredentials::DontAddCredentialsAttribute.

Now, this is still not working because the session cookies are confusing the
server.  So we'll clear the cookies when re-opening the browser

### DIFF
--- a/src/gui/creds/oauth.cpp
+++ b/src/gui/creds/oauth.cpp
@@ -22,6 +22,7 @@
 #include <QJsonDocument>
 #include "theme.h"
 #include "networkjobs.h"
+#include "creds/httpcredentials.h"
 
 namespace OCC {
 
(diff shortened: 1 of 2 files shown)


guruz:
@ckamm @ogoffart Does this need to be backported to 2.4.2 too?

ogoffart:
> Does this need to be backported to 2.4.2 too?

No, this is a very corner case issue. Not worth backporting IMHO.

src/gui/creds/oauth.cpp
```diff
@@ -85,6 +86,8 @@ void OAuth::start()
                 QString basicAuth = QString("%1:%2").arg(
                     Theme::instance()->oauthClientId(), Theme::instance()->oauthClientSecret());
                 req.setRawHeader("Authorization", "Basic " + basicAuth.toUtf8().toBase64());
+                 // We just added the Authorization header, don't let HttpCredentialsAccessManager temper with it
```
ckamm: minor: tamper

PR #6592, opened 2018-06-18
SettingsDialog: Show the page for the newly created account

ogoffart:
(I don't know if mac needs also a similar change)


guruz:
> (I don't know if mac needs also a similar change)

pushed

ogoffart:
> pushed

Great!

PR #6593, opened 2018-06-19
Updater: Fix beta channel for translated clients.

ogoffart:
The names in the combobox are being translated, so we need to rely
on the index rather than the text.

### DIFF
--- a/src/gui/generalsettings.cpp
+++ b/src/gui/generalsettings.cpp
@@ -157,12 +157,13 @@ void GeneralSettings::slotUpdateInfo()
 
     // Channel selection
     _ui->updateChannel->setCurrentIndex(ConfigFile().updateChannel() == "beta" ? 1 : 0);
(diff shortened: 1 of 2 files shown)


src/gui/generalsettings.cpp
```diff
@@ -157,12 +157,13 @@ void GeneralSettings::slotUpdateInfo()
 
     // Channel selection
     _ui->updateChannel->setCurrentIndex(ConfigFile().updateChannel() == "beta" ? 1 : 0);
-    connect(_ui->updateChannel, &QComboBox::currentTextChanged,
+    connect(_ui->updateChannel, static_cast<void (QComboBox::*)(int)>(&QComboBox::currentIndexChanged),
```
dschmidt: C++ Syntax ... yay!

ckamm:
Thanks for the fix

ogoffart:
(was cherry-picked in 2.4)

Issue #6075, opened 2017-10-04
Setting window is too tall and rannot be made smaller (because of the general settings tab)


mmattel:
### Expected behaviour
Open the client, resize horizontally and vertically should be possibe as it was in W7

### Actual behaviour
Open the client, resize horizontally = works, resize vertically = will not do
It "feels" like this is (b)locked, moving the window works

### Client configuration
Client version: 2.3.3

Operating system: W10x64

OS language: German

SamuAlfageme:
@mmattel hm, I can resize vertically, but only to make the window taller. Might have something to do with the `sizepolicy` set for https://github.com/owncloud/client/issues/5760 -> https://github.com/owncloud/client/commit/87a7642381858c298a3c3dbc4d3480701bf501e5 

Will take a look. Thanks for the tip!

ogoffart:
The problem is that the general settings tab is too big.
We should re-design it, or split it, or make the about text shorter. (There is also some space that is wasted because a bug in which way Qt computes the minimum size when word wrap is activated for the about label.)

ogoffart:
Note that we are talking about the minimum width of the settings dialog.  This is the case on every plaftorm and is not specific to Windows 10  (and Windows 7 had the same behavior)

@mmattel  can you confirm that you can indeed make the dialog veritcally bigger ?

mmattel:
I guess you mean horizontally...
To be precise:
Horizontally: Left - Right: slimmer: yes
Vertically: Bottom - Top: shorter: no

ogoffart:
@mmattel  But:  Vertically: Bottom - Top: bigger:  yes or no?

Anyway, I adjusted the title assuming yes.

mmattel:
Vertically (y-axis) no resize possible

SamuAlfageme:
@mmattel I'm testing with 2.4alpha (you also can give it a spin without conflicting with your current installation downloading the testpilotcloud version: https://download.owncloud.com/desktop/testing/testpilotcloud-2.4.0.8458alpha1-setup.exe) and I'm unable to reproduce: 

<p align="center">
  <img src="https://user-images.githubusercontent.com/2644445/31715639-f9d4aeae-b404-11e7-9a8a-e65bce991562.gif"/>
</p>

mmattel:
Just tested.

I can make the window narrow or wider (x-axis is ok)
But not shorter or higher (y-axis not)

My NB has W10x64 Enterprise edition, 8GB Ram

[...]

Issue #6454, opened 2018-04-13
Toolbar: Flatten the three activity tabs into the main tabs


guruz:
What's your opinion on this? I'm annoyed by the 3 sub-tabs in "Activity" and this could be a solution.
Maybe we'd need another tab for "Notifications" to split it away from the server activity view.
The new "About" tab is to fix https://github.com/owncloud/client/issues/6075 however is not strictly needed since we could also put that into a new section in the tray menu.

<img width="801" alt="screen shot 2018-04-13 at 17 41 32" src="https://user-images.githubusercontent.com/959327/38744541-55e7da86-3f42-11e8-914b-db55428cb394.png">

(I'm aware of the downsides when having a lot of accounts and a non-wide screen. But who has that?)

<img width="425" alt="screen shot 2018-04-13 at 17 46 07" src="https://user-images.githubusercontent.com/959327/38744616-9b991360-3f42-11e8-897e-b072598e796e.png">

michaelstingl:
I have no strong opinion, but I'm open for it. Let's try…

To give users a better overview. Add indicators for every account?

![more-status](https://user-images.githubusercontent.com/214010/38869063-b2987d1c-4249-11e8-98a6-90d982ca240c.png)

michaelstingl:
Add a number/count to display the amount of items in the Not Synced tab? (like iOS)

![](http://cdn.osxdaily.com/wp-content/uploads/2014/04/unread-mail-number-iphone.jpg)

dragotin:
Maybe a few tabs like General, Network and Account can go behind a drop down, like in Firefox?
<img width="597" alt="menu" src="https://user-images.githubusercontent.com/1070214/51740153-8a0e8700-2093-11e9-99d5-422ee6187454.png">
(see artistic red arrow, I love gimp :-/ )

guruz:
@hurradieweltgehtunter @TheOneRing FYI, what's your opinion?

michaelstingl:
@wuenschedesign ^

hurradieweltgehtunter:
Don't show multiple accounts in the main menu and don't mix menu contexts.
I'd say let's use the main menu for main tabs. 
Show multiple accounts as sub tabs in the main "Account" tab. 
Same goes for logs. Main tab is "Activity"(or "Log") and in a secondary tab level split it into "not synced", "protocol", ...

hurradieweltgehtunter:
quick mockups for better understanding

[...]

PR #6596, opened 2018-06-19
Introduce an about dialog

guruz:
Can someone check if this looks OK on Windows and Linux too?
screen shot 2018-06-19 at 17 12 43

Menu entry:

screen shot 2018-06-19 at 15 42 43

Dialog now not so high:
screen shot 2018-06-19 at 15 42 34

Every proper application has that!

For decreasing window height in #6075

### DIFF
--- a/src/gui/generalsettings.cpp
+++ b/src/gui/generalsettings.cpp
@@ -54,17 +54,6 @@ GeneralSettings::GeneralSettings(QWidget *parent)
     _ui->autostartCheckBox->setChecked(Utility::hasLaunchOnStartup(Theme::instance()->appName()));
     connect(_ui->autostartCheckBox, &QAbstractButton::toggled, this, &GeneralSettings::slotToggleLaunchOnStartup);
 
-    // setup about section
-    QString about = Theme::instance()->about();
-    if (about.isEmpty()) {
-        _ui->aboutGroupBox->hide();
-    } else {
-        _ui->aboutLabel->setTextInteractionFlags(Qt::TextSelectableByMouse | Qt::TextBrowserInteraction);
-        _ui->aboutLabel->setText("<qt><style> p{padding-bottom:0; margin-bottom:0;padding-top:2; margin-top: 2; font-size:smaller;};</style>"+about+"</qt>");
-        _ui->aboutLabel->setWordWrap(true);
-        _ui->aboutLabel->setOpenExternalLinks(true);
-    }
-
     loadMiscSettings();
     slotUpdateInfo();
 
--- a/src/gui/owncloudgui.cpp
+++ b/src/gui/owncloudgui.cpp
@@ -38,6 +38,8 @@
 #include <QDesktopServices>
 #include <QDir>
 #include <QMessageBox>
+#include <QDialog>
+#include <QHBoxLayout>
 
 #if defined(Q_OS_X11)
 #include <QX11Info>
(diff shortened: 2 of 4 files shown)


ogoffart:
Personally, I would have put the about screen as a new tab of the settings.
And I would put the "update" section in this about box, since it is related to the version.

I'll wait and hear what the design people say :-)

src/gui/generalsettings.cpp
```diff
@@ -54,17 +54,6 @@ GeneralSettings::GeneralSettings(QWidget *parent)
     _ui->autostartCheckBox->setChecked(Utility::hasLaunchOnStartup(Theme::instance()->appName()));
     connect(_ui->autostartCheckBox, &QAbstractButton::toggled, this, &GeneralSettings::slotToggleLaunchOnStartup);
 
-    // setup about section
-    QString about = Theme::instance()->about();
-    if (about.isEmpty()) {
```
ogoffart: The branding can specify an empty about, in which case we should also not show the entry in the system tray.

src/gui/owncloudgui.cpp
```diff
@@ -1131,5 +1137,24 @@ void ownCloudGui::slotRemoveDestroyedShareDialogs()
...
+#ifdef Q_OS_MAC
+    // From Qt doc: "On macOS, the window title is ignored (as required by the macOS Guidelines)."
+    msgBox->setText(title);
+#else
+    msgBox->setWindowTitle(title);
+#endif
+    msgBox->setAttribute(Qt::WA_DeleteOnClose, true);
+    msgBox->setTextFormat(Qt::RichText);
+    msgBox->setTextInteractionFlags(Qt::TextBrowserInteraction);
+    msgBox->setInformativeText("<qt><style> p{padding-bottom:0; margin-bottom:0;padding-top:2; margin-top: 2; font-size:smaller;};</style>"+about+"</qt>");
```
ogoffart: is the style  "cargo cult"?

I guess we don't need this if it is in its own dialog.

src/gui/generalsettings.cpp
guruz: Fixed

guruz:
Looks better now after removing the cargo cult:
screen shot 2018-06-20 at 11 46 22

guruz:
> Personally, I would have put the about screen as a new tab of the settings.

There will be too many tabs then -> https://github.com/owncloud/client/issues/6454
(And "About" is much less important than those tabs)

> And I would put the "update" section in this about box, since it is related to the version.

Can be done in a followup patch

michaelstingl:
Could you post screenshots of Win and Linux?

ckamm:
@michaelstingl Linux:
image

guruz:
E.g.
screen shot 2018-06-27 at 14 19 48

Will need to test this on Linux/Windows though..

guruz:
screen shot 2018-06-29 at 16 10 11

ckamm:
Updated linux screenshot:
image

guruz:
We can beautify this in a later round and optionally move the updater status/buttons there.

jnweiger:
The downsides of this solution are:

*  On some systems, e.g. Linux Gnome there is no tray icon. 
 This makes the about info (including version number!) now inaccessible. 
 Opening the settings dialog (which had the about info) was more robust. It is always reachable by  restarting the client, even if the trayicon is missing or defunct.

* Users have no hint why the the about info disappeared from the general tab, and may not easily discover where it is now..

I'd suggest to also have a small link on the bottom of the general tab that would open the about dialog.

guruz:
Hm. In theory you are right, but in practice: Maybe it's better to suggest people to install this package that @dragotin mentioned shows an alternative tray implementation?

Issue #6585, opened 2018-06-14
Disable update channel dropdown on branded builds 


SamuAlfageme:
... even if there's an updates URL. 

So far, we don't have any documentation on how to server an onpremise beta channel for branded clients and therefore the stable/beta dropdown has no use other than confusing users.

PR #6597, opened 2018-06-19
GeneralSettings: Hide update channel for other themes

guruz:
For #6585


guruz:
(ideally it should be a getter in Theme but we can do that if whitelabel customers really want it)

guruz:
Needs backport to 2.4.2 so a quick review would be appreciated

guruz:
CI failure unrelated
```

java.io.IOException: Failed to rm container '63f02690b5348f4a732c0fa8a432545cd15128227f9b4748178426b99bc87959'.
	at org.jenkinsci.plugins.docker.workflow.client.DockerClient.rm(DockerClient.java:186)
```

(CC @tboerger )

Issue #3933, opened 2015-10-08
[2.x Windows] After Hibernating - Always checks for client software update and never finishes


MegaV0lt:
Win Client does always check for update but never finishing the task. See Screenshot:
https://db.tt/1M7dwUzY

dragotin:
that looks like as if the request timed out during the hibernation and that does not get reflected to the UI.

MegaV0lt:
Same effect with current 2.02 version after some days... this does not happen every time hibernating...

MegaV0lt:
still with 5632
![2015-11-18 08_27_41-owncloud](https://cloud.githubusercontent.com/assets/2804301/11235145/15cd1bfe-8dcf-11e5-8278-1b02153f4c28.png)

MegaV0lt:
Problem still exist in 2.1
![2015-12-07 15_00_33-owncloud](https://cloud.githubusercontent.com/assets/2804301/11628659/65d73764-9cf3-11e5-86a5-d309d1e5469d.png)

MegaV0lt:
Still the same with 5785...
![2016-01-21 09_56_21-owncloud](https://cloud.githubusercontent.com/assets/2804301/12475833/db488bba-c025-11e5-85b4-e744118857d0.png)

MegaV0lt:
Problem still with #5833

hodyroff:
please reproduce with 2.3

MegaV0lt:
In the meantime i am at nextcloud client, but still the same. Today after the first wakeup after hibernation i have the same issue:
![Screenshot](https://i.imgur.com/qcC3vG4.png)

git from 6 mar 2017

guruz:
From looking at the code, there should be a 30 sec timeout when checking for updates..
```OCUpdater::checkForUpdate()```

MegaV0lt:
Was never working in Windows... Meanwhile i switched to Linux Mint. I will watch if there is the same effect.

guruz:
Will be fixed in 2.5.0 @MegaV0lt

PR #6604, opened 2018-06-21
OCUpdater: Set correct state on network error

guruz:
For #3933


guruz:
@dschmidt Why does the Abbveyor thing fail?

dschmidt:
It times out ... possibly because of a wrong cache number ... something. Need to look next week.

PR #6605, opened 2018-06-21
UpdateInfo: Remove unused code

guruz:


guruz:
FYI @dragotin

ogoffart:
This code was auto-generated.
But since the original xml file is not there, and that there was many changes to it, i guess it is fine to remove this code.

PR #6611, opened 2018-06-23
Fix app icon macro

dschmidt:
I don't know how this ever worked (and still does) on Windows - but it's broken when cross-compiling (or in a different CMake version). This is correct and should work everywhere and in any reasonable CMake version.


dschmidt:
It was broken because I set a string first and then appended to it as if it was a list. That led to messed up quotation marks.

Issue #6609, opened 2018-06-22
[Windows] 2.4.2 nightly EXE-> 2.5.0 alpha1 MSI update doesn't work


guruz:
Installed 2.4.2 daily.

Run with 
```
C:\Program Files (x86)\ownCloud>set OCC_UPDATE_URL=http://[...]/updater.xml
C:\Program Files (x86)\ownCloud>owncloud.exe --logfile=c:\github\oclog_msi.txt --logdebug
```
Click the 'install and relaunch'. in settings.

```
06-22 14:26:40:962 [ info gui.updater ]:	Running updater "C:/Users/marku/AppData/Local/ownCloud/ownCloud-2.5.0-alpha1.MSVC.238.msi"
```

MSI seems to be getting installed (I see a progress dialog quickly)

But then I get a broken ownCloud directory:

<img width="471" alt="screen shot 2018-06-22 at 14 34 42" src="https://user-images.githubusercontent.com/959327/41776818-77155224-7629-11e8-8643-8f5c50fba1e4.png">

-> Installation busted.

is it because I have the directory open in cmd.exe?

guruz:
(A reboot doesn't change anything, the leftover files are there still and there is no new one)

guruz:
Not sure if relevant, when cleaning build dir in explorer
![screen shot 2018-06-22 at 14 46 00](https://user-images.githubusercontent.com/959327/41777276-0d0ba796-762b-11e8-8b38-c4ce1b118518.png)
:

guruz:
> is it because I have the directory open in cmd.exe?

Not related to that (but I still had it open in explorer)

guruz:
> Not related to that (but I still had it open in explorer)

Not related to that either.
@dschmidt I can't update at all! :-/

guruz:
When I use the "Restart & Update" button I get more files but still a broken installation.
The client seems to exit properly before installation (maybe).

![screen shot 2018-06-25 at 12 38 18](https://user-images.githubusercontent.com/959327/41845730-c85b9056-7874-11e8-915e-c70c3de7314a.png)

guruz:
(The owncloud.exe there is the 2.5.0 one so it has succeeded partly)

guruz:
qt5keychain, Qt5Sensors, Qt5PrintSupport.dll are also missing.

This sounds similar-but-not-same as the leftover files. heh

guruz:
[...]

PR #6613, opened 2018-06-25
MSI: Always with logfile #6609

guruz:
* [ ] needs cherry-pick into 2.5 after

### DIFF
--- a/src/gui/updater/ocupdater.cpp
+++ b/src/gui/updater/ocupdater.cpp
@@ -283,9 +283,11 @@ void OCUpdater::slotStartInstaller()
             return QDir::toNativeSeparators(path);
         };
 
-        QString command = QString("&{msiexec /norestart /passive /i '%1'| Out-Null ; &'%2'}")
-            .arg(preparePathForPowershell(updateFile))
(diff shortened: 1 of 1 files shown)


src/gui/updater/ocupdater.cpp
```diff
@@ -283,9 +283,11 @@ void OCUpdater::slotStartInstaller()
             return QDir::toNativeSeparators(path);
         };
 
-        QString command = QString("&{msiexec /norestart /passive /i '%1'| Out-Null ; &'%2'}")
-            .arg(preparePathForPowershell(updateFile))
-            .arg(preparePathForPowershell(QCoreApplication::applicationFilePath()));
+        auto msiLogFile = cfg.configPath() + "msi.log";
```
ckamm: Is `configPath()` here a good choice for windows?

dschmidt: I'd say so: it's next to the .msi file itself

guruz:
I can't push the patch into the PR https://github.com/owncloud/client/pull/6578 because it's in @dschmidt 's repo instead of here in owncloud/client :-/ @dschmidt can you do it?

dschmidt:
Yes, sure. Will do it later today

PR #6614, opened 2018-06-25
Upload: asynchronious operations

ogoffart:
Implements https://github.com/owncloud/core/pull/31851


ogoffart:
(WIP)

guruz:
@jnweiger btw next time better build with branch name in file name:)
not that it matters much in this case

ogoffart:
@PVince81 I see something was merge in the server stable10.  Does that mean we should now go forward with this client side?

@ckamm : do you think this is small enough to go into 2.5.1? or should it be in master?

src/libsync/owncloudpropagator.cpp
```diff
@@ -1029,6 +1029,7 @@ void CleanupPollsJob::slotPollFinished()
             deleteLater();
             return;
         }
+        _journal->setUploadInfo(job->_item->_file, SyncJournalDb::UploadInfo());
```
ckamm: Can PollJobs have a "there was an error, don't retry" result in any way? I'm just wondering that no other branch wipes the uploadinfo.

PollJob seems to wipe the pollinfo on non-fatal non-503

test/testasyncop.cpp
```diff
@@ -0,0 +1,223 @@
...
+                };
+                return new FakeAsyncReply("/async-poll/" + file.toUtf8(), op, request, &fakeFolder.syncEngine());
+            } else if (op == QNetworkAccessManager::GetOperation) {
+                nGET++;
+            }
+            return nullptr;
+        });
+
+
+        auto successCallbak = [](TestCase *tc, const QNetworkRequest &request) {
```
ckamm: minor: Callback; decide whether CallBack or Callback

src/libsync/propagateuploadng.cpp
```diff
@@ -455,6 +455,18 @@ void PropagateUploadFileNG::slotMoveJobFinished()
...
     }
+
+    if (_item->_httpErrorCode == 202) {
+        QString path = QString::fromUtf8(job->reply()->rawHeader("OC-JobStatus-Location"));
+        if (path.isEmpty()) {
+            done(SyncFileItem::NormalError, tr("Poll URL missing"));
+            return;
+        }
+        _finished = true;
+        startPollJob(path);
```
ckamm: How does the overall sync timeout work in the presence of polls?

src/libsync/owncloudpropagator.cpp
ckamm: Speaking of `PollInfo`: It's set on `startPollJob` but only removed in this specific error case. And then when a sync is started, PollJobs are rerun? That means polls run indefinitely and block further syncs as long as the server replies "init" or "started"?

I think the idea is that we can continue polling for completion in case the user aborts the sync? (I think you tested that `abort()` case in the unittest, excellent!) PollInfos need more documentation about their purpose.

PVince81:
note: there's a capability so if you deem suitable, the client could decide to enable the feature if the server does: https://github.com/owncloud/core/pull/32414/files#diff-920833889854d937ebeda804d3eb5b19R52

currently what's in 10.0.10 is in tech preview / alpha stage

ckamm:
My current understanding is that this feature is enabled at the server's discretion in the first place: It looks like the client announces its capability to the server with the header, and then it either replies with the poll url or does a synchonous action.

The concern about having it default to off for the patch release is about potential bugs in the client implementation. In addition there could be a way of switching this feature off in case bugs are discovered in the server implementation.

src/libsync/owncloudpropagator.cpp
ogoffart: In case of error, the uploadinfo has to stay so the next sync will deal with it by retrying the final move, or use the upload info to resolve possible conflict, should the file have been updated after all (using checksum)

ogoffart: Yes, a server never finishing that always reply "started" will block the sync forever.
Do you think this is a problem?

src/libsync/propagateuploadng.cpp
ogoffart: The timeouts are per jobs.  A particular PollJob will timeout after the normal timeout time if there is not traffic on that connection. In this case, it will be re-tried again until it succeed.

ogoffart:
@ckamm: I've updated the commit:  Added comments, and disabled the behavior by default unless setting the env variable `OWNCLOUD_LAZYOPS=1` 

@DeepDiver1975 or @PVince81 :  How long does the poll url stay valid, after it once succeed.  Imagine this scenario:
 1. The client send a MOVE, the server reply with code 202 and a poll URL. 
 2. The server do the operation, the operation is finished
 3. The client request the pull url. 
 4. The server send the "finished", but the client crashes before recieving it!  (or the connection drops, or the user puts its laptop to sleep)
 5. The client restarts and poll the URL, will that work?    What if it is a month later?


Also note that the client will set the OC-LazyOps header for **all** uploads (non-chunking PUT or chunking MOVE).
We expect the server to be smart about it and not go into asynchronous operation if it know the operation will be fast.  Going into asynchronous mode for small PUT would kill performance!

src/libsync/owncloudpropagator.cpp
ckamm: UploadInfo not being wiped on failure: So if the server wants to communicate "something failed permanently, stop polling", the poll job would fail, we'd try to resend the final MOVE during the next sync and there it'd fail. Alright, that works.

Infinite polling if server replies that way: Should be ok, would be a significant server bug if that happens and working around it is complicated.

src/libsync/propagateuploadng.cpp
ckamm: I meant to confirm that if the server needs 30 minutes to finish the upload then the sync will only finish once that last poll is done. (as it should be for 2.5.1)

ckamm:
minor: commit message says "assynchronious", should be "asynchronous"

guruz:
I'm actually fine with having this in 2.5.1 if we enable this feature only by capability. (We kind-of tested this in the wild with a previous customer when it was called asynchronous PUT with poll jobs).
It could be forcible disabled by clients for testing (even if capability is on) by setting `OWNCLOUD_LAZYOPS=0`

In the server it's [disabled by default in config.sample.php](https://github.com/owncloud/core/pull/32414/files#diff-2f769bc01ffefdf14eb04404218e7582R1467) so brave server deployers need to enable it anyway in server?

if @ckamm @ogoffart agree please change milestone.

@PVince81 @ogoffart Can we still get answers to the questions above. https://github.com/owncloud/client/pull/6614#issuecomment-430921430

PVince81:
I think the server is not smart and always go to lazy mode when requested.

Not sure how long the poll URL is available, @DeepDiver1975 ?

ogoffart:
Rebased and fixed for 2.6

Issue #6610, opened 2018-06-23
[Windows] Client does not sync .mwb MySQL Workbench files since version 2.4.0


sfadschm:
### Expected behaviour
.mwb files should be recognized and synced normally (works fine up to version 2.3.4, even after a rollback from 2.4.x)

### Actual behaviour
.mwb files are not recognized at all. When it gets created, it gets labeled with the blue “syncing” symbol, but this disappears within 1-2s. Looking into the client, no activity is reported, neither a syncing process nor an error, nor an ignored file. I anyway checked the ignored files list and none of the entries suits the file. Interestingly, if I change the files extension to “.txt”, it will still not be recognized or handled.
After a rollback to 2.3.4, the file will immediately be recognized and synced. Now with the file being inside the synced folder, after an automatic client update, it will even get deleted from the server (I now disabled automatic updates).

### Steps to reproduce
1.     Open MySQL Workbench.
2.     Create new .mwb model file.
3.     Place file in OwnCloud directory.

### Server configuration
Operating system: Debian
Web server: Apache
PHP version: 7.1
ownCloud version: latest version (stable channel)
Storage backend (external storage): shared host

### Client configuration
Client version: latest
Operating system: Windows 10
OS language: german

guruz:
Does the behaviour change if you close MySQL Workbench?

Which messages related to this file do you see in the log file / log window?
https://doc.owncloud.org/desktop/2.4/troubleshooting.html#log-files

(CC @ckamm maybe related to locking?)

sfadschm:
Closing Workbench unfortunately does not change the behaviour. 

I just did the following:
1: .mwb file is synced on both server and client (with v2.3.4) and inside the folders.
2: Update client to v2.4.1.
3: Open .mwb file with Workbench, save file and exit Workbench.

Result:
Synchronisation protocol shows: ".mwb file deleted", ".mwb.bak file uploaded". (the .mwb.bak file is a backup of the old .mwb file produced by Workbench). The .mwb file is completely removed from 

After that I duplicated the .mwb inside the folder. The sync protocoll does not show any action. The log windows shows the following for the two actions:

06-25 11:43:05:270 [ info gui.folderwatcher ]:	Detected changes in paths: QSet("C:/Xampp/htdocs/warehouse/_mysql/warehouse.mwb")
[...]

PR #6616, opened 2018-06-26
Update: Report on readdir() errors

ckamm:
I found that we're ignoring a class of potentially critical errors while debugging #6610

Might warrant picking into 2.4?


guruz:
```
     if( FindNextFile(handle->hFind, &(handle->ffd)) == 0 ) {
          // might be error, check!
          int dwError = GetLastError();
          if (dwError != ERROR_NO_MORE_FILES) {
              errno = EACCES; // no more files is fine. Otherwise EACCESS
          }
          return nullptr;
      }
```

Can you log out the error in Windows VIO too?

src/csync/csync_update.cpp
```diff
@@ -712,7 +713,30 @@ int csync_ftw(CSYNC *ctx, const char *uri, csync_walker_fn fn,
...
+
+        // Note: Windows vio converts any error into EACCES
+        qCWarning(lcUpdate, "readdir failed for file in %s - errno %d", uri, errno);
+
+        // If we get an error during readdir, maybe that's a single unreadable
+        // item that we can skip? It's unclear whether that can happen.
+        // We definitely don't want to get stuck in an infinite loop here,
+        // so stop if we get the same errno twice in a row.
+        if (errno == last_errno) {
+            qCWarning(lcUpdate, "Second failure, giving up on directory");
```
ogoffart: AFAIK,  the `goto error`  will cause the whole sync to abort, isn't it?

src/csync/csync_update.cpp
```diff
@@ -712,7 +713,30 @@ int csync_ftw(CSYNC *ctx, const char *uri, csync_walker_fn fn,
...
+            goto error;
+        }
+        last_errno = errno;
+        continue;
```
ogoffart: I'm not sure if we should continue. 
If it is a temporary error, for example, we might end up skipping files in the discovery, meaning that the client would believe they were removed, so it would delete them on the server.
Aborting the sync seems a safer option,  or at least somehow ignoring any removed item in this directory.

ckamm:
@guruz Logging: Definitely
@ogoffart Yes, the continue is incorrect, an error during file iteration should be fatal - or exclude the whole directory subtree from sync actions.

I think this is a real issue. If readdir fails we currently just don't see some files locally and will proceed to delete them on the server. That's exactly the symptoms of #6610 - but it was caused by something else in the end.

Issue #6612, opened 2018-06-25
[OAuth2] Authorization fails with a Connection refused after 5min delay


mirekys:
### Expected behaviour
Client's web-side OAuth2 authorization should either succeed or
fail with a meaningful error message to the user (that time period
for authorization has expired an he should try again faster).

### Actual behaviour
Connection refused to the following callback URL is shown in the browser after clicking Authorize:
```
http://localhost:<random_port>/?code=<oauth_code_token>
```

### Steps to reproduce
1. Start the client (will open the ownCloud login page in the browser asking for Client Authorization)
2. Wait for > 5min
3. Now try to authorize the client

### Server configuration
Operating system: RHEL7

Web server: Apache

Database: PostgreSQL

PHP version: 7.1

ownCloud version: 10.0.8

Storage backend (external storage): local

OAuth2 provider app enabled on the server

### Client configuration
Client version:  2.4.1

Operating system: probably all

OS language: en_US

Qt version used by client package (Linux only, see also Settings dialog): Qt 5.11.0

Client package (From ownCloud or distro) (Linux only): distro 

### Logs

1. Client logfile: 
```
>>> started the client
06-25 13:55:51:321 [ info sync.connectionvalidator ]:	No system proxy set by OS
06-25 13:55:51:504 [ info sync.networkjob.checkserver ]:	status.php returns:  QJsonDocument({"edition":"Community","installed":true,"maintenance":false,"needsDbUpgrade":false,"productname":"ownCloud","version":"10.0.8.5","versionstring":"10.0.8"})   QNetworkReply::NetworkError(NoError)  Reply:  QNetworkReplyHttpImpl(0x560cfc369ab0)
06-25 13:55:51:505 [ info sync.connectionvalidator ]:	** Application: ownCloud found:  QUrl("https://owncloud")  with version  "10.0.8" ( "10.0.8.5" )
06-25 13:55:51:505 [ info sync.connectionvalidator ]:	QUrl("https://owncloud") has server version "10.0.8.5"
06-25 13:55:51:512 [ info gui.account.state ]:	AccountState connection status change:  "Undefined" -> "Credentials not ready"
06-25 13:55:51:512 [ info gui.account.state ]:	Invalid credentials for "https://owncloud" asking user
06-25 13:55:51:513 [ info gui.account.state ]:	AccountState state change:  "Disconnected" -> "Asking Credentials"
[...]

PR #6617, opened 2018-06-26
OAuth: Remove the timeout

ogoffart:
There is no real reason to have a timeout. The connection can stay open
as long as we are not authenticated. The User can still re-open a browser
from the UI at any time.

Issue #6612


guruz:
@ogoffart Please set milestones on issues. Why is this master, shouldnt it be 2.5 ?

ckamm:
Agree about 2.5 though!

PR #6620, opened 2018-06-27
Update libcrashreporter-qt to support Qt 5.11

dschmidt:


dschmidt:
Oh lol ... 

It failed in my windows build and I think that was Qt 5.11.1 already ... kill qt5_use_modules with fire :fire:

Issue #6619, opened 2018-06-27
Setup our own Craft cache (mirror)


dschmidt:
Our AppVeyor builds start to fail whenever the cache name changes on the KDE Mirrors because we then try to build all deps ourselves in AppVeyor which then times out. Moreover it helps us being able to update docker images after the cache has been removed from KDE mirrors.
Current issue: KDE removed the cache for Qt 5.10 builds and we need to switch to 5.11 because I need changes in the docker image.

We do not need to build any packages for the cache for now, we should simply mirror the KDE mirror and make sure that old caches are not deleted.

http://ftp.acc.umu.se/mirror/kde.org/files/craft/master/
I use this mirror as it turned out to be more reliable/faster than 
https://files.kde.org/craft/master/

We need that on a public webserver (for AppVeyor to access).

Bonus points if our script/job sets up a "latest" link, so we can use that cache version in master.

We should then also add a todo item for creating release branches, that then pins the qt version to whatever we use for the release.
We can possibly use that information at some point to update docker images for that specific branch.

guruz:
So client 2.5.0 on Windows (only there) will have Qt 5.11.x?

dschmidt:
If there are strong reasons to go back to 5.10, I can possibly make that happen - but it would be quite some effort :-( 

So yes, I would prefer to go with Qt 5.11. Craft currently provides Qt 5.11.0. 
@TheOneRing can we get Qt 5.11.1?

dschmidt:
Talked to @TheOneRing on IRC. The cache is still named Qt 5.11.0 but already contains 5.11.1. 
So yes, we would have Qt 5.11.1 for 2.5.

guruz:
@dschmidt We definitely need the .1 because of https://github.com/owncloud/client/issues/6517#issuecomment-398021244

jnweiger:
done:  https://download.owncloud.org/mirror/craft/

```$ du -sh Qt_5.11.1/windows/msvc2017_*
1,4G Qt_5.11.1/windows/msvc2017_32
1,7G Qt_5.11.1/windows/msvc2017_64

dschmidt:
I'm currently testing it...

PR #6621, opened 2018-06-27
Use Qt 5.11 in Craft

dschmidt:
I will fix this later in a less fragile way finally ... for now I wanna check if AppVeyor still builds at all...

### DIFF
--- a/appveyor.ini
+++ b/appveyor.ini
@@ -9,9 +9,9 @@ Command=craft
 #Values need to be overwritten to create a chache
 UseCache = True
 CreateCache = False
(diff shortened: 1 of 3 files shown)


dschmidt:
It also uses our new craft cache, set up by @jnweiger (https://github.com/owncloud/client/issues/6619)

When the AppVeyor build succeeds, we can merge...

jnweiger:
Looks good to me. Not sure why the last test case segfaults. Probably not an issue with the tested code. @olivier could you have a quick look there?

PR #6623, opened 2018-06-27
Drop Jenkinsfile, replaced by DroneCI

tboerger:
So far DroneCI is showing stable results, that's the point in time where
we can get rid of Jenkins :bomb:


dschmidt:
AFAICT we don't need this PR, we merge 2.5 into master anyway.
So I'd say let's close this, merge the other PR for 2.5 and merge 2.5 into master soonish - which would be good as 2.5 contains a bunch of little fixes from me too.

guruz:
> we don't need this PR, we merge 2.5 into master anyway.

ack

PR #6624, opened 2018-06-27
[2.5] Drop Jenkinsfile, replaced by DroneCI

tboerger:
So far DroneCI is showing stable results, that's the point in time where
we can get rid of Jenkins :bomb:


ogoffart:
But now, of course, Jenkins is failling. as it cannot find the jenkinsfile.

Maybe removing the jenkinsfile is not the right way to disable jenkins. There is maybe some option we need to toggle in github or in jenkins.

patrickjahns:
@ogoffart 
Can do - but then its for all branches - just let me know if thats okay

guruz:
@patrickjahns Yes please remove, if drone runs the same tests.

tboerger:
If everything is fine with this pr we have to disable the client pipeline within jenkins and we have to verify that the webhook from the repo have been dropped.

guruz:
> All checks have passed
4 successful checks

PR #6626, opened 2018-06-29
Update the minimum supported version

ogoffart:
We want to warn if the server version is not supported and did not get
appropriate QA for this client version.

https://github.com/owncloud/enterprise/issues/2687


ho4ho:
Nextcloud check interpreted as "fighting":

> Argh they now really start fighting against each other - WTF?

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=847613

PR #6630, opened 2018-07-04
csync_update cleanup

ogoffart:
In preparation for future work on the discovery phase

### DIFF
--- a/src/csync/csync_update.cpp
+++ b/src/csync/csync_update.cpp
@@ -171,10 +171,6 @@ static int _csync_detect_update(CSYNC *ctx, std::unique_ptr<csync_file_stat_t> f
 
   if (excluded > CSYNC_NOT_EXCLUDED || fs->type == ItemTypeSoftLink) {
       fs->instruction = CSYNC_INSTRUCTION_IGNORE;
-      if (ctx->current_fs) {
-          ctx->current_fs->has_ignored_files = true;
-      }
-
       goto out;
   }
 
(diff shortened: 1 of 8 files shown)


ckamm:
(intentionally `master`, I assume)

src/csync/csync_update.cpp
```diff
@@ -762,45 +751,44 @@ int csync_ftw(CSYNC *ctx, const char *uri, unsigned int depth) {
...
     if (rc < 0) {
       if (CSYNC_STATUS_IS_OK(ctx->status_code)) {
           ctx->status_code = CSYNC_STATUS_UPDATE_ERROR;
       }
       goto error;
     }
+    if (rc != 0) {
+        ctx->remote.read_from_db = read_from_db;
+        continue;
+    }
```
ckamm: Why is that new here?

ckamm:
I very much like these cleanups! Could you also update `doc/dev/sync-algorithm.md` (maybe only the `csync_ftw` change)

src/csync/csync_update.cpp
ogoffart: If rc is > 0 , the 'fs' pointer will be dangling, so we need to continue and avoid the next operation that touches 'fs'.

The read_from_db thingy is there to keep the invariant that was there before. This was like that because in the past, it was the vio code that did the read from the db.  
Anyway, this goes away in the next commit.

ogoffart:
> Could you also update doc/dev/sync-algorithm.md (maybe only the csync_ftw change

Done

ogoffart:
csync is going away anyway

Issue #5867, opened 2017-06-28
Locally copying a file should cause a COPY not a PUT


ckamm:
This is a follow up of a discussion in #3422.

Currently when you make a copy of a file that is synchronized, the copy will be uploaded from scratch. Instead, the client should send a COPY command to the server. For large files this would save a lot of bandwidth and time.

The duplicate detection should be possible by checking size, hashes and possibly mtime.

This server issue is relevant: https://github.com/owncloud/core/issues/26584

guruz:
* [ ] make sure ```COPY``` honors the ```If-match``` header so we really only copy the file if it has that ETag

ogoffart:
* [ ] Ceate an index on the contentChecksum in the DB
* [ ] Compute the checksum in _csync_detect_update for any INSTRUCTION_NEW for new local files
* [ ] Create a new INSTRUCTION_COPY that can already be set in detect_update, confirmed in reconcile
* [ ] Make a PropagateRemoteCopy job.

Also for server copy detection:

* [ ] If the server provide a "sufficiently good" checksum in the metadata, we can also do copy detection to avoid multiple download

But this algorithm, that query the database, only work if the file would have been there from a previous file. It would need some tweaking if two identical files are added at the same time.  (When should we do the decision to copy? it has to be done in the right order)

For which file do we compute the checksum in the update phase? All NEW files?  All files whose mtime has changed (currently we do that for eml files only)? 

Finaly, which hash can we use for that? Is it ok to do that with MD5 or SHA-1?

AnrDaemon:
You can't ensure file content equality on hash alone.
Especially not on known-weak hash, such as md5.

ckamm:
@ogoffart What if the decision to COPY instead of PUT was done in the PropagateUpload job? We compute the content checksum of all files we upload in that job anyway.

ckamm:
About hashes: We currently use `csync_is_collision_safe_hash` to determine whether assuming content equality based on hash equality has acceptable risk.

guruz:
> You can't ensure file content equality on hash alone.

Is SHA1+filesize good enough?

ogoffart:
> Is SHA1+filesize good enough?

It's been 10 years since SHA1 is no longer considered good enough for chryptographic hashes. And a collision was even found earlier this year.

https://en.wikipedia.org/wiki/Hash_function_security_summary
http://valerieaurora.org/hash.html

[...]

PR #6634, opened 2018-07-05
Upgrade Checksum algorithms

ogoffart:
The first commit move the checksum code from filesystembase.cpp to checksum.cpp, where they belong. I don't want to add more wrapper around QCryptographicHash in filesystem, ot include QCryptographicHash from there.

The second commit add support for "SHA2" (SHA256) and "SHA3" (SHA3-256).
SHA-1 has already been deprecated for 10 years. We should at least support new algorithm so the server will also be able to support them.

SHA-1 is still the default contents checksum as this is what the server supports and we want to use the same.

To be bikesheded: the name of the string "SHA2" and "SHA3" or should it be "SHA256" and "SHA3-256"

@IljaN  What about server support?


IljaN:
Should be easy to add but nothing planned atm. Also not sure if the security implications of SHA1 also apply to it`s usage in non-cryptographic contexts(checksums) as we also still support MD5 and ADLER32 which are even more broken in this regard.

ogoffart:
The client is already using the checksum as content hash to detect if the file was changed in order to avoid uploading the file if the checksum is the same. In the future, with #5867, we would even do a COPY instead of re-uploading the files with the same contents.
Of course we only do that if the hash algorithm is safe enough. So we do not do it with Adler32. (Only with MD5 and SHA1)

We probably do not need to worry about a collision attack (or should we?) but if users upload  these collision files, we would have a bug right now.

guruz:
> To be bikesheded: the name of the string "SHA2" and "SHA3" or should it be "SHA256" and "SHA3-256"

Isn't the longest string the most accurate one? Especially if the Qt enums are called similarly.
In the German wikipedia it sounds like SHA2 is a generic term for 4(!) algorithmns. https://de.wikipedia.org/wiki/SHA-2

ogoffart:
Right, So i changed the string value to "SHA256" and "SHA3-256".
I could have used "SHA2-256" instead of "SHA256",  but the SHA256 is what is how it is normally called.

Also i could have chosen the 224bit variant, but the 256 variant is the most used.

guruz:
@IljaN @ogoffart Who creates an issue in 'core'?

PVince81:
@IljaN please create a core issue to schedule an upgrade.

@settermjd any thoughts on the above ?

IljaN:
Core issue: https://github.com/owncloud/core/issues/32071

Issue #6644, opened 2018-07-11
owncloud --help usage output has unwrapped lines


voroyam:
![image](https://user-images.githubusercontent.com/24212810/42584049-9a8417b4-8532-11e8-8639-2259a5880e71.png)

guruz:
Same on macOS
```
ownCloud version 2.5.0alpha (build 9853)
File synchronisation desktop utility.

Options:
  -h --help            : show this help screen.
  --logwindow          : open a window to show log output.
  --logfile <filename> : write log output to file <filename>.
  --logdir <name>      : write each sync log output in a new file
                         in folder <name>.
  --logexpire <hours>  : removes logs older than <hours> hours.
                         (to be used with --logdir)
  --logflush           : flush the log file after every write.
  --logdebug           : also output debug-level messages in the log (equivalent to setting the env var QT_LOGGING_RULES="qt.*=true;*.debug=true").
  --confdir <dirname>  : Use the given configuration folder.

For more information, see http://www.owncloud.org
```

guruz:
Fixed in https://github.com/owncloud/client/pull/6647

PR #6647, opened 2018-07-12
Help Text: wrap the text

ogoffart:
But add a line of spaces to make sure it is wide enough otherwise the
default messagebox size is way too small.

Issue #6644


ogoffart:
Actually, the extra information is inexact and not so useful. So I've removed it.

guruz:
screen shot 2018-07-16 at 16 32 56

Issue #6643, opened 2018-07-11
nautilus/nemo shell integration causes python AttributeErrors


jnweiger:
while testing 2.4.4 rc1, this is seen on stdout of nemo:
<pre>
Traceback (most recent call last):
  File "/usr/share/nemo-python/extensions/syncstate-testpilotcloud.py", line 217, in get_file_items
    state_ok = state.startswith('OK')
AttributeError: 'NoneType' object has no attribute 'startswith'
</pre>
seems to happen during resync, when non-ascii characters are present in file names.

jnweiger:
Same error seen wiht nautilus extension:
<pre>
Traceback (most recent call last):
  File "/usr/share/nautilus-python/extensions/syncstate-testpilotcloud.py", line 217, in get_file_items
    state_ok = state.startswith('OK')
AttributeError: 'NoneType' object has no attribute 'startswith'

michaelstingl:
@ckamm @ogoffart Is there an easy way to improve here?

guruz:
Is your python python2 or python3?
Do you have the filename too? Feel free to obsofuscate, just the non-ASCII is important I guess

guruz:
How does it look UI wise in nemo?
No overlays shown at all? 
Any horrible error dialog popping up?

guruz:
Linking https://github.com/owncloud/client/issues/6406#issuecomment-404466187

ogoffart:
Is this a regression from 2.4.1?  Is the problem still here in 2.5?  (I was not able to reproduce the problem in 2.5)

jnweiger:
Oh, I forgot to mention the good side: overlay icons are doing fine. No errors showing on the UI.

jnweiger:
nemo says: Using python version sys.version_info(major=2, minor=7, micro=15, releaselevel='candidate', serial=1)

Example files should be
```
echo hello > 'äöü.txt'
mkdir Jü
cd Jü/
vi xxx
 ...
```
but the issue does not trigger today. Sorry.

ckamm:
I didn't dig deep enough to see why this would happen, but it seems some codepaths assign `None` to `'state'`. The above PR simply guards against this case and should be fine regardless of the source of the problem.

guruz:
> Oh, I forgot to mention the good side: overlay icons are doing fine. No errors showing on the UI.

@ckamm @ogoffart do you agree that this doesnt warrant a new rc?
[...]

PR #6655, opened 2018-07-16
Nautilus: Guard against None state

ckamm:
It seems None gets assigned to 'state' in some paths.

See #6643


ckamm:
Picked to 2.4 branch as 2a55b20b8d33ae06a924e6f5eb29680d7b4c429c

Issue #6654, opened 2018-07-13
easy setup to sync within a sync folder


jnweiger:
Tested with 242~rc1 on ubuntu 18.04 and testpilotcloud version 2.6.0daily20180713 (build 9853)

**DO NOT DO THAT ON A PRODUCTION SYSTEM**

Expected behavior: it should not be possible to have two syncs in the same folder running at the same time.

Observed: It is quite easy to do so, one client is sufficient. Havoc results, replicating the same folders inside the toplevel folder again and again, probably until MAXPATHLEN or diskquota are exhausted.

Reproduce:

* Start `testpilotcloud &`
* Connect to demo.owncloud.org user demo.
* Choose what to sync: [x] Documents 35kb, everything else unchecked.
* Wait until sync completes. Quit the client from the tray menu.
* `rm -rf ~/testpilotcloud`
* start `testpilotcloud &` and notice the expected error message:
![grafik](https://user-images.githubusercontent.com/1108546/42705127-a776bd86-86d2-11e8-9f4b-6ef316115d3f.png)
* click 'Add Folder sync connection', confirm the suggested local folder `/home/USERNAME/testpilotcloud`
* Click on `Documents`  -> Next -> Add Sync Connection

Now we have two active sync connections. The client window alternates between

![grafik](https://user-images.githubusercontent.com/1108546/42705319-36a93178-86d3-11e8-8908-50584c2ec62b.png)
and
![grafik](https://user-images.githubusercontent.com/1108546/42705557-fc9c910e-86d3-11e8-9400-43e975b1b0ad.png)

The top level folder gets synced into the Documents folder with every rediscovery loop, generating a deep hierarchy of 
```
ll testpilotcloud/Documents/Documents/Documents/Documents/Documents/
insgesamt 4968
drwxr-xr-x 4 testy testy    4096 Jul 13 19:32  ./
drwxr-xr-x 4 testy testy    4096 Jul 13 19:31  ../
drwxr-xr-x 2 testy testy    4096 Jul 13 19:32  Documents/
-rw-r--r-- 1 testy testy   36227 Jul 13 19:02  Example.odt
-rw-r--r-- 1 testy testy 5017133 Jul 13 19:02 'ownCloud Manual.pdf'
drwxr-xr-x 2 testy testy    4096 Jul 13 19:32  Photos/
```

There are two ._sync_*.db datastructures in the same folder. Both active
```
ls -la ~/testpilotcloud
total 5360
[...]

PR #6656, opened 2018-07-16
FolderMan::checkPathValidityForNewFolder: make sure to work with deleted folders

ogoffart:
Note that we also needed to adjust the server url to contains the user name
in the folder wizard. (As checkPathValidityForNewFolder expect the user name)

Issue #6654

### DIFF
--- a/src/gui/folderman.cpp
+++ b/src/gui/folderman.cpp
@@ -1306,25 +1306,45 @@ QString FolderMan::trayTooltipStatusString(
     return folderMessage;
 }
 
-QString FolderMan::checkPathValidityForNewFolder(const QString &path, const QUrl &serverUrl, bool forNewDirectory) const
+static QString checkPathValidityRecursive(const QString &path)
 {
     if (path.isEmpty()) {
-        return tr("No valid folder selected!");
+        return FolderMan::tr("No valid folder selected!");
     }
 
     QFileInfo selFile(path);
 
     if (!selFile.exists()) {
-        return checkPathValidityForNewFolder(selFile.dir().path(), serverUrl, true);
+        return checkPathValidityRecursive(selFile.dir().path());
     }
 
     if (!selFile.isDir()) {
-        return tr("The selected path is not a folder!");
+        return FolderMan::tr("The selected path is not a folder!");
     }
 
     if (!selFile.isWritable()) {
-        return tr("You have no permission to write to the selected folder!");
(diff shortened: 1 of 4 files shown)


ogoffart:
If you are wondering why I removed the wording with symlinks, it is because i now always use the canonical path. And there is no need for a different wording with symbolic links.

src/gui/folderman.cpp
```diff
@@ -1306,83 +1306,81 @@ QString FolderMan::trayTooltipStatusString(
...
         if (folderDir.isEmpty()) {
             continue;
         }
-        if (!folderDir.endsWith(QLatin1Char('/'), cs))
-            folderDir.append(QLatin1Char('/'));
-
         const QString folderDirClean = QDir::cleanPath(folderDir) + '/';
-        const QString userDirClean = QDir::cleanPath(path) + '/';
 
         // folderDir follows sym links, path not.
```
ckamm: Since both are canonicalized, that's probably an outdated comments

src/gui/folderman.cpp
```diff
@@ -1306,83 +1306,81 @@ QString FolderMan::trayTooltipStatusString(
...
     Qt::CaseSensitivity cs = Qt::CaseSensitive;
     if (Utility::fsCasePreserving()) {
         cs = Qt::CaseInsensitive;
     }
 
+    const QString userDirClean = QDir::cleanPath(canonicalPath(path)) + '/';
     for (auto i = _folderMap.constBegin(); i != _folderMap.constEnd(); ++i) {
         Folder *f = static_cast<Folder *>(i.value());
-        QString folderDir = QDir(f->path()).canonicalPath();
+        QString folderDir = canonicalPath(f->path());
```
ckamm: Would it be possible to drop this `folderDir` and then drop the `Clean` postfix from userDir and folderDir? I don't see where the non-cleaned variants are used in the first place.

ogoffart:
Thanks for the comments.  (I've updated the commit)

ckamm:
Looks good!

Minor on rereading, feel free to skip:
- maybe move the `+ '/'` inside the `cleanPath()` to make clear there'll not be a double trailing /
- `differentPathes` -> `differentPaths`

ogoffart:
> maybe move the + '/' inside the cleanPath() to make clear there'll not be a double trailing /

cleanPath removes trailing '/'

PR #6669, opened 2018-07-23
Convert p7.pl to a C++ test

ogoffart:
This is just a translation of test/scripts/txpl/t7.pl to C++ using the test
framework.


dschmidt:
Can't the perl test be removed then?

test/testpermissions.cpp
```diff
@@ -0,0 +1,306 @@
...
+    }
+    return nullptr;
+}
+
+// Check if the expected rows in the DB are non-empty. Note that in some cases they might be, then we cannot use this function
+// https://github.com/owncloud/client/issues/2038
+static void assertCsyncJournalOk(SyncJournalDb &journal)
+{
+    SqlDatabase db;
+    QVERIFY(db.openReadOnly(journal.databaseFilePath()));
```
ckamm: Interesting! I'd probably have given tests access to the underlying database, but making a new connection works as well.

PR #6671, opened 2018-07-25
Do not require server replies to contain an mtime

ckamm:
The check was added for #6317 in commit
13eb64584f5f96647ced39dcd3252860ebec5a37.

We did see missing mtimes in replies in tests with live servers though.
Possibly those were old incomplete responses cached in the stat cache?


guruz:
CC @PVince81 @DeepDiver1975

PVince81:
raised https://github.com/owncloud/core/issues/32153 to return a proper value instead of removing the property in case the mtime in filecache is 0

PR #6672, opened 2018-07-25
Settings: Fix rename issue on old macOS

guruz:
Found by @hurradieweltgehtunter

### DIFF
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -186,7 +186,8 @@ Application::Application(int &argc, char **argv)
     setAttribute(Qt::AA_UseHighDpiPixmaps, true);
 
     auto confDir = ConfigFile().configPath();
-    if (!QFileInfo(confDir).exists()) {
+    if (confDir.endsWith('/')) confDir.chop(1);  // macOS 10.11.x does not like trailing slash for rename/move.
+    if (!QFileInfo(confDir).isDir()) {
         // Migrate from version <= 2.4
         setApplicationName(_theme->appNameGUI());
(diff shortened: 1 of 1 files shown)


src/gui/application.cpp
```diff
@@ -186,7 +186,8 @@ Application::Application(int &argc, char **argv)
     setAttribute(Qt::AA_UseHighDpiPixmaps, true);
 
     auto confDir = ConfigFile().configPath();
-    if (!QFileInfo(confDir).exists()) {
+    if (confDir.endsWith('/')) confDir.chop(1);  // macOS 10.11.x does not like trailing slash for rename/move.
```
ogoffart: Note:  confDir ALWAYS ends in a slash  (by construction)

src/gui/application.cpp
```diff
@@ -186,7 +186,8 @@ Application::Application(int &argc, char **argv)
...
+    if (!QFileInfo(confDir).isDir()) {
```
ogoffart: and if it is a file?
You can't do the move in this case anyways either. So this changes change nothing.

src/gui/application.cpp
guruz: Yep, if this ever changes then this migration doesnt break :) So I prefer to keep the checks explicit.

src/gui/application.cpp
guruz: I think it was always a dir, even in old mirall versions. It's better to use the more explicit functions so that the code states the intention (further down correctly isDir is used)

Issue #6675, opened 2018-07-27
selective sync "Apply" stuck if another sync folder has error (missing)


moscicki:
MacOSX, client 2.4.1.

I have several sync folders. One of them is "broken" because it is missing completely (let's call it folder X).

![image](https://user-images.githubusercontent.com/621965/43327979-2d221f4e-91bd-11e8-929a-13adf5f9fd51.png)

However, another folder is OK (just some unresolved conflicts):
![image](https://user-images.githubusercontent.com/621965/43328014-3f9764cc-91bd-11e8-8e96-aa40ce543271.png)

When I try to use selective sync on my OK folder I am not able to because "Apply" button seems to be stuck:

![image](https://user-images.githubusercontent.com/621965/43328080-67127910-91bd-11e8-84a2-f88ed2e2eeca.png)

Whenever I click "Apply" I see these error messages (related to the "broken" folder X).
So there is some negative interaction for Selective Sync between the two under the hood:

```
07-27 16:47:34:198 [ warning sync.database.sql ]:	Error: "unable to open database file" for "/.../X/._sync_b2f43b5ea209.db"
07-27 16:47:34:198 [ warning sync.database.sql ]:	CANTOPEN extended errcode:  14
07-27 16:47:34:199 [ warning sync.database.sql ]:	CANTOPEN system errno:  2
07-27 16:47:34:199 [ warning sync.database ]:	Error opening the db:  "unable to open database file"
07-27 16:47:34:199 [ warning gui.folder.model ]:	Could not read selective sync list from db.
```

guruz:
@moscicki You can test this in a recent macOS 2.5.0 nightly.
E.g. https://download.owncloud.com/desktop/daily/ownCloud-qt5.10.1-2.5.0.10395-daily20180817.pkg

jnweiger:
### testing macos 10.14 Mojave

ownCloud-2.5.1.10814rc2.pkg

connect one account to demo.owncloud.com -> sync folder ownCloud
connect another account to demo.owncloud.org -> sync folder ownCloud2

use a shell to `rm -rf ~/ownCloud2`

force sync on the second account, the red error message `unable to find local sync folder` appears as expected.

* click open the folder tree on the other account, 
* deselect e.g. the Documents folder. 
* Click `Apply`.
-> The subfolder `Documents` gets removed. -> OK

The bug seems to be fixed. 
However, after a few minutes the client crashes. 
[...]

PR #6678, opened 2018-08-01
FolderStatusModel: Do not abort applying selective sync if one folder…

ogoffart:
… has an error

Issue #6675


ckamm:
@ogoffart Hmm, do we ever silently not apply selective sync changes that way? I guess not, since the only reason for the selective-sync list to fail to read from db is the db not existing? I wonder if there could be a better check since silently discarding changes would be unfortunate.

ogoffart:
> do we ever silently not apply selective sync changes that way?

Yes we do!  With this patch, at least, we don't silently ignore change from folder which do not have an error.

> I wonder if there could be a better check since silently discarding changes would be unfortunate.

Right, discarding the change is bad. But this is only when there is a database error, which should not be too often.

Issue #6694, opened 2018-08-06
Quick change in folder hierarchy causes sync errors


jnweiger:
Revert the hierarchy of two enabled folders **quickly**
 - web interface: create an empty folder in Documents called Empty.
 - owncloud client: Enable only Documents/Empty
 - `mv Documents/Empty . ; mv Documents Empty`

BAD:
Error message in client log
 6 Aug 2018 15:22:36,                                                 Documents/Empty,               testpilotcloud2,Error transferring https://demo.owncloud.org/remote.php/dav/files/demo/Empty/Documents/Empty - server replied:

![grafik](https://user-images.githubusercontent.com/1108546/43720817-6e846f74-9991-11e8-984a-70cc62477b93.png)

Error message in server log
<pre>
{"reqId":"fd27578c-dd20-4954-a8b1-9cab2c8bb3cb","level":2,"time":"2018-08-06T13:14:29+00:00","remoteAddr":"192.168.48.16","user":"demo","app":"dav","method":"MOVE","url":"\/remote.php\/dav\/files\/demo\/Empty\/Documents\/Empty","message":"Could not get node for path: \"files\/demo\/Empty\/Documents\/Empty\" : File with name Empty\/Documents\/Empty could not be located"}
{"reqId":"27ac4e76-b8ea-47bf-b826-414a1a85b84f","level":2,"time":"2018-08-06T13:14:44+00:00","remoteAddr":"192.168.48.16","user":"demo","app":"dav","method":"MOVE","url":"\/remote.php\/dav\/files\/demo\/Empty\/Documents\/Empty","message":"Could not get node for path: \"files\/demo\/Empty\/Documents\/Empty\" : File with name Empty\/Documents\/Empty could not be located"}
{"reqId":"22e1bf9f-3922-47b6-a881-559904da1d8d","level":2,"time":"2018-08-06T13:14:58+00:00","remoteAddr":"192.168.48.16","user":"demo","app":"dav","method":"MOVE","url":"\/remote.php\/dav\/files\/demo\/Empty\/Documents\/Empty","message":"Could not get node for path: \"files\/demo\/Empty\/Documents\/Empty\" : File with name Empty\/Documents\/Empty could not be located"}
</pre>

Almost good:
Do the same slowly 
 * `mv Documents/Empty . ; sleep 10; mv Documents Empty  # sufficient sleep so that a sync run completes.`
[...]

PR #6695, opened 2018-08-07
SyncEngine: Fix renames making hierarchy inversion

ogoffart:
Issue #6694

This is not a problem in the `new_discovery_algo` branch, but let's fix that for 2.5

### DIFF
--- a/src/libsync/syncengine.cpp
+++ b/src/libsync/syncengine.cpp
@@ -1245,6 +1245,14 @@ QString SyncEngine::adjustRenamedPath(const QString &original)
     while ((slashPos = original.lastIndexOf('/', slashPos - 1)) > 0) {
         QHash<QString, QString>::const_iterator it = _renamedFolders.constFind(original.left(slashPos));
         if (it != _renamedFolders.constEnd()) {
+            // This is a band-aid fix for a specific problem:
+            // See issue #6694: "hierarchy inversion" and discussion in PR #6695.
+            // There's a larger unfixed issue here, see testInvertFolderHierarchy().
+            auto original_it = _renamedFolders.constFind(original);
+            if (original_it != _renamedFolders.constEnd() && it->startsWith(*original_it)) {
+                return original;
+            }
+
             return *it + original.mid(slashPos);
         }
     }
--- a/test/testsyncmove.cpp
+++ b/test/testsyncmove.cpp
@@ -621,6 +621,42 @@ private slots:
 
         QCOMPARE(fakeFolder.currentLocalState(), fakeFolder.currentRemoteState());
     }
+
+    // Test for https://github.com/owncloud/client/issues/6694
+    void testInvertFolderHierarchy()
+    {
+        FakeFolder fakeFolder{ FileInfo::A12_B12_C12_S12() };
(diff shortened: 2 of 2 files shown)


src/libsync/syncengine.cpp
```diff
@@ -1245,6 +1245,11 @@ QString SyncEngine::adjustRenamedPath(const QString &original)
     while ((slashPos = original.lastIndexOf('/', slashPos - 1)) > 0) {
         QHash<QString, QString>::const_iterator it = _renamedFolders.constFind(original.left(slashPos));
         if (it != _renamedFolders.constEnd()) {
+            auto original_it = _renamedFolders.constFind(original);
+            if (original_it != _renamedFolders.constEnd() && it->startsWith(*original_it)) {
+                return original; // Issue #6694: "hirarchy inversion"
+            }
```
ckamm: I don't get this logic. Are you sure it doesn't indicate we should reverse the order of traversing paths in the `while`?

As far as I understand (haven't tested!) our `renamedFolders` will have two entries if "/Documents/Empty" is renamed to "/Empty/Documents":
```
/Documents -> /Empty/Documents
/Documents/Empty -> /Empty
```

Previously the function should have mapped "/Documents/Empty" to "/Empty/Documents/Empty" because the first parent directory rename is appied. With your change it'll be "/Empty/Documents" because the full path rename is taken instead.

There are probably corner cases now with three-level paths? Why aren't we taking the longest applicable rename path instead of the shortest?

test/testsyncmove.cpp
```diff
@@ -621,6 +621,42 @@ private slots:
...
+        // "AllEmpty" is before "C", alphabetically
+        fakeFolder.localModifier().rename("C/AllEmpty", "AllEmpty");
+        fakeFolder.localModifier().rename("C", "AllEmpty/C");
+
+        auto expectedState = fakeFolder.currentLocalState();
+        QVERIFY(fakeFolder.syncOnce());
+        QCOMPARE(fakeFolder.currentLocalState(), expectedState);
+        QCOMPARE(fakeFolder.currentRemoteState(), expectedState);
+
+        /* FIXME
```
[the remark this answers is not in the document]
ogoffart: Yes, this test is still failing. Fixing it would be quite some work and i think this can wait 2.6

src/libsync/syncengine.cpp
ogoffart: > Previously the function should have mapped "/Documents/Empty" to "/Empty/Documents/Empty" because the first parent directory rename is appied.

Yes, that's correct and that was the bug.

But with my change it will return the unchanged original .

Yes, there will be problem with three-level path when each level is renamed, or with cycle that involve several folder (as the FIXME test is testing).  But this fix for that is more complicated.

ckamm: Ok! Then the problem was that there was no indication that this was just fixing the most common problem in time for 2.5. Could you add something along the lines of `// band-aid to fix a common problem, there's a larger issue here with ... like in example ...`?

test/testsyncmove.cpp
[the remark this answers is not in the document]
ckamm: Agree, since it hasn't come up before...

PR #6698, opened 2018-08-07
OAuth: Fix infinite loop when the refresh token is expired

ogoffart:
The server reply with a code 400 when the token is invalid,
the client was understanding this error as a network error, and was retying
again with the same token.

Instead, we must rely on what the json is saying, even if the reply is
not a 200 code.

Issue https://github.com/owncloud/enterprise/issues/2777

(Need to be backported in 2.4.x)


ogoffart:
Cherry-picked to 2.4 as 235e314da707b1cbdb6af093d260885f424ffd07

PR #6700, opened 2018-08-08
use api/v2/shares with reshares=true

jnweiger:
as recommended in https://github.com/owncloud/android/issues/2264#issuecomment-407453191


guruz:
Should this PR be based on 2.5?
(We can also change it when merging)

guruz:
v1.php vs v2.php is just the ocs return code difference.

Still wondering about https://github.com/owncloud/android-library/pull/194/files#diff-f0b1c4e006d58debc400c9f64ca7f165R40 then though CC @davigonz @PVince81 why does Android use api/v1/shares ?

Issue #6699, opened 2018-08-07
Crashing after some time on open/locked big files for eg. *.pst


piojan:
### Actual behaviour
After few days of running client 2.4.2 crashes. Earlier had the same problem.

### Steps to reproduce
Not sure what's causing the issue. According to the logs the owncloud client crashes upon dealing with locked (opened) files. In this case a >200MB *.pst file. But that only happenes after few days of working correctly. The file is constantly open in Outlook.

### Server configuration
Operating system: Linux Debian
Web server: Apache
Database: MySql
PHP version: 5.7
ownCloud version: 10..0.8
Storage backend (external storage): local storage

### Client configuration
Client version: 2.4.2
Operating system: Windows 7
OS language: PL
Installation path of client: C:\Program Files 

### Logs

Please use Gist (https://gist.github.com/) or a similar code paster for longer
logs.
```
tail -1 /tmp/owncloud.log
07-25 16:01:15:819 [ info sync.propagator.upload ]:     Resuming  "Pliki programu Outlook/Archiwum 2016.pst"  from chunk  0 ; sent = 0

grep -A 3 -E'Resuming.*Archiwum 2016.pst' /tmp/owncloud.log


07-25 15:46:06:463 [ info sync.propagator.upload ]:     Resuming  "Pliki programu Outlook/Archiwum 2016.pst"  from chunk  0 ; sent = 0
07-25 15:46:06:466 [ warning sync.propagator.upload ]:  Could not prepare upload device:  "Permission denied"
07-25 15:46:06:466 [ info gui.lockwatcher ]:    Watching for lock of "C:/Users/pjanusz/Documents/Pliki programu Outlook/Archiwum 2016.pst" being released
07-25 15:46:06:466 [ warning sync.propagator ]: Could not complete propagation of "Pliki programu Outlook/Archiwum 2016.pst" by OCC::PropagateUploadFileNG(0x6acde28) with status 3 and error: "Permission denied"
--
07-25 15:50:11:433 [ info sync.propagator.upload ]:     Resuming  "Pliki programu Outlook/Archiwum 2016.pst"  from chunk  0 ; sent = 0
07-25 15:50:11:437 [ warning sync.propagator.upload ]:  Could not prepare upload device:  "Permission denied"
07-25 15:50:11:438 [ info gui.lockwatcher ]:    Watching for lock of "C:/Users/pjanusz/Documents/Pliki programu Outlook/Archiwum 2016.pst" being released
[...]

PR #6707, opened 2018-08-14
Windows: Release handle/fd when file open fails 

ckamm:
See #6699


ogoffart:
It is true that a test for currently opened files should be nice. But that's hard to do in a cross platform way.  (but it could be windows-only)

(See also t4.pl which also do some tests which are supposed to cover some of the locked files cases)

Issue #6702, opened 2018-08-09
OC-ConflictBasePath is missing when there was a conflict upload


jnweiger:
The PUT request that uploads a conflict file after the client detected a conflict, includes several headers with metadata about the conflict:

    OC-Conflict:            1
    OC-ConflictBaseEtag:    4d9bd6cb0b8d2d090aa09b06dad4944f
    OC-ConflictBaseFileId:  04400434oce9akr7u3v9
    OC-ConflictBaseMtime:   1533747765
    X-OC-Mtime:             1533747765

The path of the conflicting file is not here. But it may be helpful information later. E.g. when the file on the server gets replaced with a new incarnation but same name. 
A header with the path was promised in https://github.com/owncloud/client/issues/4557#issuecomment-349971567 but not found in current code.

jnweiger:
When the uploadConflictFiles feature is disabled, the client will silently ignore a file that I create via
`echo "not a conflict" > "filename (conflicted copy jwtest 2018-08-08 201107).txt"`
even if no corresponding `filename.txt` exists.

Triggering any logic on the structure of such names is a hack.
The conflict marker is currently hardcoded in
  https://github.com/owncloud/client/blob/master/src/common/utility.cpp#L558
Proper meta data stored in the server should help to obsolete such hacks.

ckamm:
@jnweiger I intentionally removed the path metadata as there is no guarantee that the base file location stays unchanged. The fileid and etag will clearly identify the path and even revision of the file as long as it hasn't been deleted and thus seemed much more durable than a path.

ckamm:
On the magic "(conflicted copy" tag: I agree that it's not pretty. With the added metadata the client would be able to stop depending on file name tags. But as long as the metadata isn't reliably available in all servers the client must be able to work without it.

Note also that this is not new: previously the magic tag was "_conflict-"

ogoffart:
@ckamm:  I added the BasePath header in https://github.com/owncloud/client/pull/6709

It's true that fileid should be enough, but basePath is still a quite usefull information.

ogoffart:
Longterm, we probably should support translating the "(conflicted copy"  so we should indeed not rely on that.  But that's out of scope for 2.5 I'd say.

guruz:
Related: https://github.com/owncloud/client/pull/6736

PR #6709, opened 2018-08-14
Conflict handling: add the OC-ConflictBasePath header

ogoffart:
Issue #6702


ckamm:
I worry that adding the path will lead to people using it when they should query by fileid instead. I had the path there initially and removed it, thinking it was a bad idea.

PR #6711, opened 2018-08-15
C++: Don't use auto parameters in lambdas

ckamm:
The mingw compiler isn't happy with them.


ogoffart:
how old is mingw? no C++14 support?

ckamm:
@ogoffart I was using the mingw that ships with the Qt SDK. It was 5.3.0 32bit, `g++ --version` is 5.3.0 too.

guruz:
This was also a problem for CentOS_6/x86_64.. thanks for fixing @ckamm

Issue #6718, opened 2018-08-15
Renaming virtual files does not work correctly


jnweiger:
### tested on Linux Mint Tara, testpilotcloud-client-2.5.0~daily20181014

    [General]
    showExperimentalOptions=true

* Renaming a file foo.txt.testpilotcloud_virtual to bettername.txt.testpilotcloud_virtual 
in a local sync folder has no effect. The original name is restored 2 or 3 seconds later.
* Renaming a subfolder in the local sync folder works as expected, though.
* Renaming a local file Photos/Paris.jpg.testpilotcloud_virtual to Photos/Paris.jpg almost does the right thing:
  - Message: Server version downloaded, copied changed local file into conflict file
  - Message: Conflict: Server version downloaded, local copy renamed and not uploaded.
 
OK:
The virtual file gets replaced by the original file. Nice!

BAD
A self-conflict is generated `Paris (conflicted copy 2018-08-15 220222).jpg`

* Renaming a fully downloaded file to its virtual placeholder should do the obvious.
  `mv Photos/Paris.jpg Photos/Paris.jpg.testpilotcloud_virtual`

BAD
Data loss! The file gets removed on the server and on the client
  - Message: Photos/Paris.jpg Deleted

### tested win10 owncloud-client 2.5.0beta1

C:\Users\IEUser\AppData\Roaming\ownCloud\owncloud.cfg

    [General]
    showExperimentalOptions=true

* The suffix in win10 is  .owncloud (instead of .owncloud_virtual)

BAD
Same issues including data loss through renaming.

ckamm:
Virtual file suffix is sometimes .owncloud and sometimes .owncloud_virtual -> #6719

Let's make this issue purely about 
1. renaming virtual files (a.owncloud -> b.owncloud)
2. renaming virtual files to actual files (a.owncloud -> a; a.owncloud -> b)
3. renaming files to virtual files (a -> a.owncloud; a -> b.owncloud)

jnweiger:
Edited back to the truth: Tests were done with testpilotcloud actually. That makes a difference, unfortunately. Sorry.

jnweiger:
### Linux Mint tara, ownCloud version 2.5.0filerenametovirtual6718 (build 10079)
Git revision 5d3d6d362b795e3954875d2608ad7523f31c2ea3

OK
this fixes the data loss. 

```
mv .../somewhere/stars.svg Photos
sleep 10
mv Photos/stars.svg{,.owncloud}
sleep 10
[...]

PR #6721, opened 2018-08-16
Virtual files: Renaming to virtual doesn't delete data

ckamm:
Unfortunately to do this, the local update phase must write to the
database, creating a new side-effect and order dependency (local update
must run before remote update).

For  #6718

### DIFF
--- a/src/libsync/propagatedownload.cpp
+++ b/src/libsync/propagatedownload.cpp
@@ -355,6 +355,8 @@ void PropagateDownloadFile::start()
     if (_item->_type == ItemTypeVirtualFile) {
         auto fn = propagator()->getFilePath(_item->_file);
         qCDebug(lcPropagateDownload) << "creating virtual file" << fn;
+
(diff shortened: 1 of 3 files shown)


src/libsync/propagatedownload.cpp
```diff
@@ -355,6 +355,8 @@ void PropagateDownloadFile::start()
     if (_item->_type == ItemTypeVirtualFile) {
         auto fn = propagator()->getFilePath(_item->_file);
         qCDebug(lcPropagateDownload) << "creating virtual file" << fn;
+
+        // NOTE: Other places might depend on shape of placeholder files (like csync_update)
```
guruz: shape of a file is what? rectangle? square? circle?

ckamm: everyone knows it's a rectangle, doh :P

ogoffart: yeah  s/shape/content/  ?

ckamm: Yeah, agree :)

Issue #6723, opened 2018-08-16
[VFS] Indicator needed for account created in VFS mode


michaelstingl:
Currently there is no information for the user that one of the accounts has been created in VFS mode. (tested with todays daily 2.5.0 build)

![owncloud](https://user-images.githubusercontent.com/214010/44227727-a8db2a00-a193-11e8-8c8b-57e5cb15ad39.png)

Could we add this information in a similar style as warnings?

![owncloud-2](https://user-images.githubusercontent.com/214010/44228211-ff953380-a194-11e8-9790-1ea2f7503a8b.png)

guruz:
2.5.1 as in 2.5.0 at least the virtual file mode is still experimental.

ogoffart:
Proposal, with screenshot on https://github.com/owncloud/client/pull/6733

guruz:
@michaelstingl feel free to have a look in the nightly tomorrow.

guruz:
Looks fine on macOS, however the green bar has a "Hand" icon that leads me to the "Not synced" tab. @ogoffart Can you make it open the file manager instead? (without breaking the "Not synced" feature for other bars)

michaelstingl:
LGTM

![msedge_-_win10](https://user-images.githubusercontent.com/214010/44998041-a90c5f80-afb2-11e8-895f-7d498fce9c3c.png)

PR #6733, opened 2018-08-18
Account Settings: Add a text for folder using vierual files

ogoffart:
Issue #6723


ogoffart:
screenshot_20180818_093913

PR #6735, opened 2018-08-21
Account Settings: the info text should not be clickable

ogoffart:
Adjustment for #6723

### DIFF
--- a/src/gui/folderstatusdelegate.cpp
+++ b/src/gui/folderstatusdelegate.cpp
@@ -375,13 +375,22 @@ QRect FolderStatusDelegate::optionsButtonRect(QRect within, Qt::LayoutDirection 
     return QStyle::visualRect(direction, within, r);
 }
 
-QRect FolderStatusDelegate::errorsListRect(QRect within)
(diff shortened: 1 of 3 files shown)


src/gui/folderstatusdelegate.cpp
```diff
@@ -375,13 +375,22 @@ QRect FolderStatusDelegate::optionsButtonRect(QRect within, Qt::LayoutDirection
...
     within.setTop(within.top() + FolderStatusDelegate::rootFolderHeightWithoutErrors(fm, aliasFm));
+    int margin = fm.height() / 4;
+    int h = 0;
+    for (auto role : {FolderConflictMsg, FolderErrorMsg}) {
+        auto msgs = qvariant_cast<QStringList>(index.data(role));
+        if (!msgs.isEmpty()) {
+            h += margin + 2 * margin + msgs.count() * fm.height();
+        }
+    }
+    within.setHeight(h);
```
ckamm: Looks ok. Manually drawing and interacting like this is really really awkward though. Situations like this make me wish it was either widgets or qml -- or anything that abstracts all of this away in a robust manner.

PR #6736, opened 2018-08-21
Conflicts: Rename BasePath to InitialBasePath

ckamm:
To make clearer that this path is not unlikely to be outdated.

Amends #6709


ogoffart:
Maybe one should involve some server person:  @PVince81 

Did you keep the database column name on purpose?

ckamm:
@ogoffart Yes, kept on purpose, that seemed good enough. My concern only was that we use that field when we shouldn't, or that the server uses the field when it shouldn't. I'm not aware of server work having started on storing conflict metadata - @PVince81 let me know if that's changed.

PR #6738, opened 2018-08-24
New discovery algo

ogoffart:


src/libsync/discovery.h
```diff
@@ -0,0 +1,106 @@
...
+            result._local = buildString(_local);
+            return result;
+        }
+    };
+    void process();
+    // return true if the file is excluded
+    bool handleExcluded(const QString &path, bool isDirectory, bool isHidden, bool isSymlink);
+    void processFile(PathTuple, const LocalInfo &, const RemoteInfo &, const SyncJournalFileRecord &dbEntry);
+    // Return false if there is an error and that a directory must not be recursively be taken
+    bool checkPremission(const SyncFileItemPtr &item);
```
ckamm: Permission

src/libsync/discovery.cpp
```diff
@@ -0,0 +1,1153 @@
...
+        if (errno != 0) {
+            // Note: Windows vio converts any error into EACCES
+            qCWarning(lcDisco) << "readdir failed for file in " << _currentFolder._local << " - errno: " << errno;
+            emit _discoveryData->fatalError(tr("Error while reading directory %1").arg(_discoveryData->_localDir + _currentFolder._local));
+            emit finished();
+        }
+        csync_vio_local_closedir(dh);
+    }
+    _hasLocalEntries = true;
+    if (_hasServerEntries)
```
ckamm: Add a comment about `process()` getting called when both local and server entries are available to make this clearer? Possibly just have a `processWhenDone()` func to call in both places?

src/libsync/discovery.h
```diff
@@ -0,0 +1,106 @@
...
+
+class ProcessDirectoryJob : public QObject
+{
+    Q_OBJECT
+public:
+    enum QueryMode {
+        NormalQuery,
+        ParentDontExist,
+        ParentNotChanged,
+        InBlackList
```
ckamm: Note that InBlackList can apply only to server queries

src/libsync/discovery.cpp
```diff
@@ -0,0 +1,1153 @@
...
+        }
+    } else if (_queryLocal == ParentNotChanged && dbEntry.isValid()) {
+        if (_queryServer != ParentNotChanged && !serverEntry.isValid()) {
+            // Not modified locally (ParentNotChanged), bit not on the server:  Removed on the server.
+            item->_instruction = CSYNC_INSTRUCTION_REMOVE;
+            item->_direction = SyncFileItem::Down;
+        }
+    } else if (_queryServer != ParentNotChanged && !serverEntry.isValid()) {
+        // Not locally, not on the server. The entry is stale!
+        qCInfo(lcDisco) << "Stale DB entry";
```
ckamm: move this check very far up if possible

ogoffart:
I removed the WIP tag.

I consider that this branch has feature parity with master.  
And it even support parallel PROPFIND, which makes the discovery much faster

ckamm:
I didn't get around to starting the review this week because of other issues - but next week should be good.

src/libsync/discovery.cpp
```diff
@@ -0,0 +1,1218 @@
...
+                    return;
+                }
+
+                // Now we know there is a sane rename candidate.
+                QString originalPath = QString::fromUtf8(base._path);
+
+                // Rename of a virtual file
+                if (base._type == ItemTypeVirtualFile && item->_type == ItemTypeFile) {
+                    // Ignore if the base is a virtual files
+                    return;
```
ckamm: Shouldn't this prohibit renames on the server propagating to renaming the local virtual files? But since testRenameVirtual passes - why doesn't it?

src/libsync/discovery.cpp
```diff
@@ -0,0 +1,1218 @@
...
+                }
+
+                if (item->_type == ItemTypeFile) {
+                    csync_file_stat_t buf;
+                    if (csync_vio_local_stat((_discoveryData->_localDir + originalPath).toUtf8(), &buf)) {
+                        qCInfo(lcDisco) << "Local file does not exist anymore." << originalPath;
+                        return;
+                    }
+                    if (buf.modtime != base._modtime || buf.size != base._fileSize || buf.type != ItemTypeFile) {
+                        qCInfo(lcDisco) << "File has changed locally, not a rename." << originalPath;
```
ckamm: Some above set done=true; - these don't. It's inconsistent but probably doesn't cause bugs. I'm unconvinced what the best way of doing it would be.

src/libsync/discovery.cpp
```diff
@@ -0,0 +1,1218 @@
...
+
+                        // The file do not exist, it is a rename
+
+                        // In case the deleted item was discovered in parallel
+                        auto it = _discoveryData->_deletedItem.find(originalPath);
+                        if (it != _discoveryData->_deletedItem.end()) {
+                            ASSERT((*it)->_instruction == CSYNC_INSTRUCTION_REMOVE);
+                            (*it)->_instruction = CSYNC_INSTRUCTION_NONE;
+                        }
+                        delete _discoveryData->_queuedDeletedDirectories.take(originalPath);
```
ckamm: This check-local, async-check, check-local-again pattern is a bit awkward but looks correct.

One worry: since the async job is started before done is true, could there be multiple deletion-check jobs running at the same time? Then when they finish both could consume the rename?

done=true before going async? (depends on how friendly we want to be towards duplicate fileids)

test/testsyncvirtualfiles.cpp
```diff
@@ -145,7 +145,7 @@ private slots:
         QVERIFY(fakeFolder.currentLocalState().find("A/a1m.owncloud"));
         QVERIFY(!fakeFolder.currentRemoteState().find("A/a1"));
         QVERIFY(fakeFolder.currentRemoteState().find("A/a1m"));
-        QVERIFY(itemInstruction(completeSpy, "A/a1m.owncloud", CSYNC_INSTRUCTION_RENAME));
+        //QVERIFY(itemInstruction(completeSpy, "A/a1m.owncloud", CSYNC_INSTRUCTION_RENAME));
```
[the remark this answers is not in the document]
ogoffart: I don't remember, but i write in the commit message of a299907c5926e946a03afd6df74da99eb3a4cd47 that it was implementation detail.

I wonder if i made that placeholders are no longer renamed but just re-created.  
I'll have another look.  But I guess if that's the case, it is still ok, since there is no metadata in the placeholder themselfs.

src/libsync/abstractnetworkjob.h
```diff
@@ -24,13 +24,71 @@
...
+    QString errorMessage() const
+    {
+        ASSERT(_isError);
+        return _error.string;
+    }
+    int errorCode() const
+    {
+        return _isError ? _error.code : 0;
+    }
+};
```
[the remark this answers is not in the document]
ogoffart: Right...  I just needed it, and did not want to implement too much, knowing how much bikeshed there can be for such utility class.

It's indeed based on rust's `Result`, and LLVM's `Expected` class. (there is also a std::expected proposal, but it is just a source of bikeshedding :-))
I liked the name Result better than Expected, so i named it result.  But it can be easilly changed as there are only a few use of it right now. (altough i would plan to use it everywhere in the future)

src/common/syncjournalfilerecord.h
```diff
@@ -61,7 +61,7 @@ class OCSYNC_EXPORT SyncJournalFileRecord
 
     QByteArray _path;
     quint64 _inode;
-    qint64 _modtime;
```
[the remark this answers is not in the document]
ogoffart: The other ones are initialized in the constructor. But I'll just initialize everything inline, it's much more readable.

src/libsync/discoveryphase.cpp
```diff
@@ -100,156 +87,90 @@ bool DiscoveryJob::checkSelectiveSyncNewFolder(const QString &path, RemotePermis
...
 {
-    return static_cast<DiscoveryJob *>(data)->checkSelectiveSyncNewFolder(QString::fromUtf8(path), remotePerm);
+    int slashPos = original.size();
+    while ((slashPos = original.lastIndexOf('/', slashPos - 1)) > 0) {
+        auto it = _renamedItems.constFind(original.left(slashPos));
+        if (it != _renamedItems.constEnd()) {
+            return *it + original.mid(slashPos);
+        }
+    }
+    return original;
```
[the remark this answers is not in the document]
ogoffart: We don't need them because we track the exact paths in the different part of `ProcessDirectoryJob::PathTuple`, so there is no need to look up things later.

(In theory, this should allow more complex renames to happen  (rename A/B/ to A/B2/ on the server, and A/B/C to A/B/C2 on the client) although i haven't test that yet.)

src/libsync/discovery.h
```diff
@@ -0,0 +1,114 @@
...
+#include "discoveryphase.h"
+#include "syncfileitem.h"
+#include "common/asserts.h"
+
+class ExcludedFiles;
+
+namespace OCC {
+class SyncJournalDb;
+
+class ProcessDirectoryJob : public QObject
```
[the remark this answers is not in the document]
ckamm: Doc comment looks good, minor suggestions:
- mention that results are fed outwards via DiscoveryPhase::itemDiscovered
- "invormation", "that handle"->"that handles", "what need"->"what needs", "couple with"->"coupled with", "it emit"->"it emits"

src/libsync/discoveryphase.cpp
```diff
@@ -100,156 +87,90 @@ bool DiscoveryJob::checkSelectiveSyncNewFolder(const QString &path, RemotePermis
...
-        }
-        break;
-    case 413: /* Request Entity too Large */
-        new_errno = EFBIG;
-        break;
-    default:
-        new_errno = EIO;
+    auto limit = qMax(1, _syncOptions._parallelNetworkJobs);
+    if (_currentRootJob && _currentlyActiveJobs < limit) {
+        _currentRootJob->progress(limit - _currentlyActiveJobs);
```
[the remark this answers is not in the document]
ckamm: Hm. I think it's essential that it gets renamed since currently there's `progress()` and `process()` and that's quite confusing. One is for processing the directory itself, one is for processing the sub jobs that were spawned.

Maybe `processSubJobs()` then, with some added documentation? It seems dealing with the queued up jobs is the second phase of ProcessDirectoryJob. Maybe the two aspects could even be disentangled further?

src/libsync/syncengine.cpp
```diff
@@ -365,395 +286,101 @@ void SyncEngine::conflictRecordMaintenance()
...
+            // If the 'W' remote permission changed, update the local filesystem
+            SyncJournalFileRecord prev;
+            if (_journal->getFileRecord(item->_file, &prev)
+                && prev.isValid()
+                && prev._remotePerm.hasPermission(RemotePermissions::CanWrite) != item->_remotePerm.hasPermission(RemotePermissions::CanWrite)) {
+                const bool isReadOnly = !item->_remotePerm.isNull() && !item->_remotePerm.hasPermission(RemotePermissions::CanWrite);
+                FileSystem::setFileReadOnlyWeak(filePath, isReadOnly);
             }
+
+            _journal->setFileRecordMetadata(item->toSyncJournalFileRecordWithInode(filePath));
```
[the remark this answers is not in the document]
ckamm: There used to be code like this:
```
                if (other && other->type != ItemTypeVirtualFile && other->type != ItemTypeVirtualFileDownload) {
                    // Even if the mtime is different on the server, we always want to keep the mtime from
                    // the file system in the DB, this is to avoid spurious upload on the next sync
                    item->_modtime = other->modtime;
                    // same for the size
                    item->_size = other->size;
                }
```
in `treewalkFile`. But there's a good chance the way you're populating item now takes care of it.

Yes, I think the mtime/size changes to virtual files belong here. But not in this PR.

src/libsync/syncengine.cpp
```diff
@@ -779,12 +406,18 @@ void SyncEngine::startSync()
...
+    _backInTimeFiles = 0;
+    _seenFiles.clear();
+
     _progressInfo->reset();
 
     if (!QDir(_localPath).exists()) {
         _anotherSyncNeeded = DelayedFollowUp;
         // No _tr, it should only occur in non-mirall
-        csyncError("Unable to find local sync folder.");
+        syncError("Unable to find local sync folder.");
```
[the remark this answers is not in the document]
ckamm: Oops, didn't see the comment.

src/libsync/discovery.cpp
```diff
@@ -0,0 +1,1218 @@
...
+        if (dbEntry.isValid() && localEntry.isVirtualFile) {
+            if (dbEntry._type == ItemTypeFile) {
+                // If we find what looks to be a spurious "abc.owncloud" the base file "abc"
+                // might have been renamed to that. Make sure that the base file is not
+                // deleted from the server.
+                if (dbEntry._modtime == localEntry.modtime && dbEntry._fileSize == localEntry.size) {
+                    qCInfo(lcDisco) << "Base file was renamed to virtual file:" << item->_file;
+                    Q_ASSERT(item->_file.endsWith(_discoveryData->_syncOptions._virtualFileSuffix));
+                    item->_direction = SyncFileItem::Down;
+                    item->_instruction = CSYNC_INSTRUCTION_NEW;
```
ckamm: Don't we also need to delete the local `abc.owncloud`? (could possibly use TYPE_CHANGE)

src/libsync/discovery.cpp
```diff
@@ -0,0 +1,1218 @@
...
+                    item->_type = ItemTypeVirtualFile;
+                }
+            } else if (item->_type != ItemTypeVirtualFileDownload) {
+                item->_type = ItemTypeVirtualFile;
```
ckamm: This looks redundant with the if block above where `serverModified = false; item->_type = ItemTypeVirtualFile;`

src/libsync/discovery.cpp
```diff
@@ -0,0 +1,1218 @@
...
+            } else if (!serverModified && dbEntry._inode != localEntry.inode) {
+                item->_instruction = CSYNC_INSTRUCTION_UPDATE_METADATA;
+                item->_direction = SyncFileItem::Down; // Does not matter
+            }
+        } else if (serverModified || dbEntry._type == ItemTypeVirtualFile) {
+            if (serverEntry.isDirectory && localEntry.isDirectory) {
+                // Folders of the same path are always considered equals
+                item->_instruction = CSYNC_INSTRUCTION_UPDATE_METADATA;
+            } else {
+                // It could be a conflict even if size and mtime match!
```
ckamm: Wait! If local/db size+mtime match, we won't be in this branch. We'll have gone into the above `else if`!

Possibly this comment is confusing here since the below code seems to assume that there was a local change.

src/libsync/discovery.cpp
```diff
@@ -0,0 +1,1218 @@
...
+                QByteArray remoteChecksumHeader = serverEntry.checksumHeader;
+                if (!remoteChecksumHeader.isEmpty()) {
+                    // Do we have an UploadInfo for this?
+                    // Maybe the Upload was completed, but the connection was broken just before
+                    // we recieved the etag (Issue #5106)
+                    auto up = _discoveryData->_statedb->getUploadInfo(path._original);
+                    if (up._valid && up._contentChecksum == remoteChecksumHeader) {
+                        // Solve the conflict into an upload, or nothing
+                        item->_instruction = up._modtime == localEntry.modtime && up._size == localEntry.size
+                            ? CSYNC_INSTRUCTION_UPDATE_METADATA : CSYNC_INSTRUCTION_SYNC;
```
ckamm: Using an UPDATE_METADATA here is confusing when the below comment says we can't use a typical UPDATE_METADATA and then updates the journal on the spot. Please explain why it isn't a NONE here?

src/libsync/discovery.cpp
```diff
@@ -0,0 +1,1218 @@
...
+                    item->_instruction = isConflict ? CSYNC_INSTRUCTION_CONFLICT : CSYNC_INSTRUCTION_UPDATE_METADATA;
+                    item->_direction = isConflict ? SyncFileItem::None : SyncFileItem::Down;
+                }
+            }
+            if (dbEntry._type == ItemTypeVirtualFile)
+                item->_type = ItemTypeVirtualFileDownload;
+            if (item->_file.endsWith(_discoveryData->_syncOptions._virtualFileSuffix)) {
+                item->_file.chop(_discoveryData->_syncOptions._virtualFileSuffix.size());
+                item->_type = ItemTypeVirtualFileDownload;
+            }
```
ckamm: These two ifs are intended to catch cases where the user replaces a virtual file with a real one and we produce a conflict? I'm confused.

src/libsync/discovery.cpp
```diff
@@ -0,0 +1,1218 @@
...
+                        (*it)->_instruction = CSYNC_INSTRUCTION_NONE;
+                    oldEtag = (*it)->_etag;
+                    if (!item->isDirectory() && oldEtag != base._etag) {
+                        isMove = false;
+                    }
+                }
+                if (auto deleteJob = _discoveryData->_queuedDeletedDirectories.value(originalPath)) {
+                    oldEtag = deleteJob->_dirItem->_etag;
+                    delete _discoveryData->_queuedDeletedDirectories.take(originalPath);
+                }
```
ckamm: There's a good chance we haven't visited originalPath yet, and I'm assuming the way it works is that when we get to it but have the _renamedItems entry, we won't set it up for REMOVE. If true, please add a comment to that effect here.

src/libsync/discovery.cpp
```diff
@@ -0,0 +1,1218 @@
...
+                            // Can't be a rename, leave it as a new.
+                            postProcessLocalNew();
+                        } else {
+                            // In case the deleted item was discovered in parallel
+                            auto it = _discoveryData->_deletedItem.find(originalPath);
+                            if (it != _discoveryData->_deletedItem.end()) {
+                                ASSERT((*it)->_instruction == CSYNC_INSTRUCTION_REMOVE);
+                                (*it)->_instruction = CSYNC_INSTRUCTION_NONE;
+                            }
+                            delete _discoveryData->_queuedDeletedDirectories.take(originalPath);
```
ckamm: similar check-async-check pattern; possibly make the "check" part a function again?

src/libsync/discovery.cpp
```diff
@@ -0,0 +1,1218 @@
...
+                // Special case! deleted on server, modified on client, the instruction is then NEW
+                item->_instruction = CSYNC_INSTRUCTION_NEW;
+            }
+            item->_direction = SyncFileItem::Up;
+            item->_checksumHeader.clear();
+            item->_size = localEntry.size;
+            item->_modtime = localEntry.modtime;
+            item->_previousSize = dbEntry._fileSize;
+            item->_previousModtime = dbEntry._modtime;
+            if (!_dirItem || _dirItem->_direction == SyncFileItem::Down) {
```
ckamm: There have been more of these along the way, but I still don't understand

src/libsync/discovery.cpp
```diff
@@ -0,0 +1,1218 @@
...
+                if (computeLocalChecksum(dbEntry._checksumHeader, _discoveryData->_localDir + path._local, item)
+                        && item->_checksumHeader == dbEntry._checksumHeader) {
+                    qCInfo(lcDisco) << "NOTE: Checksums are identical, file did not actually change: " << path._local;
+                    item->_instruction = CSYNC_INSTRUCTION_UPDATE_METADATA;
+                }
+            }
+        }
+    } else if (_queryLocal == ParentNotChanged && dbEntry.isValid()) {
+        if (noServerEntry) {
+            // Not modified locally (ParentNotChanged), bit not on the server:  Removed on the server.
```
ckamm: sp: but

src/libsync/discovery.cpp
```diff
@@ -0,0 +1,1218 @@
...
+            // Also restore everything
+            item->_instruction = CSYNC_INSTRUCTION_NEW;
+            item->_direction = SyncFileItem::Down;
+            item->_isRestoration = true;
+            item->_errorString = tr("Not allowed to remove, restoring");
+        } else if (!dbEntry._serverHasIgnoredFiles) {
+            item->_instruction = CSYNC_INSTRUCTION_REMOVE;
+            item->_direction = SyncFileItem::Up;
+        }
+    }
```
ckamm: Generally: I find these huge if-elseif-... chains very hard to follow. A lot of logic depends on what has previously been ruled out. Any ideas for making this more readable? Maybe asserting all the other important assumed conditions after the `if`?

Like, I've read this but:
- I'm not sure it's exhaustive / that all cases are covered.
- If there were mutually exclusive conditions I probably wouldn't have spotted them.
- Usually the branches assume a lot of context that's way up in this function.
- Maybe renames and conflict checking could be functions to limit the indent-depth and height of this thing?
- Overall feeling of fragility: due to many long-term assumptions and connections

Merging three trees in this way is genuinely hard and complex. Some of this will always be there.

src/libsync/discovery.cpp
```diff
@@ -0,0 +1,1218 @@
...
+                if (computeLocalChecksum(base._checksumHeader, _discoveryData->_localDir + path._original, item)) {
+                    qCInfo(lcDisco) << "checking checksum of potential rename " << path._original << item->_checksumHeader << base._checksumHeader;
+                    isMove = item->_checksumHeader == base._checksumHeader;
+                }
+            }
+            auto originalPath = QString::fromUtf8(base._path);
+            if (isMove && _discoveryData->_renamedItems.contains(originalPath))
+                isMove = false;
+
+            //Check local permission if we are allowed to put move the file here
```
ckamm: This move permissions check is a bit surprising at this point as other permissions are checked further down. I expect it has to be done right here though. Hmm, put into function at least?

src/libsync/discovery.cpp
```diff
@@ -0,0 +1,1218 @@
...
+        // Do not download virtual files
+        if (serverModified || dbEntry._type != ItemTypeVirtualFile)
+            item->_instruction = CSYNC_INSTRUCTION_UPDATE_METADATA;
+        serverModified = false;
+        item->_type = ItemTypeVirtualFile;
+    }
+    if (!_dirItem || _dirItem->_direction == SyncFileItem::Up) {
+        _childModified |= serverModified;
+    }
+    if (localEntry.isValid()) {
```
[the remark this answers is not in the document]
ogoffart: Any suggestions on how to do it.

Maybe the problem is the rename detection,  some could be factored out, that's true. But i'm not sure we gain something.

src/libsync/discovery.cpp
```diff
@@ -0,0 +1,1218 @@
...
+        // If the type of the entity changed, it's like NEW, but
+        // needs to delete the other entity first.
+        item->_instruction = CSYNC_INSTRUCTION_TYPE_CHANGE;
+        item->_direction = SyncFileItem::Down;
+        item->_modtime = serverEntry.modtime;
+        item->_size = serverEntry.size;
+    } else if (dbEntry._type == ItemTypeVirtualFileDownload) {
+        item->_direction = SyncFileItem::Down;
+        item->_instruction = CSYNC_INSTRUCTION_NEW;
+        item->_file = _currentFolder._target + QLatin1Char('/') + serverEntry.name;
```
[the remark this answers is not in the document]
ogoffart: Because path contains the ".owncloud"

If that's more clear i could use path, and remove the virtual file suffix.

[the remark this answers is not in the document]
ckamm: I assumed path._server never had the suffix, but maybe that assumption is mistaken.

src/libsync/discovery.cpp
```diff
@@ -0,0 +1,1218 @@
...
+        || item->_instruction == CSYNC_INSTRUCTION_RENAME || item->_instruction == CSYNC_INSTRUCTION_TYPE_CHANGE;
+
+    if ((dbEntry.isValid() && dbEntry._type == ItemTypeVirtualFile) || (localEntry.isValid() && localEntry.isVirtualFile && item->_type != ItemTypeVirtualFileDownload)) {
+        // Do not download virtual files
+        if (serverModified || dbEntry._type != ItemTypeVirtualFile)
+            item->_instruction = CSYNC_INSTRUCTION_UPDATE_METADATA;
+        serverModified = false;
+        item->_type = ItemTypeVirtualFile;
+    }
+    if (!_dirItem || _dirItem->_direction == SyncFileItem::Up) {
```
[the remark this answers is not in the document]
ogoffart: I did not know either :-). So I removed the condition to see what happens, and it turned out it is because of CSYNC_INSTRUCTION_TYPE_CHANGE.  In this case, we do not want to mark it as a conflict if the new directory has new files (which is to be expected)

I now removed the condition and checked it in another way.  Only when removing a directory I check if _childModified

Notice that while doing that I realized that we will delete the directory and proceed with the type change if the directory being removed contains ignored item...
Maybe this can be improved...

src/libsync/discovery.cpp
```diff
@@ -0,0 +1,1218 @@
...
+        }
+    } else if (_queryLocal == ParentNotChanged && dbEntry.isValid()) {
+        if (noServerEntry) {
+            // Not modified locally (ParentNotChanged), bit not on the server:  Removed on the server.
+            item->_instruction = CSYNC_INSTRUCTION_REMOVE;
+            item->_direction = SyncFileItem::Down;
+        }
+    } else if (noServerEntry) {
+        // Not locally, not on the server. The entry is stale!
+        qCInfo(lcDisco) << "Stale DB entry";
```
ckamm: I think we want to wipe the local db entry in this case (and add a remove/remove test if we don't have one yet)

src/libsync/discovery.cpp
ogoffart: That's taken care of in PropagateDownloadFile.   
I did not change the behavior with the old code, AFAIK.  
It is true that this could be improved.

src/libsync/syncengine.cpp
```diff
@@ -1064,36 +628,19 @@ void SyncEngine::slotDiscoveryJobFinished(int discoveryResult)
...
-        restoreOldFiles(syncItems);
-    } else if (!_hasForwardInTimeFiles && _backInTimeFiles >= 2
-        && _account->serverVersionInt() < Account::makeServerVersion(9, 1, 0)) {
-        // The server before ownCloud 9.1 did not have the data-fingerprint property. So in that
-        // case we use heuristics to detect restored backup.  This is disabled with newer version
-        // because this causes troubles to the user and is not as reliable as the data-fingerprint.
-        qCInfo(lcEngine) << "All the changes are bringing files in the past, asking the user";
-        // this typically happen when a backup is restored on the server
-        bool restore = false;
-        emit aboutToRestoreBackup(&restore);
```
ckamm: `aboutToRestoreBackup()` is never emitted in the new code, and `_hasForwardInTimeFiles/_backInTimeFiles` are set, but unused.

src/libsync/discovery.cpp
ogoffart: Right, this comment is copy-pasted from the reconcile code, but some code was moved around and is no longer in the same order.

src/libsync/discovery.cpp
```diff
@@ -0,0 +1,1218 @@
...
+    }
+
+    int started = 0;
+    foreach (auto *rj, _runningJobs) {
+        started += rj->processSubJobs(nbJobs - started);
+        if (started >= nbJobs)
+            return started;
+    }
+
+    while (started < nbJobs && !_queuedJobs.empty()) {
```
ckamm: Not every start()ed job will increase `_currentlyActiveJobs` by one. Maybe this should loop until the desired number for that variable is reached?

src/libsync/discovery.cpp
ckamm: Hm, I don't see it. Where? I see the remove for VirtualFileDownload, but not for this direction.

src/libsync/discovery.cpp
ckamm: I don't know. If the test is very thorough then yes, but probably we don't check the journal state in detail. An extra UPDATE_METADATA shouldn't hurt, it just confused me while reading. Since it was that way before I'd leave it as-is for now, double check later.

src/libsync/discovery.cpp
ogoffart: exactly.
You create a file with the same name as an existing file, this will cause the file to be made as a conflict and the actual file to be downloaded.
This is the same behavior as before and is tested by `testVirtualFileConflict`

src/libsync/discovery.cpp
```diff
@@ -0,0 +1,1218 @@
...
+                                           << "but looks odd, keeping";
+                        item->_instruction = CSYNC_INSTRUCTION_IGNORE;
+                    }
+                }
+            };
+
+            // Check if it is a move
+            OCC::SyncJournalFileRecord base;
+            if (!_discoveryData->_statedb->getFileRecordByInode(localEntry.inode, &base)) {
+                dbError();
```
ckamm: This emits `finished()`, but `scheduleSubJobs()` will still be called via a timer from `process()` leading to a double invocation of `finished()`. Something similar happens for the other rename detection.

src/libsync/discovery.cpp
ogoffart: We need to check that the permission are correct before calling findAndCancelDeletedJob, as this has side effect

This allow to convert a move into a remove + download immediately.

src/libsync/discovery.cpp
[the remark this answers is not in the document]
ckamm: I'm sceptical as there's lots of code like this - where _childModified gets changed after some Up/Down check. I've just never dug into what's going on exactly. Guarding against deleting stuff with ignores or other changes on the opposide side makes a lot of sense.

src/libsync/syncengine.cpp
ogoffart: True, this was done only if _account->serverVersionInt() < Account::makeServerVersion(9, 1, 0) ,  and this server is no longer supported.
So I'll remove the remains of it.

src/libsync/discovery.cpp
ogoffart: I think it's fine as we don't want to have too many jobs in parallel.
The jobs that don't increase _currentlyActiveJobs will anyway call this function again.

src/libsync/discovery.cpp
ogoffart: well spotted.

dbError is not tested at all :-(

Maybe it make sense to make a test for it...  (where we remove the db or cause it to error somehow)

src/libsync/discovery.cpp
```diff
@@ -0,0 +1,1218 @@
...
+        f->start();
+        started++;
+    }
+    return started;
+}
+
+void ProcessDirectoryJob::dbError()
+{
+    _discoveryData->fatalError(tr("Error while reading the database"));
+    emit finished();
```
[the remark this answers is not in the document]
ckamm: Like this processSubJobs will still get called and potentially start and process the sub jobs. Should it immediately abort if finished was already emitted? (preference for isFinished bool over reusing _pendingAsyncJobs in that case)

ckamm:
Overall, I think this is great. There will very likely be bugs in places that are not sufficiently covered by tests yet. But I think it's worth it.

ogoffart:
@ckamm  So do you think this is ready to be merged?

ckamm:
@ogoffart No, there are a bunch of unresolved comments. I'd like for everything that's open to be addressed (many outdated things are still open). I'll spend some time doing so right now.

src/libsync/discovery.cpp
```diff
@@ -0,0 +1,1218 @@
...
+        }
+        entriesNames.insert(name);
+        localEntriesHash[name] = std::move(e);
+    }
+    _localEntries.clear();
+
+    if (_queryServer == ParentNotChanged || _queryLocal == ParentNotChanged) {
+        // fetch all the name from the DB
+        auto pathU8 = _currentFolder._original.toUtf8();
+        // FIXME do that better (a query that do not get stuff recursively ?)
```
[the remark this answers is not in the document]
ckamm: @ogoffart Do you want to take over this part?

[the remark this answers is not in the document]
ogoffart: In order to do that, we will need to change the layout of the database.
We would need a `parent` column with a hash of the parent.

I thought to delay this change a bit. But it indeed need to be done for the release, so i guess i'll start working on this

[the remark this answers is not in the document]
ogoffart: I found a clever way which does not even need to add a column in the db.

ckamm:
The custom index is a pretty nice solution! :+1:

ckamm:
Since it's a pain to navigate: I think those are the last four unresolved conversations:
https://github.com/owncloud/client/pull/6738#discussion_r225180576
https://github.com/owncloud/client/pull/6738#discussion_r225465386
https://github.com/owncloud/client/pull/6738#discussion_r225451143
https://github.com/owncloud/client/pull/6738#discussion_r225466052

src/libsync/discovery.cpp
ogoffart: The logic here is unchanged from `_csync_detect_update`  (that's even your code :-)).
IMHO, indeed, none of the test above should have `done = true`  Should I change that now?

src/libsync/discovery.cpp
ogoffart: this condition no longer exists

src/libsync/discovery.cpp
```diff
@@ -1219,6 +1219,9 @@ DiscoverySingleDirectoryJob *ProcessDirectoryJob::startAsyncServerQuery()
 
 bool ProcessDirectoryJob::runLocalQuery()
 {
+    QByteArray localPath = (_discoveryData->_localDir + _currentFolder._local).toUtf8();
```
ckamm: @ogoffart Where's this new path used?

ckamm: Happens. 

I think it would look somewhat nicer if all the logging and error strings further down also used `localPath` over the longer string conat. But that's optional.

Issue #6726, opened 2018-08-16
[VFS] No option to replace downloaded files with placeholder files


michaelstingl:
Currently there is no option for the user to replace downloaded files with placeholder files to free-up disk space. (tested with todays daily 2.5.0 build)

![owncloud-3](https://user-images.githubusercontent.com/214010/44232136-7257dc00-a1a0-11e8-83b2-97bb4afe72a6.png)

### Expected behaviour
There should be an option (in the context menu) to replace downloaded files with placeholder files

### Actual behaviour
Currently there is no option replace individual downloaded files, but there is a workaround to replace all content of the parent folder with placeholder files.

### Workaround
1. Open settings and deselect the parent folder in the selective sync UI
2. Apply this change
3. Wait for the files to be deleted
4. Then select the parent folder in the selective sync UI
5. Apply this change
6. ==> Placeholder file gets downloaded from ownCloud server

### Here is the example from another tool

![db-ss](https://user-images.githubusercontent.com/214010/44232684-14c48f00-a1a2-11e8-8798-338160066853.png)

If multiple elements have the same state, a checkmark is shown:

![db-ss-checkmark](https://user-images.githubusercontent.com/214010/44232854-9ddbc600-a1a2-11e8-8dd4-643c1ffbefff.png)

ogoffart:
@ckamm:  here is how i'd do it:

I'd add a new option in the context menu (socket api) `tr("Replace by virtual files")`, then i see several options: 

1) Simply delete the files we want to replace, and at the same time, delete their entry in the DB.  Next sync will then restore the placeholder.

2) For each file we want to replace, just rename it by adding the ".owncloud" extension, and let the logic from #6721 make the virtual file in the next sync.

3) For each file we want to replace, delete it, replace it by a placeholder, and update the database as wall.

What approach should we take?
Thinking about it, I think I'd do 2. It is the easier. But it relies on the hack done for #6721, I guess it is fine.
3 Would be the most correct but also is a bit more work.

ckamm:
@ogoffart I'm ok with using 2. if there's a warning note about the interaction on the code that performs the action. More advanced vfs will need to do more complex actions like 3.

michaelstingl:
LGTM

![cursor_and_msedge_-_win10](https://user-images.githubusercontent.com/214010/44998118-6008db00-afb3-11e8-934d-5b6a02ede99c.png)

- real file(s) selected ==> `Replace file(s) by virtual file`
[...]

PR #6743, opened 2018-08-28
Socket API: add an option to replace existing files with virtual files

ogoffart:
Issue #6726

### DIFF
--- a/src/gui/socketapi.cpp
+++ b/src/gui/socketapi.cpp
@@ -621,6 +621,41 @@ void SocketApi::command_DOWNLOAD_VIRTUAL_FILE(const QString &filesArg, SocketLis
     }
 }
 
+/* Go over all the files ans replace them by a virtual file */
+void SocketApi::command_REPLACE_VIRTUAL_FILE(const QString &filesArg, SocketListener *)
+{
+    QStringList files = filesArg.split(QLatin1Char('\x1e')); // Record Separator
+    auto suffix = QStringLiteral(APPLICATION_DOTVIRTUALFILE_SUFFIX);
+
+    for (const auto &file : files) {
+        QString relativePath;
+        auto folder = FolderMan::instance()->folderForPath(file, &relativePath);
+        if (!folder)
+            continue;
+        if (file.endsWith(suffix))
+            continue;
+        QFileInfo fi(file);
+        if (fi.isDir()) {
+            folder->journalDb()->getFilesBelowPath(relativePath.toUtf8(), [&](const SyncJournalFileRecord &rec) {
+                if (rec._type != ItemTypeFile || rec._path.endsWith(APPLICATION_DOTVIRTUALFILE_SUFFIX))
+                    return;
+                QString file = folder->path() + '/' + QString::fromUtf8(rec._path);
+                if (!FileSystem::rename(file, file + suffix)) {
+                    qCWarning(lcSocketApi) << "Unable to rename " << file;
+                }
+            });
+            continue;
+        }
+        SyncJournalFileRecord record;
+        if (!folder->journalDb()->getFileRecord(relativePath, &record) || !record.isValid())
+            continue;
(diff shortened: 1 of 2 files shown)


src/gui/socketapi.cpp
```diff
@@ -763,12 +798,24 @@ void SocketApi::command_GET_MENU_ITEMS(const QString &argument, OCC::SocketListe
...
+        bool hasNormalFiles = false;
+        bool hasDir = false;
         for (const auto &file : files) {
-            if (file.endsWith(virtualFileSuffix) || (folder->useVirtualFiles() && QFileInfo(file).isDir()))
+            if (QFileInfo(file).isDir()) {
+                hasDir = true;
+            } else if (file.endsWith(virtualFileSuffix)) {
                 hasVirtualFile = true;
+            } else if (!hasNormalFiles) {
+                FileData fileData = FileData::get(files.first());
```
ckamm: `get(file)`

src/gui/socketapi.cpp
```diff
@@ -763,12 +798,24 @@ void SocketApi::command_GET_MENU_ITEMS(const QString &argument, OCC::SocketListe
...
+                bool isOnTheServer = fileData.journalRecord().isValid();
+                hasNormalFiles = isOnTheServer;
+            }
         }
-        if (hasVirtualFile)
+        if (hasVirtualFile || (hasDir && folder->useVirtualFiles()))
             listener->sendMessage(QLatin1String("MENU_ITEM:DOWNLOAD_VIRTUAL_FILE::") + tr("Download file(s)", "", files.size()));
+
+        if ((hasNormalFiles || hasDir) && folder->useVirtualFiles())
+            listener->sendMessage(QLatin1String("MENU_ITEM:REPLACE_VIRTUAL_FILE::") + tr("Replace file(s) by virtual file", "", files.size()));
```
ckamm: The wording isn't as immediately clear as "Download file(s)". 

"Remove local copy of file contents"
"Free space by dropping file contents"
"Store file(s) on the server only"

Unsure.

src/gui/socketapi.cpp
```diff
@@ -621,6 +621,41 @@ void SocketApi::command_DOWNLOAD_VIRTUAL_FILE(const QString &filesArg, SocketLis
...
+            continue;
+        if (file.endsWith(suffix))
+            continue;
+        QFileInfo fi(file);
+        if (fi.isDir()) {
+            folder->journalDb()->getFilesBelowPath(relativePath.toUtf8(), [&](const SyncJournalFileRecord &rec) {
+                if (rec._type != ItemTypeFile || rec._path.endsWith(APPLICATION_DOTVIRTUALFILE_SUFFIX))
+                    return;
+                QString file = folder->path() + '/' + QString::fromUtf8(rec._path);
+                if (!QFile::rename(file, file + suffix)) {
```
ckamm: use `FileSystem::rename()` otherwise you run into problems with .lnk files.

src/gui/socketapi.cpp
```diff
@@ -621,6 +621,41 @@ void SocketApi::command_DOWNLOAD_VIRTUAL_FILE(const QString &filesArg, SocketLis
...
+            });
+            continue;
+        }
+        SyncJournalFileRecord record;
+        if (!folder->journalDb()->getFileRecord(relativePath, &record) || !record.isValid())
+            continue;
+        if (!QFile::rename(file, file + suffix)) {
+            qCWarning(lcSocketApi) << "Unable to rename " << file;
+        }
+        FolderMan::instance()->scheduleFolder(folder);
```
ckamm: Probably should not happen inside the loop

ogoffart: yeah...  but this will be queued annyway.  And folder might be different (altough it shouldn't)

src/gui/socketapi.cpp
ogoffart: @michaelstingl   : What  should the wording be?

Issue #6725, opened 2018-08-16
[VFS] Migrating existing sync folder(s) to VFS


michaelstingl:
### Expected behaviour
It would be awesome to migrate pre-2.5 accounts to the new VFS behaviour

### Actual behaviour
Currently it's not possible
(tested with todays daily 2.5.0 build)

ogoffart:
Sync-engine wise, this is trivial to do. We just need to find a way to do it ui-wise.
Add an entry in the context menu with  "[] New files are created as Virtual Files" ?
Can one disable the option and then all the files are downloaded?  Does it need a message box to confirm?

michaelstingl:
Cool, good to know! 👍

tessus:
@ogoffart @michaelstingl is this supposed to be done automatically? It's just that I do not plan to use VFS and I don't want the client to automatically convert my setup to VFS.

According to the commit log, it's not a behind-the-curtain conversion, but I just wanted to make sure.

tessus:
@michaelstingl thanks for the reply. I only wanted to say that I do not want the client to convert non-VFS setups to VFS automatically - ever. If people choose not to use VFS, it should be respected.

PR #6744, opened 2018-08-28
Account Settings: Add a context menu entry to enable or disable virtual files

ogoffart:
Issue #6725

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -335,6 +335,16 @@ void AccountSettings::slotCustomContextMenuRequested(const QPoint &pos)
             OwncloudWizard::askExperimentalVirtualFilesFeature([folder, this](bool enable) {
                 if (enable && folder)
                     folder->setUseVirtualFiles(enable);
+
+                // Also wipe selective sync settings
+                bool ok = false;
+                auto oldBlacklist = folder->journalDb()->getSelectiveSyncList(SyncJournalDb::SelectiveSyncBlackList, &ok);
+                folder->journalDb()->setSelectiveSyncList(SyncJournalDb::SelectiveSyncBlackList, {});
+                for (const auto &entry : oldBlacklist) {
+                    folder->journalDb()->avoidReadFromDbOnNextSync(entry);
+                }
+                FolderMan::instance()->scheduleFolder(folder);
+
                 // Make sure the size is recomputed as the virtual file indicator changes
                 ui->_folderList->doItemsLayout();
             });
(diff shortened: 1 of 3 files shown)


ogoffart:
This adds a checkbox in the folder context menu.
The option is only visible in experimental mode.
When enabling it (but not disabling it), it will show the popup which we already show in the wizard.

@michaelstingl 

Screenshot of the menu:

screenshot_20180828_111425

When clicking on the last item, this dialog appear:

screenshot_20180828_111558

src/gui/folder.h
```diff
@@ -244,6 +244,7 @@ class Folder : public QObject
 
     /** new files are downloaded as virtual files */
     bool useVirtualFiles() { return _definition.useVirtualFiles; }
+    void setUseVirtualFiles(bool enabled) { _definition.useVirtualFiles = enabled; }
```
ckamm: Immediate `saveToSettings()` ?

ogoffart: Ooops, i forgot that.

src/gui/accountsettings.cpp
```diff
@@ -315,6 +319,29 @@ void AccountSettings::slotCustomContextMenuRequested(const QPoint &pos)
 
     ac = menu->addAction(tr("Remove folder sync connection"));
     connect(ac, &QAction::triggered, this, &AccountSettings::slotRemoveCurrentFolder);
+
+    if (ConfigFile().showExperimentalOptions() || folder->useVirtualFiles()) {
```
ckamm: probably remove `showExperimentalOptions()`

ckamm: should become a branding option though

ckamm: see #6752

ckamm:
now depends on https://github.com/owncloud/client/pull/6752

ckamm:
@ogoffart merge if it looks good to you

ogoffart:
@michaelstingl  should we merge this into 2.5.0?
Experimental settings always visible in the wizard is one thing, but always visible in the context menu is a different thing.

michaelstingl:
@ogoffart I also don't feel confident having this in 2.5.0.

ogoffart:
@michaelstingl  Do we want this patch in 2.5.1 ?  Or should we wait even more?

guruz:
I think we should have this in 2.5.1.
@ogoffart Do you need to re-base since https://github.com/owncloud/client/pull/6752 is merged? (github claims it can still merge)

michaelstingl:
okay, let's try this 👍

PR #6746, opened 2018-08-28
Fix doc error regarding adding a new Apt repo

settermjd:
This PR fixes an error in the existing documentation, which has an error in the instructions for adding a new ownCloud Debian 9 repo. If followed, Apt becomes unusable as it will display the error below, whenever you attempt to use it, and exit.

```console
E: Malformed line 1 in source list /etc/apt/sources.list.d/owncloud-client.list (type)
E: The list of sources could not be read.
E: Malformed line 1 in source list /etc/apt/sources.list.d/owncloud-client.list (type)
E: The list of sources could not be read.
```

### DIFF
--- a/doc/building.rst
+++ b/doc/building.rst
@@ -50,15 +50,14 @@ repositories`_ to see all the Linux client repos.
    repo for your Debian or Ubuntu version, like this example for Debian 9 
    (run as root)::
  
-    echo 'deb-src 
-    http://download.opensuse.org/repositories/isv:/ownCloud:/desktop/Debian_9.0/ /' >> /etc/apt/sources.list.d/owncloud-client.list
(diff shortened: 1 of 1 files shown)


settermjd:
@jnweiger, would you mind casting your eye over the changes?

ogoffart:
Should this not be rebased on 2.5?

ogoffart:
I agree with dschmidt: `-y` should not be present in the default documentation. This can lead to bad surprise to the user as it sees things being installed that he might not agree with.

Also, if this was meant to be in 2.5, it should have been done in the 2.5 branch. This cannot be backported because of conflicts.

ogoffart:
Closing outdated PR

PR #6748, opened 2018-08-28
Add the ownCloud documentation theme as a submodule

settermjd:
Without this, the HTML documentation isn't able to build successfully, as it fails on the deployment of the generated HTML files to the live server.


michaelstingl:
@settermjd [2.5](https://github.com/owncloud/client/tree/2.5) too?

tboerger:
> @settermjd 2.5 too?

2.5 should already have that, just the 2.4 branch failed to build the docs.

michaelstingl:
yes, looks good:
https://github.com/owncloud/client/blob/2.5/.gitmodules

PR #6752, opened 2018-08-30
virtual files: introduce branding option for showing ui option

ckamm:


ckamm:
Actually I shouldn't have removed it, it should have become a branding option

ckamm:
How to make the testcloud client show the option too?

guruz:
> How to make the testcloud client show the option too?

That's in an internal repo, I can do that part..

src/libsync/theme.cpp
```diff
@@ -555,4 +555,9 @@ QString Theme::versionSwitchOutput() const
     return helpText;
 }
 
+bool Theme::showVirtualFilesOption() const
+{
+    return false;
```
ogoffart: Maybe  `return ConfigFile().showExperimentalOptions();`  ?

ckamm: good point, done. makes Theme depend on ConfigFile though

Issue #6724, opened 2018-08-16
[VFS] Is selective sync useful for accounts created in VFS mode ??


michaelstingl:
What is the use case for the selective sync UI in an account that has been created  in VFS mode?

Or is it more user friendly to let users interact only in the file manager?

![owncloud](https://user-images.githubusercontent.com/214010/44227727-a8db2a00-a193-11e8-8c8b-57e5cb15ad39.png)

/cc @pmaier1 @hodyroff @felixboehm

guruz:
>  in an account that has been created in VFS mode?

[It's per folder not per account](https://github.com/owncloud/client/issues/6723#issuecomment-413809435)...

FYI now it looks like
<img width="796" alt="screen shot 2018-08-20 at 14 30 39" src="https://user-images.githubusercontent.com/959327/44340620-b0494e80-a485-11e8-8c2c-c2ed4c2d7bb3.png">

We could hide all subtree items.

michaelstingl:
> We could hide all subtree items.

As of today, I still need it as a workaround to replace "real" files with placeholder files (https://github.com/owncloud/client/issues/6726).

ogoffart:
What we probably should get rid of, is the size check for new folder.

michaelstingl:
LGTM

![msedge_-_win10](https://user-images.githubusercontent.com/214010/44998041-a90c5f80-afb2-11e8-895f-7d498fce9c3c.png)

PR #6753, opened 2018-08-30
Virtual files: Don't show selective sync

ogoffart:
Issue #6724


src/gui/accountsettings.cpp
```diff
@@ -287,6 +287,9 @@ void AccountSettings::slotCustomContextMenuRequested(const QPoint &pos)
     bool folderPaused = _model->data(index, FolderStatusDelegate::FolderSyncPaused).toBool();
     bool folderConnected = _model->data(index, FolderStatusDelegate::FolderAccountConnected).toBool();
     auto folderMan = FolderMan::instance();
+    QPointer<Folder> folder = folderMan->folder(alias);
```
ckamm: minor: why the QPointer?

ogoffart: Because I need it for #6744  .  And there we need QPointer because there will be a message box showing and it might get deleted while the message box is shown.

PR #6754, opened 2018-08-30
Virtual files: don't check for new big folders

ogoffart:
https://github.com/owncloud/client/issues/6724#issuecomment-417368475

### DIFF
--- a/src/libsync/discoveryphase.cpp
+++ b/src/libsync/discoveryphase.cpp
@@ -90,7 +90,7 @@ int DiscoveryJob::isInSelectiveSyncBlackListCallback(void *data, const QByteArra
 
 bool DiscoveryJob::checkSelectiveSyncNewFolder(const QString &path, RemotePermissions remotePerm)
 {
-    if (_syncOptions._confirmExternalStorage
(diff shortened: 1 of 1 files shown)


michaelstingl:
@ogoffart same for mounted storages?

ogoffart:
> same for mounted storages?

Yes.

Since the "choose what to sync" would be hidden (https://github.com/owncloud/client/pull/6753 ) , these feature makes no sense as there are no way to select files.

src/libsync/discoveryphase.cpp
```diff
@@ -90,7 +90,7 @@ int DiscoveryJob::isInSelectiveSyncBlackListCallback(void *data, const QByteArra
 
 bool DiscoveryJob::checkSelectiveSyncNewFolder(const QString &path, RemotePermissions remotePerm)
 {
-    if (_syncOptions._confirmExternalStorage
+    if (_syncOptions._confirmExternalStorage && !_syncOptions._moveFilesToTrash
```
ckamm: `_moveFilesToTrash`? I don't get it

ogoffart: That obviously was meant to be _newFilesAreVirtual
I guess i selected the wrong entry in auto-complete by mistake.

Issue #6757, opened 2018-08-31
Force sync should ignore or wipe the blacklist


michaelstingl:
### Expected behaviour
Force sync should try again to sync all failed files and folders that didn't sync during previous errors

### Actual behaviour
Files and folder with sync errors get blacklisted for up to 24h. Force sync triggers a new sync run, but blacklisted files are handled like in a regular sync run

@ckamm @ogoffart also relevant for F6 ?

PR #6760, opened 2018-09-03
AccountSettings: Force sync should wipe the blacklist

ogoffart:
Issue #6757


guruz:
> The other option would have been to set the blacklisted items to retry: then they would have been attempted, but if they still failed, the errors would have been muted. I 

This is how I had understood the discussion too and have expected that this was the goal.

Is this for 2.5.0 or 2.5.1?

michaelstingl:
2.5.1 would be enough (can't judge the impact)

PR #6761, opened 2018-09-03
WIP: Read default folder configuration from a file on the server

ogoffart:
This commit will read a file called `.desktopclientconfig` at the root
of the webdav, while running the wizard, before showing the last page.

If the file exist and contains a list of default folder to configure,
hide all the option to configure the remote folders, as the default
configurtation will be applied.

screenshot_20180903_112503

The config file should contains a json object. The currently supported
config can be seen in this these example:

```
{ "default_folders": [
    {
        "remote_folder" : "Documents",
        "black_list" :  [ "Archives/1999", "Archives/1998" ]
    },
    {  "remote_folder": "Music"  }
]}
```

The `default_folders` property is a list of folder configuration. The
`remote_folder` property is the path of the folder on the remote. The
`black_list` property is a list of entries for the "choose what to sync".

### DIFF
--- a/src/gui/owncloudsetupwizard.cpp
+++ b/src/gui/owncloudsetupwizard.cpp
@@ -353,7 +353,35 @@ void OwncloudSetupWizard::testOwnCloudConnect()
     // so don't automatically follow redirects.
     job->setFollowRedirects(false);
     job->setProperties(QList<QByteArray>() << "getlastmodified");
-    connect(job, &PropfindJob::result, _ocWizard, &OwncloudWizard::successfulStep);
+    connect(job, &PropfindJob::result, this, [this, account] {
+        auto *job2 = account->sendRequest("GET", Utility::concatUrlPath(account->davUrl(), ".desktopclientconfig"));
+        job2->setIgnoreCredentialFailure(true);
+        QObject::connect(job2, &SimpleNetworkJob::finishedSignal, this, [this, account] (QNetworkReply *reply) {
+            if (reply->error()) {
+                if (reply->error() == QNetworkReply::ContentNotFoundError) {
+                    // Ignore a 404 error, the file can be missing (default case)
+                    _ocWizard->successfulStep();
+                } else {
+                    // Note: slotAuthError uses sender() which should be job2
+                    slotAuthError();
+                }
+                return;
+            }
+            auto jsonData = reply->readAll();
+            QJsonParseError jsonParseError;
+            QJsonDocument json = QJsonDocument::fromJson(jsonData, &jsonParseError);
+            if (jsonParseError.error != QJsonParseError::NoError) {
+                // Invalid JSON
+                qCWarning(lcWizard) << "Error while parsing .desktopclientconfig" << jsonParseError.errorString() << jsonData;
+                // Silently ignore the error (consider the file was not there)
+            } else {
+                // TODO: make it a real parameter
+                account->setProperty("oc_serverConfig", json);
+            }
+            _ocWizard->successfulStep();
+        });
+    });
+
     connect(job, &PropfindJob::finishedWithError, this, &OwncloudSetupWizard::slotAuthError);
     job->start();
 }
@@ -362,7 +390,7 @@ void OwncloudSetupWizard::slotAuthError()
 {
     QString errorMsg;
 
-    PropfindJob *job = qobject_cast<PropfindJob *>(sender());
+    AbstractNetworkJob *job = qobject_cast<AbstractNetworkJob *>(sender());
     if (!job) {
         qCWarning(lcWizard) << "Can't check for authed redirects. This slot should be invoked from PropfindJob!";
         return;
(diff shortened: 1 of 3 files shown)


ogoffart:
@michaelstingl Before continuing, it would be nice to know the use case of what should be configured.

ChrisEdS:
Is it also possible to set a default local folder with this `.desktopclientconfig` file?

michaelstingl:
@ChrisEdS We'll collect the requirements in a different issue…

ogoffart:
@michaelstingl  Can you please create an issue tracking what needs to be configured?

src/gui/owncloudsetupwizard.cpp
```diff
@@ -353,7 +353,35 @@ void OwncloudSetupWizard::testOwnCloudConnect()
     // so don't automatically follow redirects.
     job->setFollowRedirects(false);
     job->setProperties(QList<QByteArray>() << "getlastmodified");
-    connect(job, &PropfindJob::result, _ocWizard, &OwncloudWizard::successfulStep);
+    connect(job, &PropfindJob::result, this, [this, account] {
+        auto *job2 = account->sendRequest("GET", Utility::concatUrlPath(account->davUrl(), ".desktopclientconfig"));
```
ckamm: I would prefer if this didn't store the configuration in a user-visible / editable file. I understand getting support on the server may be hard/limiting, and that this may be a reasonable tradeoff, but:
- users will change / delete this file
- users will think this contains their current desktop client configuration when it doesn't

src/gui/owncloudsetupwizard.cpp
```diff
@@ -353,7 +353,35 @@ void OwncloudSetupWizard::testOwnCloudConnect()
     // so don't automatically follow redirects.
     job->setFollowRedirects(false);
     job->setProperties(QList<QByteArray>() << "getlastmodified");
-    connect(job, &PropfindJob::result, _ocWizard, &OwncloudWizard::successfulStep);
+    connect(job, &PropfindJob::result, this, [this, account] {
```
ckamm: I would prefer if the code for triggering and handling the request checking the server here was a separate function. At least it must have a comment explaining what's going on.

src/gui/owncloudsetupwizard.cpp
```diff
@@ -362,7 +390,7 @@ void OwncloudSetupWizard::slotAuthError()
 {
     QString errorMsg;
 
-    PropfindJob *job = qobject_cast<PropfindJob *>(sender());
+    AbstractNetworkJob *job = qobject_cast<AbstractNetworkJob *>(sender());
     if (!job) {
         qCWarning(lcWizard) << "Can't check for authed redirects. This slot should be invoked from PropfindJob!";
```
ckamm: Also update this warning

src/gui/owncloudsetupwizard.cpp
```diff
@@ -610,32 +638,62 @@ void OwncloudSetupWizard::slotAssistantFinished(int result)
         FolderMan *folderMan = FolderMan::instance();
         auto account = applyAccountChanges();
 
+        QJsonDocument configJson = account->account()->property("oc_serverConfig").toJsonDocument();
+        bool hasServerConfig = configJson.isObject() && !configJson.object()["default_folders"].toArray().isEmpty();
+        qWarning() << configJson << hasServerConfig;
```
ckamm: qCWarning()

src/gui/owncloudsetupwizard.cpp
```diff
@@ -610,32 +638,62 @@ void OwncloudSetupWizard::slotAssistantFinished(int result)
...
-                if (folderDefinition.virtualFilesMode != Vfs::Off && _ocWizard->useVirtualFileSync())
-                    f->setNewFilesAreVirtual(true);
-
-                f->journalDb()->setSelectiveSyncList(SyncJournalDb::SelectiveSyncBlackList,
-                    _ocWizard->selectiveSyncBlacklist());
-                if (!_ocWizard->isConfirmBigFolderChecked()) {
-                    // The user already accepted the selective sync dialog. everything is in the white list
-                    f->journalDb()->setSelectiveSyncList(SyncJournalDb::SelectiveSyncWhiteList,
-                        QStringList() << QLatin1String("/"));
+            if (!hasServerConfig) {
```
ckamm: This is getting very deeply nested. Split the two branches into functions?

src/gui/owncloudsetupwizard.cpp
```diff
@@ -610,32 +638,62 @@ void OwncloudSetupWizard::slotAssistantFinished(int result)
...
+                    QDir(localFolder).mkpath(folderDefinition.localPath);
+                    folderDefinition.targetPath = FolderDefinition::prepareTargetPath(remoteFolder);
+                    folderDefinition.ignoreHiddenFiles = folderMan->ignoreHiddenFiles();
+                    if (folderMan->navigationPaneHelper().showInExplorerNavigationPane())
+                        folderDefinition.navigationPaneClsid = QUuid::createUuid();
+
+                    auto f = folderMan->addFolder(account, folderDefinition);
+                    if (f) {
+                        if (folderDefinition.virtualFilesMode != Vfs::Off && _ocWizard->useVirtualFileSync())
+                            f->setNewFilesAreVirtual(true);
```
ckamm: Can be removed I think. folderDefinition.virtualFilesMode is never set and useVirtualFileSync() an't be true.

src/gui/wizard/owncloudadvancedsetuppage.cpp
```diff
@@ -139,11 +125,31 @@ void OwncloudAdvancedSetupPage::initializePage()
...
+    Theme *theme = Theme::instance();
 
-    if (Theme::instance()->wizardSelectiveSyncDefaultNothing()) {
+    _ui.rDefaultFolders->setVisible(hasServerConfig);
+    _ui.rSyncEverything->setVisible(!hasServerConfig);
+    _ui.lSyncEverythingSizeLabel->setVisible(!hasServerConfig);
+    _ui.rSelectiveSync->setVisible(!hasServerConfig);
+    _ui.bSelectiveSync->setVisible(!hasServerConfig);
+    _ui.lSelectiveSyncSizeLabel->setVisible(!hasServerConfig);
+    _ui.rVirtualFileSync->setVisible(!hasServerConfig && Theme::instance()->showVirtualFilesOption() && bestAvailableVfsMode() != Vfs::Off);
```
ckamm: As far as I remember removing the layout from the wSyncStrategy widget was required for making the layout not be broken on some platforms. May need testing.

PR #6785, opened 2018-09-27
Migrate the Documentation to Antora

settermjd:


dschmidt:
Is the `doc` cmake target simply gone now or do i simply don't see newly added CMake code?

tboerger:
The docs are getting imported into the main docs and built there. The client docs don't need to be built standalone anymore with antora.

tboerger:
Looks like it got to be dropped from the root cmake file?

dschmidt:
@tboerger I'm pretty sure @dragotin will contradict. Docs are packaged with distribution packages.

dragotin:
Yes, the client docs should be remain in the client source so that it can go into the rpm/deb. What is wrong with the current way of documenting?

dragotin:
Any *reason* for that? How about putting energy into the contents, rather than the framework?

tboerger:
That's not my decision, even if I welcome this decision because the setup up Antora is much simpler than this whole sphinx stack. Besides that I personally also prefer ascidoc over rst.

michaelstingl:
Can't distros not use the docs built from adoc? @dschmidt @dragotin what format is needed?

tboerger:
I guess we need some PDF shipped with the client packages. And with the latest tooling of @settermjd we should be able to build a PDF "book".

dragotin:
In openSUSE there is no antora yet. That means I would have trouble building the docs, other distros probably similar. Sphinx is a well know and wide spread technology which everybody has.

Shipping a PDF: That is the second best option. I am still dreaming of an openSUSE branded ownCloud client, and that would include the docs. Hardly possible with PDF.

I am still trying to understand why you wanna do the change, what is the real *benefit* of antora?

dschmidt:
@michaelstingl @tboerger @settermjd I don't want to judge over the choice of Antora. If it's good it will find it's way into distributions. I'm more worried that the "command line API" to generate them has changed and packagers now need to learn how to use Antora (while they didnt need to know sphinx, just cmake)
In conclusion: please readd the cmake target with a sane default invocation so make doc still works. 
And possibly document in the readme how to find more info on the docs (Maybe just a link to https://github.com/owncloud/docs/blob/master/README.md)

P.S. asciidoc is pretty readable without generating html or pdf from it. ... y not html?

settermjd:
@dschmidt, thanks for [your most recent feedback](https://github.com/owncloud/client/pull/6785#issuecomment-425364427). It could be possible (likely is) to have the client docs both included in the rest of the documentation, as well as be able to be built standalone. I have to check the Antora docs again. If so, then I'll get a CMake target back in to build them, using the target name. 

In the main docs repo, I've assembed [a basic set of information around AsciiDoc and Antora](https://github.com/owncloud/docs/blob/master/docs/getting-started.md), along with some of the motivations for making the change (I think).

dschmidt:
@settermjd Yep, as I said just a link to that repo in the client's README would be very handy already.

settermjd:
OK, I'll get it done, @dschmidt.

settermjd:
@dschmidt, I've added a new documentation section to the readme in 5a0c042. It's not much, but it's a start.

dschmidt:
I'd argue it's not so relevant what was used before only the current state. We will rewrite it anyway when the CMake target lands .. so for now this is ok to me :+1:

README.md
```diff
@@ -61,6 +61,15 @@ Past maintainers:
 * Daniel Molkentin <daniel@molkentin.de>
 * Andreas Schneider <asn@cryptomilk.org>
 
+## Building the Documentation
+
+The documentation has been migrated from Sphinx-Doc to [Antora](https://docs.antora.org/), which is based on [the AsciiDoc format](https://github.com/owncloud/docs/blob/master/docs/getting-started.md).
```
mmattel: I would reference via link else you have to maintain two sides.
Like:
Please see the README.md in https://github.com/owncloud/docs for more information

mmattel:
@tboerger @settermjd 

I highly propose to move from ``ROOT`` to eg ``desktop_client``
``docs/modules/ROOT/pages`` --> ``docs/modules/desktop_client/pages``
We just have created in the docs repro ``administration_manual`` and moved the former contents of ``ROOT`` there and added new content with a general description into ``ROOT``.

**Bonus:**
It is clear what is meant and where to find.
Harmonisation to the docs repro.
Whenever there is the wish to move the documentation from this repro to repro docs...

michaelstingl:
Maybe related:
[Create man pages from AsciiDoc](https://asciidoctor.org/docs/user-manual/#man-pages)

tboerger:
Generating a pdf or man page is also possible without antora, if this is the only documentation required by packages we just need asciidoctor which could be already available on the major distros (and generally ruby packaging is much more mature than nodejs).

dragotin:
Asciidoctor is available in openSUSE, at least in TW. 

Still I see no reason to do this change. To me it just looks like a case of _let's do it differently_ rather than a real improvement. Also @settermjd [Link](https://github.com/owncloud/docs/blob/master/docs/getting-started.md) does not explain the motivation for me. Sorry.

dschmidt:
> Asciidoctor is available in openSUSE, at least in TW.

That's good :-) 

> Still I see no reason to do this change. To me it just looks like a case of _let's do it differently_ rather than a real improvement. Also @settermjd [Link](https://github.com/owncloud/docs/blob/master/docs/getting-started.md) does not explain the motivation for me. Sorry.

It indeed doesn't. I think the main reason to do this change is so that all of ownCloud's documentation can be built together and integrated into one big manual (with that also comes consistency which makes it easier to contribute to all areas of documentation in ownCloud fwiw). If I understood @settermjd correctly when we talked about it earlier, he said that asciidoc rendering is way more reliable and doesn't break as often as the old format.

To me it seems to make sense to switch, but it's not important to me.
Important to me is that we decide what to do *soon*. Right now it's unclear where to add new documentation and having this branch diverge from 2.5/master is a nightmare IMHO.

tboerger:
We need to define what is required on the client side, like adding docs to the packages, what are we adding now, what needs to be changed for that.

I think it is already clear that we won't drop the plans to migrate to adoc and antora for the general public available documentation, the time to discuss or change that is already over, even if @dragotin dislikes it.

dragotin:
Interesting. Where can community guys (are there still any?) like me re-read the discussion and understand the arguments?
And yes, you should not care for my likes or dislikes. But maybe for my arguments ;-) 

I am out of this now, given this attitude. Do whatever you want. I will work around that in my openSUSE client package, or just drop the docs.

felixboehm:
Public tickets https://github.com/owncloud/documentation/issues?utf8=✓&q=is%3Aissue+is%3Aopen+antora

Public discussions on documentation, please join 
https://talk.owncloud.com/channel/documentation

We value your opinion and arguments and we work open.
Reasons for the change are mainly a better build process, which is more easy to maintain, and a much better feature for writing. 
The decision was made a long time ago, and we needed time for the migration.

Please share your arguments if you see any blockers.

dragotin:
I shared my blockers above but I am happy to repeat:  

- Changing to a doc system that is not in openSUSE breakes the package of the ownCloud Client that I maintain. 
- I still do not see a *substancial* benefit of the new system over the old one, and none of the people involved here is naming one. Of course, it is new, which is always good, and maybe it is even a bit better, but for me that would not be enough to do this significant migration work and break existing workflows. The effort should rather be spent into the content rather than the framework. CONTENT is what is benefiting the users. 

Let me also repeat that I stop caring about this. Next time I update the package, I will check if I can build docs and if not, just drop it.

guruz:
> I will check if I can build docs and if not, just drop it.

I often get this feeling too with oC, changing things for the sake of changing things.
Feel free to drop the docs in your packages and just link to doc.owncloud.org ..
Thanks for continuing to maintain the openSUSE builds.

mmattel:
Before you genrealize and stomp a change into hell, we do not do this for the sake of a task or to be important. There have been many thoughts upfront before this decision has been made - and it was a good, important and necessary decision. reStructuredText/Markdown has many advantages if you ceate content like a documentation out of it. But we have reached a level of complexity where the positive nature turns into barriers. AsciiDoc with Antora as backend is much easier to install, use and maintain when you have a more complex documentation as we have now reached. As a big benefit, you have a much better quality control with asciidoc. The writing difference is not that much and if you are used to rst you will find an easy way to ascidoc.

Please see README.md in https://github.com/owncloud/docs and the getting started guide at https://github.com/owncloud/docs/blob/master/docs/getting-started.md

Personal note: I had my doubts on that path too and there was a lot of insecurity. I knew the devil and I hated it. But have to admit that I would not want to turn back after writing a bunch of documents in ascidoc.

dragotin:
> I think it's perfectly fine if @dragotin 's distro packages documentation.html just link to doc.owncloud.org/desktop/$WHATEVERMAJORVERSION

Well, usually you try to keep the distro packages independant from netwokrk, think of rural areas with bad networking where you do not wanna waste bandwidth to access docs, but be happy to have the docs on the DVD. But well, that is just a little problem, and also may vanish again once the new shiny doc system makes it into the docs, I can build it again.

michaelstingl:
> I am still trying to understand why you wanna do the change, what is the real _benefit_ of antora?

@guruz @dragotin @settermjd shared his ideas and our pain in the blog post a while ago: https://owncloud.org/news/owncloud-docs-migrating-antora-pt-1-2/

dragotin:
Thanks @michaelstingl , that blog is so useful. It explains the rationale behind. Why has nobody referenced it before? :man_facepalming: 
Ok, let's all be friends and go through this.... somehow :smile_cat:

michaelstingl:
@settermjd @dschmidt any blockers left?

mmattel:
I highly recommend to look for my comments regarding README.md and ROOT.
Especially, the topic on ROOT, if we dont do that now, it will be a big pain afterwards

tboerger:
There is only a single manual, so imho root is totally fine.

settermjd:
@mmattel, in this case, I agree with @tboerger, as there's only one manual, then ROOT makes sense, as it's the root (and only) module). If, however, there were a number of modules, then having ROOT as the overarching/coordinating module would make sense.

settermjd:
@michaelstingl, the only thing left to do is to finish the integrated site search. It's almost done, just have to fix some minor bugs in the revised build process. Aiming to have it complete this Friday at the latest.

dschmidt:
The build failure seems unrelated to this branch (would be really surprising if documentation changes triggered a failure in code tests). So I'd say: :shipit:

dschmidt:
:tada: :tada: :tada:

PR #6798, opened 2018-10-08
Fix crash relating to aborts while chunked v1 uploads are ongoing

ckamm:


src/libsync/owncloudpropagator.cpp
```diff
@@ -838,7 +841,7 @@ void PropagatorCompositeJob::slotSubJobFinished(SyncFileItem::Status status)
     // Delete the job and remove it from our list of jobs.
     subJob->deleteLater();
     int i = _runningJobs.indexOf(subJob);
-    ASSERT(i >= 0);
+    ENFORCE(i >= 0); // should only happen if this function is called more than once
```
ogoffart: These kind of changes are scary in a stable branch.

ckamm: My reasoning was: Why use `ASSERT()` when the very next line will crash if the condition is false? But yes - also adding the other `ENFORCE()` might lead to new crashes if there are more hidden bugs.

guruz: I like this approach. Better fail hard than giving inconsistent state / crash.
(@dschmidt ENFORCE will also trigger crash reporter right?)

PR #6802, opened 2018-10-09
Propagator: Add assert against duplicate done() calls

ckamm:
TODO: Does `ENFORCE()` trigger the crash reporter properly?


ckamm:
I can't easily test with the crash reporter, but I don't think current `ENFORCE` (using `qFatal`) crashes the application on windows. There's logging and then the process exits, see `qt_message_fatal`.

guruz:
@dschmidt How can we get ENFORCE/qFatal into the crash reporter?

ogoffart:
Merge?

PR #6805, opened 2018-10-10
Fix deployment with new openssl

dschmidt:


dschmidt:
You tested these commits (although they were still based on master/2.6) with a new Qt+openSSL. I will now build this branch with our old Qts to see they are still working. Then this is good to merge IMHO.

dschmidt:
Built went ok! (As expected as this is kind of a new code path with our new Qt root and doesn't change anything for the status quo)

dschmidt:
Pushed a small fix https://github.com/owncloud/client/commit/ba73f8e5d727d6aaf9b7c885cf87ab8ada97288b

This was neccessary after having symlinked a qt install dir...

Issue #6804, opened 2018-10-10
Subdirectories of directories moved into a synced directory sometimes do not sync on 2.5.0


cdhowie:
### Expected behaviour
When moving a directory with subdirectories into a directory that is being synced with ownCloud, all (unignored) files in the tree are uploaded to the server.

### Actual behaviour
When moving a directory with subdirectories into a directory that is being synced with ownCloud, sometimes the subdirectories do not get synced to the server.  No error is given and the files are not listed as ignored; it's as though the client does not see them at all.

The subdirectories themselves appear on the server, but are empty.

"Force sync now" does not resolve the problem. Restarting the client _does_ resolve the problem, and is an effective (though tedious and annoying) workaround.

### Steps to reproduce
1. Move a directory containing subdirectories with files into the directory structure of a directory being synced with ownCloud.
2. Somewhat randomly (my tests show between 25%-50% of the time) files in at least one subdirectory do not sync until the client is restarted.  If a specific directory exhibits this issue, its parent directory can be moved out of the synced directory tree and back into the tree and the problem will recur with that directory 100% of the time.

### Server configuration
Operating system: Debian Stretch

Web server: nginx + php7.0-fpm

Database: PostgreSQL 9.6.10

PHP version: 7.0

ownCloud version: 10.0.10

Storage backend (external storage): N/A

### Client configuration
Client version: 2.5.0

Operating system: Windows 10

OS language: US English

Qt version used by client package (Linux only, see also Settings dialog): N/A

Client package (From ownCloud or distro) (Linux only): N/A

Installation path of client: `C:\Program Files (x86)\ownCloud`

### Logs

1. Client logfile: I have captured the relevant log entries but, as they contain details about my private installation, I do not want to disclose them on a public issue tracker.  Due to the size of the snippet, it would take some effort to sanitize them.  If the log entries are required, I would prefer to disclose them privately.
2. Web server error log: No errors reported
3. Server logfile: No new entries in the log during the broken sync operation

ogoffart:
Thanks for the bug report.  This is going to be fixed with PR #6806 .
It is caused by the thanges to rely more on the file system watcher in 2.5.0.

Note that there is a one hour timeout for which we would anyway do a full discovery.

You can work around the problem by decreasing the timeout using the `OWNCLOUD_FULL_LOCAL_DISCOVERY_INTERVAL` environement variable, or the `fullLocalDiscoveryInterval` config entry in the fonfig file  (value in ms)

[...]

PR #6806, opened 2018-10-11
Local discovery: always recurse within new directory

ogoffart:
Regardless of what the file system watcher said

Issue #6804

### DIFF
--- a/src/libsync/syncengine.cpp
+++ b/src/libsync/syncengine.cpp
@@ -1642,6 +1642,22 @@ void SyncEngine::setLocalDiscoveryOptions(LocalDiscoveryStyle style, std::set<QB
 {
     _localDiscoveryStyle = style;
     _localDiscoveryPaths = std::move(paths);
+
+    // Normalize to make sure that no path is a contained in another.
+    // Note: for simplicity, this code consider anything less than '/' as a path separator, so for
+    // example, this will remove "foo.bar" if "foo" is in the list. This will mean we might have
+    // some false positive, but that's Ok.
+    // This invariant is used in SyncEngine::shouldDiscoverLocally
+    QByteArray prev;
+    auto it = _localDiscoveryPaths.begin();
+    while(it != _localDiscoveryPaths.end()) {
+        if (!prev.isNull() && it->startsWith(prev) && (prev.endsWith('/') || *it == prev || it->at(prev.size()) <= '/')) {
+            it = _localDiscoveryPaths.erase(it);
+        } else {
+            prev = *it;
+            ++it;
+        }
+    }
 }
 
 bool SyncEngine::shouldDiscoverLocally(const QByteArray &path) const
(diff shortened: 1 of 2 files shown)


ogoffart:
@ckamm: I'm sure this cover the case in which a directory is removed, then replaced by a new directory with subfolder. 
Should i also check for INSTRUCTION_EVAL or would that be to broad?

ckamm:
@ogoffart As far as I remember EVAL for directories happens a lot on some platforms and then later decays to NONE if no children changed? So I think it'd be too broad.

I'm surprised this is a problem: Shouldn't the filewatcher see the new directory and thus it and its contents be marked for local discovery? I'll need to check out the test.

test/testlocaldiscovery.cpp
```diff
@@ -149,6 +149,29 @@ private slots:
         QVERIFY(fakeFolder.currentRemoteState().find("A/a4"));
         QVERIFY(tracker.localDiscoveryPaths().empty());
     }
+
+    void tesdDirectoryAndSubDirectory()
```
ckamm: typo: testd

ogoffart:
> Shouldn't the filewatcher see the new directory and thus it and its contents be marked for local discovery? 

No.  The new directory are not to the filesystemwatcher before the sync.  It is only added when the sync engine sucessfully sync a directory (in `Folder::slotItemCompleted` )

But even if it was imediatly added (or on system where it is recursive) if you do`mv  folderWithManySubs  ~/sync_folder/` , you would still have only one notification from the file system that folderWithManySubs is created. but not for the sub folder (which indeed don't change)

test/testlocaldiscovery.cpp
```diff
@@ -76,19 +76,24 @@ private slots:
...
-            { "A/X", "foo bar space/touch", "foo/", "zzz" });
+            { "A/X", "A/X/beta", "foo bar space/touch", "foo/", "zzz", "zzzz" });
 
         QVERIFY(engine.shouldDiscoverLocally(""));
         QVERIFY(engine.shouldDiscoverLocally("A"));
         QVERIFY(engine.shouldDiscoverLocally("A/X"));
         QVERIFY(!engine.shouldDiscoverLocally("B"));
         QVERIFY(!engine.shouldDiscoverLocally("A B"));
         QVERIFY(!engine.shouldDiscoverLocally("B/X"));
-        QVERIFY(!engine.shouldDiscoverLocally("A/X/Y"));
```
ogoffart: This one was not valid.

src/libsync/syncengine.cpp
```diff
@@ -1642,6 +1642,18 @@ void SyncEngine::setLocalDiscoveryOptions(LocalDiscoveryStyle style, std::set<QB
 {
     _localDiscoveryStyle = style;
     _localDiscoveryPaths = std::move(paths);
+
+    // Normalize to make sure that no path is a prefix of another
+    QByteArray prev;
+    auto it = _localDiscoveryPaths.begin();
+    while(it != _localDiscoveryPaths.end()) {
+        if (!prev.isNull() && it->startsWith(prev) && (prev.endsWith('/') || *it == prev || it->at(prev.size()) == '/')) {
```
ckamm: I don't think this is right because of the dreaded ' ' < '/'. So: `"foo" < "foo bar" < "foo/bar"` and iterating would not remove the redundant "foo". The way we solved this in selective sync was to ensure directories always have a trailing '/' I think.

src/libsync/syncengine.cpp
```diff
@@ -1650,8 +1662,13 @@ bool SyncEngine::shouldDiscoverLocally(const QByteArray &path) const
         return true;
 
     auto it = _localDiscoveryPaths.lower_bound(path);
-    if (it == _localDiscoveryPaths.end() || !it->startsWith(path))
+    if (it == _localDiscoveryPaths.end() || !it->startsWith(path)) {
+        if (it != _localDiscoveryPaths.begin() && path.startsWith(*(--it))) {
```
ckamm: And this probably fails for similar reasons: if "foo" and "foo bar" are in the set, checking "foo/bar" would have <end> as lower bound, and the preceding entry isn't a prefix of the checked path. Tests for the space-case are needed.

Issue #6808, opened 2018-10-11
client 2.5.0 deletes files that are moved into a renamed folder


jnweiger:
Ubuntu 18.04 linux mint 19 tara

Connect to demo.owncloud.org, wait until the skeleton files have synced. In my case the main sync folder was `ownCloud7`.

Execute the following commands:

```
cd ownCloud7
ls -la Photos/Squirrel.jpg
  -rw-r--r-- 1 testy testy 233724 Okt 11 22:22 Photos/Squirrel.jpg
mkdir foo
sleep 2
mv foo bar
sleep 2
mv Photos/Squirrel.jpg bar
```

Check with a web browser: Squirrel.jpg is no longer on the demo.owncloud.org -- neither in Photos nor in bar.

The client Sync Protocol window logs this (from bottom to top):

```
   11.10.18 22:53:11,       Photos/Squirrel.jpg,     ownCloud7,        Deleted,          
   11.10.18 22:52:57,                       foo,     ownCloud7,   Moved to bar,          
   11.10.18 22:52:48,                       foo,     ownCloud7,       Uploaded,          
   11.10.18 22:52:22,       ownCloud Manual.pdf,     ownCloud7,     Downloaded,    4,7 MB
   11.10.18 22:52:14,       Photos/Squirrel.jpg,     ownCloud7,     Downloaded,    228 KB
   11.10.18 22:52:11,  Photos/San Francisco.jpg,     ownCloud7,     Downloaded,    211 KB
   11.10.18 22:52:08,          Photos/Paris.jpg,     ownCloud7,     Downloaded,    223 KB
   11.10.18 22:52:05,     Documents/Example.odt,     ownCloud7,     Downloaded,     35 KB
   11.10.18 22:52:02,                    Photos,     ownCloud7,     Downloaded,          
   11.10.18 22:52:02,                 Documents,     ownCloud7,     Downloaded,       
```

On the client side the file Squirrel.jpg still exists in folder bar. But it was deleted on the server, its folder bar is now empty.

```
echo hey > bar/wakeup.txt
```

This also does not sync anything to the server. The f12 log mentions wakeup.txt with only one line:

```
10-11 23:07:24:739 [ info gui.folderwatcher ]:  Detected changes in paths: QSet("/home/testy/ownCloud7/foo/wakeup.txt")
```

Note that it shows the old folder name `foo` instead of the actual folder name `bar`. This is wrong.

ckamm:
Very possibly connected to #6804 #6806

[...]

PR #6809, opened 2018-10-12
Folderwatcher: On linux, fix paths after dir renames

ckamm:
If a folder was renamed A -> B, the folder watcher for the inode
would be unaware and still report changes for A/foo. Now directory
renames in the watched folders are tracked and paths are updated
accordingly.

For  #6808

### DIFF
--- a/src/gui/folderwatcher_linux.h
+++ b/src/gui/folderwatcher_linux.h
@@ -23,6 +23,8 @@
 
 #include "folderwatcher.h"
 
+class QTimer;
+
 namespace OCC {
 
 /**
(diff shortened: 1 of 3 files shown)


src/gui/folderwatcher_linux.h
```diff
@@ -44,17 +46,34 @@ protected slots:
...
 
     QString _folder;
     QHash<int, QString> _watches;
     QScopedPointer<QSocketNotifier> _socket;
     int _fd;
+
+    // Maps inotify event cookie to rename data.
+    QHash<quint32, Rename> _potentialRenames;
+
+    QTimer *_wipePotentialRenamesSoon;
```
ogoffart: Maybe does not need to be a pointer.

PR #6813, opened 2018-10-15
Attempt to fix appveyor build

ogoffart:
It tries to compile LLVM because of qttools. (qdoc)
Compiling llvm takes way to long and is timing out.
We do not need qdoc anyway.


ogoffart:
It seems this worked

PR #6816, opened 2018-10-16
Disable autostartCheckbox if autostart is configured system wide

dschmidt:


dschmidt:
Hehe, the monoIconsCheckbox apparently was already at the bottom because it can be hidden by the theme ... suggestions how to handle this without overcomplicating the code?

dschmidt:
I like that! Will do!

dschmidt:
Tested on Win7 and Win10.

guruz:
(Forgot to mention issue ID in commit..i think with enterprise issues we often forget that)

Issue #6814, opened 2018-10-15
OAuth2 token expiring during a very long discovery makes sync impossible


ogoffart:
As reported in https://github.com/owncloud/client/issues/6677 and https://github.com/owncloud/enterprise/issues/2880:

The discovery of huge trees can takes long, and if it takes more than one hour, the oauth2 token expires and the sync aborts.
Next sync will start discovery from scratch again, which will take more than one hour again,  and the client will never sync.

Workaround: use selective sync to avoid syncing part of the tree, wait untill this syncs successfully, then enable more directories, little by little.

ogoffart:
Possible solutions:

1. Wait for 2.6 and the `new_discovery_algorithm` branch to be merged. This have a much faster discovery as it does work in parallel.  Hopefully this can be down under 1h.  In the long term (although this is not yet implemented), we can start propagation in parallel, which means that part already propagated do not need to be re-discovered on next sync.

2. Have a discovery cache ( https://github.com/owncloud/client/issues/2976#issuecomment-291930131 ) .  But this would also be a 2.6 feature and would be pointless if we have propagation in parallel of discovery.

3. Maybe the most obvious: instead of breaking the sync when the token expire, just renew the token and continue. However, this is not as simple as it sounds.
I guess this could be do-able if we:
   -  In `HttpCredentials::slotAuthentication`  when using oauth2, we need to request the new token. We need to introduce a new state while we are requesting a token.
   - intercept that particular error in `AbstractNetworkJob::slotFinished`, and redo the request and enqueue it somewhere. (where?)
   - Queue any other requests that are done while we are requesting the token.
   - Once we have the token, we can re-do all the requests. In case of error, cancel them all.

All of that is quite some work, and maybe bit difficult to fix in 2.5.1

ckamm:
I think 3 is the only approach that's not just a workaround: There could be odd situations where discovery takes a while longer, or other network activity that's broken by expiring tokens. Maybe some server admin makes the tokens expire much more frequently.

But as you say it's also error prone. There could even be multiple requests in flight that all come back with a token expired error. Restarting requests isn't too hard (done for redirects already) - but all the queuing and holding would be quite complex.

guruz:
@ogoffart @ckamm Why not blindly renew the token every 10 minutes and for each new request use the current latest one? ("viel renew hilft viel")
(I don't know how this works server side, will the old and new token be valid at same time or will this somehow then clash? CC @DeepDiver1975 )

[...]

PR #6819, opened 2018-10-17
OAuth2: Refresh the token without aborting the sync

ogoffart:
OAuth2 access token typically only has a token valid for 1 hour.
Before this patch, when the token was timing out during the sync, the
sync was aborted, and the ConnectionValidator was then requesting a new
token, so the sync can be started over.
If the discovery takes longer than the oauth2 validity, this means that
the sync can never proceed, as it would be always restarted from scratch.
    
With this patch, we try to transparently renew the OAuth2 token and restart
the jobs that failed because the access token was invalid.
    
Note that some changes were required in the GETFile job because it handled
the error itself and so it was erroring the jobs before its too late.
    
Issue #6814

### DIFF
--- a/src/libsync/creds/httpcredentials.cpp
+++ b/src/libsync/creds/httpcredentials.cpp
@@ -103,12 +103,6 @@ static void addSettingsToJob(Account *account, QKeychain::Job *job)
     job->setSettings(settings.release());
 }
 
-HttpCredentials::HttpCredentials()
-    : _ready(false)
-    , _keychainMigration(false)
-{
-}
-
 // From wizard
 HttpCredentials::HttpCredentials(const QString &user, const QString &password, const QSslCertificate &certificate, const QSslKey &key)
     : _user(user)
--- a/src/libsync/creds/httpcredentials.cpp
+++ b/src/libsync/creds/httpcredentials.cpp
@@ -45,6 +45,7 @@ namespace {
     const char clientCertificatePEMC[] = "_clientCertificatePEM";
     const char clientKeyPEMC[] = "_clientKeyPEM";
     const char authenticationFailedC[] = "owncloud-authentication-failed";
+    const char needRetryC[] = "owncloud-need-retry";
 } // ns
 
 class HttpCredentialsAccessManager : public AccessManager
@@ -84,8 +85,15 @@ protected:
             req.setSslConfiguration(sslConfiguration);
         }
 
+        auto *reply = AccessManager::createRequest(op, req, outgoingData);
 
-        return AccessManager::createRequest(op, req, outgoingData);
+        if (_cred->_isRenewingOAuthToken) {
+            // We know this is going to fail, but we have no way to queue it there, so we will
+            // simply restart the job after the failure.
+            reply->setProperty(needRetryC, true);
+        }
+
+        return reply;
     }
 
 private:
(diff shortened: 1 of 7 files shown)


ogoffart:
I have been testing this by adding this in HttpCredential::fetchFromKeychain

```
    auto timer = new QTimer(this);
    timer->setInterval(10000);
    QObject::connect(timer, &QTimer::timeout, this, [this] { qWarning() << "YOLO";  _password = "yolo"; });
    timer->start();
```

This simulate an expiration of the access token every 10 second. Then one
can verify that the sync goes through and that other side jobs such as expanding 
the "choose what to sync" tree, are working correctly.

Making an automated test for this would be quite hard as this is in the HttpCredential class, and this class is currently not automatically tested/testable. (All our sync engine test are using fake credentials)

src/libsync/creds/httpcredentials.cpp
```diff
@@ -517,12 +527,31 @@ void HttpCredentials::slotAuthentication(QNetworkReply *reply, QAuthenticator *a
 {
     if (!_ready)
         return;
+    if (_isRenewingOAuthToken) {
+        reply->setProperty(needRetryC, true);
+    } else if (isUsingOAuth() && !reply->property(needRetryC).toBool()) {
+        reply->setProperty(needRetryC, true);
+        refreshAccessToken();
```
ckamm: log this happening

src/libsync/propagatedownload.cpp
```diff
@@ -159,9 +159,14 @@ void GETFileJob::slotMetaDataChanged()
 
     int httpStatus = reply()->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
 
-    // Ignore redirects
-    if (httpStatus == 301 || httpStatus == 302 || httpStatus == 303 || httpStatus == 307 || httpStatus == 308)
+    if (httpStatus == 301 || httpStatus == 302 || httpStatus == 303 || httpStatus == 307 || httpStatus == 308 || httpStatus == 401) {
+        // Ignore redirects and authentication failures.
+        // This is handled by AbstractNetworkJob,  so don't let this job error out in this case.
+        bool ok = disconnect(reply(), &QNetworkReply::finished, this, &GETFileJob::slotReadyRead)
+            && disconnect(reply(), &QNetworkReply::readyRead, this, &GETFileJob::slotReadyRead);
```
ckamm: This works because `newReplyHook` reconnects the signals, right? Suggestion:

```
// Redirects and auth failures (oauth token renew) will end up restarting the job. We do not want to process
// further data from the initial request. newReplyHook() will reestablish signal connections for the follow-up
// request.
```

Also, what if there's a 401 that *doesn't* cause a new job to be spawned? Then `finishedSignal()` wouldn't be emitted.

ogoffart: Right, I now changed GETFileJob::finished so this does not happen

ckamm: Alright, that looks like it works. What do you think about adding the comment?

ogoffart: Comment updated

guruz:
@micbar Do you want to help us to test this in one the actual setups?

micbar:
I could send a patched client to the user with the issue.

ogoffart:
@micbar : I just merged the patch, which means that the 2.5.1 daily build from tomorrow will have the patch.  They could try the daily build.

Issue #6810, opened 2018-10-12
cannot encode <Cyrillic-names> to local encoding 2252


voroyam:
User in central reported this

https://central.owncloud.org/t/some-folders-with-cyrillic-names-are-ignored/16196
Client seems to have a problem with Cyrillic characters. Error message: "cannot encode За принтиране to local encoding 2252" 

Client version 2.5 

oC version 10.0.10

Platform win 7 pro

From his post:

For several weeks my desktop client ignores the main folders with cyrillic names and all they contain. This issue is only for one of the users. Other two have no problems. 
Interestingly some folders, again with cyrillic names, were synced after I renamed the folder they were in. 

### Expected behaviour
Sync all folders and files within to the server and from the server

### Actual behaviour
The client shows the folder as ignored with issue: "The filename cannot be encoded on your system."

### Steps to reproduce
1. Rename folder  ..\ownCloud\Представяне to "Presentation"
2. All files in it are uploaded
3. Rename again to "Представяне"
4. Client logs "Folder Presentation - deleted"
5. In not synced tab there is Issue: "The filename cannot be encoded on your system."

### Server configuration
Shared hosting
Operating system: Linux

Web server: Apache 2.4.27

Database: mySQL

Server PHP version: 5.3.24
For ownCloud: 5.6

ownCloud version: latest - 10.0.10.4

Storage backend (external storage): - 

### Client configuration
Client version: 2.5.0

Operating system: Windows 7 profesional

OS language: english

Installation path of client: C:\Program Files (x86)\ownCloud

### Logs

Client log output shows a lot of similar  lines

```10-12 11:00:09:037 [ info sync.csync.updater ]:	cannot encode За принтиране to local encoding 2252```

1. Client logfile: Output of `owncloud --logwindow` or `owncloud --logfile log.txt`
(On Windows using `cmd.exe`, you might need to first `cd` into the ownCloud directory)
(See also http://doc.owncloud.org/desktop/2.2/troubleshooting.html#client-logfile )

I hope this is enough to understand the problem.

ogoffart:
This happens because the system locale codec is Windows-1252 which cannot encode cyrilic character.  At least that's what Qt tells us.
Normally, most modern system are using UTF-8.
But perhaps this user configured his system to use Windows-1252 instead of UTF-8 ?

On Linux this is important since the file system is using that locale.  On Windows however, the file system is always UCS-2, to my knowledge. And so we probably can encode everything
[...]

PR #6820, opened 2018-10-17
Discovery: Windows: Don't check if a server file name can be encoded

ogoffart:
Because on windows, all filename can be encoded: The file system uses
UTF-16, regardless of the locale

Issue: #6810

### DIFF
--- a/src/csync/csync_update.cpp
+++ b/src/csync/csync_update.cpp
@@ -38,6 +38,7 @@
 #include "csync_update.h"
 #include "csync_util.h"
 #include "csync_misc.h"
+#include "common/utility.h"
 
 #include "vio/csync_vio.h"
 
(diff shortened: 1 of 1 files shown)


src/csync/csync_update.cpp
```diff
@@ -147,7 +148,7 @@ static int _csync_detect_update(CSYNC *ctx, std::unique_ptr<csync_file_stat_t> f
   }
 
   auto localCodec = QTextCodec::codecForLocale();
-  if (ctx->current == REMOTE_REPLICA && localCodec->mibEnum() != 106) {
+  if (!OCC::Utility::isWindows() && ctx->current == REMOTE_REPLICA && localCodec->mibEnum() != 106) {
```
ckamm: Please add a comment about why windows is excluded.

Issue #6824, opened 2018-10-18
Forbidden when Move or Rename Files or Folders


georgeoftown:
### Expected behaviour
Move or rename file or folder

### Actual behaviour
403 Forbidden MOVE

### Steps to reproduce
1. Configure Client with Server Address ended with slash (Example https://owncloud.org/)
2. Sync something
3. Move or rename folder or file synced
4. Sync Error

### Server configuration
Operating system:
Debian 9.5
Web server: 
Apache/2.4.25
Database:
mysql Ver 15.1 Distrib 10.1.26-MariaDB
PHP version:
PHP 7.0.30-0+deb9u1
ownCloud version:
10.0.10

### Client configuration
Client version:
2.5.0 (build 10560)
Operating system:
Windows 10
OS language:
ES

### Logs

1. Client logfile: Output of `owncloud --logwindow` or `owncloud --logfile log.txt`
```
#=#=#=# Syncrun started 2018-10-18T11:56:20Z
#=#=#=#=# Propagation starts 2018-10-18T11:56:20Z (last step: 196 msec, total: 196 msec)
11:56:21||Pruebas/Prueba2 -> Pruebas/1|INST_RENAME|Up|1539862451|5bc86fa3018b9|0|00057074511b5315c9cf7|2|El servidor respondió "403 Forbidden" a "MOVE https://owncloud.org/remote.php/dav/files/aaaaaa.mmmmm/Pruebas/Prueba3"|403|0|0|fb133929-61e9-4e83-b085-4cb86af6c2a1|
11:56:21||Pruebas/2 -> Pruebas/23|INST_RENAME|Up|1539862435|5bc86fa7e0977|0|00057075511b5315c9cf7|2|El servidor respondió "403 Forbidden" a "MOVE https://owncloud.org/remote.php/dav/files/aaaaaa.mmmmm/Pruebas/2"|403|0|0|6790c65c-179f-44b8-b92c-347c81aff3c8|
#=#=#=# Syncrun finished 2018-10-18T11:56:20Z (last step: 171 msec, total: 368 msec)
```
2. Web server error log:
look the 3 slash together ///
```
111.111.111.111 - - [18/Oct/2018:10:26:05 +0000] "MOVE /remote.php/dav/files/aaaaaa.mmmmm/Prueba/Prueba3 HTTP/1.1" 403 297 ///remote.php/dav/files/aaaaaa.mmmmm/Prueba/Prueba2
```
3. Expected
```
[...]

PR #6834, opened 2018-10-24
Move: Fix too many starting slashes in the destination header

ogoffart:
QDir::cleanPath does not remove starting slashes on windows.
So use account::davUrl which is already cleaned

Issue: #6824

### DIFF
--- a/src/libsync/propagateremotemove.cpp
+++ b/src/libsync/propagateremotemove.cpp
@@ -89,8 +89,7 @@ void PropagateRemoteMove::start()
     }
 
     QString source = propagator()->_remoteFolder + _item->_file;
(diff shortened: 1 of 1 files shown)


src/libsync/propagateremotemove.cpp
```diff
@@ -89,8 +89,7 @@ void PropagateRemoteMove::start()
     }
```
guruz: Would it make sense to somehow use Utility::concatUrlPath ? Maybe not.

guruz:
> I think there's a case for making clear that davPath is relative to the account url in its doc comments.

+1 Please be verbose in the .h there

PR #6840, opened 2018-10-25
Database: future-proof to allow downgrade from future version

ogoffart:
The 2.6 client will introduce an index that depends on a custom
function.
This causes insersion onto the metadata database to fail if the database
it opened with older version of the client. (Which will cause the client
to abort)

Delete this index in this version in order to allow downgrade.

OF course this commit need not to be in the 2.6 release

### DIFF
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -313,6 +313,20 @@ bool SyncJournalDb::checkConnect()
         qCInfo(lcDb) << "sqlite3 version" << pragma1.stringValue(0);
     }
 
+    {
+        // Future version of the client (2.6) will have an index 'metadata_parent' which
+        // depends on a custom sqlite function which does not exist yet in 2.5.
+        // So make sure to remove the index if it exists, otherwise we will crash when inserting
+        // rows in the metadata database.
+        // This needs to be done before the synchronous mode is enabled.
+        // The 2.6 client will anyway re-creates this index if it does not exist.
+        SqlQuery query(_db);
+        query.prepare("DROP INDEX IF EXISTS metadata_parent;");
+        if (!query.exec()) {
+            return sqlFail("updateMetadataTableStructure: remove index metadata_parent", query);
+        }
+    }
+
     pragma1.prepare("PRAGMA journal_mode=" + _journalMode + ";");
     if (!pragma1.exec()) {
         return sqlFail("Set PRAGMA journal_mode", pragma1);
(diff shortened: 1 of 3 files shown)


guruz:
AppVeyor is failing.
```
configparser.DuplicateOptionError: While reading from 'C:\\CraftMaster\\windows-msvc2017_32-cl\\etc\\CraftSettings.ini' [line 11]: option 'python' in section 'Paths' already exists
```
@dschmidt

guruz:
> OF course this commit need not to be in the 2.6 release

You will take care of having it not there after 2.5->master merge?

I wonder if there is no better solution for this whole thing. What if someone downgrades to something-lower-than-2.5.1?

ckamm:
@ogoffart @guruz For accounts I had added a "version" int that can be incremented when we do changes that are not backwards compatible. Folders already store a version int too, but it's not considered during reading. If this was the case we could increment it to avoid older clients seeing the config that's too new for them.

@ogoffart's change here is more flexible and would allow older clients to use 2.6+ configurations. But it'd still break awkwardly on <=2.5.0 - that's where the versions could be useful.

This would also need some extra work to to make sure older clients preserve too-new configuration without messing it up.

guruz:
> Maybe there's a way to check for the availability of the function in SQL? I'm currently not seeing how.

Doesn't seem to. :( 
https://stackoverflow.com/questions/15701159/how-do-i-call-user-defined-function-in-sqlite-only-if-it-exists
https://stackoverflow.com/questions/17003447/how-can-i-check-whether-a-function-exists-in-sqlite

ogoffart:
> You will take care of having it not there after 2.5->master merge?

Yes, i'll take care of that.

> What if someone downgrades to something-lower-than-2.5.1

Downgrading is not something we officially support. I'm only doing that out of kindness for beta tester.
When they want to downgrade to an older version, they'd have to manually run 
```
sqlite3 ._sync_*.db "DROP INDEX metadata_parent;"
```

(or delete the database)

> I wonder if there is no better solution for this whole thing.

Maybe. Feel free to suggest. This index is quite elegant as it avoids creating yet another column in the database. If this is a major problem, we could find a solution.

> Both accounts and folders carry a "version" in their settings

Right, we could increment the folder version in 2.6, and let 2.5.1 open the new version, but 2.5.0 and earlier would not.
But then even the manual DROP INDEX would not work anymore.

So I would not do it and document the workaround and hopefully people who downgrade find the workaround by googling for the problem.

In this case the client would assert, and not corrupt anything as it would have been the case with VFS

> Maybe there's a way to check for the availability of the function in SQL? 

I wouldn't think this would help, also because we don't want to have a corrupted index if the user re-upgrade.

I tried to install a TRIGGER on UPDATE in the version database, but one cannot have DROP INDEX in there :-(

ogoffart:
@ckamm:  updated , also allow version 2 folder

PR #6845, opened 2018-10-26
Fix #6799: Make OCUtil helper lib static and use static crt for shell extensions

### LINKED ISSUE #6799: Failed to register module OCContextMenu.dll. HRESULT - 2147024770
### Expected behaviour
Installation of owncloud client

### Actual behaviour
Error message. Installation fails

### Steps to reproduce
1. Remove previous version of owncloud client
2. Download latest version and execute installer
3. Watch error message

### Server configuration
Operating system: 

Web server: 

Database:

PHP version:

ownCloud version: 

Storage backend (external storage):

### Client configuration
Client version: 2.5.0.10598

Operating system: Windows 8.1 Pro

OS language:

Qt version used by client package (Linux only, see also Settings dialog):

Client package (From ownCloud or distro) (Linux only):

Installation path of client:

### Logs

Please use Gist (https://gist.github.com/) or a similar code paster for longer
logs.

```Template for output < 10 lines```

1. Client logfile: Output of `owncloud --logwindow` or `owncloud --logfile log.txt`
(On Windows using `cmd.exe`, you might need to first `cd` into the ownCloud directory)
(See also http://doc.owncloud.org/desktop/2.2/troubleshooting.html#client-logfile )

2. Web server error log:

3. Server logfile: ownCloud log (data/owncloud.log):

### PR DESCRIPTION
**Pull request by dschmidt**

### CODE REVIEW

File: shell_integration/windows/OCOverlays/CMakeLists.txt [hunk 1]
```diff
@@ -8,6 +8,8 @@ add_library(OCOverlays MODULE
     OCOverlays.def
 )
 
+set_static_crt(OCContextMenu)
```
ogoffart: s/OCContextMenu/OCOverlays/ !!

dschmidt: Funnily enough it still works... so it indeed might be "inherited" from OCUtill...

guruz: >  so it indeed might be "inherited" from OCUtill...

Because it's the same compilation unit then

dschmidt:
Msi is already fixed in a branch too. Will merge analoguely. NSI ... well needs updating of the submodule before something breaks at all...

I have no oldish visual studio and tbh don't care too much about it right now as it's deprecated anyway...

guruz:
> I have no oldish visual studio and tbh don't care too much about it right now as it's deprecated anyway...

So as long as we don't update the binary submodule anyway it won't break...
And NSIS is not affected since we ship the CRT..all good then

dschmidt:
Btw adding a warning is not really nice, but it seems to be simplest approach:
https://stackoverflow.com/questions/51494506/replacing-md-with-mt-in-cmake-is-not-possible-in-release-mode (no answers at this point)

shell_integration/windows/CMakeLists.txt
```diff
@@ -1,3 +1,9 @@
+# Use static runtime for all subdirectories
+foreach(buildType "" "_DEBUG" "_MINSIZEREL" "_RELEASE" "_RELWITHDEBINFO")
+    string(REPLACE "/MD" "/MT" "CMAKE_CXX_FLAGS${buildType}" "${CMAKE_CXX_FLAGS${buildType}}")
+    message("status: CMAKE_CXX_FLAGS${buildType}: ${CMAKE_CXX_FLAGS${buildType}}")
```
ogoffart: maybe remove this message

guruz:
@ogoffart updated

dschmidt:
> olivier 16:57
> well, if that's they only way to do it, so be it

Issue #6846, opened 2018-10-26
owncloudcmd: Also fetch server version


guruz:
From https://github.com/owncloud/client/pull/6835

> What about the server version received in ConnectionValidator::slotCapabilitiesRecieved?

PR #6848, opened 2018-10-29
owncloudcmd: Read the server version from the capabilities

ogoffart:
Issue: #6846


guruz:
FYI @dschmidt Appveoyr seems fixed..

PR #6849, opened 2018-10-29
Settings: Make FoldersWithPlaceholders group sticky

ckamm:
If virtual files are disabled on a folder it might still have db entries
or local virtual files that would confuse older client versions.

### DIFF
--- a/src/gui/folderman.h
+++ b/src/gui/folderman.h
@@ -302,7 +302,7 @@ private:
     // restarts the application (Linux only)
     void restartApplication();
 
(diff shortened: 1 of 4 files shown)


src/gui/folderman.h
```diff
@@ -302,7 +302,7 @@ private slots:
     // restarts the application (Linux only)
     void restartApplication();
 
-    void setupFoldersHelper(QSettings &settings, AccountStatePtr account, bool backwardsCompatible, const QStringList &ignoreKeys);
+    void setupFoldersHelper(QSettings &settings, AccountStatePtr account, const QStringList &ignoreKeys, bool backwardsCompatible, bool foldersWithPlaceholders);
```
ogoffart: Maybe consider using a QFlags

PR #6851, opened 2018-10-30
Settings: Bring back version and copyright

guruz:
For owncloud/enterprise#2941

### DIFF
--- a/src/gui/generalsettings.ui
+++ b/src/gui/generalsettings.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>785</width>
-    <height>390</height>
+    <width>794</width>
+    <height>432</height>
    </rect>
   </property>
   <property name="windowTitle">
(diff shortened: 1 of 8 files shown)


src/gui/generalsettings.ui
```diff
@@ -279,8 +363,6 @@
   <tabstop>newExternalStorage</tabstop>
   <tabstop>showInExplorerNavigationPaneCheckBox</tabstop>
   <tabstop>crashreporterCheckBox</tabstop>
-  <tabstop>updateChannel</tabstop>
-  <tabstop>restartButton</tabstop>
```
guruz: i have no idea why this is like that. When i click to edit the tabsstops in designer they are shown?

src/libsync/theme.h
```diff
@@ -203,6 +203,7 @@ class OWNCLOUDSYNC_EXPORT Theme : public QObject
      * About dialog contents
      */
     virtual QString about() const;
+    virtual bool aboutShowCopyright() const;
```
ogoffart: (I would have made this return a QString, so they can customize the name if they want (or empty to hide), I foresee that customer will want that.)

michaelstingl: Use `APPLICATION_VENDOR` here?

Should be `ownCloud GmbH` in the unbranded?
https://github.com/owncloud/client/blob/v2.5.0/OWNCLOUD.cmake#L5

michaelstingl: Yeah, even when `APPLICATION_VENDOR` is "Acme Corp LLC", copyright is still "ownCloud GmbH".

guruz:
Fix on top 62252d0012c00dbba64dbe3210f104f7ee58434c

PR #6852, opened 2018-10-31
Make qFatal() trigger the crash reporter on Windows

dschmidt:


src/libsync/logger.cpp
```diff
@@ -19,6 +19,7 @@
 #include <QDir>
 #include <QStringList>
 #include <QThread>
+#include <QtGlobal>
```
ogoffart: do we really need that?

dschmidt: it's likely to be included by something else, but technically it's where `Q_OS_WIN` is defined and I don't like it to depend on implicit includes...

dschmidt:
Is there a way to include the log file when sending a crash report? The log is written before the actual crash... 

If we want to trigger the crash reporter directly, we need to extend the interface of libcrashreporter-qt..  and passing the message directly is even more involved. 

If you don't want to rely on undefined behavior in Utitlity::crash() and that's any better to you, I can check whether std::abort() for example is also caught by breakpad..

guruz:
We can’t do logfile for privacy reasons.

But sending one of our fixed fatal strings sounds good but optional since the backtrack will lead to that log message anyway on the source.

So indeed try std::abort or whatever and if yes then I think it is good enough for 2.5.1 to merge.

dschmidt:
Yeah, seems only useful if the qFatal() contains any variable output...

dschmidt:
Nope, does not work...

ogoffart:
How abouit throwing an exception?

dschmidt:
> How abouit throwing an exception?

I'm afraid that does not trigger breakpad either (at least not on Windows in MSVC builds) ...

PR #6855, opened 2018-11-01
Add extra logging to UploadDevice::prepareAndOpen() 

ckamm:
For #6699

Stacktraces point to this function as culprit. Let's make extra sure it is the buffer allocation and that the size is sane. 

FYI @piojan


ckamm:
Not needed, bug found!

Issue #6822, opened 2018-10-18
[Windows] Prevent sync trigger when file has changed but has state open (by another app) (locked file)


mmattel:
### Expected behaviour
Client should not start trying syncing a file if the file is in an open state.
The open state hapens if you eg work on an Excel or Word document.
Changes are written to that file on every save, but the file is locked by the app.

### Actual behaviour
The file as described in the example above triggers on any change a sync but it is in an open state by the app. This conflicts and you see a red crossed button on the file sync status as long the file is open. When the file gets released by the app, syncing is fine and things turn green.

### Steps to reproduce
1. Create or open an Office document, save it, but keep it open
2. Watch the client trying to sync
3. Watch the filesytem with regards to the file sync status
4. Close the Office document and see 2) and 3)

### Server configuration
Operating system: Ubuntu 16.04

Web server: nginx

Database: mysql

PHP version: 7.0

ownCloud version: 10.0.10 (stable)

Storage backend (external storage): smb, ftp, dropbox, local

### Client configuration
Client version: 2.5.0

Operating system: W10x64

OS language: DE

michaelstingl:
It seems this really difficult in the Win/Office world. Here is an overview from an user in the heise.de forum:
https://www.heise.de/forum/iX/News-Kommentare/ownCloud-Neue-Synchronisationsfunktion-fuer-den-Desktop/Re-Ich-wuerde-Filelocking-im-Desktop-Client-dringender-benoetigen/posting-32458876/show/

mmattel:
The point is, that the friends at MS create a new file containing an ``~`` sign in the filename (somewhere in the filename, see link above).
BUT MS also locks the original file and prevents double writing on. When the application closed things get cleand up and you have a correct freed file with new content.
My request is to not only check for state ``has changed`` but also for state ``is write opened`` by another app (no care on ``is read opened``). And do not sync in case both are present, only in first case.

mmattel:
Another usecase which is much more highlighting this problematic is iTunes.
If you have your library in a folder which gets synced with the client, you see this much more clearly. 
iTunes has two files which trigger this issue: the database ``iTunes Library.itl`` and a file called ``sentinel``.
iTunes does many changes in the database while the app is open. ``sentinel`` gets deleted when iTunes is closed. Both get sync errors constantly during iTunes is open.

mmattel:
More impact (really bad)
- I opened a already synced Word document, did a change, Word still open, client tries to sync - fails.
[...]

PR #6859, opened 2018-11-05
Folder: Treat file unlock similar to external change

ckamm:
Otherwise the local discovery list might not get filled as needed.

Possibly for #6822, maybe for 2.5.1 too?

FYI @guruz  @mmattel


ogoffart:
Thanks for the test  :+1:

PR #6864, opened 2018-11-06
Add Ctrl-Alt-Shift-L as log window shortcut

ckamm:
F12 is taken on OSX and there's no other way of showing it.


ogoffart:
Looks good.

But why Ctrl+Alt+Shift ?  Usually, shortcut with many modifier might be taken by the system.
Perhaps just a simpler shortcut?  (Ctrl+L)

ckamm:
True, fixed

guruz:
Fix pushed 5c98193c63e5cdd892557ef0c74344bf314032bb

PR #6866, opened 2018-11-06
A couple of optimisations in the discovery algorithm

ogoffart:
The new discovery algorithm tried not to touch too much of the code to minimize the diff, but now, one can refactor a bit. In particular, we can do more with QString without the need to go back and forth.

### DIFF
--- a/src/csync/vio/csync_vio_local_win.cpp
+++ b/src/csync/vio/csync_vio_local_win.cpp
@@ -41,81 +41,53 @@ Q_LOGGING_CATEGORY(lcCSyncVIOLocal, "sync.csync.vio_local", QtInfoMsg)
  * directory functions
  */
 
-typedef struct dhandle_s {
+struct csync_vio_handle_t {
   WIN32_FIND_DATA ffd;
   HANDLE hFind;
   int firstFind;
-  mbchar_t *path; // Always ends with '\'
-} dhandle_t;
+  QString path; // Always ends with '\'
+};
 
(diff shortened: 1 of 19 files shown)


src/csync/vio/csync_vio_local_win.cpp
```diff
@@ -41,81 +41,53 @@ Q_LOGGING_CATEGORY(lcCSyncVIOLocal, "sync.csync.vio_local", QtInfoMsg)
...
 
-  handle = (dhandle_t*)c_malloc(sizeof(dhandle_t));
+    QScopedPointer<csync_vio_handle_t> handle(new csync_vio_handle_t{});
 
-  // the file wildcard has to be attached
-  int len_name = strlen(name);
-  if( len_name ) {
-      char *h = NULL;
+    // the file wildcard has to be attached
+    QString dirname = name + QLatin1String("/*");
```
ckamm: I don't see the `path_to_locale` call anywhere. It's the thing that also converts to UNC paths for long-filename compatibility.

ogoffart: well spotted!

ogoffart:
UNC path fixed.  
(maybe i should add a test for that)

> Re: setFileRecordMetadata: did you double check there are no other fields that might accidentally get wiped like that?

Yes, only the checksum was kept.  
Anyway, currently, all the fields are coming from the DB already everytime it is used.

PR #6868, opened 2018-11-07
Discovery: Fix downloading files when database is used for local discovery

ogoffart:
This also fix the currently failling LockedFilesTest


guruz:
@ogoffart What does this mean? Is this a blocker for 2.5.1 or not?

guruz:
(Assuming this is 2.6)

ogoffart:
The remaining test faillure is fixed in https://github.com/owncloud/client/pull/6860

I'm going to merge now so we can have a working CI  

@guruz   No, it does not apply to 2.5, this is just the fact that a test that was introduced in 2.5 is breaking when it was merged in master because master had a bug.

PR #6870, opened 2018-11-08
Remarkable tablet build support

gergap:
Hi, I'm trying to cross-compile owncloudcmd for remarkable tablet.
Here are some fixes for that.

the libraries build already.
the owncloudcmd tool still has linker errors.


src/CMakeLists.txt
```diff
@@ -15,7 +15,7 @@ if(NOT TOKEN_AUTH_ONLY)
 endif()
 
 if(NOT MSVC)
-  if(NOT (CMAKE_SYSTEM_PROCESSOR MATCHES "^(alpha|parisc|hppa)"))
+    if(NOT (CMAKE_SYSTEM_PROCESSOR MATCHES "^(alpha|parisc|hppa)") AND NOT CMAKE_CROSSCOMPILING)
```
dschmidt: The indention change looks wrong.

CMAKE_CROSSCOMPILING is way too unspecific imho. Cant you more reliably detect what you are compiling for?

gergap: When cross-compiling the compiler and linker flags are set by the CMake toolchain file.
You should not modify them in the CMakeLists.txt.
So it would make sense to move all compiler settings into a NOT CROSS_COMPILING section,
so that you don't interfere with target specific compiler settings.

Because handling all possible target settings in a CMakeLists.txt becomes a mess,
there are these toolchain files in CMake.
See
https://cmake.org/cmake/help/v3.6/manual/cmake-toolchains.7.html#cross-compiling

e.g. my toolchainfile `toolchain-linux-yocto-cortex-a9.cmake` file contains this:

    [snip]
    # Set compiler flags
    set(COMPILER_FLAGS "-march=armv7-a -marm -mfpu=neon -mfloat-abi=hard -mcpu=cortex-a9 --sysroot=${CMAKE_SYSROOT}")
    set(CMAKE_C_FLAGS "${COMPILER_FLAGS}" CACHE STRING "" FORCE)
    set(CMAKE_CXX_FLAGS "${COMPILER_FLAGS}" CACHE STRING "" FORCE)
    
    # linker settings
    set(CMAKE_EXE_LINKER_FLAGS "--sysroot=${CMAKE_SYSROOT}")
    set(CMAKE_SHARED_LINKER_FLAGS "--sysroot=${CMAKE_SYSROOT}")
    [snip]

I can fix the indentation of course.

src/CMakeLists.txt
```diff
@@ -44,7 +44,9 @@ endif()
 add_subdirectory(csync)
 add_subdirectory(libsync)
 if (NOT BUILD_LIBRARIES_ONLY)
-    add_subdirectory(gui)
+    if (BUILD_GUI)
```
ogoffart: maybe (not TOKEN_AUTH_ONLY) , so no need to add an option

src/CMakeLists.txt
ogoffart: maybe the solution was to add arm in the list of excluded cpu.

guruz:
@gergap thanks for the contribution!

PR #6882, opened 2018-11-14
Introduce VFS plugins and improvements

ckamm:
- Different kinds of vfs modes based on plugins
- Suffix-file based vfs is one of these
- Adjustments to discovery and propagation for allowing non-suffix vfs modes

### DIFF
--- a/src/common/result.h
+++ b/src/common/result.h
@@ -14,21 +14,16 @@
 
 #pragma once
 
+#include "asserts.h"
+
 namespace OCC {
 
 /**
- * A Result of type T, or an Error that contains a code and a string
- *
- * The code is an HTTP error code.
- **/
-template <typename T>
+ * A Result of type T, or an Error
+ */
+template <typename T, typename Error>
 class Result
 {
-    struct Error
-    {
-        QString string;
-        int code;
-    };
     union {
         T _result;
         Error _error;
@@ -38,12 +33,16 @@ class Result
 public:
     Result(T value)
         : _result(std::move(value))
-        , _isError(false){};
-    Result(int code, QString str)
-        : _error({ std::move(str), code })
+        , _isError(false)
+    {
+    }
+    // TODO: This doesn't work if T and Error are too similar
+    Result(Error error)
+        : _error(std::move(error))
         , _isError(true)
     {
     }
+
     ~Result()
     {
         if (_isError)
@@ -62,14 +61,31 @@ public:
         ASSERT(!_isError);
         return std::move(_result);
     }
-    QString errorMessage() const
+    const Error &error() const &
     {
         ASSERT(_isError);
-        return _error.string;
+        return _error;
     }
-    int errorCode() const
+    Error error() &&
     {
-        return _isError ? _error.code : 0;
+        ASSERT(_isError);
+        return std::move(_error);
+    }
+};
+
+namespace detail {
+struct OptionalNoErrorData{};
+}
+
+template <typename T>
+class Optional : public Result<T, detail::OptionalNoErrorData>
+{
+public:
+    using Result<T, detail::OptionalNoErrorData>::Result;
+
+    Optional()
+        : Optional(detail::OptionalNoErrorData{})
+    {
     }
 };
 
--- a/src/common/vfs.cpp
+++ b/src/common/vfs.cpp
@@ -19,6 +19,9 @@
 #include "vfs.h"
 #include "plugin.h"
 
+#include <QPluginLoader>
+#include <QLoggingCategory>
+
 using namespace OCC;
 
 Vfs::Vfs(QObject* parent)
@@ -69,7 +72,7 @@ bool OCC::isVfsPluginAvailable(Vfs::Mode mode)
     auto name = modeToPluginName(mode);
     if (name.isEmpty())
         return false;
-    return PluginLoader().load("vfs", name);
+    return QPluginLoader(pluginFileName("vfs", name)).load();
 }
 
 Vfs::Mode OCC::bestAvailableVfsMode()
@@ -82,10 +85,34 @@ Vfs::Mode OCC::bestAvailableVfsMode()
     return Vfs::Off;
 }
 
+Q_LOGGING_CATEGORY(lcPlugin, "plugins", QtInfoMsg)
+
 Vfs *OCC::createVfsFromPlugin(Vfs::Mode mode, QObject *parent)
 {
     auto name = modeToPluginName(mode);
     if (name.isEmpty())
         return nullptr;
-    return qobject_cast<Vfs *>(PluginLoader().create("vfs", name, parent));
+
+    auto pluginPath = pluginFileName("vfs", name);
+    QPluginLoader loader(pluginPath);
+    auto plugin = loader.instance();
+    if (!plugin) {
+        qCWarning(lcPlugin) << "Could not load plugin" << pluginPath << loader.errorString();
+        return nullptr;
+    }
+
+    auto factory = qobject_cast<PluginFactory *>(plugin);
+    if (!factory) {
+        qCWarning(lcPlugin) << "Plugin" << pluginPath << "does not implement PluginFactory";
+        return nullptr;
+    }
+
+    auto vfs = qobject_cast<Vfs *>(factory->create(parent));
+    if (!vfs) {
+        qCWarning(lcPlugin) << "Plugin" << pluginPath << "does not create a Vfs instance";
+        return nullptr;
+    }
+
+    qCInfo(lcPlugin) << "Created VFS instance from plugin" << pluginPath;
+    return vfs;
 }
--- a/src/common/vfs.cpp
+++ b/src/common/vfs.cpp
@@ -18,6 +18,7 @@
 
 #include "vfs.h"
 #include "plugin.h"
+#include "version.h"
 
 #include <QPluginLoader>
 #include <QLoggingCategory>
@@ -74,6 +75,8 @@ static QString modeToPluginName(Vfs::Mode mode)
     return QString();
 }
 
+Q_LOGGING_CATEGORY(lcPlugin, "plugins", QtInfoMsg)
+
 bool OCC::isVfsPluginAvailable(Vfs::Mode mode)
 {
     if (mode == Vfs::Off)
@@ -81,7 +84,30 @@ bool OCC::isVfsPluginAvailable(Vfs::Mode mode)
     auto name = modeToPluginName(mode);
     if (name.isEmpty())
         return false;
-    return QPluginLoader(pluginFileName("vfs", name)).load();
+    auto pluginPath = pluginFileName("vfs", name);
+    QPluginLoader loader(pluginPath);
+
+    auto basemeta = loader.metaData();
+    if (basemeta.isEmpty() || !basemeta.contains("IID")) {
+        qCDebug(lcPlugin) << "Plugin doesn't exist" << pluginPath;
+        return false;
+    }
+    if (basemeta["IID"].toString() != "org.owncloud.PluginFactory") {
+        qCWarning(lcPlugin) << "Plugin has wrong IID" << pluginPath << basemeta["IID"];
+        return false;
+    }
+
+    auto metadata = basemeta["MetaData"].toObject();
+    if (metadata["type"].toString() != "vfs") {
+        qCWarning(lcPlugin) << "Plugin has wrong type" << pluginPath << metadata["type"];
+        return false;
+    }
+    if (metadata["version"].toString() != MIRALL_VERSION_STRING) {
+        qCWarning(lcPlugin) << "Plugin has wrong version" << pluginPath << metadata["version"];
+        return false;
+    }
+
+    return true;
 }
 
 Vfs::Mode OCC::bestAvailableVfsMode()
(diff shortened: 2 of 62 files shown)


src/gui/folder.cpp
```diff
@@ -533,9 +584,26 @@ void Folder::downloadVirtualFile(const QString &_relativepath)
...
+    }
+    if (!enabled && _definition.virtualFilesMode != Vfs::Off) {
+        ENFORCE(_vfs);
+
+        // TODO: Must wait for current sync to finish!
+        SyncEngine::wipeVirtualFiles(path(), _journal, _vfs);
+
+        _vfs->stop();
+        _vfs->unregisterFolder();
+        delete _vfs;
```
ogoffart: and _vfs is left dangling.  Is that not a problem?
Maybe use QScopedPointer for vfs

src/gui/folder.cpp
```diff
@@ -1113,9 +1209,11 @@ void FolderDefinition::save(QSettings &settings, const FolderDefinition &folder)
     settings.setValue(QLatin1String("targetPath"), folder.targetPath);
     settings.setValue(QLatin1String("paused"), folder.paused);
     settings.setValue(QLatin1String("ignoreHiddenFiles"), folder.ignoreHiddenFiles);
-    settings.setValue(QLatin1String("usePlaceholders"), folder.useVirtualFiles);
     settings.setValue(QLatin1String(versionC), maxSettingsVersion());
 
+    settings.setValue(QStringLiteral("virtualFilesMode"), Vfs::modeToString(folder.virtualFilesMode));
+    settings.remove(QLatin1String("usePlaceholders")); // deprecated key
```
ogoffart: How about downgrading?  Should we not leave this key on for the suffix mode?

src/gui/folder.h
```diff
@@ -101,7 +107,11 @@ class Folder : public QObject
     Q_OBJECT
 
 public:
-    Folder(const FolderDefinition &definition, AccountState *accountState, QObject *parent = 0L);
+    /** Create a new Folder
+     *
+     * The vfs instance will be parented to this.
+     */
+    Folder(const FolderDefinition &definition, AccountState *accountState, Vfs *vfs, QObject *parent = 0L);
```
ogoffart: take vfs by std::unique_ptr  to document it takes ownershipt.

src/gui/folderman.cpp
```diff
@@ -949,21 +976,22 @@ Folder *FolderMan::addFolder(AccountState *accountState, const FolderDefinition
             break;
         }
     }
-    folder->setSaveBackwardsCompatible(oneAccountOnly);
 
     if (folder) {
```
ogoffart: Folder can't be null

src/gui/socketapi.cpp
```diff
@@ -608,52 +608,51 @@ void SocketApi::copyUrlToClipboard(const QString &link)
...
             continue;
+        auto journal = data.folder->journalDb();
+        auto markForDehydration = [&](SyncJournalFileRecord rec) {
+            if (rec._type != ItemTypeFile)
+                return;
+            rec._type = ItemTypeVirtualFileDehydration;
+            journal->setFileRecord(rec);
+            toSync.insert(data.folder);
+        };
+
```
ogoffart: We also need to add the path in the list of path we need to visit.  Otherwise the discovery may skip them.   I think.

src/libsync/discovery.cpp
```diff
@@ -85,8 +85,7 @@ void ProcessDirectoryJob::process()
     for (auto &e : _localNormalQueryEntries) {
         // Remove the virtual file suffix
         auto name = e.name;
-        if (hasVirtualFileSuffix(name)) {
-            e.isVirtualFile = true;
+        if (e.isVirtualFile && isVfsWithSuffix()) {
```
ogoffart: maybe change chopVirtualFileSuffix to return bool, so you can do  
`if (e.isVirtualFile && chopVirtualFileSuffix(name))`

since this pattern is repeated a lot.

Likewise the addVirtualFileSuffix could be a no-op if the mode does not have the suffix

src/libsync/propagatedownload.cpp
```diff
@@ -344,17 +359,30 @@ void PropagateDownloadFile::start()
...
+    auto vfs = syncOptions._vfs;
+
     // For virtual files just create the file and be done
+    if (_item->_type == ItemTypeVirtualFileDehydration) {
+        _item->_type = ItemTypeVirtualFile;
+        // TODO: Could dehydrate without wiping the file entirely
+        auto fn = propagator()->getFilePath(_item->_file);
+        qCDebug(lcPropagateDownload) << "dehydration: wiping base file" << fn;
+        QFile::remove(fn);
+        propagator()->_journal->deleteFileRecord(_item->_file);
```
ogoffart: I'd delete the file record before deleting the file, so then if the client crashes here, the file won't be removed on the server.

Actually, the test `test/testdatabaseerror.cpp`  could test that if we make it test virtual files (which it currently does not test)

src/libsync/vfs/suffix/vfs_suffix.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+bool VfsSuffix::updateMetadata(const QString &filePath, time_t modtime, quint64, const QByteArray &, QString *)
+{
+    FileSystem::setModTime(filePath, modtime);
+    return true;
+}
+
+void VfsSuffix::createPlaceholder(const QString &syncFolder, const SyncFileItemPtr &item)
+{
+    // NOTE: Other places might depend on contents of placeholder files (like csync_update)
```
ogoffart: csync_update lo longer exist.  maybe time to update the comment.

src/libsync/vfs/suffix/vfs_suffix.h
```diff
@@ -0,0 +1,62 @@
...
+
+namespace OCC {
+
+class VfsSuffixPrivate;
+
+class VfsSuffix : public Vfs
+{
+    Q_OBJECT
+    Q_DECLARE_PRIVATE(VfsSuffix)
+    const QScopedPointer<VfsSuffixPrivate> d_ptr;
```
ogoffart: Why do we need a VfsSuffixPrivate ?

It seems we don't.

test/testsyncvirtualfiles.cpp
```diff
@@ -46,6 +47,25 @@ void triggerDownload(FakeFolder &folder, const QByteArray &path)
...
+        return;
+    record._type = ItemTypeVirtualFileDehydration;
+    journal.setFileRecord(record);
+    journal.avoidReadFromDbOnNextSync(record._path);
+}
+
+SyncOptions vfsSyncOptions()
+{
+    SyncOptions options;
+    options._vfs = createVfsFromPlugin(Vfs::WithSuffix, nullptr);
```
ogoffart: I bet this is leaking.
Maybe a QSharedPointer should actually be appropriate

test/testsyncvirtualfiles.cpp
```diff
@@ -408,13 +418,12 @@ private slots:
         QVERIFY(!dbRecord(fakeFolder, "A/a1.owncloud").isValid());
     }
 
-    // Check what might happen if an older sync client encounters virtual files
-    void testOldVersion1()
+    // Check what happens if vfs mode is disabled
+    void testSwitchOfVfs()
     {
         QSKIP("Does not work with the new discovery because the way we simulate the old client does not work");
```
ogoffart: Maybe this test can be unskipped?   or removed?

src/gui/socketapi.cpp
ckamm: Yes!

Thinking about it, I assumed that "_invalid_" etags and `avoidReadFromDbOnNextSync` would also encompass local discovery, but at least for the new sync algorithm I don't think it does. (not sure whether it's true for the old code) What do you think about that?

ogoffart: avoidReadFromDbOnNextSync would work.  But we also don't need to do server discovery (slow).
folder->_localDiscoveryTracker->addTouchedPath should also work.


Ideally, in the future, we can actually ask the sync engine to just do these files immediatly.

src/libsync/vfs/suffix/vfs_suffix.cpp
ckamm: Can be removed now since the only place that depends on it is a few lines down, `isDehydratedPlaceholder`.

src/libsync/discovery.cpp
ckamm: I thought the same thing and looked at it again. I prefer the explicitness of checking for suffix-vfs. I feel I'd need to rename to `maybeChopSuffix` and `maybeAddSuffix` otherwise and that makes the code less clear.

test/testsyncvirtualfiles.cpp
ckamm: I didn't mind about that since it's in a test. Do you think it matters?

ogoffart: I sometimes use valgrind or other leak detector on tests to make sure the code does not leak.

But this can be fixed later.

src/common/result.h
```diff
@@ -38,12 +33,16 @@ class Result
 public:
     Result(T value)
         : _result(std::move(value))
-        , _isError(false){};
-    Result(int code, QString str)
-        : _error({ std::move(str), code })
+        , _isError(false)
+    {
+    }
+    // TODO: This doesn't work if T and Error are too similar
```
ogoffart: Yeah, the std::expected proposal includes a std::unexpected type for this purpose.
But other possibilities can be to have a  makeError(...)  function,  or a  Result::fromError

Does not have to be done in this change

src/common/vfs.cpp
```diff
@@ -72,7 +72,7 @@ bool OCC::isVfsPluginAvailable(Vfs::Mode mode)
     auto name = modeToPluginName(mode);
     if (name.isEmpty())
         return false;
-    return PluginLoader().load("vfs", name);
+    return QPluginLoader(pluginFileName("vfs", name)).load();
```
ogoffart: Is there not an option to do that without actually loading the plugin?  Perhaps querying the metadata.
Also, don't we need a mirall version check?

src/common/vfs.h
```diff
@@ -30,72 +33,138 @@ class VfsPrivate;
...
+    static Optional<Mode> modeFromString(const QString &str);
 
 public:
     Vfs(QObject* parent = nullptr);
     virtual ~Vfs();
 
     virtual Mode mode() const = 0;
 
-    // For WithSuffix modes: what's the suffix (including the dot)?
+    /// For WithSuffix modes: what's the suffix (including the dot)?
```
ogoffart: Is that a question?

src/common/vfs.h
```diff
@@ -30,72 +33,138 @@ class VfsPrivate;
...
 };
 
+/// Check whether the plugin for the mode is available.
 OCSYNC_EXPORT bool isVfsPluginAvailable(Vfs::Mode mode);
+
+/// Return the best available VFS mode.
 OCSYNC_EXPORT Vfs::Mode bestAvailableVfsMode();
-OCSYNC_EXPORT Vfs *createVfsFromPlugin(Vfs::Mode mode, QObject *parent);
+
+/// Create a VFS instance for the mode, returns nullptr on failure.
```
ogoffart: Maybe instead of returning nullptr on faillure, we should return the VFS instance for Vfs::Off.
Yes, i know currently this does not exist, but maybe it should.
This way we do not need to check if vfs is null all the time.

src/csync/vio/csync_vio_local_unix.cpp
```diff
@@ -141,8 +141,11 @@ std::unique_ptr<csync_file_stat_t> csync_vio_local_readdir(csync_vio_handle_t *d
   }
 
   // Override type for virtual files if desired
-  if (vfs)
+  if (vfs) {
+      // Directly modifiest file_stat->type.
```
ogoffart: spelling

src/gui/folder.cpp
```diff
@@ -1212,7 +1237,8 @@ void FolderDefinition::save(QSettings &settings, const FolderDefinition &folder)
     settings.setValue(QLatin1String(versionC), maxSettingsVersion());
 
     settings.setValue(QStringLiteral("virtualFilesMode"), Vfs::modeToString(folder.virtualFilesMode));
-    settings.remove(QLatin1String("usePlaceholders")); // deprecated key
+    if (folder.virtualFilesMode == Vfs::WithSuffix)
+        settings.setValue(QLatin1String("usePlaceholders"), true); // to support older versions
```
ogoffart: and set it to false otherwise?

src/common/vfs.cpp
ckamm: Yeah, I think it makes a lot of sense to ensure the plugin and client version match exactly, and that can be done with metadata without loading the plugin. I'll check it out.

src/common/vfs.h
ckamm: I had a similar thought and mentioned it in one of the commit messages. Having a VfsOff instance and being guaranteed that vfs!=nullptr would make this less error-prone. Will do.

ckamm:
Still needs improvements to plugin metadata checking and handling placeholder files from 2.5 when `usePlaceholders=false` (2.5. probably has to be treated as always in "WithSuffix" mode, but the option controls whether new files are placeholders or not)

ogoffart:
The remaining problems can still be fixed before the plugin is actually available.

PR #6890, opened 2018-11-21
SSLButton: do not hardcode the background color

ogoffart:
Reported in https://github.com/nextcloud/desktop/issues/871


guruz:
works on macOS.

Issue #6815, opened 2018-10-16
[Virtual Filesystem] Add option on folder level for downloading and deleting files


DangerSchwob:
Please add a option how the client handle files in different folders.

The default currently is, download the file if the user download it and from than update the file also.

Please also add 2 other options that can be set on folder level:
1.) Always download files in this folder
2.) Temporary files, if the file is not used for X days or it is updated on the server, replace file with virtual file

michaelstingl:
@DangerSchwob Thanks for your feedback!

Workaround for "1.) Always download …" could be multiple sync folders. (related: https://github.com/owncloud/client/pull/6744, planned for 2.5.1)

michaelstingl:
First implementation for `1)` planned for next client release…

ckamm:
Baseline proposal:
- Each directory can have a pin state: unspecified (propagates parent), available-locally, on-demand
- When vfs is enabled the folder sync root defaults to on-demand
- If a new remote file is propagated check the parents for the closest set pin state: Make it virtual if it's on-demand. Download it if the state is available-locally.
- Pin state can be set via context menu actions on folders

Further suggestions:
- Pin state is visible via icons on folders
- Prettier context menu ui (probably want submenus, checked items, icons etc, which necessitate changes in all shell integrations - ideally would be generic, similar to the current action list)

michaelstingl:
Did some brainstorming with @pmaier1 this afternoon… (no conflicts with @ckamm 's proposal from @ckamm except the wording my pov)

### VFS-options for a folder that contains only "Online only" files

![folder - online only](https://user-images.githubusercontent.com/214010/48505274-f32d7f00-e846-11e8-89b3-6903db3d5708.png)

### VFS-options for a folder that contains downloaded and "Online only" files

![folder - mixed](https://user-images.githubusercontent.com/214010/48505275-f32d7f00-e846-11e8-8563-aca152665ee1.png)

### VFS-options for multiple files that are all downloaded

![files - multi synced](https://user-images.githubusercontent.com/214010/48505278-f3c61580-e846-11e8-84c2-68d9b59b3449.png)

### VFS-options for multiple files that are all "Online only"

![files - multi online only](https://user-images.githubusercontent.com/214010/48505279-f3c61580-e846-11e8-8627-331059a71fb7.png)

### VFS-options for multiple files where some are downloaded and others are "Online only" files

[...]

PR #6898, opened 2018-11-27
Directory level control of whether new files are virtual or not

ckamm:
For #6815

This is targeted for 2.5 in the ticket but I'd like to avoid that if possible since it's a new feature.

Warts:
- Maybe `PinState` should be `NewFilesAreVirtualFlag` or something? With options `NewFilesAreVirtual` and `NewFilesAreDownloaded`? I have some hopes of replacing the action-flag like "ItemTypeVirtualFileDehydration" and "ItemTypeVirtualFileDownload" with something orthogonal to the item type.
- The root folder's state isn't stored in the db (the root has no entry). It's stored in `SyncOptions` instead.
- SocketAPI actions can't control the root's state.
- The root "newFilesAreVirtual" state controls whether selective sync is shown or not. That's now weird, since the flag may be set even though all subfolders are `AlwaysLocal`. Or it's not set even though all subfolders are `OnlineOnly`. Selective sync has lots of overlap with vfs and we need to figure out what to do about it.

### DIFF
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -463,6 +463,15 @@ bool SyncJournalDb::checkConnect()
         return sqlFail("Create table datafingerprint", createQuery);
     }
 
+    // create the flags table.
+    createQuery.prepare("CREATE TABLE IF NOT EXISTS flags ("
+                        "path TEXT PRIMARY KEY,"
+                        "pinState INTEGER"
+                        ");");
+    if (!createQuery.exec()) {
+        return sqlFail("Create table flags", createQuery);
+    }
+
     // create the conflicts table.
     createQuery.prepare("CREATE TABLE IF NOT EXISTS conflicts("
                         "path TEXT PRIMARY KEY,"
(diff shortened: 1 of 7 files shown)


ogoffart:
I don't think we should add new features in 2.5 at this state.
Merging in 2.6 will be a bit of work since he code has changed a lot.

ogoffart:
Instead of adding a new column, have you considered:

* using the selectivesync table with a new type?  (if not, maybe actually this table should be removed and added as a flag)

* reusing the type column  (no, probably not such a good idea)

I'm just worried that the sqlite database size becomes too big, but maybe a integer column is not a problem.

src/csync/csync_update.cpp
```diff
@@ -481,12 +481,19 @@ static int _csync_detect_update(CSYNC *ctx, std::unique_ptr<csync_file_stat_t> f
...
-              && fs->instruction == CSYNC_INSTRUCTION_NEW
-              && fs->type == ItemTypeFile) {
-              fs->type = ItemTypeVirtualFile;
-              fs->path.append(ctx->virtual_file_suffix);
+          // Turn new remote files into virtual files if the directory desires it.
+          if (fs->instruction == CSYNC_INSTRUCTION_NEW && fs->type == ItemTypeFile) {
+              using OCC::PinState;
+
+              // Get the path's pinstate and anchor to the root option
+              auto pinState = ctx->statedb->pinStateForPath(fs->path);
```
ogoffart: Can we avoid making a sqlite request for every new file, by keeping the state.  I guess this should be easy with the new discovery.

ckamm:
@ogoffart What do you think about making the "selectivesync" table a generic flags table?

src/csync/csync_update.cpp
ckamm: I agree, but would much prefer to do it only in the 2.6 codebase.

lazawan:
i have just tested the new "feature" on Macos 10.14 and Windows 10 and it works but the option "if the file is not used for X days or it is updated on the server, replace file with virtual file" is missing.
there is another thing : files have under owncloud-menu a few options : 
-share
-copy public link to clipboard
-copy private link to clipboard
-open in browser
-keep updated locally..download if necessary 

when one clicks "keep updated locally..download if necessary" file will be downloaded and when one shows the menu again ,then appears an option " make available on demand , clear local data " + disappears the option "keep updated locally..download if necessary ".

but folders have problem with switching between the two options and both of them are in the menu . on my opinion ,it should be one option in the menu

either folder is virtual >> one wants to download it 
or folder is already downloaded >> one wants to make it virtual 

screenshot from 2018-11-28 14-05-47

ogoffart:
Looks good to me, appart from the discovery step.
But now this should be merged in master.

ckamm:
@ogoffart I've rebased it to master and ensured there are no excessive queries during discovery. Please re-review!

I'm not convinced the caching is worth it. Would you agree that querying the pin state once for each directory is acceptable?

ckamm:
@lazawan Thanks for testing, btw, I'm glad it worked for you! This isn't yet the final state and more work will follow.

Issue #6451, opened 2018-04-11
Drop QtWebKit from build dependencies


SamuAlfageme:
After having https://github.com/owncloud/client/pull/5668 implemented and https://marketplace.owncloud.com/apps/oauth2 gaining more and more adoption, we're in the right track (platform-wise) to drop the requirement of shipping a full-blown browser together with the client. 

Today I was looking at the differences and benefits we'll get in the e.g. `.app` bundles on install time and QtWebKit bloats the size in disk more than 50% (45.7 MB build with `-DNO_SHIBBOLETH=1` vs. 92.2 MB wo/ the build flag).

jnweiger:
Will master build without QtWebKit now?

ogoffart:
The packages just need to update the package to remove this dependences.

guruz:
I'll check macOS and NSIS now

guruz:
NSIS and macOS merged https://github.com/owncloud/client/pull/7158
MSI and Linux/Drone pending.
Linux OBS won't do.

guruz:
MSI is done

guruz:
@dschmidt @jnweiger How is this with the Linux/Drone builds? Done already so we can close here?

jnweiger:
2.6.0 will still be done with the openSUSE Build service. I'll remove the dependency and see what happens.

jnweiger:
https://build.opensuse.org/package/show/isv:ownCloud:desktop:daily:2.6/owncloud-client
is doing a rebuild without qtwebkit.

jnweiger:
https://gitea.owncloud.services/client/build-linux/commit/32e709e29dd9d136067575c524bbb86ed9791516

https://gitea.owncloud.services/ownbrander/scripting/commit/a55cf01cce3f9c1de098cfded23177d081790050

done. 

## Tested ownCloud version 2.6.0daily20190604 (build 914) on Linux Mint 19
login with web token and password works. syncs fine. No more ocqt*qebkit packages on my system!

PR #6899, opened 2018-11-27
Remove Shibboleth support

ogoffart:
Issue #6451

This also removes the webkit dependency


ogoffart:
@michaelstingl   We continue with the plan to remove shibboleth in 2.6, right?

ogoffart:
oops, i merged this by mistake too eagerly when trying to merge 2.5 in master. (I forgot i still had this commit on top of my branch)
Well, we can still revert if we need to.

Issue #434, opened 2013-03-12
Add "blacklisted_files" to sync-exclude.lst


LukasReschke:
With oC 5.0 we gave the admin the possibility to configure the blacklisted files (e.g. to sync .htaccess or deny other files which was an often asked feature)
(https://github.com/owncloud/core/blob/master/config/config.sample.php#L35)

Could we sync this list with the client by exposing the list via status.php - what do you think @dragotin @danimo?

Additionally we should remove the hardcoded .htaccess from sync-exclude.lst

nickvergessen:
Could be done via capabilities?

ogoffart:
capabilities, or any other api call.
(you probably don't want the reply to capabilities to become too big)

guruz:
> Additionally we should remove the hardcoded .htaccess from sync-exclude.lst

Why? because now it comes via blacklisted_files capabilities?

nickvergessen:
> Why? because now it comes via blacklisted_files capabilities?

Yes it would then come from the server. If the server is not apache but nginx, there is no real reason to block `.htaccess`, it is just like any other hidden file then.

nickvergessen:
value has been added to capabilities for 9.1+

HanaGemela:
# Tested
Tested on macOS with 2.6.0-daily20190426 (build 11674)
sync-exclude.lst is synchronized with a client
sync-exclude.lst doesn't include .htaccess

PR #6905, opened 2018-11-29
Discovery: Handle the blacklistFiles from the server capabilities

ogoffart:
Issue #434

Ideally one could add the blacklist to the exlucde reggexp, but this
is simpler


src/libsync/discovery.h
```diff
@@ -115,8 +115,10 @@ class ProcessDirectoryJob : public QObject
      */
     void process();
 
-    // return true if the file is excluded
-    bool handleExcluded(const QString &path, bool isDirectory, bool isHidden, bool isSymlink);
+    // return true if the file is excluded.
+    // path is the full relative path of the file. localName is the base name of the local entry.
+    bool handleExcluded(const QString &path, const QString &localName , bool isDirectory,
```
ckamm: minor: stray space

Issue #6358, opened 2018-02-14
Convert cmocka-based tests to Qt test framework.


guruz:
* less dependencies
 * easier running of tests on all platforms

PR #6910, opened 2018-11-30
Convert cmocka-based tests to Qt test framework

ogoffart:
There is still a few test to port.
See individual commit messages for more info.

issue #6358


ogoffart:
The test is failling on windows:

´´´
16202FAIL!  : TestExcludedFiles::check_csync_excluded() Compared values are not the same
16203C:\projects\client\test\testexcludedfiles.cpp(203) : failure location
16204FAIL!  : TestExcludedFiles::check_csync_excluded_traversal() Compared values are not the same
16205C:\projects\client\test\testexcludedfiles.cpp(347) : failure location
```

That's because we ignore invalid char on windows and the tests were never ran on windows (no cmocka there)

Issue #6157, opened 2017-11-13
Forbidden web page when using invalid URL in setup dialog


PVince81:
### Steps
1. Setup Apache2 with ownCloud under http://localhost/owncloud
1. Setup desktop client and enter "http://localhost" (note that it's the wrong path)
1. Click "next"

### Expected
User-friendly error message

### Actual
Popup appears with 403 error page inside. I can close it to go back.
![screenshot_20171113_110915](https://user-images.githubusercontent.com/277525/32720199-167e6726-c863-11e7-83ff-1cd37d4ffa48.png)

### Versions
openSUSE Tumbleweed
plasma5-workspace-5.11.2-1.2.x86_64
owncloud-client-2.4.0~alpha1-8649.1.x86_64

ckamm:
This behavior should exist since 2.3.0 with 679ac0d26a6607a1d98024d829c0f71cc715d480: If the server replies with text the client displays that to the user. 

It would be nice
1) If it were clearer to the user that this is what's happening
2) The "Ok" button that's supposed to be there were visible

PVince81:
Maybe the client should first ping status.php of the given URL to decide whether it's even worth opening a web view ?

ckamm:
It looks like the intention was to show the 400 replies for missing client certs or the ip needing to be added as a "trusted domain" for the instance. We could filter out 404 which would cover most arbitrary urls. @guruz ?

guruz:
@PVince81 It's a Qt message box, not a web view. it is just a dialog that renders some HTML using Qt's Label HTML implementation. :-)

> The "Ok" button that's supposed to be there were visible

Will that be visible now?

> We could filter out 404 which would cover most arbitrary urls

What do you want to do for 404? @PVince81 got a 403

ckamm:
@guruz I have no idea why the "ok" button isn't visible for @PVince81, it was always there in my tests. Maybe specific html can hide it somehow?

404/403: Right. Do you think my patch 
aa9a386  (not yet merged) is sufficient here?

guruz:
.oO(Is @PVince81 getting some shibboleth web view?)

Yes, sufficient. let's re-evaluate this for 2.5 if there is any issues

ckamm:
@guruz ok, merged simple "info" patch to 2.4

PVince81:
Maybe I don't see the OK button of highdpi issues ? See my other ticket https://github.com/owncloud/client/issues/6156

TheOneRing:
[...]

PR #6914, opened 2018-12-03
Wizard: do not display HTML error page from the server in a message box

ogoffart:
It makes no sense to display arbitrary html in a message box.
It is ugly and unfriendly.
The wizard itself already show the error in red.

In case of invalid certificate/ missing clientside certificate, there
is already a dialog that suggest using client side certificates.

Issue #6157

### DIFF
--- a/src/gui/owncloudsetupwizard.cpp
+++ b/src/gui/owncloudsetupwizard.cpp
@@ -272,23 +272,6 @@ void OwncloudSetupWizard::slotNoServerFound(QNetworkReply *reply)
     }
     bool isDowngradeAdvised = checkDowngradeAdvised(reply);
 
-    // If a client cert is needed, nginx sends:
-    // 400 "<html>\r\n<head><title>400 No required SSL certificate was sent</title></head>\r\n<body bgcolor=\"white\">\r\n<center><h1>400 Bad Request</h1></center>\r\n<center>No required SSL certificate was sent</center>\r\n<hr><center>nginx/1.10.0</center>\r\n</body>\r\n</html>\r\n"
-    // If the IP needs to be added as "trusted domain" in oC, oC sends:
-    // https://gist.github.com/guruz/ab6d11df1873c2ad3932180de92e7d82
-    if (resultCode != 200 && contentType.startsWith("text/")) {
-        // FIXME: Synchronous dialogs are not so nice because of event loop recursion
-        // (we already create a dialog further below)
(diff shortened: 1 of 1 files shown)


ogoffart:
This dialog should never have been there in the first place. It is out of place, ugly, and break the wizard workflow. The wizard already reports the error.  

See what happens if you type `owncloud.org` in the URL
screenshot_20181203_142825

`qt.io`

screenshot_20181203_142926

(In that case, you don't even see the close button since the dialog is too big to fit in the screen)

In some case, there is no error (example, github.com)

screenshot_20181203_143201

When the error is already shown in the wizard

screenshot_20181203_143458

guruz:
I agree it looks ugly. But seeing a HTML at least makes people understand they're not in the "DAV-compatible" URL.

> When the error is already shown in the wizard 

Funny, the English translation of that German error message in your screenshot is "cannot be produced".  Is there a "Die Verbindung"- missing at the beginning?

In any case, that error display is not helping. 

As a solution instead: Can we make the red unuseful error message more useful?
* We could check the content-type and based on that make a recommendation ("Returned a HTML page but we expected a WebDAV response")
* We could have a hardcoded check for what nginx and Apache return when there is no proper client certificate sent?
* if we know it is an error as HTML, try to extract it from <title> or wherever? https://stackoverflow.com/questions/33560256/custom-nginx-error-page-for-the-ssl-certificate-error

>  _ocWizard->displayError(msg, isDowngradeAdvised);

.. and just append that created message there?

guruz:
See here for an example how it is useful to see the HTML error: https://github.com/owncloud/client/issues/6940#issue-393740993

ogoffart:
Sure, there could be some information extracted. But we already give a big hint for the TLS client certificate.

guruz:
FYI this is what I see with nginx, nice error for which I'd like to have the dialog or an equivalent display. Please don't remove functionality.

Forgot the 's' in https:
Screen Shot 2019-06-04 at 20 39 27

Forgot the client certificate (and will be promoted to add it):
Screen Shot 2019-06-04 at 20 39 40

ogoffart:
> If you don't like the dialog, There are ways to make it nicer, e.g. try to extract more text from it, only take part of the text etc etc.

That's right, I don't like the dialog, and not having it is better than having it in the current state.
Unless you volunteer to make it better, i suggest we merge this patch.

guruz:
OK, let's do something nicer in 2.6.1

PR #6917, opened 2018-12-06
Switch to standard opt-out BUILD_TESTING instead of opt-in UNIT_TESTING

dschmidt:
Compare https://cmake.org/cmake/help/v3.0/module/CTest.html
Craft automatically handles BUILD_TESTING, so we don't need to handle it
in our own blueprint.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -29,7 +29,7 @@ pipeline:
     commands:
       - mkdir clang-build
       - cd clang-build
-      - cmake -GNinja -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_BUILD_TYPE="Debug" -DUNIT_TESTING=1 ..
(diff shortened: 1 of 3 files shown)


dschmidt:
Copying from chat:

> right now you need to opt in via -DUNIT_TESTING=ON, standard would be opt-out via -DBUILD_TESTING=OFF
> i can maybe keep it compatible with the old option - but only with some spaghetti CMake code.. do you care about it at all?
> I'd argue that BUILD_TESTING is standard and most package macros should include it already (with whatever default a distro wants), so it shouldn't be an issue for packagers. And building them by default makes it harder to forget adjusting them when breaking apis

ogoffart:
There are still references to UNIT_TESTING in src/csync/DefineOptions.cmake and .drone.yml

dschmidt:
Good catch. Thanks. Although the default should be ON, I wanted to keep it explicit in .drone.yml.

ogoffart:
Don't we need a `option` command to define it?

PR #6919, opened 2018-12-06
Add ownCloud icon to Dolphin context menu

nicolasfella:
Most actions in Dolphin's context menu have an icon. 

image

### DIFF
--- a/shell_integration/dolphin/ownclouddolphinactionplugin.cpp
+++ b/shell_integration/dolphin/ownclouddolphinactionplugin.cpp
@@ -93,6 +93,7 @@ public:
 
         auto menuaction = new QAction(parentWidget);
         menuaction->setText(helper->contextMenuTitle());
+        menuaction->setIcon(QIcon::fromTheme(QStringLiteral("owncloud")));
         menuaction->setMenu(menu);
         return { menuaction };
     }


ogoffart:
Thanks for your merge request.  I think we should indeed show the icon there.
However, I don't think we can simply use "owncloud" because of the branding. We should query the socketapi to get the icon the same way we ask for the title or for the menus.

dschmidt:
@ogoffart What happens if several branded/branded and unbranded clients are installed? Do they all load independently or is only one loaded and handles whichever client? 

If all are loaded and only handle their respective client, we could probably also just use APPLICATION_ICON_NAME to hardcode it in the extension:
https://github.com/owncloud/client/blob/master/CMakeLists.txt#L154 https://github.com/owncloud/client/blob/master/mirall.desktop.in#L8 

In fact it looks like mirall.desktop.in uses APPLICATION_EXECUTABLE although it should use APPLICATION_ICON_NAME (which is based on APPLICATION_SHORTNAME (!)) -> 
https://github.com/owncloud/client/blob/master/src/gui/CMakeLists.txt#L235


(How) does this relate to https://github.com/owncloud/client/pull/6921?

guruz:
> We should query the socketapi to get the icon the same way we ask for the title or for the menus.

@nicolasfella You can find inspiration on how to do this in 85938ab1f15c78d86aa80ba1853674810a3afe9e

guruz:
@nicolasfella Any news? :)

TheOneRing:
Closing abandoned pr

PR #6921, opened 2018-12-07
Nautilus Shell Extension: separator and submenu

ogoffart:
This is just a WIP RFC, before doing it on other plugins.

This is an extension to the protocol to get separator and submenu

The protocol looks like this:

```
GET_MENU_ITEMS:BEGIN
MENU_ITEM:SHARE:d:Share…
SUBMENU_BEGIN::Private Links
MENU_ITEM:COPY_PRIVATE_LINK::Copy
MENU_ITEM:OPEN_PRIVATE_LINK::Open in browser
SUBMENU_END:
SEPARATOR:
MENU_ICON:/usr/share/icons/hicolor/32x32/apps/owncloud.png
MENU_ITEM:DOWNLOAD_VIRTUAL_FILE::Download
GET_MENU_ITEMS:END
```

It adds a few command:
`SEPARATOR` is self explanatory.
`MENU_ICON` will allow to set an icon to the next item
`SUBMENU_BEGIN:<flags>:<text>` will add a submenu, and all following items are
int this submenu. the submenu ends with `SUBMENU_END`
It is possible to have submenu within submenu.

I also wanted to add checkable or checked flags, but it does not seem that nautilus supports that.
Actually, even the icon does not work (i don't see it in my nautilus although the API allows it)

### DIFF
--- a/shell_integration/nautilus/syncstate.py
+++ b/shell_integration/nautilus/syncstate.py
@@ -238,7 +238,14 @@ class MenuExtension_ownCloud(GObject.GObject, Nautilus.MenuProvider):
         done = False
         start = time.time()
         timeout = 0.1 # 100ms
-        menu_items = []
+        empty = True
+
+        item_owncloud = Nautilus.MenuItem(
+            name='IntegrationMenu', label=self.strings.get('CONTEXT_MENU_TITLE', appname))
+        menu = Nautilus.Menu()
+        item_owncloud.set_submenu(menu)
+        menu_stack = [ menu ]
+        icon = "";
         while not done:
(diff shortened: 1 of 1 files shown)


ogoffart:
>  I agree about not adding "checked" to the socket api if not all platforms support it.

We could still add it. 
On nautilus, and other platform that do not support checkboxes in menu, we could simply add the  ☐ and ☑ character in front of the item names.

TheOneRing:
I'd suggest to use the json based api for new features and send icons as payload instead of a file link, with the icons being located in the qrc the file path approach only works on linux.

PR #6922, opened 2018-12-07
Fix KDEInstallDirs deprecation warnings

dschmidt:
This bugged me for quite some time :-D

### DIFF
--- a/shell_integration/dolphin/CMakeLists.txt
+++ b/shell_integration/dolphin/CMakeLists.txt
@@ -19,6 +19,7 @@ set_package_properties(DolphinVcs PROPERTIES
     PURPOSE "Provides plugin interfaces for Dolphin."
 )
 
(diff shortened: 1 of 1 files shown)


dschmidt:
Checking if this broke kcodeaddons_add_plugin...

dschmidt:
Oh god why, yeah, it did. 

https://github.com/KDE/kcoreaddons/blob/master/KF5CoreAddonsMacros.cmake#L135
is using  PLUGIN_INSTALL_DIR which is deprecated. Afaict it should use  KDE_INSTALL_PLUGINDIR according to https://api.kde.org/ecm/kde-module/KDEInstallDirs.html.

dschmidt:
Hehe, fixed after I filed a bug report: https://cgit.kde.org/kcoreaddons.git/commit/?id=1285c4bf5552405a3932eb631029de39ec91031d

Issue #5904, opened 2017-07-14
macOS Journal DB filename with dot underscore


guruz:
@dragotin did you notice that a filename like ```._sync_7d56ce96981f.db``` is COMPLETELY INVISIBLE in Finder while ```.csync_journal.db``` is visible in light grey if you have 
```defaults write com.apple.finder AppleShowAllFiles YES``` set.

Did we think about the implications this could have for end users when moving files/folders/directories etc? @michaelstingl 

But maybe normal users anyway don't have ```AppleShowAllFiles``` set?

FYI @ckamm @ogoffart @SamuAlfageme 

Feel free to close this issue, I just wanted to mention it.

michaelstingl:
> Did we think about the implications this could have for end users when moving files/folders/directories etc

If someone moves the full sync folder, it's included. I don't see a "real" problem,

ogoffart:
There was another issue because of ._  in the name:  #5633    (but that was fixed by removing the _  for samba shares)

labkode:
@guruz @michaelstingl  @ogoffart as @matthiasschroder described, we usually have to ask the user for this file while doing support as the owncloud sync longs are not enough for debugging the problem.
Having this file completely invisible (and we do not want to ask end-users to type commands in a Termninal) makes our support life more difficult. If there is not any strong technical reason for having that prefix I propose to change it back to .csync_journal.db or somethink alike.

guruz:
@ckamm maybe the code here could just be swapped around a bit to prefer the non-underscore-path?
https://github.com/owncloud/client/commit/4291ea47f77752848e31341019e9be98f59526df#diff-0b6f105ffd4493a278873f32b7c170c1R84

ckamm:
@guruz Yes.  We needed  `._sync` for client versions <= 2.3.2 apparently. Now that we don't support these versions anymore we could go for plain `.sync` by default. Since the journal file name is stored in the config it'd be easy and safe to do this for new folders without needing lots of migration code.

ckamm:
New databases will be created with the `.sync_` prefix.

labkode:
@ckamm @guruz Thank you!

HanaGemela:
New databases are created with the .sync_ prefix
Tested in 2.6.0alpha2 (build 12128), macOS 10.14.5

PR #6925, opened 2018-12-18
Database: Change path for new dbs to .sync_*

ckamm:
This is to avoid issues on OSX, where the ._ prefix has special meaning.

Originally (before 2.3.2) ._ was necessary to guarantee exclusion. But
since then the .sync_ prefix is excluded as well.

This does not affect existing database files.

For  #5904

### DIFF
--- a/doc/troubleshooting.rst
+++ b/doc/troubleshooting.rst
@@ -55,8 +55,8 @@ Identifying Basic Functionality Problems
 ---------------------
 
 If you see this error message stop your client, delete the
-``._sync_xxxxxxx.db`` file, and then restart your client.
-There is a  hidden ``._sync_xxxxxxx.db`` file inside the folder of every account
+``.sync_xxxxxxx.db`` file, and then restart your client.
+There is a  hidden ``.sync_xxxxxxx.db`` file inside the folder of every account
 configured on your client. 
 
 .. NOTE::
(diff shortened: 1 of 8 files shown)


test/csync/csync_tests/check_csync_exclude.cpp
```diff
@@ -143,16 +143,17 @@ static void check_csync_excluded(void **)
...
-    assert_int_equal(check_file_full("._sync_5bdd60bdfcfa.db.ctmp"), CSYNC_FILE_SILENTLY_EXCLUDED);
-    assert_int_equal(check_file_full("._sync_5bdd60bdfcfa.db-shm"), CSYNC_FILE_SILENTLY_EXCLUDED);
-    assert_int_equal(check_file_full("subdir/._sync_5bdd60bdfcfa.db"), CSYNC_FILE_SILENTLY_EXCLUDED);
-
     assert_int_equal(check_file_full(".sync_5bdd60bdfcfa.db"), CSYNC_FILE_SILENTLY_EXCLUDED);
     assert_int_equal(check_file_full(".sync_5bdd60bdfcfa.db.ctmp"), CSYNC_FILE_SILENTLY_EXCLUDED);
     assert_int_equal(check_file_full(".sync_5bdd60bdfcfa.db-shm"), CSYNC_FILE_SILENTLY_EXCLUDED);
     assert_int_equal(check_file_full("subdir/.sync_5bdd60bdfcfa.db"), CSYNC_FILE_SILENTLY_EXCLUDED);
 
+    /* and the older form */
```
ogoffart: Well, the much older form is checked one paragraph before :-)

ckamm:
I'll just undo the code moves then

PR #6935, opened 2018-12-20
vfs: Remove newFilesAreVirtual - use root PinState instead

ckamm:
This unifies how to deal with pin states.

Also enable reading a folders direct pin state vs its effective pin
state.

### DIFF
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -1990,24 +1990,47 @@ void SyncJournalDb::markVirtualFileForDownloadRecursively(const QByteArray &path
     query.exec();
 }
 
-PinState SyncJournalDb::pinStateForPath(const QByteArray &path)
+Optional<PinState> SyncJournalDb::rawPinStateForPath(const QByteArray &path)
 {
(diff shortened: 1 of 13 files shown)


src/common/syncjournaldb.cpp
```diff
@@ -1990,24 +1990,45 @@ void SyncJournalDb::markVirtualFileForDownloadRecursively(const QByteArray &path
     query.exec();
 }
 
-PinState SyncJournalDb::pinStateForPath(const QByteArray &path)
+PinState SyncJournalDb::rawPinStateForPath(const QByteArray &path)
 {
     QMutexLocker lock(&_mutex);
     if (!checkConnect())
-        return PinState::Unspecified;
+        return PinState::Inherited;
```
ogoffart: I'm wondering if we should not reply with an error there.

ckamm: I wondered the same thing. From reading other functions in SyncJournalDb it seemed that we assume checkConnect() errors during sync are unlikely enough to not be handled consistently.

It'd be easy to error if pinStateForPath() fails to access the db. I've added that.

PR #6939, opened 2018-12-23
Fix a couple of issue with move of nested directories

ogoffart:
Fix issue #6694 and some other move problems.

### DIFF
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -495,10 +495,12 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(
             return;
         }
 
+        QString originalPathAdjusted = _discoveryData->adjustRenamedPath(originalPath, SyncFileItem::Up);
+
         if (!base.isDirectory()) {
             csync_file_stat_t buf;
-            if (csync_vio_local_stat((_discoveryData->_localDir + originalPath).toUtf8(), &buf)) {
-                qCInfo(lcDisco) << "Local file does not exist anymore." << originalPath;
+            if (csync_vio_local_stat((_discoveryData->_localDir + originalPathAdjusted).toUtf8(), &buf)) {
+                qCInfo(lcDisco) << "Local file does not exist anymore." << originalPathAdjusted;
                 return;
             }
             // NOTE: This prohibits some VFS renames from being detected since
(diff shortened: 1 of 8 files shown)


src/libsync/discovery.cpp
```diff
@@ -496,8 +496,8 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(
         bool wasDeletedOnServer = _discoveryData->findAndCancelDeletedJob(originalPath).first;
 
         auto postProcessRename = [this, item, base, originalPath](PathTuple &path) {
-            auto adjustedOriginalPath = _discoveryData->adjustRenamedPath(originalPath);
-            _discoveryData->_renamedItems.insert(originalPath, path._target);
+        auto adjustedOriginalPath = _discoveryData->adjustRenamedPath(originalPath, SyncFileItem::Up);
+            _discoveryData->_renamedItemsRemote.insert(originalPath, path._target);
```
ckamm: odd indent

src/libsync/owncloudpropagator.h
```diff
@@ -506,6 +506,10 @@ class OwncloudPropagator : public QObject
     bool createConflict(const SyncFileItemPtr &item,
         PropagatorCompositeJob *composite, QString *error);
 
+
+    QMap<QString, QString> _renamedDirectories;
```
ckamm: Will not splitting this into local and remote renames lead to bugs similar to in the discovery? 

Please add a explanatory comment too: Why is this necessary when `adjustRenamedPath()` is already called on _item->_file during discovery?

test/testsyncmove.cpp
```diff
@@ -660,7 +659,57 @@ private slots:
...
+        modifier.rename("FolA", "FolA_Renamed");
+        modifier.rename("FolA_Renamed/FolB", "FolB_Renamed");
+        modifier.rename("FolB_Renamed/FolC", "FolA");
+        modifier.rename("FolA/FolD", "FolA/FolD_Renamed");
+        modifier.mkdir("FolB_Renamed/New");
+        modifier.rename("FolA/FolD_Renamed/FolE", "FolB_Renamed/New/FolE");
+        auto expected = local ? fakeFolder.currentLocalState() : fakeFolder.currentRemoteState();
+        QVERIFY(fakeFolder.syncOnce());
+        QCOMPARE(fakeFolder.currentLocalState(), expected);
+        QCOMPARE(fakeFolder.currentRemoteState(), expected);
```
ckamm: Do we already have a test for interacting renames on different sides?
```
local: A -> B
remote: A/foo -> A/bar
```
and the other way around.

src/libsync/owncloudpropagator.h
ogoffart: > Will not splitting this into local and remote renames lead to bugs similar to in the discovery?

No, because in the discovery we map  original -> current local or server path, which might be different from target.

But here we always map origin->target and that is the same for both local and server.

I'm adding a comment

test/testsyncmove.cpp
ogoffart: That's right,  but that'd be another issue.

ogoffart: Fixed now

ckamm: Great, looks good

ogoffart:
@ckamm  ok to merge?

PR #6954, opened 2019-01-08
Vfs <=> selective sync switching

ckamm:
There is now a local/online availability switch on folder context menus in the settings as well as a menu entry to switch into vfs mode and away from it.

I looked at migrating between the two analogous settings (selective sync excluded <-> vfs online-only) but it's complex and will probably not produce exactly what users want. So instead I just start with blank vfs and selective sync settings after switching.

Details: When switching on vfs it seems natual to make all excluded paths online-only, the rest available-locally. But then switching on vfs has no visible effect for users that didn't use selective sync unless they manually also switch folders to online-only.

Migrating from vfs to selective sync is just hard since vfs can have arbitrary switches back and forth between online/local folders. Also, online-only folders can totally have locally available files - so setting up selective sync options automatically from that is likely not worth it.

@michaelstingl FYI

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -336,35 +336,27 @@ void AccountSettings::slotCustomContextMenuRequested(const QPoint &pos)
     ac = menu->addAction(tr("Remove folder sync connection"));
     connect(ac, &QAction::triggered, this, &AccountSettings::slotRemoveCurrentFolder);
 
-    if ((Theme::instance()->showVirtualFilesOption() && folder->supportsVirtualFiles()) || folder->newFilesAreVirtual()) {
-        ac = menu->addAction(tr("Create virtual files for new files (Experimental)"));
+    if (folder->supportsVirtualFiles()) {
+        auto availabilityMenu = menu->addMenu(tr("Availability"));
+        ac = availabilityMenu->addAction(tr("Local"));
+        ac->setCheckable(true);
+        ac->setChecked(!folder->newFilesAreVirtual());
+        connect(ac, &QAction::triggered, this, [this]() { slotSetCurrentFolderAvailability(PinState::AlwaysLocal); });
+
+        ac = availabilityMenu->addAction(tr("Online only"));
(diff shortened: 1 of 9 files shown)


ogoffart:
> optional: data on heap, unions don't like non-trival dtors

Why?  
Normally this should work if we call the destructor manyally, which Error does.

src/gui/accountsettings.cpp
```diff
@@ -541,6 +533,103 @@ void AccountSettings::slotOpenCurrentLocalSubFolder()
...
+        for (const auto &entry : oldBlacklist) {
+            folder->journalDb()->avoidReadFromDbOnNextSync(entry);
+        }
+
+        // Change the folder vfs mode and load the plugin
+        folder->setSupportsVirtualFiles(true);
+
+        // Wipe pin states to be sure
+        folder->journalDb()->wipePinStateForPathAndBelow("");
+        folder->journalDb()->setPinStateForPath("", PinState::OnlineOnly);
```
ogoffart: should we not set the oldBlackList as Online Only ?

src/gui/accountsettings.cpp
```diff
@@ -541,6 +533,103 @@ void AccountSettings::slotOpenCurrentLocalSubFolder()
...
+    QPointer<Folder> folder = folderMan->folder(selectedFolderAlias());
+    QModelIndex selected = ui->_folderList->selectionModel()->currentIndex();
+    if (!selected.isValid() || !folder)
+        return;
+
+    auto msgBox = new QMessageBox(
+        QMessageBox::Question,
+        tr("Disable virtual file support?"),
+        tr("This action will disable virtual file support. As a consequence contents of folders that "
+           "are currently marked as 'available online only' will either be downloaded or excluded "
```
ogoffart: I believe they will all be downloaded

ckamm:
@ogoffart Please rereview

ogoffart:
I've merged this since other pr depends on it.

PR #6959, opened 2019-01-09
Fix lookup of system wide settings

dschmidt:
ConfigFile sets the application name to `appNameGUI()`, so we should also use that for looking up settings in the registry. 

It's also what is used in NSIS to write the `skipUpdateCheck` setting in the registry. [1] For the unbranded ownCloud client it doesn't matter, but if application name and application shortname differ, the current code fails to retrieve the setting correctly. 

Worst of all: it's completely inconsistent: for user policies the shortname is used `appName()` but for machine policies the full application name is used via `APPLICATION_NAME`. Replaced that also with `appNameGUI()` to be more consistent.

[1] https://github.com/owncloud/client/blob/master/cmake/modules/NSIS.template.in#L819

### DIFF
--- a/src/libsync/configfile.cpp
+++ b/src/libsync/configfile.cpp
@@ -248,14 +248,14 @@ QVariant ConfigFile::getPolicySetting(const QString &setting, const QVariant &de
     if (Utility::isWindows()) {
         // check for policies first and return immediately if a value is found.
         QSettings userPolicy(QString::fromLatin1("HKEY_CURRENT_USER\\Software\\Policies\\%1\\%2")
-                                 .arg(APPLICATION_VENDOR, Theme::instance()->appName()),
+                                 .arg(APPLICATION_VENDOR, Theme::instance()->appNameGUI()),
             QSettings::NativeFormat);
         if (userPolicy.contains(setting)) {
             return userPolicy.value(setting);
         }
 
         QSettings machinePolicy(QString::fromLatin1("HKEY_LOCAL_MACHINE\\Software\\Policies\\%1\\%2")
-                                    .arg(APPLICATION_VENDOR, APPLICATION_NAME),
+                                    .arg(APPLICATION_VENDOR, Theme::instance()->appNameGUI()),
             QSettings::NativeFormat);
         if (machinePolicy.contains(setting)) {
             return machinePolicy.value(setting);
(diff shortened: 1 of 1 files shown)


dschmidt:
@guruz Well, IMHO it was broken before. It unbreaks certain branded apps. Let's assume ownCloud's shortname was `ownCloud` and the application name was `own Cloud` then settings written and read through would go to `ownCloud/own Cloud` while the current code tries to read it from `ownCloud/ownCloud`. This already breaks with `skipUpdateCheck` as you can see if you look at  `NSIS.template.in`. 

I'm obviously sending this for review because I want feedback, if this actually breaks something, then we should do it differently of course. But please point out where you think it would break anything.

dschmidt:
FTR: appName() and appNameGUI() are the same for unbranded ownCloud client.
NSIS writes to appNameGUI():
`WriteRegDWORD HKLM "Software\${APPLICATION_VENDOR}\${APPLICATION_NAME}"`
so does MSI.

Where would you expect anything to break? Yes, admins may have found this out and put reg values in the wrong location. That's why we will put a notice in the customer mailing when this is released.

dschmidt:
Ah and to make it clear: this is not about user settings or anything. This is only for system settings set by an administrator or the installer.

ogoffart:
and we can't change everything to just appName?  because that would break reading the config from previous version?

dschmidt:
That would break more IMHO. This only changes installer and/or admin settings. Changing the `QApplication::applicationName()` would also break paths in the filesystem, that seems way worse to me.

As you can see, this is only a lookup in HKEY_LOCAL_MACHINE, no user setting.

dschmidt:
Except for the policy, but that's also nothing we set manually but it's only set via some GPO afaict.

dschmidt:
As I said: this affects only branded clients and we will mention it in customer mailings. 
The documentation at https://github.com/owncloud/client/blob/master/doc/autoupdate.rst only says`ownCloud/ownCloud` which is ambiguous. The installer(s) use(s) `appNameGUI()` since forever or even the beginning. So it likely has never worked for branded builds where shortname and applicationname differ more than in casing (that seems to still work, apparently because the registry is case insensitive). I'm not aware that this is even used for anything else but the auto update check by us - although it can be used to give defaults for other values as well of course. 

IMHO this is at least a less intrusive change than changing the QApplication name. That would be a can of worms.

guruz:
Appvoyor still fails with this qtransla* stuff https://ci.appveyor.com/project/ownclouders/client/builds/21495813/job/wt8vat1pwhxldm1j

dschmidt:
@guruz exactly! 

Yeah, will take care of AppVeyor after macos symbols.

Issue #6881, opened 2018-11-14
[Windows] EXCEPTION_IN_PAGE_ERROR Crash in sqlite3 WAL since 2.5.0


guruz:
See https://sentry.io/owncloud/desktop-win-and-mac/issues/705087323/events/

The [backtrace shown in sentry is a bit confusing](https://github.com/owncloud/enterprise/issues/2986) but from raw one you can see:
```
Thread 0 (crashed)
 0  owncloud_csync.dll!static int walIndexTryHdr(struct Wal *, int *) [sqlite3.c : 59283 + 0x1]
    eip = 0x6d68d122   esp = 0x00afc27c   ebp = 0x00afc378   ebx = 0x08090178
    esi = 0x007b0000   edi = 0x00000000   eax = 0x007b0000   ecx = 0x007b0000
    edx = 0x00000000   efl = 0x00010286
    Found by: given as instruction pointer in context
 1  owncloud_csync.dll!static int walIndexReadHdr(struct Wal *, int *) [sqlite3.c : 59367 + 0xf]
    eip = 0x6d68cb3c   esp = 0x00afc300   ebp = 0x00afc378
    Found by: call frame info
 2  owncloud_csync.dll!static int walTryBeginRead(struct Wal *, int *, int, int) [sqlite3.c : 59685 + 0x7]
    eip = 0x6d68dd88   esp = 0x00afc320   ebp = 0x00afc378
    Found by: call frame info
 3  owncloud_csync.dll!static int sqlite3WalBeginReadTransaction(struct Wal *, int *) [sqlite3.c : 59956 + 0xb]
    eip = 0x6d680abb   esp = 0x00afc344   ebp = 0x00afc378
    Found by: call frame info
 4  owncloud_csync.dll!static int pagerBeginReadTransaction(struct Pager *) [sqlite3.c : 52804 + 0x10]
    eip = 0x6d644f8a   esp = 0x00afc364   ebp = 0x00afc378
    Found by: call frame info
 5  owncloud_csync.dll!static int sqlite3PagerSharedLock(struct Pager *) [sqlite3.c : 54898 + 0x6]
    eip = 0x6d667de1   esp = 0x00afc380   ebp = 0x00afc378
    Found by: call frame info
 6  owncloud_csync.dll!static int lockBtree(struct BtShared *) [sqlite3.c : 65032 + 0x13]
    eip = 0x6d641587   esp = 0x00afc3ac   ebp = 0x00afc378
    Found by: call frame info
[...]

PR #6960, opened 2019-01-10
sqlite: Set exclusive locking_mode to avoid WAL issues

ckamm:
Please double check that this is safe to do for a patch release. I think it is.

For #6881

### DIFF
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -327,6 +327,18 @@ bool SyncJournalDb::checkConnect()
         }
     }
 
+    // Set locking mode to avoid issues with WAL on Windows
+    static QByteArray locking_mode_env = qgetenv("OWNCLOUD_SQLITE_LOCKING_MODE");
+    if (locking_mode_env.isEmpty())
+        locking_mode_env = "EXCLUSIVE";
+    pragma1.prepare("PRAGMA locking_mode=" + locking_mode_env + ";");
+    if (!pragma1.exec()) {
+        return sqlFail("Set PRAGMA locking_mode", pragma1);
+    } else {
+        pragma1.next();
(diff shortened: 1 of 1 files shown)


guruz:
Are we really sure this works for our multi-threaded access? (vs multi-process)

https://www.sqlite.org/wal.html

> Specialized applications for which the default implementation of shared memory is unacceptable can devise alternative methods via a custom VFS. For example, if it is known that a particular database will only be accessed by threads within a single process, the wal-index can be implemented using heap memory instead of true shared memory.

Doesnt this kind of imply we need a wal-index if we want multi-thread?

>  If the VFS does not support shared-memory methods, then the attempt to open a database that is already in WAL mode, or the attempt convert a database into WAL mode, will fail. 

Hm.

guruz:
(see also my comments on https://github.com/owncloud/client/issues/6881#issuecomment-453145542 )

ckamm:
By default I'd have assumed that an exclusive locking_mode on an sqlite connection prevents other connections, even in the same process. The bugreport suggests otherwise but I haven't found explicit mention of it in the documentation.

In the client all db access goes through a single `SyncJournalDb` instance and thus one connection. (The separate connection in `SocketApi` is gone)

Poking around a bit I see that the shared-cache mode (https://www.sqlite.org/sharedcache.html) allows multiple connections in the same process to share a cache. According to the source code that seems to relate to the sqlite pager, where the locking is located. It's plausible that this mode would allow multiple connections in the process (if we needed it).

The quote about WAL mode opens failing if there's no shared-memory methods is about normal locking_mode. With exclusive locking mode we'd get a heap wal.

guruz:
You kind of have me convinced. I'd +1.
But we can wait a bit more for @ogoffart to agree.

Issue #6953, opened 2019-01-08
Vfs with suffix files crashes in discovery when server has virtual files uploaded


ckamm:
To reproduce upload
```
test
test.owncloud
test.owncloud.owncloud
```
to the server. The discovery will get confused and crash in `ASSERT(_dirItem && _dirItem->_instruction == CSYNC_INSTRUCTION_RENAME);`.

HanaGemela:
macOS 10.14.5 and Windows 10 when VFS off:
File type is virtual file - Is that OK?
![image](https://user-images.githubusercontent.com/49001702/61534266-6ec37980-aa2f-11e9-81bd-ccce8bae8e72.png)

macOS 10.14.5 when VFS on:
only test.owncloud in the client

Windows 10 when VFS on:
only test and test.owncloud in the client

What exactly is the expected result here?

ckamm:
> macOS 10.14.5 and Windows 10 when VFS off

This looks right, all three files are synced as-is.

> macOS 10.14.5 when VFS on

Looks good too, all .owncloud files on the server are ignored, so only "test" gets propagated as "test.owncloud".

> Windows 10 when VFS on

That's odd. I'd have imagined all three files to be around - with the cfapi plugin the client doesn't even know about the .owncloud suffix. Will try to reproduce.

HanaGemela:
Thanks, let me know if you can reproduce

ckamm:
@HanaGemela I can't reproduce. Steps:
- Create "test", "test.owncloud", "test.owncloud.owncloud" on the server
- sync on a 2.6 branch client with cfapi vfs enabled
-> see test, test.owncloud, test.owncloud.owncloud in the filesystem

I've also tested that the "Free up space" and "Always available on this device" actions work on them.

HanaGemela:
Tried again and cannot reproduce either. 
But I double checked 

Looks good too, all .owncloud files on the server are ignored, so only "test" gets propagated as "test.owncloud".

And the content of the file is empty so I can't verify. The same issue as https://github.com/owncloud/client/issues/7338

ckamm:
@HanaGemela I don't quite understand your message. What exactly can't you verify? How does this issue relate to #7338. If you still see an issue (in particular with today's build) can you give me updated steps (or create a new issue if unrelated)?

HanaGemela:
[...]

PR #6962, opened 2019-01-10
Fix discovery crash when non-virtual but suffixed files exist on server or in db

ckamm:
Depends on #6954 - that must merge first.

For #6953


ogoffart:
The content of the commit  01ba53d looks good.  But You are changing the submodule, I guess that's not intentional!

ckamm:
Indeed, fixing!

Issue #6911, opened 2018-12-01
Client does not send client TLS certificate


albatross0:
### Expected behaviour
A new account is created.

### Actual behaviour
I tried to add a new account using the client TLS ceritificate but it failed.
(No client certificates in the TLS handshake packets)
Owncloud client version 2.5.0 does not have this problem.

### Steps to reproduce
1.  Install owncloud client version 2.5.1
2.  Add a new account using the client TLS certificate

### Client configuration
Client version: 2.5.1

Operating system: Windows 10

OS language: Japanese

### Logs

1. Client logfile: https://gist.github.com/albatross0/2758870f54f5edd935baaef981202b95

guruz:
@albatross0 Which error message do you get when adding a NEW acccount (with same URL)?

guruz:
>  (No client certificates in the TLS handshake packets) 

How did you find this out? Network sniffer?

Does the log on the server confirm this?

FYI @dschmidt you changed the SSL version in 2.5.1?

albatross0:
> How did you find this out? Network sniffer?

Yes. I used Wireshark.

I think probably #6836 affects this issue.

dschmidt:
> FYI @dschmidt you changed the SSL version in 2.5.1?

I think so. From 1.1.0x to 1.1.1

guruz:
```Client version 2.5.1 (build 10818)  Qt 5.10.1  SSL  OpenSSL 1.1.1  11 Sep 2018```
@dschmidt Did you check if Qt's OpenSSL 1.0 backend has feature parity with the old 1.0 OpenSSL backend before updating? AFAIK they're using different classes.

I spent 5 minutes to search on https://bugreports.qt.io/secure/Dashboard.jspa but didnt see anything

ckamm:
Could be related to ef99800a944ed31d60c9783212c7d85d37dcda62 which is in 2.5.1

guruz:
@albatross0 Do you want to test the rc2 which contains a fix for this?
https://download.owncloud.com/desktop/testing/ownCloud-2.5.2.11135rc2.pkg

guruz:
Woops, meant to give you the Windows link :)

guruz:
@albatross0 Do you want to test the rc2 which contains a fix for this?
https://download.owncloud.com/desktop/testing/ownCloud-2.5.2.11335.11186-rc2.en-US.msi

albatross0:
@guruz Thank you for letting me know.
[...]

PR #6965, opened 2019-01-11
Wizard: Fix setting up accounts with ssl client certificates

ckamm:
See #6911 

Please review carefully since it touches the wizard. I tested with my server, my server+ssl cert and demo.owncloud.com.


guruz:
I can confirm SSL client certs work again after this PR.
About the wizard code itself though I'm not very knowledged..

PR #6966, opened 2019-01-11
Fix debug symbols on macOS

dschmidt:


guruz:
> Do not run macdeployqt.py from CMake. 

How does make install work then?

dschmidt:
The CI Script calls it directly. The other changes I did require that the script is run after *all* targets are installed. 

I need to know which files have been built by ourselves so I can run dsymutil on them. And as the script does not get passed the target it's enough to run it once in the end.
Also I would consider pulling in deps at install time an antipattern - this concrete scenario is a good example why.

dschmidt:
... and as the script is not target specific but run on the whole bundle it's enough to run it once after all is installed instead of running it trice after every application target is installed.

guruz:
Hm would have been better to also check ENABLE_CRASHREPORTS always before CopyDSYM,

Issue #6936, opened 2018-12-21
Data loss in vfs/selective sync interaction


ckamm:
See https://github.com/owncloud/enterprise/issues/3065 written up by @jnweiger:

> Windows desktop client 2.5.1. User wants to explore the experimental VFS featue. His client is connected to cloud.owncloud.com, and he uses selective sync to hide parts of the tree. Several GB of data are synced to his local disk.
>
> Now
> - he disables more folders to reduce his local disk usage.
> - Without waiting, he immediately switches the account to virtual files.
> 
> DATA LOSS: The client now removes the newly disabled folders from his local storage and also on the server.

ckamm:
I'd target this for 2.6.0 since there are large differences in vfs support between 2.5.1 and that and 2.5 vfs support was experimental to start with. If it ends up being easy to backport we can do that, but let's do the cost-benefit analysis of that after the primary fix is done.

ckamm:
This relates to the general choice of how to switch between vfs modes and what to do about selective sync in #6932. A simple "bugfix" for 2.5.x could be to remove the ability to enable vfs mode for existing folders.

ckamm:
This is still a problem with PR #6954: The user can trigger a vfs on/off switch while the folder is syncing. That should be prohibited.

ckamm:
I can't prevent bad things from happening if the user deletes files instead of waiting - but the sync client is more careful about the vfs on/off switch now.

HanaGemela:
1. Turn on VFS (wait until it's done) 
2. Turn off VFS
3. Immediately (when the sync is in progress) deselect all folders in the selective sync and confirm

This triggers 'All files were deleted on the server' message. User can select 'Keep all' so no data loss on the server but the message should not be shown.
After selecting 'Keep all':

1. Checkboxes in the selective sync list are not checked
2. Large (1G) files are missing in the file system (but present on the server) as they didn't manage to be downloading after switching off VFS
3. There are no icons against the files even thought the account is connected to the server (I can't recreate this one)

macOS 10.14.5, client 2.6.0alpha2 (build 12128)

HanaGemela:
Apart from point 2 (Large files missing) I think it is not related to this issue. They seem to be issues with selective sync. Do you want me to raise it separately or will you fix everything here?

ckamm:
I think this new issue is different from the original one. This one relates to terminating the sync run that happens when vfs is switched off. When vfs is switched off:
[...]

PR #6973, opened 2019-01-14
Switch vfs state only when no sync is running / fix discovery aborts

ckamm:
Depends on https://github.com/owncloud/client/pull/6954 which must merge first.

See #6936


src/gui/wizard/owncloudwizard.cpp
```diff
@@ -246,6 +246,8 @@ void OwncloudWizard::askExperimentalVirtualFilesFeature(const std::function<void
                "When a file is opened its contents will be downloaded automatically. "
                "Alternatively files can be downloaded manually by using their context menu."
                "\n\n"
+               "Switching to this mode will abort any currently running synchronization."
```
ogoffart: (optional) Maybe we could be smart and show this message only if the sync is running

ogoffart:
Nice cleanups \o/

PR #6976, opened 2019-01-15
vfs: Update pinning context menu to be less confusing

ckamm:
Seeing "Currently available online only" for a currently hydrated file
was odd. It makes sense since current hydration status and pin state are
independent.

The new text will say something like "Currently available, but marked
online only" to better indicate that the file might be dehydrated later
since it wasn't pinned.

FYI @michaelstingl. (This will become easier for users once there are file icons; essentially there are three states: "pinned & hydrated", "unpinned & hydrated", "unpinned & dehydrated".)

### DIFF
--- a/src/gui/socketapi.cpp
+++ b/src/gui/socketapi.cpp
@@ -908,8 +908,11 @@ void SocketApi::command_GET_MENU_ITEMS(const QString &argument, OCC::SocketListe
     if (folder && folder->supportsVirtualFiles()) {
         bool hasAlwaysLocal = false;
         bool hasOnlineOnly = false;
+        bool hasHydratedOnlineOnly = false;
+        bool hasDehydratedOnlineOnly = false;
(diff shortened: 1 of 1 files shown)


src/gui/socketapi.cpp
```diff
@@ -919,20 +922,52 @@ void SocketApi::command_GET_MENU_ITEMS(const QString &argument, OCC::SocketListe
...
                 hasOnlineOnly = true;
+                auto record = fileData.journalRecord();
+                if (record._type == ItemTypeFile)
+                    hasHydratedOnlineOnly = true;
+                if (record.isVirtualFile())
+                    hasDehydratedOnlineOnly = true;
             }
         }
 
+        auto makePinContextMenu = [listener](QString currentState, QString availableLocally, QString onlineOnly) {
```
ogoffart: (minor nitpick) codeing style: `const QString &`

PR #6992, opened 2019-01-21
Add a noPreFlight branding option

ogoffart:
For https://github.com/owncloud/enterprise/issues/2951

### DIFF
--- a/src/libsync/theme.h
+++ b/src/libsync/theme.h
@@ -358,6 +358,15 @@ public:
      */
     virtual bool showVirtualFilesOption() const;
 
+    /**
+     * @brief Open directly the browser for oauth2 without doing any unauthenticated request
+     *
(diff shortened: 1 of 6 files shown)


src/libsync/theme.h
```diff
@@ -358,6 +358,15 @@ class OWNCLOUDSYNC_EXPORT Theme : public QObject
...
     virtual bool showVirtualFilesOption() const;
 
+    /**
+     * @brief Open directly the browser for oauth2 without doing any unauthenticated request
+     *
+     * This is a very specific option which only make sense if overrideServerUrl is set.
+     * It also impact redirect handling
+     * https://github.com/owncloud/enterprise/issues/2951
+     */
+    virtual bool noPreFlight() const;
```
guruz: Why not a clearer name? something like ```noStatusPhpAlwaysOAuth2()``` ?

Issue #6776, opened 2018-09-18
Connection using certificates not working after restart


wdehoog:
### Expected behaviour
Using the installation wizard I import my certificate and complete the setup. After the client or my pc restarts the connection must be restored.

### Actual behaviour
The connection is not restored and cannnot be. I have to remove the account and recreate it. The next restart the same problem occurs.

### Server configuration
Operating system: ubuntu 18.04
Web server:  apache2
Database: sqlite
PHP version: php 7.2
ownCloud version: 13.06 (nextcloud)

### Client configuration
Client version: 2.5.0
Operating system: windows 10

Note that the Nextcloud client suffers from the same problem.

I have build branch 2.4 of the nextcloud client (probably not much different from the owncloud one) (QT 5.11.1 MinGW 32 bits).  When building qtkeychain with -DUSE_CREDENTIAL_STORE=OFF the certificates are saved in the .cfg file (as PEM) and are loaded after a restart  so login works.

ckamm:
@wdehoog Does storing the basic credentials work (test with demo.owncloud.org demo/demo)? If disabling the credential store in qtkeychain helps this might be a bug upstream. :/

wdehoog:
@ckamm yes when using -DUSE_CREDENTIAL_STORE=OFF storing the basic credentials still works. I agree the problem is likely to be upstream. I am really sorry but the coming days I have no time to reinstall and provide logs.

ckamm:
@wdehoog Ok! I was wondering whether *with* the credential store enabled in qtkeychain it was only the certificate storing that was broken or also the basic http credential saving.

wdehoog:
@ckamm that is hard to tell since the credentials are only used after the certificates are ok and since the certificates could not be loaded (they were not saved) the credentials are not asked for.

Anyway I just wanted to let you know there is a problem and building the client in a slightly other way could solve it.

guruz:
@wdehoog Then please test if the credential saving works when you add a sample account from https://demo.owncloud.com/login (see that URL in browser for credentials)

guruz:
> I have build branch 2.4 of the nextcloud client 

Ooops. Wrong repo then here. Please build from https://github.com/owncloud/client.git .. also you mentioned 2.5 so you need to use the 2.4 branch.

Please re-open if ownCloud client also has issues.

wdehoog:
[...]

PR #6995, opened 2019-01-22
HttpCreds: Warn in log if keychain-write jobs fail

ckamm:
Also, calling deleteLater() on jobs is unnecessary (they autodelete
after finished()) and deleting the attached QSettings is also
unnecessary because the settings object is parented to the job.

See  #6776


guruz:
Changed base to 2.5 as we're probably doing another release there.

PR #6999, opened 2019-01-23
Db: Move drop-index to after pragmas are set 

ckamm:
Maybe a fix for @jnweiger's observation of present -wal/-shm files even though I couldn't reproduce it. #6881

Your code had a "must be done before synchonous mode is set" note - I couldn't see why, please check.


ogoffart:
Did you test that using master, then downgrading to 2.5.x works?
If yes, then it's fine

I recall that in my test, this did not work.  But I may remember wrong, or did something else wrong.

ckamm:
Yes, that worked fine for me.

PR #7007, opened 2019-01-30
[macOS] Prepare notarization

dschmidt:


dschmidt:
You realize that I already did a fully notarized build of this branch?

guruz:
Cert already looks good (not sure when this changed).
Will try to install.

screen shot 2019-01-30 at 13 54 08

guruz:
@dschmidt In Mojave I thought I should see one of https://eclecticlight.co/2018/09/24/has-that-app-been-notarized-and-what-are-the-benefits-to-the-user/ but I don't..
Or did I click them away before and forgot or so?

michaelstingl:
> Or did I click them away before and forgot or so?

I think you only see this when you open an `*.app` directly. When you use the `*.pkg` installer, and the installer asks for your password, you won't see the message.

guruz:
let's target this for 2.6 instead when we have QtWebKit removed
(CC @ogoffart )

dschmidt:
FTR if we need to readd QtWebKit for whatever reason, it has a compile time option to disable JIT. 

``` 
<domme|Dominik Schmidt>
12:30 PM hey. has anyone experience with notarizing apps shipping qtwebkit on macos?
12:31 PM I saw upstream WebKit has https://github.com/WebKit/webkit/blob/master/Source/JavaScriptCore/allow-jit-macOS.entitlements but I could not find it in QtWebkit (5.212)
12:33 PM apparently that and adding MAP_JIT to mmap calls is needed. No idea if more. 
12:33 PM or is there a way to disable JIT that makes things only slower but not broken?
12:34 PM annulen: ^
12:38 PM https://lists.webkit.org/pipermail/webkit-dev/2012-March/019943.html <-- is this still up to date?
12:42 PM or .. this https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-qt/qtwebkit/qtwebkit-5.212.0_pre20180120.ebuild#n117 ?
2:52 PM A<annulen|Konstantin Tokarev> domme: good catch. while allow-jit-macOS.entitlements is missing in qtwebkit branch, it's not enough to have that file, it also should be used somehow and I have no idea what to do with that
2:52 PM as for disabling JIT, there is a public cmake option for that
2:53 PM also, there is USE_APPSTORE_COMPLIANT_CODE option, which wasn't really tested but should in theory work for you
2:55 PM after thinking a bit about entitlements file, it may be actually inappropriate, because if Apple gave such entitlement to its own project it doesn't mean that forks retain it
```

dschmidt:
I retargeted this to master, afaict it should be good to merge now.

@guruz @ogoffart Thoughts?

Issue #6885, opened 2018-11-14
Downloading files with content-encoding: gzip fails with HTTP2


lazawan:
Virtualfile aborts during Downloading with Error message saying :                                                                   " The file could not be downloaded completely " 
the bug happened on Ubuntu 18.10 box
 
![screenshot_20181114_163641](https://user-images.githubusercontent.com/44814084/48497482-798c9580-e834-11e8-9753-df35e6347325.png)

michaelstingl:
Please make sure you enabled `[x] Permanently save logs` and `[x] Capture debug messages` options in the F-12 window while performing tests. Here you can find more information: https://doc.owncloud.org/desktop/2.5/troubleshooting.html#logging-to-a-temporary-directory

ogoffart:
This happens if the downloaded file does not have the same size as the content-lenght header.  This should not happen under normal condition, unless there is a bug in the server that does not give the full file, or a proxy or so cuts the connection.
A complete log could be useful.

lazawan:
yesterday i enabled logging but it didn't provide informations ,that are useful .
i have tried  to download several files , all of them was properly downloaded except two ,that were txt files .

ogoffart:
I see that you are using demo.owncloud.com,  this server resets itself every now and then. Maybe it was reset while the file was downloading, causing the connection to close too early.
This is just a theory.

My feeling is that the error was genuine.
Maybe the error message could be improved.

dialognewmedia-mws:
Hi there,

I'm seeing this error as well, but only on Linux clients trying to use virtual files that are text/plain. The error does not happen on Mac clients accessing text/plain virtual files.

Alas, the error message from the server isn't very helpful. The error only appears to happen with text files.

japerki:
On Dec 15, 2018 7:55 PM, Matthew Sheffield <notifications@github.com> wrote:

Hi there,

I'm seeing this error as well, but only on Linux clients trying to use virtual files that are text/plain. The error does not happen on Mac clients accessing text/plain virtual files.

Alas, the error message from the server isn't very helpful. The error only appears to happen with text files.

ogoffart:
It would be nice to have more detailed step to reproduce, or log available.

dialognewmedia-mws:
1. Install Linux OC client
2. Sign into server and use virtual files option
3. Attempt to download files ending in .txt
4. See error captured in the original screenshot.

lazawan:
[...]

PR #7008, opened 2019-01-30
Download: Ignore content-length for compressed HTTP2/SPDY replies #6885

ckamm:
It contains the compressed size.

See https://bugreports.qt.io/browse/QTBUG-73364


ckamm:
Yeah, probably there are no other encodings... It's mirroring the check done in `QHttpNetworkReply::isCompressed()`.

PR #7017, opened 2019-02-06
Vfs adjustments for deeper shell integration

ckamm:
This is
- some code organization changes (moves to common/) to make classes accessible in the plugin
- significant pin-state updates to improve behavior and make it compatible among plugins
- discovery change to allow vfs-actions originating from local files
- a propagatedownload attribute propagation fix
- vfs interface additions and changes
- proper vfs unregistering on folder removal


src/libsync/owncloudpropagator.cpp
```diff
@@ -976,9 +983,7 @@ void PropagateDirectory::slotSubJobsFinished(SyncFileItem::Status status)
                     _item->_fileId = mkdir->_item->_fileId;
                 }
             }
-            SyncJournalFileRecord record = _item->toSyncJournalFileRecordWithInode(propagator()->_localDir + _item->_file);
-            bool ok = propagator()->_journal->setFileRecord(record);
-            if (!ok) {
+            if (!propagator()->updateMetadata(*_item)) {
                 status = _item->_status = SyncFileItem::FatalError;
```
ogoffart: Some of the error handling could also be moved to the helper
Edit:  Maybe not

src/libsync/owncloudpropagator.cpp
```diff
@@ -1025,14 +1030,17 @@ void CleanupPollsJob::slotPollFinished()
...
+        QString filesystemPath = _localPath + job->_item->_file;
+        if (!_journal->setFileRecord(job->_item->toSyncJournalFileRecordWithInode(filesystemPath))) {
             qCWarning(lcCleanupPolls) << "database error";
             job->_item->_status = SyncFileItem::FatalError;
             job->_item->_errorString = tr("Error writing metadata to the database");
             emit aborted(job->_item->_errorString);
             deleteLater();
             return;
         }
+        // TODO: Is syncfilestatustracker notified somehow?
```
ogoffart: This is old unused code.
It is going to be updated by https://github.com/owncloud/client/pull/6614

But anyway, no need to update the syncfilestatustracker because this is not about files from this sync, this happen before the discovery.

src/common/vfs.h
```diff
@@ -31,7 +31,6 @@ typedef QSharedPointer<Account> AccountPtr;
 class SyncJournalDb;
 class VfsPrivate;
 class SyncFileItem;
-typedef QSharedPointer<SyncFileItem> SyncFileItemPtr;
```
ogoffart: (Putting a message on the first line of the commit, but i want to comment on the whole commit)

I do not understand the relation between the commit message and this commit.
It does not seem to touch directory, it is just some refactoring, is it not?

src/libsync/propagatedownload.cpp
```diff
@@ -972,8 +965,15 @@ void PropagateDownloadFile::downloadFinished()
     // Apply the remote permissions
     FileSystem::setFileReadOnlyWeak(_tmpFile.fileName(), !_item->_remotePerm.isNull() && !_item->_remotePerm.hasPermission(RemotePermissions::CanWrite));
 
-    // Make the file a hydrated placeholder if possible
-    propagator()->syncOptions()._vfs->convertToPlaceholder(_tmpFile.fileName(), *_item, fn);
+    bool isConflict = _item->_instruction == CSYNC_INSTRUCTION_CONFLICT
```
ogoffart: Could there be a test for this?

src/csync/std/c_private.h
```diff
@@ -42,8 +42,12 @@
 #include <errno.h>
 
 #ifdef __MINGW32__
+#ifndef EDQUOT
```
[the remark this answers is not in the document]
ckamm: Seems like it, will push fixup commits and fold in later

src/common/common.cmake
```diff
@@ -11,6 +11,7 @@ set(common_SOURCES
     ${CMAKE_CURRENT_LIST_DIR}/remotepermissions.cpp
     ${CMAKE_CURRENT_LIST_DIR}/vfs.cpp
     ${CMAKE_CURRENT_LIST_DIR}/plugin.cpp
+    ${CMAKE_CURRENT_LIST_DIR}/syncfilestatus.cpp
```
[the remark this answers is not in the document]
ckamm: The reason I moved it was because the `Vfs` declarations are in `common` as they're still used from `csync/vio`

src/libsync/syncfileitem.cpp
```diff
@@ -24,10 +24,10 @@ namespace OCC {
 
 Q_LOGGING_CATEGORY(lcFileItem, "sync.fileitem", QtInfoMsg)
 
-SyncJournalFileRecord SyncFileItem::toSyncJournalFileRecordWithInode(const QString &localFileName)
+SyncJournalFileRecord SyncFileItem::toSyncJournalFileRecordWithInode(const QString &localFileName) const
 {
     SyncJournalFileRecord rec;
-    rec._path = _file.toUtf8();
+    rec._path = destination().toUtf8();
```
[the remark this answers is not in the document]
ckamm: Yes, would make sense to add db-checks to the SyncMove testsuites. Will do.

src/libsync/owncloudpropagator.cpp
ckamm: I was wondering whether `SyncFileStatusTracker::slotItemCompleted` would ever be emitted for a file that is finished only after polling completes. You're saying that's working fine and the comment can be removed?

src/common/vfs.h
ckamm: The significant change that happens in "Vfs: Make files that end up in db placeholders" is that `convertToPlaceholder()` is called for each `updateMetadata()`. Previously it was called much more sparingly and that would cause trouble down the line.

src/libsync/propagatedownload.cpp
ckamm: Added one!

PR #7021, opened 2019-02-08
Add doc CMake target

dschmidt:


dschmidt:
Possibly not perfect, but a start.. if we need more, I can take care of it later.

PR #7031, opened 2019-02-13
Fix compiler warnings

ckamm:
Many warnings originated in the fact that the code wasn't consistent about whether it used `qint64` or `quint64` for file sizes. It now consistently uses `qint64`.

### DIFF
--- a/src/cmd/cmd.cpp
+++ b/src/cmd/cmd.cpp
@@ -83,7 +83,7 @@ struct CmdOptions
     int downlimit;
     int uplimit;
     bool deltasync;
-    quint64 deltasyncminfilesize;
+    qint64 deltasyncminfilesize;
 };
 
 // we can't use csync_set_userdata because the SyncEngine sets it already.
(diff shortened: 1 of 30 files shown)


dschmidt:
I was just wondering why you chose the signed variant in most of the cases? Less changes?
Because for a size I'd imagine it should never be negative...

ckamm:
@dschmidt Yes, I went with signed primarily because it was the smaller change and because Qt, csync and zsync agree it should be int64_t. 

In the end, both choices work, though I also have a preference for signedness here: While "can't be negative" makes sense from a type-contract point of view, the way the contract is enforced by overflowing is not helpful at all.

A proper NonNegativeInt type would actively error on contract violation (@ogoffart this kind of type is probably what you also had in mind?). I had the itch to make it, but it would have been too large a change for this PR.

@dschmidt @ogoffart But now that we're talking about it, what about a simple one that transparently decays to int64_t, asserts when constructed from a negative value and sets itself to zero, has no arithmetic and is usually only interacted with via its constructor function `assureNonNegative(int64) -> NonNegative<int64>`? To me, it feels like it might be worth it although we have more important fish to fry.

dschmidt:
Thanks, for the detailed explanation, @ckamm! 

Poor :fish:! But yeah, not so important, I'm happy with your explanation.

PR #7032, opened 2019-02-13
Tests: Fix permission propagation check

ckamm:
It was using the wrong path to the conflict file.


dschmidt:
`All checks have passed.` woah :tada:

PR #7035, opened 2019-02-13
Update windows build instructions

dschmidt:
[Rendered](https://github.com/owncloud/client/blob/windows-build-doc/docs/modules/ROOT/pages/building.adoc#windows-development-build-with-kde-craft)

### DIFF
--- a/docs/modules/ROOT/pages/building.adoc
+++ b/docs/modules/ROOT/pages/building.adoc
@@ -1,4 +1,16 @@
 = Appendix A: Building the Client
+:kde-craft-url: https://community.kde.org/Craft
+:kde-craft-build-from-source-url: https://community.kde.org/Guidelines_and_HOWTOs/Build_from_source/Windows
+:install-powershell-url: https://docs.microsoft.com/en-us/powershell/scripting/install/installing-windows-powershell?view=powershell-6
+:python-2_7-url: https://www.python.org/download/releases/2.7/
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/building.adoc
```diff
@@ -107,11 +107,131 @@ Contrary to earlier versions, version 1.7 and later are packaged as a `pkg` inst
 Do not call `make package` at any time when compiling for OS X, as this will build a disk image, which will not work correctly.
 ====
 
-[[windows-development-build]]
-== Windows Development Build
+[[windows-development-build-craft]]
```
settermjd: You don't need to add a custom anchor, as [they're generated automatically](https://asciidoctor.org/docs/user-manual/#auto-generated-ids).

dschmidt: I was just following the pattern from the rest of the file :man_shrugging:

dschmidt:
@ckamm yeah, it was about time :-)

docs/modules/ROOT/pages/building.adoc
settermjd: Fair call. It's a new best practice that I'm doing my best to phase in.

dschmidt:
@msetter Totally, as I said I'm not a technical writer - I'm happy if someone takes over with the right skills :)

PR #7038, opened 2019-02-14
Discovery on Windows: Fix detection of case-only renames

ckamm:
Previously they were detected as DELETE+NEW because if "a" is renamed to
"A" then QFile::exists("a") will still return true on Windows.

### DIFF
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -799,9 +799,14 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
                // Directories and virtual files don't need size/mtime equality
                || localEntry.isDirectory || localEntry.isVirtualFile);
 
+    auto originalPath = QString::fromUtf8(base._path);
     if (isMove) {
-        //  The old file must have been deleted.
-        isMove = !QFile::exists(_discoveryData->_localDir + base._path);
+        // The old file must have been deleted.
+        isMove = !QFile::exists(_discoveryData->_localDir + base._path)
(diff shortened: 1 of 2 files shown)


src/libsync/discovery.cpp
```diff
@@ -800,8 +800,14 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
                || localEntry.isDirectory || localEntry.isVirtualFile);
 
     if (isMove) {
-        //  The old file must have been deleted.
-        isMove = !QFile::exists(_discoveryData->_localDir + base._path);
+        // The old file must have been deleted.
+        isMove = !QFile::exists(_discoveryData->_localDir + base._path)
+#ifdef Q_OS_WIN
```
ogoffart: I believe we should use Utility::fsCasePreserving,  not?

ckamm: Yes

ckamm: Updated.

ogoffart:
```

../src/libsync/discovery.cpp:808:35: error: no member named 'compare' in 'QByteArray'
17s
70
                    && base._path.compare(path._local.toUtf8(), Qt::CaseInsensitive) == 0);
17s
71
                       ~~~~~~~~~~ ^

```

That was added in Qt 5.12, which we still don't require.  
before that, i guess `qustrcmp`

ckamm:
@ogoffart Updated to use QString::compare - qstricmp assumes latin1 encoding.

Issue #6977, opened 2019-01-16
Macos 10.4 , file virtualisation on Folders on client 2.5.2 stops after specific times of repeating 


lazawan:
Hi there ,
i was just testing the new client 2.5.2 on macOs 10.4 and looks good but there is a problem with virtualisation .
Virtualisation for normal files works fine but the actual problem is with Virtualisation for folders.

summary for simplifying : 
------------------------------   
there is a problem and you can reproduce it as following : 
make all files over Client virtual (while account-creating) 
go to the sync folder
make a folder available by downloading over owncloud-drop-menu
make the folder virtual again (now it goes wrong ,files become virtual but have the full size not 1kb)
download the folder again , here our game ends ,it won't be downloaded 

and that is the problem i meant .

regards,

ckamm:
@lazawan I couldn't reproduce it with master or 2.5 on linux. Concrete steps I tried:
- Set up a sync folder with vfs enabled
- Create new test/ and test/foo/ directories and test/file1, test/file2, test/foo/test3 files.
- Use the shell context menu to "Download files" and "Replace files by virtual file" on the test/ folder several times in sequence.

As far as I can observe the contained files switch between placeholders and content files as expected. Did your steps differ somewhere?

lazawan:
@ckamm i think ,bug was on Macos 10.4 reproducible , i can't remember  if it is on Linux as well

ckamm:
@lazawan If you can reproduce, could you generate a client logfile? Otherwise I might need someone on OSX to try.

lazawan:
@ckamm  sure , i will try to reproduce it and share log-file with you

lazawan:
@ckamm 

> 02-12 13:12:58:998 [ info gui.socketapi ]:	Received SocketAPI message <-- "GET_MENU_ITEMS:/Users/testy/ownCloud3/Photos" from SocketApiSocket(0x600003977bc0)
02-12 13:12:59:028 [ info gui.socketapi ]:	Sending SocketAPI message --> "GET_MENU_ITEMS:BEGIN" to SocketApiSocket(0x600003977bc0)
02-12 13:12:59:030 [ info gui.socketapi ]:	Sending SocketAPI message --> "MENU_ITEM:SHARE::Share..." to SocketApiSocket(0x600003977bc0)
[...]

PR #7040, opened 2019-02-15
Vfs: Fix folder dehydration requests on OSX

ckamm:
For #6977, already fixed in master.

Since the rename is triggered by the client process the folder watcher
didn't pick up on it on OSX.


ckamm:
Yes, there'll be conflicts since I had to rewrite the function for vfs handling in master. Just prefer the master code.

PR #7048, opened 2019-02-19
Client certs: Store pkcs12 in config, password in keychain

ckamm:
It still reads and writes the old format too, but all newly stored
client certs will be in the new form.

For #6776 because Windows limits credential data to 512 bytes in older
versions.

### DIFF
--- a/src/libsync/creds/httpcredentials.cpp
+++ b/src/libsync/creds/httpcredentials.cpp
@@ -42,6 +42,8 @@ Q_LOGGING_CATEGORY(lcHttpCredentials, "sync.credentials.http", QtInfoMsg)
 namespace {
     const char userC[] = "user";
     const char isOAuthC[] = "oauth";
+    const char clientCertBundleC[] = "clientCertPkcs12";
+    const char clientCertPasswordC[] = "_clientCertPassword";
     const char clientCertificatePEMC[] = "_clientCertificatePEM";
     const char clientKeyPEMC[] = "_clientKeyPEM";
     const char authenticationFailedC[] = "owncloud-authentication-failed";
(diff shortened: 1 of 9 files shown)


src/libsync/creds/httpcredentials.cpp
```diff
@@ -191,7 +196,20 @@ void HttpCredentials::fetchFromKeychain()
...
+        ReadPasswordJob *job = new ReadPasswordJob(Theme::instance()->appName());
+        addSettingsToJob(_account, job);
+        job->setInsecureFallback(false);
+        job->setKey(keychainKey(_account->url().toString(), _user + clientCertPasswordC, _account->id()));
+        connect(job, &Job::finished, this, &HttpCredentials::slotReadClientCertPasswordJobDone);
+        job->start();
+        return;
+    }
+
+    // Old case: Read client cert and then key from keychain
```
ogoffart: Maybe mention the version (< 1.6)

src/libsync/creds/httpcredentials.cpp
```diff
@@ -467,10 +517,31 @@ void HttpCredentials::persist()
...
+        job->setInsecureFallback(false);
+        connect(job, &Job::finished, this, &HttpCredentials::slotWriteClientCertPasswordJobDone);
+        job->setKey(keychainKey(_account->url().toString(), _user + clientCertPasswordC, _account->id()));
+        job->setBinaryData(_clientCertPassword);
+        job->start();
+        _clientCertBundle.clear();
+        _clientCertPassword.clear();
+    } else if (_account->credentialSetting(QLatin1String(clientCertBundleC)).isNull() && !_clientSslCertificate.isNull()) {
+        // Option 2, older configs: We used to store the raw cert/key in the keychain and
+        // still do so if no bundle is available.
```
ogoffart: So that means we are not migrating away from that setting?

ckamm:
I'd also prefer migrating. The issue is that I don't think we currently have functions for creating an encrypted pkcs12 bundle or some other strong symmetric encryption available.

Issue #7020, opened 2019-02-08
Big files do not sync - chunks do not sum up


kerlerm:
### Expected behaviour
Big files should sync.

### Actual behaviour
Bigger files do not sync when using the sync client 2.5.3 
Upload is possible by Webbrowser & GUI. No changes where made on the server since 10/2018 and I synced files sized about 320 MB in 12/2018. No quota is applied to the user. 

### Steps to reproduce
Take file "test" with about 100-150 MB and initiate sync. Error "Chunks on server do not sum up to ..." is thrown after upload of chunks is complete.

### Server configuration

Operating system:
Linux

Web server: 
Apache

Database:
MySQL 5.7

PHP version:
7.2 FPM 

ownCloud version:
10.0.10.4

Storage backend (external storage):

### Client configuration

Client version:
2.5.3 build 342

Operating system:
Linux 64

OS language:
German

Qt version used by client package (Linux only, see also Settings dialog):
Qt 5.10.1

Client package (From ownCloud or distro) (Linux only):
GIT-Revision 80478b auf Feb 7 2019, 17:25:04 verwendet Qt 5.10.1, OpenSSL 1.0.2n 7 Dec 2017

Installation path of client:
/usr/bin/owncloud

### Logs

1. Client logfile: Output of `owncloud --logwindow` or `owncloud --logfile log.txt`
https://gist.githubusercontent.com/kerlerm/bb259a92d63e4b926b7ad3fe30ab0f62/raw/fd2740e5ffebfbef2d608ce28613317c3e76ee68/client_log.txt

2. Web server error log:
https://gist.githubusercontent.com/kerlerm/9bc1dfcc073be59966b2c2d70592f29a/raw/4b60d3d6744c623c8ce43587bdea8bd3784d3989/gistfile1.txt

3. Server logfile: ownCloud log (data/owncloud.log):
https://gist.githubusercontent.com/kerlerm/99f1ae603170e8ce0669987cc54cf055/raw/0711f8b296e7a68a902125dd0216ce98fa9d487b/server_log.txt

michaelstingl:
@kerlerm Do you know the version that worked? `2.5.1`? `2.4.3`?

kerlerm:
Unfortunately, I did not follow installed versions on my machine. I assume 2.4.3 still worked, 2.5.2 and 2.5.3 certainly do not. Should I try downgrading to verify?

ckamm:
@kerlerm Thanks for the report and for including client logs!

[...]

PR #7053, opened 2019-02-22
Disable HTTP2 by default

ckamm:
Due to QTBUG-73947 and #7020.

Use OWNCLOUD_HTTP2_ENABLED=1 to enable anyway.

### DIFF
--- a/src/libsync/accessmanager.cpp
+++ b/src/libsync/accessmanager.cpp
@@ -96,7 +96,11 @@ QNetworkReply *AccessManager::createRequest(QNetworkAccessManager::Operation op,
 #if QT_VERSION >= QT_VERSION_CHECK(5, 9, 4)
     // only enable HTTP2 with Qt 5.9.4 because old Qt have too many bugs (e.g. QTBUG-64359 is fixed in >= Qt 5.9.4)
     if (newRequest.url().scheme() == "https") { // Not for "http": QTBUG-61397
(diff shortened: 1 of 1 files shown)


guruz:
@ckamm So why does the server even send a GOAWAY? Is this just happening randomly or is this a regular thing?
Is it feasible to keep HTTP2 enabled, but as soon as we detect this kind of error (heuristic) disable it (for this run)?

ckamm:
The server may send GOAWAY frames to signal it wants to gracefully shut down the connection. This happens frequently for the server I was testing with.

Yes, I think we could detect the error: either by QNetworkReply::error() being emitted with NoError or by finished() being emitted without the body being fully consumed, content-length not matching the actualy body length or reply headers not having been received.

ckamm:
@guruz I've investigated, but would prefer outright disabling http2 for now. 

It's easy to catch QNetworkReply::error(NoError) emissions and we could do something unsanitary to adjust the value returned by QNetworkReply::error(), making our existing logic in the finished() slots work again. But I'm worried that there might be other, less critical cases where the error() signal is emitted with NoError that we're just not aware of. If we always translate NoError to SomeError, that'd break these cases.

Unfortunately the HTTP2WasUsedAttribute is not set for these error(NoError) emissions; filtering that way doesn't work.

guruz:
@ckamm Please create a follow-up issue in 2.5.x milestone so we re-evaluate if Qt 5.12.* fixes this..

guruz:
Or maybe not needed if we have ther other issue

Issue #7023, opened 2019-02-10
[Windows client] Can not copy path while owncloud client is running


xtianus:
### Expected behaviour
When I click on the "Copy path" button in Explorer, I should be able to copy the path of a file or folder anywhere on my drive.

![image](https://user-images.githubusercontent.com/7997535/52534143-f77e1100-2d3d-11e9-98bd-8d6ec8f9115f.png)

### Actual behaviour
If the owncloud client is running and the file or folder is within the owncloud repository tree, nothing is copied.

### Steps to reproduce
1. Start the owncloud client on windows 10
2. Browse to a folder inside the owncloud repository using Windows Explorer
3. Click on the "Copy path" button
4. Paste in a text editor: nothing is pasted

### Server configuration
Operating system: Windows 10 Pro version 1803 (OS Build 17134,523)

ownCloud version: Version 2.5.3 (build 11250)

### Logs

```
02-10 14:07:13:676 [ info gui.socketapi ]:	New connection QLocalSocket(0x8590848)
02-10 14:07:13:676 [ info gui.socketapi ]:	Sending SocketAPI message --> "REGISTER_PATH:C:\\ownCloud" to QLocalSocket(0x8590848)
02-10 14:07:13:677 [ info gui.socketapi ]:	Received SocketAPI message <-- "GET_STRINGS:CONTEXT_MENU_TITLE" from QLocalSocket(0x8590848)
02-10 14:07:13:677 [ info gui.socketapi ]:	Sending SocketAPI message --> "GET_STRINGS:BEGIN" to QLocalSocket(0x8590848)
02-10 14:07:13:677 [ info gui.socketapi ]:	Sending SocketAPI message --> "STRING:CONTEXT_MENU_TITLE:ownCloud" to QLocalSocket(0x8590848)
02-10 14:07:13:677 [ info gui.socketapi ]:	Sending SocketAPI message --> "GET_STRINGS:END" to QLocalSocket(0x8590848)
02-10 14:07:13:677 [ info gui.socketapi ]:	Received SocketAPI message <-- "GET_MENU_ITEMS:C:\\ownCloud\\Studio\\Programmi" from QLocalSocket(0x8590848)
02-10 14:07:13:678 [ info gui.socketapi ]:	Sending SocketAPI message --> "GET_MENU_ITEMS:BEGIN" to QLocalSocket(0x8590848)
[...]

Issue #7004, opened 2019-01-26
[Windows] Cannot open pinned OwnCloud folder or Recent Folder ownCloud folder from Windows taskbar Explorer right click menu while OwnCloud client is running


NoNaym:
I like to right click on the pinned Windows Explorer icon that is on the Windows taskbar in order to open one of my pinned (or “Quick access”) folders. Recently I noticed that while the OwnCloud Windows client is running (on Windows 10), my primary owncloud folder (and any other folder inside) which I have pinned to Windows Explorer refuse to open when right clicking on the Window Explorer taskbar icon and then clicking on the applicable folder (i.e. owncloud folder or any folder within). Nothing happens at all. All other (non-owncloud) folders open without issue. If I kill the OwnCloud client, then right clicking the Windows Explorer taskbar icon and selecting one of my owncloud folders work (the folder opens in Windows Explorer normally)!

This leads me to believe that the problem is being caused by the Windows OwnCloud client, and not Windows itself. Has anybody else experienced this issue? Anybody know of a fix? And/or does the OwnCloud team have any plans on fixing this issue?

Client OS: Windows 10 Pro (x64)

Windows ownCloud client version: 2.5.1 (build 10807)

OS language: US English

Installation path of client: C:\Program Files (x86)\ownCloud

guruz:
@NoNaym Would you say this is related to #6982 ?

guruz:
@NoNaym Could you also run the client with logging, especially checking for the ```SocketAPI``` related messages?

NoNaym:
Yes, I would say these issues are related. They sounds very similar.

NoNaym:
I'll try to do this as soon as I can and will post the log shortly...

MikeKnife:
I have an issue, that might be related to the one above.
It started when I switched from 
ownCloud-2.4.3.10188-setup.exe 
to 
ownCloud-2.5.0.10359-setup.exe.

I use to have only shortcuts of simple text files (program code) on the desktop. Right click and choose to open it with Scite (auto-it-editor) or with Ultra Edit Studio opens the file. (The entries for this come with the editors.)
Since the last owncloud version this function is gone, nothing happens. Clicking on the file itself instead of its shortcut still works though.
In my case deselect the option "Integration in den Windows Explorer" during install fixed the problem.

Mike Knife

guruz:
Pinned folder:
```
[...]

PR #7056, opened 2019-02-27
Windows: Fix context menu handling only own verbs

ckamm:
Previously it'd handle all verbs as if they were our own.

For #7004, #7023


guruz:
Nice catch!

shell_integration/windows/OCContextMenu/OCContextMenu.cpp
```diff
@@ -182,20 +182,38 @@ IFACEMETHODIMP OCContextMenu::InvokeCommand(LPCMINVOKECOMMANDINFO pici)
 {
     std::wstring command;
 
+    CMINVOKECOMMANDINFOEX *piciEx = nullptr;
+    if (pici->cbSize == sizeof(CMINVOKECOMMANDINFOEX))
```
guruz: maybe could be a >= check instead in case Microsoft introduces an EX2 structure at a later point?

Issue #7015, opened 2019-02-05
Possible propagation bug for conflicts when the temporary can't be renamed


ckamm:
The download propagation does this:
```
    if (!FileSystem::uncheckedRenameReplace(...)) {
        ...
        // If we moved away the original file due to a conflict but can't
        // put the downloaded file in its place, we are in a bad spot:
        // If we do nothing the next sync run will assume the user deleted
        // the file!
        // To avoid that, the file is removed from the metadata table entirely
        // which makes it look like we're just about to initially download
        // it.
        if (isConflict) {
            propagator()->_journal->deleteFileRecord(fn);
```

But this shouldn't work since `fn` is the filesystem path, not the path to the db item. Needs fixing and probably tests!

ogoffart:
I guess the scenario to test is the one from https://github.com/owncloud/client/issues/2428

PR #7059, opened 2019-02-28
Download: Remove useless code and add a test

ogoffart:
From issue #7015, the code is wrong because the path is the file system path and
not the path on the DB.
But since this is a conflict, this means the reconcile will still want to download
the file from the server next sync, so we need not to worry about this case

### DIFF
--- a/test/testdownload.cpp
+++ b/test/testdownload.cpp
@@ -8,6 +8,7 @@
 #include <QtTest>
 #include "syncenginetestutils.h"
 #include <syncengine.h>
+#include <owncloudpropagator.h>
 
 using namespace OCC;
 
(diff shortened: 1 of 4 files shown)


test/testdownload.cpp
```diff
@@ -145,6 +146,63 @@ private slots:
         QCOMPARE(getItem(completeSpy, "A/broken")->_status, SyncFileItem::FatalError);
         QVERIFY(getItem(completeSpy, "A/broken")->_errorString.contains("System in maintenance mode"));
     }
+
+    void testConflictAndLocked() {
+        // Test for https://github.com/owncloud/client/issues/7015
+        // We want to test the case in which the renaming of the original to the clonflict file succeeds,
```
ckamm: conflict

test/testdownload.cpp
```diff
@@ -145,6 +146,63 @@ private slots:
...
+        // but renaming the temporary file fails.
+        // This tests use the fact that a "touchedFile" notification will be sent at the right moment.
```
ckamm: uses

test/testdownload.cpp
```diff
@@ -145,6 +146,63 @@ private slots:
...
+                if (s.contains("conflicted copy")) {
+                    QCOMPARE(conflictFile, QString());
+                    conflictFile = s;
+                    return;
+                }
+                if (!conflictFile.isEmpty()) {
+                    // Check that the temporary file is still there
+                    QCOMPARE(QDir(fakeFolder.localPath() + "A/").entryList({"*.~*"}, QDir::Files | QDir::Hidden).count(), 1);
+                    // Set the permission to read only on the folder, so the rename of the temporary file will fail
+                    QFile(fakeFolder.localPath() + "A/").setPermissions(QFile::Permissions(0x5555));
```
ckamm: I'm not sure whether the test will succeed on windows.

ogoffart: i'm gonna see what appveyor says

ogoffart: Appveyer says it is not working!

What can i do?  Disable that test on windows? 
Another trick i can use?

ogoffart: Skipping the test on windows

Issue #7054, opened 2019-02-26
Reliably persist cookies


corneliusweiss:
The client looses cookies after close and reopen. They can be found in the cookies.db but the client doesn't send them.

### Expected behaviour
When a Load Balancer or Proxy adds a cookie it should be preserved/send by the client.

### Actual behaviour
The cookie is not send to the server after the client got closed and opened again.

### Steps to reproduce
1. introduce cookie via proxy or lb
2. watch traffic
3. cookie is lost after quitting and opening client

Please see related discussion in #2808

guruz:
@corneliusweiss Thanks for creating the bug, but you really need to provide version numbers etc, else how can we know you are using a current client version?

As @michaelstingl wrote in the comment already:
> What client version? What operating system?

guruz:
@corneliusweiss Are you sure your proxy/loadbalancer is sending a persistant cookie?
http://www.allaboutcookies.org/cookies/cookies-the-same.html
(Although I don't remember how strict is cookies.db about this)

corneliusweiss:
sorry for missing this. I testet with Version 2.5.1 (build 10818) on macOS (is think it's the latest version from autoupdates)

And yes, the cookie is persistent. Here is the sniffed cookie from the lb:
    Set-Cookie: be=app03; expires=Wed, 01-Jan-2020 00:00:00 GMT; Max-Age=26491080; path=/

I already played with the path option - but w.o. success

ogoffart:
I can reproduce.

guruz:
@corneliusweiss Want to test a current daily build? https://download.owncloud.com/desktop/daily/

corneliusweiss:
sure, but which is the right one? testpilotclient? which version to test? I need a macOS package.

guruz:
@corneliusweiss https://download.owncloud.com/desktop/testing/ownCloud-2.5.4.11420rc1.pkg

corneliusweiss:
@guruz works like a charm 
Thanks a lot!

guruz:
Great!

PR #7060, opened 2019-03-01
AccountManager: load the cookies

ogoffart:
For issue #7054

### DIFF
--- a/src/gui/accountmanager.cpp
+++ b/src/gui/accountmanager.cpp
@@ -16,6 +16,7 @@
 #include "configfile.h"
 #include "sslerrordialog.h"
 #include "proxyauthhandler.h"
+#include "common/asserts.h"
 #include <theme.h>
 #include <creds/credentialsfactory.h>
 #include <creds/abstractcredentials.h>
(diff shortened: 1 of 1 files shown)


ogoffart:
Note:  when re-authenticating, we send a request without any cookies in any case:

https://github.com/owncloud/client/blob/4d0e4c0a18ace992c641f7805b61f611301285c6/src/libsync/networkjobs.cpp#L853

But this should not be too often.

src/gui/accountmanager.cpp
```diff
@@ -81,9 +82,13 @@ bool AccountManager::restore()
         if (!skipSettingsKeys.contains(settings->group())) {
             if (auto acc = loadAccountHelper(*settings)) {
                 acc->_id = accountId;
-                if (auto accState = AccountState::loadFromSettings(acc, *settings)) {
-                    addAccountState(accState);
-                }
+                auto accState = AccountState::loadFromSettings(acc, *settings);
```
ckamm: I'd prefer if you kept the `if`. Currently loading AccountStates can't fail, but it might in the future. (Otherwise I'd use ENFORCE here, since adding a null AccountState to the list would lead to trouble down the line.)

Issue #6987, opened 2019-01-17
[Linux] trailing newline or carriage in filename are neither rejected nor handled correctly


jnweiger:
owncloud-client 2.5.2rc1 on Linux mint happily uploads files with trailing newline 0x0a or carriage return 0x0d to the owncloud server. The server web ui handles them, and replaces them with an _ when requesting a download via web.
owncloud-client 2.5.2rc1 on windows fails with error messages.
A trailing carriage return causes "Access denied" (very misleading)
A trailing newline causes "The filename, directory name, or volume label syntax is incorrect"

![image](https://user-images.githubusercontent.com/1108546/51325620-a136ee80-1a6d-11e9-91ad-5d17de09c35a.png)

ogoffart:
You seem to describe a problem on the server rather than a problem on the client.

Btw, When I tried with the demo server, i get the error `File name contains at least one invalid character` which seems correct.

ckamm:
I can reproduce the problem, "foo\n" uploads successfully (created with `touch $'foo\n'` in zsh). We could add `\x00-\x1F` to our invalid filename regex? Or just `\R`? Opinions? @jnweiger @ogoffart @guruz 

The current pattern is `[\\:?*"<>|]`

guruz:
I'd say "Anything in the unicode control characters category" but I'm not an expert.

ckamm:
We have two options here:
1) consider this invalid early, not even uploading these files
2) consider this invalid filenames for downloading to Windows systems

We currently do a mix of both. The pattern I mentioned above is for 1) but there are also some 2)-style filters (like windows reserved words).

For control characters I'd suggest 1) since they also not great to have on the server.

guruz:
@jnweiger Please test with rc1

jnweiger:
## tested 254rc1 on win10
 A trailing newline causes an entry as not synced with the message "The filename is a reserved name in this filesystem" -> green checkmark indicates a successful sync nevertheless ->
OK

A trailing carriage return causes the file to be ignored without entry in the not synced.
Renaming a "foo.txt" to "foo\r" on the server (via a different client) causes the windows client to simply delete the file locally.
 -> green checkmark indicates a successful sync nevertheless ->
BAD -> There should the same entry in the not synced tab as with a trailing.

ckamm:
@jnweiger Nice catch, there files should definitely appear in the not-synced tab! It's very odd, because there's no distinction between \n and \r in the code. I'll try to reproduce.

[...]

PR #7062, opened 2019-03-01
Windows: Forbid chars 0-31 in filenames 

ckamm:
For #6987


ckamm:
@guruz We might make that change in 2.6, but it could break behavior for people who're simply syncing between linux clients. The change here just improves the error messaging for Windows clients.

Issue #6983, opened 2019-01-17
Windows SocketAPI named pipe path has fixed "ownCloud" string


ckamm:
See the `SocketAPI` constructor. This makes the shell integrations not work properly if branded or derived (nextcloud) clients are running at the same time.

guruz:
Windows only indeed..
On macOS we use the app domain name: https://github.com/owncloud/client/blob/master/src/gui/socketapi.cpp#L188

I remember there was a reason for this.. maybe because the Windows plugin was built in the 'binary' submodule before, but now that we switched to MSVC we could always build it with the branding instead?

CC @ogoffart @jturcotte @dschmidt

dschmidt:
FTR probably `APPLICATION_SHORTNAME` or `Theme::appName()` should be used instead. The shell extension needs to be adjusted as well then probably.

The reason why this was not branded in the first place was probably that we used to ship binaries of the shell extension instead of building them when building the rest of the client. That made branding it impossible and so always the hardcoded ownCloud name needed to be used. This is obviously not the case anymore.

guruz:
@dschmidt commented at the same time :D

guruz:
Hm, the commit above only makes sense if we also change the DLL names, e.g.
```OCOverlays.dll``` -> ```ownCloudOverlays.dll``` 
This would involve more changes, for example in the Makefiles and in MSI.
Postponing to 2.5.4?
CC @ogoffart @dschmidt

dschmidt:
@guruz What does this have to do with dll names?

guruz:
Our goal is to have completely different brandings, e.g. if ```CERNBoxOverlays.dll``` and ```ownCloudOverlays.dll``` diverge.
(Also looks less mysterious as a DLL name)

guruz:
I'm not so sure. I thought we use a single global instance of the DLL (via some identifier).
I think @ogoffart @jturcotte know this stuff, I don't.

guruz:
OK maybe @dschmidt is right, I created a PR for 2.6

HanaGemela:
I have the below clients running at the same time:
Testpilot 2.6.0daily20190429 (build 11645) 
ownCloud (not branded) 2.6.0daily20190429 (build 11644) 

ownCloud client doesn't have the ownCloud submenu

![Untitled](https://user-images.githubusercontent.com/49001702/56897089-da7bf300-6a8d-11e9-9219-8c46f5809166.png)

HanaGemela:
[...]

PR #7066, opened 2019-03-04
Windows: Use better name for SocketAPI socket #6983

guruz:


ogoffart:
Why the two dashes?

guruz:
@ogoffart Good question. I forgot, I had resurected that commit from somewhere. Maybe to be 100% sure we're not clashing with the old version (which should be compatible though)

guruz:
@ogoffart Can we get this merged so we can test  https://github.com/owncloud/enterprise/issues/3209 ?

Issue #6677, opened 2018-07-30
Owncloud deletes files & tries to sync again (with little success)


Phlos:
I have a 250 Gb storage allowance through a university system which I use to sync my files between several computers that I work on. The university system supplies a tailored client for windows, but for linux I use OwnCloud. I noticed in the last few days that OwnCloud is deleting files from a local computer and then tries to redownload them. This is an issue I have had several times in the past and somehow keeps popping up. **[edit: That it started happening again is possibly related to the fact that I updated to 2.4.2 on 19 July).]** I have only experienced this on my linux machines, never with Windows (although it should be noted that I sync only a rather small subset of my stuff to the Windows OS).

### Expected behaviour

I have the OwnCloud client on my laptop. It syncs my files. If I am offline for a while, it syncs everything as soon as I'm back online.

### Actual behaviour

I have the OwnCloud client on my laptop. It deletes files that I had on the local machine _and_ on the server from the local machine, then tries to redownload them. 

The sync for redownloading is not unproblematic:

- at some point, it stops downloading and loses the connection, and then sits stuck -- as far as I can see, until I forcefully do something about it. The connection issue is not solved by clicking log out and log in alone (as far as I can see); I have to restart owncloud and then it connects.
- at each of the retries, it tells me for a different set of directories that "a new directory > 500MB has been discovered" which I should approve the download of  (500 MB is the prompt limit in my client). 
- This is especially problematic because in general (also when OC is not deleting files) it takes ages to initialise OC after a restart - it takes absolutely forever to work through all the directories and files to check whether any changes have taken place. 
- Owncloud has, up until now, not realised the new folders that I have created during my time offline. 

Consequently, I wait for hours to get a sync that maybe 30 minutes, after which it stops working, I restart everything, etc.

### Steps to reproduce

Since this is not a very straightforward issue, I'm afraid there are no straightforward steps either. I did the following, though:
1. Have owncloud on my laptop (Ubuntu 16.04; currently I have OC 2.4.2)
2. Work on it without internet for a while. 
3. Connect back to the internet
4. Discover the file & directory deletion mayhem.

### Server configuration

Unfortunately, I do not know what specs the university system has.

### Client configuration
Client version: 2.4.2

Operating system: Ubuntu 16.04

OS language: English (United States)

Qt version used by client package (Linux only, see also Settings dialog): 
[...]

PR #7067, opened 2019-03-04
OwnSql: Distinguish no-data from error 

ckamm:
For #6677, fyi @Phlos

This could fix a problem where the client incorrectly decides to delete
local data.

Previously any sqlite3_step() return value that wasn't SQLITE_ROW would
be interpreted as "there's no more data here". Thus an sqlite error at a
bad time could cause the remote discovery to fail to read an unchanged
subtree from the database. These files would then be deleted locally.

With this change sqlite errors from sqlite3_step are detected and
logged. For the particular case of SyncJournalDb::getFilesBelowPath()
the error will now be propagated and the sync run will fail instead of
performing spurious deletes.

Note that many other database functions still don't distinguish
not-found from error cases. Most of them won't have as severe effects on
affected sync runs though.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -4,6 +4,7 @@ ChangeLog
 version 2.5.4 (unreleased)
 * Crash fix: Infinite recursion for bad paths on Windows (#7041)
 * Crash fix: SocketApi mustn't send if readyRead happens after disconnected (#7044)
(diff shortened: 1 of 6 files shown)


ckamm:
Thanks for the review. The AppVeyor build failure looks unrelated.

Issue #7068, opened 2019-03-05
New folders aren't watched immediately, potentially missing notifications


ckamm:
Only on Linux:

1. Create a new folder, wait for a sync to start and finish local discovery.
2. Before the new folder is processed and gets an inotify watch, add files to it.
3. The client sync finishes without syncing the new files (as expected)
4. No new sync will be scheduled and the next sync won't locally discover the new folder (bad)

This will be particularly bad if the user moved files into the new folder within the failure window. Then the next sync will assume these files were deleted and delete them on the server. Only the next full-discovery sync would upload the moved files and correct the issue.

FYI @PVince81

guruz:
@PVince81 should be fixed in https://download.owncloud.com/desktop/testing/owncloud-client-2.5.4.11614.484rc1.linux-repo.html

lazawan:
@ckamm i tried to reproduce the bug on client 2.5.3 but i did not manage it (on ubuntu 16.04)
here what i did : 
i limited the bandwidth in order that sync is slower 
i pasted a folder with files into sync folder 
during sync i added files into the new pasted folder 
but the files that i added afterwards ,were synchronized too 
did i do that what you meant ?

ckamm:
@lazawan That depends on when exactly you pasted the new files, it needs to happen before the creation of the new folder on the server is completed. Propagation of the contents of the folder happens afterwards. Possibly adding large files to the sync root might do it, but it could depend on job ordering. I'll look more closely tomorrow.

PR #7069, opened 2019-03-05
FolderWatcher: Work around missing notifications on linux 

ckamm:
There was a period of time between local discovery of a new folder and
the folder receiving a filesystem watcher. Any changes to its contents
during that time were missing, leading to misbehaviors.

For #7068

Note: I'm working on a nicer solution on the folderwatcher level, but that may be too large or not in time for 2.5.4.

### DIFF
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -925,15 +925,30 @@ void Folder::slotItemCompleted(const SyncFileItemPtr &item)
         return;
     }
 
-    // add new directories or remove gone away dirs to the watcher
+#ifdef Q_OS_LINUX
+    // The linux folderwatcher isn't recursive: it can just watch individual folders.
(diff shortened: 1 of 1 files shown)


src/gui/folder.cpp
```diff
@@ -925,15 +925,30 @@ void Folder::slotItemCompleted(const SyncFileItemPtr &item)
         return;
     }
 
-    // add new directories or remove gone away dirs to the watcher
+#ifdef Q_OS_LINUX
```
ogoffart: So mac and windows are already recursive?

ckamm: Yes, I double checked. Their addFolder/removeFolder functions are literally {} (and the elaborate follow up patch removes them and makes the linux folderwatcher recursive too).

PR #7070, opened 2019-03-05
FolderWatcher linux: Make automatically recursive

ckamm:
Previously it depended on addFolder() / removeFolder() calls to adjust
watchers when new folders were added or removed. There also needed to be
complex move handling.

Now, any folder creation/move-in notifications automatically trigger
watcher additions and folder deletion/move-out triggers removal.

For  #7068 and a replacement of the 2.5 workaround #7069. I think this is too high risk for 2.5.4.

### DIFF
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -1061,16 +1061,6 @@ void Folder::slotItemCompleted(const SyncFileItemPtr &item)
         return;
     }
 
-    // add new directories or remove gone away dirs to the watcher
-    if (item->isDirectory() && item->_instruction == CSYNC_INSTRUCTION_NEW) {
-        if (_folderWatcher)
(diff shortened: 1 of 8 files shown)


ogoffart:
Relying on the syncengine to tell which folder to watch meant that ignored folders were not watched, which was a good thing.
Now, maybe we could hook the exclude path engine in there so we do not recurse in ignored folder?  Or we can ignore this issue.

ckamm:
@ogoffart There already is a `!_parent->pathIsIgnored(p)` in the folder watcher to exclude ignored files. I imagine selective-sync excluded folders would now be watched though... I'll check that out.

ckamm:
Since now linux works the same way OSX and Windows do file watching I'm not going to do extra changes for selective sync exclusions. Looks right to me as-is.

Issue #6518, opened 2018-05-12
ownCloud client icon not shown in Ubuntu 18.04


samiux:
### Expected behaviour
ownCloud client icon show on Ubuntu 18.04 panel when gnome-shell-extension-top-icons-plus is installed and enabled.

### Actual behaviour
ownCloud client is install from ownCloud official site via respo.  After enabled start when startup, the icon could not be shown on the panel in Ubuntu 18.04 even gnome-shell-extension-top-icons-plus is installed and enabled.

### Steps to reproduce
1. sudo apt install gnome-shell-extension-top-icons-plus
2. reboot and enable the extension
3. reboot but the ownCloud icon does not show on the panel

michaelstingl:
@samiux Thank you for reporting the issue. Which exact version of the ownCloud client are you using? Did you install from the [ownCloud repos](https://software.opensuse.org/download/package?project=isv:ownCloud:desktop&package=owncloud-client) or from the Ubuntu repos?

samiux:
@michaelstingl 

The ownCloud client is from both ownCloud's Ubuntu repos and Ubuntu's official repos.  After several experiments, I confirmed that it cannot work with Gnome Shell Extension only.  Without Gnome Shell Extension, it is working flawless.

ckamm:
@samiux It seems like top-icons-plus isn't in development anymore and might not work with 18.04, see https://extensions.gnome.org/extension/1031/topicons/ https://github.com/phocean/TopIcons-plus - it's quite likely that this is not a ownCloud client issue.

samiux:
@ckamm 

I have tested ownCloud client without "top icons plus" but with other Gnome Shell Extensions.  When Gnome Shell Extensions are enabled, ownCloud client will not be displayed.

I need to disable Gnome Shell Extension feature in "Gnome Tweaks".

Any idea?

ckamm:
@samiux This really sounds like an issue with the software managing the icons. Are there other programs with not gnome-native tray icons that don't have this problem?

samiux:
@ckamm 

I use RadioTray-NG (https://github.com/ebruck/radiotray-ng) which works flawlessly with and without Gnome Shell Extension.  Sometimes, indicator-cpufreq works properly with Gnome Shell Extension.

ckamm:
Okay, hmm. RadioTray-NG uses gtk and appindicator. 

I don't know what that "Gnome Shell Extension" option does.

[...]

PR #7072, opened 2019-03-06
Tray: Try to establish tray after 10s if failed initially

ckamm:
When owncloud is started during desktop startup the tray may not yet
be running when the client starts. This will make the client attempt
to create a tray icon again after 10 seconds if there's no tray
during initial startup.

For  #6518

### DIFF
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -741,6 +741,13 @@ void Application::openVirtualFile(const QString &filename)
     });
 }
 
+void Application::tryTrayAgain()
+{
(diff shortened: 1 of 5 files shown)


ogoffart:
This looks like black magic to me.  Has this been tested by users with problems?

ckamm:
@ogoffart No. It's a guess based on the symptoms presented in #6518. Since it won't hurt to do a `hide(); show();` after 10s if the tray is marked as unavailable I'd like to include it. I believe it has a good chance of making the icon appear in the slow-tray case.

Issue #7061, opened 2019-03-01
"Copy public link to clipboard" should autocreate a link share even if expiry is required


ckamm:
Follow up from #6356 due to concerns raised by @lexo-mfleuti. See https://github.com/owncloud/client/issues/6356#issuecomment-468580876.

Currently the client doesn't autocreate a link share if expiry is required. It could do it though and use the specified default expiry offset.

lexo-mfleuti:
**Context menu idea**
Screenshot: draft of how it could look like
![owncloud-context-menu-draft](https://user-images.githubusercontent.com/8611779/53632812-23dace00-3c16-11e9-98d2-97831f78c1da.jpg)

**Configurable context menu**
It would be great if the user could select which context menu elements are needed to keep the context menu as compact (efficient) as possible.
![tortoise option dialog](https://user-images.githubusercontent.com/8611779/53632864-42d96000-3c16-11e9-8be5-449446088442.jpg)

michaelstingl:
> Currently the client doesn't autocreate a link share if expiry is required. It could do it though and use the specified default expiry offset.

@ckamm works fo me!

2nd step could be, to bring all menu entries to the root level of the context menu. Visually I think this is a mess, but for efficiency it totally makes sense…

guruz:
@lexo-mfleuti Can you test the shares with expiry using https://download.owncloud.com/desktop/testing/ownCloud-2.5.4.11614.11431-rc1.msi and report back if it works?
For your other comments (configurable context menu), I think it's better to create a new issue as this one here is really just for the bugfix of the link creation (thanks!).

lexo-mfleuti:
@guruz Just tried it out. The bug is fixed. Very nice thank you! I can now share files much faster already. Though it would be very cool if you could also change the layout of the context menu to something like I explained above. That would be the dream ;)

michaelstingl:
Is the link creator aware that the created link will silently disappear after a while?

lexo-mfleuti:
No, what do you mean by "silently disappear"? You mean like "removed from clipboard"?

michaelstingl:
In this case, a max. expiration date has been set by the ownCloud server admin. The user then creates a link with the desktop client that's available in the clipboard. But the user isn't aware the link has an expiration date. After expiration, the link disappears and is no longer valid.

@ckamm adding the expiration date/period to the context menu entry?

ckamm:
[...]

PR #7074, opened 2019-03-06
Shares: "copy link" action can create shares with expiry

ckamm:
Previously it gave up if "expiry required" was enabled. Now it'll create
a link share per day with the default expiry for these setups.

For  #7061

@michaelstingl I think this is safe enough that we could include it in 2.5.4 if desired.


michaelstingl:
> @michaelstingl I think this is safe enough that we could include it in 2.5.4 if desired.

👍

Issue #7027, opened 2019-02-11
virtual-files feature on Owncloud client v2.6.0 with Qt 5.12.1 is missing completely


lazawan:
simply VFS is not there :-`(  " i mean on all linux Platforms )

regards,

ogoffart:
It is!

Maybe you are using a branding that disable it?

lazawan:
@ogoffart  no i used the daily build for 2.6.0 client , i'm aware that branded-client doesn't have VFS

ckamm:
Maybe the client can't open any of the vfs plugins. @lazawan Could you give me the first 20 lines of the client log file? It should say "VFS windows plugin is available" and "VFS suffix plugin is available".

jnweiger:
Same here: 2.6.0 daily 20190211 on Linux Mint, Ubuntu 18.04

* The firstrun wizard no longer mentions any expermimental features.
* The three dots menu no longer allows switching to/from virtual mode.
* renaming a file to end in .owncloud just renames the file. It is uploaded to the server with this suffix.

Is there now a VFS plugin that is needed for Linux? Maybe packagaging needs to adapt for structural changes?

ckamm:
@jnweiger Does the package contain the libowncloudsync_vfs_* shared libraries? @dschmidt

jnweiger:
No. Packaging cannot pick up new additions. I need to locate everything manually. Thanks for the hint, I have to ho searching now..

ckamm:
@jnweiger The plugin is called "owncloudsync_vfs_suffix.so" built from "src/libsync/vfs/suffix" and should be installed next to the binary.

jnweiger:
DEB packages have it in the build-log
[  368s] [ 21%] Linking CXX shared library ../../../../bin/owncloudsync_vfs_suffix.so
[ 1182s] -- Installing: /usr/src/packages/BUILD/debian/tmp/opt/ownCloud/ownCloud/lib/x86_64-linux-gnu/plugins/owncloudsync_vfs_suffix.so

but it neither breaks packages with "unpackaged file error", nor is it auto-included anywhere. That is odd.

jnweiger:
Is there a simple test to see where the plugin is searched at runtime?
should e.g. `strace owncloudcmd -v` show something?

jnweiger:
It gets loaded, if placed in 
```
ll /opt/ownCloud/ownCloud/bin/
insgesamt 2604
[...]

PR #7076, opened 2019-03-07
Add OC_EXTRA_PLUGIN_PATH cmake and define 

ckamm:
By default, plugins are only searched next to the binary or next to the
other Qt plugins. This optional build variable allows another path to be
configured.

The idea is that on linux the oC packaging probably wants the binary in
something like /opt/owncloud/bin and the plugins in
/opt/owncloud/lib/plugins.

Similarly, distribution packagers probably don't want the plugins next
to the binary or next to the other Qt plugins. This flag allows them to
configure another path that the executable will look in.

For #7027


src/gui/application.cpp
```diff
@@ -254,6 +254,15 @@ Application::Application(int &argc, char **argv)
         AbstractNetworkJob::httpTimeout = cfg.timeout();
 
     // Check vfs plugins
+#define MACRO_TO_STRING2(x) #x
```
ogoffart: maybe ifdef OC_EXTRA_PLUGIN_PATH,  and only add the -D if it is set

CMakeLists.txt
```diff
@@ -102,6 +102,9 @@ if(NO_MSG_HANDLER)
    add_definitions(-DNO_MSG_HANDLER=1)
 endif()
 
+set(OC_EXTRA_PLUGIN_PATH "" CACHE STRING "Extra path to look for Qt plugins")
+add_definitions(-DOC_EXTRA_PLUGIN_PATH=${OC_EXTRA_PLUGIN_PATH})
```
dschmidt: Please don't do it like this. If you absolutely want to use -D instead of putting it in a config.h file, then only apply it to the relevant target not to all.

dschmidt: We should probably also define a sane default that just works.

ckamm: I'll make these changes. What default value would you suggest? `/usr/lib/owncloud-client`? I imagine it might be different per distro.

ckamm:
@ogoffart I think for most linux packaging we'd want an absolute path. But I'll also allow relative ones, could be useful for osx bundles.

ckamm:
updated

jnweiger:
Could that be relative to our LIBDIR, so if LIBDIR is /opt/ownCloud/ownCloud/lib/x86_64-linux-gnu/
it would become /opt/ownCloud/ownCloud/lib/x86_64-linux-gnu/plugins ?

ckamm:
@jnweiger I could set the default to `${CMAKE_INSTALL_LIBDIR}/plugins`, which matches the current plugin install location. Does that seem good?

dschmidt:
Isn't `applicationDirPath()` == ` ${PREFIX}/bin`? So the lookup would be in `${PREFIX}/bin/lib/plugins`? 

If it's indeed correct I think this is ok. A different possibly simpler approach would be to use https://cmake.org/cmake/help/v3.7/module/GNUInstallDirs.html#command:gnuinstalldirs_get_absolute_install_dir

basically something around these lines:
``` 
set(PLUGINDIR "${CMAKE_INSTALL_LIBDIR}/plugins")
gnuinstalldirs_get_absolute_install_dir(FULL_PLUGINDIR PLUGINDIR)
``` 
And then add `FULL_PLUGINDIR` to the `config.h.in`. This should also handle both relative and absolute paths in and outside of the actual install prefix.

ckamm:
@dschmidt Yes, the relative path support is relative to the current location of the executable file, meaning the current default isn't useful since CMAKE_INSTALL_LIBDIR is a relative path. I'll fix by using CMAKE_INSTALL_FULL_LIBDIR for the default.

ckamm:
@dschmidt Should we change this path and the default plugin install path to LIBDIR/owncloud-client/plugins? Then the default config would install and search them in something like /usr/local/lib/x86_64-linux-gnu/owncloud-client/plugins which seems nicer than without the extra owncloud-client/ subfolder.

PR #7081, opened 2019-03-13
Linux: Add autostart delay to avoid tray issues

ckamm:
It seems that sometimes the tray implementation isn't ready on system
startup. Retrying later seems to not help. Delaying the start of the
client is the workaround that people have reported as effective.

For  #6518 since the internal retry seems to be uneffective.


ckamm:
@guruz Yeah, I also only found it when I was looking around for this ticket! :)

Agreed about "not in 2.5.4", it's not critical.

Issue #7090, opened 2019-03-15
Fix plugin packaging on OSX


ckamm:
Now that #7076 is merged it should be possible to properly include the vfs plugin into the generated bundle for OSX.

@guruz already suggested that Contents/PlugIns looks like the right place according to https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFBundles/BundleTypes/BundleTypes.html#//apple_ref/doc/uid/10000123i-CH101-SW1 .

1) I would imagine that we want a relative PLUGINDIR "../PlugIns" for that.
2) Need to make sure the plugin binaries are placed in the right location.
3) Also, is the plugin found properly when doing developer builds without bundle creation?

Since I can't test the results I don't want to be the one implementing the changes.

Relates to #7027 which is about plugin packaging for linux.

HanaGemela:
@ckamm @guruz is this something QA team can test? If so, please provide steps to reproduce

guruz:
If the virtual file system mode works in your build on macOS, then this implicitly works too :-)

PR #7094, opened 2019-03-15
macOS: Fix vfs suffix plugin paths #7090

guruz:


ckamm:
@guruz Would it be a fair summary to say:
1) Set the PLUGINDIR to be relative so the bundle works wherever it may be installed
2) Fix the developer builds by changing the build output location (has to work with same PLUGINDIR as in the final bundle, after all)
3) Fix the plugin install location so it ends up in the right place in the bundle
4) FixPlugin in macdeployqt, where I don't know what it does

If so, this seems good. It was probably a mistake then for me to use PLUGINDIR for the plugin install location.

guruz:
@ckamm correct :)
@dschmidt thanks. we'll see if the macbuilder makes something good out of it

Issue #7001, opened 2019-01-24
reproducible data loss when file is renamed and virtual file suffix is added at the same time


lazawan:
Hi there ,

operating system  : Ubuntu 18.10;  client 2.5.2rc2

i think title says a lot , simply follow the steps to reproduce the data loss : 
- enable virtual-files Feature
- delete all files in sync-folder then a message pops up ,please select keep files
- let's suppose we have 2 files with following extensions : 
something.txt.owncloud_virtual
something.pdf.owncloud_virtual
- please download the files (listed above) over owncloud-menu
- files are now  so : something.pdf  and something.txt
-now rename files to something.owncloud_virutal   (please delete the extensions .pdf and .txt )
- force sync
-files are not existing any more on the server 

regards

jnweiger:
Also reproducable with a windows client. Ouch!

lazawan:
reproducible with Macos client as well , i have just tested it

ckamm:
Yes, simpler reproduction:
- have test.txt synced
- rename to other.txt.owncloud
test.txt will be deleted and other.txt.owncloud will be ignored.

The sync engine sees this as a deletion of test.txt and the creation of an erroneous placeholder file other.txt.owncloud.

What's the expected behavior? Renaming test.txt to other.txt on the remote and dehydrating other.txt.owncloud?

I expect a similar problem to occur if one renamed test.txt.owncloud to other.txt: test.txt.owncloud would be recreated and a silly other.txt would be uploaded.

guruz:
(Issue moved per @ckamm internal comment)

ckamm:
Reasons: This is complicated to fix. A fix would take a bit and probably not be in time for 2.5.3. Also, a fix in 2.5 would need to be redone against the new discovery code in 2.6, doubling the work. I think moving to 2.6 is acceptable since the feature is experimental.

lazawan:
@ckamm unfortunately it's still reproducible with  ownCloud-2.6.0.11958.11762-daily20190510.msi
and there is another thing : build number in the link is 11762 and when one shows General settings ---> build number is 11713

![Screenshot from 2019-05-10 14-16-46](https://user-images.githubusercontent.com/44814084/57527021-3e74a600-732f-11e9-9503-076115a47bc7.png)

ckamm:
@lazawan On Windows with cfapi based vfs .owncloud files should have no special behavior at all. What exactly are you seeing?

lazawan:
[...]

PR #7103, opened 2019-03-20
VFS: Unbreak behavior for rename+hydrate

ckamm:
Users can rename a file *and* add/remove the vfs suffix at the same time
leading to very complex sync actions. This patch doesn't add support for
them, but adds tests and makes sure these cases do not cause unintened
behavior.

The rename will be propagated, but the users's hydrate/dehydrate request
will be ignored.

For  #7001. I still need to test it with winvfs, and will also drop the commented-out code once it's clear it's not needed.


ckamm:
Thanks for the review. I've tested it on windows (and added tests there), done some final touch-ups of the code and will merge now.

Issue #7101, opened 2019-03-19
Review wording with VFS


jnweiger:
Seen with 2.6.0daily 20190313 on win10:

When VFS is enabled, and a new folder appears on the server, the client still says 'downloaded.'
that is misleading, if default is online only.
![grafik](https://user-images.githubusercontent.com/1108546/54637063-1764cb00-4a88-11e9-9898-a93eb2b77962.png)

When I select 'Immer auf diesem Gerät behalten' the client pops up a notification with exactly the same wording.

Making the folder virtual again with 'Speicherplatz freigeben' the message is a vague 'updated':
![grafik](https://user-images.githubusercontent.com/1108546/54637485-08cae380-4a89-11e9-96fe-b9a75d92132c.png)

ckamm:
Indeed, I had updated the sync protocol entries in f786ef1430f85de02899dd7f4db4c0e6ae7fad27 but haven't adjusted the notifications. 

Now, what would be a good text for notifications though? Currently each action-kind in (new, deleted, updated, renamed, conflict, error) triggers a single notification that mentions the first such file by name. I wonder whether that's all that useful? But let's leave that for a different ticket.

For now I propose to adjust the wording for the "new" case to not say "downloaded". Continue to have "new-virtual" grouped with "new-regular". But put "newly-hydrated" into "updated", where "newly-dehydrated" already is.

HanaGemela:
Adding a new virtual file says 'added'
Adding a new always downloaded file says 'added'
The above was done by adding a new file via the webUI

When I free up local space or make files always available locally, the message is still updated

jnweiger:
'Immer auf diesem Gerät behalten' == 'Make alway available locally'
'Speicherplatz freigeben' == 'Free up local space'

Please do not invent new terms. I'd suggest to re-use the existing vocabulary as much as possible.
E.g. like this:

1) Server adds files, client "downloads" placeholders
        Notification "Foo.txt and 13 other files were added."
2) user changes to 'Make always available locally'
           Notification "Foo.txt and 13 other files are now always available locally."
3) user changes to 'Free up local space'
            Notification "Foo.txt and 13 others are now dehydrated."
or
                                                     Notification "Foo.txt and 13 others are now virtual."


We already have a redundancy of terms for e.g. hydrated=physical and dehydrated=virtual, 
I personally would prefer the pair (physical/virtual) over (hydrated/dehydrated).

ckamm:
[...]

PR #7107, opened 2019-03-25
Improve notification wording for hydrating vfs files

ckamm:
For #7101

### DIFF
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -416,9 +416,9 @@ void Folder::createGuiLog(const QString &filename, LogStatus status, int count,
             break;
         case LogStatusNew:
             if (count > 1) {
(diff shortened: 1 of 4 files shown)


src/gui/folder.cpp
```diff
@@ -416,9 +416,9 @@ void Folder::createGuiLog(const QString &filename, LogStatus status, int count,
             break;
         case LogStatusNew:
             if (count > 1) {
-                text = tr("%1 and %n other file(s) have been downloaded.", "", count - 1).arg(file);
+                text = tr("%1 and %n other file(s) are new.", "", count - 1).arg(file);
```
ogoffart: I think `are new` is a bit strange.
maybe `have been added` or `have been created` 
What do you think?

ckamm: Yeah, updated

Issue #7112, opened 2019-03-26
Improve error for missing data during discovery


ckamm:
"The server file discovery reply is missing data" errors are surprisingly frequent and often need our attention to diagnose.

1) Improve the error message to mention the file name and what data is missing. This will give users more hints for attempting to fix the error themselves.
2) Potentially continue syncing other items

lazawan:
@ckamm where should this Error-Message be shown and how could we reproduce this Error ?

michaelstingl:
I think you could remove the `etag` entry for file in the ownCloud server database manually. Not sure if this is a good issue for qa? /cc @HanaGemela @ckamm

ckamm:
Yes, this is not straightforward to QA. There's an autotest for the errors in testMissingData in testRemoteDiscovery.cpp and that's how I verified the new behavior works. It's a quality-of-life improvement for debugging a fairly rare error that we mostly see in enterprise contexts.

To test with a real server one would need to make it send replies that lack the size, permissions, etag or fileid. Manually adjusting the server db as suggested may work. Otherwise you'd need a proxy that filters replies.

Test cases:
1. If server doesn't send data for one of these properties, the sync errors with "server reported no X". The error should also be visible per affected file in the sync protocol.
2. Such an error does not abort the sync run: other files will be up/downloaded as normal.

michaelstingl:
> Otherwise you'd need a proxy that filters replies.

Yeah, you're right. "Faking" this in `mitmproxy` should be the recommended method. I think @jnweiger played with it before.

lazawan:
@ckamm I have just tried to remove Etag from Server response and  I didn't get an Error-Message, I got only in logs that Etag is invalid 

> 05-28 05:50:18:979 [ info gui.folder ]:	#### Propagation start ####################################################
05-28 05:50:18:994 [ info gui.application ]:	Sync state changed for folder  "https://demo.owncloud.org/remote.php/dav/files/demo/" :  "Sync Running"
05-28 05:50:18:994 [ info sync.propagator ]:	Starting INSTRUCTION_NEW propagation of "ldpaspdlapda+ldsad" by OCC::PropagateLocalMkdir(0x7a82af0)
[...]

PR #7113, opened 2019-03-26
Detect missing server data during discovery

ckamm:
This has two positive effects:
- We can put the error on the particular file that has missing data
- We can sync all other files

For  #7112

### DIFF
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -368,6 +368,27 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(
     item->_directDownloadUrl = serverEntry.directDownloadUrl;
     item->_directDownloadCookies = serverEntry.directDownloadCookies;
 
+    // Check for missing server data
+    {
+        QStringList missingData;
+        if (serverEntry.size == -1)
+            missingData.append(tr("size"));
+        if (serverEntry.remotePerm.isNull())
+            missingData.append(tr("permissions"));
+        if (serverEntry.etag.isEmpty())
+            missingData.append(tr("etag"));
+        if (serverEntry.fileId.isEmpty())
+            missingData.append(tr("file id"));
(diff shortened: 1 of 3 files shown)


src/libsync/discovery.cpp
```diff
@@ -368,6 +368,29 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(
     item->_directDownloadUrl = serverEntry.directDownloadUrl;
     item->_directDownloadCookies = serverEntry.directDownloadCookies;
 
+    // Check for missing server data
+    if (serverEntry.size == -1
```
ogoffart: I would remove this `if`, and add a condition later: `if (!missingData.isEmpty())`,  so one avoids to repeat every condition.
(But that's just cosmetic)

test/testremotediscovery.cpp
```diff
@@ -63,7 +73,6 @@ private slots:
         // 200 should be an error since propfind should return 207
         QTest::newRow("200") << 200 << httpErrorMessage;
         QTest::newRow("InvalidXML") << +InvalidXML << "error while reading directory 'B' : Unknown error";
-        QTest::newRow("MissingPermissions") << +MissingPermissions << "error while reading directory 'B' : The server file discovery reply is missing data.";
```
ogoffart: Eh, why do you remove this test?  Is there no way to keep it?
I guess, if you remove the test,  the `MissingPermissions` value should also go away.

src/libsync/discovery.cpp
ckamm: Yep

test/testremotediscovery.cpp
ckamm: I'll remove the enum. I removed it from testRemoteDiscoveryError since it's no longer in the same class as the timeout or broken-xml error. It now has its own test.

PR #7126, opened 2019-04-02
Vfs: Better handling and more tests for suffix file renames

ckamm:
Previously removing the vfs suffix of a file always triggered a
conflict. Now it may just cause a file download.

This was done because users expected symmetry in the rename actions and
renaming foo -> foo.owncloud already triggers the "make the file
virtual" action. Now foo.owncloud -> foo triggers the "download the
contents" action.


ckamm:
Possibly didn't get ported to the new discovery because that case had no tests.

Issue #6826, opened 2018-10-18
PROPFIND timeout: show folder with error, and don't abort sync run


michaelstingl:
An user added 25k PDF files to a single folder. (7GB in total) this is obviously too much for the ownCloud server and causes erratic behaviour. Needs to be fixed on the server-side, but maybe the client can also be improved a little bit.

ownCloud client 2.4.2 shows timeout error, but doesn't display which folder caused the timeout:

![pasted_image_18_10_18__21_45](https://user-images.githubusercontent.com/214010/47179740-1f74df00-d31f-11e8-8a72-ac6b9cd3d1d5.png)

![pasted_image_18_10_18__21_51](https://user-images.githubusercontent.com/214010/47180191-45e74a00-d320-11e8-9c06-ebaf6e5e0e0e.png)

```
10-18 07:43:37:234 [ info sync.networkjob.propfind ]: PROPFIND of QUrl("https://cloud.example.com/remote.php/dav/files/user123@domain.com/") FINISHED WITH STATUS QNetworkReply::NetworkError(NoError) ""
10-18 07:44:04:792 [ warning sync.networkjob ]: Network job timeout QUrl("https://cloud.example.com/remote.php/dav/files/user123@domain.com/PDF-1")
10-18 07:44:04:793 [ warning sync.networkjob ]: QNetworkReply::NetworkError(OperationCanceledError) "Connection timed out" QVariant(Invalid)
10-18 07:44:04:794 [ info sync.networkjob.lscol ]: LSCOL of QUrl("https://cloud.example.com/remote.php/dav/files/user123@domain.com/PDF-1") FINISHED WITH STATUS QNetworkReply::NetworkError(OperationCanceledError) "Connection timed out"
10-18 07:44:04:795 [ warning sync.discovery ]: LSCOL job error "Operation canceled" 0 QNetworkReply::NetworkError(OperationCanceledError)
10-18 07:44:04:796 [ warning sync.csync.updater ]: opendir failed for PDF-1 - errno 5
10-18 07:44:04:797 [ warning sync.engine ]: ERROR during csync_update : "An error occurred while opening a folder Operation canceled"
```

The new ownCloud client 2.5 produced the same error.

It would be smart to list at least the folder name in the error message. It took the user pretty long time to find out which folder of the customer was causing this problem. By unselecting this folder, the sync could continue and finished successfully.

The really annoying issue is that the ownCloud client starts checking the content over and over again - but this time at least we could see the folder name which was causing the problem.

[...]

PR #7134, opened 2019-04-09
Discovery: No fatal error on bad server response 

ckamm:
Previously receiving unexpected replies from the server during discovery
would terminate the sync run with a fatal error. Now it only causes an
error on the affected item.

For #6826

### DIFF
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -383,7 +383,6 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(
             missingData.append(tr("file id"));
         if (!missingData.isEmpty()) {
             item->_instruction = CSYNC_INSTRUCTION_ERROR;
-            item->_status = SyncFileItem::NormalError;
             _childIgnored = true;
             item->_errorString = tr("server reported no %1").arg(missingData.join(QLatin1String(", ")));
             emit _discoveryData->itemDiscovered(item);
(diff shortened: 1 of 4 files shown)


ogoffart:
The test was making sure that the filename where the problem occurs is in the error message. This should stay.

ckamm:
Since the error is now located on the affected item it's no longer necessary to include the file name in the message itself. We do `appendErrorString(QObject::tr("%1: %2").arg(item->_file, item->_errorString));`

ogoffart:
This sounds like a dangerous change.

It is true that this is now made possible to ignore part of the tree with the new discovery.

What happens if the server suddenly goes in maintenance or away? Do we get many ignore "error".

If a file is moved to a directory which now errors, the file will simply be deleted. Is that ok?

ckamm:
Yes, doing discovery with partial remote information would affect things like renames/moves in the way you mention. 

And thinking about it more, I agree that it's a bad idea to open the sync algorithm for this kind of complication. If we look a the user issue "my remote discovery takes very very long and if it has any kind of error it has to start from scratch" it's plausible that we can solve this in another way (like the discovery cache we talked about) without starting to sync when we know we don't have an up to date remote snapshot.

Note: The maintenance mode (503) behavior already works in this way: it was already handled by ignoring the folders where the error appeared.

ckamm:
I'll extract some changes from this patch and open a new PR with these.

PR #7135, opened 2019-04-09
Add Makefiles for handling transifex translations

dschmidt:


admin/win/nsi/l10n/Makefile
```diff
@@ -0,0 +1,24 @@
+#
+# Translation
+#--------------------------------------
+
+.PHONY: l10n-push
+l10n-push: ;
+	tx -d push -s --skip --no-interactive
+
+.PHONY: l10n-pull
+l10n-pull: ;
```
tboerger: ; should be dropped?

admin/win/nsi/l10n/Makefile
```diff
@@ -0,0 +1,24 @@
+#
+# Translation
+#--------------------------------------
+
+.PHONY: l10n-push
+l10n-push: ;
```
tboerger: Same here?

admin/win/nsi/l10n/Makefile
dschmidt: No, this is how you specify an empty recipe: https://www.gnu.org/software/make/manual/html_node/Empty-Recipes.html

admin/win/nsi/l10n/Makefile
dschmidt: Same answer as above ;-)

admin/win/nsi/l10n/Makefile
dschmidt: Ah, yes. That was for debugging. Good catch!

dschmidt:
Wtf?! I did not close this and I cannot reopen it now o.O

PR #7152, opened 2019-04-17
Link shares: Adjust creation flow

ckamm:
The important thing is that, similar to the web link share creation
flow, the share properties can be set before creation is done. This
allows better handling of error conditions. One such condition is the
password policy app restricting valid expire dates.

See owncloud/enterprise#3271

This significantly adjusts how the link share creation flows, will post images below.


ckamm:
Can now set properties before clicking "create":
image

Share editing behavior is unchanged:
image

Issue #7154, opened 2019-04-17
Permanently save logs is not permanent


jnweiger:
![grafik](https://user-images.githubusercontent.com/1108546/56292869-48d4c300-6128-11e9-926a-ad49e4e62bf0.png)

The mouse over hover text contradicts the onscreen text.
If it expires after a few hours, then it cannot be permanent.
Please get that in sync.

guruz:
That's the old log window btw.
@ckamm Maybe the time the logs are stored should also be increased?

michaelstingl:
> @ckamm Maybe the time the logs are stored should also be increased?

Checkbox for auto-removal of log entries?

lazawan:
@ckamm unfortunately after this modification i don't get any Logs at all ( i get only a window with Info about Logs ) . another thing : when i click on open Folder button ---> nothing happens because no folder was created yet.
by the way i tested this new Feature with ownCloud-2.6.0.11958.11762-daily20190510.msi

![4444](https://user-images.githubusercontent.com/44814084/57618449-ba672c00-7583-11e9-9cb4-bea1a5f232ee.png)

ckamm:
@lazawan :( I'll need to revisit.

ckamm:
@lazawan I've tried again with my 2.6 build and logging setup works for me. Maybe you used command-line logging arguments when you tried?

lazawan:
@ckamm which build did you use ? maybe i used the wrong one . and did you get live Logs ? i mean logs without saving

ckamm:
@lazawan The issue I could not reproduce with my 2.6 build was this:

> unfortunately after this modification i don't get any Logs at all ( i get only a window with Info about Logs ) . another thing : when i click on open Folder button ---> nothing happens because no folder was created yet.

The live log view is removed intentionally: having it lead to people copying data from there into bug reports. These log excerpts were nearly always too short to be useful. The hope is that the on-disk log files will enable people to report useful data more easily while also enabling people with intermittent issues to capture logs preemptively.

What's the situation in which users would want a live log view?

I'm wary around spending too much extra time on settings and ui of the log window. It's an ancillary place that most users won't interact with. Let's only add features we're sure will be needed.

ckamm:
Let's iterate on allowing users to easily send us logs - but this particular bug looks fixed to me.

HanaGemela:
Text is now 'Enable logging to temporary folder'
[...]

PR #7157, opened 2019-04-23
LogWindow: Option to disable automatic deletion

ckamm:
For  #7154

### DIFF
--- a/src/gui/logbrowser.cpp
+++ b/src/gui/logbrowser.cpp
@@ -36,6 +36,8 @@ namespace OCC {
 
 // ==============================================================================
 
+const std::chrono::hours defaultExpireDuration(4);
+
 LogBrowser::LogBrowser(QWidget *parent)
     : QDialog(parent)
 {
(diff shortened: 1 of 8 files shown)


src/gui/logbrowser.cpp
```diff
@@ -36,6 +36,8 @@ namespace OCC {
 
 // ==============================================================================
 
+const int defaultExpireHours = 4;
```
ogoffart: why not std::chrono::hours ?

src/gui/logbrowser.cpp
```diff
@@ -104,24 +112,46 @@ LogBrowser::~LogBrowser()
...
+    auto logger = Logger::instance();
+
+    if (config.automaticLogDir()) {
+        // Don't override other configured logging
+        if (logger->isLoggingToFile())
+            return;
+
+        logger->setupTemporaryFolderLogDir();
+        if (auto deleteOldLogsHours = config.automaticDeleteOldLogsAge()) {
+            logger->setLogExpire(deleteOldLogsHours->count());
```
ogoffart: Too bad that setLogExpire also is not taking a std::chrono arg.  
now, std::chrono does not help because we do not see that it is in hours.

ckamm:
@ogoffart I changed some of the touched apis to pass std::chrono types now.

PR #7186, opened 2019-05-14
Migrate v2.5 of the documentation to Antora / AsciiDoc

settermjd:
This fixes https://github.com/owncloud/docs/issues/1197.


tboerger:
That will kill the whole 2.5 packaging since this relies on sphinx. /cc @jnweiger @dschmidt

PR #7188, opened 2019-05-14
OAuth2: Better error logging

guruz:
This does not fix a bug, just was found while spotting a bug that was no bug.
For https://github.com/owncloud/enterprise/issues/2951

### DIFF
--- a/test/testoauth.cpp
+++ b/test/testoauth.cpp
@@ -33,6 +33,8 @@ class FakePostReply : public QNetworkReply
 public:
     std::unique_ptr<QIODevice> payload;
     bool aborted = false;
+    bool redirectToPolicy = false;
+    bool redirectToToken = false;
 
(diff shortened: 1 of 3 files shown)


test/testoauth.cpp
```diff
@@ -52,6 +54,24 @@ class FakePostReply : public QNetworkReply
             emit metaDataChanged();
             emit finished();
             return;
+        } else if (redirectToPolicy) {
+            setHeader(QNetworkRequest::LocationHeader, "/my.policy");
+            setAttribute(QNetworkRequest::RedirectionTargetAttribute, "/my.policy");
+            setAttribute(QNetworkRequest::HttpStatusCodeAttribute, 307);
```
patrickjahns: would a `302` make a difference here? As far as I understood per default bigip will use 302

guruz: it doesn't make a difference but i updated the PR anyway:)

Issue #7204, opened 2019-05-23
After renaming the original file is synchronised as well


HanaGemela:
Server: 10.2.0
Client: 2.6.0-daily20190522 (build 11801)
macOS: 10.14.4.

Steps to recreate:
1) Make sure you have only one single file in your ownCloud
2) Rename the file

Expected result: File has been renamed

Actual result: Pop-up message "All files in the sync folder 'ownCloud' folder were deleted on the server....". When user clicks 'Keep files', the original file gets synchronised as well. That means now user has two files - original one and renamed one

ogoffart:
PR #7215

HanaGemela:
Works as expected in 2.6.0alpha2 (build 12128), macOS 10.14.5

PR #7215, opened 2019-06-04
SyncEngine: Fix renaming a single file cause the "delete all file" popup

ogoffart:
Possibly a regression, since the new discovery discovers rist the renamed
files as removed

Issue #7204

### DIFF
--- a/test/testallfilesdeleted.cpp
+++ b/test/testallfilesdeleted.cpp
@@ -271,6 +271,30 @@ private slots:
 
         QCOMPARE(fakeFolder.currentLocalState(), fakeFolder.currentRemoteState());
     }
+
+    void testSingleFileRenamed() {
+        FakeFolder fakeFolder{FileInfo{}};
+
+        int aboutToRemoveAllFilesCalled = 0;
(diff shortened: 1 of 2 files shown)


ogoffart:
I'm setting the milestone to 2.6 since it is probably a regression

test/testallfilesdeleted.cpp
```diff
@@ -271,6 +271,31 @@ private slots:
...
+        // add a single file
+        fakeFolder.localModifier().insert("hello.txt");
+        QVERIFY(fakeFolder.syncOnce());
+        QCOMPARE(aboutToRemoveAllFilesCalled, 0);
+        QCOMPARE(fakeFolder.currentLocalState(), fakeFolder.currentRemoteState());
+
+        // rename it
+        fakeFolder.localModifier().rename("hello.txt", "goodbye.txt");
+
+        // Now, aboutToRemoveAllFiles with down as a direction
```
ckamm: You mean that it's *not* triggered, right?

ogoffart: Bad copy-paste

Issue #7199, opened 2019-05-22
One broken folder stops the entire sync


HanaGemela:
Server: 10.2.0
Client: 2.6.0-daily20190513 (build 11776)
macOS: 10.14.4

Steps to recreate:
1) Add a new account that has one broken folder on the server

Expected result: All items apart from the broken folder have been synced

Actual result: Nothing has been synced - see log below

Note: Currently working on how to recreate that broken folder on server. Once known, the issues will be linked together

![image](https://user-images.githubusercontent.com/49001702/58174968-62c17280-7c9f-11e9-863a-8b3e5fd596e0.png)

![image](https://user-images.githubusercontent.com/49001702/58174989-6c4ada80-7c9f-11e9-8e8a-1eabe7a6f1ac.png)

michaelstingl:
Looks like a candidate for my collection in https://github.com/owncloud/client/issues/5859

HanaGemela:
Follow steps 1-5 in https://github.com/owncloud/core/issues/35305 to create that problematic folder

jnweiger:
It should also be possible to poison one folder using mitmproxy with a python script that fakes a 500 response, whenever a specific folder is accessed.

ogoffart:
What was the server error code.  Looking at the error message it seems to be a 404. But 404 should not cause the sync to stop, neither should 500.
The only errors causing the sync to abort is a 503 if the server is in maintainnence, or a connection drop (connection timeout or similar)


Could you paste the log from the client?

HanaGemela:
@ogoffart I've just emailed you the logs

ogoffart:
Thanks for the log. I realized the problem was during the discovery and not during the propagation.

Fix PR: https://github.com/owncloud/client/pull/7219

HanaGemela:
The folder is not synced but it is still shown in the selective sync list as synced 

![image](https://user-images.githubusercontent.com/49001702/61793425-68fad900-ae1f-11e9-9657-5d2948795939.png)

HanaGemela:
@michaelstingl please decide if it is worth the effort to fix the selective sync list

michaelstingl:
The remaining issue is only cosmetic, so I'd close it for now.

PR #7219, opened 2019-06-04
Discovery: Do not abort the sync in case of error 404 (or 500)

ogoffart:
Issue: #7199


ogoffart:
This is not a fix for a regression. 2.5 had the same problem and worse, as the previous discovery had more of trouble when ignoring part of the tree.

However, this means that if files were moved from or to the directories that are now unavailable because of an error, they would not be recognized as moves, but instead deleted or duplicated.

Issue #6877, opened 2018-11-13
[Windows?] On Startup: Assert in SyncJournalDb::checkConnect via getSelectiveSyncList(?)


guruz:
https://sentry.io/owncloud/desktop-win-and-mac/issues/759409929/

ogoffart:
```
  File "syncjournaldb.cpp", line 324, in OCC::SyncJournalDb::checkConnect()
```

Actually, this is the DROP INDEX that was added in 2.5.1, so maybe this is a regression

ogoffart:
But we'd still need a log to find out what the problem is.

guruz:
Woops, wrong milestone, adding to 2.5.x(=2.5.2)

ckamm:
Yeah, why would prepare of `DROP INDEX IF EXISTS metadata_parent;` fail? Hmm.

ckamm:
@ogoffart Can you explain why that statement must execute before synchronous mode is enabled? I'd prefer if it was done after all the setup pragmas but you seem to have put it into this location for a reason.

ogoffart:
> Yeah, why would prepare of DROP INDEX IF EXISTS metadata_parent; fail?

Maybe the database is corrupted and it is the first statement that runs so that's where it fails.
But if it were not there it would fail later.  (this is just a hypotheses, i don't know if that's true)

> Can you explain why that statement must execute before synchronous mode is enabled?

I don't recall exactly, but it it tried and it does not work. Maybe I missed something.

guruz:
Related https://github.com/owncloud/client/pull/6999

ckamm:
Seems to still exist in 2.5.3: https://sentry.io/owncloud/desktop-win-and-mac/issues/890266316/?query=release%3A%222.5.3%20(build%2011250)%22%20crash

guruz:
@ogoffart What to do?

ckamm:
I think what's happening in https://sentry.io/organizations/owncloud/issues/862054725 and https://sentry.io/organizations/owncloud/issues/866394733 is this:
- db is closed on sync-finish and needs to be reopenend
- pragma journal_mode=wal fails (probably with BUSY or LOCKED, maybe antivirus?)
- we run into the ENFORCE and log and crash

To fix: don't close-and-reopen the db all the time. This is already done in 2.6 and could possibly be backported to 2.5.5.

ckamm:
[...]

PR #7224, opened 2019-06-05
Backport change to stop closing the db connection when a sync finishes

ckamm:
it seems likely to decrease the severety of #6877: if the database isn't closed and reopened as frequently users are less likely to run into crashes when the sqlite connection switches to wal journaling mode.


ckamm:
@dschmidt Apparently appveyor failed at "Action: compile for libs/qt5/qttranslations:5.11.1 FAILED" for both the 32 bit and 64 bit builds. Assuming it's unrelated.

Issue #6880, opened 2018-11-14
[2.5.1] Unhandled C++ Exception in PropagateUploadFileNG::startNextChunk()


guruz:
https://sentry.io/owncloud/desktop-win-and-mac/issues/655442309/events/36380829151/
```
﻿﻿Crash: Unhandled C++ Exception at 0x773017d2
  File "propagateuploadng.cpp", line 319, in OCC::PropagateUploadFileNG::startNextChunk()
  File "propagateuploadng.cpp", line 442, in OCC::PropagateUploadFileNG::slotPutFinished()
```

We were discussing a similar backtrace here: https://github.com/owncloud/client/issues/6699#issuecomment-420513324

guruz:
Related https://github.com/owncloud/enterprise/issues/2985

ogoffart:
Since it is the location of one of the bigger allocation, it is likely to be triggered in a out of memory situation.
We were hoping that the leaks were fixed in 2.5.1. Maybe there's still leaks. Or maybe it's just the peak memory which is too big.

ckamm:
The sentry issue shows that this is still regularly reported with 2.5.1 on Windows. However, I don't think we can do much without further information. I think both @ogoffart and me have already done a bunch of analysis around this and I would imagine further undirected work to not be worth the cost. I'd be different if there was a hint, through user reports of heap tracking tools.

One thing we might try would be including memory tracking code in the released client, so users can more easily provide us with useful data on memory issues.

ckamm:
Ideas:
- We could catch all exceptions and log them
- We could add extra logging like #6855 

Do you know of a way to make the stackframe for `prepareAndOpen()` show up? Stacktraces always point to propagateuploadng:319 and then into QtCore.

ckamm:
Interestingly there seems to be no case of this issue reported for >=2.5.2

ckamm:
On further investigation this is likely to be the 2.5.2, 2.5.3 version of this crash: https://sentry.io/organizations/owncloud/issues/866377501 . It points towards not enough memory being available for allocating the uploaddevice.

ckamm:
How about we try-catch around that particular allocation - which is way more likely than others to fail - and report it as a sync error?

HanaGemela:
@ckamm @guruz is this something I can test? If so, please provide steps to recreate. Otherwise please close

PR #7225, opened 2019-06-05
Upload: Catch out-of-memory error

ckamm:
This will avoid a rare-but-regular crash in low-memory situations.

Note that this is a workaround: In 2.6 this code should be adjusted to
not read a full file chunk into memory immediately and instead load more
data on demand.

For  #6880


guruz:
But how does this provide a solution?
Could it not set some things to continue syncing?
Will this abort the sync or only avoid the single file to be uploaded in this sync run?

ckamm:
It creates a soft error and continues syncing, very likely creating more errors for other large files. It'd at least tell users what's going on. They could then reduce parallelism or maxChunkSize. 

The correct solution is to avoid having such a huge buffer in the first place (#7226) which we may also do in 2.5 if the change is low-risk enough.

Issue #7174, opened 2019-05-03
Resend network requests on ContentReSendError


ckamm:
Now that https://codereview.qt-project.org/#/c/254111/ is merged and will apear in Qt 5.12.4 we can prepare the client by catching the error and doing a manual resend. This will allow us to re-enable HTTP2 support once that Qt is available.

ogoffart:
(moved to 2.6.1 since Qt 5.12.4 is not yet released)

ckamm:
This is extremely hard to test, to the point where I would skip it. One needs to run with 5.12.4 and have a server that's enabled for http2 and sends GOAWAY frames in the way that triggers the Qt bug that used to exist.

HanaGemela:
@ckamm How come all the platforms are using different version of Qt? 
macOS uses 5.12.1
Window uses Qt 5.12.4
Ubuntu uses Qt 5.12.2

Does this need testing on lower version than 5.12.4?

ckamm:
@HanaGemela I don't know the exact reason. (@dschmidt ?)

This enhancement can only be tested with Qt >=5.12.4, can't test it with an earlier version.

HanaGemela:
@ckamm does that mean that the client with Qt 5.12.4 has HTTP2 enabled by default? 
What if I'll connect it to the server that does not support it?
Do you know how to send GOAWAY frames in the way that triggers the Qt bug?

ckamm:
@HanaGemela Yes, a 2.6.0 client with Qt >= 5.12.4 will check whether a server supports http2 and use it if possible. If the server doesn't support it will work as before.

GOAWAY: I think this nginx config triggered the issue, but only sometimes:
```
http2_max_concurrent_streams 2;
http2_max_requests 2;
http2_idle_timeout 3m;
```

HanaGemela:
Closing as exact steps to recreate are not known.

guruz:
> @ckamm How come all the platforms are using different version of Qt?

The Qt version is defined by the build systems not by the client source code :-)
We try to get this all to 5.12.4 though

guruz:
@ckamm Please evaluate https://codereview.qt-project.org/c/qt/qtbase/+/272754 when you have the time..

PR #7227, opened 2019-06-06
Http2: Resend requests on ContentReSend error 

ckamm:
Since Qt does not yet transparently resend HTTP2 requests in some cases
we do it manually.

The test showed a problem where the initial non-200 reply would close
the target temporary file and the follow-up request couldn't store any
data. Removing that close() call is safe because there also is a
_saveBodyToFile flag that guards writes to the target file.

For #7174

### DIFF
--- a/src/libsync/abstractnetworkjob.cpp
+++ b/src/libsync/abstractnetworkjob.cpp
@@ -53,7 +53,6 @@ AbstractNetworkJob::AbstractNetworkJob(AccountPtr account, const QString &path, 
     , _ignoreCredentialFailure(false)
     , _reply(0)
     , _path(path)
-    , _redirectCount(0)
 {
     // Since we hold a QSharedPointer to the account, this makes no sense. (issue #6893)
     ASSERT(account != parent);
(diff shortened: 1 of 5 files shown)


src/libsync/abstractnetworkjob.cpp
```diff
@@ -164,6 +164,27 @@ void AbstractNetworkJob::slotFinished()
...
+        }
+        sendRequest(
+            verb,
+            _reply->request().url(),
+            _reply->request(),
+            _requestBody);
+        return;
+    }
+
     if (_reply->error() != QNetworkReply::NoError) {
```
ogoffart: The previous chunk of code could be moved in this block, maybe

src/libsync/abstractnetworkjob.cpp
```diff
@@ -164,6 +164,27 @@ void AbstractNetworkJob::slotFinished()
         qCWarning(lcNetworkJob) << "SslHandshakeFailedError: " << errorString() << " : can be caused by a webserver wanting SSL client certificates";
     }
 
+    // Qt doesn't yet transparently resend HTTP2 requests, do so here
+    QByteArray verb = requestVerb(*reply());
+    if (_reply->error() == QNetworkReply::ContentReSendError
+        && _reply->attribute(QNetworkRequest::HTTP2WasUsedAttribute).toBool()
+        && (!_requestBody || _requestBody->isSequential())
+        && !verb.isEmpty()) {
```
ogoffart: Should we also not increment and check _redirectCount (or another variable) in case there is a problem with Qt or the server that causes QNetworkReply::ContentReSendError to be send over and over again.

src/libsync/propagatedownload.cpp
```diff
@@ -176,7 +172,6 @@ void GETFileJob::slotMetaDataChanged()
         // Disable the buffer limit, as we don't limit the bandwidth for error messages.
         // (We are only going to do a readAll() at the end.)
         reply()->setReadBufferSize(0);
-        _device->close();
```
ogoffart: Is that safe? 

As the comment say, we do not want to write to the file because we need to keep the data for error reporting.

src/libsync/abstractnetworkjob.cpp
ckamm: I don't want to do it because it'd increase nesting unnecessarily.

src/libsync/abstractnetworkjob.cpp
ckamm: Yes, definitely!

src/libsync/propagatedownload.cpp
ckamm: See commit message: I think it's safe because there's an extra `_saveBodyToFile` flag that's only set to true if `slotMetaDataChanged` doesn't detect any error.

ckamm:
Failures look unrelated, merging.

Issue #7226, opened 2019-06-05
Read file chunk data on demand to reduce peak memory use


ckamm:
This is a continuation of #6880.

Some users, usually on windows, experienced crashes during parallel uploads of large files. This was because a chunk of each file was loaded into memory and chunk size goes up to 100 MB. A workaround was added to avoid the crash and instead error on the upload.

The better solution is to keep only a much smaller buffer of file data in memory and read from disk as needed. This would significantly reduce the peak amount of memory used by the client during upload.

ckamm:
QA: Add many large new local files and trigger an upload. Observe memory use. Increase during sync should be < 100MB.

HanaGemela:
Works as expected in 2.6.0alpha2 (build 12128) on macOS and Windows

PR #7229, opened 2019-06-06
Upload: Read file chunks gradually 

ckamm:
Instead of all at once, to reduce peak memory use.

Changing UploadDevice in this way requires keeping the file open for the
duration of the upload. It also means changes to open(), seek(), close()
to ensure that uses of the device work right when a request needs to
be resent.

Let's consider whether this is safe for 2.5 or should be a 2.6 change!

For #7226

### DIFF
--- a/src/libsync/propagateupload.cpp
+++ b/src/libsync/propagateupload.cpp
@@ -316,8 +316,11 @@ void PropagateUploadFileCommon::slotStartUpload(const QByteArray &transmissionCh
     doStartUpload();
 }
 
-UploadDevice::UploadDevice(BandwidthManager *bwm)
-    : _read(0)
+UploadDevice::UploadDevice(const QString &fileName, qint64 start, qint64 size, BandwidthManager *bwm)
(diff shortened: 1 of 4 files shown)


src/libsync/propagateupload.cpp
```diff
@@ -381,7 +383,12 @@ qint64 UploadDevice::readData(char *data, qint64 maxlen)
         }
         _bandwidthQuota -= maxlen;
     }
-    std::memcpy(data, _data.data() + _read, maxlen);
+
+    auto c = _file.read(data, maxlen);
+    if (c != maxlen) {
```
ogoffart: should we not just do `_read += c; return c;` ?

ckamm: @ogoffart If `c >= 0` yeah. I mimiced the previous code here but that should be unnecessary.

ckamm:
Moved to 2.6 after everyone was "mh, maybe?" on inclusion in 2.5.

Issue #7241, opened 2019-06-13
Detect crippled folder watchers to avoid bad sync behavior


ckamm:
From https://github.com/owncloud/enterprise/issues/3303

Some filesystems may have limitations in file and folder watchers, like some Windows network drives. This leads to two main problems:
- Changes to files in the sync folder don't trigger a sync run. Users must trigger it manually or wait for a time-based one.
- To avoid slow local discovery the client normally trusts the folder watcher about changes. When notifications are missing the client may perform a sync run and not notice local changes.

Workarounds: Use the `forceSyncInterval` and `fullLocalDiscoveryInterval` settings to reduce the duration between time-based sync runs and to require full local discovery every time.

This issue is about automatically detecting bad folder watchers. The client should then notify the user and switch into the existing "unreliable folder watcher" mode.

ckamm:
QA: I tested by verifying no warning comes up after startup on Windows and linux. Then I commented out these lines in folderwatcher.cpp
```
    if (QFile::exists(path)) {
        auto mtime = FileSystem::getModTime(path);
        FileSystem::setModTime(path, mtime + 1);
    } else {
        QFile f(path);
        f.open(QIODevice::WriteOnly | QIODevice::Append);
    }
```
and checked that the warning notification was triggered.

To test for real one would need to set up a network drive like in the enterprise ticket.

HanaGemela:
@ckamm Does it mean that any Windows network drive will not sync the changes until the client does the full local discovery which is once an hour?

ckamm:
@HanaGemela No. It should only come up for network drives where folder watchers are unsupported. I don't know which those are. Unfortunately I didn't hear back in owncloud/enterprise#3303 about what kind of network drive was affected specifically and couldn't reproduce their problem. That's why I could only test that the check works in principle by changing code.

HanaGemela:
OK, this issue will be closed as soon as https://github.com/owncloud/client/issues/7305 has passed QA

HanaGemela:
#7305 has passed QA

PR #7242, opened 2019-06-13
FolderWatcher: Become unreliable if test notification fails

ckamm:
Necessary for some filesystems on windows that don't have full file
watching capabilities.

 #7241 https://github.com/owncloud/enterprise/issues/3303


ckamm:
There's a genuine test failure I need to address.

guruz:
Note that since Windows supports something like mounts, this could still fail somewhere deeper in the directory

michaelstingl:
@cdamken You could test with next weeks 2.6.0 nightlies

Issue #7230, opened 2019-06-06
Clarify "Apply manual changes" when asking user about large folders


HanaGemela:
Client: 2.6.0-daily20190522 (build 11801)
Server: 10.2.0
OS: Mojave 10.14.5

Steps to recreate:
1. After completing an initial sync, log out the client
2. Upload an additional folder to the server that exceeds the limit beyond which the client asks for confirmation if folders larger than that should be synced (Advanced Setting in General tab) 
3. Log in the client to the server
4. Select "Apply manual changes" in the warning message

Expected result: Not really sure. Maybe the client should offer a selection which files inside that folder should be synced?
Actual result: The folder is not synced. Basically identical behaviour as 'Synchronize none' button. If the behaviour is really the same, then there is no need for 'Apply manual changes' button

ckamm:
Indeed, if you make no changes to the selective sync list "Apply manual changes" will be the same as "Synchronize none". Possibly we could disable "Apply manual changes" until the user has actually made a change.

HanaGemela:
@ckamm What should I do to see the difference between these two buttons?

ckamm:
@HanaGemela When the "new large folder" ui state is activated you should be seeing some sync folders subfolders pop up. Each has a checkbox for its current selective sync state. The new large folders have are unchecked by default. You could have two new large folders, check only one and use "Apply manual changes" to confirm your manual choice (instead of syncing both or neither).

HanaGemela:
The same when syncing External storages

guruz:
@HanaGemela Please test

ckamm:
Behavior change is that now the "Apply manual changes" button only becomes enabled when a change was made.

HanaGemela:
@ckamm Does it mean I'll see the button when there are two large folder to be synced? 
Are the below steps to recreate correct?

1. After completing an initial sync, select only some folders to be synced, log out the client
2. Upload an additional folder to the server that exceeds the limit 
3. Log in the client to the server
Expected result: the button is shown

ckamm:
@HanaGemela When you do these steps the button will be shown (as before), but it will be disabled (new). It will become enabled when you change the checked status of at least one subfolder checkbox.

HanaGemela:
@ckamm I've checked with macOS 2.6.0-daily20190616 (build 11930), I run exactly the steps described above and the button is enabled. 
And what subfolder do you mean?

ckamm:
[...]

PR #7244, opened 2019-06-14
BigFolders: Enable "apply manual changes" only if dirty

ckamm:
Currently it's confusing because without user changes "Sync None" and
"Apply manual changes" have the same meaning. Now the "Apply" action
only becomes available if changes were made.

For #7230

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -985,6 +985,7 @@ void AccountSettings::refreshSelectiveSyncStatus()
 
         ui->selectiveSyncButtons->setVisible(true);
         ui->bigFolderUi->setVisible(false);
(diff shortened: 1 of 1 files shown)


src/gui/accountsettings.cpp
```diff
@@ -985,6 +985,7 @@ void AccountSettings::refreshSelectiveSyncStatus()
 
         ui->selectiveSyncButtons->setVisible(true);
         ui->bigFolderUi->setVisible(false);
+        ui->selectiveSyncApply->setEnabled(_model->isDirty());
```
ckamm: Since selectiveSyncApply is a child of selectiveSyncButtons it's unnecessary to setEnabled() on it in the bigFolder case.

Issue #7245, opened 2019-06-15
Public sharing does not work anymore with 10.2.0 when password is enforced


lexo-mfleuti:
### Expected behaviour
When the option "Enforce expiration date" is enabled the Windows client should automatically set the latest expiration date and create a valid public share link. This worked up until 10.2.0.

### Actual behaviour

- Creating a public share link directly via context menu: does not work at all
- Creating a public share link via "share" dialog: does not work at all. The "expiration date" field is greyed out and cannot be activated - not even manually. Instead a red "expiration date is enforced" error message is shown from the beginning.

### Steps to reproduce

1. Enable this setting in OC Settings: Sharing > Enforce expiration date
2. Right click on a file -> Select "ownCloud" within context menu
3. Select "Share..." (Teilen...)
4. Enjoy the show

### Server configuration
Operating system: 
Ubuntu 19.04 (GNU/Linux 5.0.0-16-generic x86_64)

Web server: 
Apache/2.4.38 (Ubuntu)

Database: mysql  Ver 15.1 Distrib 10.3.13-MariaDB, for debian-linux-gnu (x86_64) using readline 5.2

PHP version: 
PHP 7.2.19-0ubuntu0.19.04.1 (cli) (built: Jun  4 2019 14:44:42) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.2.0, Copyright (c) 1998-2018 Zend Technologies
    with Zend OPcache v7.2.19-0ubuntu0.19.04.1, Copyright (c) 1999-2018, by Zend Technologies

ownCloud version:
10.2.0

Storage backend (external storage):
Local

### Client configuration
Client version:
I tried 2.5.4 and 2.6.0alpha1 from here: https://owncloud.org/download/

OS language:
English

Installation path of client:
C:\Program Files (x86)\ownCloud
![oc_error](https://user-images.githubusercontent.com/8611779/59550368-34ffde80-8f6a-11e9-8bf0-16ae2ae3088d.png)

ckamm:
@lexo-mfleuti So you're saying with server version 10.2.0 it starts failing?

lexo-mfleuti:
@ckamm Yes, we updated to 10.2.0 about 2-3 weeks ago and it's no longer working ever since. 
There was an issue with this in client version 2.5.3 already which I reported and was solved shortly after in 2.5.4. But with 10.2.0 it's now failing completely. I cannot even set a date manually anymore.

[...]

PR #7249, opened 2019-06-17
Sharing: Fix fast share creation with enforced expiry and 10.2.0

ckamm:
Fixes the "Copy public link to clipboard" context menu action for newer
servers.

Previously the server would fill the expiration date for us, since
version 10.2.0 it reports an error. To workaround that the action now
always sends an expiration date if default expiration is enabled in
the settings.

For #7245


ckamm:
@ogoffart Yeah: https://github.com/owncloud/core/pull/35564

Issue #7250, opened 2019-06-17
Copy virtual files - Virtual files get uploaded on server, client keeps both


HanaGemela:
Client: macOS 2.6.0-daily20190613 (build 11917)
Server: 10.2.0
macOS: 10.14.5

Steps to recreate:
1) Turn on VFS
2) Go to Photos folder
3) Copy Paris.jpg.owncloud
4) Paste it into the same location

Actual result: Virtual file is uploaded on the server
Expected result: Paris copy.jpg has been created and it's virtual on the client and normal on the server 

5) Copy Squirrel.jpg (note it is hydrated)
6) Paste it into the same location

Actual = Expected result: Squirrel copy.jpg.owncloud has been created on the client. Squirrel copy.jpg is on the server

7) Copy Squirrel.jpg (note it is hydrated)
8) Paste it into the same location

Actual result: Both Squirrel copy.jpg.owncloud and Squirrel copy.jpg are on the client
Expected result: Another copy of the file is Squirrel copy 2.jpg.owncloud. And it is virtual on the client and normal on the server

![image](https://user-images.githubusercontent.com/49001702/59599075-f864ec00-90fd-11e9-9419-afc5476e79db.png)

ckamm:
@HanaGemela What's the difference between step 5/6 and step 7/8?

HanaGemela:
> @HanaGemela What's the difference between step 5/6 and step 7/8?

No difference. Just repeat the same process again. 
The issue is that in steps 5/6 it creates a virtual file and then the same file is created in steps 7/8 but this time as normal file. I would expect that it increases the number in the filename (the first file is copy, the second file is copy 2 etc.)

ckamm:
I can't reproduce what happens after step 4). For me (on linux), the new virtual file is deleted. Virtual files being uploaded would be a significant bug and if you can reproduce it, please share logs!

I also see something odd happening in 5-8 and will look at it.

HanaGemela:
@ckamm do you want the logs from my mac or do you want me to try on linux?

ckamm:
both :)

HanaGemela:
I can't recreate now. The difference is that yesterday when I copy pasted a file, I got virtual file immediately. Today when I copy a normal file, the copy is normal as well. It takes a while until it's changed to virtual. How can I make it to create a virtual file straightaway?

ckamm:
[...]

PR #7263, opened 2019-06-21
Set pin state of new and renamed virtual files

ckamm:
That means new and renamed files and folders in online-only directories will no longer become dehydrated.

For #7250


ckamm:
Still need to test on windows

PR #7274, opened 2019-06-25
Fix persisting pin states after file deletion

ckamm:
The adjustments to the `metadata` table are rather complex, since one can't add a PRIMARY KEY after the table has been created.


ckamm:
Moving to 2.7 due to safety concerns.

TheOneRing:
We will migrate users away from the suffix placeholders in an upcoming release so there is no need for this any more.

Issue #7275, opened 2019-06-27
Cannot reshare a folder with restricted permissions via User and Groups share


HanaGemela:
Server: 10.2.1 RC1
Client: macOS testpilot 2.6.0-daily20190626 (build 11996)
macOS: 10.14.5

Steps to recreate:
1) Share a folder with one subfolder (e.g. Test/Subfolder) with User1 and remove Edit permissions
2) Login as User1
3) Reshare Subfolder with User2 via User and Groups share

Actual result: Entry in Users and Groups sharing list cannot be created
Expected result: Entry in Users and Groups sharing list with the same permissions as the recipient has 

![image](https://user-images.githubusercontent.com/49001702/60577704-27c35c00-9d80-11e9-91ff-44da00d5e98f.png)

michaelstingl:
@micbar looks like the thing we found. Maybe open a server issue too…

micbar:
## Background
- Desktop client tries to reshare a subfolder. 
- The Share permissions on the incoming parent folder are for example `17`
- The desktop client (2.5.x) tries to create the share with default permissions from the Capabilities API, which are for example `31`
- The creation of the share fails

## Solution
- Do a PROPFIND on the Subfolder you want to share first and request the sharing-permissions
- Respect the current permissions when creating a share

```
<?xml version="1.0" ?>
<d:propfind xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns" xmlns:x1="http://open-collaboration-services.org/ns">
  <d:prop>
    <d:getetag/>
    <oc:fileid/>
    <oc:permissions/>
    <oc:share-types/>
    <oc:owner-id/>
    <d:getlastmodified/>
    <oc:size/>
    <oc:checksums/>
    <x1:share-permissions/>
  </d:prop>
</d:propfind>```

jnweiger:
Reproduced with client 2.6.0daily on Linux:

User Admin shares AdminPhotos with user jwtest, allowing reshare, but not edit.
User jwtest tries to reshare AdminPhotos/subfolder with user testy, he gets this error message:

![image](https://user-images.githubusercontent.com/1108546/60578372-5db51000-9d81-11e9-8c81-badc8e95eed8.png)

ckamm:
@jnweiger I've tried to reproduce with 240d97686808800ec51fa170d7816b5860e74fd6 (current 2.6) and 10.2.1RC1.
[...]

PR #7298, opened 2019-07-08
Sharing: Create shares with appropriate permissions

ckamm:
If user A shares something with user B and allows resharing but removes
some other permission then user B should be able to share the item or
any of its contents with user C and the new share permissions should
automatically be <= the permissions of the original share.

Currently this works correctly when resharing the original item but
fails for any subitem, preventing the reshare entirely.

With this patch the reshare will always be created with appropriately
limited permissions.

For #7275


ckamm:
This could potentially go into 2.5.

ckamm:
Retargeted to 2.5.

Issue #7313, opened 2019-07-16
Cannot edit public link expiration date when it is enforced


HanaGemela:
Client: 2.6.0alpha2 (build 12128)
macOS: 10.14.5
Server: 10.2.1. (stable)

Steps to recreate:
1) In the client create a public link without an expiration date
2) On the server go to Settings -> Admin -> Sharing
3) Tick 'Enforce expiration date' checkbox
4) Go back to the client and set expiration against the link created in the step 1 (Client restart might be needed to get the new settings from the server)

Actual result: 'Set expiration date' is greyed out so it cannot be changed
Expected result: Expiration date can be set

TheOneRing:
@HanaGemela can you have a look and close this issue if its fixed :)

Issue #7247, opened 2019-06-16
New "password required" capabilities lead to incorrect behavior


ckamm:
The server 10.2.0 (edit: actually since 10.0.8, see below) has new fine-grained capabilities for requiring passwords for link shares. This setting can now be controlled independently for "read-only", "read/write" and "upload only" link shares.

The client should support these new capabilities.

However, there's an additional complication: it seems that if any one of these settings is enabled, the server will set the previous files_sharing/public/password/enforced capability to 1. This is unfortunate because it'll make the client require passwords for *all* public link shares when the server may only require it for one of the three types.

Found while testing with 10.2.0 for #7245.

ckamm:
Moving to 2.6.1 since it doesn't need to block 2.6.0 release. Users can still create shares, they just may be required to set passwords when that's not actually required for the particular link share type.

pmaier1:
This has been added with 10.0.8 more than one year ago. https://doc.owncloud.com/server/10.2/admin_manual/release_notes.html#new-option-to-granularly-configure-public-link-password-enforcement

ckamm:
@pmaier1 Thanks, I hadn't looked it up yet. Unfortunate we didn't see it earlier, but at least not urgent then.

michaelstingl:
Can this be solved with https://github.com/owncloud/client/issues/7487

michaelstingl:
fixed

PR #7318, opened 2019-07-17
Link sharing ui improvements

ckamm:
- Respect fine-grained "password required" capability #7247
- Allow password/expiration to be added to shares that were created
  without these at a time when the server allowed that #7313


guruz:
Pending discussion in https://github.com/owncloud/client/pull/7564
CC @michaelstingl

michaelstingl:
> Pending discussion in #7564
> CC @michaelstingl

"Hybrid" approach planned mid-term, so I think we'll need this before.

ogoffart:
I fixed the merge conflict based on what i think the code was supposed to do, but this should be reviewed again i guess.

TheOneRing:
I tested the share dilaog and it looks good

could you add a changelog entry?

PR #7319, opened 2019-07-17
Vfs: Improve strings for availability states

ckamm:
"Currently always" read very oddly.


guruz:
@michaelstingl @lefherz this means we need to take some time translating between alpha2 and release

Issue #7329, opened 2019-07-19
Virtual files stay in 'Sync pending' state after sync run


HanaGemela:
Client: 2.6.0alpha2 (build 12097)
Server: 10.2.1 (stable)
OS: Windows 10

Steps to recreate:
1) Turn on VFS
2) Force sync now or just wait for next sync

Actual result: Files and folders are in 'Sync pending' state
Expected result: Files and folders are not pending anymore

![image](https://user-images.githubusercontent.com/49001702/61518905-f6e35800-aa0a-11e9-8e2a-7e7e4b35becc.png)

HanaGemela:
Works as expected on Windows 10, client 2.6.0beta1 (build 12201)

PR #7334, opened 2019-07-23
Vfs: Add hook to allow update-metadata for unchanged files

ckamm:
Allows winvfs to convert files to placeholders when vfs is enabled. This
is required to mark files as in-sync #7329.

Needs matching winvfs change.


ckamm:
@ogoffart There are tests on the winvfs side. Neither suffix or off vfs modes need the hook.

PR #7344, opened 2019-07-25
Vfs: Require local discovery after disabling vfs

ckamm:
Without it local files aren't guaranteed to be downloaded #6936.


ckamm:
I still don't yet know what exactly goes wrong without it, but this should help. Hope to add tests later.

Issue #7345, opened 2019-07-26
When a new folder is created it always syncs twice, even on success


ckamm:
Create a new sync folder. Wait for the initial sync to finish. A second sync will be triggered.

The log shows:
```
Compare etag with previous etag: last: "\"foo\"", received: "foo" -> CHANGED
```

That means there's still a place where etag parsing isn't done consistently. This appears now because I fixed a different (more rare) case where etag parsing wasn't done.

HanaGemela:
Works as expected on Windows 10, 2.6.0beta1 (build 12201)

PR #7346, opened 2019-07-26
Discovery: Parse etag to be consistent with RequestEtagJob

ckamm:
This avoids unnecessary sync runs.

For #7345


ogoffart:
(Perhaps we should change etag to be a QByteArray everywhere...)

Issue #7336, opened 2019-07-23
Selective sync menu not opened by default


HanaGemela:
Another issue on icons https://github.com/owncloud/client/issues/7321

Client: 2.6.0alpha2 (build 12128)
macOS: 10.14.5
Server: 10.2.1. (stable)

Steps to recreate:
1) Add a new account 
2) Check the account tab

Actual result: The arrow points down which means that the menu is expanded and the list is empty - there are no folders in the root directory

![image](https://user-images.githubusercontent.com/49001702/61721138-45c12280-ad68-11e9-8597-3797ba3d6f43.png)

User has to click on the arrow twice to open the list
This hides not only the selective sync list but also the warning message about what should be synced 

![image](https://user-images.githubusercontent.com/49001702/61721197-625d5a80-ad68-11e9-9077-ecdfd0d81a05.png)

Expected result: The list is expanded and all warning messages are shown by default once the account has been added so that the user knows that some actions are required and it is obvious why the files have not been synced yet

ckamm:
indeed. Does not happen with folder wizard, only if the folder is created as part of the account setup. It does not matter whether there are "big" folders in the folder.

ckamm:
Let's leave this for 2.6.1 to move ahead with 2.6.0 now.

HanaGemela:
Works as expected on macOS 10.14.6, client 2.6.0beta1 (build 12241)

HanaGemela:
Broken again in 2.6.1sprint1 (build 12882), macOS 10.15.1
P2 as it is a regression

TheOneRing:
@HanaGemela can't repoduce

HanaGemela:
@TheOneRing Still broken in 2.6.1 (build 13096), macOS. Steps to reproduce are correct

<img width="718" alt="Screenshot 2020-02-04 at 09 28 03" src="https://user-images.githubusercontent.com/49001702/73726944-c528a900-4730-11ea-9489-7f91e474fdd6.png">

michaelstingl:
fixed

PR #7347, opened 2019-07-26
AccountSettings: Fetch subitems after wiping them

ckamm:
That helps avoid empty lists after account creation #7336


ckamm:
Was requested to be in 2.6.0

Issue #7363, opened 2019-08-01
Sync errors not always cleared before sync starts


ckamm:
From @HanaGemela at https://github.com/owncloud/client/issues/6972#issuecomment-516816927

Steps to recreate:
1) Pause the sync when "Checking for changes in ..."  message is shown
2) Resume the sync
Actual result: The message says Aborted and Synchronizing at the same time
![image](https://user-images.githubusercontent.com/49001702/62209667-7faeac00-b39a-11e9-94b8-6e5cb1342746.png)

TheOneRing:
I haven't seen this issue recently.
I assume it no longer occurs.

PR #7364, opened 2019-08-01
Fix syncstatus and progress reporting errors

ckamm:
Relates to #7363


src/libsync/theme.cpp
```diff
@@ -81,9 +84,6 @@ QString Theme::statusHeaderText(SyncResult::Status status) const
     case SyncResult::SetupError:
         resultStr = QCoreApplication::translate("theme", "Setup Error");
         break;
-    case SyncResult::SyncPrepare:
-        resultStr = QCoreApplication::translate("theme", "Preparing to sync");
```
ogoffart: I think the idea is that user complains "Why is the sync running when there should be nothing to sync".
Maybe that is no longer relevant because of other changes of the UI.  But probablt it still deserve a different message.

Issue #7368, opened 2019-08-06
Crash after switching between VFS and Selective sync 


HanaGemela:
I've moved the comments from https://github.com/owncloud/client/issues/7281 to a separate issue

Switching between VFS and Selective sync quickly causes a crash:
bf781858-2481-49d7-85b9-f7f058bc5ea9

When turning on the client after previous crash caused by the scenario described above:
640a9068-e964-4cd6-bd5c-3627b382ab6c 

Another scenario:
Server: 10.2.2 (stable)
Client: macOS 2.6.0.12241beta1
User has an unavailable folder, an unavailable file (503 errors), a big file... then switches between VFS and selective sync and then the client crashes. 
This time I cannot start the client again. On each start the client crashes immediately (e.g. 341ceedf-abad-47d1-b035-462f7e8c7f92 or cb0728c9-8c84-467d-9b0c-4d79d52f7db3). Logs shares

HanaGemela:
I've tried with newer build with Ot 5.12.4 (2.6.0-daily20190806 (build 12251))
crash: 8a0c1b20-e22d-4071-ad4d-ce8034e7c130

guruz:
Havent looked at the bt, but maybe 2.6.0?! @ckamm

ckamm:
Seems to be a crash in `calcCryptoHash()`
- https://sentry.io/organizations/owncloud/issues/1142050016/events/8a0c1b20e22d4071ad4dce8034e7c130/?project=79001&statsPeriod=14d
- with deeper stacktrace: https://sentry.io/organizations/owncloud/issues/1142008010/events/cb0728c98c84467d9b0c4d79d52f7db3/?project=79001&statsPeriod=14d

ckamm:
It crashes in `QObject::objectName()`. I assume the QIODevice is somehow deleted while the checksum computation is running.

HanaGemela:
Cannot recreate in 2.6.0rc1 (build 12411), believed to be fixed. mac 10.14.6

PR #7369, opened 2019-08-07
Checksums: Fix crash due to threading issue

ckamm:
The checksum computation thread was potentially using a QFile that was
deleted in the gui thread.

For #7368


guruz:
Doesn't Qt Concurrent have some inherent machine+platform specific limits on concurrency etc?
I remember QThreadPool had that.
I'm asking because of this "Bug" comment..
Would be a pity if things get 'stuck' somehow

ogoffart:
> Doesn't Qt Concurrent have some inherent machine+platform specific limits on concurrency etc?

Yes, by default, it only uses as many thread as logical cores on the machine.

But that's unrelated to this issue.

Issue #7372, opened 2019-08-09
[2.6.0 beta1] Sync aborts with "Too many open files"


dragotin:
Sync aborts with error messages "Too many open files" in one of my sync dirs. It  contains files of the crypto container cryfs with around 5000 files:

```
kf:~/ownCloud] $ find vault | wc -l
4948
[kf:~/ownCloud] $ find vault -type f| wc -l
2902
[kf:~/ownCloud] $ find vault -type d| wc -l
2046
```
After a while of syncing that dir, the client bails out with error. The logfile contains entries like this:

```
...
08-09 08:41:22:638 [ info sync.propagator ]:    Starting INSTRUCTION_CONFLICT propagation of "vault/A46/0348425F253D030C94633DD718E50" by OCC::PropagateDownloadFile(0x55a138eda310)
08-09 08:41:22:638 [ info sync.checksums ]:     Computing "SHA1" checksum of "/home/kf/ownCloud/vault/A46/0348425F253D030C94633DD718E50" in a thread
08-09 08:41:22:638 [ info sync.checksums ]:     Computing "SHA1" checksum of iodevice in a thread
08-09 08:41:22:638 [ info sync.propagator ]:    Starting INSTRUCTION_CONFLICT propagation of "vault/A47/D1ED17C084F24555047ECF3C7AB1E" by OCC::PropagateDownloadFile(0x55a138edb1f0)
08-09 08:41:22:638 [ info sync.checksums ]:     Computing "SHA1" checksum of "/home/kf/ownCloud/vault/A47/D1ED17C084F24555047ECF3C7AB1E" in a thread
08-09 08:41:22:638 [ info sync.checksums ]:     Computing "SHA1" checksum of iodevice in a thread
08-09 08:41:22:638 [ info sync.propagator ]:    Starting INSTRUCTION_CONFLICT propagation of "vault/A47/E7C0D8E9317B6E049C3DC0FAED675" by OCC::PropagateDownloadFile(0x55a138edc080)
08-09 08:41:22:638 [ info sync.checksums ]:     Computing "SHA1" checksum of "/home/kf/ownCloud/vault/A47/E7C0D8E9317B6E049C3DC0FAED675" in a thread
08-09 08:41:22:638 [ warning sync.checksums ]:  Could not open file "/home/kf/ownCloud/vault/A47/E7C0D8E9317B6E049C3DC0FAED675" for reading to compute a checksum "Too many open files"
[...]

PR #7373, opened 2019-08-12
Download: Don't trigger too many concurrent hash computations

ckamm:
Previously the job would only become "active" when the downloads
started. That meant that arbitrarily many hash computations could be
queued at the same time.

Since the the file was opened during future creation this could lead to
a "too many open files" problem if there were lots of new-new conflicts.

To change this:
- Make PropagateDownload become active when computing a hash
  asynchronously.
- Make the computation future open the file only once it gets run. This
  will make it less likely for this problem to occur even if thousands
  of these futures are queued.

For #7372


guruz:
Yeah that's what I had meant in https://github.com/owncloud/client/pull/7369#issuecomment-519064743 ...

PR #7378, opened 2019-08-12
Updates to doc for configuration file

ckamm:
For owncloud/docs#1365

### DIFF
--- a/docs/modules/ROOT/pages/advanced_usage/configuration_file.adoc
+++ b/docs/modules/ROOT/pages/advanced_usage/configuration_file.adoc
@@ -41,14 +41,33 @@ Some interesting values that can be set on the configuration file are:
 | Variable | Default | Meaning 
 
 | `chunkSize` 
-| `5242880` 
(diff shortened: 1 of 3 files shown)


HanaGemela:
@ckamm I'm looking at my 2.6. owncloud.cfg and I don't have [ownCloud] section at all. Also [General] includes different options. And [BWLimit] is missing in the documentation

ckamm:
I was assuming we'd only document options that can't easily be set from the ui. If we want to document all options there are indeed a lot of missing ones.

HanaGemela:
@michaelstingl what exactly do we want to document? Everything or just things that cannot be set in the ui?

michaelstingl:
> @michaelstingl what exactly do we want to document?

@patrickjahns opened the issue.

> Everything or just things that cannot be set in the ui?

For me, this would be sufficient. Configure stuff automatically that can be set in the UI is maybe part of a bigger task about provisioning… (https://github.com/owncloud/client/pull/6761)

PR #7379, opened 2019-08-13
Print critical and fatal messages to stderr

TheOneRing:
That way we might get a reason why the client crashes (dev setup)

### DIFF
--- a/src/libsync/logger.cpp
+++ b/src/libsync/logger.cpp
@@ -22,6 +22,8 @@
 #include <QtGlobal>
 #include <qmetaobject.h>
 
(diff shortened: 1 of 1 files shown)


guruz:
Appveoyor borks while downloading libpcre
https://ci.appveyor.com/project/ownclouders/client/builds/26658848/job/dmjhfa473olpapdh

guruz (approved):
This looks good enough for targetting 2.6 instead of master?
(Do you agree @ckamm ?)

ckamm:
Yes, I'm fine with this targeting 2.6

PR #7384, opened 2019-08-14
Fix update in case auto updater URL does a redirect (e.g. if trailing slash in URL is missing)

TheOneRing:


guruz:
I'm not sure I understand what this fixes.. see rocket chat.
I agree with better log output though...

guruz:
@TheOneRing Can you prefix commit messages with the area they are in? e.g. "Autoupdater: Improve logging" and "Autoupdater: Follow HTTP Redirects"

guruz:
@ogoffart @ckamm I'm not sure we want redirect handling for the auto updater URL. Could be a security problem?

guruz:
It seems that Sparkle also follows the redirects, so maybe it's not an issue.
(Or maybe Sparkle does a check if it is on same host?! I haven't tried..)

guruz:
> It seems that Sparkle also follows the redirects, so maybe it's not an issue.

Sparkle uses file signatures though (the .sig linked from the XML)....

jnweiger:
Regarding the securiy aspect:
- Qt's FollowRedirectsAttribute does not follow insecure downgrades from https to http.
- I'd argue that it is a fixed string URL pointing only to servers under our or customer admin control.
- If an end user makes an attempt to point the client elsewhere to fetch updates via OCC_UPDATER_URL environment variable, then he would bypass any policies anyhow.

PR #7386, opened 2019-08-14
Updater: Improve logging for version comparison

TheOneRing:
As requested by qa


TheOneRing:
Well as we would only getting this message on windows its pointless to say "Windows:".
And if you really want it to be more specific we should introduce a new category.

guruz:
@TheOneRing I'm talking about the COMMIT message :-) the code is fine

PR #7394, opened 2019-08-15
OwncloudWizard: Mark vfs as tech preview instead of experimental

TheOneRing:
Related issue: https://github.com/owncloud/client/issues/7406

### DIFF
--- a/src/gui/folderwizard.cpp
+++ b/src/gui/folderwizard.cpp
@@ -483,7 +483,8 @@ FolderWizardSelectiveSync::FolderWizardSelectiveSync(const AccountPtr &account)
     layout->addWidget(_selectiveSync);
 
     if (Theme::instance()->showVirtualFilesOption() && bestAvailableVfsMode() != Vfs::Off) {
-        _virtualFilesCheckBox = new QCheckBox(tr("Use virtual files instead of downloading content immediately (experimental)"));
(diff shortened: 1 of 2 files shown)


HanaGemela:
The change of the button in the three dots menu in Account tab is missing. The button is 'Enable virtual file support (experimental)'

michaelstingl:
Can you confirm, it's only "tech preview" with Win10 1709+ where we use `CFAPI` ?

TheOneRing:
The code should only be triggered when the plugin is available so we should only see those messages on recent version of windows.

src/gui/folderwizard.cpp
```diff
@@ -483,7 +483,8 @@ FolderWizardSelectiveSync::FolderWizardSelectiveSync(const AccountPtr &account)
     layout->addWidget(_selectiveSync);
 
     if (Theme::instance()->showVirtualFilesOption() && bestAvailableVfsMode() != Vfs::Off) {
-        _virtualFilesCheckBox = new QCheckBox(tr("Use virtual files instead of downloading content immediately (experimental)"));
+        _virtualFilesCheckBox = new QCheckBox(tr("Use virtual files instead of downloading content immediately %1").arg(
+                                                  Utility::isWindows() ? tr("(tech preview)") : tr("(experimental)")));
```
ckamm: Check if `bestAvailableVfsMode()` is the `WindowsCfApi` one. Otherwise older windows versions will show tech preview here even though they're using suffix files.

PR #7396, opened 2019-08-15
Checksums: Add back QIODevice api

ckamm:
Because the winvfs plugin needs it. But be more careful about the
device's lifetime this time.

This will need matching winvfs changes. I'll also merge 2.6 into master and update both winvfs branches.


TheOneRing:
LGTM

PR #7404, opened 2019-08-16
Utility: Implement hasDarkSystray_private on Windows

TheOneRing:


TheOneRing:
Naming and QWidget dep handling might be improved....

TheOneRing:
Naming should be discussed 😅

src/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@
 if(NOT TOKEN_AUTH_ONLY)
 endif()
 
-find_package(Qt5 5.6 COMPONENTS Core Network Xml Concurrent REQUIRED)
+find_package(Qt5 5.6 COMPONENTS Core Network Xml Concurrent Gui REQUIRED)
```
guruz: Hm I'm not sure about this one :-/

src/common/winrthelper/winrthelper.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include <Windows.UI.ViewManagement.h>
+#include <QColor>
+
+namespace  {
+
+QColor GetThemeColor(ABI::Windows::UI::ViewManagement::UIColorType type)
+{
+    // Error checking has been elided for expository purposes.
+    Microsoft::WRL::ComPtr<ABI::Windows::UI::ViewManagement::IUISettings3> settings;
+    if (SUCCEEDED(Windows::Foundation::ActivateInstance(Microsoft::WRL::Wrappers::HStringReference(
```
guruz: Is this a problem when executed on older windows versions I wonder?

TheOneRing: The code is com based so it should not make a problem, the only thing is the new library I link

src/CMakeLists.txt
TheOneRing: Me neither, I just use it for QColor, could implement it manually but...
In general I'm not 100% happy how the utility code is handled here.

ckamm: We don't want libsync to depend on QtGui. I think this was a problem before. Possibly `Theme` should be split into a sync and a gui part, ideally in a way that makes ownbrander continue to work. 

You could look at doing that, but I guess that avoiding QColor here will be easier.

TheOneRing: I removed the dep to qtgui, and computate the lightness myself

TheOneRing:
> 
> 
> Why do we need a new library, cant we just add winrthelper.cpp to csync?
> 
> I can't tell if linking against runtimeobject is allowed on all supported windows. I trust you on that.

I tried to make it even simpler by just setting the c++ standard for the utility_win.cpp to c++17 but it seems that has only an effect on targets, not source files.

TheOneRing:
For the support of older versions of Windows, I'll try it in a vm. If it fails I could make the helper lib dynamic and try to load it during runtime?

dschmidt:
> I tried to make it even simpler by just setting the c++ standard for the utility_win.cpp to c++17 but it seems that has only an effect on targets, not source files.

Out of curiosity what in your code does need C++17?

TheOneRing:
> 
> 
> > I tried to make it even simpler by just setting the c++ standard for the utility_win.cpp to c++17 but it seems that has only an effect on targets, not source files.
> 
> Just out of curiosity: why do you need C++17?

Because the window rt headers require it....

TheOneRing:
image
So a plugin or dynamic loaded lib?

dschmidt:
> So a plugin or dynamic loaded lib?

Isn't a plugin a dynamically loaded lib? :-P 
If you mean QPlugin vs. a simple lib exporting a single function ... I don't mind.

TheOneRing:
> 
> 
> > So a plugin or dynamic loaded lib?
> 
> Isn't a plugin a dynamically loaded lib? :-P
> If you mean QPlugin vs. a simple lib exporting a single function ... I don't mind.

Thats what I meant

ogoffart:
> I tried to make it even simpler by just setting the c++ standard for the utility_win.cpp to c++17 but it seems that has only an effect on targets, not source files.

Would it not be even simpler to set the C++17 support to the whole project on windows?
If windows rt requires it anyway....

TheOneRing:
Well in my experience c++14 is already quite progressive and I'm against different c++ standards for different platforms.
But bumping the requirements on linux is a bit harder, with all the distributions out there.

TheOneRing:
My test was a custom build and is not representative.
I'll test it again with an official ci build, as far as I can see we would normally ship those libs.

TheOneRing:
Tested on Windows 7 and 8.
Depends on Windows 8.

guruz:
We could make it depend on the 64bit build and say that one will only have it.

TheOneRing:
I would rather introduce an explicit option to set the minimum Windows version required

guruz:
> I would rather introduce an explicit option to set the minimum Windows version required

What does that mean? Break old Windows versions?

TheOneRing:
option(BUILD_WINRT_FEATURES...

TheOneRing:
I fixed the issues by dynamically loading the component

changelog/unreleased/7356
```diff
@@ -0,0 +1,5 @@
+Change: Detect the Windows 10 thme for the system tray
```
dschmidt, line 1: *theme

src/common/utility_win.cpp
```diff
@@ -29,10 +29,29 @@
...
 
-static const char systemRunPathC[] = "HKEY_LOCAL_MACHINE\\Software\\Microsoft\\Windows\\CurrentVersion\\Run";
-static const char runPathC[] = "HKEY_CURRENT_USER\\Software\\Microsoft\\Windows\\CurrentVersion\\Run";
+namespace {
+const char systemRunPathC[] = "HKEY_LOCAL_MACHINE\\Software\\Microsoft\\Windows\\CurrentVersion\\Run";
+const char runPathC[] = "HKEY_CURRENT_USER\\Software\\Microsoft\\Windows\\CurrentVersion\\Run";
+const QString rtHelperNameC()
+{
+    return QStringLiteral(APPLICATION_EXECUTABLE "_winrt");
+}
```
dschmidt, line 43: could be a const char[] just like the defines above, no?

dschmidt (approved):

src/common/utility_win.cpp
TheOneRing: the first parameter from QLibaray takes a QString not char array, anyhow I will pass a absolute string to ensure we are loading the right lib...

PR #7405, opened 2019-08-16
Don't set gnu style flags with msvc

TheOneRing:
The settings are enabled by default with msvc

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -35,8 +35,10 @@ endif()
 
 if(WIN32)
   # Enable DEP & ASLR
-  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,--nxcompat -Wl,--dynamicbase")
-  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--nxcompat -Wl,--dynamicbase")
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -35,8 +35,10 @@ endif()
 
 if(WIN32)
   # Enable DEP & ASLR
-  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,--nxcompat -Wl,--dynamicbase")
-  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--nxcompat -Wl,--dynamicbase")
+  if (MINGW)
+    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,--nxcompat -Wl,--dynamicbase")
+    set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--nxcompat -Wl,--dynamicbase")
```
ckamm: Meaning they did nothing? Do we already enable `/DYNAMICBASE` for msvc in another place then?

TheOneRing: Nope they are just the msvc default so setting them would be a noop.
Setting the gnu flags however creates a whole bunch of warnings.

PR #7415, opened 2019-08-20
AccountSettings: Use switch, case to ensure we handle all cases

TheOneRing:


TheOneRing:
Can we introduce a clang format config?

src/gui/accountsettings.cpp
```diff
@@ -845,19 +845,28 @@ void AccountSettings::slotAccountStateChanged()
             serverWithUser = tr("%1 as <i>%2</i>").arg(server, Utility::escape(user));
         }
 
-        if (state == AccountState::Connected) {
+        switch(state)
```
ogoffart: (nitpick:  we follow the Qt style which put the brace on the same line  (here and after the case)

ogoffart:
> Can we introduce a clang format config?

There is one  (`.clang-format` in the repo root)

I recommand you use `git-clang-format HEAD^`  to fixup the commits

PR #7418, opened 2019-08-21
Don't obfuscate code with mutable lambda

TheOneRing:


src/libsync/discovery.cpp
```diff
@@ -439,7 +439,7 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(
     item->_modtime = serverEntry.modtime;
     item->_size = serverEntry.size;
 
-    auto postProcessServerNew = [item, this, path, serverEntry, localEntry, dbEntry] () mutable {
+    auto postProcessServerNew = [=, &path] () {
```
ogoffart: This was a lambda because it was called twice before commit b9f17aeaf1d39c4a4fb1da3b15c1c9ea91b85bc8,  but i guess now the code no longer need to be in a lambda and can just be moved in the if bellow.

src/libsync/discovery.cpp
```diff
@@ -572,7 +572,7 @@ void ProcessDirectoryJob::processFileAnalyzeRemoteInfo(
             // we need to make a request to the server to know that the original file is deleted on the server
             _pendingAsyncJobs++;
             auto job = new RequestEtagJob(_discoveryData->_account, originalPath, this);
-            connect(job, &RequestEtagJob::finishedWithResult, this, [=](const HttpResult<QString> &etag) mutable {
+            connect(job, &RequestEtagJob::finishedWithResult, this, [=, &path](const HttpResult<QString> &etag) {
```
ogoffart: I don't think that's right since the job will be done async so the path variable will be a dangling pointer.

src/libsync/discovery.cpp
```diff
@@ -943,7 +943,8 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
             if (base.isVirtualFile() && isVfsWithSuffix())
                 chopVirtualFileSuffix(serverOriginalPath);
             auto job = new RequestEtagJob(_discoveryData->_account, serverOriginalPath, this);
-            connect(job, &RequestEtagJob::finishedWithResult, this, [=](const HttpResult<QString> &etag) mutable {
+            connect(job, &RequestEtagJob::finishedWithResult, this, [=, &path](const HttpResult<QString> &etag) {
```
ogoffart: ditto.

TheOneRing:
So yes it only made the appearance that path was changed in the last line of the lambda, which would not have made any sense.

While my code now not fixes any issue, it still breaks one unittest, so it looks like we really need a stateful/mutable unittest which in my opinion is not a good code style...

TheOneRing:
I now use local copies instead of mutable lambdas.
In my opinion its clearer what happens.
Also when a mutable lambda is executed multiple times, it has a different state each time...

ogoffart:
Right, if you think it is clearer... :+1: 
Note that it would be even more clear if `postProcessServerNew` was not a lambda (since it does not need to be called from two location anymore)
And that all of these labmda are only called once.

TheOneRing:
Its still used in 2 locations so....
With the size of the function it might even make sense to refactor to use some more member functions to split things up.

PR #7424, opened 2019-08-22
Setup: Change Windows virtual files description to tech preview

TheOneRing:


HanaGemela:
@TheOneRing I still think the button in three dots menu is not changed

PR #7425, opened 2019-08-22
Update appveyor settings

TheOneRing:


TheOneRing:
So we need to fix the unit tests but that's unrelated.
@dschmidt ?

PR #7426, opened 2019-08-22
Port to Drone 1.x and handle translations in this repo

dschmidt:


ckamm:
@dschmidt Why is there a transifex commit in the PR?

dschmidt:
> @dschmidt Why is there a transifex commit in the PR?

Basically "because the PR works"  :trollface: 

I configured the cron job so it runs on my branch. Probably I forked between a commit and the next transifex run, that's why there are a few changes. Subsequent runs did not trigger further commits as you can see:
https://drone.owncloud.com/owncloud/client/

(also because I made the order of NSIS translations stable)

When merged I will change the cron job to the 2.6 branch and raise the execution interval back to `@daily`.

ckamm:
Ok!

guruz:
Please next time prefix your commit messages with the topic, e.g. "drone:" or "translations:"
thanks :)

Issue #7419, opened 2019-08-21
Recipient cannot access shared files


HanaGemela:
Windows 10
Client: 2.6.0daily20190820 (build 12286)
Server: 10.3.0 alpha, 10.2

Steps to recreate:

1. User A shares folder Test with user B, granting edit-create + edit-delete  permissions.
2. User A creates folders Test/Sub, Test/testfolder and file Test/file.txt
3. User B is connected with a desktop client, folder Test is synced.

Actual result: Access to files denied
Expected result: Files are accessible

<img width="702" alt="Screenshot 2019-08-21 at 11 53 37" src="https://user-images.githubusercontent.com/49001702/63422402-929d2500-c40a-11e9-98ed-fd1407263370.png">

ckamm:
Probably relates to https://github.com/owncloud/core/pull/35884

ckamm:
@HanaGemela Do you have a log file? Particularly the lines around "Received sharing permissions for" when user A shares the folder initially are interesting. 

The client is intended to discover the maximum possible permissions for sharing of an item and then create a user share with that. See `ShareDialog::slotPropfindReceived` for maximum sharing permissions retrieval.

ckamm:
I can't reproduce the problem. Spoke to @HanaGemela, she'll also try to reproduce and save logs.

HanaGemela:
I cannot recreate either. The below is the error for the file not synced yesterday 
```08-22 15:44:27:380 [ debug gui.socketapi ]	[ OCC::SocketApi::slotUpdateFolderView ]:	Not sending UPDATE_VIEW for "2" because status() is 3
08-22 15:44:27:380 [ info gui.application ]:	Sync state changed for folder  "http://10.42.16.185:8080/remote.php/dav/files/carol/" :  "Sync Running"
08-22 15:44:27:380 [ info sync.propagator ]:	Starting INSTRUCTION_NEW propagation of "Test4/file.txt.txt" by OCC::PropagateDownloadFile(0x8e82ee0)
08-22 15:44:27:380 [ debug sync.propagator.download ]	[ OCC::PropagateDownloadFile::start ]:	"Test4/file.txt.txt" 0
08-22 15:44:27:380 [ debug sync.propagator ]	[ OCC::OwncloudPropagator::localFileNameClash ]:	CaseClashCheck for  "C:/Users/Hana Gemela/ownCloud17/Test4/file.txt.txt"
[...]

PR #7427, opened 2019-08-23
PropagateDownload: Don't try to open readonly temporaries

ckamm:
This situation could arrise when receiving a read-only share and the
temporary rename failed for some reason.

See #7419


ckamm:
We set files shared read-only to read-only on the filesystem. Typically we adjust the permissions of the temporary at the end and then do the rename-replace. So when the rename-replace fails for some reason the temporary is still left read-only.

ckamm:
I'll write a test separately so QA can continue sooner.

Issue #7429, opened 2019-08-26
Permissions options not shown when only single option is available for a public link


HanaGemela:
Server: 10.3.0 alpha
Client: 2.5.4 (build 11415)

Steps to recreate:
1) Disable 'Allow public uploads' in server sharing setting
2) Create a new public link with Download/View permissions

Actual result: No options to pick permissions
Expected result: User can see what permissions are set on the public link - Download/View option is shown, pre-selected and greyed out. Other options not shown at all

<img width="412" alt="Screenshot 2019-08-26 at 11 04 05" src="https://user-images.githubusercontent.com/49001702/63679537-cd7bd000-c7f1-11e9-9a7c-8a4a46a96d86.png">

HanaGemela:
The same for files

HanaGemela:
The below is server 10.2. where user can select the permissions

<img width="410" alt="Screenshot 2019-08-26 at 17 29 30" src="https://user-images.githubusercontent.com/49001702/63702382-26b22680-c827-11e9-90e1-30a59cdd7e98.png">

ckamm:
For files one never sees the permission combo box. Users cannot upload via file link shares ever, as far as I'm aware.

For link-shared directories:
- if the `files_sharing/public/upload` capability is enabled, the permission selection is shown
- if the `files_sharing/public/supports_upload_only` capability is enabled, the "Upload only" option is shown

ckamm:
Ok, I think I understand what you mean: You'd prefer if the single available option was visible.

HanaGemela:
@ckamm yes, I think users should know what permissions the link has. So we should show them all available options (even though in this case it is only one single option)

HanaGemela:
Works as expected on client 2.6.0rc1 (build 12411), mac 10.14.6

PR #7432, opened 2019-08-27
ShareLinks: Always show at least readonly permissions

ckamm:
For #7429

The only reason I'd have this in 2.6.0 is because there was a bug where the new sharing permission labels weren't hidden along with the radio button options.


ckamm:
@HanaGemela This patch makes unavailable permissions invisible, they're not just disabled. As far as I know that's usually what's done for things that admins have disabled via capabilities.

Issue #7410, opened 2019-08-20
Renaming deletes the original file when user has delete but not change/create permissions


HanaGemela:
Windows 10
Client: 2.6.0daily20190820 (build 12286)
Server: 10.3.0 alpha, 10.2.1.4

Steps to recreate:
1) User1 shares a folder with a group without change and create permissions. The folder includes one file 
2) User2 (member of that group) renames the file

Actual result: File has been deleted from User1
Expected result: Account of User1 unchanged, User2 got a warning message that renaming (changing) not allowed

guruz:
If this only happens with 10.3 alpha,are you in the right repo?

sharidas:
At least in the webUI where user2 requests for MOVE is made with url `remote.php/dav/files/$username/$foldername` dav Forbidden exception is thrown.

michaelstingl:
Compare requests from client 2.6.0 and 10.x web UI ? @jnweiger Could you have a look?

patrickjahns:
@HanaGemela 
Is this only an issue on 10.3.0alpha - or also with 10.2.x ?

HanaGemela:
@patrickjahns also 10.2 as mentioned in the description

jnweiger:
Server involved should be irrelevant here. It is a dataloss, that the client can easily prevent all by itself, 

10.3 web interface: Rename of a file or folder is not available in this case -> No problem.
10.2 web interface: Rename of a file or folder is not available in this case -> No problem.

Client probably goes down the pipe when MOVE throws a forbidden exception and it tries to workaround with copy and delete locally.

The new local file won't sync to the server, but the delete action will. 
By breaking an atomic move action up into two unrelated actions, we create a data loss. -> reassign @guruz

jnweiger:
The code responsible for this behaviour is
https://github.com/owncloud/client/blob/2.6/src/libsync/discovery.cpp#L1231

I read this as: 
* we see, that server does not allow the new file to be created. Missing hasPermission(RemotePermissions::CanAddFile) for the new name.
* we decide that a rename would not work, and to avoid more rename attempts we remove the etag.
* another effect of removing the etag is: The "move transaction" breaks apart into whatever independant(!) actions the next sync run will decide to do.
* The next sync run discovers changes on both oldname and newname and schedules actions for each of them. One delete, one create.
* Bummer.

jnweiger:
@ogoffart do you remember if this is intended behaviour? I see your name near that code...

[...]

PR #7436, opened 2019-08-28
Restore source when move to target fails

ckamm:
For #7410

### DIFF
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -644,7 +644,7 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
         // conflict we don't need to recurse into it. (local c1.owncloud, c1/ ; remote: c1)
         if (item->_instruction == CSYNC_INSTRUCTION_CONFLICT && !item->isDirectory())
             recurse = false;
-        if (_queryLocal != NormalQuery && _queryServer != NormalQuery && !item->_isRestoration)
+        if (_queryLocal != NormalQuery && _queryServer != NormalQuery)
             recurse = false;
 
         auto recurseQueryLocal = _queryLocal == ParentNotChanged ? ParentNotChanged : localEntry.isDirectory || item->_instruction == CSYNC_INSTRUCTION_RENAME ? NormalQuery : ParentDontExist;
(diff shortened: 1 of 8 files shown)


test/testsyncengine.cpp
```diff
@@ -714,6 +726,114 @@ private slots:
...
+
+    static void setAllPerm(FileInfo *fi, RemotePermissions perm)
+    {
+        fi->permissions = perm;
+        for (auto &subFi : fi->children)
+            setAllPerm(&subFi, perm);
+    }
+
+    // What happens if the source can't be moved or the target can't be created?
+    void testForbiddenMoves()
```
ogoffart: move to testpermissions.cpp ?  since testsyncengine is big already, and that's where i would expect it.

src/libsync/discovery.cpp
```diff
@@ -890,77 +890,105 @@ void ProcessDirectoryJob::processFileAnalyzeLocalInfo(
...
+        // Here we know the new location can't be uploaded: must prevent the source delete.
+        // Two cases: either the source item was already processed or not.
+        auto wasDeletedOnClient = _discoveryData->findAndCancelDeletedJob(originalPath);
         if (wasDeletedOnClient.first) {
-            recurseQueryServer = wasDeletedOnClient.second == base._etag ? ParentNotChanged : NormalQuery;
-            processRename(path);
+            // More complicated. The REMOVE is canceled. Restore will happen next sync.
+            qCInfo(lcDisco) << "Undid remove instruction on source" << originalPath;
+            _discoveryData->_statedb->deleteFileRecord(originalPath, true);
+            _discoveryData->_anotherSyncNeeded = true;
```
ogoffart: Is it not kind of simple to do that now?

We can just create a new item with CSYNC_INSTRUCTION_NEW and _isRestoration = true.
Or is there something more complicated i don't see?

ckamm: We may need to restore a whole subtree recursively.

ckamm:
There's also another issue in permissionstest that I need to fix.

ckamm:
The remaining problem is that when:
```
Tree:
A/
A/S
A/S/f

With:
A is renamed to B, which is forbidden but creating B is fine

so:
- remote mkdir B
- remote move A/S -> B/S
- local restore A recursively -> ERROR, A/S/f can't be downloaded (it's been moved on the server)

src/libsync/discovery.cpp
ogoffart: Good point. Although that shouldn't be too complicated to recurse. We already do it when we delete a subtree.

Issue #7439, opened 2019-08-29
Syncing large accounts in VFS mode leads to "Not responding" in UI


HanaGemela:
OS: Windows 7, Windows 10
Client: 2.5.4, 2.6 daily
VFS on

Steps to recreate:
1) Add a new large account to your desktop client
2) Turn on VFS before initial sync
3) Run initial sync

Actual result: Client not responding, connection times out, client freezes 
Expected result: Account has been synced 

<img width="714" alt="Screenshot 2019-08-29 at 10 21 25" src="https://user-images.githubusercontent.com/49001702/63929058-0ebdeb00-ca51-11e9-82f8-22218ee64339.png">
<img width="709" alt="Screenshot 2019-08-29 at 10 30 53" src="https://user-images.githubusercontent.com/49001702/63929062-1087ae80-ca51-11e9-9107-dc9411afabb2.png">

guruz:
Also spotted by @lefherz

guruz:
I'm putting this into 2.6.0 in case it's an easy fix.

HanaGemela:
I can't rename the shared log file
In 20190828_1936_owncloud.log.6 "Not responding" happened around 11:21

guruz:
I don't see a timestamp stall there, only a lot of DB activity.

guruz:
@ogoffart @ckamm I wonder if `OwncloudPropagator::scheduleNextJob()` is good like that?
It will directly create job after job (with a 0 single shot timer), and one job in this initial sync case for the virtual files is created quite fast (incl DB write).

This is the 'loop':
```
08-29 11:21:00:096 [ debug sync.database ]      [ OCC::SyncJournalDb::commitInternal ]: Transaction commit  "download file start2" and starting new transaction
08-29 11:21:00:096 [ info sync.propagator ]:    Completed propagation of "test/1015066234.
wem.owncloud" by OCC::PropagateDownloadFile(0x15a56a98) with status 4
08-29 11:21:00:096 [ debug sync.statustracker ] [ OCC::SyncFileStatusTracker::slotItemCompleted ]:      Item completed "test/1015066234.wem.owncloud" 4 8
08-29 11:21:00:111 [ info sync.propagator ]:    Starting INSTRUCTION_NEW propagation of "test/101507196.wem.owncloud" by OCC::PropagateDownloadFile(0x15a56c00)
[...]

PR #7440, opened 2019-08-30
Propagator: Delay job execution a bit #7439

guruz:
Possible avoids "Not responding"

### DIFF
--- a/src/libsync/owncloudpropagator.cpp
+++ b/src/libsync/owncloudpropagator.cpp
@@ -594,7 +594,7 @@ void OwncloudPropagator::scheduleNextJob()
 {
     if (_jobScheduled) return; // don't schedule more than 1
     _jobScheduled = true;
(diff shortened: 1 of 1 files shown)


guruz:
@HanaGemela says it fixes the 'Do not responding' but it does not fix the real bug, so let's see if we really need it or not

guruz:
@HanaGemela reports that even with https://github.com/owncloud/client/pull/7442 she is still getting the 'Not responding'.
So we might need this PR(?), maybe with lower msec(?).

guruz:
https://superuser.com/questions/961843/how-does-windows-know-if-a-program-is-not-responding/961910

guruz:
@ckamm I changed the delay to 3 msec.
I think the bigger "Not responding" is things like in https://github.com/owncloud/enterprise/issues/3500 though

PR #7442, opened 2019-08-30
Propagator: Make sure we schedule only one job #7439

guruz:
To not starve the event loop.

(Idea by @ogoffart )

### DIFF
--- a/src/libsync/owncloudpropagator.cpp
+++ b/src/libsync/owncloudpropagator.cpp
@@ -491,6 +491,7 @@ void OwncloudPropagator::start(const SyncFileItemVector &items)
 
     connect(_rootJob.data(), &PropagatorJob::finished, this, &OwncloudPropagator::emitFinished);
 
+    _jobScheduled = false;
     scheduleNextJob();
 }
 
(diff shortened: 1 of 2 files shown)


src/libsync/owncloudpropagator.cpp
```diff
@@ -591,6 +592,8 @@ QString OwncloudPropagator::getFilePath(const QString &tmp_file_name) const
 
 void OwncloudPropagator::scheduleNextJob()
 {
+    if (m_jobScheduled>0) return; // don't schedule more than 1
```
ckamm: minor: would prefer " > "

src/libsync/owncloudpropagator.h
```diff
@@ -571,6 +572,7 @@ private slots:
     AccountPtr _account;
     QScopedPointer<PropagateDirectory> _rootJob;
     SyncOptions _syncOptions;
+    int m_jobScheduled;
```
ckamm: minor: would prefer = 0; here instead of in ctor
minor: we use only `_`, not `m_` in OCC

ckamm:
oh, maybe also set to 0 if it's ever negative, just for sanity

ogoffart:
Looks ok, but it need to be tested.
It could simply be a `bool` 
As for the negative, it should be an assert, perhaps

Issue #7441, opened 2019-08-30
Server directories can be deleted before renames happen


ckamm:
Have the following tree:
```
A/
A/f
```
pause the sync and perform these actions
```
local mkdir B/
local move A/f -> B/f
local remove A/
```
restart the sync.

It's possible that A/ will be removed on the remote before the remote rename from A/f to B/f happens. 

That's because the `PropagateDirectory` job for B/ (containing the remote mkdir, and the move) has `FullParallelism` and thus the delete of A/ can happen while the mkdir B/ is still running and before the move has started.

guruz:
Didn't @jnweiger report this some time ago?

ckamm:
If he did I wasn't aware of it. :(

HanaGemela:
@ckamm Do you know if I can reliably test it?

ckamm:
@HanaGemela It was reliable in an autotest. I haven't tried manual testing. Just check whether the setup above causes an error with current 2.6 builds.

HanaGemela:
@ckamm But if it works, can I be sure it is thanks to the fix or just this time rename happened before delete?

ckamm:
I meant that you could test with the current builds that don't have the patch yet to figure out whether testing is reliable.

jnweiger:
Is this related? https://github.com/owncloud/client/issues/7438

ckamm:
I don't think so. In this issue the client decides to do the right thing but shoots itself in the foot by executing in the wrong order.

HanaGemela:
Cannot recreate in 2.6.0rc1 (build 12411), macOS 10.14.6. Believed to be fixed

PR #7443, opened 2019-08-30
Propagation: Fix delete-before-rename bug

ckamm:
By introducing a PropagateRootDirectory job that explicitly
separates the directory deletion jobs from all the other jobs.

For #7441


ckamm:
The test is in testsyncmove.cpp - fails before, works now.

ckamm:
One of the permissions tests fails now, will need to investigate.

Issue #7445, opened 2019-09-01
SyncEngine::slotDiscoveryFinished() and SyncFileStatusTracker::slotAboutToPropagate block the main thread ("Not responding")


guruz:
We kind of have this already as an issue, but now @michaelstingl reported it in his video (see rocket chat).

Big loop of:
```
08-30 15:52:18:917 [ debug sync.statustracker ]	[ OCC::SyncFileStatusTracker::slotAboutToPropagate ]:	Investigating "Shared/owncloud/xxxx.pdf" OCC::SyncFileItem::NoStatus 8
08-30 15:52:18:918 [ warning sync.vfs ]:	fileStatusChanged: could not make handle for "C:/Users/mail/ownCloud/Shared/owncloud/xxxx.pdf"
```
Created following @HanaGemela 's https://github.com/owncloud/client/issues/7439

probably not fixable for 2.6.0

guruz:
Actually I think this is https://github.com/owncloud/client/issues/5356
Why @jnweiger only finds this now and not earlier I don't know..did 2.6 regress in that regard?

guruz:
Those are the functions:
https://github.com/owncloud/client/blob/v2.6.0-rc1/src/libsync/syncengine.cpp#L618
calls both
https://github.com/owncloud/client/blob/v2.6.0-rc1/src/libsync/syncfilestatustracker.cpp#L220
and
https://github.com/owncloud/client/blob/f7d4b29df93451e16969bd2167cd5f2c6c437269/src/libsync/owncloudpropagator.cpp#L368

guruz:
On my fast macbook with syncing down cloud.owncloud.com (in placeholder mode) I get:

```
09-06 13:03:36:108 [ info sync.engine ]:	#### Discovery end #################################################### Discovery Finished 425791 ms
09-06 13:03:36:917 [ info sync.engine ]:	#### Discovery end #################################################### Reconcile (sorted OK) 426600 ms
09-06 13:03:37:173 [ info sync.engine ]:	#### Discovery end #################################################### Reconcile (aboutToPropagate OK) 426856 ms
09-06 13:03:37:277 [ info sync.engine ]:	#### Post-Reconcile end #################################################### Post-Reconcile Finished  426959 ms
09-06 13:03:37:277 [ info gui.folder ]:	#### Propagation start ####################################################
[...]

PR #7457, opened 2019-09-06
Reconcile: Show more timings #7445

guruz:
To prove that we block the main thread too long and where.

### DIFF
--- a/src/libsync/syncengine.cpp
+++ b/src/libsync/syncengine.cpp
@@ -53,6 +53,7 @@
 #include <QProcess>
 #include <QElapsedTimer>
 #include <qtextcodec.h>
+#include <qtconcurrentrun.h>
 
 namespace OCC {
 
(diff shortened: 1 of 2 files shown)


src/libsync/syncengine.cpp
```diff
@@ -622,7 +623,18 @@ void SyncEngine::slotDiscoveryFinished()
         return;
     }
 
-    qCInfo(lcEngine) << "#### Discovery end #################################################### " << _stopWatch.addLapTime(QLatin1String("Discovery Finished")) << "ms";
+    qCInfo(lcEngine) << "#### Discovery end #################################################### Discovery Finished" << _stopWatch.addLapTime(QLatin1String("Discovery Finished")) << "ms" << QThread::currentThread();
+    QFuture<void> future = QtConcurrent::run([=]() {
```
ckamm: The only thing this captures is `this` right? That means this could crash if SyncEngine is deleted while the thread is running.

Capture a copy of `_syncItems`, sort that, and make that the future result.

The call to `invokeMethod` is also unsafe - again, `this` could be gone. Connect the future finished signal to the continuation slot.

guruz:
Take https://github.com/owncloud/client/pull/7461 instead

Issue #7456, opened 2019-09-05
infinite 'Connection timed out' errors during large upload


jnweiger:
Client 2.6.0rc1 with server 10.3.alpha

Empty server, client exited.
rm -rf syncfolder; mkdir syncfolder
rm -rf ~/.config/ownCloud\ Testpilot\ Edition

This is a very deep structure, 500000 files in ca 100 subdirectory levels:
`mv randfile_500k_105 syncfolder/`
-> start client using this syncfolder, connect to server

client grows to 3.158 GB rss during discovery
  #### Discovery end #### 211855 ms
  Connection timed out

This procedure repeats ca. every 5 Minutes. Nothing appears on the server.
Typical messages from the client log are like this:
```
09-04 18:10:43:572 [ info sync.engine ]:        #### Discovery end ####################################################  134185 ms
09-04 18:11:03:895 [ info sync.engine ]:        #### Post-Reconcile end ####################################################  154508 ms
09-04 18:11:04:031 [ info gui.folder ]: #### Propagation start ####################################################
09-04 18:11:04:031 [ info gui.application ]:    Sync state changed for folder  "http://localhost/owncloud/remote.php/dav/files/admin/" :  "Sync Running"
09-04 18:11:04:056 [ info sync.networkjob.propfind ]:   PROPFIND of QUrl("http://localhost/owncloud/remote.php/dav/files/admin/") FINISHED WITH STATUS "OperationCanceledError Connection timed out"
09-04 18:11:04:065 [ info gui.account.state ]:  AccountState connection status change:  OCC::ConnectionValidator::Connected -> OCC::ConnectionValidator::Timeout
09-04 18:11:04:066 [ info gui.account.state ]:  AccountState state change:  "Connected" -> "Network error"

```

Setting kernel TCP timeouts higher, or setting server php timeouts higher does not change anything. 
A local discovery that takes 90 seconds or more is very likely to result in a Timeout error. 

Also: During such a long local discovery, the client gui is unresponsive. It may take a minute for the tray icon to respond, also for the settings window to switch tab.

jnweiger:
Workaround:
Compile with the attached crude hack.
Now upload starts immediately after the first discovery. No errors in the log anymore.
[extended_timeouts_jw.patch.zip](https://github.com/owncloud/client/files/3581016/extended_timeouts_jw.patch.zip)

jnweiger:
Tested on windows 10:
[...]

PR #7458, opened 2019-09-06
ConnectionValidator: increase timeout to 57s

ckamm:
When the gui thread blocks for several seconds it's possible for the
ConnectionValidator to timeout and decide that the account is
unreachable. It will then terminate all sync runs.

Increasing the timeout makes this less likely to happen. The tradeoff is
that real disconnects will not be detected as quickly.

This does not address the root cause but makes the symptom less likely
to appear.

For #7456

### DIFF
--- a/src/gui/connectionvalidator.h
+++ b/src/gui/connectionvalidator.h
@@ -97,7 +97,7 @@ public:
     Q_ENUM(Status);
 
     // How often should the Application ask this object to check for the connection?
-    enum { DefaultCallingIntervalMsec = 32 * 1000 };
+    enum { DefaultCallingIntervalMsec = 62 * 1000 };
 
 public slots:
     /// Checks the server and the authentication.


src/gui/connectionvalidator.h
```diff
@@ -97,7 +97,7 @@ class ConnectionValidator : public QObject
     Q_ENUM(Status);
 
     // How often should the Application ask this object to check for the connection?
-    enum { DefaultCallingIntervalMsec = 32 * 1000 };
+    enum { DefaultCallingIntervalMsec = 62 * 1000 };
```
jnweiger: How does a change from 32 to 62 implement a 57 seconds timeout?
Some hidden magic eating 5secs?

ckamm: Yes, there's some magic that makes the timeout 5s less than the calling interval.

PR #7464, opened 2019-09-09
windows default is still 32bit.

jnweiger:
For windows, we are still at 32bits, afaik. Maybe it is better to let the adventurous people try that first before venturing into 64bit builds?


dschmidt:
I don't think any of this is accurate.

The ownCloud client is always built and shipped as 32bit executable. Only the explorer integration is shipped as 32bit and 64bit dlls alongside each other. 

mingw-w64 provides 32bit and 64bit compilers. There is no such thing as mingw-w32. There is mingw.org but I've never tested that because we already had mingw-w64.

dschmidt:
Dailies and next beta build are shipped as 32bit and 64bit MSIs. I think this is even less relevent now.
Can we close this?

dschmidt:
Actually I'm just closing it, reopen if you disagree.

PR #7469, opened 2019-09-11
Fix issues with threaded local discovery

ckamm:
Again thanks to @jnweiger for testing and stack traces.

### DIFF
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -1423,7 +1423,7 @@ DiscoverySingleDirectoryJob *ProcessDirectoryJob::startAsyncServerQuery()
 void ProcessDirectoryJob::startAsyncLocalQuery()
 {
     QString localPath = _discoveryData->_localDir + _currentFolder._local;
-    auto localJob = new DiscoverySingleLocalDirectoryJob(_discoveryData->_account, localPath, _discoveryData->_syncOptions._vfs.data(), this);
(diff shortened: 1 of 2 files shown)


guruz:
Note that this is only for the crashes, your other reported problems are probably not adressed yet.
(The Windows-only problems you reported)

jnweiger:
@guruz understood. I'll try to crash this the usual way.

jnweiger:
Testing on ubuntu 18.04
```
testpilotcloud version 2.6.0pr7469 (build 1471)
Git revision c454bfdf0ef71d0136c79c88c5e599863a09e996
Using Qt 5.12.4, built against Qt 5.12.4
Using 'OpenSSL 1.1.1  11 Sep 2018'
```

* synced empty folder with server
* mv 500k files in
* discovery starts, rss reaches ca 800 MB
* upload starts
* pause sync (GUI hangs for 8 sec, 100% cpu)
* resume sync, 2 seconds discovery, then upload.
* upload finishes within 30 sec. Only ca 1700 files were uploaded. Bad, but no crash!
* exit client
* restart client
* discovery starts, rss reaches ca 800 MB
* upload starts
* pause sync (GUI hangs for ca 10 sec, 100% CPU)
* exit client.
* restart client; 
* resume sync.
* discovery takes ca 1 minute, then upload starts
image

That looks great. It seems you fixed a crash. Thank you!

PR #7474, opened 2019-09-13
max_user_watches is also an isue on Mac

jnweiger:
@ChrisEdS please confirm if the solution here, also works for Mac.


ogoffart:
> macOS does not use inotify.

Therefore i believe we should not merge this PR and just close it.

ogoffart:
Closing as this is actually not correct.

PR #7477, opened 2019-09-16
Compile fix for compilers that don't like brace-init

ckamm:


ckamm:
Already covered by  #7476

Issue #7481, opened 2019-09-20
Weird disk/network condition can cause a reverse of the sync direction


jnweiger:
Server 10.3.0 alpha1, Client windows10 2.6.0daily20190920 msi

* Created a random file tree on the server with ca 500.000 files.
* start a client to sync down the files into a new (empty) local syncfolder.
* after discovery is done, and the first ca 5000 files were synced, switch to virtual files mode.
* while syncing, the network goes away and re-appears (suspended/resumed VM) -> provoking timeout and abort errors. 
* while syncing, we run out of disk space. The client suddenly exits (or crashed?)
* after restaring with sufficient free disk space and stable network, the client continues apparently without errors. 
![image](https://user-images.githubusercontent.com/1108546/65334415-a6cb7200-dbc2-11e9-8dd9-37c42fbe8f03.png)

Looking at the server, we see, that files are being moved into the user's trashcan.
Apparently the client at some point decided to change direction. Instead of bringing client and server in sync by downloading from the server, it now deletes from the server.

Sorry for the weird description. This may be hard to reproduce. 
I am reporting this, as it is a potential data loss. (Modulo restoring files on the server from the trashcan before expiry).

ckamm:
Do you have the log file?

The client should only delete files if they are stored to the local database. But if there are errors (the vfs switch, network errors, disk space errors), the directory etags should not be written to the database, avoiding this situation.

ogoffart:
I believe ckamm fixed this issue, right?

PR #7483, opened 2019-09-23
Propagate dir: Never write the etag on remote mkdir 

ckamm:
It must always only be written once all children are successfully
propagated.

For #7481


ogoffart:
I believe this could be tested

guruz:
^^ drone fails in AllFilesDeletedTest

ckamm:
Drone failure looks unrelated, retesting.

ckamm:
Also added a trivial test.

ckamm:
Now a different test fails on drone. All tests pass for me locally.

Issue #7468, opened 2019-09-11
craft windows builds fail in test suite


jnweiger:
There are long log file dump in the logs, but no apparent error. Still the testsuite fails in two cases:

The following tests FAILED:
	 27 - PermissionsTest (Failed)
	 29 - DatabaseErrorTest (Exit code 0xc0000374
)
Errors while running CTest
Action: test for owncloud/owncloud-client:master FAILED
*** Craft test failed: owncloud/owncloud-client after 2 minutes 7 seconds ***

Should we disable these, or is this fixable?

jnweiger:
Edited https://jenkins.int.owncloud.com/job/client-daily/
to say BUILD_TESTING=false
with testpilotcloud 2.6

jnweiger:
The testsuite seems to have random failures, sometimes more, somtimes less.
@TheOneRing just ran into https://ci.appveyor.com/project/ownclouders/client/builds/27495161/job/5rk2wxk214h2ygl3

with more failures:

 The following tests FAILED:
	 13 - SyncVirtualFilesTest (Failed)
	 20 - ZsyncTest (Failed)
	 27 - PermissionsTest (Failed)
	 29 - DatabaseErrorTest (Failed)
Errors while running CTest

guruz:
Indeed:
@ckamm wrote:
> The three tests pass for me with current 2.6 on windows 10

I saw 'The process cannot access the file because it is being used by another process.' in one of the tests, is it always the same one?
Maybe we need some artificial sleeps() again somewhere?

jnweiger:
All daily builds fail on this issue for quite a while. It is never exaclty the same test that fail, so fighting some race conditions is probably a good approach. Make this very long sleeps for a start, we can optimize them later :stuck_out_tongue:

TheOneRing:
The issue is not in the unittests, Appveyor provides a slow vm with a single core.
The tests fail as the checksum process has the tmp file still open while it should be renamed

michaelstingl:
@TheOneRing @jnweiger remove from 2.6.0 milestone? Drop AppVeyor?

guruz:
@TheOneRing not sure I understand your reasoning.

If the issue is VM slowness, then a sleep/wait-longer as suggested above will help :-)

If the issue is with the test that the test fails if there is a slow VM, then our code has a bug and we should check our use of threading in case of certain OS/CPU timings.

[...]

PR #7484, opened 2019-09-23
Checksums: Explicitly close file before reporting result

ckamm:
To ensure it's no longer open when the finished signal fires.

For #7468


guruz:
Nice!
Let's wait for AppVeyor to succeed @TheOneRing

Issue #7371, opened 2019-08-09
[Windows] [Linux] 2.6.0beta1 (build 12201) the client only show a white (or black) window at second opening of tray icon


DangerSchwob:
Hi,

Client Version: 2.6.0beta1 (build 12201)
Server Version: 10.2.1.4
OS Version: Microsoft Windows 10 Pro - 10.0.18362 Build 18362

Issue: When you open the ownCloud client window via tray icon, the first time all working fine, but the second time and later you get only a white window

![grafik](https://user-images.githubusercontent.com/22342104/62759153-d4c87d00-ba80-11e9-9618-ebafa62018b9.png)

Resizing the window or click somewhere in the navbar, where a navigation button is, bring everything back, but if you close and open the window, you see the same issue again.

If you find in the white window the "account" button, you see also all again, but when the menu closing, the whole client crashing. CrashID: bf541872-6a27-4ea9-a8cb-910da4f9ff2f

![grafik](https://user-images.githubusercontent.com/22342104/62759561-efe7bc80-ba81-11e9-97d9-ea2f19cdaf0c.png)

jnweiger:
Similar, but black instead of white:
2.6.0daily 20190916 on debian 8

 - connect, then close the settings dialog.
 - run the client again to trigger reopening the settings dialog. 
 - It appears all black,
 - Workaround: click the mouse near the top left corner of the black area. This causes a refresh.

guruz:
@HanaGemela @jnweiger Let's try to achieve this for 2.6.0, it sounds rather annoying?

TheOneRing:
I'll do

guruz:
> If you find in the white window the "account" button, you see also all again, but when the menu closing, the whole client crashing. CrashID: bf541872-6a27-4ea9-a8cb-910da4f9ff2f

Sounds also very Qt-ish..:

```
EXCEPTION_ACCESS_VIOLATION_READ

Fatal Error: EXCEPTION_ACCESS_VIOLATION_READ



    Qt5Widgets0x6a3e5c50QWidget::windowHandle qwidget.cpp:2660

Qt5Widgets0x6a3f6fe6QWidgetWindow::handleMouseEvent qwidgetwindow.cpp:591
Qt5Widgets0x6a3f5446QWidgetWindow::event qwidgetwindow.cpp:281
Qt5Widgets0x6a3b4889QApplicationPrivate::notify_helper qapplication.cpp:3737
[...]

Issue #7043, opened 2019-02-15
[macOS] Several glitches when switching dark/light mode


islamadel:
I use the latest version of ownCloud Client (2.5.3) on MacOS Mojave (10.14.3) Dark Mode activated
Under Settings Folder Names are blank
### Expected behaviour
Folder Names should be displayed

### Actual behaviour
Folder Names blank, can't choose what to sync

### Steps to reproduce
1. Just Install and enable dark mode
2. 
3. 

### Server configuration
Operating system:

Web server: 

Database:

PHP version:

ownCloud version:

Storage backend (external storage):

### Client configuration
Client version:

Operating system:

OS language:

Qt version used by client package (Linux only, see also Settings dialog):

Client package (From ownCloud or distro) (Linux only):

Installation path of client:

### Logs

Please use Gist (https://gist.github.com/) or a similar code paster for longer
logs.

```Template for output < 10 lines```

1. Client logfile: Output of `owncloud --logwindow` or `owncloud --logfile log.txt`
(On Windows using `cmd.exe`, you might need to first `cd` into the ownCloud directory)
(See also http://doc.owncloud.org/desktop/2.2/troubleshooting.html#client-logfile )

2. Web server error log:

3. Server logfile: ownCloud log (data/owncloud.log):

See Screenshot:

<img width="794" alt="2019-02-15 20 06 51" src="https://user-images.githubusercontent.com/1722747/52875563-491f0500-315d-11e9-9358-5bf28241cc68.png">

michaelstingl:
The team is working on a 2.5.4 release with Qt 5.12.x. You can already download a pre-release build here:
 https://download.owncloud.com/desktop/daily/?C=M;O=D

Could you test and provide feedback?

guruz:
@islamadel to be exact please try https://download.owncloud.com/desktop/daily/ownCloud-2.5.4.11337-daily20190219.pkg

lazawan:
@guruz @michaelstingl 
I used the build from guruz 
https://download.owncloud.com/desktop/daily/ownCloud-2.5.4.11337-daily20190219.pkg

it looks so right now : 

![screen shot 2019-02-19 at 11 48 39](https://user-images.githubusercontent.com/44814084/53010058-2894ca80-343d-11e9-8a53-e884aa277174.png)

![screen shot 2019-02-19 at 11 50 10](https://user-images.githubusercontent.com/44814084/53010075-321e3280-343d-11e9-8b9c-008c95b90c8c.png)

[...]

PR #7492, opened 2019-09-25
No MacSettingsDialog (WiP) #7371

guruz:
We could have this for 2.6.0 and fix the issue mentioned in https://github.com/owncloud/client/issues/7043#issuecomment-534757893 and https://github.com/owncloud/client/issues/7371#issuecomment-534888737 for 2.6.1 or so :/ 

Bildschirmfoto 2019-09-25 um 10 15 43

Maybe the elliding is a bit too agressive?

FYI @ChrisEdS @TheOneRing @dschmidt


guruz:
@ChrisEdS i only tested High Sierra, so maybe you want to make a build and check out Dark Mode with it... I'll update the PR for that

ChrisEdS:
> @ChrisEdS i only tested High Sierra, so maybe you want to make a build and check out Dark Mode with it... I'll update the PR for that

Yes, I will test that on Mojave.

TheOneRing:
Hm is eliding in the middle a mac thing? I always prefer it at the end.

guruz:
> Hm is eliding in the middle a mac thing? I always prefer it at the end.

I think we do it on purpose to show both the username and the server name in case you are hanna@ on multiple servers

(Check the source..)

TheOneRing:
> > Hm is eliding in the middle a mac thing? I always prefer it at the end.
> 
> I think we do it on purpose to show both the username and the server name in case you are hanna@ on multiple servers
> 
> (Check the source..)

Makes sense.
With the space available it looks reasonable

guruz:
I think we should merge this if it fixes the UI glitches. So maybe it's not a WiP at all?

guruz:
continuous-integration/drone/pr — Build is passing 

@ckamm @ogoffart merge?

guruz:
Great, failure only in SelectiveSyncTest but that might be fixed in another PR

Issue #6722, opened 2018-08-16
[2.5.0] server URL without <Enter> ==> Invalid URL


michaelstingl:
### Expected behaviour
Clicking the [Next] button after entering the server url (without leading `https://`) should complete the URL automagically. (same behaviour as with <Enter> key today)

### Actual behaviour
Clicking the [Next] button after entering the server url (without leading `https://`) results in an "Invalid URL" error message. Only when confirming with <Enter> key, the leading `https://` gets added to the server URL automagically.

### Steps to reproduce
1. enter server url without leading `https://`
2. click [Next] button

![screenflow](https://user-images.githubusercontent.com/214010/44217585-1711f380-a178-11e8-9a91-fcfd18c82359.gif)

![blank_skitch_document](https://user-images.githubusercontent.com/214010/44217626-2e50e100-a178-11e8-88af-5842e54d2054.png)

![about_this_mac](https://user-images.githubusercontent.com/214010/44217669-4d4f7300-a178-11e8-8726-9dfa1494d591.png)

ogoffart:
I cannot reproduce this problem.  The 'next' button auto-complete the field.
Maybe something mac-specific?

michaelstingl:
Yeah, macOS only (tested with High Sierra & Mojave). Didn't happen with `ownCloud-qt5.6.2-2.4.3.10347.pkg`.

PR #7493, opened 2019-09-25
Ensure the url is complete before we validate it

TheOneRing:
Fixes: #6722


TheOneRing:
@HanaGemela @michaelstingl the code is not tested but it should ensure a valid url schema

michaelstingl:
> @HanaGemela @michaelstingl the code is not tested but it should ensure a valid url schema

yes, solves my problem 👍

PR #7501, opened 2019-10-07
Investigation around drone test failures

ckamm:
Turns out the drone failures are because of seemingly random errno 11 failures during readdir. These changes should provide better diagnostics and make the source of the failure clearer.

### DIFF
--- a/src/libsync/discoveryphase.cpp
+++ b/src/libsync/discoveryphase.cpp
@@ -247,9 +247,13 @@ void DiscoverySingleLocalDirectoryJob::run() {
         emit finishedFatalError(errorString);
         return;
     }
-    errno = 0;
+
     QVector<LocalInfo> results;
-    while (auto dirent = csync_vio_local_readdir(dh, _vfs)) {
+    while (true) {
+        errno = 0;
+        auto dirent = csync_vio_local_readdir(dh, _vfs);
+        if (!dirent)
+            break;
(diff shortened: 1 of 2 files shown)


src/libsync/discoveryphase.cpp
```diff
@@ -247,9 +247,13 @@ void DiscoverySingleLocalDirectoryJob::run() {
...
         return;
     }
-    errno = 0;
+
     QVector<LocalInfo> results;
-    while (auto dirent = csync_vio_local_readdir(dh, _vfs)) {
+    while (true) {
+        errno = 0;
+        auto dirent = csync_vio_local_readdir(dh, _vfs);
+        if (!dirent)
```
ogoffart: nitpick: We actually should check the errno here.  This only works now because there is only one break.

test/syncenginetestutils.h
```diff
@@ -1204,7 +1204,7 @@ class FakeFolder
         // A new folder will update the local file state database on first sync.
         // To have a state matching what users will encounter, we have to a sync
         // using an identical local/remote file tree first.
-        syncOnce();
+        Q_ASSERT(syncOnce());
```
ogoffart: Actually, i think that since we build in release mode in some configuration this does not fire.

(We should probably replace all the Q_ASSERT in this file with ENFORCE or something)

PR #7504, opened 2019-10-07
Revert "Windows: Use better name for SocketAPI socket #6983"

guruz:
This reverts commit ed1e534d24f871a0ab6ea02e4c88eefe3126bf59.

Doesn't make sense to spend effort on this, we keep it as it is.


guruz:
Thanks for review!

PR #7509, opened 2019-10-09
Support for openid connect

ogoffart:


ogoffart:
From drone:

> /drone/src/src/gui/creds/oauth.cpp:26:10: fatal error: QRandomGenerator: No such file or directory

QRandomGenerator was added in Qt 5.10.  I don't really want to add a fallback with the insecure `qrand`
So i'd like to make Qt 5.12 a minimum required Qt version for the client 2.7

michaelstingl:
> So i'd like to make Qt 5.12 a minimum required Qt version for the client 2.7

Okay, works for me 👍

ogoffart:
Make Qt 5.12 minimum requirement: https://github.com/owncloud/client/pull/7521
Update to CI: https://github.com/owncloud-ci/client/pull/1

mmattel:
@TheOneRing mind review/approving?

src/gui/creds/oauth.cpp
```diff
@@ -91,10 +101,9 @@ void OAuth::start()
 
                 auto requestBody = new QBuffer;
                 QUrlQuery arguments(QString(
-                    "grant_type=authorization_code&code=%1&redirect_uri=http://localhost:%2")
-                                        .arg(code, QString::number(_server.serverPort())));
+                    "grant_type=authorization_code&code=%1&redirect_uri=http://localhost:%2&code_verifier=%3&scope=openid offline_access")
```
guruz: Is it safe to construct a query string like that? (yes i see the FullyEncoded below but still) (In code further further below you use proper setQueryItems)

src/gui/creds/oauth.cpp
```diff
@@ -110,7 +119,9 @@ void OAuth::start()
                         || jsonData.isEmpty() || json.isEmpty() || refreshToken.isEmpty() || accessToken.isEmpty()
                         || json["token_type"].toString() != QLatin1String("Bearer")) {
                         QString errorReason;
-                        QString errorFromJson = json["error"].toString();
+                        QString errorFromJson = json["error_description"].toString();
+                        if (errorFromJson.isEmpty())
+                            QString errorFromJson = json["error"].toString();
```
guruz: BUG! two times "QStrings errorFromJson" in different scope.

src/gui/creds/oauth.cpp
```diff
@@ -129,59 +140,130 @@ void OAuth::start()
...
+
+        emit this->authorisationLinkChanged(authorisationLink());
+    });
+}
+
+void OAuth::openBrowser()
+{
+    authorisationLinkAsync([this](const QUrl &link) {
+        if (!QDesktopServices::openUrl(link)) {
+            // We cannot open the browser, then we claim we don't support OAuth.
```
guruz: if this is realistic, maybe it should show a qWarning?

src/gui/creds/oauth.h
```diff
@@ -27,15 +27,20 @@ namespace OCC {
  *
  *   --> start()
  *       |
- *       +----> openBrowser() open the browser to the login page, redirects to http://localhost:xxx
+ *       +----> fetchWellKnown() query the ".well-known/openid-configuration" endpoint
```
guruz: Please also document what happens if this does does not exist and how this is with regards to OpenID vs normal OAuth?

src/libsync/creds/httpcredentials.cpp
```diff
@@ -412,47 +412,64 @@ bool HttpCredentials::refreshAccessToken()
...
+        req.setHeader(QNetworkRequest::ContentTypeHeader, "application/x-www-form-urlencoded");
+
+        QString basicAuth = QString("%1:%2").arg(
+            Theme::instance()->oauthClientId(), Theme::instance()->oauthClientSecret());
+        req.setRawHeader("Authorization", "Basic " + basicAuth.toUtf8().toBase64());
+        req.setAttribute(HttpCredentials::DontAddCredentialsAttribute, true);
+
+        auto requestBody = new QBuffer;
+        QUrlQuery arguments(QString("grant_type=refresh_token&refresh_token=%1").arg(_refreshToken));
+        requestBody->setData(arguments.query(QUrl::FullyEncoded).toLatin1());
```
guruz: Same as comment above, maybe it's cleaner to use setQueryItems

src/libsync/creds/httpcredentials.cpp
```diff
@@ -412,47 +412,64 @@ bool HttpCredentials::refreshAccessToken()
...
+        auto job = _account->sendRequest("POST", requestTokenUrl, req, requestBody);
+        job->setTimeout(qMin(30 * 1000ll, job->timeoutMsec()));
+        QObject::connect(job, &SimpleNetworkJob::finishedSignal, this, [this](QNetworkReply *reply) {
+            auto jsonData = reply->readAll();
+            QJsonParseError jsonParseError;
+            QJsonObject json = QJsonDocument::fromJson(jsonData, &jsonParseError).object();
+            QString accessToken = json["access_token"].toString();
+            if (jsonParseError.error != QJsonParseError::NoError || json.isEmpty()) {
+                // Invalid or empty JSON: Network error maybe?
+                qCWarning(lcHttpCredentials) << "Error while refreshing the token" << reply->errorString() << jsonData << jsonParseError.errorString();
```
guruz: Is there any error state you need to set in some variable or return value?

src/gui/creds/oauth.cpp
ogoffart: Yeah, this is simple enough to be safe.

src/gui/creds/oauth.cpp
ogoffart: well spotted.

src/libsync/creds/httpcredentials.cpp
ogoffart: The idea is that the jobs get re-tried and will then process to the normal authentication method.
Yeah, maybe something better could be done, but we really don't want to disturb the user in this case anyway. This is not new code, it just got one more level of identation.

guruz:
@michaelstingl @HanaGemela FYI

Issue #7512, opened 2019-10-10
Too low contrast when tab is selected


HanaGemela:
Client: 2.6.0-daily20191009 (build 12541)
macOS: 10.15

When a tab is selected, contrast is too low. 

<img width="860" alt="Screenshot 2019-10-10 at 10 59 26" src="https://user-images.githubusercontent.com/49001702/66554330-17740780-eb4d-11e9-810b-35569a50fa77.png">

ogoffart:
PR: https://github.com/owncloud/client/pull/7513

HanaGemela:
Works as expected in 2.6.1-daily20191127 (build 12791), macOS 10.15

PR #7513, opened 2019-10-10
Fix too low contrast when tab is selected

ogoffart:
Issue: #7512

(I only tested on linux where it makes no difference. But that should solve the problem described in #7512)


guruz:
Looks better!

Bildschirmfoto 2019-10-11 um 12 57 21

Issue #7403, opened 2019-08-16
Change the color of VFS box so the account doesn't appear online and synced 


HanaGemela:
'Virtual file support is enabled.' text is displayed in a green box. That makes the account look connected and synced. So even though the user is signed out, their account is green which evokes online state. 

<img width="790" alt="Screenshot 2019-08-16 at 14 59 04" src="https://user-images.githubusercontent.com/49001702/63170093-cfe36a80-c038-11e9-837d-c2d7be46556d.png">

ogoffart:
What color do you suggest?

HanaGemela:
Blue or grey to match the rest of the dialog

ogoffart:
How about  this:  (this is what one get if we invert the green and blue component in the color)

![image](https://user-images.githubusercontent.com/959326/66814874-98f5dc00-ef37-11e9-98f2-ee8370c9b74d.png)

HanaGemela:
Nice

PR #7526, opened 2019-10-15
Account Settings: change the color of info message from green to blue

ogoffart:
To avoid confusion with the color of "success"

Issue #7403


ogoffart:
image

Issue #7018, opened 2019-02-06
Starting client while already running sometimes doesn't show settings window


lazawan:
steps to reproduce : 
after client-configuration and client runs properly : 
- go to  tray-icon in taskbar and quit owncloud
- doppel-click on owncloud-icon on Desktop >> nothing opens just owncloud-client is started in background and tray-icon in taskbar appears again
- again doppel-click >> nothing happens
- again doppel-click >> now owncloud-window opens 

regards

windows 10 client 2.5.3rc1

ckamm:
It's intentional that starting the client usually doesn't lead to a window popping up. We don't want the settings window to be shown every windows startup.

What's odd is that the second time you start the client nothing happens. The intended behavior is that starting the client while it's already running will lead to the settings window popping up.

guruz:
Maybe some windows popup protection?
Moving to 2.6.1 though

ogoffart:
There is a 10 second delay from startup for which we won't show the settings:

https://github.com/owncloud/client/blob/e82dc141f7696b208102324750a052f55a5e7ccf/src/gui/application.cpp#L523-L529


The reason is that some auto-start script tends to start the application twice :-/

guruz:
> session restore and auto start

That's a linux thing though?
Or I heard that Windows might do this too?

In general, our auto start setting could in theory launch the client with something like  `--from-autostart` and then adjust behaviour somehow?
But maybe this is all not worth the trouble.

HanaGemela:
@TheOneRing is this issue fixed? There isn't `ReadyToTest` label
If it is fixed, could you specify the expected result here? When should settings dialog open?

TheOneRing:
I don't think so it would probably require changes to the installer to invoke the client with a -s

TheOneRing:
Adopt desktop files and windows shortcuts to pass -s to the client.

jnweiger:
PR #7784 fixes this issue for me.
It nicely implements the general UX-rule that every user interaction should have an immediate reaction.
Regarding Windows and Mac: do we do the same there?

HanaGemela:
What is the expected results for macOS? 

Windows - setting window opens on every double click
macOS - setting window never opens on double click
Linux  - setting window opens on every double click

michaelstingl:
> macOS - setting window never opens on double click

Okay for now 👍

PR #7528, opened 2019-10-15
Application: add an option to start with the settings visible

ogoffart:
Relates to #7018

### DIFF
--- a/docs/modules/ROOT/pages/advanced_usage/options.adoc
+++ b/docs/modules/ROOT/pages/advanced_usage/options.adoc
@@ -10,6 +10,8 @@ The other options are:
 [options="header"]
 |===
 | Option | Description
+| `-s` +
+`--showsettings` | Open the settings dialog while starting.
 | `--logwindow` | Opens a window displaying log output.
 | `--logfile <filename>` | Write log output to the file specified. To write to stdout, specify `-` as the filename.
 | `--logdir <name>` | Writes each synchronization log output in a new file in the specified directory.
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -68,6 +68,7 @@ namespace {
     static const char optionsC[] =
         "Options:\n"
         "  -h --help            : show this help screen.\n"
+        "  -s --showsettings    : show the settings dialog while starting.\n"
         "  --logwindow          : open a window to show log output.\n"
         "  --logfile <filename> : write log output to file <filename>.\n"
         "  --logfile -          : write log output to stdout.\n"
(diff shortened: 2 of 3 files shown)


docs/modules/ROOT/pages/advanced_usage/options.adoc
```diff
@@ -10,6 +10,7 @@ The other options are:
 [options="header"]
 |===
 | Option | Description
+| `--showsettings` | Open the settings dialog while starting.
```
mmattel: Why do you only show the long form in help but dont show the short form `-s` ?
There are many examples in eg. occ commands in docs where we show both, just in the way it has been designed.

ogoffart: Yes you are right, i should add the `-s`  in the docs

docs/modules/ROOT/pages/advanced_usage/options.adoc
```diff
@@ -10,6 +10,7 @@ The other options are:
 [options="header"]
 |===
 | Option | Description
+| `--showsettings` or `-s` | Open the settings dialog while starting.
```
mmattel: Only a detail: we never use `or` when having a short and long option.
We would write:
|  `-s` +
`--showsettings` | Open the settings dialog while starting

See https://doc.owncloud.com/server/10.3/admin_manual/configuration/server/occ_command.html#background-jobs-selector
Executing a background job

mmattel: I know you will not like it, but you have to make a new line after the plus symbol, exactly as written. This is due to the way how ascidoc works. The `+` sign at the end of a line shows, that the next line is part of that block but a new line (like you have in MS Word shift-return making a new line in the same paragraph while a return makes a new line with a new paragraph). Having that, the table renders correctly.

ogoffart: Fixed, thanks for explaining me. (I thought the + was meant as a literal +)

mmattel: Approved from the docs pov 😄 
Marking as resolved

TheOneRing:
Looks good but I'm  not 100% sure about the use case.
Should the desktop entry (shortcut/desktop file) pass -s?
I mean especially on gnome where we no longer have a system tray the settings window should show when we start from the desktop file.

Btw: should we create a low lvl ticket to port to QCommandLineParser

ogoffart:
> Should the desktop entry (shortcut/desktop file) pass -s?

Yes, i'd say so.

> Btw: should we create a low lvl ticket to port to QCommandLineParser

Not sure how good QCommandLineParser is when we pass the option accross process if the application is already running.

TheOneRing:
> > Should the desktop entry (shortcut/desktop file) pass -s?
> 
> Yes, i'd say so.
> 
> > Btw: should we create a low lvl ticket to port to QCommandLineParser
> 
> Not sure how good QCommandLineParser is when we pass the option accross process if the application is already running.
Should be possible https://doc.qt.io/qt-5/qcommandlineparser.html#parse

PR #7531, opened 2019-10-16
Port drone config to starlark

dschmidt:


guruz:
@tboerger @TheOneRing review? :)

PR #7534, opened 2019-10-18
Start using extra cmake modules

TheOneRing:


CMakeLists.txt
```diff
@@ -1,7 +1,24 @@
 cmake_minimum_required(VERSION 3.1)
 set(CMAKE_CXX_STANDARD 14)
 
-project(client)
+project(client LANGUAGES CXX VERSION 2.7)
+include(FeatureSummary)
+find_package(ECM 5.50.0  NO_MODULE)
```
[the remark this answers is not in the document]
ogoffart: KF 5.50 is from sept 2018.     Given that we will require Qt 5.12 which is from dec 2018, i guess that's fine.

CMakeLists.txt
```diff
@@ -1,7 +1,24 @@
 cmake_minimum_required(VERSION 3.1)
 set(CMAKE_CXX_STANDARD 14)
 
-project(client)
+project(client LANGUAGES CXX VERSION 2.7)
```
ogoffart: does that mean we need to maintain this number as well as in VERSION.cmake?

ogoffart:
Needs https://github.com/owncloud-ci/client/pull/1

CMakeLists.txt
TheOneRing: good point I changed the version.cmake.
In the long run we might wan't do drop the MIRALL variables.
I hope that no script parse that cmake file to determine the version... @dschmidt ?

src/cmd/CMakeLists.txt
```diff
@@ -1,5 +1,4 @@
 project(cmd)
-set(CMAKE_AUTOMOC TRUE)
```
dschmidt: <3

VERSION.cmake
```diff
@@ -1,6 +1,6 @@
-set( MIRALL_VERSION_MAJOR 2 )
-set( MIRALL_VERSION_MINOR 7 )
-set( MIRALL_VERSION_PATCH 0 )
+set( MIRALL_VERSION_MAJOR ${PROJECT_VERSION_MAJOR} )
+set( MIRALL_VERSION_MINOR ${PROJECT_VERSION_MINOR} )
+set( MIRALL_VERSION_PATCH ${PROJECT_VERSION_PATCH} )
```
dschmidt: Nice, didn't know this.

src/csync/CMakeLists.txt
```diff
@@ -106,8 +98,6 @@ set_target_properties(
         ${LIBRARY_VERSION}
       SOVERSION
         ${LIBRARY_SOVERSION}
-      RUNTIME_OUTPUT_DIRECTORY
-        ${BIN_OUTPUT_DIRECTORY}
```
dschmidt: Personally I don't care about this, because I always install also during development but some people might be annoyed because this changes the way executables are run from the build directory.

src/gui/CMakeLists.txt
```diff
@@ -1,8 +1,6 @@
 project(gui)
 find_package(Qt5 REQUIRED COMPONENTS Widgets)
-set(CMAKE_AUTOMOC TRUE)
 set(CMAKE_AUTOUIC TRUE)
```
dschmidt: Why do we still need this? Should we set this per target while at it?

src/gui/CMakeLists.txt
```diff
@@ -258,10 +256,6 @@ add_library(updater STATIC ${updater_SRCS})
 target_link_libraries(updater ${synclib_NAME} ${updater_DEPS} Qt5::Widgets Qt5::Network Qt5::Xml)
 target_include_directories(updater PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
 
-set_target_properties( ${APPLICATION_EXECUTABLE} PROPERTIES
-        RUNTIME_OUTPUT_DIRECTORY  ${BIN_OUTPUT_DIRECTORY}
-)
```
dschmidt: See comment above regarding `RUNTIME_OUTPUT_DIRECTORY`.

src/gui/CMakeLists.txt
```diff
@@ -231,13 +229,13 @@ if(NOT BUILD_OWNCLOUD_OSX_BUNDLE)
     # we may not add MACOSX_BUNDLE here, if not building one
```
dschmidt: Building a bundle is the only supported option for a long long time, maybe we should get rid of `BUILD_OWNCLOUD_OSX_BUNDLE` and simplify this code a bit?
And also adjust comments?

src/libsync/CMakeLists.txt
```diff
@@ -153,7 +151,6 @@ target_include_directories(${synclib_NAME} PUBLIC ${CMAKE_CURRENT_SOURCE_DIR} ${
 set_target_properties( ${synclib_NAME}  PROPERTIES
 	VERSION ${MIRALL_VERSION}
 	SOVERSION ${MIRALL_SOVERSION}
-	RUNTIME_OUTPUT_DIRECTORY ${BIN_OUTPUT_DIRECTORY}
```
dschmidt: See above.

test/owncloud_add_test.cmake
```diff
@@ -1,12 +1,11 @@
 find_package(Qt5 COMPONENTS Core Test Xml Network REQUIRED)
 
 macro(owncloud_add_test test_class additional_cpp)
-    set(CMAKE_AUTOMOC TRUE)
     set(OWNCLOUD_TEST_CLASS ${test_class})
     string(TOLOWER "${OWNCLOUD_TEST_CLASS}" OWNCLOUD_TEST_CLASS_LOWERCASE)
 
     add_executable(${OWNCLOUD_TEST_CLASS}Test test${OWNCLOUD_TEST_CLASS_LOWERCASE}.cpp ${additional_cpp})
-    set_target_properties(${OWNCLOUD_TEST_CLASS}Test PROPERTIES RUNTIME_OUTPUT_DIRECTORY ${BIN_OUTPUT_DIRECTORY})
```
dschmidt: See above.

test/owncloud_add_test.cmake
```diff
@@ -24,12 +23,11 @@ macro(owncloud_add_test test_class additional_cpp)
 endmacro()
 
 macro(owncloud_add_benchmark test_class additional_cpp)
-    set(CMAKE_AUTOMOC TRUE)
     set(OWNCLOUD_TEST_CLASS ${test_class})
     string(TOLOWER "${OWNCLOUD_TEST_CLASS}" OWNCLOUD_TEST_CLASS_LOWERCASE)
 
     add_executable(${OWNCLOUD_TEST_CLASS}Bench benchmarks/bench${OWNCLOUD_TEST_CLASS_LOWERCASE}.cpp ${additional_cpp})
-    set_target_properties(${OWNCLOUD_TEST_CLASS}Bench PROPERTIES RUNTIME_OUTPUT_DIRECTORY ${BIN_OUTPUT_DIRECTORY})
+    ecm_mark_nongui_executable(${OWNCLOUD_TEST_CLASS}Bench)
```
dschmidt: See above.

src/gui/CMakeLists.txt
TheOneRing: ecm sets it globaly for some platforms, for other platforms it should not be needed but I'll check

src/gui/CMakeLists.txt
TheOneRing: I think I'll have to rework some stuff anyhow..... ( mac is currently broken on this branch)

src/csync/CMakeLists.txt
ogoffart: I believe you can't run the client without installing it anyway because it needs to access some files from the installed directory. (or maybe this was a csync problem and it is no longer the case)

But I suppose this may also change the location of the tests. But in a way, it is better to have the tests in their directory.

ogoffart:
I allowed myself to rebase the branch on top of master to re-run drone now that we upgraded the CI

CMakeLists.txt
dschmidt: actually ... probably yes. That was the reason to move everything to VERSION.cmake.

ogoffart:
Drone CI fails because it no longer find SQLite.

TheOneRing:
> Drone CI fails because it no longer find SQLite.

And mac is probably broken, this needs some more work on my side before we can really us it

dschmidt:
Btw drone fails because it can't find an sqlite find/config module. You removed this in this PR, why and how is it supposed to work now?

src/libsync/vfs/suffix/CMakeLists.txt
```diff
@@ -1,34 +1,9 @@
...
-    set(vfs_installdir "${LIB_INSTALL_DIR}/../PlugIns")
-else()
-    set(vfs_installdir "${PLUGINDIR}")
-endif()
-
-INSTALL(TARGETS "${synclib_NAME}_vfs_suffix"
-  LIBRARY DESTINATION "${vfs_installdir}"
-  RUNTIME DESTINATION "${vfs_installdir}"
-)
+INSTALL(TARGETS "${synclib_NAME}_vfs_suffix" DESTINATION ${QT_PLUGIN_INSTALL_DIR})
```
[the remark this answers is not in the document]
TheOneRing: Yes it is the global plugin dir, but both on Windows and on mac we move the files to that dir anyhow, when deploying.
We could also use KDE_INSTALL_PLUGINDIR which can be a separate directory or the qt system plugin dir depending on a cmake flag.

src/gui/MacOSXBundleInfo.plist
```diff
@@ -56,14 +56,14 @@
...
         <string>Editor</string>
         <key>LSHandlerRank</key>
         <string>Owner</string>
         <key>LSItemContentTypes</key>
         <array>
-            <string>@APPLICATION_REV_DOMAIN@.VIRTUALFILE</string>
+            <string>${APPLICATION_REV_DOMAIN}.VIRTUALFILE</string>
         </array>
     </dict>
 </array>
```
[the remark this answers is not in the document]
TheOneRing: CMake now directly handles the file now so no need for a configure step

src/cmd/CMakeLists.txt
```diff
@@ -9,28 +6,17 @@ set(cmd_SRC
...
-elseif(NOT BUILD_LIBRARIES_ONLY)
-    install(TARGETS ${cmd_NAME}
-	    RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
-	    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
-	    ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
-endif()
+install(TARGETS ${cmd_NAME}
+        RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
+        LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+        ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
```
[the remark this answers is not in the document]
TheOneRing: because, thank you for spotting this one

CMakeLists.txt
```diff
@@ -1,24 +1,45 @@
...
+if (APPLE OR WIN32)
+    set(USE_OUR_OWN_SQLITE3 TRUE)
+    set(SQLITE3_INCLUDE_DIR ${CMAKE_SOURCE_DIR}/src/3rdparty/sqlite3)
+    set(SQLITE3_LIBRARIES "")
+    set(SQLITE3_SOURCE ${SQLITE3_INCLUDE_DIR}/sqlite3.c)
+    MESSAGE(STATUS "Using own sqlite3 from " ${SQLITE3_INCLUDE_DIR})
+else()
+    find_package(SQLite3 3.9.0 REQUIRED)
+endif()
+
```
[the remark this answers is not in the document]
TheOneRing: We had that before for some packages but it is really irritating when a find_package which looks like a standard cmake find_package behaves really really strang...

CMakeLists.txt
```diff
@@ -1,24 +1,45 @@
...
+set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules ${ECM_MODULE_PATH})
+
+include(KDEInstallDirs)
+include(KDECMakeSettings)
+include(ECMMarkNonGuiExecutable)
+
+include(KDECompilerSettings NO_POLICY_SCOPE)
+
+set(POSITION_INDEPENDENT_CODE ON)
+set(CMAKE_AUTOUIC ON)
```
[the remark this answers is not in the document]
TheOneRing: It looks to me like our code behaves quite well with this setting so I didn't see a reason to try it on a target base.

CMakeLists.txt
```diff
@@ -1,24 +1,45 @@
...
 set(CMAKE_CXX_STANDARD 14)
 
-project(client)
+project(client LANGUAGES CXX C VERSION 2.7.0)
+include(FeatureSummary)
+find_package(ECM 5.50.0  NO_MODULE)
+set_package_properties(ECM PROPERTIES TYPE REQUIRED DESCRIPTION "Extra CMake Modules." URL "https://projects.kde.org/projects/kdesupport/extra-cmake-modules")
+feature_summary(WHAT REQUIRED_PACKAGES_NOT_FOUND FATAL_ON_MISSING_REQUIRED_PACKAGES)
+
+set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules ${ECM_MODULE_PATH})
```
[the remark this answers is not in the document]
TheOneRing: Prepend is rather new and this is how it was done always since the beginning of time :D

CMakeLists.txt
```diff
@@ -1,24 +1,45 @@
 cmake_minimum_required(VERSION 3.1)
 set(CMAKE_CXX_STANDARD 14)
 
-project(client)
+project(client LANGUAGES CXX C VERSION 2.7.0)
```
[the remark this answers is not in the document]
TheOneRing: Hm now I remember you mentioned this before... 
Both ways sound bad, most ci scripts looked like they already get the info from git or something similar, could you provide an example for a script using this?

src/cmd/CMakeLists.txt
```diff
@@ -9,28 +6,17 @@ set(cmd_SRC
    )
 
 
-if(UNIX AND NOT APPLE)
-  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fPIE")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIE")
-  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -pie")
-endif()
```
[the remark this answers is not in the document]
TheOneRing: yes

src/crashreporter/CMakeLists.txt
```diff
@@ -32,7 +32,7 @@ list(APPEND crashreporter_RC "${CMAKE_CURRENT_BINARY_DIR}/resources.qrc")
 
 
 if(NOT BUILD_LIBRARIES_ONLY)
-    add_executable( ${CRASHREPORTER_EXECUTABLE} WIN32
+    add_executable( ${CRASHREPORTER_EXECUTABLE}
```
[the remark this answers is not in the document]
TheOneRing: yes thx your comment made me fix the crash reporter on mac

CMakeLists.txt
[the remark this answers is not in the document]
dschmidt: ... and manually patching buildsystems was also needed since the beginning of time. 
I prefer `list(APPEND ...)` but if you insist on `set(...)` please add the initial value there.

CMakeLists.txt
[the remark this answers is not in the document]
dschmidt: It enables autouic for non-Qt-targets. It's not a big deal, but it's not nice either. Especially because it's a new addition while *cleaning* *up* our CMake code.

src/gui/MacOSXBundleInfo.plist
```diff
@@ -9,25 +9,25 @@
...
         <string>????</string>
         <key>CFBundleVersion</key>
-        <string>@MIRALL_VERSION_FULL@</string>
+        <string>${MIRALL_VERSION_FULL}</string>
         <key>CFBundleShortVersionString</key>
-        <string>@MIRALL_VERSION_STRING@</string>
+        <string>${MIRALL_VERSION_STRING}</string>
         <key>NSHumanReadableCopyright</key>
-        <string>(C) 2014-2019 @APPLICATION_VENDOR@</string>
+        <string>(C) 2014-2019 ${APPLICATION_VENDOR}</string>
```
dschmidt: It's 2020 now :trollface:

src/gui/MacOSXBundleInfo.plist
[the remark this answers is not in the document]
dschmidt: Interesting. Didn't know this... 

For reference: CMake upstream does it like that, but still calls it .plist.in ;-)
https://github.com/Kitware/CMake/blob/master/Modules/MacOSXBundleInfo.plist.in

src/libsync/vfs/suffix/CMakeLists.txt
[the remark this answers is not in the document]
dschmidt: Sounds better!

src/gui/MacOSXBundleInfo.plist
TheOneRing: different commit

TheOneRing:
> Btw drone fails because it can't find an sqlite find/config module. You removed this in this PR, why and how is it supposed to work now?

It worked in my setups because craft installs a cmake config file...
So the sqlite version we provide on mac doesn't provide one?

TheOneRing:
> > Btw drone fails because it can't find an sqlite find/config module. You removed this in this PR, why and how is it supposed to work now?
> 
> It worked in my setups because craft installs a cmake config file...
> So the sqlite version we provide on mac doesn't provide one?

Also cmake seems to provide one since 3.14
https://cmake.org/cmake/help/v3.14/module/FindSQLite3.html?highlight=sqlite

TheOneRing:
@guruz @ogoffart @dschmidt 
Regarding the last commit, I think a better solution might be to compile this c_stuff into a static lib and link that? The exports are now needed as the build flags are now more strict and the visibility is set to hidden by default.

TheOneRing:
> @guruz @ogoffart @dschmidt
> Regarding the last commit, I think a better solution might be to compile this c_stuff into a static lib and link that? The exports are now needed as the build flags are now more strict and the visibility is set to hidden by default.

Well it was also only fixed on my mac, linux is still broken...

cmake/modules/ECMAddAppIcon.cmake
```diff
@@ -1,397 +0,0 @@
...
-                    if (offset GREATER -1)
-                        list(APPEND ${type}_at_${size}px "${icon_full}")
-                    elseif()
-                        message(STATUS "not found ${type}_at_${size}px ${icon_full}")
-                    endif()
-                endif()
-            endif()
-        endif()
-    endforeach()
-endmacro()
```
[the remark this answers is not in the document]
dschmidt: Did you?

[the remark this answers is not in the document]
TheOneRing: yes

CMakeLists.txt
```diff
@@ -153,9 +153,7 @@ if(BUILD_CLIENT)
     endif(APPLE)
 
     if(UNIX AND NOT APPLE)
-        find_package(INotify REQUIRED)
-    else()
-        find_package(INotify)
```
dschmidt: lol ... why was this there in the first place? Good catch.

TheOneRing: Hysteric reasons?

CMakeLists.txt
```diff
@@ -1,24 +1,44 @@
...
 
-project(client)
+include(VERSION.cmake)
+project(client LANGUAGES CXX C VERSION ${MIRALL_VERSION_MAJOR}.${MIRALL_VERSION_MINOR}.${MIRALL_VERSION_PATCH})
+include(FeatureSummary)
+find_package(ECM 5.50.0  NO_MODULE)
+set_package_properties(ECM PROPERTIES TYPE REQUIRED DESCRIPTION "Extra CMake Modules." URL "https://projects.kde.org/projects/kdesupport/extra-cmake-modules")
+feature_summary(WHAT REQUIRED_PACKAGES_NOT_FOUND FATAL_ON_MISSING_REQUIRED_PACKAGES)
+
+set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules ${ECM_MODULE_PATH} ${CMAKE_MODULE_PATH})
```
dschmidt: I still think this should be `list(APPEND` :-P

CMakeLists.txt
```diff
@@ -170,25 +170,27 @@ endif()
...
 # Handle Translations, pick all client_* files from trans directory.
 file( GLOB TRANS_FILES ${CMAKE_SOURCE_DIR}/translations/client_*.ts)
 set(TRANSLATIONS ${TRANS_FILES})
 
+
+if(APPLE)
+    list(APPEND OWNCLOUD_BUNDLED_RESOURCES ${CMAKE_CURRENT_SOURCE_DIR}/sync-exclude.lst)
+elseif(BUILD_CLIENT)
+    configure_file(sync-exclude.lst bin/sync-exclude.lst COPYONLY)
+    install( FILES sync-exclude.lst DESTINATION ${SYSCONFDIR}/${APPLICATION_SHORTNAME} )
```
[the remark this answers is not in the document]
TheOneRing: Not needed with Destination I think but I will add it

PR #7535, opened 2019-10-18
[CMAKE] Deprecate old custom flag for unittests

TheOneRing:


TheOneRing:
I decided to keep it around to not break ci scripts

dschmidt:
I'm a fan of the deprecation and the approach you've chosen here! 
Big thumbs up!

Issue #7169, opened 2019-04-30
Proxy settings are ignored without warning when connected to localhost 


HanaGemela:
Client: 2.6.0-daily20190429 (build 11690)
macOS 10.14.4

Steps to recreate:
1) Run ownCloud server and client on your machine
2) Connect the client to the server using localhost:8080
3) Set up the proxy in the Network tab in the client

Actual result: User can set proxy but this settings is ignored
Expected result: Logs mention that proxy settings has been ignored. Also user interface warns that proxy is not used - e.g. if connected to localhost and proxy has been set up, the Account tab displays 'Connected to http://localhost:8080 as admin (no proxy)'

ckamm:
Indeed, proxys are not used for local host by Qt (see https://doc.qt.io/qt-5/qnetworkproxy.html#details). The client could notify about that.

ogoffart:
How should we notify about it?  A label in the networksetting with a warning that is shown if at least one account has a localhost url?

guruz:
Either that or in the label that mentions the account state for that specific localhost account?

butonic:
um what is the rationale behind this? if I want to debug connections I locally run mitmproxy as well as my apache .... but how do I make the client use mitmproxy?

hm ok I had expected to make it use the proxy by giving it the public ip of my machine...  that did not work ... but using the public ip for the oc server did ...

:confused: ... anyway ... workaround is good enough ... maybe list this as a warning when running `owncloudcmd`...

ogoffart:
> um what is the rationale behind this? 

I don't know. That's what the network library we are using (Qt) is doing.

I guess the rationale is maybe that network settings are used globally for things that need access to the internet, and localhost does not need that.

ogoffart:
A warning was added when localhost is used as an host.
But we can't apply the proxy for localhost because Qt does not allow it.

PR #7538, opened 2019-10-21
Network Settings: Show a warning that proxy settings do not apply to localhost

ogoffart:
Only show this if at least one account is detected to have an url that looks
like localhost, because this could otherwise be confusing

Issue #7169


mmattel:
No reviewer?

ogoffart:
@guruz  ping

Issue #3111, opened 2015-04-14
Make PROPFIND parsing incremental on readyRead()


guruz:
... also change `LsColXMLParser` to be without signals and slots and just feed it data until we received `finished()` from QNetworkReply.

FYI @dragotin

guruz:
This is not 100% useful as the server does first have the complete PROPFIND Response XML in memory and only then serializes it to XML and has it sent out.

I remember we discussed with @evert that there is actually a way with PHP generators to improve this on the server side.

@PVince81 also mentioned that sometimes a node that gets put into the WebDAV XML response tree can take some time to generate (e.g. if remote share or external storage)

PVince81:
> the complete PROPFIND Response XML in memory and only then serializes it to XML and has it sent out.

Not sure if that's still the case in recent Sabre versions. From what I remember there was work done there to stream the response.
But at least I think that all nodes are in memory before being converted to XML.

evert:
We're changing this for an upcoming version actually. PROPFIND and REPORT will almost entirely be streamed, which means that the first byte gets sent out much sooner and memory usage will go down drastically.

guruz:
I've started working on the client side changes for this.

PR #7559, opened 2019-10-28
PROPFIND: Parse while data is coming in via readyRead()

guruz:
Issue: https://github.com/owncloud/client/issues/3111

It's quite broken and I didn't touch it for some months, but should get it out here so I don't lose it.

Got reminded about it because of https://github.com/sabre-io/dav/pull/1204
Thanks @DeepDiver1975 @petrkotek ;-)

### DIFF
--- a/src/libsync/discoveryphase.cpp
+++ b/src/libsync/discoveryphase.cpp
@@ -321,6 +321,7 @@ static void propertyMapToRemoteInfo(const QMap<QString, QString> &map, RemoteInf
 
 void DiscoverySingleDirectoryJob::directoryListingIteratedSlot(QString file, const QMap<QString, QString> &map)
 {
+    qDebug() << Q_FUNC_INFO << _ignoredFirst << file << map;
     if (!_ignoredFirst) {
         // The first entry is for the folder itself, we should process it differently.
         _ignoredFirst = true;
--- a/src/libsync/networkjobs.cpp
+++ b/src/libsync/networkjobs.cpp
@@ -150,102 +150,126 @@ bool MkColJob::finished()
     return true;
 }
 
-/*********************************************************************************************/
-// supposed to read <D:collection> when pointing to <D:resourcetype><D:collection></D:resourcetype>..
-static QString readContentsAsString(QXmlStreamReader &reader)
-{
-    QString result;
-    int level = 0;
-    do {
-        QXmlStreamReader::TokenType type = reader.readNext();
-        if (type == QXmlStreamReader::StartElement) {
-            level++;
-            result += "<" + reader.name().toString() + ">";
-        } else if (type == QXmlStreamReader::Characters) {
-            result += reader.text();
-        } else if (type == QXmlStreamReader::EndElement) {
-            level--;
-            if (level < 0) {
-                break;
-            }
-            result += "</" + reader.name().toString() + ">";
-        }
-
-    } while (!reader.atEnd());
-    return result;
-}
-
 
 LsColXMLParser::LsColXMLParser()
 {
+    currentPropsHaveHttp200 = false;
+    insideHref = false;
+    insidePropstat = false;
+    insideProp = false;
+    insideStatus = false;
+    insideMultiStatus = false;
+    insideResourcetype = false;
+}
+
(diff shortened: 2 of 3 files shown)


src/libsync/discoveryphase.cpp
```diff
@@ -321,6 +321,7 @@ static void propertyMapToRemoteInfo(const QMap<QString, QString> &map, RemoteInf
 
 void DiscoverySingleDirectoryJob::directoryListingIteratedSlot(QString file, const QMap<QString, QString> &map)
 {
+    qDebug() << Q_FUNC_INFO << _ignoredFirst << file << map;
```
ogoffart: please remove or ad a category.

src/libsync/networkjobs.cpp
```diff
@@ -150,131 +150,171 @@ bool MkColJob::finished()
...
-                break;
-            }
-            result += "</" + reader.name().toString() + ">";
-        }
 
-    } while (!reader.atEnd());
-    return result;
+LsColXMLParser::LsColXMLParser()
+{
+    currentPropsHaveHttp200 = false;
```
ogoffart: i'd say it is better to initialize all of them in the header, then we do not forget to initialize some

src/libsync/networkjobs.cpp
```diff
@@ -150,131 +150,171 @@ bool MkColJob::finished()
...
-    bool currentPropsHaveHttp200 = false;
-    bool insidePropstat = false;
-    bool insideProp = false;
-    bool insideMultiStatus = false;
+bool LsColXMLParser::parse(const QByteArray &xml, QHash<QString, qint64> *sizes, const QString &expectedPath)
+{
+    _expectedPath = expectedPath;
+    bool ok = parseSome(xml);
+    if (reader.hasError() && reader.error() == QXmlStreamReader::PrematureEndOfDocumentError) {
+        qDebug() << "fuXXXX";
```
ogoffart: !!!!

src/libsync/networkjobs.cpp
```diff
@@ -150,131 +150,171 @@ bool MkColJob::finished()
...
+        return false;
+    }
+    *sizes = this->sizes;
+    return ok;
+}
 
+bool LsColXMLParser::parseSome(const QByteArray &xml)
+{
+    qDebug() << Q_FUNC_INFO << this << xml << reader.lineNumber() << reader.columnNumber();
```
ogoffart: ditto (and there is more everywhere)

src/libsync/networkjobs.cpp
```diff
@@ -344,32 +384,60 @@ void LsColJob::start()
...
-// not all in one big blob at the end.
-bool LsColJob::finished()
+void LsColJob::newReplyHook(QNetworkReply *reply)
 {
-    qCInfo(lcLsColJob) << "LSCOL of" << reply()->request().url() << "FINISHED WITH STATUS"
-                       << replyStatusString();
+    qDebug() << Q_FUNC_INFO << reply;
+    connect(reply, &QNetworkReply::metaDataChanged, this, &LsColJob::slotMetaDataChanged);
+    connect(reply, &QIODevice::readyRead, this, &LsColJob::slotReadyRead);
+    // FIXME networkActivity for timeout reset?
```
ogoffart: is timeout not handled  already by the abstract network job?

src/libsync/networkjobs.cpp
```diff
@@ -344,32 +384,60 @@ void LsColJob::start()
...
+    }
+}
 
-        QString expectedPath = reply()->request().url().path(); // something like "/owncloud/remote.php/webdav/folder"
-        if (!parser.parse(reply()->readAll(), &_sizes, expectedPath)) {
-            // XML parse error
+void LsColJob::slotReadyRead() {
+    qDebug() << Q_FUNC_INFO;
+    if (_parser.isInitialized()) {
+        QByteArray snippet = reply()->readAll();
```
ogoffart: Note: there is a function QXmlStreamReader::setDevice you probably should use that instead of using slotReadyRead and putting the whole buffer in a QByteArray.

Issue #7557, opened 2019-10-25
Data loss when switching on VFS for .owncloud files


HanaGemela:
Similar to https://github.com/owncloud/client/issues/6953

Client: 2.6.0rc2 (build 12577)
Server: 10.3.0 stable
macOS 10.15

Steps to recreate:

1) upload the below files on the server 

```
test
test.owncloud
test.owncloud.owncloud
```

2) sync them to the client
3) turn on VFS
4) Account tab -> three dots menu -> availability -> free up local space 
5) Force sync

Actual files: One file is gone from the client, one file has lost its content, the third file test.owncloud.owncloud remain the same
Expected result: No data loss

ogoffart:
I wouldn't call that as data loss since the data is still on the server and you get the data back if you turn off VFS again

ogoffart:
When having files on the server named with the VFS extention, they are not properly synchronized to the client when using VFS. I'd say it is an expected limitation. I'm not sure what to do. I guess it should at least appear in the "not sinchronized"

jnweiger:
https://github.com/owncloud/client/issues/7556 has a reproducer for the data-loss. We start with three files, then we have two files. I don't think that switching off VFS afterwards can recreate the third file.

ogoffart:
So what should we do?

Suggestion:

1. If we detect such case, abort the sync and tell the user that virtual files cannot be enabled because there are files that ends with `.owncloud` in the sync directory.

2. Immediatly remove all the files that ends with `.owncloud`, as they are bining ignored from now on.

3. Leave it as is: the `.owncloud` files are no longer synchronized, but they are also ignored.  They can also be overwritten by an actual placeholder, which could be considered a dataloss if the file is not on the server.

4. Leave it almost as it, but do not overwrite existing file with a placeholder. Dehydrating or creating new placeholder for files with content would result as an error.

HanaGemela:
Suggestion number 1 is the most user friendly as it explains to the user what is going on and there are no surprises or wrong assumptions what will happen to .owncloud files.
Basically I like the idea of having a message telling the user what turning on VFS means for .owncloud files

ogoffart:
I actually went for option 4. in https://github.com/owncloud/client/pull/7562
There is still an error message that explains the problem in the not sync'ed dialog.

HanaGemela:
Files are not deleted but they are ignored now in 2.6.1sprint2 (build 12978), macOS 10.15.2

[...]

Issue #7556, opened 2019-10-25
[QA] Files with .owncloud suffix on the server cause a syncing loop


jnweiger:
* Connect Linux client 2.6.0rc2 to owncloud server 10.3
* Wait until initial sync is done.
* Pause syncing.
* upload the same file three times to the server with the web GUI like this:
   * upload adler05.jpg and rename on the server to adler05.jpg.owncloud.owncloud
   * upload adler05.jpg and rename on the server to adler05.jpg.owncloud
   * upload adler05.jpg
* resume syncing.

The client downloads all three of them:
![image](https://user-images.githubusercontent.com/1108546/67575889-c9aaf200-f73d-11e9-85c2-049dbdd7a2e4.png)
```
3& [testy:~/ownCloud5] $ ll adl*
-rw-r--r-- 1 testy testy 99226 Mai 12 01:10 adler05.jpg
-rw-r--r-- 1 testy testy 99226 Mai 12 01:10 adler05.jpg.owncloud
-rw-r--r-- 1 testy testy 99226 Mai 12 01:10 adler05.jpg.owncloud.owncloud
```
OK so far.

* enable virtual files ; wait for a sync -> Nothing changes.
* switch 'Availability' -> 'Free up local space'

The client starts looping. 
![image](https://user-images.githubusercontent.com/1108546/67576217-6cfc0700-f73e-11e9-8f49-e7b79d1d460d.png)

and contents of at least one file is lost:
```
3& [testy:~/ownCloud5] $ ll adl*
-rw-r--r-- 1 testy testy     1 Mai 12 01:10 adler05.jpg.owncloud
-rw-r--r-- 1 testy testy 99226 Mai 12 01:10 adler05.jpg.owncloud.owncloud
3& [testy:~/ownCloud5] $ 
```

jnweiger:
related issue: #6953

jnweiger:
when we switch back to 'Availability' -> 'Make always available locally' the looping stops, but we lost one file:
```3& [testy:~/ownCloud5] $ ll adl*
-rw-r--r-- 1 testy testy 99226 Mai 12 01:10 adler05.jpg
-rw-r--r-- 1 testy testy 99226 Mai 12 01:10 adler05.jpg.owncloud.owncloud
3& [testy:~/ownCloud5] $

jnweiger:
@ogoffart Force sync is not needed here to reproduce. The client immediately starts looping after switching the availability.

jnweiger:
[...]

PR #7562, opened 2019-10-30
VFS: Do not overwrite existing files by placeholder

ogoffart:
For issue #7557 and #7556

Note: this change the API of the VFS plugin, so the VFS plugin needs small
adaptations

### DIFF
--- a/src/common/result.h
+++ b/src/common/result.h
@@ -129,6 +129,18 @@ public:
 };
 
 namespace detail {
+    struct NoResultData{};
+}
+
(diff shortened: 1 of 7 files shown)


guruz:
Assuming 2.6.1 since bug is on 2.6.1

ogoffart:
Was merged in https://github.com/owncloud/client/pull/7568

TheOneRing:
Why was this merged when the ci was failing?

PR #7564, opened 2019-10-30
Share dialog: Remove and open the browser instead

ogoffart:
The Share Dialog tries to look like the page on the server, but by simply
opening the server page, we get rid of a big class of bug
since it is really hard to keep that dialog in sync with the server.
(As the server do changes from version to versions)

That way we can concentrate on doing the work on the sync engine, and
do not have to worry about the share dialog.


ogoffart:
Basically fixes all these bugs: https://github.com/owncloud/client/issues?q=is%3Aissue+is%3Aopen+label%3ASharing

ogoffart:
Of course, we do not have single sing on yet. But the browser is most likely logged on already.

guruz:
@michaelstingl poke

michaelstingl:
> @michaelstingl poke

I'm in conversations with other departments. I'd schedule this for post-2.7 if we plan to go for a quick 2.7.

DeepDiver1975:
Will this be the approach for mobile apps as well? 

My point is: the server has to deliver any necessary information to the clients so that they can properly perform any sharing operations. If information is missing we better add this (actually we did this already . see roles api).

Killing the sharing feature from desktop client and using the webui feels like a short circuit reaction of a little child that doesn't want to play with others anymore ...

michaelstingl:
@DeepDiver1975 product team is working on a concept. PR is not the right place for this fundamental discussion. Also mobile clients are out of scope of this PR.

ogoffart:
The issue is that over the years, there were many changes to the sharing dialog, both feature-wise and UI-wise. We aren't usually notified of such changes. And even if we were, it is still never gonna match the currently used server version.

Since everyone has a browser anyhow, and that this is anyway not something that can be done offline, i see no problem in using the actual server interface from the browser for this dialog.

Note that we are not "killing the sharing feature from desktop client", there will still be the option to share from the context menu. The difference is that we propose opening the actual owncloud UI instead of trying to mimic it.

michaelstingl:
Superseded by https://github.com/owncloud/client/pull/10241

Issue #7567, opened 2019-10-31
[QA] [Hardening] VFS plugin can be loaded from current directory


jnweiger:
owncloud client 2.6.0RC2 on Ubuntu 18.04

```
# run the owncloud client
$ owncloud &

# inspect from which path the vfs plugin is loaded.

$ jobs -l
[1]+  4826 Running                 owncloud &
$ lsof -p 4826 | grep vfs
owncloud 4826 testy  mem       REG              259,5    26776    321447 /opt/ownCloud/ownCloud/lib/x86_64-linux-gnu/ownCloud/plugins/owncloudsync_vfs_suffix.so
$ kill 4826
```

Place a copy of that library in my home, run again. For this example, I am using an exact copy, but some different code could be easily injected here.

```
$ cp /opt/ownCloud/ownCloud/lib/x86_64-linux-gnu/ownCloud/plugins/owncloudsync_vfs_suffix.so $HOME

# run the owncloud client again, and inspect the path of the vfs plugin used

$ cd $HOME
$ owncloud &
$ jobs -l
[1]+  4894 Running                 owncloud &
$ lsof -p 4894 | grep vfs
owncloud 48294testy  mem       REG               0,53    26776  11161695 /home/testy/owncloudsync_vfs_suffix.so
```

A library was picked up from my home.

Expected result: The vfs plugin is still picked up from the original place. No foreign code executed from a user writable directory.

jnweiger:
The ELF headers of the binaries and libraries we ship all have a nice and short RPATH. Not exactly sure why loading from the current working directory works at all.
```
$ ls -ld /opt/ownCloud/ownCloud/lib/*/* /opt/ownCloud/ownCloud/bin/* 
-rwxr-xr-x 1 root root 2562328 Oct 19 02:56 /opt/ownCloud/ownCloud/bin/owncloud
-rwxr-xr-x 1 root root   88312 Oct 19 02:56 /opt/ownCloud/ownCloud/bin/owncloudcmd
lrwxrwxrwx 1 root root      26 Oct 19 02:56 /opt/ownCloud/ownCloud/lib/x86_64-linux-gnu/libowncloud_csync.so.0 -> libowncloud_csync.so.2.6.0
-rw-r--r-- 1 root root  371008 Oct 19 02:56 /opt/ownCloud/ownCloud/lib/x86_64-linux-gnu/libowncloud_csync.so.2.6.0
[...]

PR #7568, opened 2019-11-01
VFS: use absolute path for the plugins

ogoffart:
Issue #7567


ogoffart:
Note that due to a mistake in my branch pushing commands, the title of this pull request was not matching with the commit it contained.

PR #7571, opened 2019-11-04
VFS: use absolute path for the plugins

ogoffart:
Issue #7567


ogoffart:
Yeah right, that works for me because i usually install the client.

- We could run `make install` as part of the CI tests.
- We could only use the absolute path when not compiling the tests.

ogoffart:
This is not a good idea. Better to patch Qt instead, see #7567

PR #7583, opened 2019-11-12
Update link to troubleshooting docs

michaelstingl:


settermjd:
Does this need to be backported, @ogoffart / @michaelstingl?

michaelstingl:
> Does this need to be backported, @ogoffart / @michaelstingl?

Don't think so. Used only for GH internally.

PR #7587, opened 2019-11-13
[318] Document how to show/hide the Crash Reporter from the config.ini

settermjd:
This PR updates the documentation, adding a link from the GUI config window to the config.ini file documentation. The main reason for doing so is to help ensure that the user knows that they can show and hide the crash reporter from the config file, just in case they need to hide it so that they can debug on-startup crashes. It fixes https://github.com/owncloud/docs/issues/318.

### DIFF
--- a/docs/modules/ROOT/pages/advanced_usage/configuration_file.adoc
+++ b/docs/modules/ROOT/pages/advanced_usage/configuration_file.adoc
@@ -14,7 +14,7 @@ You can locate this configuration file as follows:
 | `$HOME/.config/ownCloud/owncloud.cfg`
 | *Microsoft Windows* 
 | `%APPDATA%\ownCloud\owncloud.cfg`
-| *macOS X* 
+| *macOS* 
 |`$HOME/Library/ApplicationSupport/ownCloud/owncloud.cfg`
 |===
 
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/advanced_usage/configuration_file.adoc
```diff
@@ -1,13 +1,24 @@
...
+[cols="25%,75%",options="header"]
+|===
+| System
+| Location
+
+| *Linux* 
+| `$HOME/.config/ownCloud/owncloud.cfg`
+| *Microsoft Windows* 
+| `%APPDATA%\ownCloud\owncloud.cfg`
+| *macOS X*
```
dschmidt: I think it's either `macOS`  without `X` or `(Mac) OS X` 

@guruz @michaelstingl correct me if I'm wrong here

settermjd: `macOS` [it is](https://en.wikipedia.org/wiki/MacOS)!

settermjd: Fixed.

Issue #7547, opened 2019-10-23
Cannot exit owncloud client on gnome desktop


jnweiger:
Gnome deprecates the tray icon. There are workarounds and tricks to bring back a tray icon, as e.g. discussed in  https://github.com/owncloud/client/issues/6607

The Gnome desktop should work without tray icons. 
To control an app running in background, like our owncloud-client, the user is supposed to "start the application again". This works fine to bring up the settings page, as we already always run as one single application process.

A user can find the owncloud client icon and add it to favourites. that makes it easer to reach.

There is however one important entry that used to be in the tray-icon menu, that is no longer available now: Exit the application.

There are some generic options, when right clicking the icon in the activities, but exit is missing:

![image](https://user-images.githubusercontent.com/1108546/67400519-0a740100-f5ae-11e9-949e-6fef73659856.png)

When the settings window is open and in foreground, the application is shown in the title bar. 
There we have a "Beenden" (Quit) entry:
![image](https://user-images.githubusercontent.com/1108546/67400889-9f76fa00-f5ae-11e9-9dd0-532d8ffee246.png)

But this only closes the settings window, and the client continues to run.

Expected behaviour: One or multiple of
   * Make the existing "Beenden" (Quit) menue entry really exit the client (and expand the menu there to include other items from the original tray icon menu here) That is what most other apps have.
   * Add an exit option to the right click of the activities icon.
   * Add an exit option to the settings windows.

guruz:
What shall be the milestone for this?

ogoffart:
> Make the existing "Beenden" (Quit) menue entry really exit the client

This would mean that closing the settings windows exits owncloud. We probably do not want that.

> Add an exit option to the right click of the activities icon

Is that even possible?  How?

> Add an exit option to the settings windows

Probably the easier. But where to put this?
Add a button in the General settings tab? Should it only be visible in Gnome?

TheOneRing:
Add an exit button to the config window.

Investigate https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#extra-actions

TheOneRing:
![image](https://user-images.githubusercontent.com/200626/69726601-2de81980-1121-11ea-9131-fd61a410d3fd.png)
Obviously we'd need a new icon.

TheOneRing:
![image](https://user-images.githubusercontent.com/200626/69737656-0e5aec00-1135-11ea-859b-e66a28af779d.png)
[...]

PR #7611, opened 2019-11-27
Gnome allow to exit the application without systray

TheOneRing:
Issue https://github.com/owncloud/client/issues/7547


ogoffart:
Looks good to me...
I'm wondering however if we should show this only if the systray is not there.
It can be confusing to have both "Close" button, and Quit

michaelstingl:
> I'm wondering however if we should show this only if the systray is not there.

Yes, if we can detect this, I'd prefer it…

TheOneRing:
@dschmidt will transifex break the desktop file now?
Also do we need to define the icon for every language? It's always the same.

PR #7631, opened 2019-12-04
Prepend PLUGINDIR else its pointless

TheOneRing:


guruz:
I don't think this is correct..
What is the sympton you are trying to fix?

TheOneRing:
I had a special setup and different versions of the plugins in 2 locations and the wrong one was selected. I can also imagine that this could have an effect on unit tests run in the build folder.

PR #7644, opened 2019-12-10
[Settings] As the instance is created in the constructor we don't nee…

TheOneRing:
…d to check for it

### DIFF
--- a/src/gui/owncloudgui.cpp
+++ b/src/gui/owncloudgui.cpp
@@ -51,19 +51,16 @@ const char propertyAccountC[] = "oc_account";
 
 ownCloudGui::ownCloudGui(Application *parent)
     : QObject(parent)
-    , _tray(nullptr)
+    , _tray(new Systray(this))
     , _settingsDialog(new SettingsDialog(this))
     , _logBrowser(nullptr)
     , _recentActionsMenu(nullptr)
     , _app(parent)
 {
-    _tray = new Systray();
-    _tray->setParent(this);
-
     // for the beginning, set the offline icon until the account was verified
     _tray->setIcon(Theme::instance()->folderOfflineIcon(/*systray?*/ true, /*currently visible?*/ false));
 
-    connect(_tray.data(), &QSystemTrayIcon::activated,
+    connect(_tray, &QSystemTrayIcon::activated,
         this, &ownCloudGui::slotTrayClicked);
 
     setupActions();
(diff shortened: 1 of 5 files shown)


src/gui/owncloudgui.cpp
```diff
@@ -1025,8 +1020,7 @@ void ownCloudGui::slotShutdown()
     // that saving the geometries happens ASAP during a OS shutdown
 
     // those do delete on close
-    if (!_settingsDialog.isNull())
-        _settingsDialog->close();
+    _settingsDialog->close();
```
jnweiger: Is it safe to ->close()M if .isNull() ?

src/gui/settingsdialog.cpp
```diff
@@ -84,13 +84,11 @@ QString shortDisplayNameForSettings(OCC::Account *account, int width)
 namespace OCC {
 
 SettingsDialog::SettingsDialog(ownCloudGui *gui, QWidget *parent)
-    : QDialog(parent)
+    : QDialog(parent, Qt::WindowFlags() & ~Qt::WindowContextHelpButtonHint)
     , _ui(new Ui::SettingsDialog)
```
jnweiger: Should we also setAttribute(Qt::WA_DeleteOnClose, true); here?

TheOneRing: The widget is only initialised once so we don't want it to be deleted.

src/gui/owncloudgui.cpp
TheOneRing: The dialogue is initialised in the constructor and should always exist.

src/gui/settingsdialog.cpp
ogoffart: But don't we want to delete it when we exit?

src/gui/owncloudgui.cpp
```diff
@@ -97,12 +94,17 @@ ownCloudGui::ownCloudGui(Application *parent)
         this, &ownCloudGui::slotShowGuiMessage);
 }
 
+ownCloudGui::~ownCloudGui()
+{
+    _settingsDialog->deleteLater();
+}
```
dschmidt: Why do we need this? Isn't the SettingsDialog parented to the ownCloudGui anyway?

TheOneRing: ownCloudGui is a QObject and the dialog a Widget so it doesn't work here.

TheOneRing:
> I don't think the title/description of the PR is accurate.
> 
> It should be more around the lines of:
> 
> > Create the settings dialog only once and don't delete it when it's hidden
> > This allows us to simplify the code
> 
> I'm not sure whether this is a good or a bad thing. 🤷‍♂

We had quite some code to recreate the dialog in case it was deleted.
But the current code never deleted it so it was just dead code.

> 
> Before approving I would like to know why we now need `_settingsDialog->deleteLater()`.

src/gui/owncloudgui.cpp
dschmidt: Huh? Can you elaborate a bit more?

TheOneRing:
ping

src/gui/systray.h
```diff
@@ -34,6 +34,7 @@ class Systray : public QSystemTrayIcon
 {
     Q_OBJECT
 public:
+    Systray(QObject *parent = nullptr);
```
ogoffart: Btw, one can do `using QSystemTrayIcon::QSystemTrayIcon;`  to inherit the constructor.

PR #7653, opened 2019-12-17
document how to disable chunking

jnweiger:
@TheOneRing reading the source, can you confirm, that this should disable chunking?

### DIFF
--- a/docs/modules/ROOT/pages/advanced_usage/environment_variables.adoc
+++ b/docs/modules/ROOT/pages/advanced_usage/environment_variables.adoc
@@ -19,7 +19,7 @@ The environment variables are:
 | 10000000 (or 10 MB)
 | Specifies the initial chunk size of uploaded files in bytes.
 The client will dynamically adjust this size within the maximum and minimum bounds (see below).
-To disable chunking completely, set these four variables to 0: `OWNCLOUD_CHUNK_SIZE=0 OWNCLOUD_MAX_CHUNK_SIZE=0 OWNCLOUD_MIN_CHUNK_SIZE=0 OWNCLOUD_CHUNKING_NG=0`
+To disable chunking completely, set `OWNCLOUD_CHUNK_SIZE=0` .
 
 | `OWNCLOUD_MAX_CHUNK_SIZE`
 | 100000000 (or 100 MB)
--- a/docs/modules/ROOT/pages/advanced_usage/environment_variables.adoc
+++ b/docs/modules/ROOT/pages/advanced_usage/environment_variables.adoc
@@ -19,6 +19,7 @@ The environment variables are:
 | 10000000 (or 10 MB)
 | Specifies the initial chunk size of uploaded files in bytes.
 The client will dynamically adjust this size within the maximum and minimum bounds (see below).
+To disable chunking completely, set these four variables to 0: `OWNCLOUD_CHUNK_SIZE=0 OWNCLOUD_MAX_CHUNK_SIZE=0 OWNCLOUD_MIN_CHUNK_SIZE=0 OWNCLOUD_CHUNKING_NG=0`
 
 | `OWNCLOUD_MAX_CHUNK_SIZE`
 | 100000000 (or 100 MB)


jnweiger:
Then please leave a pointer: 'See [FAQ](URL) for more details'
The FAQ did not come up when googling for 'owncloud chunking environment variables'.

mmattel:
My pov, this perfectly fits where it is made and a approval / merging would be totally fine

docs/modules/ROOT/pages/advanced_usage/environment_variables.adoc
```diff
@@ -19,6 +19,7 @@ The environment variables are:
 | 10000000 (or 10 MB)
 | Specifies the initial chunk size of uploaded files in bytes.
 The client will dynamically adjust this size within the maximum and minimum bounds (see below).
+To disable chunking completely, set these four variables to 0: `OWNCLOUD_CHUNK_SIZE=0 OWNCLOUD_MAX_CHUNK_SIZE=0 OWNCLOUD_MIN_CHUNK_SIZE=0 OWNCLOUD_CHUNKING_NG=0`
```
ogoffart: I think setting OWNCLOUD_CHUNK_SIZE=0 should be enough to disable chunking.
What makes you think you need to set all of this. Did you try it and OWNCLOUD_CHUNK_SIZE=0 was not enough?

jnweiger: Not tried. I suggested all 4 of them, just to be sure. 
What should the user check, to learn if chunking is on or off?

ogoffart: OWNCLOUD_CHUNK_SIZE=0 is the only required one that disable the chunking completely.

Setting `OWNCLOUD_MIN_CHUNK_SIZE`  or `OWNCLOUD_MAX_CHUNK_SIZE` to 0 would be a very bad idea as it would mean that the client will want to send an infinite amount of chunk of size 0. These are only used in the computation of the chunk size and not in the decision to use chunking or not.
Setting `OWNCLOUD_CHUNKING_NG` would just force a chunking algorithm but not decide wether or not use chunks.

So in other words, only `OWNCLOUD_CHUNK_SIZE=0` is required and should be documented.

mmattel:
Ping, any progress ?

docs/modules/ROOT/pages/advanced_usage/environment_variables.adoc
michaelstingl:
```suggestion
To disable chunking completely, set `OWNCLOUD_CHUNK_SIZE=0` .
```

mmattel: Suggestion has been incorporated

mmattel:
@settermjd mind to take a look why the pdf build in drone is failing?

mmattel:
Seems that another font has an issue:
`No such file or directory - mplus-1mn-regular.ttf not found in fonts`

docs/resources/themes/owncloud-theme.yml
```diff
@@ -5,20 +5,20 @@ font:
       normal: notoserif-regular.ttf
       bold: notoserif-bold.ttf
       italic: notoserif-italic.ttf
-      bold_italic: notoserif-bold_italic.ttf
+      bold_italic: notoserif-bolditalic.ttf
```
TheOneRing: Why are the fonts touched here?

TheOneRing:
I rebased the code

docs/resources/themes/owncloud-theme.yml
settermjd: Because the configuration file uses incorrect font file names.

settermjd:
> Seems that another font has an issue:
> `No such file or directory - mplus-1mn-regular.ttf not found in fonts`

I don't get the error locally. Not sure how the Drone build and local build work differently, yet.

Issue #7574, opened 2019-11-05
Logo doesn't fit in About dialog


HanaGemela:
Client 2.6.0 rc4

Steps to recreate:
1) Go to General tab
2) Click on the version

Actual result: Only part of the logo is displayed
Windows 10
<img width="839" alt="Screenshot 2019-11-04 at 16 54 47" src="https://user-images.githubusercontent.com/49001702/68197314-58c7cd80-ffba-11e9-822b-5cdb0c4dcb21.png">

Ubuntu 18.04
<img width="613" alt="Screenshot 2019-11-05 at 10 49 30" src="https://user-images.githubusercontent.com/49001702/68197316-5b2a2780-ffba-11e9-91e9-e8384d8c4e91.png">

Expected result: Logo has been scaled down to fit

michaelstingl:
@ChrisEdS I think you worked on this before?

ChrisEdS:
@guruz fixed this in the 2.5.4 release before. It looks like a regression.

ogoffart:
Note that the previous issue was https://github.com/owncloud/enterprise/issues/3257
But the fix was only valid for the owncloud icon from the default theme.
But these icons that are used now are even much bigger than before.

So the fix in #7582 should make sure the icon is small enough.

HanaGemela:
@TheOneRing please provide testpilot Sprint 01 build

HanaGemela:
@TheOneRing please provide testpilot Sprint 02 build

TheOneRing:
https://jenkins.int.owncloud.com/job/client-trigger/3992/

HanaGemela:
Still broken, Windows 10, client 2.6.1 sprint2

<img width="847" alt="Screenshot 2020-01-08 at 16 04 41" src="https://user-images.githubusercontent.com/49001702/71988696-a4a02880-3230-11ea-98df-ff8308ddf9fc.png">

HanaGemela:
Fixed in 2.6.1sprint2 (build 12923), Windows 10
And in 2.6.1sprint2 (build 2001), Ubuntu 19.10

PR #7663, opened 2020-01-08
[GUI] Limit about dialog image size to 1/4 of the screen

TheOneRing:
Fixes: #7574


TheOneRing:
Or we just set a reasonable size?

PR #7676, opened 2020-01-15
Include vfs plugin in issue template

TheOneRing:


.github/issue_template.md
```diff
@@ -33,6 +33,8 @@ Storage backend (external storage):
 ### Client configuration
 Client version:
 
+Virtual File(No, Suffix, Windows):
```
michaelstingl:
```suggestion
Virtual File enabled (No, Suffix, Windows):
```

TheOneRing:
is it ok now?

HanaGemela:
It mixes too many things into one. I'd say `Virtual files enabled (yes or no)`.
We can guess the plugin from operating system. If that is not enough, I'd add `Virtual files plugin (suffix or Wincfapi)` and added a hint how to find it in General tab

TheOneRing:
I think in general we should just require the info string from the about dialog.
--version should in future report the same string.
This will basically make the whole questionair obsolete.

TheOneRing:
@HanaGemela Instead of that questionnaire I'd prefer to ask users to always share the --version string which will tell us about the used libraries the exact version and the used plugin.
What do you think?

HanaGemela:
@TheOneRing Yeah, it might be easier

PR #7680, opened 2020-01-16
Only set plugin search path in dev builds

TheOneRing:


CMakeLists.txt
```diff
@@ -89,13 +89,13 @@ set(DATADIR "share")
 endif(WIN32)
 set(SHAREDIR ${DATADIR})
 
-if (NOT APPLE)
-    set(PLUGINDIR "${CMAKE_INSTALL_FULL_LIBDIR}/${APPLICATION_SHORTNAME}/plugins" CACHE STRING "Extra path to look for Qt plugins like for VFS. May be relative to binary.")
-else()
-    # Inside the .app bundle
-    set(PLUGINDIR "../PlugIns" CACHE STRING "Extra path to look for Qt plugins like for VFS. May be relative to binary.")
+if (MIRALL_VERSION_SUFFIX)
```
ogoffart: I don't thin that's the right condition.  QA and community are mainly testing rc build and such. So it is best if that did not do any of such change.

TheOneRing: The only issue here occurs on a dev system, when plugins are loaded from the build folder instead of the install location, this shouldn't be an issue for qa.
But yes we could add an option instead. What do you think?

TheOneRing:
ping

src/gui/application.cpp
```diff
@@ -263,14 +263,11 @@ Application::Application(int &argc, char **argv)
 
 #ifdef PLUGINDIR
     // Setup extra plugin search path
-    QString extraPluginPath = QStringLiteral(PLUGINDIR);
-    if (!extraPluginPath.isEmpty()) {
-        if (QDir::isRelativePath(extraPluginPath))
```
dschmidt: Any idea why this was made explicitly absolute here?

CMakeLists.txt
```diff
@@ -89,13 +89,13 @@ set(DATADIR "share")
 endif(WIN32)
 set(SHAREDIR ${DATADIR})
 
-if (NOT APPLE)
-    set(PLUGINDIR "${CMAKE_INSTALL_FULL_LIBDIR}/${APPLICATION_SHORTNAME}/plugins" CACHE STRING "Extra path to look for Qt plugins like for VFS. May be relative to binary.")
-else()
-    # Inside the .app bundle
-    set(PLUGINDIR "../PlugIns" CACHE STRING "Extra path to look for Qt plugins like for VFS. May be relative to binary.")
```
dschmidt: I assume this is because we do not build the full bundle inside the build folder anymore? Or why is this not needed anymore?

src/gui/application.cpp
TheOneRing: We used a relative path on Mac, I'm not sure how relative plugin paths are defined(what are they relative to).

PR #7681, opened 2020-01-16
Allow OIDC without wellknown url

TheOneRing:


src/libsync/CMakeLists.txt
```diff
@@ -109,13 +110,13 @@ add_library(${synclib_NAME} SHARED ${libsync_SRCS})
 target_link_libraries(${synclib_NAME}
     "${csync_NAME}"
     ${OS_SPECIFIC_LINK_LIBRARIES}
-    Qt5::Core Qt5::Network
+    Qt5::Core Qt5::Network Qt5::Widgets
```
ogoffart: Eh, wait!  we don't want to add more widget dependency in libsync.

Well, then we no longer need TOKEN_AUTH_ONLY.

(We used to have customer project that needed to work without QtGui,  but these are no longer existing at the moment,  still would be good to keep linsync working withuout gui)

test/testoauth.cpp
```diff
@@ -199,7 +199,7 @@ class OAuthTestCase : public QObject
                 { "access_token", "123" },
                 { "refresh_token" , "456" },
                 { "message_url",  "owncloud://success"},
-                { "user_id", "789" },
+                { "user_id", "admin" },
```
ogoffart: Can you explain why this change is needed

src/libsync/CMakeLists.txt
TheOneRing: Hm I'd expect that tokenauth only means also no oauth? in that case I'll just move the QtWidget back into the if

src/libsync/creds/oauth.cpp
```diff
@@ -210,13 +210,14 @@ void OAuth::finalize(QPointer<QTcpSocket> socket, const QString &accessToken,
...
-                                    authorisationLink().toString(QUrl::FullyEncoded));
-        httpReplyAndClose(socket, "200 OK", message.toUtf8().constData());
+        qCWarning(lcOauth) << "We expected the user" << _account->davUser() << "but the server answered with user" << user;
+        const QString message = tr("<h1>Wrong user</h1>"
+                                   "<p>You logged-in with user <em>%1</em>, but must login with user <em>%2</em>.<br>"
+                                   "Please log out of %3 in another tab, then <a href='%4'>click here</a> "
+                                   "and log in as user %2</p>")
+                                    .arg(user, _account->davUser(), Theme::instance()->appNameGUI(),
+                                        authorisationLink().toString(QUrl::FullyEncoded));
+        httpReplyAndClose(socket, "401 Unauthorized", message.toUtf8().constData());
```
ogoffart: 401 means the browser should retry with a password.
I think the right code here would be 403

PR #7687, opened 2020-01-21
Cleanup

TheOneRing:


TheOneRing:
They are not

TheOneRing:
I tried to look for unreferenced source files with 
```git ls-files | %{ if (-not (git grep  (Get-Item $_).Name)) { write-host "Found potential dead file $_" }}```
And found some more stuff to remove

TheOneRing:
I tried to look for unreferenced source files with 
```git ls-files | %{ if (-not (git grep  (Get-Item $_).Name)) { write-host "Found potential dead file $_" }}```
And found some more stuff to remove

PR #7692, opened 2020-01-22
Fix for forked repos

micbar:
# Description
Fixes the changelog pipeline for forked repositories

## Before
Screenshot_2020-01-22 Build #3294 1 1 - owncloud client - Drone CI

## After
Screenshot_2020-01-22 Build #3302 1 1 - owncloud client - Drone CI(1)

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -292,8 +292,6 @@ def build_client_docs(ctx):
     }
 
 def changelog(ctx, trigger = {}, depends_on = []):
-    pipelines = []
-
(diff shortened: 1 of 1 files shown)


micbar:
@tboerger works as you suggested, value is always populated.

micbar:
@tboerger actually it needed an enhancement #7694 

Merge commits in master have and empty `ctx.build.source_repo`.
So added a conditional.

PR #7699, opened 2020-01-24
[WIP]Add previously missing changelog entries

TheOneRing:
What changed https://github.com/owncloud/client/compare/v2.6.1...master


TheOneRing:
TODO:
https://github.com/owncloud/client/commit/54d1bb95a0cb36fac0d4638ed0fbb893ce69228a

Maybe:
https://github.com/owncloud/client/commit/d0a165c22fe92d58360655228ba1e13279ffc653 ?
The fact that we now use extra cmake modules?
The new generation of the mac packager?

HanaGemela:
I'd rather talk about owncloud desktop client than introducing OpenID Connect

Something like: OwnCloud desktop client now uses OpenID Connect 1.0. to verify the identity of the end user... Benefits for the users/admins... Why we are using it...

HanaGemela:
We should discuss who we write the changelog for. The server changelog is written for end users and includes only changes affecting the users - new features or bugfixes that can be seen. That means no tests, no infrastructure, no stuff for developers etc.

Issue #7700, opened 2020-01-24
Saving cookies on Windows might fail


TheOneRing:

HanaGemela:
@TheOneRing is this something to test? If so, please provide steps

TheOneRing:
On a fresh install on Windows the client failed to store the cookies.
The log files showed something about a QDevice error or something similar.
As I added a unit test to ensure that this is working I don't think there is a need for manual testing.

PR #7701, opened 2020-01-24
Fix saving of cookies

TheOneRing:
Fixes: #7700

And again we need to discuss the change log


TheOneRing:
@micbar and how am I supposed to reference the pr before I create a review?
Do I need to create a issue if there was none?
Can I just use a random name instead?

micbar:
Yes. This has some limitations.

Calens needs one primary issue id. It is always the id of the first url in the list.

That could be an issue that you create before creating the pr.
This issue could point then to the pr.
The url should be github.com/vendor/repo/issues/1234  which BTW always resolves both types, issues and PRs.

TheOneRing:
Yes we now use the changelog.md but it's only generated on Windows

micbar:
> Yes we now use the changelog.md but it's only generated on Windows

Windows?

dschmidt:
Are we losing the old changelog completely? Can we somehow migrate the old changelog to the new autogenerated one?

micbar:
> Are we losing the old changelog completely? Can we somehow migrate the old changelog to the new autogenerated one?

Already done. See on master.

test/testcookies.cpp
```diff
@@ -0,0 +1,42 @@
...
+        QNetworkCookie cookieA = QNetworkCookie("foo", "bar");
+        // tomorrow rounded
+        cookieA.setExpirationDate(QDateTime(QDateTime::currentDateTimeUtc().addDays(1).date()));
+        const QList<QNetworkCookie> cookies = {cookieA, QNetworkCookie("foo2", "bar")};
+        CookieJar jar;
+        jar.setAllCookies(cookies);
+        QCOMPARE(cookies, jar.allCookies());
+        QVERIFY(jar.save(tmp.filePath("test.db")));
+        QVERIFY(jar.save(nonexistingPath));
+
```
dschmidt: At first it was a bit confusing to me why you saved to two different paths and only verified loading only in one of them afterwards... maybe leave a comment what you are testing exactly.

guruz:
Please note that the old hand-made changelog contains (well, should) REAL CHANGES in USER UNDERSTANDABLE language and not just random git log messages.

TheOneRing:
> Please note that the old hand-made changelog contains (well, should) REAL CHANGES in USER UNDERSTANDABLE language and not just random git log messages.

As you can see in https://github.com/owncloud/client/blob/master/CHANGELOG.md the new change log does it too, it is supposed to be even more detailed.

micbar:
@guruz What brought you to the conclusion that we are using git log messages?

TheOneRing:
If the change log is sufficient this one is ready to be merged?

TheOneRing:
ping

PR #7702, opened 2020-01-27
[OAuth] Improve logging

TheOneRing:


src/libsync/creds/oauth.cpp
```diff
@@ -260,12 +262,21 @@ QByteArray OAuth::generateRandomString(size_t size) const
...
+    return *out;
+}
+
 QUrl OAuth::authorisationLink() const
 {
     Q_ASSERT(_server.isListening());
     QUrlQuery query;
     QByteArray code_challenge = QCryptographicHash::hash(_pkceCodeVerifier, QCryptographicHash::Sha256)
-        .toBase64(QByteArray::Base64UrlEncoding | QByteArray::OmitTrailingEquals);
+    .toBase64(QByteArray::Base64UrlEncoding | QByteArray::OmitTrailingEquals);
```
ogoffart: that looked better before

TheOneRing: I think that's caused by clang format

src/libsync/creds/oauth.cpp
```diff
@@ -182,25 +171,39 @@ void OAuth::startAuthentification()
...
-                // Invalid or empty JSON: Network error maybe?
-                qCWarning(lcOauth) << "Error while refreshing the token" << reply->errorString() << jsonData << jsonParseError.errorString();
-            } else if (accessToken.isEmpty()) {
-                // If the json was valid, but the reply did not contain an access token, the token
-                // is considered expired. (Usually the HTTP reply code is 400)
-                qCDebug(lcOauth) << "Expired refresh token. Logging out";
-                Q_EMIT refreshFinished(QString(), QString());
+            const QString errorReason = errorFromJson(json);
+            if (!errorReason.isEmpty()) {
+                if (errorReason == QStringLiteral("invalid_grant")) {
```
TheOneRing: Can I assume this response or should I rather check for a http error code?

TheOneRing: To answer my question no.
Is there a proper way to detect a rejection getting proper json without a token feels wrong?

src/libsync/creds/oauth.cpp
```diff
@@ -260,12 +264,33 @@ QByteArray OAuth::generateRandomString(size_t size) const
...
+    return *out;
+}
+
+QString OAuth::errorFromJson(const QJsonObject &json)
+{
+    if (json.isEmpty()) {
+        return {};
+    }
+    QString errorFromJson = json[QStringLiteral("error_description")].toString();
+    if (errorFromJson.isEmpty())
```
TheOneRing: TODO: don't use error_description to identify the error, this is a user targeting string.

src/libsync/creds/oauth.cpp
```diff
@@ -182,25 +176,40 @@ void OAuth::startAuthentification()
...
-                // Invalid or empty JSON: Network error maybe?
-                qCWarning(lcOauth) << "Error while refreshing the token" << reply->errorString() << jsonData << jsonParseError.errorString();
-            } else if (accessToken.isEmpty()) {
-                // If the json was valid, but the reply did not contain an access token, the token
-                // is considered expired. (Usually the HTTP reply code is 400)
-                qCDebug(lcOauth) << "Expired refresh token. Logging out";
-                Q_EMIT refreshFinished(QString(), QString());
+            const QString error = json.value(QLatin1String("error")).toString();
+            if (!error.isEmpty()) {
+                if (error == QLatin1String("invalid_grant") ||
```
TheOneRing: @DeepDiver1975 can we assume this?

src/libsync/abstractnetworkjob.h
```diff
@@ -98,6 +98,10 @@ class OWNCLOUDSYNC_EXPORT AbstractNetworkJob : public QObject
      */
     static int httpTimeout;
 
+    /** whether or noth this job can be retried */
+    bool retryAble() const;
```
ogoffart: Perhaps this should be more explicit that this is only about retrying after renewing the auth token.
So the name should be something like `isRetryableAfterAuthToken`
And actually, i would change it for something different as to describe that this is actually means that the job should work when there is no authentication:  `isUnauthenticatedJob`

(as for Qt API naming convention,  boolean starts with 'is' and are named in such a way that the default is false)

TheOneRing: hm maybe it would be best to call it  isAuthenticationJob() as only those are not allowed to be retried?

ogoffart: :+1:

src/libsync/creds/oauth.cpp
```diff
@@ -182,25 +176,40 @@ void OAuth::startAuthentification()
...
-                Q_EMIT refreshFinished(accessToken, json["refresh_token"].toString());
+                if (jsonParseError.error != QJsonParseError::NoError || json.isEmpty()) {
+                    // Invalid or empty JSON: Network error maybe?
+                    qCWarning(lcOauth) << tr("Error while refreshing the token: %1 : %2").arg(jsonParseError.errorString(), QString::fromUtf8(jsonData));
+                } else {
+                    QString error;
+                    accessToken = getRequiredField(json, QStringLiteral("access_token"), &error).toString();
+                    newRefreshToken = getRequiredField(json, QStringLiteral("refresh_token"), &error).toString();
+                    if (!error.isEmpty()) {
+                        qCWarning(lcOauth) << tr("The reply from the server did not contain all expected fields\n:%1").arg(error);
```
ogoffart: We need to be carefull not to get in an infinite loop that would flood the server with token refresh queries.

We probably should if the server is returning another error for a reason or another, that would happen.

Issue #7710, opened 2020-01-31
Closing VFS dialog behaves like confirmation 


HanaGemela:
macOS 10.15.2
Windows 10
client 2.6.1

Steps to recreate:
1) Click on Enable (or disable) VFS -> Confirmation dialog will pop up
2) Click the cross icon to close the dialog

Actual result: VFS has been enabled (respectively disabled). 
Expected result: No changes to the files when the dialog has been closed

michaelstingl:
fixed

HanaGemela:
Broken in 2.6.3263_rc1 (build 13516) macOS 10.15.3

Steps to recreate:

1. Click on Disable VFS -> Confirmation dialog will pop up
2. Click the cross icon to close the dialog

HanaGemela:
Still broken in 2.6.3263rc2 (build 13648)

TheOneRing:
![vfs](https://user-images.githubusercontent.com/200626/82817003-78620d00-9e9c-11ea-9335-9579f0f1d4d2.gif)

HanaGemela:
Oh, we have more VFS dialogs :) I'm opening the dialog from the Account tab

TheOneRing:
That should behave the same 😭 I'll see

TheOneRing:
You where talking about the disable vfs dialog?

HanaGemela:
Yes. Enable is correct

PR #7717, opened 2020-02-03
Fix vfs wizard issues

TheOneRing:
@HanaGemela did I fix the correct issue with the fix for #7710

### DIFF
--- a/changelog/unreleased/7709
+++ b/changelog/unreleased/7709
@@ -1,0 +1,5 @@
+Bugfix: Fix crash if setup wizard is closed while the virtual file system  dialog is open
+
+
+https://github.com/owncloud/client/issues/7709
+https://github.com/owncloud/client/issues/7711
--- a/changelog/unreleased/7710
+++ b/changelog/unreleased/7710
@@ -1,0 +1,6 @@
+Bugfix: Don't show virtual file system warning again when the radio button is triggered a second time
+
+Declining the dialog had no effect as the radio button was already checked
+
+https://github.com/owncloud/client/issues/7710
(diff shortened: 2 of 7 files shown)


ogoffart:
Why was it crashing?

I'm not a fan of using exec() as it makes the callstack deeper and sometimes can have even worse problems.

If some pointer get dangling, maybe they need to be wrapped in a QPointer and tested.

TheOneRing:
I use exec() as it should be a modal dialog.
Users should be forced to accept or decline before they can do anything else.
If we didn't do soemthing evil I wouldn't expect issues by calling exec().

The crash was caused by by the caller being deleted before the signal was emitted.
So the receiver didn't exist anymore, in a connect the receiver should always be specified to determine the live time of the lambda.

Making the dialog blocking fixes the issue with multiple dialogs as the user can't click on the radio button while the dialog is displayed.

ogoffart:
`QDialog::open` should also be model. It is documented as such.  Maybe the problem is that the parent is not properly set?

And indeed, it may be worth to specify a better receiver.  

But what exactly is deleted too early in this case?

TheOneRing:
> `QDialog::open` should also be model. It is documented as such. Maybe the problem is that the parent is not properly set?
> 
> And indeed, it may be worth to specify a better receiver.
> 
> But what exactly is deleted too early in this case?

The non modal dialog is displayed, the user ignores it, and closes the wizard, or ownCloud.
During destruction of the dialog the the finished signal is emitted, the callback is executed but the variables referenced in the lambda do no longer exist.

TheOneRing:
@ogoffart should be ready now

PR #7719, opened 2020-02-04
Fix tree

TheOneRing:


src/gui/folderstatusmodel.cpp
```diff
@@ -1090,10 +1090,7 @@ void FolderStatusModel::slotFolderSyncStateChange(Folder *f)
     if (f->syncResult().folderStructureWasChanged()
         && (state == SyncResult::Success || state == SyncResult::Problem)) {
         // There is a new or a removed folder. reset all data
-        auto &info = _folders[folderIndex];
-        auto idx = index(folderIndex);
-        info.resetSubs(this, idx);
-        fetchMore(idx);
+        fetchMore(index(folderIndex));
```
ogoffart: this does nothing if the older was already fetched. 
We need to clean the cache or do something so we are sure to actually do a request and remove the folders that no longer exist.

src/gui/folderstatusmodel.cpp
```diff
@@ -1200,7 +1197,7 @@ void FolderStatusModel::slotNewBigFolder()
         return;
     }
 
-    _folders[folderIndex].resetSubs(this, index(folderIndex));
```
ogoffart: i still believe the reset is required.

TheOneRing: As said, I think its covered by https://github.com/owncloud/client/blob/58d54d68e09e5eca9ef83bf9173fe2ac79732a44/src/gui/folderstatusmodel.cpp#L564

src/gui/folderstatusmodel.cpp
TheOneRing: I don't get it.

TheOneRing:
how ab out now?

src/gui/folderstatusmodel.cpp
```diff
@@ -589,6 +589,13 @@ void FolderStatusModel::fetchMore(const QModelIndex &parent)
     QTimer::singleShot(1000, this, &FolderStatusModel::slotShowFetchProgress);
 }
 
+void FolderStatusModel::resetAndFetch(const QModelIndex &parent)
+{
+    auto info = infoForIndex(parent);
+    info->_fetched = false;
```
ogoffart: This would still not re-fetch if there was already a fetch in progress  
(info->_fetchingJob)

I think the best is to simply call `resetSubs` here.
Calling it twice does not hurt, and it is even more explicit of what it does.

TheOneRing: ok

PR #7721, opened 2020-02-05
Fix warnings and modernize code

ogoffart:
Run clang-tidy over the code since we know get warning when there are missing override

Also fix a couple of deprecated warning as more function has been marked as deprecated in Qt 5.14

See individual commit.


ogoffart:
(the QSet change needs Qt 5.14 so i'll remove it)

PR #7727, opened 2020-02-06
OAuth: the well-known url is always at the host root

ogoffart:
So this means that even id the owncloud URL is
https://some.server/something/something,
the well-known url should be
https://some.server/.well-known/openid-configuration and not
https://some.server/somethins/somethins/.well-known/openid-configuration

### DIFF
--- a/src/libsync/creds/oauth.cpp
+++ b/src/libsync/creds/oauth.cpp
@@ -326,7 +326,8 @@ void OAuth::fetchWellKnown()
     }
     else
     {
-        QUrl wellKnownUrl = Utility::concatUrlPath(_account->url(), QStringLiteral("/.well-known/openid-configuration"));
+        QUrl wellKnownUrl = _account->url();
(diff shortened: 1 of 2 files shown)


src/libsync/creds/oauth.cpp
```diff
@@ -326,7 +326,8 @@ void OAuth::fetchWellKnown()
     }
     else
     {
-        QUrl wellKnownUrl = Utility::concatUrlPath(_account->url(), QStringLiteral("/.well-known/openid-configuration"));
+        QUrl wellKnownUrl = _account->url();
```
TheOneRing: could you constify it?

ogoffart: no, we modify it in the next line

ogoffart:
It seems this is not correct.  .well-known path is not required to be in the root, as shown in this document:
https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-protocols-oidc#fetch-the-openid-connect-metadata-document

Issue #7729, opened 2020-02-06
2.6.1 on Ubuntu 18.04 fails to load vfs extension


jnweiger:
![image](https://user-images.githubusercontent.com/1108546/73947587-fbfceb80-48f7-11ea-9766-8a9395631460.png)

It should say vfs plugin: suffix

The plugin is installed as `/opt/ownCloud/ownCloud/lib/x86_64-linux-gnu/ownCloud/plugins/owncloudsync_vfs_suffix.so` but does not seem to get loaded.

jnweiger:
Instead a file that happens to be in my homedirectory gets loaded:
`-rw-r--r-- 1 testy testy 26776 Oct 31 15:44 owncloudsync_vfs_suffix.so`

and it does not work. Didn't we fix that long ago?

ogoffart:
It was fixed in Qt 5.14.0
(The fix will also be in  Qt 5.12.7)

jnweiger:
Yeay! looking forward to 5.12.7 ! 

Refernence: https://github.com/owncloud/client/issues/7567

michaelstingl:
Qt 5.12.7 update: https://github.com/owncloud/client/issues/7708
/cc @TheOneRing

michaelstingl:
Will be fixed with #7708 in 2.6.2

PR #7730, opened 2020-02-06
update dependecies to Qt-5.12.7

jnweiger:
By changing our dependencies to Qt version 5.12.7 we resolve several known
issues with older Qt releases:

* plugins were picked up from the current directory.

https://github.com/owncloud/client/issues/7729
https://github.com/owncloud/client/issues/7567

### DIFF
--- a/changelog/unreleased/issue-7729
+++ b/changelog/unreleased/issue-7729
@@ -1,0 +1,9 @@
+Change: Update to Qt-5.12.7
+
+By changing our dependencies to Qt version 5.12.7 we resolve several known
+issues with older Qt releases:
+
+* plugins were picked up from the current directory.
+
+https://github.com/owncloud/client/issues/7729
+https://github.com/owncloud/client/issues/7567


TheOneRing:
Not done yet, also should we add updated 3rdparty libs to the change log?

TheOneRing:
@michaelstingl ?

michaelstingl:
> Not done yet, also should we add updated 3rdparty libs to the change log?

Yeah, I think so.

guruz:
> also should we add updated 3rdparty libs to the change log?

Note that we never consistently did this as there is so many variables involved (platforms, distros etc..)

TheOneRing:
> > also should we add updated 3rdparty libs to the change log?
> 
> Note that we never consistently did this as there is so many variables involved (platforms, distros etc..)

Exactly the changelog will also be part of distro packages that still use older versions of qt

michaelstingl:
@guruz @TheOneRing No strong opinion. Up 2 U…

changelog/unreleased/issue-7729
```diff
@@ -0,0 +1,9 @@
+Change: Update to Qt-5.12.7
```
TheOneRing: Change: Bundle Qt 5.12.7

We updated the bundled version of Qt tot 5.12.7.
This resolves the following issues:
* plugins were picked up from the current directory.

https://github.com/owncloud/client/issues/7729
https://github.com/owncloud/client/issues/7567

PR #7734, opened 2020-02-10
Oidc

TheOneRing:


src/libsync/creds/oauth.h
```diff
@@ -59,7 +59,7 @@ class OWNCLOUDSYNC_EXPORT OAuth : public QObject
     OAuth(Account *account, QObject *parent);
     ~OAuth();
     void startAuthentification();
-    void refreshAuthentification(const QString &refreshToken);
+    void refreshAuthention(const QString &refreshToken);
```
ogoffart: I did not know Authention is a word.  What's the reason behind this  change?

ogoffart: Maybe it should be "Authentication" ?

src/libsync/creds/oauth.h
```diff
@@ -59,7 +59,7 @@ class OWNCLOUDSYNC_EXPORT OAuth : public QObject
     OAuth(Account *account, QObject *parent);
     ~OAuth();
     void startAuthentification();
```
ogoffart: This should also be renamed.

src/libsync/creds/oauth.h
TheOneRing: thx 🤦‍♀

PR #7736, opened 2020-02-10
[OAuth] Pass scope parameter also on token refresh

TheOneRing:


src/libsync/creds/oauth.cpp
```diff
@@ -25,6 +25,11 @@
 #include "creds/httpcredentials.h"
 #include <QRandomGenerator>
 
+namespace {
+// TODO: make theming parameter
+QString SCOPE = QStringLiteral("openid offline_access email");
```
ogoffart: One should avoid non-trivial static global object.
Use Q_GLOBAL_STATIC,  or return from a function, or just plain char*

src/libsync/creds/oauth.cpp
```diff
@@ -101,7 +106,6 @@ void OAuth::startAuthentication()
                     { QStringLiteral("code"), args.queryItemValue(QStringLiteral("code")) },
                     { QStringLiteral("redirect_uri"), QStringLiteral("http://localhost:%1").arg(_server.serverPort()) },
                     { QStringLiteral("code_verifier"), _pkceCodeVerifier },
-                    { QStringLiteral("scope"), args.queryItemValue(QStringLiteral("scope")) },
```
ogoffart: Wait, was that not the correct thing to do?

TheOneRing: I have no idea and it looks like we would need the full set anyhow....
@DeepDiver1975 ?

ogoffart: But then we would need to use `SCOPE()` here.
If you don't use it here, there is also no reason to use in the other call.

TheOneRing: It's now set in postTokenRefresh

PR #7740, opened 2020-02-12
fix notoserif font

xoxys:
Same fixes as:
https://github.com/owncloud/docs/pull/1889
https://github.com/owncloud/docs/pull/1796


dschmidt:
Ugly that we need to duplicate these things :-/ 

But let's unbreak PRs...

PR #7742, opened 2020-02-13
Rename the changelog unreleased direcrory

ogoffart:
Otherwise it will be hard to track when merging with master


TheOneRing:
will be moved on the 2.6 branch once we release

PR #7743, opened 2020-02-13
[CMake] Directly install sync-exclude.lst on Windows to the correct d…

TheOneRing:
…irectory


ogoffart:
sync-exclude.lst should actually be placed in $XDG_DATA_DIR  instead of sysconfdir

CMakeLists.txt
```diff
@@ -183,7 +183,11 @@ if(APPLE)
     list(APPEND OWNCLOUD_BUNDLED_RESOURCES "${CMAKE_CURRENT_SOURCE_DIR}/sync-exclude.lst")
 elseif(BUILD_CLIENT)
     configure_file(sync-exclude.lst bin/sync-exclude.lst COPYONLY)
```
ogoffart: Not sure if this is needed then.

Issue #7744, opened 2020-02-14
quit button not shown, when name is too long


jnweiger:
Seen with a branded fablab client 2.6.1 on CentOS8

![image](https://user-images.githubusercontent.com/1108546/74545200-bd46e100-4f48-11ea-97b6-bd74c17e0b8c.png)

This branded client has its Application name set to: "FabLab Cloud Nbg -- Application name in the Desktop section --"

When the Application Name is set to a string longer than what would fit on the remaining space, the Quit button is not shown, instead a small arrows at the right hand side indicates an overhang menu having the Quit button.
Technically this is correct behaviour,...

Improvement: shorten the Application name in the same way as we shorten the addresses for in the connection tabs.

TheOneRing:
![image](https://user-images.githubusercontent.com/200626/74819768-2fc90f80-5301-11ea-96e5-9c30cc94f64e.png)

TheOneRing:
![image](https://user-images.githubusercontent.com/200626/74820060-a6660d00-5301-11ea-92da-a6b05cc07a8f.png)

TheOneRing:
ok works with your string and two accounts

PR #7750, opened 2020-02-19
[GUI] Dynamically elide actions

TheOneRing:
Fixes: #7744


ogoffart:
Does it works if you have many accounts?

TheOneRing:
image
image

PR #7754, opened 2020-02-21
Branded no experimental

jnweiger:
hard coded theme names ownCloud and testpilotcloud
are already used to show crashreporter options. We could do the same for experimental options.


TheOneRing:
Lets do that properly

jnweiger:
obsoleted by #7757, thanks!

Issue #7304, opened 2019-07-11
Remove all files message is too complicated


jnweiger:
client 260daily20190711 on win10, with demo.owncloud.org

The client was off for a while. The server reset itself, then the client is connected again.
This message pops up.

![image](https://user-images.githubusercontent.com/1108546/61042464-8ae26d80-a3d4-11e9-8d83-45ac91588d9d.png)

The second sentence gives me a hard time. Syncing a "delete" probably means delete the files. If so, please say so. What "right to restore" do I need. Does "making ... unavailable" mean, it is moved to some hidden trashcan, where I need admin rights? If it is simply deleted, then we should say so.

Why is there a warning about "rights to do so" when choosing "keep the files"? Is that different from a normal sync?

The last sentence somehow contradicts the second. It seems to say, that nothing is deleted unless I am the owner. Client runs undermy account, so I am the owner of almost all files here. I don't believe that.

Suggestion:

    Remove All Files?
     
    All files in the sync folder 'ownCloud' were deleted on the server.
    Choose keep  to re-upload to the server.
    Chose remove to delete them locally too.
     
    [Keep]   [Remove]

ckamm:
Simplifying like this looks good to me. Any concerns @michaelstingl ?

dragotin:
How about this:

```
All files in the sync folder 'ownCloud' were deleted on the server.

Do you want to restore them from your local files or also remove locally?

[Restore]    [Remove]
```

michaelstingl:
@hodyroff @pmaier1 could you have a look?

hodyroff:
I remember that this was carefully worded ;) Need to think about it ... at least folder seems twice in the first sentence. The challenge is that if you don't have write access there will be nothing restored, thats why its worded also for read-only users. Was there not a similar message which got cleaned up already?

dragotin:
Does this message also show if a share from another user was removed, and as a result the sync client does not longer see it? That might also influence the wording, because fi the user decides to restore the files, she will create the removed share locally in her space (not longer as a share obviously) what is a bit against the idea ...

ckamm:
The message shows exactly if it's determined that a sync action is performed on every single known file and at least one of them is a deletion.

So in the "share removed" scenario the message would only show if the user has no other files.

TheOneRing:
Another nice piece of information would be the account name/url...

jnweiger:
Dragotin suggested:
> ```
[...]

PR #7760, opened 2020-02-24
Fix potential null pointer access

TheOneRing:
@ogoffart do you think too that could be the cause?
https://sentry.io/organizations/owncloud/issues/1529161263/events/02509984b5ca42ffb3960d9c9e161414/?project=79001&statsPeriod=14d

### DIFF
--- a/changelog/unreleased/7760
+++ b/changelog/unreleased/7760
@@ -1,0 +1,4 @@
+Bugfix: Fixed a potential crash on cancelation of discovery jobs
+
+
+https://github.com/owncloud/client/pull/7760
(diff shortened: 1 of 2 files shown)


ogoffart:
I don't see how this could be related to the linked crash.

TheOneRing:
@ogoffart https://github.com/owncloud/client/blob/68c6d85a3f8299b20e4299b7617be3d6fd9af170/src/libsync/discoveryphase.cpp#L218

ogoffart:
Ah yes, maybe it is related

ogoffart:
Note that the stack trace has a QDialog::exec() and re-enter the event loop (so maybe some event are not processed in the order we would expect them)

TheOneRing:
I left a  note in https://github.com/owncloud/client/issues/7304

HanaGemela:
@TheOneRing please provide test to recreate

TheOneRing:
I don't think this one can be tested manually.

PR #7767, opened 2020-02-24
Reuqire qt 5.12

TheOneRing:
See https://github.com/owncloud/client/pull/7730
Ci will fail until we update our dependencies.


ogoffart:
Maybe not 5.12.7,  but at least 5.12

TheOneRing:
Yesterday we talked about documenting 3rd party updates.
The only way we can claim something got fixed through a 3rd party update is by enforcing at least that version.
Or come up with a proper way to describe deployment changes in the changelog.

PR #7769, opened 2020-02-25
Run changelog ci on 2.6

TheOneRing:
This is a backport from master.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -9,7 +9,19 @@ def main(ctx):
     translations_trigger = {
         "cron": ["translations"],
     }
+    changelog_trigger = {
+         "ref": [
+            "refs/heads/master",
+            "refs/pull/**",
+        ],
+    }
     pipelines = [
+        # Build changelog
+        changelog(
+            ctx,
+            trigger = changelog_trigger,
+            depends_on = [],
+        ),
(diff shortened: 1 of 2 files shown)


.drone.star
```diff
@@ -262,6 +275,63 @@ def build_client_docs(ctx):
         },
     }
 
+def changelog(ctx, trigger = {}, depends_on = []):
+    repo_slug = ctx.build.source_repo if ctx.build.source_repo else ctx.repo.slug
+    result = {
+        'kind': 'pipeline',
+        'type': 'docker',
+        'name': 'changelog',
+        'clone': {
```
TheOneRing: @dschmidt this is a backport from master I just removed the publish step.
Can you explain to me why clone is disabled and implemented manually?

micbar: Clone is disabled because:
- Default clone clones a commit id
- This custom clone step clones the head of the branch
- This is important for the build on master, there could be other merge commits "in betwen.
- So you make sure that you do not build the changelog on an outdated commit and publish changes that are behind the current tip of master

TheOneRing: OK I realised that the clone step takes quite some time. I guess a shallow lone would help.

Issue #7762, opened 2020-02-24
+ gets lost when requesting login with oAuth2


HanaGemela:
Server: 10.4.0 RC3
Client: 2.6.1 (build 13096)

Steps to recreate:
1) Install oAuth2 on the server
2) Add a new account with username `name+2@test.com` in the client
3) Log out
4) Log in

Actual result: Client requests login for `name 2@test.com` -> not possible to login
Expected result: Client requests login for `name+2@test.com`

![Screenshot 2020-02-24 at 11 00 10](https://user-images.githubusercontent.com/49001702/75146678-dfd0bb00-56fb-11ea-891d-9f552cee0674.png)

Found when testing https://github.com/owncloud/core/pull/36613

ogoffart:
We use `QUrlQuery` to encode all our fields (in `OAuth::authorisationLink`) 
`+` normaly does not need to be encoded. The documentation of QUrlQuery has a documentation about that: https://doc.qt.io/qt-5/qurlquery.html#handling-of-spaces-and-plus
But QUrlQuery does not seem to allow to force the percent-encoding of `+`

The owncloud server understands an unencoded `+` as a space, which i think is a server bug.

We could workaround the problem by doing a search and replace of `+` with `%2B` in the query.

michaelstingl:
> The owncloud server understands an unencoded `+` as a space, which i think is a server bug.

Open issue?

jnweiger:
A '+' (plus) is well known but historic url-encoding for a ' ' (whitespace). A raw plus character should always be percent-encoded, to avoid confusion.
 
https://de.wikipedia.org/wiki/URL-Encoding mentions that form-url-encoding differs from normal url-encoding because of the use of '+'. I disagree.

michaelstingl:
fixed in https://github.com/owncloud/client/pull/7771

PR #7771, opened 2020-02-25
OAuth: make sure to encode the '+' in query string

ogoffart:
The server understands '+' as a space.

Issue #7762

### DIFF
--- a/changelog/unreleased/7762
+++ b/changelog/unreleased/7762
@@ -1,0 +1,7 @@
+Bugfix: Fixed OAuth2 login of user with `+`
+
+Make sure that the `+` in the user name is properly encoded in the URL
+opened by the browser when doing an OAuth2 authentication.
+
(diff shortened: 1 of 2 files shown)


ogoffart:
For reference, as i wrote in this comment: https://github.com/owncloud/client/issues/7762#issuecomment-590801153

We use QUrlQuery to encode all our fields (in OAuth::authorisationLink)
+ normaly does not need to be encoded. The documentation of QUrlQuery has a documentation about that: https://doc.qt.io/qt-5/qurlquery.html#handling-of-spaces-and-plus
But QUrlQuery does not seem to allow to force the percent-encoding of +

The owncloud server understands an unencoded + as a space, which i think is a server bug.

michaelstingl:
> The owncloud server understands an unencoded + as a space, which i think is a server bug.

Open server issue?

PR #7785, opened 2020-03-03
Debug

TheOneRing:


TheOneRing:
Maybe I should use a shorter namespace name.
```03-03 16:55:47:430 [ info sync.discovery ]:	Processing "ownCloud Manual.pdf.owncloud" | valid: true/true/true | mtime: 1580914383/1580914383/1580914383 | size: 5083143/1/5083143 | etag: "a434f8acaca1190999e812a3029e6c3a"//"a434f8acaca1190999e812a3029e6c3a" | checksum: "SHA1:82c49d440ac4c034e0a2d8c2b9dd6a6402c86c5e"//"SHA1:82c49d440ac4c034e0a2d8c2b9dd6a6402c86c5e" | perm: WDNVR//WDNVR | fileid: "00000013ocfn2qccq5uk"//"00000013ocfn2qccq5uk" | inode: 22589414/22589414/ | type: CSyncEnumMetaData::ItemTypeVirtualFile/CSyncEnumMetaData::ItemTypeVirtualFile/CSyncEnumMetaData::ItemTypeFile```

File: src/csync/csync_util.cpp [hunk 1]
```diff
@@ -198,3 +198,5 @@ bool csync_is_collision_safe_hash(const QByteArray &checksum_header)
     return checksum_header.startsWith("SHA")
         || checksum_header.startsWith("MD5:");
 }
+
+#include "moc_csync.cpp"
```
ogoffart: Why do we need this?  And why here?

I guess we need this because the file does not have Q_OBJECT which is recognized by cmake.
maybe something like
```
#if 0
Q_OBJECT
#endif 
```

would work...


But anyway, if including this file work as well, that's ok too.

TheOneRing: Have to fix windows builds :)

TheOneRing: Sigh...
To use that on windows we need qt 5.14...
https://doc.qt.io/qt-5/qobject.html#Q_NAMESPACE_EXPORT

ogoffart: before, you can still do 
`CSYNC_EXPORT Q_NAMESPACE`
with the same effect

TheOneRing: Your right so there was no reason at all to add jet another macro...

PR #7797, opened 2020-03-11
Simplify oauth test

TheOneRing:


TheOneRing:
Well but its simpler :D
And quter xD

PR #7800, opened 2020-03-12
Update year in rc files and cleanup win extension code

TheOneRing:


shell_integration/windows/OCUtil/RegistryUtil.cpp
```diff
@@ -12,10 +12,10 @@
  * details.
  */
 
-#include "stdafx.h"
-
 #include "RegistryUtil.h"
 
+#include <Windows.h>
```
dschmidt: Nitpick: for consistency's sake you could make this lowercase

TheOneRing: Uh juck thx

PR #7804, opened 2020-03-13
Fix crash reporter in craft

dschmidt:


dschmidt:
This is not correct, TARGET_FILE_DIR is in the build dir, we need the install dir.

src/cmd/CMakeLists.txt
```diff
@@ -14,6 +14,11 @@ if(NOT BUILD_LIBRARIES_ONLY)
 
     # Need tokenizer for netrc parser
     target_include_directories(${cmd_NAME} PRIVATE ${CMAKE_SOURCE_DIR}/src/3rdparty/qtokenizer)
-    install(TARGETS ${cmd_NAME} ${KDE_INSTALL_TARGETS_DEFAULT_ARGS})
+
+    if(APPLE)
+      install(TARGETS ${cmd_NAME} RUNTIME DESTINATION "${KDE_INSTALL_BUNDLEDIR}/${APPLICATION_EXECUTABLE}.app/Contents/MacOS/")
```
TheOneRing: For the shell extension I cam up with https://github.com/owncloud/client/blob/master/shell_integration/MacOSX/CMakeLists.txt#L18
I think the generator expression is nicer, I can't remember why I decided to use the copy command but I think I had issues with the install command....

TheOneRing:
Well it sucks that cmake is still not providing something proper...

PR #7810, opened 2020-03-18
Cleanup  csync_instructions_e

TheOneRing:


src/csync/csync.h
```diff
@@ -103,26 +103,27 @@ Q_ENUM_NS(csync_status_codes_e)
...
-  CSYNC_INSTRUCTION_RENAME          = 0x00000004,  /* The file need to be renamed (RECONCILE) */
-  CSYNC_INSTRUCTION_EVAL_RENAME     = 0x00000800,  /* The file is new, it is the destination of a rename (UPDATE) */
-  CSYNC_INSTRUCTION_NEW             = 0x00000008,  /* The file is new compared to the db (UPDATE) */
-  CSYNC_INSTRUCTION_CONFLICT        = 0x00000010,  /* The file need to be downloaded because it is a conflict (RECONCILE) */
-  CSYNC_INSTRUCTION_IGNORE          = 0x00000020,  /* The file is ignored (UPDATE|RECONCILE) */
-  CSYNC_INSTRUCTION_SYNC            = 0x00000040,  /* The file need to be pushed to the other remote (RECONCILE) */
-  CSYNC_INSTRUCTION_STAT_ERROR      = 0x00000080,
-  CSYNC_INSTRUCTION_ERROR           = 0x00000100,
-  CSYNC_INSTRUCTION_TYPE_CHANGE     = 0x00000200,  /* Like NEW, but deletes the old entity first (RECONCILE)
+enum _csync_instructions_e {
```
ogoffart: SyncInsturction

ogoffart:
Do we want that in 2.6 or should it go to master instead?
maybe safer in master. Although it is true that this probably won't break much

TheOneRing:
Lets merge it to master

src/csync/csync_util.cpp
```diff
@@ -40,27 +40,9 @@ Q_LOGGING_CATEGORY(lcCSyncUtils, "sync.csync.utils", QtInfoMsg)
 
 typedef struct {
   const char *instr_str;
-  enum csync_instructions_e instr_code;
+  SyncInsturctions instr_code;
 } _instr_code_struct;
```
ogoffart: this is no longer used.

src/gui/syncrunfilelog.cpp
```diff
@@ -41,7 +41,7 @@ QString SyncRunFileLog::directionToStr(SyncFileItem::Direction dir)
     return re;
 }
 
-QString SyncRunFileLog::instructionToStr(csync_instructions_e inst)
+QString SyncRunFileLog::instructionToStr(SyncInsturctions inst)
```
ogoffart: Note that this function only consider that it has only one value,  not used as a flag.  So it should not take the QFlags

src/libsync/syncfileitem.h
```diff
@@ -252,7 +252,7 @@ class OWNCLOUDSYNC_EXPORT SyncFileItem
     // usually this value is 1, but for removes on dirs, it might be much higher.
 
     // Variables used by the propagator
-    csync_instructions_e _instruction;
+    SyncInsturctions _instruction;
```
ogoffart: As far as i know, the item can only have only one instruction at the time.
This should be a `SyncInsturction`, not the QFlags.

TheOneRing: Well I can't come up with many use cases besides for tests like
`ENFORCE(instruction & (NEW | REMOVE))`
But yes maybe we should not encourage that.

PR #7812, opened 2020-03-19
ActivityListModel: Fix assertion on remove

TheOneRing:


src/gui/activitylistmodel.cpp
```diff
@@ -217,17 +217,19 @@ void ActivityListModel::slotRefreshActivity(AccountState *ast)
...
+        int i = 0;
         while (it.hasNext()) {
             Activity activity = it.next();
             if (activity._accName == accountToRemove) {
-                beginRemoveRows(QModelIndex(), i, i + 1);
+                beginRemoveRows(QModelIndex(), i, i);
                 it.remove();
                 endRemoveRows();
+            } else {
+                ++i;
```
TheOneRing: Not sure about i, but thats the only way I could make sense of it.

PR #7813, opened 2020-03-19
Notification 

TheOneRing:


TheOneRing:
See https://github.com/owncloud/client/issues/7043#issuecomment-600733357

TheOneRing:
It breaks windows :(
image

Issue #7799, opened 2020-03-12
Crash with vfs_suffix


TheOneRing:
When moving files around it is possible that a virtual file is recreated after it was moved.
When that now non existing virtual file is moved the client crashes with an assertion.

This crash was produced using an external dropbox share.
In order to reproduce it the same file needs to be moved until we get a duplicate, then the duplicate needs to be moved.


![Clipboard - March 12, 2020 11_10 AM](https://user-images.githubusercontent.com/200626/76511219-23912780-6453-11ea-94e7-da031eae2079.png)

![Clipboard - March 12, 2020 11_13 AM](https://user-images.githubusercontent.com/200626/76511216-22f89100-6453-11ea-8e82-f2a4090a969d.png)

https://github.com/owncloud/client/blob/master/src/libsync/discoveryphase.cpp#L162

[log.txt](https://github.com/owncloud/client/files/4323562/log.txt)

```1   __pthread_kill                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             (x86_64) /usr/lib/system/libsystem_kernel.dylib       0x7fff72b407fa 
2   pthread_kill                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               (x86_64) /usr/lib/system/libsystem_pthread.dylib      0x7fff72bfdbc1 
3   abort                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      (x86_64) /usr/lib/system/libsystem_c.dylib            0x7fff72ac7a1c 
4   qt_message_fatal(QtMsgType, QMessageLogContext const&, QString const&)                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     qlogging.cpp                                     1907 0x10207e7a8    
5   QMessageLogger::fatal(const char *, ...) const                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             qlogging.cpp                                     888  0x102583a2b    
6   OCC::DiscoveryPhase::findAndCancelDeletedJob(QString const&)                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               discoveryphase.cpp                               162  0x100ff0a5a    
[...]

PR #7814, opened 2020-03-20
Fix vfs sync crash

TheOneRing:
Issue #7799

### DIFF
--- a/src/libsync/discoveryphase.cpp
+++ b/src/libsync/discoveryphase.cpp
@@ -159,14 +159,24 @@ QPair<bool, QByteArray> DiscoveryPhase::findAndCancelDeletedJob(const QString &o
     QByteArray oldEtag;
     auto it = _deletedItem.find(originalPath);
     if (it != _deletedItem.end()) {
-        ENFORCE((*it)->_instruction == CSYNC_INSTRUCTION_REMOVE
+        const csync_instructions_e instruction = (*it)->_instruction;
+        if (instruction == CSYNC_INSTRUCTION_IGNORE && (*it)->_type == ItemTypeVirtualFile) {
             // re-creation of virtual files count as a delete
-             || ((*it)->_type == ItemTypeVirtualFile && (*it)->_instruction == CSYNC_INSTRUCTION_NEW)
-             || ((*it)->_isRestoration && (*it)->_instruction == CSYNC_INSTRUCTION_NEW)
-               );
-        (*it)->_instruction = CSYNC_INSTRUCTION_NONE;
-        result = true;
-        oldEtag = (*it)->_etag;
+            // a file might be in an error state and thus gets marked as CSYNC_INSTRUCTION_IGNORE
+            // after it was initially marked as CSYNC_INSTRUCTION_REMOVE
+            // return true, to not trigger any additional actions on that file that could elad to dataloss
+            result = true;
+            oldEtag = (*it)->_etag;
+        } else {
+            ENFORCE(instruction == CSYNC_INSTRUCTION_REMOVE
+                // re-creation of virtual files count as a delete
+                || ((*it)->_type == ItemTypeVirtualFile && instruction == CSYNC_INSTRUCTION_NEW)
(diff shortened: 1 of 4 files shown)


src/csync/csync.h
```diff
@@ -103,7 +103,7 @@ Q_ENUM_NS(csync_status_codes_e)
   * Instruction enum. In the file traversal structure, it describes
   * the csync state of a file.
   */
-enum csync_instructions_e {
+enum _csync_instructions_e {
```
ogoffart: maybe CSyncInstruction.  Type are camel case in the Qt coding style.

upercase for the first character means member in the owncloud coding style.

src/csync/csync_util.cpp
```diff
@@ -71,7 +71,7 @@ struct csync_memstat_s {
   int dt;
 };
 
-const char *csync_instruction_str(enum csync_instructions_e instr)
+const char *csync_instruction_str(csync_instructions_e instr)
```
ogoffart: (maybe we no longer need this function)

test/testsyncmove.cpp
```diff
@@ -857,6 +857,58 @@ private slots:
         QVERIFY(fakeFolder.syncOnce());
         QCOMPARE(fakeFolder.currentLocalState(), fakeFolder.currentRemoteState());
     }
+
+    void testMovedWithError()
+    {
+#define VFS 1
```
ogoffart: maybe can be a QFETCH  and have a _data function to test both.

test/testsyncmove.cpp
```diff
@@ -857,6 +857,58 @@ private slots:
...
+
+        // sync1 file gets detected as error, instruction is still NEW_FILE
+        fakeFolder.syncOnce();
+
+        // sync2 file is in error state, checkErrorBlacklisting sets instruction to IGNORED
+        fakeFolder.syncOnce();
+
+#if VFS
+        // TODO: why does the comparison need local files
+        // comparison of states fails as local files have a suffix
```
ogoffart: I guess that's because the local state has the suffix and the remote test doesn't.
Maybe this was just not implemented in the test framework.  Check if the suffix virtual file test do the comparison.

src/csync/csync_util.cpp
TheOneRing: true

ogoffart:
> QVERIFY has no effect outside of a test, use Q_ASSERT

But all these QVERIFY should only be run from tests.
They should return and log the failure even if they are not in the main test functions.

QVERIFY is better than the assert because it logs the faillure, and it lets further test run.

src/libsync/discoveryphase.cpp
```diff
@@ -159,14 +159,20 @@ QPair<bool, QByteArray> DiscoveryPhase::findAndCancelDeletedJob(const QString &o
...
-            // re-creation of virtual files count as a delete
-             || ((*it)->_type == ItemTypeVirtualFile && (*it)->_instruction == CSYNC_INSTRUCTION_NEW)
-             || ((*it)->_isRestoration && (*it)->_instruction == CSYNC_INSTRUCTION_NEW)
-               );
-        (*it)->_instruction = CSYNC_INSTRUCTION_NONE;
-        result = true;
-        oldEtag = (*it)->_etag;
+        const csync_instructions_e instruction = (*it)->_instruction;
+        if (instruction == CSYNC_INSTRUCTION_IGNORE && (*it)->_type == ItemTypeVirtualFile) {
+            result = true;
```
ogoffart: I think the result needs to be false here.  We must not consider this as a delete anymore since it is not.
On the other hand, the calling code will assume that it is a move regardless so this may not matter actually.

Either way, i think this deserve a comment.

TheOneRing: I added an explaining comment.
Returning false triggers actual action, the placeholder file gets removed. See https://github.com/owncloud/client/pull/7814/files#diff-04e3bc4ea34b9a9a993967604f3d74faR924

TheOneRing:
> 👍
> 
> You could not reproduce the bug without virtual files?

Yes this issue only occurs with virtual files.

PR #7818, opened 2020-03-23
Tests: Fail on error

TheOneRing:
QVERIFY has no effect besides a return outside of the test function itself.

> Note: This macro can only be used in a test function that is invoked by the test framework.

https://doc.qt.io/qt-5/qtest.html#QVERIFY

Therefore use Q_ASSERT in the testing stubs.


TheOneRing:
Cool github created two pr's...

PR #7819, opened 2020-03-23
Tests: Fail on error

TheOneRing:
QVERIFY has no effect besides a return outside of the test function itself.

> Note: This macro can only be used in a test function that is invoked by the test framework.

https://doc.qt.io/qt-5/qtest.html#QVERIFY

Therefore use Q_ASSERT in the testing stubs.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,3 +1,4 @@
+add_definitions(-DQT_FORCE_ASSERTS)
 find_package(SQLite3 3.8.0 REQUIRED)
 include_directories(${CMAKE_SOURCE_DIR}/src
                     ${CMAKE_SOURCE_DIR}/src/3rdparty/qtokenizer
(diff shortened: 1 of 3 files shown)


ogoffart:
I wouldn't change the QVERIFY to ASSERT. I think QVERIFY is fine there.
It is true that QVERIFY is not really meant to be used in a inner function, but it does work, as it will make the test fail if it is false.
Have you seen test succeeding despite QVERIFY was false?

Q_ASSERT is bad because the content of a Q_ASSERT is not supposed to have side effect (and many of these do). And also Q_ASSERT will not run the other tests.

So i'm not really in favor of this change.

TheOneRing:
No the tests failed but just continued to run, so my fault here I did not see the error location in the tons of log we create.

In general I'd expect a wrong api call here to fail with an assert, as an error in the fake server should not be part of the test itself.

Also see the comment by @guruz https://github.com/owncloud/client/pull/7819/files#diff-414178231286915abf220bbb1c7f89eaR26 

Also, we are using QVERIFY as well as Q_ASSERT in this code...

PR #7821, opened 2020-03-25
Add the missing documentation content to the PDF configuration file

settermjd:
This commit adds the remainder of the client documentation to the PDF configuration file, which were previously missing. I found this while investigating #7653.


TheOneRing:
If that is also missing in 2.6 please apply the patch against 2.6

settermjd:
> If that is also missing in 2.6 please apply the patch against 2.6

Shall do.

TheOneRing:
> > If that is also missing in 2.6 please apply the patch against 2.6
> 
> Shall do.

I meant, please apply this patch to 2.6 and not to master.....

Issue #7759, opened 2020-02-21
Unclear sync status after renaming a resource to match excluded regex


HanaGemela:
Client: 2.6.1 (build 13096)
Server: 10.4.0 RC4
macOS 10.15.3

Steps to recreate:
1) Add `'excluded_directories_regex' => [ '.*pdf' ],` in config.php on the server
2) Open Finder -> go to ownCloud sync folder -> create a new folder 
3) Rename the folder to e.g. `aaapdf`
4) Sync

Actual result: Folder has been renamed in Finder. Failed sync message shows the old name. Overlay icon is green - successfully synced. Selective sync list shows the old name. Confusing

<img width="979" alt="Screenshot 2020-02-21 at 15 52 39" src="https://user-images.githubusercontent.com/49001702/75045940-7adf5000-54c4-11ea-8ca5-33bc60ef3236.png">

When a file in added in such folder, nothing happens. No message, no overlay icon... 

Expected result: Either renamed everywhere and not synced, including red overlay icon or not renamed anywhere and still syncing, green overlay icon. 

There is already a separate issue for the broken overlay icons https://github.com/owncloud/client/issues/7703

michaelstingl:
@ogoffart anything we can improve here?

ogoffart:
It appears that the problem is that the destination of the move is not marked as a faillure as far as the socketapi is concerned.
That should be easy to fix

HanaGemela:
@ogoffart how exactly has this been fixed? What is the expected result?

Current behaviour: The overlay icon is red that's correct. But if I move the file into that folder, the file has a green icon and looks synced. But not synced in web ui

TheOneRing:
The bug is that the icon of that file is not updated.
If the page is refreshed the icon disappears as the whole folder is not tracked....

michaelstingl:
> If the page is refreshed

What exactly do you refresh? Finder view?

PR #7822, opened 2020-03-25
SocketAPI: fix status after a failure to move

ogoffart:
Issue #7759

### DIFF
--- a/src/libsync/syncfilestatustracker.cpp
+++ b/src/libsync/syncfilestatustracker.cpp
@@ -229,10 +229,10 @@ void SyncFileStatusTracker::slotAboutToPropagate(SyncFileItemVector &items)
         _dirtyPaths.remove(item->destination());
 
         if (hasErrorStatus(*item)) {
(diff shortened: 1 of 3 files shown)


src/libsync/syncfilestatustracker.cpp
```diff
@@ -229,10 +229,10 @@ void SyncFileStatusTracker::slotAboutToPropagate(SyncFileItemVector &items)
         _dirtyPaths.remove(item->destination());
 
         if (hasErrorStatus(*item)) {
-            _syncProblems[item->_file] = SyncFileStatus::StatusError;
+            _syncProblems[item->destination()] = SyncFileStatus::StatusError;
```
TheOneRing: Hm is that only an issue here or everywhere where _file is used?

ogoffart: _file is usually the same as destination(),  unless this is a move operation, then _file is the source.
But yes, in many case we want destination, while in some other case we want the source.  Perhaps having a source() function to make it explicit and avoid the use of _file should be done.

PR #7826, opened 2020-03-30
Restart mac client after update

TheOneRing:


TheOneRing:
See https://github.com/owncloud/enterprise/issues/3922

guruz:
Since macosx.pkgproj.cmake is a beast, could you please document in the oC handbook how you edited it :-)
Last time I tried some years ago the tools were messy

TheOneRing:
Well it is terribly messy.
I generated a pkgproj, opened it with the tool, and manually reapplied our cmake parameters....

TheOneRing:
@michaelstingl could you give https://jenkins.int.owncloud.com/job/client-osx/13435/ a try

admin/osx/post_install.sh.cmake
```diff
@@ -1,7 +1,9 @@
-#!/bin/sh
+#!/bin/bash
```
dschmidt, line 1: Even better usually is /usr/bin/env bash, but probably doesnt matter here.

admin/osx/post_install.sh.cmake
```diff
@@ -10,4 +12,10 @@ if [ -x "$(command -v pluginkit)" ]; then
     pluginkit -e use -i @APPLICATION_REV_DOMAIN@.FinderSyncExt
 fi
 
+if [[ -f ${INSTALLER_TEMP}/OC_RESTART_NEEDED ]]; then
```
dschmidt, line 15: quotes

PR #7828, opened 2020-03-30
Fix: Don't pass references to local variables to lambda

TheOneRing:


src/gui/notificationwidget.cpp
```diff
@@ -81,10 +81,10 @@ void NotificationWidget::setActivity(const Activity &activity)
             return;
         });
     } else {
-        foreach (auto link, activity._links) {
+        for (const auto &link : activity._links) {
             QPushButton *b = _ui._buttonBox->addButton(link._label, QDialogButtonBox::AcceptRole);
             b->setDefault(link._isPrimary);
-            connect(b, &QAbstractButton::clicked, this, [&]{
+            connect(b, &QAbstractButton::clicked, this, [&, b]{
```
ogoffart: actually, you should also remove the `&`  and take this explicitly by value.

```suggestion
            connect(b, &QAbstractButton::clicked, this, [this, b]{
```

PR #7831, opened 2020-03-31
Cookies: Don't override cookies with outdated values

TheOneRing:
This code was actually not breaking most cookie handling by accident.
As the raw cookies where not split properly we added cookies with values like
"key: val; key2 = val2; key3 = val3"
When the code was corrected we overwrote the newer values in the jar with
the old ones from a request.

### DIFF
--- a/changelog/unreleased/7831
+++ b/changelog/unreleased/7831
@@ -1,0 +1,6 @@
+Bugfix: Don't override cookies with old values
+
+We fixed a bug where a client somteimes overrode the content of the cookie jar
+with outdated or corrupted values
+
+https://github.com/owncloud/client/pull/7831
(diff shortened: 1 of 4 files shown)


ogoffart:
The code must have been like this for a reason... But i cannot remember why (maybe @guruz  knows)

TheOneRing:
The code was introduced in https://github.com/owncloud/client/commit/b037e6356e8053acb89c32037367c21b3a25aa49

And it can never have worked when more than one cookie was set.
And with that code fixed, it was possible that we received a new cookie in a reply, and overwrote it with the old one.

HanaGemela:
@TheOneRing  @guruz Please provide steps to test

TheOneRing:
We don't really have a test case for this besides running the client with any server.
As this seems to work and the patch fixed the client behaviour with a more complex cookie setup (f5) I expect it to work.

PR #7836, opened 2020-04-02
Mac Package: Kill the correct process

TheOneRing:


admin/osx/post_install.sh.cmake
```diff
@@ -1,4 +1,8 @@
 #!/bin/bash
+#exec   > ~/owncloud-install-post.log
+#exec  2>&1
+#BASH_XTRACEFD=1
+#set -x
```
dschmidt, line 5: what's this?

admin/osx/pre_install.sh.cmake
```diff
@@ -1,6 +1,12 @@
 #!/bin/bash
+#exec   > ~/owncloud-install-pre.log
+# exec  2>&1
+# BASH_XTRACEFD=1
+# set -x
 
-OC_INSTANCE=$(ps aux | grep "/Applications/@APPLICATION_EXECUTABLE@.app/Contents/MacOS/@APPLICATION_EXECUTABLE@")
+# don't grep in one line, to avaoid grepping the grep process...
+PROCESSES=$(ps aux)
+OC_INSTANCE=$(echo "${PROCESSES}" | grep "/Applications/@APPLICATION_EXECUTABLE@.app/Contents/MacOS/@APPLICATION_EXECUTABLE@")
```
dschmidt, line 9: i like this - my first attempt probably would have been to use `grep -v grep` breaking brandings containing grep in the name. ^_°

michaelstingl:
Helps for https://github.com/owncloud/enterprise/issues/3922

PR #7843, opened 2020-04-16
Tests: Fix rare crash in OAuthTest

TheOneRing:


TheOneRing:
Fixed in https://github.com/owncloud/client/commit/73bc702249c5c41c34f008935b49be31caadc830

PR #7855, opened 2020-05-05
Mac: Allow multiple instances so that we can detect a second invocation

TheOneRing:


guruz:
I recommend to have a ifdef for mac around here https://github.com/owncloud/client/blob/327fcf7a3133dd890bf34f5f2679624826701804/src/gui/owncloudgui.cpp#L1055

And there try to call:

[NSApp activateIgnoringOtherApps:YES];

(Which you can only do from a .mm file.. hmmm)

TheOneRing:
I'll look for an alternative solution as the side effects are to high

PR #7861, opened 2020-05-12
Fix order of pluginkit commands and modernise

TheOneRing:
Load the plugin file before we load the plugin by its id.


TheOneRing:
> I'm not against the change, but does this actually fix anything? Also: Because I see it adds more lines than it removes. Using system() in a platform-specific codepath is not worse than using QProcess.

If loading the file fails, for example the file does not exist, we now get a entry in the log.
I stumbled over this code when I was wondering why the integration was not working.

PR #7864, opened 2020-05-13
Uploads: correctly handle the bigfilechunking capability

TheOneRing:
The current implementation disables chunking v1 and uses the basic upload.
Is support for unchunked v1 support required?

### DIFF
--- a/changelog/unreleased/7862
+++ b/changelog/unreleased/7862
@@ -1,0 +1,5 @@
+Change: The client uploads chunks even though the server repports lack of support
+
+We now correctly handle the bigfilechunking capability
+
+https://github.com/owncloud/client/issues/7862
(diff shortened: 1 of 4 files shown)


PVince81:
> Is support for unchunked v1 support required?

there is no capability entry by design for a plain Webdav single-shot PUT upload, if that's what you mean
without Webdav PUT no uploads would be possible at all.

PVince81:
as far as I can understand the logic in this PR looks correct

changelog/unreleased/7862
```diff
@@ -0,0 +1,5 @@
+Change: The client uploads chunks even so the server repports lack of support
```
dschmidt, line 1: Do you mean `even though`? 

Also I honestly can't tell from this change what was the "previous" handling was and what happens "after" this change.

TheOneRing: thx

PR #7868, opened 2020-05-15
Mac: Display docker icon when settings are visible

TheOneRing:


guruz:
Nice!
I wanted to do this for a long time, but never did it :-)
Only downside I see is that people might not understand that there is a menu bar icon. (they probably didnt understand before either..)
I solved this in ocsmount by force-showing the menu after the configuration dialog is closed.

jnweiger:
Crude screenshots:
First with the settings window open, second with the settings window closed:





In both cases, we have a small icon in the top icon bar, and a large owncloud icon in the bottom dock panel.

jnweiger:
@TheOneRing @michaelstingl should the big owncloud icon in the dock disappear, when we close the settings window? On Linux it disappears.

TheOneRing:
It will disappear, but the dock displays the last few used applications when not more than n windows are open.

jnweiger:
That is more userfriendly than expected. OK.

Issue #7748, opened 2020-02-18
[VFS] Enabling native VFS on Windows create two shortcuts in Explorer's left pane


Zoddo:
### Expected behaviour
When enabling native VFS, a single shortcut per synced directory should be created in the left pane of the Windows Explorer.

### Actual behaviour
Two shortcuts are created: one pointing to the synced directory, and one pointing to the parent of the synced directory.

![image](https://user-images.githubusercontent.com/5215545/74717942-9252da80-5231-11ea-8806-53e76c35357c.png)
"1" is the synced directory  
"2" is the parent directory

### Steps to reproduce
1. Enable native VFS on a synced directory

### Server configuration
Operating system: Debian 10.3

Web server: Apache 2.4.38

Database: MariaDB 10.3.22

PHP version: 7.3.14

ownCloud version: 10.3.2

Storage backend (external storage): No external storage

### Client configuration
Client version: 2.6.1 (build 12992)

Operating system: Windows 10 1909 (build 18363.657)

OS language: French

Installation path of client: C:\Program Files (x86)\ownCloud

### Logs

1. Client logfile: https://gist.github.com/Zoddo/c131dcf1432beadb8030bf5f3f45602e (logs created when adding a new synced directory with native VFS enabled - redacted names are replaced by `<REDACTED>`)

2. Web server error log: N/A

3. Server logfile: ownCloud log (data/owncloud.log): N/A

TheOneRing:
I haven't seen this before. 
Are you using multiple folder sync connections?

lex2k0:
Today I updated to version 2.7.1 of the windows client and now I have two times the same ownCloud link in the explorer.
Maybe this comes due to a change in that issue?
![grafik](https://user-images.githubusercontent.com/12857178/99887476-04cf6b80-2c45-11eb-916f-6ea1455665c9.png)

PR #7906, opened 2020-06-18
VFS: Prevent duplicated navigation panel icon

TheOneRing:
Issue: #7748


TheOneRing:
# Old
image

# New
image

PR #7907, opened 2020-06-18
VFS: Tell the vfs plugin whether we have multiple accounts

TheOneRing:
This allows us to decide on the presentation of the account

### DIFF
--- a/src/common/vfs.h
+++ b/src/common/vfs.h
@@ -62,6 +62,11 @@ struct OCSYNC_EXPORT VfsSetupParams
     QString providerName;
     QString providerVersion;
 
+    /** when registering with the system we might use
+     *  a different presentaton to identify the accounts
+     */
+    bool multipleAccountsRegistered = false;
+
     /** Whether native shell integration shall be enabled
      *
      * For some plugins that doesn't work well in tests.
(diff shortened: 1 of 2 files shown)


TheOneRing:
Related: https://gitea.owncloud.services/client/client-plugin-vfs-win/pulls/9

src/common/vfs.h
```diff
@@ -62,6 +62,11 @@ struct OCSYNC_EXPORT VfsSetupParams
     QString providerName;
     QString providerVersion;
 
+    /** when registering with the system we might use
+     *  a different presentaton to identify the accounts
+     */
+    bool multipleAccountsKnow = false;
```
dschmidt, line 68: Strange var name

TheOneRing: multipleAccountsKnown... thats we I ask to review my code 🙈

TheOneRing: > 
> 
> This needs restarting after adding an account for the display name to update for the old account, right?

Usually only a restart of the client, as the registration is updated every time.

dschmidt: Yes, that's what I meant

Issue #7888, opened 2020-06-05
2.7 client switches from oidc to oauth, when re-opening the browser


jnweiger:
When pointing a 2.7daily client to connect to an instance with openid connect, the `/signin/v1/identifier/` route is correctly used:
 
![image](https://user-images.githubusercontent.com/1108546/83864864-a6114680-a725-11ea-8629-bdef5140db23.png)

For these screenshots I have instrumented xdg-open to intercept and display the URL that we pass to `QDesktopServices::openUrl(url)` 

The client GUI offers a 'Click here to re-open the browser'--  very helpful, in case the user needs to restart, e.g. with a fresh browser.
Clicking there, I get

![image](https://user-images.githubusercontent.com/1108546/83865671-d1e0fc00-a726-11ea-9de8-967072ce5f01.png)

This time the route `/index.php/apps/oauth2/authorize` is taken, and the browser reacted with an OAuth attempt.
Expected behavior: Same route as before, using oidc every time.

PR #7908, opened 2020-06-19
OAuth: Fix reopening the browser from the account page

TheOneRing:
Issue: #7888


src/libsync/creds/httpcredentials.h
```diff
@@ -171,6 +171,8 @@ private Q_SLOTS:
     bool _keychainMigration = false;
     bool _retryOnKeyChainError = true; // true if we haven't done yet any reading from keychain
 
+    bool _useOauth = false;
```
dschmidt, line 174: I don't understand how this var is used.

When is it true, when false?

TheOneRing: Sorry that was bullshit, it should be more sane now.

PR #7918, opened 2020-06-25
Compile resources in the exe not the core lib

TheOneRing:


dschmidt:
I don't care too much if we link the resources to the executable or if we initialize them explicitly in the static lib.
Personally I would choose the latter, because when they are used in the lib, they should be available in the lib ... but in the end, I don't care too much.

FTR https://doc.qt.io/qt-5/resources.html#using-resources-in-a-library

Issue #7920, opened 2020-06-25
Layout issue in the advanced account setup page


TheOneRing:
<img width="558" alt="Screenshot 2020-06-20 at 11 15 33" src="https://user-images.githubusercontent.com/200626/85713054-40dbc000-b6e9-11ea-93d9-c0549d90a883.png">

PR #7921, opened 2020-06-25
Wizard: Improve layout

TheOneRing:
Instead of using a fixed size grid layout use a dynamic layout.

Fixes: #7920


changelog/unreleased/7920
```diff
@@ -0,0 +1,5 @@
+Change: The layout in the advannced account setup page might be broken
```
dschmidt, line 1: advanced

PR #7922, opened 2020-06-25
Wizard: Minimize the Window instead of hiding it

TheOneRing:


dschmidt:
Does this behave well on all platforms?

TheOneRing:
Seems good on Windows and Mac

jnweiger:
Also looks fine on Linux.

PR #7923, opened 2020-06-25
Remove ifdef(Q_OS_X11) code, there never was such a define.

TheOneRing:
As noone complained since I don't think anyone will miss the code


TheOneRing:
Merged with https://github.com/owncloud/client/pull/7924

PR #7924, opened 2020-06-25
Wizard: Make the wizard a child dialog of the settings

TheOneRing:


TheOneRing:
image

TheOneRing:
image

TheOneRing:
I changed the default size of the dialog, one issue is that the client saves the size so most users will keep the small one.
We could set a minimum size, the current value of 800x500 should fit on all screens.
Also I guess the Wizard should get a cancel button as currently the is no indication on how to abort (Esc).

michaelstingl:
Yeah, cancel makes sense.

TheOneRing:
image

TheOneRing:
Or the cancel button on the left?
image

michaelstingl:
Yeah, cancel better left 👍

dschmidt:
I agree, the outer right button should ne next, so cancel left.

TheOneRing:
Done ready for review

PR #7925, opened 2020-06-26
Gui: hide the selective sync pop up by default

TheOneRing:
This fixes a strange flickering when the client is started with the gui visible.

### DIFF
--- a/changelog/unreleased/7925
+++ b/changelog/unreleased/7925
@@ -1,0 +1,5 @@
+Change: The settings ui shows a prompt for a few seconds
+
+We now hide that prompt by default and only show it if needed.
+
+https://github.com/owncloud/client/pull/7925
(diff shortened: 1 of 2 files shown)


changelog/unreleased/7925
```diff
@@ -0,0 +1,5 @@
+Change: The settings ui shows a promt for a prompt for a few seconds
```
dschmidt, line 1: I don't understand this sentence

jnweiger:
cannot conirm fix. changelog message is unclear, reproducer is unclear. Sorry.

TheOneRing:
The widget for selective sync, (sync folders bigger than) in the lower pane was visible when starting the application with the -s for a second until we had a defined state.
Flashing widgets are not cool

jnweiger:
## confirmed in testpilotcloud version 2.7.0daily20201019
started with -s, connecting to demo.owncloud.org
while the setup wizard is shown, nothing flashes. 

(cannot reproduce)

PR #7931, opened 2020-06-29
Make the password dialog modal

TheOneRing:


TheOneRing:
https://github.com/owncloud/client/pull/7924 needs review first

src/gui/creds/httpcredentialsgui.cpp
```diff
@@ -113,7 +117,7 @@ void HttpCredentialsGui::showDialog()
             + QLatin1String("<br>");
     }
 
-    QInputDialog *dialog = new QInputDialog();
+    QInputDialog *dialog = new QInputDialog(qobject_cast<Application *>(qApp)->gui()->settingsDialog());
```
dschmidt, line 120: Do we need the cast here because qApp returns a QCoreApplication?
Maybe use QApplication::instance() instead?

TheOneRing: Its QApplication vs Application....

PR #7942, opened 2020-07-03
Get rid of gridlayout

TheOneRing:
The first few changes include src changes so they should be reviewed with a bit more care.
I know that is impossible to review ui files

### DIFF
--- a/src/gui/sharedialog.cpp
+++ b/src/gui/sharedialog.cpp
@@ -29,6 +29,7 @@
 #include <QPointer>
 #include <QPushButton>
 #include <QFrame>
+#include <QRegularExpression>
 
 namespace OCC {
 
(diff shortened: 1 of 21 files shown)


src/gui/sharedialog.cpp
```diff
@@ -91,19 +92,16 @@ ShareDialog::ShareDialog(QPointer<AccountState> accountState,
     QString ocDir(_sharePath);
     ocDir.truncate(ocDir.length() - fileName.length());
 
-    ocDir.replace(QRegExp("^/*"), "");
-    ocDir.replace(QRegExp("/*$"), "");
+    // reove leading and trailing spaces
```
dschmidt, line 95:
```suggestion
    // remove leading and trailing spaces
```

PR #7948, opened 2020-07-08
Watcher

TheOneRing:


src/gui/folderwatcher_win.cpp
```diff
@@ -97,47 +97,47 @@ void WatcherThread::watchChanges(size_t fileNotifyBufferSize,
...
+            const int len = curEntry->FileNameLength / sizeof(wchar_t);
+            QString longfile = longPath + QString::fromWCharArray(curEntry->FileName, len);
 
             // Unless the file was removed or renamed, get its full long name
             // TODO: We could still try expanding the path in the tricky cases...
-            QString longfile = file;
             if (curEntry->Action != FILE_ACTION_REMOVED
                 && curEntry->Action != FILE_ACTION_RENAMED_OLD_NAME) {
-                size_t longNameSize = GetLongPathNameW(reinterpret_cast<LPCWSTR>(file.utf16()), fileNameBuffer, fileNameBufferSize);
+                const auto wfile = longfile.toStdWString();
```
TheOneRing, line 127: yes toStdWString creates a copy and as the code is windows only the result is supposed to be the same, but ....

src/gui/folderwatcher_win.cpp
```diff
@@ -97,47 +97,56 @@ void WatcherThread::watchChanges(size_t fileNotifyBufferSize,
...
             }
+
+#if QT_VERSION < QT_VERSION_CHECK(5, 14, 0)
+            // The prefix is needed for native Windows functions before Windows 10, version 1607
+            const bool hasLongPathPrefix = longPath.startsWith(QStringLiteral("\\\\?\\"));
+            if (hasLongPathPrefix)
+            {
+                longfile.remove(0, 4);
+            }
+#endif
```
dschmidt, line 143: Any more info on this?

src/gui/folderwatcher_win.cpp
```diff
@@ -62,25 +63,24 @@ void WatcherThread::watchChanges(size_t fileNotifyBufferSize,
         ResetEvent(_resultEvent);
 
         FILE_NOTIFY_INFORMATION *pFileNotifyBuffer =
-            (FILE_NOTIFY_INFORMATION *)fileNotifyBuffer.data();
+                reinterpret_cast<FILE_NOTIFY_INFORMATION *>(fileNotifyBuffer.data());
         DWORD dwBytesReturned = 0;
-        SecureZeroMemory(pFileNotifyBuffer, fileNotifyBufferSize);
```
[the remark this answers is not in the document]
TheOneRing: As far as I can see we always know the size of the file entries ReadDirectoryChangesW succeeds we always should have well defined memory?

src/gui/folderwatcher_win.cpp
TheOneRing: I changed the longFile name to start with longPath which is a windows file path including the \\\\?\\ prefix which was needed on windows <10 to allow paths longer than 260 chars.

dschmidt: I rather meant the qt_version_check, what changed when and how etc. If you know more, could be added as comments.

TheOneRing: I had to dive into the code to find out why the unit tests where failing and why it was working with my Qt 5.15 setup and not on the 5.12.9 ci...

TheOneRing: https://github.com/qt/qtbase/commit/3966b571ff54100923a479050a38ac906e0ed1e3

PR #7950, opened 2020-07-10
CMake: Make VIRTUAL_FILE_SYSTEM_PLUGINS accessible to cmake tools

TheOneRing:


TheOneRing:
> When adding external plugins, you need to pass "suffix win /some/path" right? That's a bit unusual I'd say...

Well I added that bit last year as I prefer it over cloning into a sub directory of a git repo.
The option is in general only relevant for internal devs, but it would make live easier for me to configure it that way :D

PR #7954, opened 2020-07-14
Fix warning

TheOneRing:
warning C4573: the usage of 'QObject::disconnect' requires the compiler to capture 'this' but the current default capture mode does not allow it

### DIFF
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -777,8 +777,8 @@ void Application::openVirtualFile(const QString &filename)
     folder->implicitlyHydrateFile(relativePath);
     QString normalName = filename.left(filename.size() - virtualFileExt.size());
     auto con = QSharedPointer<QMetaObject::Connection>::create();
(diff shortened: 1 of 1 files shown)


src/gui/application.cpp
```diff
@@ -777,8 +777,8 @@ void Application::openVirtualFile(const QString &filename)
     folder->implicitlyHydrateFile(relativePath);
     QString normalName = filename.left(filename.size() - virtualFileExt.size());
     auto con = QSharedPointer<QMetaObject::Connection>::create();
-    *con = QObject::connect(folder, &Folder::syncFinished, [con, normalName] {
-        QObject::disconnect(*con);
+    *con = connect(folder, &Folder::syncFinished, folder, [folder, con, normalName] {
```
dschmidt, line 780: is the folder as target object really needed?

capturing it should be enough, no?

TheOneRing: yes but if I capture it and don't use it the compiler will complain about an unused capture...

PR #7962, opened 2020-07-15
Gui: Rename "General" -> "Settings" and move "Network" -> "Settings"

TheOneRing:
image

### DIFF
--- a/changelog/unreleased/7962
+++ b/changelog/unreleased/7962
@@ -1,0 +1,5 @@
+Change: Reorgenize Settings
+
+We rename "General" to "Settings" and move the "Network" into "Settings"
+
+https://github.com/owncloud/client/pull/7962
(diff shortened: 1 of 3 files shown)


jnweiger:
That is a good improvement! I like it! :+1:

Next improvement idea for that topic: Allow different settings per account?

dschmidt:
The scroll area is just to make this fit on a small screen, right?

Is it neccessary to scroll on a big enough screen?

dschmidt:
Screenshot_20200720_133552

to be honest, I think this is too big and that the version info at the bottom being fixed makes it overall a bit awkward to use.

TheOneRing:
Right now the window is not full screen but 800x500 so the scroll area is needed.
The version stuff is important and should always be visible.
With the scroll area the network stuff is partly hidden but I'd expect that most user will never interact with them at all.

TheOneRing:
Les see how it feels, we can always change it.

PR #7964, opened 2020-07-15
Remove legacy C code

TheOneRing:


guruz:
I'm quite sure there was a reason behind this :-)

Let's not merge before @ogoffart confirms

TheOneRing:
The last remaning stuff can probably only be killed once we can use std::path wich requires c++17 and mac 10.15

test/csync/vio_tests/check_vio_ext.cpp
```diff
@@ -125,8 +101,7 @@ static int teardown(void **state) {
     rc = wipe_testdir();
     assert_int_equal(rc, 0);
 
-    SAFE_FREE(((statevar*)*state)->result);
-    SAFE_FREE(*state);
+    delete reinterpret_cast<statevar*>(*state);
```
dschmidt, line 104: result doesn't need to be freed anymore?

TheOneRing: I now use new and delete instead of malloc

PR #7977, opened 2020-07-22
CI:  Remove nsis translation job

TheOneRing:


dschmidt:
typo in the commit message:
s/where/were/

In general I find the commit message rather suboptimal. It describes a pre-condition of the commit, not what the commit actually does.

What do you think of "CI: Remove nsis translation job

as NSIS scripts where removed before already."?

PR #7989, opened 2020-07-24
Use the servers checksum type by default

TheOneRing:
Local computation of sha1 checksum
Comparison with the checksum the remote provided after upload (sha256), therefore re computation of local checksum with sha256. 

Related: https://gitea.owncloud.services/client/client-plugin-vfs-win/pulls/11

### DIFF
--- a/src/libsync/propagatedownload.cpp
+++ b/src/libsync/propagatedownload.cpp
@@ -847,7 +847,7 @@ namespace { // Anonymous namespace for the recall feature
 
 void PropagateDownloadFile::transmissionChecksumValidated(const QByteArray &checksumType, const QByteArray &checksum)
 {
-    const auto theContentChecksumType = contentChecksumType();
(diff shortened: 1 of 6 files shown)


src/libsync/propagatedownload.cpp
```diff
@@ -847,7 +847,7 @@ namespace { // Anonymous namespace for the recall feature
 
 void PropagateDownloadFile::transmissionChecksumValidated(const QByteArray &checksumType, const QByteArray &checksum)
 {
-    const auto theContentChecksumType = contentChecksumType();
+    const QByteArray theContentChecksumType = propagator()->account()->capabilities().preferredUploadChecksumType();
```
dschmidt, line 850: Why did you change this to an explicit type?

TheOneRing: I love auto but sometimes it helps to see what type you are using :D

PR #7991, opened 2020-07-27
Use strict QString handling

TheOneRing:


src/common/syncjournaldb.cpp
```diff
@@ -107,13 +107,13 @@ QString SyncJournalDb::makeDbName(const QString &localPath,
...
-    QString journalPath = QLatin1String(".sync_");
+    QString journalPath = QStringLiteral(".sync_");
 
-    QString key = QString::fromUtf8("%1@%2:%3").arg(user, remoteUrl.toString(), remotePath);
+    QString key = QStringLiteral("%1@%2:%3").arg(user, remoteUrl.toString(), remotePath);
 
     QByteArray ba = QCryptographicHash::hash(key.toUtf8(), QCryptographicHash::Md5);
-    journalPath.append(ba.left(6).toHex());
-    journalPath.append(".db");
+    journalPath.append(QString::fromUtf8(ba.left(6).toHex()));
```
dschmidt, line 115: hex values are ascii, so you could use fromLatin1

src/common/syncjournaldb.cpp
```diff
@@ -950,6 +950,7 @@ bool SyncJournalDb::setFileRecord(const SyncJournalFileRecord &_record)
     }
 }
 
+// TODO: filename -> QBytearray?
```
dschmidt, line 953: Why?

src/common/utility.cpp
```diff
@@ -555,22 +557,22 @@ QString Utility::makeConflictFileName(
...
+    if (dotLocation <= conflictFileName.lastIndexOf(QLatin1Char('/')) + 1) {
         dotLocation = conflictFileName.size();
     }
 
     QString conflictMarker = QStringLiteral(" (conflicted copy ");
     if (!user.isEmpty()) {
         // Don't allow parens in the user name, to ensure
         // we can find the beginning and end of the conflict tag.
-        const auto userName = sanitizeForFileName(user).replace('(', '_').replace(')', '_');
+        const auto userName = sanitizeForFileName(user).replace(QRegularExpression(QStringLiteral("\\(|\\)")), QStringLiteral("_"));
```
dschmidt, line 570: Not sure this is more readable ...

if this is supposed to be a performance improvement, have you benchmarked it?

src/common/syncjournaldb.cpp
TheOneRing: Looks like its basically always used as a byte array and converted back and forth but it would have bloated this review....

PR #7992, opened 2020-07-28
Fix gnome shell crash by minimizing and showing the main window

TheOneRing:


src/gui/application.h
```diff
@@ -148,6 +148,13 @@ protected slots:
     QScopedPointer<FolderMan> _folderManager;
 };
 
+inline Application *ocApp()
+{
+    auto instance = qobject_cast<Application *>(qApp);
+    OC_ENFORCE(instance);
```
dschmidt, line 154: <3

src/gui/owncloudgui.cpp
```diff
@@ -1055,11 +1055,16 @@ void ownCloudGui::slotHelp()
 
 void ownCloudGui::raiseDialog(QWidget *raiseWidget)
 {
-    auto window = raiseWidget->window();
-    Q_ASSERT(window);
+    auto window = ocApp()->gui()->settingsDialog();
+    OC_ASSERT(window);
```
dschmidt, line 1059: :+1:

PR #7994, opened 2020-07-29
Gracefully handle missing sync root

TheOneRing:


src/gui/accountmanager.h
```diff
@@ -58,7 +58,7 @@ class AccountManager : public QObject
      * Return a list of all accounts.
      * (this is a list of QSharedPointer for internal reasons, one should normally not keep a copy of them)
      */
-    QList<AccountStatePtr> accounts() { return _accounts; }
+    const QList<AccountStatePtr> &accounts() { return _accounts; }
```
dschmidt, line 61: This is only ever used from a single thread, right?

src/gui/application.cpp
```diff
@@ -360,6 +354,14 @@ Application::Application(int &argc, char **argv)
...
     // Cleanup at Quit.
     connect(this, &QCoreApplication::aboutToQuit, this, &Application::slotCleanup);
+
+    QTimer::singleShot(0, this, [this] {
+        FolderMan::instance()->setupFolders();
+        _proxy.setupQtProxyFromConfig(); // folders have to be defined first, than we set up the Qt proxy.
+        for (const auto &ai : AccountManager::instance()->accounts()) {
+            slotAccountStateAdded(ai.data());
+        }
+    });
```
dschmidt, line 364: Why are things initialized through the event loop now? 
Maybe worth a comment...

src/gui/folder.cpp
```diff
@@ -166,18 +165,42 @@ void Folder::checkLocalPath()
...
         if (!FileSystem::fileExists(_definition.localPath, fi)) {
-            _syncResult.appendErrorString(tr("Local folder %1 does not exist.").arg(_definition.localPath));
-            _syncResult.setStatus(SyncResult::SetupError);
+            if (ocApp()) {
+                if (FileSystem::fileExists(fi.path() + "/..")) {
+                    int button = QMessageBox::question(ocApp()->gui()->settingsDialog(),
+                        tr("Your %1 folder %2 does not exist.").arg(ocApp()->applicationName(), _definition.localPath),
+                        tr("The folder %1 does not exist, do you want to create it or quit %2.").arg(_definition.localPath, ocApp()->applicationName()),
+                        tr("Create Folder"), tr("Quit"));
+                    if (button == 0) {
```
dschmidt, line 178:
```suggestion
                    if (button == QMessageBox::NoButton) {
```

src/gui/folder.cpp
```diff
@@ -166,18 +165,42 @@ void Folder::checkLocalPath()
...
         qCDebug(lcFolder) << "Checked local path ok";
+        return true;
     } else {
         // Check directory again
         if (!FileSystem::fileExists(_definition.localPath, fi)) {
-            _syncResult.appendErrorString(tr("Local folder %1 does not exist.").arg(_definition.localPath));
-            _syncResult.setStatus(SyncResult::SetupError);
+            if (ocApp()) {
+                if (FileSystem::fileExists(fi.path() + "/..")) {
+                    int button = QMessageBox::question(ocApp()->gui()->settingsDialog(),
```
dschmidt, line 174:
```suggestion
                    auto button = QMessageBox::question(ocApp()->gui()->settingsDialog(),
```
or
```suggestion
                    QMessageBox::StandardButton button = QMessageBox::question(ocApp()->gui()->settingsDialog(),
```

src/gui/folder.cpp
```diff
@@ -166,18 +165,42 @@ void Folder::checkLocalPath()
...
+                        QThread::sleep(1);
+                        return checkLocalPath();
+                    }
+                } else {
+                    QMessageBox::warning(ocApp()->gui()->settingsDialog(),
+                        tr("Your %1 folder %2 does not exist.").arg(ocApp()->applicationName(), _definition.localPath),
+                        tr("The folder %1 and its parent folder do not exist. %2 will quit now.").arg(_definition.localPath, ocApp()->applicationName()));
+                }
+                _syncResult.setStatus(SyncResult::SetupError);
+                qApp->quit();
```
dschmidt, line 192: Do we need to quit here?

Can't we disable the account or something?

This way you can't remove an offending account if you can't recreate the folder on disk?

src/gui/folder.cpp
TheOneRing: Hm not really in that case, https://code.woboq.org/qt5/qtbase/src/widgets/dialogs/qmessagebox.cpp.html#2250
I don't think switching the non obsolete question box here would be a good idea as a question that can answered with a yes or no here would be quite complex.

src/gui/folder.cpp
TheOneRing: We ignored the folder before and it went magically back online once a folder appeared. https://github.com/owncloud/enterprise/issues/4048
And in general I guess most people will only have one account.
That you can't remove the account yes that needs to be handled....

src/gui/accountmanager.h
TheOneRing: I seriously hope so....

src/gui/application.cpp
TheOneRing: Doing complex initialization in the constructor of qApp.....

Also this makes sure that qApp/ocApp is available one the folders get initialized.

TheOneRing: Maybe I should move that change in a separate review and also initialise the ui delayed.

src/gui/folder.cpp
dschmidt: Didn't see that in the docs ... hrm... .....

src/gui/folder.cpp
```diff
@@ -166,18 +165,47 @@ void Folder::checkLocalPath()
...
     } else {
         // Check directory again
         if (!FileSystem::fileExists(_definition.localPath, fi)) {
-            _syncResult.appendErrorString(tr("Local folder %1 does not exist.").arg(_definition.localPath));
-            _syncResult.setStatus(SyncResult::SetupError);
+            if (ocApp()) {
+                if (FileSystem::fileExists(fi.path() + "/..")) {
+                    const int button = QMessageBox::question(ocApp()->gui()->settingsDialog(),
+                        tr("Your %1 sync folder %2 does no longer exist").arg(ocApp()->applicationName(), _accountState->account()->displayName()),
+                        tr("The %1 sync folder %2 for the account %3 does not exist.\nDo you want to create it, remove the account %2 or quit.")
```
dragotin, line 176: This is a question ("Do you want to...") and should end with a question mark.

src/gui/folder.cpp
dragotin: I am not quite getting why you remove the account at all, instead of just removing the sync folder. But I am sure you thought about that... ;-)

src/gui/folder.cpp
```diff
@@ -166,18 +165,47 @@ void Folder::checkLocalPath()
...
+        return true;
     } else {
         // Check directory again
         if (!FileSystem::fileExists(_definition.localPath, fi)) {
-            _syncResult.appendErrorString(tr("Local folder %1 does not exist.").arg(_definition.localPath));
-            _syncResult.setStatus(SyncResult::SetupError);
+            if (ocApp()) {
+                if (FileSystem::fileExists(fi.path() + "/..")) {
+                    const int button = QMessageBox::question(ocApp()->gui()->settingsDialog(),
+                        tr("Your %1 sync folder %2 does no longer exist").arg(ocApp()->applicationName(), _accountState->account()->displayName()),
```
dragotin, line 175: Wording - for the header, I would short it: "%1 sync folder %2 does not exist."

michaelstingl:
Maybe not only deleted sync dirs are the problems. I assume users also move or rename their sync dir. Instead of the "Create" button, we could also show "Select Sync Folder(on Disk)" to let the users find/re-connect a moved sync dir…

PR #8009, opened 2020-08-10
Remove leftover from shiboleth support from the cmd client

TheOneRing:


src/libsync/theme.h
```diff
@@ -256,7 +256,6 @@ class OWNCLOUDSYNC_EXPORT Theme : public QObject
      * it has a trailing slash, for example "remote.php/webdav/".
      */
     virtual QString webDavPath() const;
-    virtual QString webDavPathNonShib() const;
```
TheOneRing, line 259: we might need to touch ownbrander.

dschmidt: https://github.com/owncloud/ownbrander/search?q=webDavPathNonShib&unscoped_q=webDavPathNonShib

```We couldn’t find any code matching 'webDavPathNonShib' in owncloud/ownbrander```

PR #8015, opened 2020-08-14
Improve checksum handling

TheOneRing:


TheOneRing:
Actual fix in https://gitea.owncloud.services/client/client-plugin-vfs-win/pulls/13

PR #8018, opened 2020-08-17
Advances Wizard: Move choose what to sync to Sync everything

TheOneRing:
# Old:
image

# New:
image

TODO: add changelog entry


jnweiger:
Not sure, if 'Choose what to sync' belongs inside 'Synchronize everything'

for me the logic is, When I decide to 'Sync everything' I dont need to think about what to sync any more. It is everything.
So "choose what to sync" should be an alternative to 'sync everything'

But YMMV, I am no UX expert. As long as the element is there. I am fine.

PR #8025, opened 2020-08-19
Move sync hidden files to advanced settings

TheOneRing:


dschmidt (approved):
Hard to review but nothing caught my attention especially ..

Issue #8021, opened 2020-08-17
[UI] Add option to upload conflict files to the "Advanced" Settings


michaelstingl:
![Capto_Capture 2020-08-17_07-19-38_PM copy](https://user-images.githubusercontent.com/214010/90425967-e0517980-e0c0-11ea-8654-1878160e28c4.png)

I'd like to try "default=on", at least in the 2.7.0-beta3.

We already have an Environment Variable for it:
`OWNCLOUD_UPLOAD_CONFLICT_FILES`
(Set to "1" to enable uploading conflict files to the server.)

TheOneRing:
```
bool Capabilities::uploadConflictFiles() const
{
    static auto envIsSet = !qEnvironmentVariableIsEmpty("OWNCLOUD_UPLOAD_CONFLICT_FILES");
    static int envValue = qEnvironmentVariableIntValue("OWNCLOUD_UPLOAD_CONFLICT_FILES");
    if (envIsSet)
        return envValue != 0;

    return _capabilities[QStringLiteral("uploadConflictFiles")].toBool();
}```
Seems to be a server setting, I guess we should coordinate with the server team?

michaelstingl:
> Seems to be a server setting, I guess we should coordinate with the server team?

Should be a server setting, but was never implemented.

TheOneRing:
I see

TheOneRing:
TODO: Investigate conflict file logic.

glpatcern:
Hello, may I ask what is the status of this feature, now that the sync client is v4+?

michaelstingl:
> Hello, may I ask what is the status of this feature, now that the sync client is v4+?

The latest implementation didn't work, so it was dropped:
- https://github.com/owncloud/client/pull/8986#issuecomment-953708676

glpatcern:
> The latest implementation didn't work, so it was dropped:

Thanks Michael for the quick reaction. That implementation was from 2 years ago, what about current plans?

michaelstingl:
No plans currently

PR #8028, opened 2020-08-19
Add upload conflict file to the settings and enable it by default

TheOneRing:
Fixes: #8021
image


TheOneRing:
Well, looks like we have a problem.
As far as I can tell the unit tests use the standard config ....

TheOneRing:
Depends on https://github.com/owncloud/client/issues/6213

TheOneRing:
See https://github.com/owncloud/client/pull/8986

PR #8029, opened 2020-08-19
Allow to directly use svg instead of export

TheOneRing:


src/libsync/owncloudtheme.cpp
```diff
@@ -53,7 +53,7 @@ QColor ownCloudTheme::wizardHeaderTitleColor() const
 
 QPixmap ownCloudTheme::wizardHeaderLogo() const
 {
-    return QPixmap(hidpiFileName(":/client/theme/colored/wizard_logo.png"));
+    return QIcon(QStringLiteral(":/client/theme/colored/wizard_logo.png")).pixmap({132, 63});
```
dschmidt, line 56: As discussed on RocketChat:

We need to change ownBrander for this anyway and so while at it, we want to make this function return a QIcon directly, so the resolution can be hardcoded wherever it's needed where the pixmap is actually consumed.

dschmidt (changes requested):

src/libsync/theme.cpp
```diff
@@ -430,28 +430,19 @@ QColor Theme::wizardHeaderBackgroundColor() const
...
-        // Indeed, this size need to be big enough to for the banner height, and the wizard's width
-        auto ratio = screen->logicalDotsPerInch() / 96.;
-        if (ratio > 1.)
-            size *= ratio;
-    }
-    QPixmap pix(size);
-    pix.fill(wizardHeaderBackgroundColor());
-    return pix;
+        return QIcon();
+    QPixmap pix(QSize{750, 78} * 3);
```
dschmidt, line 443: This function is used in exactly one location afaict, shouldn't we move this pixmap conversion code there too? 
Then we can keep the size hardcoding out of this place just like for the other wizardHeaderLogo.

dschmidt: Or pass in the size at least. 

Anyway: what is the `*3` good for?

TheOneRing: we request the actual size there, the image it self is a solid color, I compute a bigger one so that QIcon::pixmap is able to compute a hdpi version, the *3 is just to have a litle extra.

TheOneRing:
image
there are issues ...

TheOneRing:
> 
> 
> image
> there are issues ...

That doesn't make any sense.
This is caused by the *3 .

From the doc for ::pixmap()
> Returns a pixmap of size QSize(extent, extent). The pixmap might be smaller than requested, but never larger.

> Setting the Qt::AA_UseHighDpiPixmaps application attribute enables this function to return pixmaps that are larger than the requested size. Such images will have a devicePixelRatio larger than 1.

src/gui/guiutility.cpp
```diff
@@ -98,16 +99,36 @@ QString Utility::vfsFreeSpaceActionText()
...
+    OC_ASSERT(!icon.isNull());
+    auto availableSize = icon.availableSizes();
+    if (availableSize.empty())
+    {
+        availableSize = {{48, 48}, {64, 64}, {128, 128}, {256, 256}};
+    }
+    QImage img = icon.pixmap(availableSize.first()).toImage();
+    if (!img.isGrayscale()) {
+        return icon;
+    } else {
```
TheOneRing, line 114: I guess we should just add a second set of colored svgs in to the pot and use them instead of this madness.

TheOneRing:
Superseded by https://github.com/owncloud/client/pull/8038

PR #8030, opened 2020-08-19
Move the management of the account wizard to ownCloudGui

TheOneRing:


src/gui/owncloudgui.cpp
```diff
@@ -124,26 +124,16 @@ void ownCloudGui::slotTrayClicked(QSystemTrayIcon::ActivationReason reason)
 
     // Left click
     if (reason == QSystemTrayIcon::Trigger) {
-        if (OwncloudSetupWizard::bringWizardToFrontIfVisible()) {
```
TheOneRing, line 127: as we only have modal dialogs for those now, we don't need this hackery

src/gui/accountsettings.cpp
```diff
@@ -227,7 +227,7 @@ QString AccountSettings::selectedFolderAlias() const
 
 void AccountSettings::slotOpenAccountWizard()
 {
-    OwncloudSetupWizard::runWizard(qApp,  SLOT(slotownCloudWizardDone(int)), this);
+    ocApp()->gui()->runNewAccountWizard();
```
dschmidt, line 230: Not a fan of getter chains and especially not on singletons - but this PR gets rid of another singleton, so I guess it's still an improvement :-D

dschmidt (approved):

PR #8031, opened 2020-08-19
Add add account button

TheOneRing:
image


michaelstingl (approved):
Nice one 👍

guruz:
Nice!
I always hated that this functionality was hidden in the menu for an existing account where it logically didnt belong it all.

guruz:
Kind of related: https://github.com/owncloud/client/issues/6454

TheOneRing:
> 
> 
> Kind of related: #6454

Hmm I'm more in favor of a complete rethink, activity is account specific so it should be per account and not global. At least something like that, but I guess redoing the ui would require multiple long meetings :D

Issue #7967, opened 2020-07-16
Crash when moving file within shared folder


m0ritz3:
### Expected behaviour
I move a file with a shared folder and the shared folder is updated.

### Actual behaviour
The file is moved. Owncloud shows the error symbol with a message "MOVE: file move_location/file

### Steps to reproduce
1. Move file within shared folder.

ownCloud version 2.6.3 (build 2668)
Git revision f28dd972e941986edc3e1803a25cf067f520b7e3
Using Qt 5.12.7, built against Qt 5.12.7
Using 'OpenSSL 1.1.1  11 Sep 2018'

Operating system: Ubuntu 18.04.0 LTS

OS language: English

Terminal output on crash:

07-16 16:35:31:557 [ fatal default ]:	ENFORCE: "instruction == CSYNC_INSTRUCTION_REMOVE || ((*it)->_type == ItemTypeVirtualFile && instruction == CSYNC_INSTRUCTION_NEW) || ((*it)->_isRestoration && instruction == CSYNC_INSTRUCTION_NEW)" in file /usr/src/packages/BUILD/src/libsync/discoveryphase.cpp, line 175

For additional information please tell me how I can provide them.

michaelstingl:
@m0ritz3 could you post the ID from the crash report window?

m0ritz3:
@michaelstingl There was no crash report window after restarting owncloud. Or do I have to eplicitly report the crash? If yes, how can I do that?

I would like to share the log file, but there is a lot of sensitive information. Is there a way to easily anonymize the log file?

TheOneRing:
`
07-17 13:27:34:051 [ fatal default ]:	ENFORCE: "instruction == CSYNC_INSTRUCTION_REMOVE || ((*it)->_type == ItemTypeVirtualFile && instruction == CSYNC_INSTRUCTION_NEW) || ((*it)->_isRestoration && instruction == CSYNC_INSTRUCTION_NEW)" in file /usr/src/packages/BUILD/src/libsync/discoveryphase.cpp, line 175
`
I hate that crash

TheOneRing:
Hi are you able to reproduce the bug?
As the crash is bugging me, and the enforce is quite complex I added an extra level of verbosity. https://github.com/owncloud/client/commit/b8a4d0c3ff72b075609fe9fe20d1fe19d9c01535
So from tomorrow on the client should tell me more on what is happening here.
[...]

PR #8033, opened 2020-08-20
Don't crash if a delete job marked as ignore is cancelled

TheOneRing:
Issue: #7967

The enforce was introduced in https://github.com/owncloud/client/commit/4d9748080958aaee55810f52a3e14dda024dffa1

### DIFF
--- a/changelog/unreleased/7967
+++ b/changelog/unreleased/7967
@@ -1,0 +1,6 @@
+Bugfix: Don't crash if move is undone due to permissions
+
+We fixed a crash when a file in a folder with certain permissions
+was restored.
+
+https://github.com/owncloud/client/issues/7967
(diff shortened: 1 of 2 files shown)


dschmidt (commented):
Better let someone else comment on this, no idea about this.

ogoffart:
It should probably be possible to write a test for that and I highly recommend doing that to avoid regressions.
I guess the patch is correct although I don't explain how the ignore item gets in the delete items list.

TheOneRing:
Yes we should probably write a test for that, but I was not able to trick the test system into those conditions.
The actual scenario is 
Clipboard - August 18, 2020 3_20 PM

Now we move a file in this folder. The move fails with a 403, we have a delete on the original file (legacy code?) which gets marked as ignore.

Here a log 
[log.txt](https://github.com/owncloud/client/files/5107630/log.txt)

TheOneRing:
See https://github.com/owncloud/client/issues/8066

PR #8044, opened 2020-08-31
allow displayname instead of display-name

jnweiger:
Seen in https://github.com/owncloud/product/issues/181#issuecomment-680223599
Please check which one is correct.


TheOneRing:
That indicates a server bug I guess.
@butonic

butonic:
It is, but this would make the client a little more forgiving. Because the ocs endpoints in oc10 cannot even agree on how to spell things: https://github.com/owncloud/product/issues/181#issuecomment-685789391

The server would like to fix this and make it `displayname` everywhere ... but clients should allow that by checking both ... :chicken: :egg: problem ...

TheOneRing:
Ah ok

TheOneRing (approved):

TheOneRing (changes requested):
Please target 2.7

PR #8049, opened 2020-09-07
Ensure dialog is visible

TheOneRing:


dschmidt (approved):
Neverending story ... I didn't get how this wasn't straightforward ten years ago, I don't get it even less today ten years later...

PR #8050, opened 2020-09-07
Modal share dialog

TheOneRing:


dschmidt (approved):

jnweiger:
## confirmed in rc1. 
sharing dialogs are always shown ontop of the settings window.  OK.

But multiple sharing dialogs can be opened simultaneously. and their stacking order can be disturbed. A later sharing dialog blocks an earlier sharing dialog, but can be hidden behind the earliear one. Okayish.

PR #8055, opened 2020-09-08
Ci: github actions for linux

TheOneRing:


TheOneRing:
Ah the exclude file ...
image

dschmidt (approved):

PR #8064, opened 2020-09-09
Remove the theming option `wizardSelectiveSyncDefaultNothing`

TheOneRing:
We recommend `newBigFolderSizeLimit` together with `wizardHideFolderSizeLimitCheckbox`
as an alternative

### DIFF
--- a/changelog/unreleased/8064
+++ b/changelog/unreleased/8064
@@ -1,0 +1,6 @@
+Change: Remove the branding option `wizardSelectiveSyncDefaultNothing`
+
+The branding option was removed as believe that it did not provide a good user experience.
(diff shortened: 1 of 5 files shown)


dschmidt (approved):

jnweiger:
```
gitea/ownbrander/themes] master* ± grep wizardSelectiveSyncDefaultNothing */*/*-theme.h | wc -l
136
gitea/ownbrander/themes] master* ± grep wizardSelectiveSyncDefaultNothing */*/*-theme.h | grep true | wc -l
14
```

This setting is still used heavily in ownbrander. @TheOneRing  @michaelstingl please schedule removal there too.

TheOneRing:
The old setting won't have an effect and we can only clean up ownbrander after the release

PR #8075, opened 2020-09-11
Http log setting

TheOneRing:


TheOneRing:
https://github.com/owncloud/client/issues/8073

src/libsync/configfile.cpp
```diff
@@ -113,6 +116,13 @@ ConfigFile::ConfigFile()
 
     QSettings settings(config, QSettings::IniFormat);
     settings.beginGroup(defaultConnection());
+
+    // run init only once
+    static bool init = [this]() {
+        setLogHttp(logHttp());
+        return false;
+    }();
+    Q_UNUSED(init);
```
dschmidt, line 125: I really dislike this singleton-ish behavior. 

Can't you call this explicitly from main.cpp or from application or something?

src/libsync/logger.cpp
```diff
@@ -210,14 +201,18 @@ void Logger::setLogFlush(bool flush)
...
+        removeLogRule(rules);
+    }
     _logDebug = debug;
 }
 
 QString Logger::temporaryFolderLogDirPath() const
 {
-    QString dirName = APPLICATION_SHORTNAME + QString("-logdir");
-    return QDir::temp().filePath(dirName);
+    return QDir::temp().filePath(QStringLiteral(APPLICATION_SHORTNAME "-logdir"));
```
dschmidt, line 215: :+1:

src/libsync/configfile.cpp
TheOneRing: Well yes with 2.8 I will rework the whole class.
For now I'd like to just do this...

PR #8076, opened 2020-09-14
Ensure we don't enter empty permissions to the db

TheOneRing:


dschmidt (approved):

TheOneRing:
Well looks like I need to fix the unit tests first :D

TheOneRing:
We decided to kill the poll jobs so this is ready to go @dragotin @dschmidt

dschmidt (approved):

PR #8077, opened 2020-09-14
Don't clear cookies during oauth

TheOneRing:
The cookies are required by the load balancer and there for a reason.
Since #6574 we fixed the back button in https://github.com/owncloud/client/pull/8056 .


dschmidt (approved):
sounds good

Issue #8081, opened 2020-09-14
Docs are missed for 2.7 version


davitol:
Built from Git revision 4cb5cf on Sep 14 2020, 02:12:32 using Qt 5.12.9, OpenSSL 1.1.1g 21 Apr 2020 using Mac Client

<img width="976" alt="Screenshot 2020-09-14 at 14 04 14" src="https://user-images.githubusercontent.com/10596560/93083730-5383e680-f693-11ea-8662-2caf2cdee9b0.png">

<img width="1279" alt="Screenshot 2020-09-14 at 14 05 37" src="https://user-images.githubusercontent.com/10596560/93083783-672f4d00-f693-11ea-820b-401fdceac727.png">

michaelstingl:
@TheOneRing please remove version string:
https://doc.owncloud.org/desktop/

TheOneRing:
We have versioned docs https://doc.owncloud.org/desktop/2.6/ , but the 2.7 docs are not generated yet, part of the release process?
@jnweiger

jnweiger:
https://confluence.owncloud.com/display/OG/Documentation

TheOneRing:
It is step two in the ticket https://github.com/owncloud/client/issues/7966 so it is covered by the release ticket.
Duplicate of https://github.com/owncloud/client/issues/7966

TheOneRing:
As mentioned by @jnweiger , this is not about maintaining multiple versions of the doc, but keeping the old one but also providing the new ones.
@michaelstingl

TheOneRing:
https://github.com/owncloud/docs/pull/2788

PR #8087, opened 2020-09-15
Always use latest doc

TheOneRing:
Fixes: #8081


dschmidt (commented):
Don't think this change is a very good idea, but technically ok.

TheOneRing:
See https://github.com/owncloud/docs/pull/2788

PR #8090, opened 2020-09-15
Don't abort sync when a token refresh is needed

TheOneRing:
https://github.com/owncloud/ocis/issues/576

### DIFF
--- a/src/libsync/creds/httpcredentials.cpp
+++ b/src/libsync/creds/httpcredentials.cpp
@@ -371,8 +371,7 @@ void HttpCredentials::slotReadJobDone(QKeychain::Job *incomingJob)
         return;
     }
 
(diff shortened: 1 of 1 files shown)


dschmidt (approved):
Don't know the code, so the actual change is hard to review for me but there's nothing obviously wrong with the PR.

guruz:
It would be nice if " Don't abort sync when a token refresh is needed " could also be a code comment before your code block, just so it's intuitively clear what it is about.
I can't comment on the code itself..

TheOneRing:
I added a comment and added a log message

ogoffart:
I'm not sure chat this fixes.  I thought that we would already refresh the authentication without aborting. Maybe that was not working good?
Would be nice if the commit message could detail the problem and why this fixes it.

PR #8091, opened 2020-09-16
Ensure receiver exists

TheOneRing:


dschmidt (approved):
Interesting solution!

Issue #8085, opened 2020-09-15
Activity view displays notifications for removed account


TheOneRing:

jnweiger:
## Retested with rc1 on win10 and ubuntu 18.04

* client is connected to demo.owncloud.org and demo.owncloud.com
* via web gui, several files get added, removed, renamed on both serves.
* The client pops up system notifications for each and lists te action in the activity view.

* user removes the .org account from the client.

all entries related to .org disappear from the activity panel. OK.
Linux: system notifications are all erased as soon as the settings window of the client is opend. BAD, but unrelated.
Windows: system notifications remain for both accounts. BAD. Is it possible to remove there too?

TheOneRing:
Windows: Not with the current balloon messgae based notifications

PR #8094, opened 2020-09-17
Remove notifications when account was removed

TheOneRing:
Fixes: #8085


TheOneRing:
Maybe I should have added a TODO: rewrite notifications

dschmidt (approved):
I guess you tested it ... looks okayish, but no idea tbh

PR #8096, opened 2020-09-17
Remove support for legacy private links

TheOneRing:


dschmidt:
Which was the last ownCloud Server version that required this?

src/gui/accountsettings.cpp
```diff
@@ -263,7 +263,7 @@ void AccountSettings::slotCustomContextMenuRequested(const QPoint &pos)
         QString path = info->_folder->remotePathTrailingSlash();
         path += info->_path;
         connect(ac, &QAction::triggered, this, [this, path]{
-            fetchPrivateLinkUrl(_accountState->account(), path, {}, this, [](const QString &url) {
```
dschmidt, line 266: Uargh 🙈

dschmidt (approved):

TheOneRing:
This will go to master now, in 2.8 we will officially remove support for all servers < 10

dschmidt:
Sounds good!

PR #8098, opened 2020-09-18
Add stale bot

TheOneRing:
Currently its in debug mode so it will only perform dry runs.
It is also configured to not mark issues as stale by itself.
Only an issue marked as stale by us will be closed after 7 days.

See https://github.com/actions/stale/blob/main/action.yml

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,16 @@
+name: "Close stale issues"
+on:
+  schedule:
+  - cron: "0 0 * * *"
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


dschmidt:
Can the 7 days be explicitly configured? 
Otherwise I'm feeling a bit uneasy about mentioning the 7 days interval in the message while relying on some default value, that might change 🙈

TheOneRing:
> 
> 
> Can the 7 days be explicitly configured?
> Otherwise I'm feeling a bit uneasy about mentioning the 7 days interval in the message while relying on some default value, that might change 🙈

yes can be changed

dschmidt (approved):

PR #8100, opened 2020-09-18
Cache Plugin availability

TheOneRing:
We abandoned the review as it uses a static var for caching and @dschmidt doesn't like statics
but this code starts to become used more and more often and what we do currently is meh...


dschmidt:
What does "used more and more often" mean? Is this a bottleneck somewhere? 

And yes, I really dislike static vars in functions. A testing nightmare...

TheOneRing:
I still don't like what we do here, the proper solution would be to move it to an object instance ...

PR #8115, opened 2020-09-28
WIP: Enable sonarcloud.io

TheOneRing:


TheOneRing:
I guess I should target 2.7?

C0rby (approved):

TheOneRing:
@C0rby I need to rebase against 2.7 and the full depth clone did not work, which is needed for git blame support.
The rest of the config is ok? Anything we should enable /disable?

C0rby:
The config looks fine to me.
What we need to keep in mind is the quality gate in sonar for future pull requests.
Per default it is set to something like 80% test coverage x% duplication and if the PRs don't match that gate you won't be able to merge stuff.
So that needs to be configured to sensible values.

TheOneRing:
kk, for coverage we should probably add a step to generate a llvm coverage file and well then write more tests :D

dschmidt (approved):
Ok I guess 🙈

PR #8119, opened 2020-09-29
Correctly handle error in fetch user data during oauth flow

TheOneRing:


dschmidt (approved):
Looks more reasonable than the first version :)

Issue #8107, opened 2020-09-24
Clarify that "ask for confirmation before syncronizing folders" > and external storrages only applies for non vfs


TheOneRing:
![CleanShot 2020-09-24 at 12 20 33](https://user-images.githubusercontent.com/200626/94133666-3ef8d880-fe61-11ea-8186-c1ec7077cfda.png)

TheOneRing:
Suggestion
![image](https://user-images.githubusercontent.com/200626/94820102-59522980-0400-11eb-98ea-cbc38e69a9ac.png)

PR #8123, opened 2020-10-01
Gui: Mark non vfs settings explicitly

TheOneRing:
See #8107 for screenshots

### DIFF
--- a/src/gui/generalsettings.ui
+++ b/src/gui/generalsettings.ui
@@ -64,115 +64,128 @@
          <property name="title">
           <string>Advanced</string>
          </property>
-         <layout class="QVBoxLayout" name="verticalLayout">
+         <layout class="QHBoxLayout" name="horizontalLayout_5">
           <item>
-           <layout class="QHBoxLayout" name="horizontalLayout_3">
+           <layout class="QVBoxLayout" name="verticalLayout_6">
             <item>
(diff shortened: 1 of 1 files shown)


dschmidt:
I'm torn:
I think the structure is good but box in box in box isn't really optimal ... maybe we can get away with a regular column based layout and headlines at the top?

dschmidt (commented):
I'm torn:
I think the structure is good but box in box in box isn't really optimal ... maybe we can get away with a regular column based layout and headlines at the top?

TheOneRing:
Hm there is no real column based design element.
I also think that it is a bit cramped...
I could move the non virtual files only group out to its own seperate section like netowrk.
Something like "Advamced - Non virtual files"

TheOneRing:
image

michaelstingl:
Maybe new section "Approval of folders (Non virtual file mode only)" or "Folder download approvals (Non virtual file mode only)"

TheOneRing:
"Approval of folders sync (Non virtual file mode only)" ?

michaelstingl:
"Approval of folder sync (Non virtual file mode only)" / "Approval of folder download (Non virtual file mode only)" ?

We’re getting closer…

TheOneRing:
image

dschmidt (approved):
fine with me, if fine with @michaelstingl

Issue #8104, opened 2020-09-21
Windows: Display name for folder sync pairs does not include remote path


TheOneRing:
# Current:
![image](https://user-images.githubusercontent.com/200626/93753579-a82de100-fc00-11ea-9eb5-b709fd00659d.png) 

# New:
![image](https://user-images.githubusercontent.com/200626/94824567-776e5880-0405-11eb-9bc1-982e8ed4e7bb.png)
![image](https://user-images.githubusercontent.com/200626/94901894-a503e200-0497-11eb-8010-1e4cf8e483e4.png)
![image](https://user-images.githubusercontent.com/200626/94901920-b1883a80-0497-11eb-8b01-a9238c7d051b.png)

michaelstingl:
@TheOneRing oCIS? Should be fixed in latest version. (`display-name` vs `displayname`)

TheOneRing:
no thats unrelated to the wrong user name, I'm talking about multiple sync folder pairs.
here / and /sigh/

michaelstingl:
I'd follow more the Federated Cloud ID:
`test@demo.owncloud.com`
(https://demo.owncloud.com/settings/personal)

Then add the full path or only the last path segment maybe…

![CleanShot 2020-09-21 at 13 44 09](https://user-images.githubusercontent.com/214010/93763035-fc40c180-fc10-11ea-950a-af3d29f02ca2.png)

@pmaier1 what do you think how to display the left sidebar in Win Explorer?

michaelstingl:
Alternative would be, to use the name of the local folder the user choose:

![CleanShot 2020-09-21 at 14 37 18](https://user-images.githubusercontent.com/214010/93767318-2649b200-fc18-11ea-878d-5586ed4d74cd.png)

pmaier1:
I basically like the idea in the example and I also like the idea of using the Fed ID here. I'd say we do one of the following:

**a) Like today but with sync connection location**

ownCloud - `<domain>` (`<user>`) - `<sync connection location>`

**b) With Fed ID and sync connection location**

ownCloud - `<user@domain>` - `<sync connection location>`

=> ownCloud - demo@cloud.owncloud.com - /Documents

I'd like to see b) in action. Actually we should embrace federation but not sure if it maybe looks too technical here.
Maybe we can drop the last part (`- <sync connection location>`) if it's only one sync connection and that's `/`. I'd expect that mostly power users will have more than one sync connection and those will understand.

For the sync folder naming default, we could go for

[...]

PR #8124, opened 2020-10-02
Account display name `servername (username) -> username@servername`

TheOneRing:
Issue: #8104

# New:
image

### DIFF
--- a/src/libsync/account.cpp
+++ b/src/libsync/account.cpp
@@ -114,13 +114,11 @@ QString Account::displayName() const
     if (user.isEmpty())
         user = davUser();
     QString host = _url.host();
-    int port = url().port();
+    const int port = url().port();
     if (port > 0 && port != 80 && port != 443) {
-        host.append(QLatin1Char(':'));
-        host.append(QString::number(port));
+        host += QStringLiteral(":%1").arg(QString::number(port));
     }
-    QString dn = QStringLiteral("%1 (%2)").arg(host, user);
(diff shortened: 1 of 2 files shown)


src/libsync/account.cpp
```diff
@@ -114,13 +114,11 @@ QString Account::displayName() const
...
-    int port = url().port();
+    const int port = url().port();
     if (port > 0 && port != 80 && port != 443) {
-        host.append(QLatin1Char(':'));
-        host.append(QString::number(port));
+        host += QStringLiteral(":%1").arg(QString::number(port));
     }
-    QString dn = QStringLiteral("%1 (%2)").arg(host, user);
-    return dn;
+    return tr("%1@%2").arg(user, host);
```
TheOneRing, line 121: I switched to a translatable string even so that syntax now should be pretty universal

michaelstingl (approved):
Changelog 👍 
(can't judge the actual code)

pmaier1 (approved):
<3 (can't judge on code :x)

src/libsync/account.cpp
dschmidt: Yeah, not sure this brings actual benefit ...

dschmidt (approved):

TheOneRing:
yes the previous string probably needed translations for right to left languages

TheOneRing:
but mail address like things probably not xD

PR #8128, opened 2020-10-02
Remove the update channel settings from the unbranded client

TheOneRing:


pmaier1 (approved):
:+1:

dschmidt:
So the settings is still available for testpilotcloud? Maybe you can reflect that in the changelog more explicitly?

changelog/unreleased/8127
```diff
@@ -0,0 +1,8 @@
+Change: Remove update channels from the ownCloud client
+
+To ensure a maximum of stability user should not replace their productive client with a preview build.
+For that exact reason we offer tespilotcloud clients, they can be installed in parallel.
```
dschmidt, line 4:
```suggestion
For that exact reason we offer tespilotcloud clients, they can be installed in parallel and updated via the beta channel.
```

dschmidt (approved):

PR #8132, opened 2020-10-02
Evaluate enableExperimentalFeatures when handling suffix vfs

TheOneRing:
This change kind of decouples the vfs modes a bit.
If enableExperimentalFeatures == false, we will hide the suffix implementation.
showVirtualFilesOption can still be used to completely disable vfs support.

We need to set showVirtualFilesOption to true per default in ownbrander.

WIP because its Friday evening and those conditions are complex....

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -337,11 +337,12 @@ void AccountSettings::slotCustomContextMenuRequested(const QPoint &pos)
     }
 
     if (Theme::instance()->showVirtualFilesOption()
-        && !folder->supportsVirtualFiles()
-        && bestAvailableVfsMode() != Vfs::Off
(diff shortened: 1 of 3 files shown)


src/gui/accountsettings.cpp
```diff
@@ -337,11 +337,12 @@ void AccountSettings::slotCustomContextMenuRequested(const QPoint &pos)
     }
 
     if (Theme::instance()->showVirtualFilesOption()
-        && !folder->supportsVirtualFiles()
-        && bestAvailableVfsMode() != Vfs::Off
```
TheOneRing, line 341: Line 341 and 342 are already covered by supportsVirtualFiles

dschmidt (approved):

Issue #8103, opened 2020-09-18
[QA] client sows UID before authentication succeeds, then display name


jnweiger:
Tested with
 * ocis master from 7h ago
 * appliance from https://github.com/owncloud/client/actions/runs/255830173

<details>
<summary> ocis master version details</summary>

<pre>
OCIS_VERSION:         master
ocis --version:       ocis version 7b8e9bc
git log:              commit 7b8e9bc2981c0fd9917bc576a291ab5231c02685 (HEAD -> refs/heads/master, refs/remotes/origin/master, refs/remotes/origin/HEAD)
eos --version:        EOS 4.6.5 (2019)
xrootd -v:            v4.11.0
bin/ocis contains:
	owncloud/ocis-accounts@v0.4.2-0.20200911161616-e55b8ae2f440
	owncloud/ocis-glauth@v0.5.1-0.20200909064150-0735ec933777
	owncloud/ocis-graph-explorer@v0.0.0-20200210111049-017eeb40dc0c
	owncloud/ocis-graph@v0.0.0-20200318175820-9a5a6e029db7
	owncloud/ocis-hello@v0.1.0-alpha1.0.20200828085053-37fcf3c8f853
	owncloud/ocis-konnectd@v0.3.2
	owncloud/ocis-migration@v0.2.0
	owncloud/ocis-ocs@v0.3.2-0.20200916110615-7ca52baa61c4
	owncloud/ocis-phoenix@v0.13.1-0.20200910090722-8e3ed0e966a5
	owncloud/ocis-pkg/v2@v2.4.1-0.20200902134813-1e87c6173ada
	owncloud/ocis-proxy@v0.7.1-0.20200907105449-201b9a652685
	owncloud/ocis-reva@v0.14.0
	owncloud/ocis-settings@v0.3.2-0.20200903035407-ad5de8264f91
	owncloud/ocis-store@v0.1.1
	owncloud/ocis-thumbnails@v0.3.0
	owncloud/ocis-webdav@v0.1.1
</pre></details>

Reproducer:
* connect client to server
* stop client
* restart client
![image](https://user-images.githubusercontent.com/1108546/93623855-aafdbb80-f9df-11ea-8e25-79097d24743d.png)

* the browser opens to authenticate

![image](https://user-images.githubusercontent.com/1108546/93623896-bbae3180-f9df-11ea-8e4c-d1f909d35791.png)

* after authentication, the UID gets replaced with the display name

[...]

PR #8141, opened 2020-10-06
    Ensure that we always have a display name ready when we register a folder

TheOneRing:
Fixes: #8103

### DIFF
--- a/src/gui/owncloudsetupwizard.cpp
+++ b/src/gui/owncloudsetupwizard.cpp
@@ -308,6 +308,16 @@ void OwncloudSetupWizard::testOwnCloudConnect()
     connect(job, &PropfindJob::result, _ocWizard, &OwncloudWizard::successfulStep);
     connect(job, &PropfindJob::finishedWithError, this, &OwncloudSetupWizard::slotAuthError);
     job->start();
+
+    // get the display name so that the folder we register with the file browser has a pretty name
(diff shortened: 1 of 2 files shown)


src/gui/owncloudsetupwizard.cpp
```diff
@@ -308,6 +308,16 @@ void OwncloudSetupWizard::testOwnCloudConnect()
     connect(job, &PropfindJob::result, _ocWizard, &OwncloudWizard::successfulStep);
     connect(job, &PropfindJob::finishedWithError, this, &OwncloudSetupWizard::slotAuthError);
     job->start();
+
+    // get the display name so that the folder we register with the file browser has a pretty name
+    auto userJob = new JsonApiJob(account, QStringLiteral("ocs/v2.php/cloud/user"), this);
+    connect(userJob, &JsonApiJob::jsonReceived, this, [account](const QJsonDocument &json, int status) {
+        if (status == 200) {
+            const QString user = json.object().value(QStringLiteral("ocs")).toObject().value(QStringLiteral("data")).toObject().value(QStringLiteral("display-name")).toString();
```
dschmidt, line 316: does this need any check in the chain or at the end if user is empty?

dschmidt (approved):

src/gui/owncloudsetupwizard.cpp
TheOneRing: if its our server and we got a 200 its really supposed to be there and if we didn't get a 200 well probably the first check failed and nothing is lost

PR #8152, opened 2020-10-12
Fix warning icons

TheOneRing:


TheOneRing:
The actual error here was that the warning icon was removed as I didn't find references to that png...
Should I reword and keep the url without the extension?

dschmidt (approved):

dschmidt:
> The actual error here was that the warning icon was removed as I didn't find references to that png...
> Should I reword and keep the url without the extension?

I don't mind :-D

PR #8157, opened 2020-10-13
Oauth use `prompt=select_account` instead of `consent`

TheOneRing:
The idp will now provide a selection dialog.
This fixes the issue where the login is done with the wrong account.
Also provide `login_hint` so that providers supporting it can skip that dialog.

### DIFF
--- a/src/gui/owncloudsetupwizard.cpp
+++ b/src/gui/owncloudsetupwizard.cpp
@@ -313,8 +313,13 @@ void OwncloudSetupWizard::testOwnCloudConnect()
     auto userJob = new JsonApiJob(account, QStringLiteral("ocs/v2.php/cloud/user"), this);
     connect(userJob, &JsonApiJob::jsonReceived, this, [account](const QJsonDocument &json, int status) {
         if (status == 200) {
-            const QString user = json.object().value(QStringLiteral("ocs")).toObject().value(QStringLiteral("data")).toObject().value(QStringLiteral("display-name")).toString();
-            account->setDavDisplayName(user);
+            const auto data = json.object().value(QStringLiteral("ocs")).toObject().value(QStringLiteral("data")).toObject();
+            const QString userName = data.value(QStringLiteral("username")).toString();
+            const QString displayName = data.value(QStringLiteral("display-name")).toString();
+            account->setDavDisplayName(displayName);
+            if (!userName.isEmpty()) {
+                account->setIdpUserName(userName);
+            }
         }
     });
     userJob->start();
--- a/src/libsync/account.cpp
+++ b/src/libsync/account.cpp
@@ -78,6 +78,16 @@ void Account::setSharedThis(AccountPtr sharedThis)
     _sharedThis = sharedThis.toWeakRef();
 }
 
+QString Account::idpUserName() const
+{
+    return _idpUserName;
+}
+
+void Account::setIdpUserName(const QString &userName)
+{
+    _idpUserName = userName;
+}
+
 AccountPtr Account::sharedFromThis()
 {
     return _sharedThis.toStrongRef();
--- a/src/libsync/theme.cpp
+++ b/src/libsync/theme.cpp
@@ -539,6 +539,11 @@ QString Theme::openIdConnectScopes() const
     return QStringLiteral("openid offline_access email profile");
 }
 
+QString Theme::openIdConnectPrompt() const
+{
+    return QStringLiteral("select_account");
+}
+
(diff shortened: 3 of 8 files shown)


TheOneRing:
the dav user might not be what the idp actually uses, do we need to introduce something besides user_id and display name?

TheOneRing:
Well looks like I'd need to use both consent and select_account separated by space, but yay konnect does a simple string compare...

https://github.com/Kopano-dev/konnect/blob/77149f57529b9c94f5d1c8b18a40eeac2bb04fe9/identifier/src/containers/Login/Login.js#L51

DeepDiver1975:
This might be yet another branding option. Further more this can be part of the well known openid configuration afaik.
This is why I am always stating: use a proper library .... ( I know there is none in at)

dschmidt (approved):
i like

TheOneRing:
The new changes uses the username provided by ocis for the hint.
The issue with multiple prompts ist still not solved and I'm not sure how we will.... branding?
At least with konnect the current setup promt seems to work as we still get the consent page after the user page.
any ideas?

DeepDiver1975 (commented):
Did you read up https://openid.net/specs/openid-connect-core-1_0.html (search for login_hint)

```
login_hint
    OPTIONAL. Hint to the Authorization Server about the login identifier the End-User might use to log in (if necessary). This hint can be used by an RP if it first asks the End-User for their e-mail address (or other identifier) and then wants to pass that value as a hint to the discovered authorization service. It is RECOMMENDED that the hint value match the value used for discovery. This value MAY also be a phone number in the format specified for the phone_number Claim. The use of this parameter is left to the OP's discretion. 
```

This sounds like this is data to be given to us by the end user ....

TheOneRing:
love it:
```
When your app knows which user it is trying to authenticate, it can provide this parameter as a hint to the authentication server. Passing this hint suppresses the account chooser and either pre-fills the email box on the sign-in form, or selects the proper session (if the user is using multiple sign-in), which can help you avoid problems that occur if your app logs in the wrong user account. The value can be either an email address or the sub string, which is equivalent to the user's Google ID.
```
https://developers.google.com/identity/protocols/oauth2/openid-connect#re-consent

So yes every idp will do something, and no one will know what.

TheOneRing:
https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-protocols-oidc
Azure ad only supports consent

src/gui/owncloudsetupwizard.cpp
```diff
@@ -313,8 +313,13 @@ void OwncloudSetupWizard::testOwnCloudConnect()
     auto userJob = new JsonApiJob(account, QStringLiteral("ocs/v2.php/cloud/user"), this);
     connect(userJob, &JsonApiJob::jsonReceived, this, [account](const QJsonDocument &json, int status) {
         if (status == 200) {
-            const QString user = json.object().value(QStringLiteral("ocs")).toObject().value(QStringLiteral("data")).toObject().value(QStringLiteral("display-name")).toString();
-            account->setDavDisplayName(user);
+            const auto data = json.object().value(QStringLiteral("ocs")).toObject().value(QStringLiteral("data")).toObject();
+            const QString userName = data.value(QStringLiteral("username")).toString();
+            const QString displayNamer = data.value(QStringLiteral("display-name")).toString();
```
dschmidt, line 318: displayName**r**? :trollface:

src/libsync/account.cpp
```diff
@@ -78,6 +78,16 @@ void Account::setSharedThis(AccountPtr sharedThis)
     _sharedThis = sharedThis.toWeakRef();
 }
 
+QString Account::idpUuserName() const
```
dschmidt, line 81: idpU**u**serName?

src/libsync/theme.cpp
```diff
@@ -539,6 +539,11 @@ QString Theme::openIdConnectScopes() const
     return QStringLiteral("openid offline_access email profile");
 }
 
+QString Theme::openIdConnectPromt() const
```
dschmidt, line 542: openIdConnectProm**p**t?

dschmidt (approved):
No idea about the code, but please fix the typos

src/libsync/account.cpp
TheOneRing: 🤦

src/libsync/theme.cpp
TheOneRing: 🤦

PR #8161, opened 2020-10-14
Try add a badge for the daily builds

jnweiger:
It seems not to work. Sigh. But even in the broken state the reference is valuable.


TheOneRing:
Well it wont work as our internal ci is behind a vpn :D

dschmidt:
Yeah this doesn't belong into the public repository.

Issue #8154, opened 2020-10-12
[QA] newline missing at the end of --version output


jnweiger:
## Seen on Ubuntu 20.04

```
[testy:~] $ 
[testy:~] $ 
[testy:~] $ owncloud --version
ownCloud 2.7.0v270rc1 (build 2254) Oct  9 2020 12:56:43
https://github.com/owncloud/client/commit/ec4341372eb026bb21e38a246e6698f6a684270c
Libraries Qt 5.12.9, OpenSSL 1.1.1  11 Sep 2018
Using virtual files plugin: off[testy:~] $ 
[testy:~] $ 
[testy:~] $ 
```

The last line of the output is `Using virtual files plugin: off`
The shell prompt should be in the next line.

jnweiger:
Seen again with 2.7.1:
```
[root@cd1386ccd51c /]# owncloudcmd --version
ownCloud 2.7.1 (build 2560) Nov 18 2020 08:46:25
https://github.com/owncloud/client/commit/afe5732722f5303889933f7baab0014f661d4d93
Libraries Qt 5.12.9, OpenSSL 1.1.1g FIPS  21 Apr 2020
Using virtual files plugin: off[root@cd1386ccd51c /]# 
[root@cd1386ccd51c /]# 
[root@cd1386ccd51c /]# 
[root@cd1386ccd51c /]# 
```

please try harder :-)

fmoc:
Cannot reproduce in latest master. @jnweiger could you please re-check whether the issue has been fixed in the meantime?

PR #8165, opened 2020-10-14
Always append a newline to the help text

TheOneRing:
Fixes: #8154

### DIFF
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -71,7 +71,9 @@ Q_LOGGING_CATEGORY(lcApplication, "gui.application", QtInfoMsg)
 
 namespace {
 
-    static const char optionsC[] =
+    const QString optionsC()
+    {
+        return QStringLiteral(
         "Options:\n"
         "  -h --help            : show this help screen.\n"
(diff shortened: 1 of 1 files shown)


src/gui/application.cpp
```diff
@@ -640,12 +643,13 @@ void Application::showHelp()
 
     stream << QLatin1String("File synchronisation desktop utility.") << endl
            << endl
-           << QLatin1String(optionsC);
+           << optionsC();
 
-    if (_theme->appName() == QLatin1String("ownCloud"))
+    if (_theme->appName() == QLatin1String("ownCloud")) {
```
dschmidt, line 648: This is the second time I read more or less this exact same comparison in different parts of the code.
Should we maybe have something like theme->isBranded() and/or theme->isUnbranded()? 

Totally not needed for this PR obviously...

dschmidt (approved):

src/gui/application.cpp
TheOneRing: Thought the same when I did the last change....
But we have also a check for testpilotcloud

PR #8167, opened 2020-10-14
Use platform icon theme only on the unbranded client

TheOneRing:
Fixes: https://github.com/owncloud/ownbrander/issues/715

There are also two other cases where the system style might be used:
```
git grep fromTheme
src/gui/sharelinkwidget.cpp:    auto deleteIcon = QIcon::fromTheme(QStringLiteral("user-trash"),
src/gui/shareusergroupwidget.cpp:    _ui->deleteShareButton->setIcon(QIcon::fromTheme(QStringLiteral("user-trash"),


dschmidt:
This is apparently currently only applied to openSUSE builds, but  @jnweiger 's comment here suggests that there is no special reason for that ......... https://github.com/owncloud/enterprise/issues/3318#issuecomment-491202639

So I think this patch is accurate.

dschmidt (approved):

dschmidt:
https://gitea.owncloud.services/client/build-linux/pulls/7

PR #8172, opened 2020-10-15
Make dialogs non blocking

TheOneRing:
Remaining blocking dialogs are the ssl dialog which turned out to be quite complicated...
And an error dialog during initialization...


dschmidt (approved):
Hard to review in detail, but nothing really obviously wrong with this

PR #8175, opened 2020-10-19
Replace custome date parser QDateTime::fromString(value, Qt::RFC2822D…

TheOneRing:
…ate)

Should we wait until after 2.7.0?


dschmidt:
Let's wait

jnweiger:
Do you have an example, what is fixed by this?

TheOneRing:
It replaces a whole lot of unmaintained custom copy paste code with a Qt function.
It introduces, error detection on the level of the log.
Its supposed to properly handle other time zones while the old code only assumed gmt.

TheOneRing:
Targeting master now

dschmidt (approved):

PR #8179, opened 2020-10-20
Add branding option to enforce VFS on Windows

TheOneRing:
image

### DIFF
--- a/src/gui/folderwizard.cpp
+++ b/src/gui/folderwizard.cpp
@@ -480,8 +480,9 @@ FolderWizardSelectiveSync::FolderWizardSelectiveSync(const AccountPtr &account)
     layout->addWidget(_selectiveSync);
 
     const auto vfsMode = bestAvailableVfsMode();
-    if (Theme::instance()->showVirtualFilesOption() && vfsMode != Vfs::Off && (vfsMode == Vfs::WindowsCfApi || Theme::instance()->enableExperimentalFeatures())) {
-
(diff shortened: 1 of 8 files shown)


src/gui/folderwizard.cpp
```diff
@@ -479,8 +479,9 @@ FolderWizardSelectiveSync::FolderWizardSelectiveSync(const AccountPtr &account)
     layout->addWidget(_selectiveSync);
 
     const auto vfsMode = bestAvailableVfsMode();
-    if (Theme::instance()->showVirtualFilesOption() && vfsMode != Vfs::Off && (vfsMode == Vfs::WindowsCfApi || Theme::instance()->enableExperimentalFeatures())) {
-
+    if (Theme::instance()->forceVirtualFilesOption() && vfsMode == Vfs::WindowsCfApi) {
+        // using an else if too make the condition not even more complex...
+    } else if (Theme::instance()->showVirtualFilesOption() && vfsMode != Vfs::Off && (vfsMode == Vfs::WindowsCfApi || Theme::instance()->enableExperimentalFeatures())) {
```
dragotin, line 484: Using && and || in one statement is a killer... IMHO.

dragotin:
On a general note: If we have an option "forceVFS" that makes the "enableExperimentalFeature" option obsolete automatically. How would we have a way to force an experimental feature? That does not make sense IMHO.

TheOneRing:
The branch is horrible outdated, by now windows vfs is no longer an experimental feature...
Only the unix vfs is considered experimental.

dschmidt (approved):

PR #8184, opened 2020-10-21
Append .exe to crashreporter path so it is better located

TheOneRing:


src/gui/application.cpp
```diff
@@ -246,8 +246,15 @@ Application::Application(int &argc, char **argv)
...
 
 #if defined(WITH_CRASHREPORTER)
-    if (ConfigFile().crashReporter())
-        _crashHandler.reset(new CrashReporter::Handler(QDir::tempPath(), true, CRASHREPORTER_EXECUTABLE));
+    if (ConfigFile().crashReporter()) {
+        auto reporter = QStringLiteral(CRASHREPORTER_EXECUTABLE);
+#ifdef Q_OS_WIN
+        if (reporter.endsWith(QLatin1String(".exe"))) {
+            reporter.append(QLatin1String(".exe"));
+        }
```
dschmidt, line 254: I think this is the opposite of what you want to do  :trollface:

dschmidt (changes requested):

src/gui/application.cpp
dschmidt: AFAICT it should never have .exe anyway tho...

TheOneRing: your right the if is wrong 🙈 
But the crashreporter.exe is an exe so ....
if the full file name is passed it uses the exe close to the binary if available, else it falls back to path.

TheOneRing: Ah yes we don't append exe yet, but ...

dschmidt (approved):

Issue #8188, opened 2020-10-22
[QA] client enters busy loop, when migrating back from openidconnect to oauth2


jnweiger:
Seen with testpilotcloud 2.7.0daily20201019 (build 2311) and server 10.5 with openidconnect-1.0.0RC5 and oauth2-0.4.4RC1

* have openidconnect enabled at the server
* connect desktop client to server (openidconnect is used).
* enable oauth2 at the server, disable openidconnect
* wait or force resync at the client

The client shows the blue circular busy arrow, but shows neither progress nor error.
The server log starts scrolling error messages.

[20201022_1440_owncloud.log.0.gz](https://github.com/owncloud/client/files/5422874/20201022_1440_owncloud.log.0.gz)
[server.log.gz](https://github.com/owncloud/client/files/5422889/server.log.gz)

jnweiger:
Remove connection at the client, reconnect to the same server, same account.
Now oauth2 gets used. Normal operation. The setup itself seems good.

Issue #8186, opened 2020-10-21
[QA] client enters busy loop, when oauth2 user gets deleted from owncloud server


jnweiger:
Seen with testpilotcloud 2.7.0daily20201019 (build 2311) Oct 15 2020 00:08:32 and oc10.5 with enabled oauth2-0.4.4RC1

* desktop client logs in via oauth2
* owncloud admin deletes the account of the user
* wait 15 seconds. when a token refresh is due, the client enter a busy loop:

The client log has
```
10-21 17:06:02:132 [ info sync.httplogger ]:	"3e888bf0-cc81-45f2-bf3b-0cba7cb4842f: Response: PROPFIND 401 https://oc-10-5-0.oidc-1-0-0rc5-20201021.jw-qa.owncloud.works/remote.php/dav/files/aaliyah_adams/ Header: { Cache-Control: no-store, no-cache, must-revalidate, Content-Length: 476, Content-Security-Policy: default-src 'none';, Content-Type: application/xml; charset=utf-8, Date: Wed, 21 Oct 2020 15:06:02 GMT, Expires: Thu, 19 Nov 1981 08:52:00 GMT, Pragma: no-cache, Server: Caddy, Apache, Strict-Transport-Security: max-age=15552000;, Www-Authenticate: Bearer realm=\"ownCloud\", Basic realm=\"ownCloud\", charset=\"UTF-8\", X-Content-Type-Options: nosniff, X-Download-Options: noopen, X-Frame-Options: SAMEORIGIN, X-Permitted-Cross-Domain-Policies: none, X-Robots-Tag: none, X-Xss-Protection: 1; mode=block, } Data: []"
10-21 17:06:02:132 [ info sync.networkjob ]:	Restarting "PROPFIND" QUrl("https://oc-10-5-0.oidc-1-0-0rc5-20201021.jw-qa.owncloud.works/remote.php/dav/files/aaliyah_adams/")
10-21 17:06:02:133 [ info sync.accessmanager ]:	6 "PROPFIND" "https://oc-10-5-0.oidc-1-0-0rc5-20201021.jw-qa.owncloud.works/remote.php/dav/files/aaliyah_adams/" has X-Request-ID "f9b1cb62-60a9-4f93-9b4f-ddc88650184e"
[...]

Issue #8122, opened 2020-10-01
Oauth: Connection does not recover after network loss


TheOneRing:

TheOneRing:
See https://github.com/owncloud/client/issues/8186

PR #8189, opened 2020-10-22
Oauth: Reset login state on error

TheOneRing:
Fixes: #8188, Fixes: #8122, Fixes: #8186


jnweiger:
Logout seems to be a good idea in general.
In the case of #8188, I observe a strange sideeffect when logging out manually:

* user clicks the logout entry in the accounts menu
* the loop ends, and "aborted" is shown. OK.
image

* user tries to log in again with the 'accounts' menu
* but fails:
image

Expected behaviour: the user can log in. Normal operation (without busy loop).


Workaround: After waiting a while, a timeout error appears, now the next login attempt starts a fresh oauth2 cycle and succeeds.

TheOneRing:
it takes a few seconds until the login is displayed in the ui but yes you will get a reopen browser button now.
You can probably already test using the appimage

dschmidt (approved):
hard to judge from just looking at the code ... but if @jnweiger  is happy, I am too.

Issue #8187, opened 2020-10-22
Dektop Client crash when local folder has no read/write permissions (cfapi)


iHerrmann77:
### Actual behaviour
When i add a folder sync connection without permissions to that folder, the sync client crashes.

![Screenshot from 2020-10-22 09-37-14](https://user-images.githubusercontent.com/63246306/96842110-ed863e00-144c-11eb-964f-aca892416ec2.png)

### Expected behaviour
When i type or choose the path to that folder, check permissions before I am able to select the owncloud and "Add sync connection" with any kind of error-message

ownCloud 2.7.0v270rc1 (build 2254) ec4341 Oct 9 2020 06:15:11
Libraries Qt 5.12.9, OpenSSL 1.1.1g 21 Apr 2020
Using virtual files plugin: wincfapi

crash-report
fbbf6656-f09a-4151-8309-7c3f7b3b7294

logfile
[ info sync.networkjob.lscol ]:	LSCOL of QUrl("https://demo.owncloud.com/remote.php/dav/files/demo/") FINISHED WITH STATUS "OK"
[ debug sync.networkjob ]	[ OCC::AbstractNetworkJob::slotFinished ]:	Network job OCC::LsColJob finished for ""
[ info gui.account.settings ]:	Folder wizard completed
[ info sync.utility ]:	Creating "C:/Folder1/Folder2/Folder3//Desktop.ini" to set a folder icon in Explorer.
[ warning default ]:	QIODevice::write (QFile, "C:\Folder1\Folder2\Folder3\\Desktop.ini"): device not open
[ warning default ]:	QIODevice::write (QFile, "C:\Folder1\Folder2\Folder3\\Desktop.ini"): device not open
[ warning default ]:	QIODevice::write (QFile, "C:\Folder1\Folder2\Folder3\\Desktop.ini"): device not open
[ info sync.utility ]:	Creating favorite link from "C:/Folder1/Folder2/Folder3/" to "C:/Users/tester/Links/Folder3.lnk"
[ warning sync.database ]:	Could not find a writable database path "C:/Folder1/Folder2/Folder3/.sync_8da9d25ef3ff.db" "Access is denied."
[ info plugins ]:	Created VFS instance from plugin "owncloudsync_vfs_win"
[ info sync.database ]:	Detected filesystem "NTFS" for "C:/Folder1/Folder2/Folder3/.sync_8da9d25ef3ff.db"
[...]

PR #8191, opened 2020-10-22
Fix folder permission check on NTFS

TheOneRing:
Fixes: #8187

### DIFF
--- a/src/common/utility_win.cpp
+++ b/src/common/utility_win.cpp
@@ -29,6 +29,8 @@
 
 #include <QLibrary>
 
+extern Q_CORE_EXPORT int qt_ntfs_permission_lookup;
+
 static const char systemRunPathC[] = "HKEY_LOCAL_MACHINE\\Software\\Microsoft\\Windows\\CurrentVersion\\Run";
 static const char runPathC[] = "HKEY_CURRENT_USER\\Software\\Microsoft\\Windows\\CurrentVersion\\Run";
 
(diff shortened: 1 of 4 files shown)


src/common/utility_win.cpp
```diff
@@ -293,4 +295,31 @@ QString Utility::formatWinError(long errorCode)
...
+    }
+}
+
+void Utility::NtfsPermissionLookupRAII::disable()
+{
+    if (_enabled) {
+        _enabled = false;
+        qt_ntfs_permission_lookup--;
+    }
+}
```
dschmidt, line 323: Wouldn't it be enough to simply increment/decrement `qt_ntfs_permission_lookup` in the ctor/dtor?

What is _enabled good for?

And without that do we really need enable and disable methods? Looks kinda verbose ... but do whatever you like

dschmidt (changes requested):

src/common/utility_win.cpp
TheOneRing: the boolean is to prevent negative values for qt_ntfs_permission_lookup as he code there relally just checks for  qt_ntfs_permission_lookup > 0

dschmidt: Yes, that's what I'm saying:

```
Utility::NtfsPermissionLookupRAII::NtfsPermissionLookupRAII()
{
   qt_ntfs_permission_lookup++;
}
Utility::NtfsPermissionLookupRAII::~NtfsPermissionLookupRAII()
{
   qt_ntfs_permission_lookup--;
}
```

... should be enough.

dschmidt:
> in theory it allows to reset before the scope runs out

I think it's too complicated for what it achieves, if you need a smaller scope inside a function e.g., you can still do a "sub-scope":

```cpp
Foo::bar() 
{
  ... 
  {
      // this is a scope inside the function
      NtfsPermissionLookupRAII nl{};
  }
  ...
}
```

dschmidt (approved):

PR #8193, opened 2020-10-23
Ci: Improve sonar settings

TheOneRing:


dschmidt (approved):
not convinced the loop is more readable than the original version ...

PR #8194, opened 2020-10-23
[WIP]Cache sonarcloud

TheOneRing:


TheOneRing:
I don't get why the cache isnt uploading anything

.github/workflows/main.yml
```diff
@@ -47,11 +47,19 @@ jobs:
         fetch-depth: ${{ matrix.fetch-depth }}
         submodules: true
 
+    - name: Restore cache
```
fmoc, line 50: This name is also used for the "post" action that is registered, so we now have a "Restore cache" and a "Post restore cache" that actually stores the cache. Most workflows just seem to use "Cache".

fmoc (approved):
Caching appears to work, the runtime between the last two builds was cut down by over 11 minutes (from 13:21 to 2:02).

PR #8199, opened 2020-10-28
Use new ownCloud theme

TheOneRing:
We need to add the dark colored theme and we need to ensure that the icons are reloaded, maybe 2.7.1


TheOneRing:
Icons that don't get updated on theme change.

The lock:
image
The tabs, and all scroll areas:
image

src/libsync/theme.cpp
```diff
@@ -233,7 +250,7 @@ QString Theme::systrayIconFlavor(bool mono, bool sysTrayMenuVisible) const
         }
 #endif
     } else {
-        flavor = QStringLiteral("colored");
+        flavor = (!Utility::hasDarkSystray() && _hasDarkColoredTheme) ? QStringLiteral("dark") : QStringLiteral("colored");
```
dschmidt, line 253: Can you add a comment here explaining the logic?

dschmidt (approved):

PR #8200, opened 2020-10-29
Fix vfs experimental text

TheOneRing:


src/gui/wizard/owncloudadvancedsetuppage.cpp
```diff
@@ -86,7 +86,7 @@ OwncloudAdvancedSetupPage::OwncloudAdvancedSetupPage()
     }
     _ui.lLocalIcon->setPixmap(Utility::getCoreIcon(QStringLiteral("folder-sync")).pixmap(_ui.lLocalIcon->size()));
 
-    _ui.rVirtualFileSync->setText(tr("Use &virtual files instead of downloading content immediately").arg(bestAvailableVfsMode() == Vfs::WindowsCfApi ? QString() : tr(" (experimental)")));
+    _ui.rVirtualFileSync->setText(tr("Use &virtual files instead of downloading content immediately%1").arg(bestAvailableVfsMode() == Vfs::WindowsCfApi ? QString() : tr(" (experimental)")));
```
michaelstingl, line 89: needs space?
```suggestion
    _ui.rVirtualFileSync->setText(tr("Use &virtual files instead of downloading content immediately %1").arg(bestAvailableVfsMode() == Vfs::WindowsCfApi ? QString() : tr(" (experimental)")));
```

TheOneRing: the space is in the experimental, if we add the space there we will also have a trailing space in the version without experimental

dschmidt (approved):

PR #8205, opened 2020-11-04
Generate theme on the client side, not in ownBrander

TheOneRing:


dschmidt (approved):
A few concerns discussed in chat, but basically looking really good.

Thanks a lot!

dschmidt (approved):

PR #8215, opened 2020-11-09
Do not sync core files

jnweiger:
Linux core files can be huge and are not normally meant to sync. 
The syntax here seems to be globbing patterns.
I'd like `core.*` to match e.g. `core.freecad-daily.1604943483` according to Ubuntu default /proc/sys/kernel/core_pattern = `core.%e.%t`

### DIFF
--- a/sync-exclude.lst
+++ b/sync-exclude.lst
@@ -41,6 +41,5 @@ System Volume Information
 
 My Saved Places.
 
-# do not sync Linux core files. They can come in multiple flavours.
-core
-core.*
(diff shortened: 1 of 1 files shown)


michaelstingl:
Do such files cause problems for the sync process? I can imagine other applications use files with similar names, and I would avoid block those without very serious reason.

jnweiger:
A core file from a CAD package can be several hundred Mbytes. 
This blocks the client from syncing useful files, consumes bandwidth, fills the server's disk and warms up our planet.

jnweiger:
I did not suggest to block a suffix *.core -- and I did not suggest anything with UPPERCASE letters.

TheOneRing:
Too generic for a general ignore rule

PR #8220, opened 2020-11-10
WIP: Fix icon size in activity log

TheOneRing:


TheOneRing:
Hmm actually thi didn't change anything but the issue still seems to be gone, I'll check again with the artifacts.

dschmidt (approved):

PR #8232, opened 2020-11-17
Don't follow redirects on .well-known

TheOneRing:


dschmidt:
Why do we need this? Any related issue?

dschmidt (approved):

dschmidt (approved):

PR #8249, opened 2020-11-23
Don't recurse the whole tree looking for a git dir, use PROJECT_SOURC…

TheOneRing:
…E_DIR


dschmidt (approved):

dschmidt:
While I agree with this change, it probably doesn't solve the issue where the wrong hash was found ... needs to be addressed still.

Issue #8248, opened 2020-11-23
Owncloud - virtual files smaller <1KB - problems with syncing


hdijkema:
### Expected behaviour
Syncing a file <1Kb (827bytes) should sync its 827 bytes to the windows drive when I try to open it and it resides on the owncloud server (virtual files mode).

### Actual behaviour
The local file becomes 0bytes. This is with virtual file support. So, when I try to open the file, which resides on the server and is smaller than 1KB (827 bytes in this case), it first is downloaded, but when I open it it is empty; and when I look it's 0 bytes. When I open it in the web interface, it is oke. When I sync a new file from local to server both files are the same size (client and server), but when I clean up the space and reopen it with e.g. notepad++ or notepad, the syncing starts, and next, notepad(++) is openened with an empty file. If I choose 'make available always', it will sync nicely and when I open the file, it contains the 827 bytes it had. 

### Steps to reproduce
1.  Create a virtual files share.
2.  Put a file of <1KB (827 bytes?) in it, a text file. 
3.  Make room, i.e. cleanup the files to the server
4. Open a file from the cloud by opening it with notepad.
5. Nothing will be shown.

### Server configuration
Operating system: Debian 9
Web server: Apache 2.4, with Reverse proxy.
Database: PostgreSQL 
PHP version: 7.2
ownCloud version: 10.5.0 stable
Storage backend (external storage): not applicable, just normal ext4 file system.

### Client configuration
Client version: 2.7.1
Operating system: Windows 10
OS language: Dutch

Qt version used by client package (Linux only, see also Settings dialog): not applicable

Client package (From ownCloud or distro) (Linux only): not applicable

Installation path of client: Windows, standard

### Logs

? Client or server? Let's just show what happens.

These are the files...

![image](https://user-images.githubusercontent.com/745693/99939529-8d93f780-2d6a-11eb-9338-75cf940d641c.png)

Making room...

![image](https://user-images.githubusercontent.com/745693/99939599-bfa55980-2d6a-11eb-99ae-96eb16c3d5b3.png)

It has been done...

![image](https://user-images.githubusercontent.com/745693/99939648-d3e95680-2d6a-11eb-9089-2f031aa8c2a6.png)

Now double click on the item...and the result...

![image](https://user-images.githubusercontent.com/745693/99939747-157a0180-2d6b-11eb-9f5b-06be0947029d.png)

[...]

PR #8268, opened 2020-12-01
Owncloud - virtual files smaller <1KB - problems with syncing

TheOneRing:
The issue was caused by gziped responses not providing a
content lenght header.

Fixes: #8248


dragotin (approved):
LGTM

PR #8287, opened 2020-12-07
start of BDD UI tests with squish

individual-it:
this PR shows how a automated UI tests could work with [squish](https://www.froglogic.com/squish/)

## things that are nice
- full BDD support
- build-in support for pageObjects https://kb.froglogic.com/squish/howto/using-page-object-test-design-pattern/
- selectors can be found by UI and without knowledge of the code
- documentation (sometimes hard to find, buts its actually there)

## challenges
- dedicated IDE need to be used
- JS support does not go very far, so we went with Python (test also could be written in Perl, Rubby or Tcl)
- setup Python3 was difficult (pip3 libs can be used but don't show up correctly in the IDE)
- no docker image with squish binary (its not hard to create but we would have to talk with frologic if they are happy for us to publish their binary e.g. on hub.docker.com)
- the generated selector names are confusing https://github.com/owncloud/client/pull/8287/files#diff-580ee8d0fac518d9abc9cb198b4c6d45d6c953e84a718d55bc64700bea182326 (but can be edited)
- price https://www.froglogic.com/squish/prices/floating-license/

## general discussion (for the future)
- do we have explicit feature files or do we create client independent feature files that should be the same on all clients?
- ~~this is the third time (after core & web) we would rewrite certain `Given` steps to setup the backend system (create/delete user/group, upload files via webdav, ...) - should we maybe have some GRPC style proxy that receives those steps and does the work? This proxy could be used for all current and future test suites to setup the backend.~~


individual-it:
alternative approach without squish #8408

individual-it:
ToDo: 
-[ ] use https://github.com/owncloud/owncloud-test-middleware to set up the backend system
-[ ] run tests in drone (maybe separate PR)

individual-it:
the current version of this PR needs the test middleware
in some places the steps are exactly the same as in the middleware e.g. ` @Given("user '|any|' has been created with default attributes")`
in other places there are e.g `@Then("the file '|any|' should exist on the server for user '|any|' with following content")`

places that need to know user attributes e.g. the password need to know it now from the feature file e.g. `@Given("user '|any|' has set up a client with these settings and password |any|:")`
the question for me is should we add more endpoints to the middleware to be able to retrieve those things, so that it would be more adaptable and the knowledge would not need to be in the feature file

test/suite_oc-desktop/shared/steps/steps.py
```diff
@@ -0,0 +1,343 @@
...
+    )
+
+@Then("the file '|any|' should exist on the server for user '|any|' with following content")
+def step(context, path, username):
+    fileContent = "\n".join(context.multiLineText)
+    executeStepThroughMiddleware(
+        context,
+        "Then as \"" + username + "\" the file \"" + path + "\" should have the content \"" + fileContent + "\"",
+        "Failed getting file from server or comparing the content"
+    )
```
dpakach, line 243: Since the middleware server is responsible for parsing the step and run the appropriate step definition, I would prefer if there was a common step definition here to execute a step on middleware

test/suite_oc-desktop/shared/steps/steps.py
```diff
@@ -0,0 +1,343 @@
...
+    "step": step}
+    params = json.dumps(body).encode('utf8')
+        
+    req = urllib.request.Request(
+        context.userData['middlewareUrl'] + 'execute',
+        data=params,
+        headers={"Content-Type": "application/json"}, method='POST'
+    )
+    res = urllib.request.urlopen(req)
+    if res.getcode() != 200:
```
dpakach, line 165: if something fails on the middleware it returns the error stack on the body, you can log that here for easier debug

michaelstingl (approved):
LGTM

TheOneRing (approved):
LGTM

test/suite_oc-desktop/shared/steps/steps.py
individual-it: addressed in e1177ee2839256afebec6a85ff2ea492b7fef475

test/suite_oc-desktop/shared/steps/steps.py
individual-it: done in ae37abdb130b835becbbbba6215190f16030803e

dpakach (approved):
Looks good

test/gui/shared/steps/steps.py
```diff
@@ -0,0 +1,334 @@
...
+    waitForObject(names.settings_OCC_SettingsDialog)
+    waitForObjectExists({
+            "column": 1,
+            "container": names.oCC_IssuesWidget_treeWidget_QTreeWidget,
+            "text": RegularExpression(buildConflictedRegex(filename)),
+            "type": "QModelIndex"
+    })
+
+
+@When('user selects the unsynced files tab with |integer| unsynced files')
```
phil-davis, line 329:
```suggestion
@When('the user selects the unsynced files tab with |integer| unsynced files')
```

phil-davis (commented):
This is all minor stuff - I will make a PR to adjust words here and there.

phil-davis:
PR #8492 has the minor changes that I suggested.

PR #8293, opened 2020-12-10
JobQueue

TheOneRing:


src/libsync/jobqueue.h
```diff
@@ -0,0 +1,42 @@
...
+class OWNCLOUDSYNC_EXPORT JobQueue : public QObject
+{
+    Q_OBJECT
+public:
+    JobQueue();
+
+    void block(bool block);
+    bool isBlocked() const;
+    bool retry(AbstractNetworkJob *job);
+    bool queue(AbstractNetworkJob *job);
```
TheOneRing, line 33: enqueue

src/libsync/jobqueue.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include <QObject>
+
+namespace OCC {
+
+class AbstractNetworkJob;
+
+class OWNCLOUDSYNC_EXPORT JobQueue : public QObject
+{
+    Q_OBJECT
```
TheOneRing, line 26: no need for qobject

TheOneRing:
Doesn't work yet...

src/libsync/jobqueue.cpp
```diff
@@ -0,0 +1,168 @@
...
+    }
+}
+
+size_t JobQueue::size() const
+{
+    return _jobs.size();
+}
+
+bool JobQueue::needsRetry(AbstractNetworkJob *job) const
+{
```
dragotin, line 106: I was wondering why this method is in `JobQueue` and not in the `Job` class directly. The job should be the source if retry is needed or not. 

And the implementation indeed only uses job data to find out. So this method could be moved to `Job::needsRetry()`

dragotin (approved):
Apart from the one comment I made (which is minor) I can not see anything suspicious.

Very good work. Makes the code more clear and solves problems :1st_place_medal: 

However, this is a significant change that needs good testing.

src/libsync/jobqueue.cpp
TheOneRing: good point thx

Issue #8298, opened 2020-12-12
Crash on sync .ini files (virtual file system)


efrenbg1:
Constant crash on new client 2.7.3 with virtual file system (.ini files causing crash?).

### Expected behaviour
Client should sync all files.

### Actual behaviour
When syncing client stops working and crashes. Last file synced is .ini and removing it makes it possible to advance the sync process until another .ini files is synced. Both x86 and x86_64 clients crash.

### Steps to reproduce
1. Place .ini file in owncloud server
2. Sync client with virtual file system
3. Wait until client crash

### Server configuration
Operating system: debian 10 (Linux 4.19.0 - amd64)

Web server: nginx/1.14.2

Database: 10.3.25-MariaDB

PHP version: 7.2.34

ownCloud version: 10.5.0

Storage backend (external storage): none

### Client configuration
Client version: 2.7.3

Operating system: Windows 10 Pro 20H2

OS language: Spanish

Qt version used by client package (Linux only, see also Settings dialog): none

Client package (From ownCloud or distro) (Linux only): none

Installation path of client: C:\Program Files (x86)\ownCloud

### Logs
1. Client logfile: 
12-12 11:08:36:078 [ debug sync.networkjob ]	[ OCC::AbstractNetworkJob::slotFinished ]:	Network job OCC::GETFileJob finished for "/10_TRABAJOS/00_ARCHIVAR/1000. TCSD - {sensible data ommited}/Documentación técnica/190906_DOC PARA 5.1, 5.2/5.2.3_Documentación técnica/desktop.ini"
12-12 11:08:36:079 [ debug sync.networkjob.get ]	[ OCC::GETFileJob::slotReadyRead ]:	Actually finished!

2. Web server error log: none

3. Server logfile: ownCloud log (data/owncloud.log): none

jnweiger:
Since client 2.4.0 the name `desktop.ini` is hardcoded as an excluded file. Strange.

jnweiger:
When /Photos/desktop.ini was already synced, and I add Photos/Desktop.ini on the server, we get an error (maybe unrelated to your crash):
```
12.12.2020 04:06:09, Photos/Desktop.ini, ownCloud2,CfCreatePlaceholders error WindowsError: ffffffff80070780: Das System kann auf die Datei nicht zugreifen. : WindowsError: ffffffff80070780: Das System kann auf die Datei nicht zugreifen. for Photos/Desktop.ini

dragotin:
[...]

PR #8299, opened 2020-12-12
Ignore the desktop.ini file in every directory, not only in top dir.

dragotin:
See https://github.com/owncloud/client/issues/8298  for reasons.

### DIFF
--- a/src/csync/csync_exclude.cpp
+++ b/src/csync/csync_exclude.cpp
@@ -201,8 +201,8 @@ static CSYNC_EXCLUDE_TYPE _csync_excluded_common(const QString &path, bool exclu
 #endif
 
     /* We create a Desktop.ini on Windows for the sidebar icon, make sure we don't sync it. */
-    if (blen == 11 && path == bname) {
-        if (bname.compare(QLatin1String("Desktop.ini"), Qt::CaseInsensitive) == 0) {
+    if (blen == 11) {
+        if (bname.compare(QStringLiteral("Desktop.ini"), Qt::CaseInsensitive) == 0) {
             return CSYNC_FILE_SILENTLY_EXCLUDED;
         }
     }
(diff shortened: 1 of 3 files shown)


jnweiger:
The comment `/* We create a Desktop.ini on Windows for the sidebar icon, make sure we don't sync it. */` should be updated.
It refers to one specific file. It does not explain why all Desktop.ini files in any subfolder must be suppressed.

michaelstingl:
> See #8298 for reasons.

Why exclude `desktop.ini` in subfolders? Should be sufficient to fix the crash. `desktop.ini` in subfolders were intentionally allowed in 2.5. (https://github.com/owncloud/client/pull/6404)

dragotin:
The desktop.ini file only concerns the local Explorer. If you sync it and the "other" computer does have a different screen resolution for example, things mit go wrong.

Check this: https://support.microsoft.com/en-us/office/invalid-file-names-and-file-types-in-onedrive-and-sharepoint-64883a5d-228e-48f5-b3d2-eb39e07630fa

OneDrive also does not sync `desktop.ini`. I still think we also should not.

jnweiger:
Imo, it should be always synced, but renamed to a safe name. E.g. desktop.ini._RENAMED

src/csync/csync_exclude.cpp
```diff
@@ -201,8 +201,8 @@ static CSYNC_EXCLUDE_TYPE _csync_excluded_common(const QString &path, bool exclu
 #endif
 
     /* We create a Desktop.ini on Windows for the sidebar icon, make sure we don't sync it. */
```
TheOneRing, line 203: The comment is now misleading.

TheOneRing (approved):

TheOneRing:
Internal ci issues are fixed on 2.7 and master

PR #8301, opened 2020-12-14
Add a changelog entry for a cfapi fix

TheOneRing:


dschmidt (approved):
is VirtualFile a fixed term? Otherwise I'd spell it "virtual file" ... apart from that, ok 👍🏻

PR #8306, opened 2020-12-16
add note to install-doc for windows desktop client

mschreml:
add note to the installation-doc for windows desktop client, on how to find the path for autostart settings under windows.

(Computer\HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Run)

---

benefit: path will be easier to find for users

From issue: https://github.com/owncloud/docs/issues/2322

backport needed for:

2.7
2.6

### DIFF
--- a/docs/modules/ROOT/pages/installing.adoc
+++ b/docs/modules/ROOT/pages/installing.adoc
@@ -209,3 +209,6 @@ The client will attempt to connect to your ownCloud server, and when it is succe
 * One to open your local folder.
 
 It will also start synchronizing your files.
+
+
+NOTE: Path to find the autostart settings for the windows desktop client: `Computer\HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Run`


docs/modules/ROOT/pages/installing.adoc
```diff
@@ -209,3 +209,6 @@ The client will attempt to connect to your ownCloud server, and when it is succe
 * One to open your local folder.
 
 It will also start synchronizing your files.
+
+
+NOTE: Path to find the autostart settings for the windows desktop client: `Computer\HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Run`
```
mmattel, line 214: What is the reaon / drive for this addition?
You already can set the autostart settings in the client...

image

mschreml: I think @voroyam just wanted to show where the settings to this are stored. He is on vacation this week. We can wait with the merge to be sure if you want.

mmattel: Just writing this in the way it is done is very unprofessional und confusing customers. If there are (very) special reasons for a particular audience, the reason and the audience must be declared. Please also see that this is a „Note“ and thererfore highlighted. Ordinary user DO NOT have any reason for using regedit and manipulating keys! In case this is a must have info, this must be rewritten and redesigned.
You can if you want make a proposal change which speeds up the process.

michaelstingl:
@mschreml maybe add it next to the related setting in the GUI:
https://doc.owncloud.com/desktop/navigating.html#general-window

mmattel:
@michaelstingl first it must be defined which audience the info is for and for what purpose. Then rewrite the text and find a appropriate position.

mschreml:
Thank You @michaelstingl @mmattel. 

I'm going to wait for next week and do better when I have more Information ;) Thank You for helping!

voroyam:
where is the follow up question is the issue regarding this?

PR #8309, opened 2020-12-17
Handle errors in convertToPlaceholder

TheOneRing:


dschmidt (approved):
Looking at this patch in isolation ... it looks *very* reasonable, can't judge what error handling it triggers

PR #8310, opened 2020-12-17
OC_REQUIRED_RESULT

TheOneRing:
Depends on https://github.com/owncloud/client/pull/8309/files

### DIFF
--- a/src/common/asserts.h
+++ b/src/common/asserts.h
@@ -3,6 +3,13 @@
 
 #include <qglobal.h>
 
+#if 0 && defined(Q_CC_MSVC)
+// requires c++2017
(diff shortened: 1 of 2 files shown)


src/common/asserts.h
```diff
@@ -3,6 +3,13 @@
 
 #include <qglobal.h>
 
+#if 0 && defined(Q_OS_WIN)
```
dschmidt, line 6: `Q_CC_MSVC`? 

https://infektor.net/posts/2017-01-19-using-cpp17-attributes-today.html#attribute-detection

TheOneRing: stimmt

PR #8314, opened 2020-12-18
Fix a possible crash with the remove all files dialog

TheOneRing:


dschmidt (approved):

GeraldLeikam:
Confirmed fixed with client 2.7.5rcX on [win10/linux mint20]

PR #8322, opened 2020-12-21
Ensure that our dialogs are modal

TheOneRing:


dschmidt (changes requested):
This introduces separate code paths for development and production builds.

We agreed in a call that you remove all custom handling and only keep the assert.

Issue #8335, opened 2020-12-29
2.7.4 build failure at cmake generate stage: multiple cmake errors for Qt5::DBus not found


GvMariani:
Trying to build ownCloud client 2.7.4, to create an updated rpm package for the future release of Rosa Linux (Fresh R12), using cmake 3.19.2 and extra-cmake-modules 5.75.0.

### Expected behaviour
Successful build of ownCloud client.

### Actual behaviour
The build fails while doing the cmake generate step, in a number of places.
The majority of them(31) is related to tests, eg:
`CMake Error at /usr/share/ECM/modules/ECMAddTests.cmake:89 (add_executable):
  Target "OAuthTest" links to target "Qt5::DBus" but the target was not found.  Perhaps a find_package() call is missing for an IMPORTED target, or an ALIAS target is missing?`
Then there are 5 errors involving various client components: src/libsync/CMakeLists.txt, src/libsync/vfs/suffix/CMakeLists.txt, src/gui/CMakeLists.txt (2 times) and src/cmd/CMakeLists.txt.

Since cmake is finding the installed Qt5DBus library, something is missing elsewhere...
See the full log for all the gory details.

### Steps to reproduce
1. Install the sources
2. Run cmake (see the log for the command line used)
3. Watch the errors

### Server configuration
N/A

### Client configuration
Client version: 2.7.4.2934
Operating system: Linux - Rosa Linux Fresh R12 x86_64
OS language: english
Qt version used by client package (Linux only, see also Settings dialog): 5.15.2
Client package (From ownCloud or distro) (Linux only): distro
Installation path of client: /usr/bin

### Logs
Full log of a failed build: https://gist.github.com/GvMariani/ec1bd291b4fa4f861b4fcf3eb51f6150

TheOneRing:
Will be fixed with https://github.com/owncloud/client/issues/8335 but in general it looks like you are missing a dependency.

GvMariani:
> Will be fixed with #8335 but in general it looks like you are missing a dependency.
What one?
As I already wrote, Qt5DBus devel files (and all the other qt5 dev files AFAICT) are installed on the system...

TheOneRing:
Can you share your CMakeCache.txt

GvMariani:
Sure...
[CMakeCache.txt](https://github.com/owncloud/client/files/5810320/CMakeCache.txt)

fmoc:
I can kind of "reproduce" this issue on an Ubuntu 18.04 machine that:

[...]

PR #8337, opened 2021-01-04
CMake: Only link Qt5::DBus if it was found

TheOneRing:


TheOneRing:
https://github.com/owncloud/client/issues/8335

dschmidt (approved):

PR #8341, opened 2021-01-05
Fix warnings

TheOneRing:


shell_integration/windows/OCContextMenu/OCContextMenu.cpp
```diff
@@ -72,7 +72,7 @@ IFACEMETHODIMP_(ULONG) OCContextMenu::Release()
 
 // Initialize the context menu handler.
 IFACEMETHODIMP OCContextMenu::Initialize(
-    LPCITEMIDLIST pidlFolder, LPDATAOBJECT pDataObj, HKEY hKeyProgID)
+    LPCITEMIDLIST, LPDATAOBJECT pDataObj, HKEY)
```
dschmidt, line 75: We probably won't find a consensus here ...
but I would like to point out that I think it's a bad idea to remove parameter names here ... I would prefer `Q_UNUSED` or adding `/* */ ` around the parameter names.

shell_integration/windows/OCContextMenu/OCContextMenu.cpp
```diff
@@ -88,7 +88,7 @@ IFACEMETHODIMP OCContextMenu::Initialize(
         HDROP hDrop = static_cast<HDROP>(GlobalLock(stm.hGlobal));
         if (hDrop) {
             UINT nFiles = DragQueryFile(hDrop, 0xFFFFFFFF, NULL, 0);
-            for (int i = 0; i < nFiles; ++i) {
+            for (UINT i = 0; i < nFiles; ++i) {
```
dschmidt, line 91: Why in caps?

dschmidt (approved):

shell_integration/windows/OCContextMenu/OCContextMenu.cpp
TheOneRing: I just use the same type as nFIles

shell_integration/windows/OCContextMenu/OCContextMenu.cpp
TheOneRing: Anyone who touches the code will need to open the doc anyhow

Issue #7646, opened 2019-12-11
DB-WAL keeps growing


martupeppeadmin:
Since I have updated to 2.6 the db-wa file keeps growing and it's not cleared after a sync session.

theclem54:
This issue has also been identified in a couple of posts at ownCloud Central
https://central.owncloud.org/t/large-sync-db-wal-file-18gb/23085/3
https://central.owncloud.org/t/huge-db-wal-file-size-218gb-most-of-the-times-is-400gb/23902

I have found that if I quit ownCloud client, the db-wal file is removed.

I am using Windows Embedded on the system on which I have personally encountered a full drive issue and on Windows 10 I have also confirmed that quitting ownCloud client removes the file.

Expected behaviour is that the file is cleaned/removed at some regular interval, depending on it's function. If the file is not needed between multiple synchronisations, then it should be removed as soon as a synchronisation is completed.

jnweiger:
https://www.sqlite.org/wal.html suggests that sqlite checkpointing should be performed periodically. either automtically or manually. PRAGMA schema.wal_checkpoint;

NicolasGoeddel:
Any chance that this can be fixed soon? Our wal file grows to a few gigabytes after just a month. I then have to stop the client and start it again which takes over half an hour, where 20 minutes is just the shutdown itself. The other 10 minutes is the discovery of the file system and creating those inotify watchers.

jnweiger:
Confirmed fixed with testpilotcloud 2.9.0 beta3 on win10

* the wal file never grows beyond 4052KB
* This size was quickly reached by repeatedly sharing and unsharing a gallery with 80 photos with this user.

PR #8364, opened 2021-01-15
Always reset prepared sql statements

TheOneRing:
This allow the creation of checkpoints and fixes the growing wal issue
We might need to play with the wal_autocheckpoint to improve the performance 

Fixes: #7646


TheOneRing:
I cant measure any significant difference when comparing my test run.

TheOneRing:
Also have a look on https://github.com/owncloud/client/pull/8372

TheOneRing:
#8372

PR #8372, opened 2021-01-18
Manage prepared sql

TheOneRing:


TheOneRing:
Basically an alternative solution to https://github.com/owncloud/client/pull/8364

src/common/syncjournaldb.cpp
```diff
@@ -953,20 +958,26 @@ bool SyncJournalDb::deleteFileRecord(const QString &filename, bool recursively)
...
-            return false;
+        {
+            const auto query = _queryManager.get(PreparedSqlQueryManager::DeleteFileRecordPhash, QByteArrayLiteral("DELETE FROM metadata WHERE phash=?1"), _db);
+            if (!query) {
+                return false;
+            }
 
-        qlonglong phash = getPHash(filename.toUtf8());
-        _deleteFileRecordPhash.bindValue(1, phash);
+            qlonglong phash = getPHash(filename.toUtf8());
```
dragotin, line 967: the method returns qint64, which is probably the same as qlonglong, but should probably be consistent at some point.

dragotin (approved):
This one is hard to review. But LGTM.

PR #8381, opened 2021-01-21
Fix theme detection with branded clients

TheOneRing:


TheOneRing:
The changes to the enum will require a change to ownbrander but I really did not like the names I came up with in the fist try

PR #8384, opened 2021-01-21
Fix branding fallback

TheOneRing:
https://github.com/owncloud/ownbrander/pull/961
https://github.com/owncloud/ownbrander/pull/962


src/libsync/theme.cpp
```diff
@@ -129,39 +158,51 @@ QString Theme::configFileName() const
...
+    return loadIcon(QStringLiteral("universal"), name, iconType);
+}
+
+QIcon Theme::themeTrayIcon(const QString &name, bool sysTrayMenuVisible, IconType iconType) const
+{
+    auto &icon = loadIcon(systrayIconFlavor(_mono, sysTrayMenuVisible), name, iconType);
+#ifdef Q_OS_MAC
+    // This defines the icon as a template and enables automatic macOS color handling
+    // See https://bugreports.qt.io/browse/QTBUG-42109
+    icon.setIsMask(_mono && !sysTrayMenuVisible);
```
dschmidt, line 188: Maybe worth checking whether it's correct to use `setIsMask` on the reference or if it is better to use it on the copy that is returned later on.

src/libsync/theme.h
```diff
@@ -118,7 +119,15 @@ class OWNCLOUDSYNC_EXPORT Theme : public QObject
...
-    bool isUsingDarkTheme(IconType fallbackType = IconType::BrandedIcon) const;
+    bool isUsingDarkTheme() const;
+
+    /**
+    * Whether the branding allows the dark theme
+    */
+    bool allowDarkTheme() const
+    {
+        return _allowDarkTheme;
+    };
```
dschmidt, line 130: -> .cpp

src/libsync/theme.h
```diff
@@ -420,11 +430,20 @@ class OWNCLOUDSYNC_EXPORT Theme : public QObject
...
+    bool hasTheme(IconType type, const QString &theme) const;
+
     static Theme *_instance;
-    bool _mono;
+    bool _mono = false;
 #ifndef TOKEN_AUTH_ONLY
-    bool _hasDarkColoredTheme = hasTheme(QStringLiteral("dark"));
-    mutable QHash<QString, QIcon> _iconCache;
+    mutable QMap<QString, QIcon> _iconCache;
+    // <is vanilla, theme name>
```
dschmidt, line 441: maybe some more doc what maps to what here and what it reflects

dschmidt (approved):

Issue #8379, opened 2021-01-19
Mac: Fix for #8149 causes issues when display scaling is used


TheOneRing:
![Clipboard - 19  Januar 2021 14_53](https://user-images.githubusercontent.com/200626/105044877-84bc5b00-5a67-11eb-9f44-7d6ef54a24c6.png)

TheOneRing:
#8149

TheOneRing:
Scaling:
![image](https://user-images.githubusercontent.com/200626/105046312-3ad47480-5a69-11eb-83fc-1a9eec053024.png)

PR #8387, opened 2021-01-22
Fix typo

TheOneRing:
#8379 was unrelated to the dpi fix, due to the typo the 32x32 icon was missing and Qt decided to do something with the 16x16 icon.


michaelstingl (approved):
Great catch 👍  - looks better know!

PR #8397, opened 2021-01-27
Implement a central credential manager

TheOneRing:


src/libsync/creds/oauth.cpp
```diff
@@ -79,23 +78,16 @@ class RegisterClientJob : public QObject
...
-                QJsonParseError error;
-                const auto doc = QJsonDocument::fromJson(keychainJob->binaryData(), &error);
-                if (error.error == QJsonParseError::NoError) {
-                    rgisterClientFinished(doc);
-                    return;
-                }
-                qCWarning(lcOauth) << Q_FUNC_INFO << error.errorString();
-            } else if (keychainJob->error() != QKeychain::EntryNotFound) {
-                qCCritical(lcOauth) << "Failed to read client id" << keychainJob->errorString();
+        auto job = _account->credentialManager()->getCredentials(clientSecretC());
```
dragotin, line 81: Why do export the credentialsManager() at all? And not use `_account->getCredentials(...)` ?

dragotin (commented):
Only one little comment. Nice stuff so far.

src/libsync/creds/oauth.cpp
TheOneRing: Hm instead of exporting the credential manager provide a set and get method? We wouldn't really gain much?

src/libsync/account.cpp
```diff
@@ -40,10 +41,12 @@ Q_LOGGING_CATEGORY(lcAccount, "sync.account", QtInfoMsg)
 
 Account::Account(QObject *parent)
     : QObject(parent)
+    , _uuid(QUuid::createUuid())
```
dschmidt, line 44: Why is a new Uuid created whenever an account object is constructed?

src/libsync/creds/credentialmanager.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+Q_LOGGING_CATEGORY(lcCredentaislManager, "sync.credentials.manager", QtDebugMsg)
+
+namespace {
+QString credentialKeyC()
+{
+    return QStringLiteral("%1_credentials").arg(Theme::instance()->appName());
+}
+
+QString accoutnKey(const Account *acc)
```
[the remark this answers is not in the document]
dschmidt: still typo

dschmidt (approved):

src/libsync/account.cpp
dschmidt: afaict `acc->_uuid = settings.value(userUUIDC(), acc->_uuid).toUuid();` this is where it's overridden anyway. 
Maybe we should leave it empty/invalid and fallback to QUuid::createUuid() where it's overridden

TheOneRing: But a newly created account object that is not loaded from settings should also have one, thats why I generate one.
I could also make it lazy and generate it in the getter but...

PR #8398, opened 2021-01-27
Remove PollJob support

TheOneRing:


guruz:
Would have been nice to keep it :-( But also good to see code cleaner

TheOneRing:
In the current implementation remotePermissions are not properly handled.
Which can lead to db entries without a valid permission string which can lead to the following enforce crash https://github.com/owncloud/client/blob/7d1107932b450a29aeea9734ce18fca7c1c699f5/src/libsync/discoveryphase.cpp#L177

I discovered it in https://github.com/owncloud/client/pull/8076 where I tried to prevent empty permissions.

For oci a different approach is planned.

dschmidt (approved):

PR #8404, opened 2021-01-30
Properly escape response texts in socket api

### LINKED ISSUE #8399: [QA] socket api lacks string quoting
This one I just spotted when looking at @GeraldLeikam 's screen:

image

Widgets are reported in a comma-separated list. String labels are included verbatim per widget and these strings may contain a comma too. Example highlighted in red.
Expected behaviour: Escaping/quoting quoting so that the list can be properly parsed.

jnweiger:
The characters `|` and `:` are reserved characters in the protocol. In case of the ASYNC_LIST_WIDGETS command, a comma separated list is returned, so `,` is also a reserved character.
With the suggested url-percent-encoding, the replacement sequences for `| : ,` are `%7C %3A %2C`
This also ensures that no unexpected linebreaks or other control characters could violate the protocol.

Fixes #8399 
Other commands may also be affected. Not checked.

### DIFF
--- a/src/gui/socketapi.cpp
+++ b/src/gui/socketapi.cpp
@@ -91,11 +91,15 @@ QObject *findWidget(const QString &queryString, const QList<QWidget*> &widgets =
 
     QList<QObject*>::const_iterator foundWidget;
 
+    qCDebug(lcSocketApi) << "findWidget ...";
     if (queryString.contains('>')) {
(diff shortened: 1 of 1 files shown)


TheOneRing:
Other socket functions use '\x1e' as list separator so we should probably align that.

PR #8408, opened 2021-02-02
[WIP] POC socket based GUI tests

individual-it:
alternative approach to #8287

this is based on the work in https://gitea.owncloud.services/client/build-linux
for now I've striped it down to only the nessecery parts to run the tests locally, docker images, builds on different systems, VNC, etc. can be added later here as well as in #8287

run tests locally:
`CLIENT_SYNC_PATH=<full path of the folder the client should connect to> BACKEND_HOST=<URI of oC server with trailing slash> yarn run acceptance-tests features/addAccounts.feature:7`

advantages:
 - no squish needed
 - seem to run faster (because mouse don't need to be moved)
 - can reuse code from web tests because the steps are written in JS using cucumber-js
 - TBC

disadvantages (at the current stage of the test infrastructure development):
 - how to find names of the elements (@TheOneRing you mentioned some tool that could help with that, can you give me a hint?)
 - the state of the app is also only queried by the socket API, so the app tell the tests what state its in, no check from the outside
 - probably more work is needed to add test-functionality that otherwise squish would provide
 - typing text into input boxes does not trigger the events that are triggered when a user is typing the same text see https://github.com/owncloud/client/pull/8408#issuecomment-777234801
 - cannot identify unnamed widgets see discussion in the chat https://chat.owncloud.com/channel/client-dev?msg=x7rn5QC5CLsoGnmFd
 - TBC


jnweiger:
Hints for how to find the names:
* `echo 'ASYNC_LIST_WIDGETS:42|' | netcat -U -w1 /run/user/$UID/ownCloud/socket`
* https://github.com/KDAB/GammaRay

test/gui/features/support/rsvp-party/LICENSE
```diff
@@ -0,0 +1,17 @@
+Code is taken from https://github.com/dschmidt/rsvp-party/tree/the-cleanup
+
+(Forked from https://github.com/stefanpenner/rsvp-party/tree/the-cleanup)
+
+ISC-License:
+
+Permission to use, copy, modify, and/or distribute this software for any
+purpose with or without fee is hereby granted, provided that the above
+copyright notice and this permission notice appear in all copies.
```
jnweiger, line 9: A copyright notice is missing.

TheOneRing:
Yes the tool is gammaray, I can easily add it to our prepared binaries in our sdk.
https://github.com/owncloud/ownbuild

individual-it:
Triting text into the sharing dialog seems not to trigger the event to display the user list, so the sharee can not be selected

https://user-images.githubusercontent.com/2425577/107608579-76c3b980-6c64-11eb-8d2f-c4a5e00b06aa.mp4

individual-it:
as discussed in the meeting on 16.02.2021 closing this in favour of #8287

PR #8422, opened 2021-02-08
Fix multiple modal dialogs

TheOneRing:


TheOneRing:
Doesnt work...

TheOneRing:
I'd rather hope for an upstream fix https://bugreports.qt.io/browse/QTBUG-91059

PR #8406, opened 2021-02-01
Update removing.adoc

voroyam:
added the location of the configuration file in case user want to remove the client completely.

PR #8430, opened 2021-02-10
Update removing.adoc

mmattel:
Superseeds #8406 using the correct branch 2.7 (instead of master)

Added the location of the configuration file in case user wants to remove the client completely.

Backporting to 2.6 and master


TheOneRing (approved):

mmattel:
@TheOneRing: info: you said before, that 2.6 is closed. But these changes are client docs only and 2.6 is part of official docs. I please you therefore to backport (or however you do it) these changes to 2.6 too. Else this looks quite inconsistant. Many thanks.

PR #8433, opened 2021-02-12
additional logging option: CLI

enbrnz:
on Windows you need to open a powershell
on Linux just a standard terminal

Just a quick addition to the docs. Wording can probably be improved

### DIFF
--- a/docs/modules/ROOT/pages/troubleshooting.adoc
+++ b/docs/modules/ROOT/pages/troubleshooting.adoc
@@ -121,7 +121,10 @@ As an example, to define a test where you keep log data for two days, you can is
 If the ownCloud client isn't able to start and immediately crashes the first two options are not available. 
 Therefore it might need to be necessary to start the ownCloud client using the command line in order to be see the error message
 
-On Linux and Mac simply open the terminal and run `owncloud`.
+On Linux and Mac simply open the terminal and run: 
+```
+owncloud --logfile - --logflush`.
+```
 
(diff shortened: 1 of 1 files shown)


michaelstingl (approved):
LGTM, but I’d prefer to wait for @TheOneRing ‘s approval

docs/modules/ROOT/pages/troubleshooting.adoc
```diff
@@ -118,6 +118,21 @@ owncloud --logdir /tmp/owncloud_logs --logexpire 48
...
 Adding the `--logdebug` flag increases the verbosity of the generated log files.
 
+==== Logging in the Console
+
+If the ownCloud client isn't able to start and immediately crashes the first two options are not available. 
+Therefore it might need to be necessary to start the ownCloud client using the command line in order to be see the error message
+
+On Linux and Mac simply open the terminal and run `owncloud`.
+
+On Windows open a PowerShell and run the following command:
```
TheOneRing, line 128: --logfile - --logflush is also recommended on unix we will just get a more detailed log.

--logdebug will create a more detailed log.

The pipe on Windows is required as usually no console is attached to a gui process so the output is lost.

enbrnz: Thanks for your input!
Better like that?

TheOneRing:
I rebased the change to 2.7 so we will get it in 2.7 and master

TheOneRing (approved):

PR #8434, opened 2021-02-12
Fix Windows long path issue introduced in …

TheOneRing:
dd641fae997d71c8396b77def2fa25ad96fdf47f

Test on master https://github.com/owncloud/client/pull/8435

PR #8435, opened 2021-02-12
    Add test for csync_vio_local_stat with long path

TheOneRing:
Depends on https://github.com/owncloud/client/pull/8434


TheOneRing:
https://github.com/owncloud/client/actions/runs/561167866 was with a fix for the bug 
https://github.com/owncloud/client/actions/runs/561170464 is the unit test without the fix

dschmidt (approved):

PR #8437, opened 2021-02-15
Allow to trigger a  backup sync using the socket api

TheOneRing:
Test app output:
```    
-> V2/LIST_ACCOUNTS
<- V2/ACCOUNTS:{"accounts":[{"id":"0","name":"admin@localhost"},{"id":"1","name":"admin@demo.owncloud.org"}]}
```

```
-> V2/UPLOAD_FILES_FROM:{"account":{"id":"0","name":"admin@localhost"},"localPath":" "C:/user/hanna","pattern":".*\\.stl","remotePath":".snapshots/19 Feb 2021 13:31:05 +0000"}

<- V2/UPLOAD_FILES_RESULT:{"error":"","localPath":"C:/Users/hanna","syncedFiles":["Desktop/untitled2.stl"]}

### CODE REVIEW

**Comment by TheOneRing** [comment]
Is camel case in json a bad idea?

File: src/gui/socketapi/socketuploadjob.cpp [hunk 1]
```diff
@@ -0,0 +1,90 @@
...
+SocketUploadJob::SocketUploadJob(OCC::SocketListener *listener, const QString &argument, QObject *parent)
+    : QObject(parent)
+    , _listener(listener)
+{
+    const auto args = QJsonDocument::fromJson(argument.toUtf8()).object();
+    _localPath = args[QLatin1String("localPath")].toString();
+    _remotePath = args[QLatin1String("remotePath")].toString();
+    _pattern = args[QLatin1String("pattern")].toString();
+    // TODO: use uuid
+    const auto accname = args[QLatin1String("account")][QLatin1String("name")].toString();
```
dragotin, line 37: Nitpick: QLatin1String -> QStringLiteral ?

src/libsync/owncloudpropagator.cpp
```diff
@@ -369,15 +370,36 @@ qint64 OwncloudPropagator::smallFileSize()
     return smallFileSize;
 }
 
-void OwncloudPropagator::start(const SyncFileItemVector &items)
+void OwncloudPropagator::start(const SyncFileItemVector &_items)
```
dragotin, line 373: is that underscore intentional? I usually use the underscore for member variables only.

dragotin (approved):
Awesome work!

src/gui/socketapi/socketuploadjob.cpp
TheOneRing: Json has an overload and as the string is not actually used, besides the lookup, QLatin1String should be slightly better 🙈

src/libsync/owncloudpropagator.cpp
TheOneRing: I create a copy one line later and well...

src/gui/socketapi/socketapi_p.h
```diff
@@ -29,6 +29,13 @@
 
 namespace OCC {
 
+namespace SoketApiHelper {
```
dschmidt, line 32: `SocketApiHelper`? ;-)

src/gui/socketapi/socketapi.cpp
```diff
@@ -866,8 +882,7 @@ void SocketApi::sendSharingContextMenuOptions(const FileData &fileData, SocketLi
     // If sharing is globally disabled, do not show any sharing entries.
     // If there is no permission to share for this file, add a disabled entry saying so
     if (isOnTheServer && !record._remotePerm.isNull() && !record._remotePerm.hasPermission(RemotePermissions::CanReshare)) {
-        listener->sendMessage(QLatin1String("MENU_ITEM:DISABLED:d:") + (!record.isDirectory()
-            ? tr("Resharing this file is not allowed") : tr("Resharing this folder is not allowed")));
+        listener->sendMessage(QLatin1String("MENU_ITEM:DISABLED:d:") + (!record.isDirectory() ? tr("Resharing this file is not allowed") : tr("Resharing this folder is not allowed")));
```
dschmidt, line 885: This looks like a "Verschlimmbesserung". 
maybe break it up into two real lines instead of a single statement.

src/gui/socketapi/socketuploadjob.cpp
dschmidt: https://forum.qt.io/topic/78540/qstringliteral-vs-qlatin1string/3

I think in the end it doesn't matter. This is unlikely to ever become a "hot path"

src/gui/socketapi/socketuploadjob.h
```diff
@@ -0,0 +1,49 @@
...
+
+private:
+    SocketListener *_listener;
+    QString _localPath;
+    QString _remotePath;
+    QString _pattern;
+    QTemporaryFile tmp;
+    SyncJournalDb *db;
+    SyncEngine *engine;
+    QStringList syncedFiles;
```
dschmidt, line 45: Why do the last 4 ones lack the `_` prefix? Confused me a lot during review because I was wondering where these came from.

src/libsync/syncoptions.cpp
```diff
@@ -53,3 +56,20 @@ void SyncOptions::verifyChunkSizes()
     _minChunkSize = qMin(_minChunkSize, _initialChunkSize);
     _maxChunkSize = qMax(_maxChunkSize, _initialChunkSize);
 }
+
+const QSharedPointer<QRegularExpression> &SyncOptions::fileRegex() const
```
dschmidt, line 60: QRegularExpression is implicitly shared, why do we need a shared pointer?

dschmidt (commented):
What happens, when the remote path is not empty? 

Should we define what should happen and possibly write a test for that? 🙈

src/gui/socketapi/socketapi.cpp
TheOneRing: Clang format does that for some reason and as I moved the file it was touched...

src/gui/socketapi/socketuploadjob.h
TheOneRing: https://code.woboq.org/qt5/qtbase/src/corelib/text/qstring.h.html#QLatin1Literal

src/gui/socketapi/socketuploadjob.cpp
TheOneRing: https://code.woboq.org/qt5/qtbase/src/corelib/text/qstring.h.html#QLatin1Literal

src/gui/socketapi/socketuploadjob.cpp
```diff
@@ -0,0 +1,90 @@
...
+        return;
+    }
+    _engine->setSyncOptions(opt);
+
+    // create the dir, don't fail if it already exists
+    auto mkdir = new OCC::MkColJob(_engine->account(), _remotePath);
+    connect(mkdir, &OCC::MkColJob::finishedWithoutError, _engine, &OCC::SyncEngine::startSync);
+    connect(mkdir, &OCC::MkColJob::finishedWithError, this, [this](QNetworkReply *reply) {
+        if (reply->error() == 202) {
+            // folder does exist, we could also assert here if we want to enforce a backup over a sync
```
TheOneRing, line 74: I could fail if the dir already exist as we don't really know what will happen in that case...

src/gui/socketapi/socketapi.cpp
dschmidt: Well then move `(!record.isDirectory() ? tr("Resharing this file is not allowed") : tr("Resharing this folder is not allowed"))` to a variable?

src/gui/socketapi/socketuploadjob.cpp
dschmidt: We don't know?

TheOneRing: Well I expect unintended things, but I guess it will mainly depend on the state of the remote and the local folder...

src/gui/socketapi/socketapi.cpp
TheOneRing: I didn't change that line I can also try to revert the format change....

dschmidt (approved):

PR #8440, opened 2021-02-16
Prevent copy of PathTuple

TheOneRing:
As far as I can tell PathTuple was not passed as a const reference to for some special optimizations.
This change is meant to make the behavior more explicit.


dschmidt (commented):
🤷🏻

dschmidt:
Why did you close in the end? Maybe worth recording it here...

TheOneRing:
I plan to go for const ref and explicitly copy in the two cases it is needed

PR #8446, opened 2021-02-19
Don't crash when the listener disconnected

TheOneRing:
Depends on: https://github.com/owncloud/client/pull/8437

In an async scenario the listener can disconnect before we answer


dschmidt (approved):
Cool, I think I like this :-D

PR #8469, opened 2021-03-03
Dump the last 20 lines of logs to a file when we crash

TheOneRing:
I'll also modify the crash reporter once this is discussed.

image


dragotin (approved):
Nice! Again nothing to moan about from my side ;-)

PR #8471, opened 2021-03-04
Always call doLog even if we don't have a file stream

TheOneRing:
The only overhead added is the call to qFormatMessage... and we lock one time less if one is attached.

Depends on #8470


dschmidt (approved):
While at it, should we get rid of NO_MSG_HANDLER? 

if we need the feature, it should be a runtime flag/env var probably...

PR #8480, opened 2021-03-09
add autostart to troubleshooting

voroyam:
instruction what to do when disabling the autostart feature does not work.

removed the note about the server log file [it was a note to self from Matthew]


voroyam:
@michaelstingl need review

voroyam:
> How about
> image

How about reading what I wrote?

TheOneRing:
In that case we should debug why it does not work

TheOneRing:
I don't think we need to document a temporary workaround.
Can you open an issue for this instead and document what and how it is broken?

PR #8482, opened 2021-03-11
Instead of renaming the targets, we rename their output 

TheOneRing:
This fixes brandings with spaces in the name

Do we need a changelog entry for that?

### DIFF
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -129,7 +129,7 @@ endif()
 add_library(owncloudCore STATIC ${final_src})
 set_target_properties(owncloudCore PROPERTIES AUTOUIC ON AUTORCC ON)
 target_link_libraries(owncloudCore PUBLIC Qt5::Widgets Qt5::Network Qt5::Xml
-                                          ${synclib_NAME})
+                                          libsync)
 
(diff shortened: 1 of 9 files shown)


src/gui/CMakeLists.txt
```diff
@@ -189,37 +189,40 @@ if(WITH_CRASHREPORTER)
...
 endif()
 
-add_executable(${APPLICATION_EXECUTABLE} main.cpp)
-set_target_properties(${APPLICATION_EXECUTABLE} PROPERTIES AUTOUIC ON AUTORCC ON)
-target_link_libraries( ${APPLICATION_EXECUTABLE} owncloudCore )
+add_executable(owncloud main.cpp)
+set_target_properties(owncloud PROPERTIES
+    OUTPUT_NAME "${APPLICATION_EXECUTABLE}"
+    AUTOUIC ON
+    AUTORCC ON)
```
dschmidt, line 196: ultra-nitpick: maybe put ) on the next line

dschmidt (approved):

jnweiger:
@TheOneRing do you have an example branding with a whitespace?

TheOneRing:
"own Cloud" its for the .exe name and the dll names

PR #8493, opened 2021-03-16
Make language configurable

### LINKED ISSUE #8466: Add language picker to the cli and the ui
Implement a language picker to the ci.
The selected language is stored in the settings and loaded at start up.
Additionally add a `--language` switch to the command line interface which temporarily overrides the selection.

fmoc:
Closes #8466.

This PR adds two new methods to select the language within the GUI application: a config option in the general settings page, and a `--language` parameter.

screenshot_2021-03-16_17-29-30

The current implementation of listing all available locales relies on the translation file's filenames. It extracts the locale part and tries to resolve the native language name using `QLocale::nativeLanguageName`. This seems to work relatively well, except for one file called `TW`. I suspect we have to rename that file to `tw`.

### DIFF
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -86,6 +86,7 @@ set(client_SRCS
     servernotificationhandler.cpp
     guiutility.cpp
     elidedlabel.cpp
+    translations.cpp
     creds/httpcredentialsgui.cpp
     wizard/postfixlineedit.cpp
     wizard/abstractcredswizardpage.cpp
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -698,11 +698,15 @@ void Application::setupTranslations()
 {
     QStringList uiLanguages = QLocale::system().uiLanguages();
 
+    // the user can also set a locale in the settings, so we need to load the config file
+    ConfigFile cfg;
+
     // allow user and theme to enforce a language via a commandline parameter
     const auto themeEnforcedLocale = Theme::instance()->enforcedLocale();
-    // note that the user enforced language is prioritized over the theme enforced one
+    // note that user-enforced languages are prioritized over the theme enforced one
+    // to make testing easier, --language overrides the setting from the config file
     // as we are prepending to the list of languages, the list passed to the loop must be sorted with ascending priority
-    for (const auto &enforcedLocale : { themeEnforcedLocale, _userEnforcedLanguage }) {
+    for (const auto &enforcedLocale : { themeEnforcedLocale, cfg.uiLanguage(), _userEnforcedLanguage }) {
         if (!enforcedLocale.isEmpty()) {
             uiLanguages.prepend(enforcedLocale);
         }
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -42,6 +42,7 @@
 #include "csync_exclude.h"
 #include "common/vfs.h"
 #include "settingsdialog.h"
+#include "translations.h"
 
 #include "config.h"
 
@@ -88,23 +89,6 @@ namespace {
             "  --logdebug           : also output debug-level messages in the log.\n"
             "  --language <locale>  : override UI language\n"
             "  --confdir <dirname>  : Use the given configuration folder.");
-    }
-
-    QString applicationTrPath()
-    {
-        QString devTrPath = qApp->applicationDirPath() + QString::fromLatin1("/../src/gui/");
-        if (QDir(devTrPath).exists()) {
-            // might miss Qt, QtKeyChain, etc.
-            qCWarning(lcApplication) << "Running from build location! Translations may be incomplete!";
-            return devTrPath;
-        }
-#if defined(Q_OS_WIN)
-        return QApplication::applicationDirPath();
-#elif defined(Q_OS_MAC)
-        return QApplication::applicationDirPath() + QLatin1String("/../Resources/Translations"); // path defaults to app dir.
-#elif defined(Q_OS_UNIX)
-        return QString::fromLatin1(SHAREDIR "/" APPLICATION_EXECUTABLE "/i18n/");
-#endif
     }
 }
 
@@ -731,8 +715,8 @@ void Application::setupTranslations()
     foreach (QString lang, uiLanguages) {
         lang.replace(QLatin1Char('-'), QLatin1Char('_')); // work around QTBUG-25973
         lang = substLang(lang);
-        const QString trPath = applicationTrPath();
-        const QString trFile = QLatin1String("client_") + lang;
+        const QString trPath = Translations::applicationTrPath();
+        const QString trFile = Translations::translationsFilePrefix() + lang;
         if (translator->load(trFile, trPath) || lang.startsWith(QLatin1String("en"))) {
             // Permissive approach: Qt and keychain translations
             // may be missing, but Qt translations must be there in order
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -74,19 +74,20 @@ namespace {
     const QString optionsC()
     {
         return QStringLiteral(
-        "Options:\n"
-        "  -h --help            : show this help screen.\n"
-        "  -s --showsettings    : show the settings dialog while starting.\n"
-        "  -q --quit            : quit the running instance\n"
-        "  --logfile <filename> : write log output to file <filename>.\n"
-        "  --logfile -          : write log output to stdout.\n"
-        "  --logdir <name>      : write each sync log output in a new file\n"
-        "                         in folder <name>.\n"
-        "  --logexpire <hours>  : removes logs older than <hours> hours.\n"
-        "                         (to be used with --logdir)\n"
-        "  --logflush           : flush the log file after every write.\n"
-        "  --logdebug           : also output debug-level messages in the log.\n"
-        "  --confdir <dirname>  : Use the given configuration folder.");
+            "Options:\n"
+            "  -h --help            : show this help screen.\n"
+            "  -s --showsettings    : show the settings dialog while starting.\n"
+            "  -q --quit            : quit the running instance\n"
+            "  --logfile <filename> : write log output to file <filename>.\n"
+            "  --logfile -          : write log output to stdout.\n"
+            "  --logdir <name>      : write each sync log output in a new file\n"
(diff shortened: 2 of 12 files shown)


michaelstingl:
1. Better move to "Advanced" section?
2. I think we'd need a branding option to hide this. (default=on)
3. Regarding displaying the language name, in oC10, the name of the language is part of the language file itself:
  `    "__language_name__" : "Deutsch (Förmlich: Sie)",`
  (Source: https://github.com/owncloud/core/blob/621a09501d9afd81afcc5f62765b9858ed5a7cce/lib/l10n/de_DE.json#L45)
4. In the desktop sync app, we only have [`client_de.ts`](https://github.com/owncloud/client/blob/master/translations/client_de.ts). With the selection in the UI, we could also make the other German translations available?

TheOneRing:
> 
> 
>     1. Better move to "Advanced" section?

I only would expect technically advanced settings there not ui settings.

> 
>     2. I think we'd need a branding option to hide this. (default=on)

Do you see issues caused by the feature?
I'd rather see it as an accessibility feature.

> 
>     3. Regarding displaying the language name, in oC10, the name of the language is part of the language file itself:
>        `    "__language_name__" : "Deutsch (Förmlich: Sie)",`
>        (Source: https://github.com/owncloud/core/blob/621a09501d9afd81afcc5f62765b9858ed5a7cce/lib/l10n/de_DE.json#L45)
> 
>     4. In the desktop sync app, we only have [`client_de.ts`](https://github.com/owncloud/client/blob/master/translations/client_de.ts). With the selection in the UI, we could also make the other German translations available?

We currently have Austrian and German in here I guess, we could add explicit translations for some of them.
Currently the code is translating the short language code de_DE to its official name.

src/gui/CMakeLists.txt
```diff
@@ -86,6 +86,8 @@ set(client_SRCS
     servernotificationhandler.cpp
     guiutility.cpp
     elidedlabel.cpp
+    translationshelper.h
```
TheOneRing, line 89: don't add the header

TheOneRing (commented):
> 
> 
> > ```
> > 1. Better move to "Advanced" section?
> > ```
> 
> I only would expect technically advanced settings there not ui settings.
> 
> > ```
> > 2. I think we'd need a branding option to hide this. (default=on)
> > ```
> 
> Do you see issues caused by the feature?
> I'd rather see it as an accessibility feature.

I see we have, enforcedLocale but I'd expect that was rather a workaround for a lack of a language selection?

> 
> > ```
> > 3. Regarding displaying the language name, in oC10, the name of the language is part of the language file itself:
> >    `    "__language_name__" : "Deutsch (Förmlich: Sie)",`
> >    (Source: https://github.com/owncloud/core/blob/621a09501d9afd81afcc5f62765b9858ed5a7cce/lib/l10n/de_DE.json#L45)
> > 
> > 4. In the desktop sync app, we only have [`client_de.ts`](https://github.com/owncloud/client/blob/master/translations/client_de.ts). With the selection in the UI, we could also make the other German translations available?
> > ```
> 
> We currently have Austrian and German in here I guess, we could add explicit translations for some of them.
> Currently the code is translating the short language code de_DE to its official name.

src/gui/application.cpp
```diff
@@ -707,9 +698,18 @@ void Application::setupTranslations()
...
+    // the user can also set a locale in the settings, so we need to load the config file
+    ConfigFile cfg;
+
+    // allow user and theme to enforce a language via a commandline parameter
+    const auto themeEnforcedLocale = Theme::instance()->enforcedLocale();
+    // note that user-enforced languages are prioritized over the theme enforced one
+    // to make testing easier, --language overrides the setting from the config file
+    // as we are prepending to the list of languages, the list passed to the loop must be sorted with ascending priority
+    for (const auto &enforcedLocale : { themeEnforcedLocale, cfg.uiLanguage(), _userEnforcedLanguage }) {
+        if (!enforcedLocale.isEmpty())
```
TheOneRing, line 710: I personally prefer to always use {} with if statements

src/gui/generalsettings.cpp
```diff
@@ -52,6 +53,25 @@ GeneralSettings::GeneralSettings(QWidget *parent)
...
+    // initialize map of locales to language names
+    //
+    const auto availableLocales = TranslationsHelper::listAvailableTranslations();
+    for (const auto &availableLocale : availableLocales) {
+        auto nativeLanguageName = QLocale(availableLocale).nativeLanguageName();
+
+        // fallback if there's a locale whose name Qt doesn't know
+        // this indicates a broken filename
+        if (nativeLanguageName.isEmpty()) {
+            qDebug() << "Warning: could not find native language name for locale" << availableLocale;
```
TheOneRing, line 65: Use qCDebug

src/gui/generalsettings.cpp
```diff
@@ -62,6 +82,7 @@ GeneralSettings::GeneralSettings(QWidget *parent)
     connect(_ui->newFolderLimitCheckBox, &QAbstractButton::toggled, this, &GeneralSettings::saveMiscSettings);
     connect(_ui->newFolderLimitSpinBox, static_cast<void (QSpinBox::*)(int)>(&QSpinBox::valueChanged), this, &GeneralSettings::saveMiscSettings);
     connect(_ui->newExternalStorage, &QAbstractButton::toggled, this, &GeneralSettings::saveMiscSettings);
+    connect(_ui->languageDropdown, QOverload<int>::of(&QComboBox::activated), this, &GeneralSettings::slotLanguagePicked);
```
TheOneRing, line 85: maybe be indexChanged?

As slotLanguagePicked is quite short, I'd prefer a lambda here :)

src/gui/generalsettings.cpp
```diff
@@ -52,6 +53,25 @@ GeneralSettings::GeneralSettings(QWidget *parent)
...
+        if (nativeLanguageName.isEmpty()) {
+            qDebug() << "Warning: could not find native language name for locale" << availableLocale;
+            nativeLanguageName = tr("unknown (%1)").arg(availableLocale);
+        }
+
+        localesToLanguageNamesMap.insert(availableLocale, nativeLanguageName);
+    }
+
+    // needs to be called before settings are loaded
+    loadLanguageNamesIntoDropdown();
```
TheOneRing, line 73: Better move the above code also in to the function.

src/gui/generalsettings.cpp
```diff
@@ -138,6 +159,29 @@ void GeneralSettings::loadMiscSettings()
     _ui->newFolderLimitSpinBox->setValue(newFolderLimit.second);
     _ui->newExternalStorage->setChecked(cfgFile.confirmExternalStorage());
     _ui->monoIconsCheckBox->setChecked(cfgFile.monoIcons());
+
+    // we assume the language names have been loaded into the dropdown already (by making sure that the corresponding method is called before this one)
+    const auto &locale = cfgFile.uiLanguage();
```
TheOneRing, line 164: I guess this could also be moved to loadLanguageNamesIntoDropdown?
By now we have 3 separate code parts which in the end all do one thing.

as the loading always happens at the beginning I don't expect any issues by populating the combobox in loadMiscSettings

src/gui/translationshelper.cpp
```diff
@@ -0,0 +1,74 @@
...
+#include <QApplication>
+#include <QLoggingCategory>
+#include <QDir>
+#include <QDirIterator>
+
+namespace OCC {
+
+Q_LOGGING_CATEGORY(lcTranslationsHelper, "gui.translationshelper", QtInfoMsg)
+
+const QString TranslationsHelper::TRANSLATIONS_FILE_PREFIX = "client_";
```
TheOneRing, line 32: Use an anonymous namespace for constants.
```
namespace {
const QString TranslationFilePrefixC() { return QStringLiteral("client_");}
}
```
For legacy reasons our const values end with a bit C even so they are no longer c strings.
Try to avoid global variables https://github.com/KDE/clazy/blob/master/docs/checks/README-non-pod-global-static.md .

src/gui/translationshelper.h
```diff
@@ -0,0 +1,43 @@
...
+
+#ifndef TRANSLATIONSMANAGER_H
+#define TRANSLATIONSMANAGER_H
+
+#include <QSet>
+#include <QString>
+
+namespace OCC {
+
+class TranslationsHelper
```
TheOneRing, line 23: namespace TranslationsHelper

TheOneRing (changes requested):
Looks pretty good :)

fmoc:
> Better move to "Advanced" section?

UX wise, I think it makes sense to make this setting as accessible as possible. The reason you want to change language through the UI is, usually, because an application is using the wrong language by default. Imagine someone has to find this dropdown while the application is in a foreign language they don't understand.

src/gui/generalsettings.cpp
fmoc: So you're suggesting to either call the utility method that populates the dropdown in `loadMiscSettings` or even integrate it here?

src/gui/translationshelper.cpp
fmoc: The prefix value is actually needed in another location as well (`application.cpp`, `setupTranslations`). I forgot to integrate it there so far. Do you want me to create the anonymous namespace in the header then? So far, the namespace method is only used in implementation but not header files.

src/gui/translations.cpp
```diff
@@ -0,0 +1,85 @@
...
+            // might miss Qt, QtKeyChain, etc.
+            qCWarning(lcTranslationsHelper) << "Running from build location! Translations may be incomplete!";
+            return devTrPath;
+        }
+#if defined(Q_OS_WIN)
+        return QApplication::applicationDirPath();
+#elif defined(Q_OS_MAC)
+        return QApplication::applicationDirPath() + QLatin1String("/../Resources/Translations"); // path defaults to app dir.
+#elif defined(Q_OS_UNIX)
+        return QString::fromLatin1(SHAREDIR "/" APPLICATION_EXECUTABLE "/i18n/");
```
TheOneRing, line 57: Please update to QStringLiteral

TheOneRing (approved):

TheOneRing (changes requested):
Changelog entry is missing

TheOneRing (approved):

Issue #7459, opened 2019-09-06
Progress wheel shown when unselecting password protection


HanaGemela:
Client: 2.6.0rc1 (build 12411)
mac 10.14.6
server: 10.2.0 stable

Steps to recreate:
1) Right click on a file 
2) Select Share from ownCloud menu
3) Select Public link tab
4) Tick the Password protect checkbox 
5) Untick the Password protect checkbox 

Actual result: Loading wheel next to the checkbox
Expect result: Checkbox has been just unchecked. No loading wheel

<img width="409" alt="Screenshot 2019-09-06 at 20 27 32" src="https://user-images.githubusercontent.com/49001702/64451985-a7a1d580-d0e5-11e9-9cd5-c87c8fd4e098.png">

michaelstingl:
I'd only touch this when covered by tests…

HariBhandari07:
Test for checking "progress wheel while unselecting password protection checkbox" has been added in PR https://github.com/owncloud/client/pull/8496.

All the remaining public link share test will be covered by https://github.com/owncloud/client/issues/8501

TheOneRing:
The sharing dialog is considered deprecated

PR #8496, opened 2021-03-17
[tests-only] Adds tests for selecting and unselecting password protection checkbox

SwikritiT:
## Description
This PR adds UI tests to select and unselect the password protection box while public sharing to see the status of the progress indicator.

## Related Issue

- https://github.com/owncloud/client/issues/7459

## How Has This Been Tested?

- locally

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [ ] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

### DIFF
--- a/test/gui/shared/steps/steps.py
+++ b/test/gui/shared/steps/steps.py
@@ -155,7 +155,7 @@ def executeStepThroughMiddleware(context, step):
     body = {
     "step": step}
     params = json.dumps(body).encode('utf8')
-        
+
     req = urllib.request.Request(
         context.userData['middlewareUrl'] + 'execute',
         data=params,
@@ -185,7 +185,7 @@ def step(context, receiver, resource, permissions):
         clickButton(waitForObject(names.scrollArea_permissionsEdit_QCheckBox))
     if ('share' in permissionsList and shareChecked == False) or ('share' not in permissionsList and shareChecked == True):
         clickButton(waitForObject(names.scrollArea_permissionShare_QCheckBox))
-    
+
     clickButton(waitForObject(names.sharingDialog_Close_QPushButton))
 
 @Then('user "|any|" should be listed in the collaborators list for file "|any|" with permissions "|any|" on the client-UI')
@@ -194,7 +194,7 @@ def step(context, receiver, resource, permissions):
     socketConnect = syncstate.SocketConnect()
     socketConnect.sendCommand("SHARE:" + resource + "\n")
     permissionsList = permissions.split(',')
-    
+
     test.compare(str(waitForObjectExists(names.scrollArea_sharedWith_QLabel).text), receiver)
     test.compare(waitForObjectExists(names.scrollArea_permissionsEdit_QCheckBox).checked, ('edit' in permissionsList))
     test.compare(waitForObjectExists(names.scrollArea_permissionShare_QCheckBox).checked, ('share' in permissionsList))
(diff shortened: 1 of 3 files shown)


swoichha (commented):
LGTM

test/gui/tst_sharing/test.feature
```diff
@@ -23,3 +27,29 @@ Feature: Sharing
             """
         When the user adds "Brian Murphy" as collaborator of resource "%client_sync_path%/textfile0.txt" with permissions "edit,share" using the client-UI
         Then user "Brian Murphy" should be listed in the collaborators list for file "%client_sync_path%/textfile0.txt" with permissions "edit,share" on the client-UI
+
+    @issue-7459
+    Scenario: Progress indicator should not visible after unselecting the password protection checkbox while sharing through public link
```
Talank, line 32:
```suggestion
    Scenario: Progress indicator should not be visible after unselecting the password protection checkbox while sharing through public link
```

Talank (commented):
Just a small suggestion, other than that, LGTM :+1:

test/gui/shared/steps/steps.py
```diff
@@ -332,3 +332,18 @@ def step(context, number):
...
+    waitFor(lambda: shareResource(resource), context.userData['clientSyncTimeout'] * 1000)
+    mouseClick(waitForObject(names.qt_tabwidget_tabbar_Public_Links_TabItem), 0, 0, Qt.NoModifier, Qt.LeftButton)
+
+@When("the user toggles the password protection using the client-UI")
+def step(context):
+    clickButton(waitForObject(names.oCC_ShareLinkWidget_checkBox_password_QCheckBox))
+
+@Then('the progress indicator should not be visible in the client-UI')
+def step(context):
+    test.compare(waitForObjectExists(names.oCC_ShareLinkWidget_checkBox_password_QProgressIndicator).visible, False)
```
jasson99, line 349: ?

jasson99:
LGTM

test/gui/shared/steps/steps.py
jasson99, line 349: can we make this line shorter?

SwikritiT: the line with test.compare? I think we can if we change the name of the object.

SwikritiT: But I think it's probably better if we don't, as that name is auto generated and other objects might have some dependency with it.

dpakach (commented):
lgtm :+1:

test/gui/shared/steps/steps.py
jasson99: No, I actually meant if you could split it into two lines. Just to make it easier to read/review. Everything else looks good.

individual-it, line 349: does s.th. like this work in Python?
```suggestion
    test.compare(
         waitForObjectExists(names.oCC_ShareLinkWidget_checkBox_password_QProgressIndicator).visible,
         False
    )
```

individual-it (approved):

test/gui/shared/steps/steps.py
SwikritiT: It's not giving me any error so I think it works.

PR #8500, opened 2021-03-18
Imrpove performance of ExcludedFiles::isExcluded

TheOneRing:


src/csync/csync_exclude.cpp
```diff
@@ -319,29 +319,31 @@ bool ExcludedFiles::isExcluded(
     const QString &basePath,
     bool excludeHidden) const
 {
+    const QFileInfo fileInfo(filePath);
+    if (!fileInfo.exists(filePath)) {
+        return true;
```
TheOneRing, line 324: Here I have the question what the actual correct behavior for non exisitng files is.
With the backup scenario I discovered that we are running expansive actions like isDir on files that don't even exist.

TheOneRing:
Currently the unit test fails as the files tested with don't exist...

dragotin (approved):
LGTM

PR #8502, opened 2021-03-19
Use more stringref

TheOneRing:
Depends on https://github.com/owncloud/client/pull/8500

### DIFF
--- a/src/common/utility.cpp
+++ b/src/common/utility.cpp
@@ -578,9 +578,9 @@ QString Utility::makeConflictFileName(
     return conflictFileName;
 }
 
-bool Utility::isConflictFile(const QString &name)
+bool Utility::isConflictFile(const QStringRef &name)
 {
-    auto bname = name.midRef(name.lastIndexOf(QLatin1Char('/')) + 1);
+    auto bname = name.mid(name.lastIndexOf(QLatin1Char('/')) + 1);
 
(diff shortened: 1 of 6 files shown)


test/testexcludedfiles.cpp
```diff
@@ -42,26 +42,25 @@ class TestExcludedFiles: public QObject
...
+        return excludedFiles->fullPatternMatch({ &path }, ItemTypeDirectory);
+    }
 
-static auto check_dir_traversal(const char *path)
-{
-    return excludedFiles->traversalPatternMatch(path, ItemTypeDirectory);
-}
+    static auto check_file_traversal(const QString &path)
+    {
+        return excludedFiles->traversalPatternMatch({ &path }, ItemTypeFile);
```
fmoc, line 57: I'm not sure you need these braces here. There's no conflicting constructor overloads. It also builds fine on my machine without those.

src/common/utility.cpp
```diff
@@ -636,4 +636,9 @@ QString Utility::sanitizeForFileName(const QString &name)
     return result;
 }
 
+bool Utility::isConflictFile(const QString &name)
```
fmoc, line 639: I'd like to move this wrapper near the wrapped method, so it's easier to find it.

fmoc (approved):
Code and style wise, the PR looks good now.

Issue #8366, opened 2021-01-17
Share window: vertical window resize not possible


mmattel:
When using the context menu to share an item, a window opens to do so. This window is quite large and can´t be vertically resized. Having a standard Full HD resolution 1920x1080, the window is so big, that it is hard to get it fully visible. Trying to resize it is not possible. This is not a good solution. The window should be vertically resizeable (with a minimum size eg, half the size of the vertical resolution) and a scrollbar should be added if the actual size is lower than the needed one to show all elements.

### Expected behaviour
Vertical window resize possible

### Actual behaviour
Vertical window resize not possible

### Steps to reproduce
1. Open share settings via context menu

### Client configuration
Client version: 2.7.4

Operating system: W10x64Pro

OS language: DE

TheOneRing:
That's not how it is supposed to look like, and while the Window can be enlarged it can't be reduced below its initial value.
What you describe sounds like a hdpi scaling issue, are you using a hdpi laptop with a 1080 screen attached?

![image](https://user-images.githubusercontent.com/200626/104894590-2f078600-5975-11eb-93b2-f83798c33794.png)

mmattel:
@TheOneRing
I am using a Fujitsu Lifebook U748, see the specs at: https://www.fujitsu.com/fts/products/computing/pc/notebooks/lifebook-u748/
No external monitor used.
The middle field as shown in your image above is so big, that you run into the issue described.

Q: Why not having the content in a frame and if the frame fits into the physical screen no scrollbar is shown and if it does not fit a scrollbar shows up - and you can resize the window down to a min height which is always half of the physical screen height?

TheOneRing:
In the first place the dialog should not get as big as you describe it :D 
And yes it should allow resizes :D

mmattel:
@TheOneRing any progress on that?

TheOneRing:
Do you have many recipients or is there any reason why that should grow that big?
Could you send a screenshot to apps@owncloud.com :)

mmattel:
It is the initial screen having NO sharees = list is empty...

TheOneRing:
I don't see how the dialog should become that big.
The dialog can be resized, but not below a minimum size.
The minimum size is supposed to be lower than what you describe.

Scroll bars don't solve all problems and cause some https://github.com/owncloud/client/issues/7043 .

mmattel:
The following screenshot is the SMALLEST size I can get on my screen. Intel UHD 1920x1080x60Hz.
[...]

PR #8506, opened 2021-03-22
Fix initial size of share dialoig

TheOneRing:
@mmattel could you give the build artifact a try to see whether it fixes your issue?

### DIFF
--- a/changelog/unreleased/8366
+++ b/changelog/unreleased/8366
@@ -1,0 +1,5 @@
+Bugfix: Fix issue with initial size of share dialogs
+
+We fixed an issue where the initial size of share dialogs was not restored.
+
+https://github.com/owncloud/client/issues/8366
--- a/src/gui/settingsdialog.cpp
+++ b/src/gui/settingsdialog.cpp
@@ -246,10 +246,6 @@ SettingsDialog::SettingsDialog(ownCloudGui *gui, QWidget *parent)
     customizeStyle();
 
     cfg.restoreGeometry(this);
-    if (size().width() < minimumSizeHint().width() || size().height() < minimumSizeHint().height())
-    {
-        resize(minimumSizeHint());
-    }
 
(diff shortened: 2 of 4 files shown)


mmattel:
> give the build artifact a try

can you give me a jump start what this means and how to try it?

mmattel:
If it works, will fix https://github.com/owncloud/client/issues/8366

TheOneRing:
Go to https://github.com/owncloud/client/actions/runs/675876872 download a windows artifact, don't try to run the installer but just extract the contained .7z for example owncloud-client-HEAD-9d921dea8-2.8-windows-msvc2017_64-cl.7z

Make sure owncloud is not running and run the owncloud.exe form that directory, check the about page to ensure your running the right build :)

mmattel:
I tested with the artifact you provided. To get a better impression, I moved the mouse to the taskbar for the screenshot to make the size of the screen more visible. This is how it looks directly after opening the share dialog for the file. I CANNOT reduce the height of the sharing window = smallest size. IMHO there should be a dynamic math behind the screen height with a min value but dependent on the screen height. Just my 2c
image

TheOneRing:
Thx for the screenshot ....
For some reason we set a fixed size for the warning label.
You issue probably also only occurrences when sharing from that share.

TheOneRing:
Hmm no, I can resize it....
image

mmattel:
This is strange. I am using W10x64 latest build an patches but this issue is present for some time now. And all other programs I use do not have this issue.


The interesting thing is, that I can resize the window horizontally, so something must be set to disable it vertically. You see in the image shown the minimum horizontal size I can shrink to:

dschmidt (approved):

fmoc (approved):

TheOneRing:
Fixed in 2.9

PR #8507, opened 2021-03-22
Don't duplicate fsCasePreserving

TheOneRing:
I know you don't like local statics, but I guess this is a compromise and an improvement to what we had before.

### DIFF
--- a/src/libsync/syncfilestatustracker.cpp
+++ b/src/libsync/syncfilestatustracker.cpp
@@ -27,33 +27,12 @@ namespace OCC {
 
 Q_LOGGING_CATEGORY(lcStatusTracker, "sync.statustracker", QtInfoMsg)
 
-static int pathCompare( const QString& lhs, const QString& rhs )
-{
-    // Should match Utility::fsCasePreserving, we want don't want to pay for the runtime check on every comparison.
-    return lhs.compare(rhs,
-#if defined(Q_OS_WIN) || defined(Q_OS_MAC)
-        Qt::CaseInsensitive
-#else
-        Qt::CaseSensitive
-#endif
-        );
-}
-
-static bool pathStartsWith( const QString& lhs, const QString& rhs )
-{
-    return lhs.startsWith(rhs,
-#if defined(Q_OS_WIN) || defined(Q_OS_MAC)
-        Qt::CaseInsensitive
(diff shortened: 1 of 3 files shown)


src/libsync/syncfilestatustracker.cpp
```diff
@@ -27,33 +27,12 @@ namespace OCC {
...
-#endif
-        );
-}
-
 bool SyncFileStatusTracker::PathComparator::operator()( const QString& lhs, const QString& rhs ) const
 {
     // This will make sure that the std::map is ordered and queried case-insensitively on macOS and Windows.
-    return pathCompare(lhs, rhs) < 0;
+    // we want don't want to pay for the runtime check on every comparison.
+    static const Qt::CaseSensitivity sensitivity = Utility::fsCasePreserving() ? Qt::CaseInsensitive : Qt::CaseSensitive;
```
fmoc, line 34: You can use `auto` here.

src/libsync/syncfilestatustracker.h
```diff
@@ -79,6 +79,8 @@ private slots:
     // We'll show a file/directory as SYNC as long as its sync count is > 0.
     // A directory that starts/ends propagation will in turn increase/decrease its own parent by 1.
     QHash<QString, int> _syncCount;
+
+    Qt::CaseSensitivity _caseSensitivity;
```
fmoc, line 83: You might want to add a little documentary comment.

src/libsync/syncfilestatustracker.cpp
```diff
@@ -112,6 +91,7 @@ static inline bool hasExcludedStatus(const SyncFileItem &item)
 
 SyncFileStatusTracker::SyncFileStatusTracker(SyncEngine *syncEngine)
     : _syncEngine(syncEngine)
+    , _caseSensitivity(Utility::fsCasePreserving() ? Qt::CaseInsensitive : Qt::CaseSensitive)
```
fmoc, line 94: This snippet is duplicated above. You might want to implement a little utility function instead that wraps `fsCasePreserving` and returns the appropriate enum entry.

fmoc (changes requested):

fmoc (approved):

Issue #7423, opened 2019-08-22
Cannot unshare without permissions to share


HanaGemela:
Windows 10
Client: 2.5.4 (build 11415)
Server: 10.3.0 alpha

Steps to recreate:
1) User1 shares a resource with User2 with full permissions
2) User2 shares the resource with User3
3) User1 removes Share permissions from User2
4) User2 removes the share from User3

Actual result: User2 cannot unshare the resource with User3
Expected result: User2 can unshare the resource

michaelstingl:
Behaviour should be same as server. Can we compare with server tests?

dpakach:
Test added in https://github.com/owncloud/client/pull/8508

individual-it:
on API level its possible for User2 to remove its own share that goes to User3, but its not possible in the old webUI

individual-it:
API tests added in https://github.com/owncloud/core/pull/38652

TheOneRing:
The sharing dialog is considered deprecated

PR #8508, opened 2021-03-23
[tests-only]unshare without permissions to share

swoichha:
### Description
This PR adds tests for deleting a share after changing permission of the shared file.

### Related Issue
https://github.com/owncloud/client/issues/7423


test/gui/tst_sharing/test.feature
```diff
@@ -53,3 +53,36 @@ Feature: Sharing
         And the user toggles the password protection using the client-UI
         And the user toggles the password protection using the client-UI
         Then the progress indicator should not be visible in the client-UI
+
+    Scenario: unshare a reshared file
+        Given the setting "shareapi_auto_accept_share" on the server of app "core" has been set to "no"
+    	And the administrator on the server has set the default folder for received shares to "Shares"
```
dpakach, line 59: fix indentation here

test/gui/shared/steps/steps.py
```diff
@@ -350,3 +350,16 @@ def step(context):
...
     )
+  
+@When('user "|any|" opens the sharing dialog of "|any|" to delete the folder shared using the client-UI')
+def step(context, receiver, resource):
+    resource = substituteInLineCodes(context, resource).replace('//','/')
+    waitFor(lambda: isFolderSynced(resource), context.userData['clientSyncTimeout'] * 1000)
+    waitFor(lambda: shareResource(resource), context.userData['clientSyncTimeout'] * 1000)   
+    
+
+@Then('"|any|" should be displayed')
```
dpakach, line 361: this step is very generic and I suspect cannot be used everywhere. Can you make it a bit more descriptive?

test/gui/tst_sharing/test.feature
swoichha: This is weird.  Locally the indention is fine but git is showing the wrong indentation.

SwikritiT:
LGTM :+1:

dpakach (commented):
lgtm :+1:

test/gui/tst_sharing/test.feature
individual-it: check tabs vs spaces

test/gui/shared/steps/steps.py
```diff
@@ -350,3 +350,16 @@ def step(context):
         waitForObjectExists(names.oCC_ShareLinkWidget_checkBox_password_QProgressIndicator).visible,
         False
     )
+  
+@When('user "|any|" opens the sharing dialog of "|any|" to delete the folder shared using the client-UI')
```
individual-it, line 354: it really does not matter in this step for what reason the user opens the sharing dialog. the reason and what will happen in that dialog will be determent in the next steps

```suggestion
@When('user "|any|" opens the sharing dialog of "|any|" using the client-UI')
```

individual-it (changes requested):

individual-it (approved):

PR #8517, opened 2021-03-24
Remove unsupported build instructions

TheOneRing:
The instructions are outdated and likely won't work.
We should not encourage people to waste their time.


TheOneRing:
And we should add https://github.com/owncloud/ownbuild to the doc

michaelstingl:
> And we should add https://github.com/owncloud/ownbuild to the doc

include in your PR?

TheOneRing:
> 
> 
> > And we should add https://github.com/owncloud/ownbuild to the doc
> 
> include in your PR?

I can do many things... but explaining stuff is not one of them...

mmattel:
> And we should add https://github.com/owncloud/ownbuild to the doc

In case, this should be made in another PR.
You can file a issue (pls mark it as docs) so it will not be forgotten

mmattel (approved):

PR #8518, opened 2021-03-24
Fix reset of sorting of tables

TheOneRing:


TheOneRing:
Will be handle in an upcoming refactoring

Issue #8501, opened 2021-03-19
Add tests to cover public link functionality


HariBhandari07:
### Description
Add tests to cover all the public link functionality

#### Files
- [x] permissions #8519 
- [x] expiration date  #8527
- [x] password protection #8519 

#### Folders
- [x] permissions (https://github.com/owncloud/client/pull/8561 )
- [x] expiration date https://github.com/owncloud/client/pull/8560 
- [x] password protection https://github.com/owncloud/client/pull/8545 
For reference check the public link tests of web

swoichha:
For now tests has been added for public link functionality. So, closing this issue for now.

PR #8519, opened 2021-03-24
[tests-only]added tests for public link

swoichha:
### Description
This PR adds tests for sharing file using public link.

### Related Issue

### DIFF
--- a/test/gui/shared/steps/steps.py
+++ b/test/gui/shared/steps/steps.py
@@ -233,6 +233,13 @@ def step(context, stepPart1, stepPart2):
     )
 
 
+@Then(r"^(.*) on the server", regexp=True)
+def step(context, stepPart1):
+    executeStepThroughMiddleware(
+        context,
+        "Then " + stepPart1
+    )
+
 @Then('the file "|any|" should exist on the file system with the following content')
 def step(context, filePath):
     expected = "\n".join(context.multiLineText)
@@ -333,16 +340,23 @@ def step(context, number):
     snooze(10)
     clickTab(waitForObject(names.stack_QTabWidget), "Not Synced ({})".format(number))
 
-@When('the user opens the public links dialog of "|any|" using the client-UI')
-def step(context, resource):
+
+def openPublicLinkDialog(context, resource):
     resource = substituteInLineCodes(context, resource).replace('//','/')
     waitFor(lambda: isFileSynced(resource), context.userData['clientSyncTimeout'] * 1000)
     waitFor(lambda: shareResource(resource), context.userData['clientSyncTimeout'] * 1000)
     mouseClick(waitForObject(names.qt_tabwidget_tabbar_Public_Links_TabItem), 0, 0, Qt.NoModifier, Qt.LeftButton)
 
+
+@When('the user opens the public links dialog of "|any|" using the client-UI')
+def step(context, resource):
+    openPublicLinkDialog(context,resource)
+
+
 @When("the user toggles the password protection using the client-UI")
 def step(context):
     clickButton(waitForObject(names.oCC_ShareLinkWidget_checkBox_password_QCheckBox))
+
 
 @Then('the progress indicator should not be visible in the client-UI')
 def step(context):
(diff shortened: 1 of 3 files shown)


test/gui/shared/steps/steps.py
```diff
@@ -233,6 +233,13 @@ def step(context, stepPart1, stepPart2):
     )
 
 
+@Then(r"^(.*) on the server", regexp=True)
```
dpakach, line 236: I'm not fully sure, but you may make this work with any step with `on the server` in it
```suggestion
@Then(r"^(.*)on the server(.*)", regexp=True)
```

test/gui/shared/steps/steps.py
```diff
@@ -362,4 +369,16 @@ def step(context, receiver, resource):
...
+ 
+@When('the user creates a new public link for file "|any|" without password using the client-UI')
+def step(context, resource):
+    test.compare(str(waitForObjectExists(names.sharingDialog_label_name_QLabel).text), resource)
+    clickButton(waitForObject(names.oCC_ShareLinkWidget_createShareButton_QPushButton))     
+    
+@When('the user creates a new public link for file "|any|" with password "|any|" using the client-UI')
+def step(context, resource, password):
+    test.compare(str(waitForObjectExists(names.sharingDialog_label_name_QLabel).text), resource)
+    snooze(10)
```
dpakach, line 381: why is the sleep here?

swoichha: That was added to get object for password field but we won't be needing it now. I will remove it in my next commit.

test/gui/shared/steps/steps.py
swoichha: This way doesn't seem to work. So, I think we should keep this as it is.

dpakach (commented):
lgtm :+1:

test/gui/tst_sharing/test.feature
```diff
@@ -87,3 +87,56 @@ Feature: Sharing
...
+            0\display-name=Alice
+            0\http_oauth=false
+            0\http_user=alice
+            0\url=%local_server%
+            0\user=Alice
+            0\version=1
+            version=2
+            """
+        When the user opens the public links dialog of "%client_sync_path%/textfile0.txt" using the client-UI
+        And the user creates a new public link for file "textfile0.txt" without password using the client-UI
```
individual-it, line 113: I think this step should do the complete action. Opening the dialog + sharing

test/gui/shared/steps/steps.py
```diff
@@ -350,16 +357,27 @@ def step(context):
...
 @Then('the error text "|any|" should be displayed in the sharing dialog')
 def step(context, fileShareContext):
     test.compare(str(waitForObjectExists(names.sharingDialog_The_file_can_not_be_shared_because_it_was_shared_without_sharing_permission_QLabel).text), fileShareContext)
 
-                 
\ No newline at end of file
+
+@When('the user creates a new public link for file "|any|" without password using the client-UI')
+def step(context, resource):
+    test.compare(str(waitForObjectExists(names.sharingDialog_label_name_QLabel).text), resource)
```
individual-it, line 375: we should first wait till the file is synced, in case its not yet

individual-it (changes requested):

test/gui/tst_sharing/test.feature
```diff
@@ -87,3 +87,55 @@ Feature: Sharing
...
+            0\http_oauth=false
+            0\http_user=alice
+            0\url=%local_server%
+            0\user=Alice
+            0\version=1
+            version=2
+            """
+        When the user opens the public links dialog of "%client_sync_path%/textfile0.txt" using the client-UI
+        And the user toggles the password protection using the client-UI
+        And the user creates a new public link for file "textfile0.txt" with password "pass123" using the client-UI
```
individual-it, line 139: same here, lets have a combined high-level step that does it all

individual-it (changes requested):

test/gui/shared/steps/steps.py
```diff
@@ -350,16 +357,38 @@ def step(context):
...
\ No newline at end of file
+
+@When('the user creates a new public link for file "|any|" without password using the client-UI')
+def step(context, resource):
+    resource = substituteInLineCodes(context, resource).replace('//','/')
+    waitFor(lambda: isFileSynced(resource), context.userData['clientSyncTimeout'] * 1000)
+    waitFor(lambda: shareResource(resource), context.userData['clientSyncTimeout'] * 1000)
+    mouseClick(waitForObject(names.qt_tabwidget_tabbar_Public_Links_TabItem), 0, 0, Qt.NoModifier, Qt.LeftButton)
+    
+    test.compare(str(waitForObjectExists(names.sharingDialog_label_name_QLabel).text), resource.replace(context.userData['clientSyncPath'], ''))
```
individual-it, line 380: duplicated code

individual-it (changes requested):

individual-it (approved):

PR #8521, opened 2021-03-24
Cleanup the activityx item and use uuid

TheOneRing:


src/gui/servernotificationhandler.cpp
```diff
@@ -70,42 +70,41 @@ void ServerNotificationHandler::slotNotificationsReceived(const QJsonDocument &j
...
         ActivityLink al;
         al._label = tr("Dismiss");
-        al._link  = Utility::concatUrlPath(ai->account()->url(), notificationsPath + "/" + QString::number(a._id)).toString();
+        al._link = Utility::concatUrlPath(ai->account()->url(), notificationsPath + "/" + QString::number(a.id())).toString();
         al._verb  = "DELETE";
         al._isPrimary = false;
-        a._links.append(al);
-
+        linkList.append(al);
+        a.setLinks(linkList);
```
fmoc, line 107: How about adding a method that returns a mutable reference on the list?

TheOneRing: hmm well I added the functions to make it more clear who modifies what.
A non const ref would break that, I don't like to have a setter here but the alternative would be a separate constructor...

TheAssassin: Makes sense. I'd go for the second constructor then, I guess. It's more laborious but also cleaner in the end, as your object then implements the value object pattern.

fmoc: Perhaps the other constructor is the way to go then? The current constructor offers the option to pass a single link, but I think it's perfectly viable to just accept a list, then change all existing code to pass a single-element list instead.

This way, you can implement the value object pattern, and make the objects basically immutable, which might come in handy in the future.

fmoc: If you want to stick with a setter, you should probably accept r-value references, and use `std::move` to pass the list. This avoids unnecessary copying.

dschmidt (approved):

PR #8524, opened 2021-03-25
Port activities to treeview

TheOneRing:
Depends on https://github.com/owncloud/client/pull/8521
image


michaelstingl (commented):
Very nice 👍 

Use the account filter from the "Not Synced" tab instead of the account column? Or have both? (Sort-order bay account doesn't make that much sense with many activity entries?)

TheOneRing:
Oh and I guess I fixed a whole lot of bugs...

TheOneRing:
> 
> 
> Very nice 👍
> 
> Use the "Account" filter from the "Not Synced" tab instead of the "Account" column? Or have both? (Sort-order by account doesn't make that much sense with plenty of activity entries?)

The account column indicates to which account the activity belongs. Filtering is possible we just need to add the ui for that, probably in a more exel kind of way than we have in the other tab.

Right now I replicated the old view, also the account icon close to the event doesn't make that much sense...

We should discuss this with @tbsbdr in a later call?

michaelstingl (approved):

TheOneRing:
I used a treview while a table view actually makes more sense I guess, looks a bit different
image

TheOneRing:
Switched to tableview

changelog/unreleased/8158
```diff
@@ -0,0 +1,7 @@
+Enhancement: We reworked the server activity table
+
+We redone the server activity table, its now behaves as a proper table,
```
dschmidt, line 3: `redid` or `have redone`

dschmidt: maybe `(have) revised`

dschmidt (approved):

PR #8525, opened 2021-03-25
Include function signature in assert

TheOneRing:
03-25 14:20:01:023 [ fatal default ]:ENFORCE: "1 == 0" in file C:\CraftRoot\build\_\30b36579\owncloud-client\src\gui\application.cpp:void __cdecl OCC::Application::slotCrashEnforce(void), line 473

### DIFF
--- a/src/common/asserts.h
+++ b/src/common/asserts.h
@@ -19,29 +19,29 @@
 // Default assert: If the condition is false in debug builds, terminate.
 //
 // Prints a message on failure, even in release builds.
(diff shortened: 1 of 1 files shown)


fmoc (approved):
LGTM, although the [Qt docs](https://doc.qt.io/qt-5/qtglobal.html#Q_FUNC_INFO) note that the macro might expand not necessarily to the function signature but maybe also file/line number, since it's compiler dependent.

Nevertheless it'll be a great help in tracing down issues like #8526.

PR #8526, opened 2021-03-25
Remove unnecessary stripping of leading slashes

fmoc:
Once the slash is gone, an assert is triggered, making the software terminate unexpectedly:

```
[...] [ fatal default ]: ASSERT: "relativePath.startsWith(QLatin1Char('/'))" in file .../client/src/libsync/abstractnetworkjob.cpp, line 196
```

This happens when trying to add a directory synchronization, in the second step where the remote directory is selected, once you click on the next button.

Related to #8459.

### DIFF
--- a/src/gui/folderwizard.cpp
+++ b/src/gui/folderwizard.cpp
@@ -498,9 +498,6 @@ FolderWizardSelectiveSync::~FolderWizardSelectiveSync()
 void FolderWizardSelectiveSync::initializePage()
 {
     QString targetPath = wizard()->property("targetPath").toString();
-    if (targetPath.startsWith('/')) {
-        targetPath = targetPath.mid(1);
-    }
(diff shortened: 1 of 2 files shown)


TheOneRing:
do things break if its not a root path? 
/foo/bar/ for example :)

fmoc:
Yes, I've tried subdirectories as well, e.g., `/Documents`. The leading slash is stripped, and the software crashes.

TheOneRing (approved):

TheOneRing (changes requested):
See https://github.com/owncloud/client/blob/master/src/gui/folder.h#L161 so we need to remove the trailing slash, BUT not if we only have /

TheOneRing (approved):

fmoc:
This is about the unnecessary removal of a leading, not a trailing slash. I don't think there'll be any issues with trailing slashes.

Issue #8367, opened 2021-01-17
[FR] Share window: Expire Date DD.MM.YYY:  Format is hardcoded --> Format OS System Settings


mmattel:
This Feature Request is a matter of the expire date representation when setting up a share via the context menu.
At the moment, the date is shown hardcoded as DD.MM.YYYY
This should be changed according the System Settings of the OS
If the setting can not be aquired, a default of YYYY.MM.DD should be used.

### Expected behaviour
Date representation according the OS System Setting

### Actual behaviour
Hardcoded date representation DD.MM.YYYY

### Steps to reproduce
1. Open share dialogue via context menu
2. Set an exiration date

### Client configuration
Client version: 2.7.4

Operating system: W10x64Pro

OS language: DE

@TheOneRing

fmoc:
@mmattel I am pretty sure I fixed the issue in #8541, but we couldn't really reproduce the issue without the ability to set a custom language in #8493. The only plausible scenario is that the date picker used the system locale whereas the UI used a theme-enforced locale. Please check whether that was the case, so we can estimate whether #8541 really fixes this issue, alternatively we'll have to dig deeper.

jnweiger:
confirmed fixed. These three incovations produce three differnt date formats in the public link dialog
* `LANG=C LANGUAGE=C LC_ALL=C testpilotcloud-2.9.0-beta3.5019.AppImage` -> 26 Aug 2021
* `LANG=de_DE LANGUAGE=de_DE LC_ALL=de_DE testpilotcloud-2.9.0-beta3.5019.AppImage` -> 26.08.21
* `LANG=en_US LANGUAGE=en_US LC_ALL=en_US testpilotcloud-2.9.0-beta3.5019.AppImage` -> 8/26/21

PR #8527, opened 2021-03-26
[tests-only]added tests for public link share with expiration date

swoichha:
### Description
This PR adds tests for sharing file with expiration dateusing public link.

### Related Issue


SwikritiT (commented):
LGTM :+1:

test/gui/shared/steps/steps.py
```diff
@@ -392,3 +395,22 @@ def step(context, resource, password):
     mouseClick(waitForObject(names.oCC_ShareLinkWidget_lineEdit_password_QLineEdit), 0, 0, Qt.NoModifier, Qt.LeftButton)
     type(waitForObject(names.oCC_ShareLinkWidget_lineEdit_password_QLineEdit), password)
     clickButton(waitForObject(names.oCC_ShareLinkWidget_createShareButton_QPushButton))
+
+
+@When('the user edits the public link named "|any|" of file "|any|" changing following')
+def step(context, publicLinkName, resource):
+    snooze(5)
```
individual-it, line 402: can we do without that static wait?

test/gui/tst_sharing/test.feature
```diff
@@ -137,3 +137,34 @@ Feature: Sharing
...
+            """
+        And user "Alice" has created a public link on the server with following settings
+            | path       | textfile0.txt |
+            | name       | Public link   |
+            | expireDate | 2038-10-14    |
+        When the user opens the public links dialog of "%client_sync_path%/textfile0.txt" using the client-UI
+        And the user edits the public link named "Public link" of file "textfile0.txt" changing following
+            | expireDate | 2038-07-21 |
+        Then the fields of the last public link share response of user "Alice" should include on the server
+            | expireDate | 2038-07-21 |
```
individual-it, line 170: that step fails for me with
```
AssertionError [ERR_ASSERTION]: Expiry Date Mismatch: [object Object]
    at Object.assertUserLastPublicShareDetails (/home/artur/www/owncloud-test-middleware/src/helpers/sharingHelper.js:431:16)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async /home/artur/www/owncloud-test-middleware/src/app.js:56:7 {
  generatedMessage: false,
  code: 'ERR_ASSERTION',
  actual: '2038-02-07',
  expected: '2038-07-21',
  operator: 'strictEqual'
}

File: test/gui/shared/steps/steps.py [hunk 3]
```diff
@@ -392,3 +395,22 @@ def step(context, resource, password):
...
+    test.compare(str(waitForObjectExists(names.sharingDialog_label_name_QLabel).text), resource)
+    test.compare(str(waitForObjectExists(names.linkShares_0_0_QModelIndex).text), publicLinkName)
+    expDate = []
+    for row in context.table:
+        if row[0] == 'expireDate':
+            expDate = row[1].split("-")
+
+    doubleClick(waitForObject(names.oCC_ShareLinkWidget_qt_spinbox_lineedit_QLineEdit), 5, 9, Qt.NoModifier, Qt.LeftButton)
```
individual-it, line 410: this doubleclick seems not to be needed

individual-it (changes requested):
the error output in the log is not so nice 
`Detail Exception: Step execution through test middleware failed. Error: AssertionError [ERR_ASSERTION]: Expiry Date Mismatch: [object Object]`

but I guess that needs to be changed in the middleware

test/gui/shared/steps/steps.py
swoichha: I don't think there is another way for it.

test/gui/shared/steps/steps.py
swoichha: The `doubleClick` is needed because we can only update the expiration date value by double clicking and selecting the current date and then typing the new value to replace the old value of dates.

test/gui/tst_sharing/test.feature
swoichha: Did you run the test after removing the `snooze(5)` step?

individual-it: the problem seems to be my location settings. my date format is set to "d/m/yy" yours seems to be "m/d/yy"

test/gui/shared/steps/steps.py
individual-it: hmm when I try it manually I even need to triple click to select the complete date

individual-it: need a way to set the env. variable `LC_TIME=en_US.utf8` before starting the app under test

test/gui/tst_sharing/test.feature
swoichha: Will it be a problem in future if we run tests on CI?

individual-it: we need a way to set the env. variable `LC_TIME=en_US.utf8` before starting the app under test to make sure its always the same in all test runs

individual-it: I've fixed the branch, now the tests work for me. I had to set the env. variable in ` test/gui/envvars ` these are the variables that get passed on to the AUT see https://kb.froglogic.com/squish/howto/specifying-environment-variables-application/

individual-it (approved):
please rebase and squash the commits

test/gui/tst_sharing/test.feature
michaelstingl: > the problem seems to be my location settings. my date format is set to "d/m/yy" yours seems to be "m/d/yy"

related to https://github.com/owncloud/client/issues/8367 ?

individual-it: no, my system is set to Australian date format, and I believe @swoichha is set to US, so the app behaives correctly, I just had to adjust the app under test

PR #8529, opened 2021-03-29
Refactor protocol and issue widget

TheOneRing:
Depends on #8524

Some parts are still missing and will be handled in a follow up

### DIFF
--- a/src/common/fixedsizeringbuffer.h
+++ b/src/common/fixedsizeringbuffer.h
@@ -1,0 +1,147 @@
+/*
+ * Copyright (C) by Hannah von Reth <hannah.vonreth@owncloud.com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+ * for more details.
+ */
+#pragma once
+
+#include <QtGlobal>
+
+#include <array>
+#include <functional>
+#include <vector>
+
+namespace OCC {
+
+/**
+ * A Fixed sized ring buffer optimized on continouous insertion
+ */
+template <typename TYPE, size_t _Size>
+class FixedSizeRingBuffer
+{
+public:
+    FixedSizeRingBuffer() = default;
+
+    constexpr size_t capacity() const
+    {
+        return _Size;
+    }
+
+    constexpr bool isFull() const
+    {
+        return size() >= _Size;
+    }
+
+    constexpr size_t size() const
+    {
+        return _end - _start;
+    }
+
+    constexpr bool empty() const
+    {
+        return size() == 0;
+    }
+
+    // iterators pointing to the raw data, the range might be smaller than _Size
+    // if ever needed we could implement a iterator for the data window
+    auto begin()
+    {
+        return _data.begin();
+    }
+
+    auto end()
+    {
+        return isFull() ? _data.end() : _data.begin() + size();
+    }
+
+    auto cbegin() const
+    {
+        return _data.cbegin();
+    }
+
+    auto cend() const
+    {
+        return isFull() ? _data.cend() : _data.cbegin() + size();
+    }
+
+    void pop_front()
+    {
+        // move the windows without changing the underlying data
+        _start++;
+        Q_ASSERT(_start < _end);
+        // adjust offset to prevent overflow
+        const auto s = size();
+        _start %= _Size;
+        _end = _start + s;
+    }
+
+    void push_back(TYPE &&data)
+    {
+        Q_ASSERT(!isFull());
+        _data[convertToIndex(size())] = data;
+        _end++;
+    }
+
+    const TYPE &at(size_t index) const
+    {
+        return _data.at(convertToIndex(index));
+    }
+
+    /*
+     * Remove items if f returns true
+     * The filtered result is unordered
+     */
+    void remove_if(const std::function<bool(const TYPE &)> &f)
+    {
+        size_t pos = 0;
+        const auto start = convertToIndex(0);
+        const auto end = size() - 1;
+        // filter and move the elements from to the corners of the buffer
+        for (auto it = _data.begin(); it != _data.begin() + start; ++it) {
+            if (f(*it)) {
+                _end--;
+            } else {
+                std::swap(*it, _data[pos++]);
+            }
+        }
+        pos = end;
+        for (auto it = std::make_reverse_iterator(this->end()); it != _data.rend() - start; ++it) {
+            if (f(*it)) {
+                _start++;
+            } else {
+                std::swap(*it, _data[pos--]);
+            }
+        }
+    }
+
+    void reset(std::vector<TYPE> &&data)
+    {
+        Q_ASSERT(data.size() <= _Size);
+        _start = 0;
+        _end = data.size();
+        std::move(data.begin(), data.end(), _data.begin());
+    }
+
+private:
+    std::array<TYPE, _Size> _data;
+
+    // the sliding window of the ring buffer
+    size_t _start = 0;
+    size_t _end = 0;
+
+    // converts an array index to the underlying array index
+    constexpr size_t convertToIndex(size_t i) const
+    {
+        return (_start + i) % _Size;
+    }
+};
+
+}
--- a/src/gui/models.cpp
+++ b/src/gui/models.cpp
@@ -1,0 +1,49 @@
+/*
+ * Copyright (C) by Hannah von Reth <hannah.vonreth@owncloud.com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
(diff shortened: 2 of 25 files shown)


src/gui/protocolitemmodel.cpp
```diff
@@ -0,0 +1,212 @@
...
+int ProtocolItemModel::columnCount(const QModelIndex &parent) const
+{
+    Q_ASSERT(checkIndex(parent));
+    if (parent.isValid()) {
+        return 0;
+    }
+    return static_cast<int>(ProtocolItemRole::Count);
+}
+
+QVariant ProtocolItemModel::data(const QModelIndex &index, int role) const
```
fmoc, line 62: Wouldn't it make more sense to use `IdemDataRole role` instead? If you really need an `int` overload, you could add one that implements the `static_cast`.

src/gui/protocolitemmodel.cpp
```diff
@@ -0,0 +1,212 @@
...
+                Q_UNREACHABLE();
+            }
+        };
+    }
+    return QAbstractTableModel::headerData(section, orientation, role);
+}
+
+void ProtocolItemModel::addProtocolItem(const ProtocolItem &&item)
+{
+    Q_ASSERT(actualSize() == _data.size());
```
fmoc, line 170: The logic within this function doesn't quite get clear to me. It's very confusing. I understand why you call these `begin`/`end` callbacks. However, it seems a bit inconsistent to me.

I am furthermore pretty sure you should replace `_data` with a better suited structure that encapsulates the logic of keeping track of the current and max sizes. Makes the entire thing easier to use and test. How about a ring buffer class that implements an `append()` operation...? You don't have to care about the underlying data structure then, too.

You could implement the `begin*`/`end*` stuff with signals/slots.

In fact, an `std::deque` would work a lot better, too. It already features an `append`, and `pop`ing from the front gets rid of the first element.

Edit: oh, now I see. You want to keep all the previous entries, right? That's why you use `_start` to keep track of the start.
In fact, this reinforces my opinion that the data structure that stores the logs should be separated. I think you want a class that wraps an `std::deque` (because of its fast appending) and implements the logic of `_start` and `_maxLogSize` (i.e., basically a view on said `deque`), which is used as a data structure in here...

Edit 2: strike that. It's a ring buffer implemented in a vector, `_start` is just used to keep track of the current position for new inserts.

src/gui/protocolwidget.cpp
```diff
@@ -183,142 +47,91 @@ ProtocolWidget::ProtocolWidget(QWidget *parent)
...
+                    if (QFileInfo::exists(localPath)) {
+                        showInFileManager(localPath);
+                    }
+                });
+            }
+        }
+        // "Open in Browser" action
+        {
+            fetchPrivateLinkUrl(folder->accountState()->account(), folder->remotePathTrailingSlash() + data.path(), parent, [parent, menu = QPointer<QMenu>(menu)](const QString &url) {
+                // make sure menu still exists
```
fmoc, line 108: Is it okay if the menu does not exist any more then? I think you should include that in your comment.

test/modeltests/testactivitymodel.cpp
```diff
@@ -0,0 +1,58 @@
...
+    void testInsert()
+    {
+        auto model = new ActivityListModel(this);
+
+        new QAbstractItemModelTester(model, this);
+
+        auto manager = AccountManager::instance();
+
+        auto createAcc = [&] {
+            // don't use the account managher to create the account, it would try to use widgets
```
fmoc, line 31:
```suggestion
            // don't use the account manager to create the account, it would try to use widgets
```

fmoc: Perhaps the account manager perhaps needs to be extended with such functionality?

test/modeltests/testactivitymodel.cpp
```diff
@@ -0,0 +1,58 @@
...
+            auto acc = Account::create();
+            acc->setUrl(QUrl(QStringLiteral("http://admin:admin@localhost/owncloud")));
+            acc->setDavDisplayName(QStringLiteral("fakename") + acc->uuid().toString());
+            acc->setServerVersion(QStringLiteral("10.0.0"));
+            manager->addAccount(acc);
+            return acc;
+        };
+        auto acc1 = createAcc();
+        auto acc2 = createAcc();
```
fmoc, line 40: I would variate the data a bit. Perhaps you can generate some random data in the factory?

test/modeltests/testprotocolmodel.cpp
```diff
@@ -0,0 +1,48 @@
...
+    {
+        // no need to test with 20000 lines
+        const auto TestBacklogSize = 111;
+        auto model = new ProtocolItemModel(this, false, TestBacklogSize);
+
+        new QAbstractItemModelTester(model, this);
+
+        // populate with dummy data
+        auto item = SyncFileItemPtr::create();
+        for (int i = 0; i < TestBacklogSize * 1.1; ++i) {
```
fmoc, line 31: I think you could also just `+2`.

fmoc (dismissed):
Tough one, because of the size. I am not too much into these model views, so I limited my review on general code style.

src/gui/protocolitemmodel.cpp
fmoc: Efficiency wise, an `std::deque` might be sufficient to implement your logic. But a dedicated ring buffer interface is surely nicer to use, and the underlying data structure is hidden behind abstraction. Any kind of performance improvement can be done later on.

fmoc: Here's a basic implementation based on an `std::deque`: https://gist.github.com/fmoc/9b52f831c11835f93276b301211dcd8e

src/gui/protocolwidget.cpp
```diff
@@ -183,142 +47,91 @@ ProtocolWidget::ProtocolWidget(QWidget *parent)
...
+                menu->addAction(tr("Show in file browser"), parent, [localPath] {
+                    if (QFileInfo::exists(localPath)) {
+                        showInFileManager(localPath);
+                    }
+                });
+            }
+        }
+        // "Open in Browser" action
+        {
+            fetchPrivateLinkUrl(folder->accountState()->account(), folder->remotePathTrailingSlash() + data.path(), parent, [parent, menu = QPointer<QMenu>(menu)](const QString &url) {
```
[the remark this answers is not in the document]
TheOneRing: Sometimes clan-format doesn't care for column counts.

src/gui/models.cpp
```diff
@@ -0,0 +1,50 @@
...
+ */
+#include "models.h"
+
+#include <QApplication>
+#include <QItemSelectionRange>
+#include <QTextStream>
+
+QString OCC::Models::formatSelection(const QModelIndexList &items)
+{
+    // TODO: fine tune
```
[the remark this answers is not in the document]
TheOneRing: I actually meant the column width, which isn't that easy.
```
         Action,                                                            File,                        Folder,  Size,             Account,                Time,
The downloaded file does not match the checksum, it will be resumed. 'd4c9b536a6d082dcef9b96a9956986bd3472196f' != 'adc83b19e793491b1c6ea0fd8b46cd9f32e592fc',                                             ownCloud Manual.pdf,                    ownCloud17,6,4 MB,     admin@localhost,2021-04-01T11:08:17.000Z,

File: src/common/fixedsizeringbuffer.h [hunk 9]
```diff
@@ -0,0 +1,156 @@
...
+    FixedSizeRingBuffer() = default;
+
+    constexpr size_t capacity() const
+    {
+        return _Size;
+    }
+
+    constexpr bool isFull() const
+    {
+        return size() >= _Size;
```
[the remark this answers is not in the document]
TheOneRing: the current hold size can be smaller than the maximum size.

dschmidt (changes requested):
please fix ifdef otherwise ok

src/common/fixedsizeringbuffer.h
```diff
@@ -0,0 +1,156 @@
...
+        Q_ASSERT(data.size() <= _Size);
+        _start = 0;
+        _end = data.size();
+        std::move(data.begin(), data.end(), _data.begin());
+    }
+
+private:
+    std::array<TYPE, _Size> _data;
+
+    // the sliding window of the ring buffer
```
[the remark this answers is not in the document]
TheOneRing: I now reset the index based on the max size on a pop_start so it can no longer overflow.

src/common/fixedsizeringbuffer.h
```diff
@@ -0,0 +1,156 @@
...
+{
+public:
+    size_t creationId() const
+    {
+        return __creationId;
+    }
+
+private:
+    size_t __creationId = [] {
+        static size_t _counter = 0;
```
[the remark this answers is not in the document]
TheOneRing: I removed the counter.
It was only needed to find the start point after we removed items.
As we don't need the data to be sorted we have a sort prox model, and the filter function is supposed to be called in a modelReset we don't need to keep track of the index of elements.

src/common/fixedsizeringbuffer.h
```diff
@@ -0,0 +1,156 @@
...
+    void pop_front()
+    {
+        // move the windows without changing the underlying data
+        _start++;
+        Q_ASSERT(_start < _end);
+    }
+
+    void push_back(TYPE &&data)
+    {
+        Q_ASSERT(!isFull());
```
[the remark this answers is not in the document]
fmoc: I see. Well, at least it's checked.

src/common/fixedsizeringbuffer.h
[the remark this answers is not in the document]
fmoc: True. I'm not sure what I was thinking.

src/common/fixedsizeringbuffer.h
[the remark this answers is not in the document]
fmoc: Yup, looks good at a first glance. Thanks!

fmoc (commented):
I had another *very* quick look (well, quick is relative, I mostly just checked the additions). Looks okay to me. Perhaps @dschmidt can have another quick look and approve the PR.

dschmidt (approved):

PR #8537, opened 2021-04-07
[tests-only] multiple small fixes in UI tests

individual-it:
multiple small issues fixed in the GUI tests:
- move path sanitization to own function
- don't rely on `openPublicLinkDialog` to sanitize path - that function should only do its own thing
- wait till the public share is created
- avoid that multiple matching step definitions


swoichha (approved):
LGTM

PR #8545, opened 2021-04-09
[tests-only]password protection public link share for folder

swoichha:
### Description
This PR adds tests for sharing folder using public link:
- with password protection
- without password protection

### Related Issue

### DIFF
--- a/test/gui/tst_sharing/test.feature
+++ b/test/gui/tst_sharing/test.feature
@@ -168,3 +168,53 @@ Feature: Sharing
             | expireDate | 2038-07-21 |
         Then the fields of the last public link share response of user "Alice" should include on the server
             | expireDate | 2038-07-21 |
+
+
+	Scenario: simple sharing of a folder by public link without password
+        Given user "Alice" has been created on the server with default attributes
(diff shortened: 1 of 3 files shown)


test/gui/tst_sharing/test.feature
```diff
@@ -168,3 +168,53 @@ Feature: Sharing
...
+            0\Folders\1\virtualFilesMode=off
+            0\dav_user=alice
+            0\display-name=Alice
+            0\http_oauth=false
+            0\http_user=alice
+            0\url=%local_server%
+            0\user=Alice
+            0\version=1
+            version=2
+            """
```
dpakach, line 192: this is an unrelated issue but can't we hide this? only variable here is `dav_user` and `display_name`

dpakach (commented):
Looks good

test/gui/tst_sharing/test.feature
swoichha: Maybe we can do that in next PR. @individual-it What do you think about this?

individual-it: sounds good, please open an issue and plan it

individual-it (approved):

PR #8547, opened 2021-04-09
Show last sync date in tray menu

### LINKED ISSUE #5644: Tray menu: Clarify "Up to date" item with time of last E-Tag check
When I wake up my laptop while being on flaky WiFi I don't know to what exact time the "Up to date" confirms to.

fmoc:
Fixes #5644.

screenshot_2021-04-09_16-05-50

### DIFF
--- a/src/gui/folderman.cpp
+++ b/src/gui/folderman.cpp
@@ -1230,11 +1230,9 @@ void FolderMan::setDirtyNetworkLimits()
     }
 }
 
-void FolderMan::trayOverallStatus(const QList<Folder *> &folders,
-    SyncResult::Status *status, bool *unresolvedConflicts)
+TrayOverallStatusResult FolderMan::trayOverallStatus(const QList<Folder *> &folders)
 {
-    *status = SyncResult::Undefined;
-    *unresolvedConflicts = false;
+    TrayOverallStatusResult result;
 
     int cnt = folders.count();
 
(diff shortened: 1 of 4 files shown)


src/gui/folderman.cpp
```diff
@@ -1230,11 +1230,12 @@ void FolderMan::setDirtyNetworkLimits()
     }
 }
 
-void FolderMan::trayOverallStatus(const QList<Folder *> &folders,
-    SyncResult::Status *status, bool *unresolvedConflicts)
+void FolderMan::trayOverallStatus(const QList<Folder *> &folders, SyncResult::Status &status, bool &unresolvedConflicts, QDateTime &lastSyncDone)
 {
-    *status = SyncResult::Undefined;
-    *unresolvedConflicts = false;
+    status = SyncResult::Undefined;
```
TheOneRing, line 1235: For reasons I usually also prefer to use pointers for out parameter, it kind of indicates this parameter is special...
I don't like functions that modify my passed reference....

TheOneRing: Maybe just return a tuple or a struct rather than out parameters?

fmoc: References are however safer to use. Passing a mutable reference also suggests it may be modified. I think something like this should be documented, though.

Using a tuple would work in this case, as we don't require the caller to create any of the parameters. Creating a little "return value class" would also do the job, but I'll go for the tuple.

TheOneRing: great, documenting that parameters might get modified is only good in theory.
1) It requires someone to write the doc
2) It requires someone to read the doc....

src/gui/folderman.h
```diff
@@ -134,8 +145,11 @@ class FolderMan : public QObject
     /// Produce text for use in the tray tooltip
     static QString trayTooltipStatusString(SyncResult::Status syncStatus, bool hasUnresolvedConflicts, bool paused);
 
-    /// Compute status summarizing multiple folders
-    static void trayOverallStatus(const QList<Folder *> &folders, SyncResult::Status &status, bool &unresolvedConflicts, QDateTime&lastSyncDone);
+    /**
+     * Compute status summarizing multiple folders
+     * @return tuple containing folders, status, unresolvedConflicts and lastSyncDone
+     */
+    static TrayOverallStatusResult trayOverallStatus(QList<Folder *> folders);
```
TheOneRing, line 152: can we use a const & for the list?

fmoc: Yeah, sure, my bad. Although copying a list of pointers isn't that big of a deal, it's of course better to use references where possible.

TheOneRing (approved):

TheOneRing:
Ship it, mac used to compile. I'll try to fix the cache somehow...

PR #8550, opened 2021-04-12
Set the logfile encoding to utf8

TheOneRing:


src/libsync/logger.cpp
```diff
@@ -148,7 +149,15 @@ void Logger::setLogFile(const QString &name)
...
 
     bool openSucceeded = false;
+    auto codec = "UTF-8";
     if (name == QLatin1String("-")) {
+#ifdef Q_OS_WIN
+        if (AttachConsole(ATTACH_PARENT_PROCESS)) {
+            freopen("CONOUT$", "w", stdout);
+        }
+        _setmode(_fileno(stdout), _O_U8TEXT);
+        codec = "UTF-16";
```
TheOneRing, line 159: Needed to prevent a crash in ucrt `buf.size() % 2 == 0`

src/libsync/logger.cpp
```diff
@@ -148,7 +149,15 @@ void Logger::setLogFile(const QString &name)
     }
 
     bool openSucceeded = false;
+    auto codec = "UTF-8";
     if (name == QLatin1String("-")) {
+#ifdef Q_OS_WIN
+        if (AttachConsole(ATTACH_PARENT_PROCESS)) {
```
TheOneRing, line 155: Only attach in a smart terminal, not if we pipe to a file.

changelog/unreleased/8550
```diff
@@ -0,0 +1,6 @@
+Enhancement: Encode the log file as UTF-8
+
+We fixed an issue where the log file might not have been encode as UTF-8 and thus
```
fmoc, line 3:
```suggestion
We fixed an issue where the log file might not have been encoded as UTF-8 and thus
```

src/libsync/logger.cpp
fmoc: This snippet doesn't make a lot of sense to me. Why first enable UTF-8 (I suppose that's the case), then use UTF-16? If this is some "Windows thing", please put a comment above this to explain what you're doing here. A link to wherever you found this snippet would also be nice to have.

Also, you must check the return value of `_setmode`. It might return a non-zero value.

fmoc (changes requested):

src/libsync/logger.cpp
TheOneRing: Looks like that wasn't needed, only my initial piping to Write-Host broke the encoding.

TheOneRing:
Old:
image

New:
image

dschmidt (approved):

Issue #8552, opened 2021-04-13
Remove unwanted values from py string while setting up client


swoichha:
### Description
Currently while setting up client we send py string in `Given` step that looks something like this 
```
And user "Alice" has set up a client with these settings and password "1234":
            """
            [Accounts]
            0\Folders\1\ignoreHiddenFiles=true
            0\Folders\1\localPath=%client_sync_path%
            0\Folders\1\paused=false
            0\Folders\1\targetPath=/
            0\Folders\1\version=2
            0\Folders\1\virtualFilesMode=off
            0\dav_user=alice
            0\display-name=Alice
            0\http_oauth=false
            0\http_user=alice
            0\url=%local_server%
            0\user=Alice
            0\version=1
            version=2
            """
```
Here we don't have to send all these values in py string. We can just send simply send `user=Alice` and do the rest behind the scene while implementing the `Given` step.

swoichha:
For reference look at https://github.com/owncloud/client/pull/8545#discussion_r611521152

PR #8553, opened 2021-04-13
[Tests-Only]Refactor step for setting up client

swoichha:
### Description
This PR removes unwanted values from py string while setting up client

### Related Issue
- Fixes https://github.com/owncloud/client/issues/8552

### DIFF
--- a/test/gui/shared/steps/steps.py
+++ b/test/gui/shared/steps/steps.py
@@ -48,7 +48,7 @@ def step(context, displayname, host):
             ), displayname + "\n" + host
         )
 
-def setUpClient(context, username, password, configContent):
+def setUpClient(context, username, password, pollingInterval):
     userSetting = '''
     [Accounts]
     0/Folders/1/ignoreHiddenFiles=true
@@ -67,7 +67,7 @@ def setUpClient(context, username, password, configContent):
     version=2
     '''
     userFirstName = username.split()
-    userSetting = userSetting + configContent
+    userSetting = userSetting + pollingInterval
     args = {'displayUserName': username,
         'davUserName': username.lower(),
         'displayUserFirstName': userFirstName[0],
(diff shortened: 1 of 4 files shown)


test/gui/tst_sharing/test.feature
```diff
@@ -65,100 +31,34 @@ Feature: Sharing
...
-            0\user=Brian
-            0\version=1
-            version=2
-            [ownCloud]
-            remotePollInterval=5000
-            """
+        """
+        [ownCloud]
+        remotePollInterval=5000
+        """
```
dpakach, line 37: I think we can make another step  that explicitly states what poll interval to use.

test/gui/shared/steps/steps.py
```diff
@@ -65,6 +88,17 @@ def step(context, username, password):
     except LookupError:
         pass
 
+@Given('user "|any|" has set up a client with poll interval settings and password "|any|"')
```
dpakach, line 91: I think we can use a step where you can provide any poll interval value from here rather than using the default 5000

test/gui/tst_syncing/test.feature
```diff
@@ -64,26 +26,7 @@ Feature: Syncing files
...
-            0\http_oauth=false
-            0\http_user=alice
-            0\url=%local_server%
-            0\user=Alice
-            0\version=1
-            version=2
-            [ownCloud]
-            remotePollInterval=5000
-            """
+		And user "Alice" has set up a client with poll interval settings and password "1234"
```
dpakach, line 29:
```suggestion
		And user "Alice" has set up a client with default settings and poll interval of 5000
```
As for password we can have a global default user values and use one from there.

test/gui/shared/steps/steps.py

test/gui/tst_syncing/test.feature
swoichha: Done

dpakach (commented):
lgtm 👍🏼

test/gui/shared/steps/steps.py
```diff
@@ -20,6 +20,29 @@
...
+        'displayname': 'Brian Murphy',
+        'password': passwords['alt2'],
+        'email': 'brian@example.org'
+    },
+    'Carol': {
+        'displayname': 'Carol King',
+        'password': passwords['alt3'],
+        'email': 'carol@example.org'
+    }
+}
```
individual-it, line 45: note: in future that information should come out of the middleware see https://github.com/owncloud/owncloud-test-middleware/issues/16

individual-it (approved):

PR #8554, opened 2021-04-13
Prevent GitHub actions from aborting remaining jobs on single failures

fmoc:
Sometimes, builds for single OSes might fail because of issues unrelated to a PR. In that case, it makes sense to have the CI finish the remaining jobs in order to be able to see whether these would have failed, too, or finished normally.


TheOneRing:
Changes unrelated to the PR should happen rather seldom.
So I'm not a huge fan.

fmoc:
Closing as unnecessary. For now, at least.

Issue #7595, opened 2019-11-15
Support opentelemetry.io


butonic:
https://opentelemetry.io/ allow tracing requests from client to server.

michaelstingl:
@TheOneRing effort estimation? Should this be part of the client or should this be better a Qt thingy?

michaelstingl:
Ah, they have a C++ library: https://github.com/open-telemetry/opentelemetry-cpp

TheOneRing:
There would also be https://api.kde.org/frameworks/kuserfeedback/html/index.html

TheOneRing:
> 
> 
> Ah, they have a C++ library: https://github.com/open-telemetry/opentelemetry-cpp

Its just a bunch of empty folders at the moment.

butonic:
opencensus has merged with opentracing into opentelemetry. maybe they will continue to use the opencensus stuff in https://github.com/census-instrumentation/opencensus-cpp

butonic:
related: https://github.com/owncloud/client/issues/4606

tboerger:
IMHO this doesn't really make sense for clients, this only makes sense for complex micro service architectures where it's hard to debug the flow of requests.

ogoffart:
What exactly is the client supposed to log?
I believe logging request would be better done on the server side.

michaelstingl:
W3C specification that defines standard HTTP headers and a value format to propagate context information that enables distributed tracing scenarios:
https://www.w3.org/TR/trace-context/#examples-of-http-traceparent-headers

PR #8556, opened 2021-04-15
Introduce X-Original-Request-ID

TheOneRing:
Add an additional header to track reuests that had to be restarted

### DIFF
--- a/src/libsync/abstractnetworkjob.cpp
+++ b/src/libsync/abstractnetworkjob.cpp
@@ -181,6 +181,7 @@ void AbstractNetworkJob::adoptRequest(QNetworkReply *reply)
     setReply(reply);
     setupConnections(reply);
     newReplyHook(reply);
+    _request = reply->request();
 }
 
 QUrl AbstractNetworkJob::makeAccountUrl(const QString &relativePath) const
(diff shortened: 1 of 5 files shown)


fmoc (approved):
Looks good to me.

michaelstingl:
According to [`RFC-6648`](https://tools.ietf.org/html/rfc6648) (from 2012), we should better avoid introducing new custom headers with "`X-`" prefix. Also see https://tonyxu.io/posts/2018/http-deprecate-x-prefix/ for more information.

fmoc:
Does either server support such a parameter at this point? I couldn't find anything in both code bases.

fmoc:
@michaelstingl is right, according to RFC 6648 the parameter should just be called `Original-Request-ID`.

michaelstingl:
Could also be solved with https://github.com/owncloud/client/issues/7595

opentelemetry uses https://w3c.github.io/trace-context/ (`traceparent`, `tracestate` and `traceresponse`)

michaelstingl:
> Does either server support such a parameter at this point? I couldn't find anything in both code bases.

No, only `X-Request-ID` are displayed in the `owncloud.log`.

But any header can be added to the Apache log:
https://doc.owncloud.com/server/admin_manual/configuration/server/logging/request_tracing.html#web-server-configuration-example
(most middleware can do too)

`opentelemetry` (https://github.com/owncloud/client/pull/8556#issuecomment-824634291) is supported in oCIS with _Jaeger_:
- https://owncloud.dev/ocis/deployment/monitoring-tracing/
- https://owncloud.dev/ocis/development/tracing/

TheOneRing:
I'd still like to merge the request as it really helps to debug the code.
Independent of this change we should discuss proper standard tracing and add it to the roadmap in a joined effort with the server team.

michaelstingl:
> I'd still like to merge the request as it really helps to debug the code.

sure 👍

PR #8557, opened 2021-04-15
Clear job queue when server is unreachable

TheOneRing:


src/libsync/jobqueue.cpp
```diff
@@ -36,9 +36,11 @@ void JobQueue::block()
 
 void JobQueue::unblock()
 {
+    if (!_blocked) {
```
fmoc, line 39: As `_blocked` is an unsigned int, not a bool, I'd rather use an explicit comparison against 0.

PR #8561, opened 2021-04-19
[tests-only]public link share for folder with different permissions

swoichha:
### Description
This PR  adds tests for sharing folder using public link with various permissions

### Related Issue
#8501

### DIFF
--- a/test/gui/shared/steps/steps.py
+++ b/test/gui/shared/steps/steps.py
@@ -522,9 +522,11 @@ def step(context, resource, role):
         radioObjectName = names.oCC_ShareLinkWidget_radio_readOnly_QRadioButton
     elif role == 'Editor':
         radioObjectName = names.oCC_ShareLinkWidget_radio_readWrite_QRadioButton
+    elif role == 'Contributor':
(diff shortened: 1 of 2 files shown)


test/gui/shared/steps/steps.py
```diff
@@ -511,3 +511,21 @@ def step(context, permissions, resource, password):
...
+def step(context, resource, role):
+    resource = sanitizePath(substituteInLineCodes(context, resource))
+    radioObjectName = ''
+    
+
+    if role == 'Viewer':
+        radioObjectName = names.oCC_ShareLinkWidget_radio_readOnly_QRadioButton
+    elif role == 'Editor':
+        radioObjectName = names.oCC_ShareLinkWidget_radio_readWrite_QRadioButton
+    else:
```
individual-it, line 525: lets throw an exception is an unexpected role is given

individual-it (changes requested):

test/gui/shared/steps/steps.py
swoichha: Done

individual-it (approved):

PR #8572, opened 2021-04-23
start the client with correct language setting

individual-it:
the current client of the master branch does not get its language settings from env variable but from a commandline switch

### DIFF
--- a/test/gui/shared/steps/steps.py
+++ b/test/gui/shared/steps/steps.py
@@ -71,6 +71,10 @@ def step(context, displayname, host):
             ), displayname + "\n" + host
         )
 
+def startClient():
+    startApplication("owncloud -s --logfile - --language en_US.utf8 --confdir " + confdir)
+    snooze(1)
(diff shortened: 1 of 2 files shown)


swoichha (approved):
lgtm

test/gui/shared/steps/steps.py
```diff
@@ -71,6 +71,10 @@ def step(context, displayname, host):
             ), displayname + "\n" + host
         )
 
+def startClient():
+    startApplication("owncloud -s --logfile - --language en_US.utf8 --confdir " + confdir)
+    snooze(1)
```
TheOneRing, line 76: The client takes language arguments like `en_US` not `en_US.utf8` .
@fmoc please add the format to the command line help text, see https://github.com/owncloud/client/issues/8574 .

individual-it: @TheOneRing fixed in #8580

PR #8575, opened 2021-04-26
Cleanup ui_lang

TheOneRing:


src/gui/openfilemanager.cpp
```diff
@@ -160,7 +162,11 @@ void showInFileManager(const QString &localPath)
 
         static QString name;
         if (name.isEmpty()) {
-            name = desktopFile.value(QString::fromLatin1("Desktop Entry/Name[%1]").arg(qApp->property("ui_lang").toString())).toString();
+            auto displayLanguage = ocApp()->displayLanguage();
+            if (displayLanguage.isEmpty()) {
+                displayLanguage = QLatin1Char('C');
+            }
+            name = desktopFile.value(QString::fromLatin1("Desktop Entry/Name[%1]").arg(displayLanguage)).toString();
```
dschmidt, line 169: Maybe you can check what this fallback actually does and whether it makes sense here ... otherwise fine with me.

dschmidt (approved):

PR #8576, opened 2021-04-26
Finish backup on soft error

TheOneRing:


jnweiger (approved):
Error propagation, yummy!

fmoc (approved):
I'm a little worried about the fact the "error files" are just sent in some failure message, but there won't be any further actions. But anyway, not putting them into the synced files could/should fix the original issue.

PR #8578, opened 2021-04-27
[tests-only]refactor tests to use scenario outline

swoichha:
### Description
This PR refactors test `simple sharing of folder by public link with different roles` to use data table in `When` step.

### DIFF
--- a/test/gui/shared/steps/steps.py
+++ b/test/gui/shared/steps/steps.py
@@ -515,31 +515,10 @@ def step(context, permissions, resource, password):
     waitFor(lambda: (findObject(names.linkShares_0_0_QModelIndex).displayText == "Public link"))
 
 
-@When('the user creates a new public link for folder "|any|" using the client-UI with these details:')
-def step(context, resource):
-    resource = sanitizePath(substituteInLineCodes(context, resource))
-    radioObjectName = ''
-    for row in context.table:
-        if row[0] == 'role':
-            if row[1] == 'Viewer':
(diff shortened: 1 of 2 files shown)


SwikritiT (commented):
LGTM :+1:

Talank (commented):
LGTM

Talank (commented):
looks good

test/gui/shared/steps/steps.py
```diff
@@ -529,8 +527,20 @@ def step(context, resource, role):
...
     openPublicLinkDialog(context, resource)
     clickButton(waitForObject(radioObjectName))
-    clickButton(waitForObject(names.oCC_ShareLinkWidget_createShareButton_QPushButton))       
+    clickButton(waitForObject(names.oCC_ShareLinkWidget_createShareButton_QPushButton))  
+    
+@When('the user creates a new public link for folder "|any|" using the client-UI with these details:')
+def step(context, resource):
+    for row in context.table:
+        if row[0] == 'role':       
+            createPublicShare(context, resource, row[1])
```
individual-it, line 539: that would only create the publiLink if the "role" is given and and if multiple "role" lines are given it would create multiple public links. no other settings are are taken in account

individual-it (changes requested):

test/gui/shared/steps/steps.py
swoichha: This step actual is used just to create a public link based on role. For multiple role line I can add a break so that it works for only one role.

individual-it (approved):

PR #8582, opened 2021-04-27
Fix index

TheOneRing:


src/common/fixedsizeringbuffer.h
```diff
@@ -101,25 +101,23 @@ class FixedSizeRingBuffer
      */
     void remove_if(const std::function<bool(const TYPE &)> &f)
     {
-        size_t pos = 0;
+        // filter and sort the data
+        auto tmp = decltype(_data) {};
```
fmoc, line 105: Out of curiosity, shouldn't it be possible to write `decltype(_data) tmp{};`?

src/common/fixedsizeringbuffer.h
```diff
@@ -101,25 +101,23 @@ class FixedSizeRingBuffer
...
+        auto pos = 0;
```
fmoc, line 106: `auto` will most likely make this an `int`, since the type is inferred from the `0`. It should be explicitly declared as `size_t`, I think. Otherwise, it might overflow with really really large buffers.

fmoc (changes requested):
The algorithm should work now. We give up some efficiency (it now has a memory complexity of O(n)), but at least the implementation is correct now.

I found a potential buffer overflow, please have a look. The rest looks good.

fmoc (approved):
I'm not sure `remove_if` is a very accurate name for the function, but the function looks good now.

PR #8584, opened 2021-04-28
Add Retry action to the protocol items

TheOneRing:


TheOneRing:
image

michaelstingl:
Nice one 👍

Does it work with blacklisted items? Does it maybe reset the blacklist status?

TheOneRing:
That's exactly what it does :)
It removes them from the ignore list and starts a new sync.
For now it only available if a single file is selected, we might allow to select multiple.
The old code added a button on items which where due to quota issues, now it should be available on all errors.

michaelstingl (approved):
LGTM

fmoc (approved):

PR #8585, opened 2021-04-28
Add context menu to activity list

TheOneRing:
image

### DIFF
--- a/src/gui/activitywidget.cpp
+++ b/src/gui/activitywidget.cpp
@@ -57,10 +57,10 @@ ActivityWidget::ActivityWidget(QWidget *parent)
     _ui->setupUi(this);
 
     _model = new ActivityListModel(this);
-    auto sortModel = new QSortFilterProxyModel(this);
-    sortModel->setSourceModel(_model);
-    _ui->_activityList->setModel(sortModel);
-    sortModel->setSortRole(Models::UnderlyingDataRole);
+    _sortModel = new QSortFilterProxyModel(this);
+    _sortModel->setSourceModel(_model);
+    _ui->_activityList->setModel(_sortModel);
+    _sortModel->setSortRole(Models::UnderlyingDataRole);
     _ui->_activityList->hideColumn(static_cast<int>(ActivityListModel::ActivityRole::Path));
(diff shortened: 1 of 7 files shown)


TheOneRing:
Depends on https://github.com/owncloud/client/pull/8584

src/gui/activitywidget.cpp
```diff
@@ -174,55 +173,6 @@ void ActivityWidget::slotAccountActivityStatus(AccountState *ast, int statusCode
     showLabels();
 }
 
-// FIXME: Reused from protocol widget. Move over to utilities.
-QString ActivityWidget::timeString(QDateTime dt, QLocale::FormatType format) const
```
fmoc, line 178: This change appears to be unrelated cleanup and should be in a separate commit. Otherwise, the Git history becomes quite useless.

src/gui/protocolwidget.cpp
```diff
@@ -129,7 +131,6 @@ void ProtocolWidget::showContextMenu(QWidget *parent, ProtocolItemModel *model,
 
 void ProtocolWidget::slotItemContextMenu()
 {
-    QModelIndexList list;
```
fmoc, line 132: This seems like some unrelated cleanup as well.

src/gui/activitywidget.cpp
```diff
@@ -313,24 +253,23 @@ void ActivityWidget::slotBuildNotificationDisplay(const ActivityList &list)
 
     const auto accId = list.first().uuid();
     QList<Activity::Identifier> strayCats;
-    for (const auto &id : _widgetForNotifId.keys()) {
-        NotificationWidget *widget = _widgetForNotifId[id];
+    for (auto it = _widgetForNotifId.cbegin(); it != _widgetForNotifId.cend(); ++it) {
```
fmoc, line 256: Is this change somehow related to the topic of the PR? I don't see the relation between using an iterator here and the new context menu.

fmoc (approved):

src/gui/activitywidget.cpp
TheOneRing: I just removed custom date handling, we just provide a qdate to the view and it takes care of the presentation.
As we no longer sort by the display string we don't need to have seconds in the display string.

PR #8592, opened 2021-05-03
Log if setting up launch on start fails on mac

TheOneRing:


src/common/utility_mac.cpp
```diff
@@ -82,35 +84,40 @@ void setLaunchOnStartup_private(const QString &appName, const QString &guiName,
...
+            } else {
+                qCWarning(lcUtility) << "Failed to insert ourself into launch on startup list";
+            }
+        } else {
+            // We need to iterate over the items and check which one is "ours".
+            UInt32 seedValue;
+            CFArrayRef itemsArray = LSSharedFileListCopySnapshot(loginItems, &seedValue);
+            CFStringRef appUrlRefString = CFURLGetString(urlRef);
+            for (int i = 0; i < CFArrayGetCount(itemsArray); i++) {
+                LSSharedFileListItemRef item = (LSSharedFileListItemRef)CFArrayGetValueAtIndex(itemsArray, i);
```
fmoc, line 106: Should be converted into proper C++-style cast.

fmoc (commented):
Program flow for the new log message looks good. I just found an old C-style cast that should be "upgraded" to C++.

src/common/utility_mac.cpp
TheOneRing: Another day, maybe with clang tidy

PR #8596, opened 2021-05-05
[Tests-Only]Fix user name in dav_user and http_user

Talank:
make `davUserName` to take first name only in lower case.


swoichha (commented):
LGTM :+1:

individual-it (approved):

PR #8598, opened 2021-05-05
Updated CHANGELOG.md

jnweiger:
change unreleased to 2.8 section, 
and added missing 2.7.0 ... 2.7.6 sections (huch?)

@TheOneRing Not sure why the old sections were missing here. Hope it merges back fine :see_no_evil:


TheOneRing:
Will only happen in the tag and on master.
As discussed in every release.

PR #8599, opened 2021-05-05
bump antora to 2.8 here

jnweiger:
@TheOneRing is this correct?

I hope, this cause https://doc.owncloud.com/desktop/2.8 to appear? Or do we need some ansible magic too?

### DIFF
--- a/docs/antora.yml
+++ b/docs/antora.yml
@@ -1,6 +1,6 @@
 name: desktop
 title: Desktop Client
-version: '2.7'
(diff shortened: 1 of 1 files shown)


TheOneRing:
puhhh no idea, it doesn't look wrong though😅 
@dschmidt @mmattel is that the only switch?

mmattel:
This is as far I know the only one 😅

mmattel (approved):

PR #8600, opened 2021-05-05
Windows: Write to debug if a debugger is attached

TheOneRing:
With the recent changes to the logger, Qt creator stopped printing the output if a debugger was attached.

### DIFF
--- a/src/libsync/logger.cpp
+++ b/src/libsync/logger.cpp
@@ -30,10 +30,24 @@
 #ifdef Q_OS_WIN
 #include <io.h> // for stdout
 #include <fcntl.h>
+#include <comdef.h>
 #endif
 
 namespace {
 constexpr int CrashLogSize = 20;
+
+bool isDebuggerPresent()
+{
+#ifdef Q_OS_WIN
+    BOOL debugged;
+    if (!CheckRemoteDebuggerPresent(GetCurrentProcess(), &debugged)) {
+        const auto error = GetLastError();
+        qDebug() << "Failed to detect debugger" << QString::fromWCharArray(_com_error(error).ErrorMessage());
+    }
+    return debugged;
+#endif
+    return false;
+}
 }
 namespace OCC {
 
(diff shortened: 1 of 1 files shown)


src/libsync/logger.cpp
```diff
@@ -80,6 +80,12 @@ void Logger::doLog(QtMsgType type, const QMessageLogContext &ctx, const QString
             if (_doFileFlush)
                 _logstream->flush();
         }
+#if defined(Q_OS_WIN)
+        if (IsDebuggerPresent()) {
```
fmoc, line 84: You may want to extend this check a bit more with other Win32 API debugger checks, e.g., `CheckRemoteDebuggerPresent`.

fmoc (approved):
Perhaps the debugger checks should be moved into a central static method. The concept could also be extended on other platforms.

TheOneRing:
> 
> 
> Perhaps the debugger checks should be moved into a central static method. The concept could also be extended on other platforms.

This is mainly a workaround for the qt debugger.
The client actually should not behave differently when a debugger is present, but I assume the change here is small enough.

fmoc:
Hence I approved the PR. Just wanted to note that `IsDebuggerPresent` is a bit limited, it only returns "true" when the process is launched by a debugger directly. "Attached" debuggers can be recognized with `CheckRemoteDebuggerPresent` only (at least with the Win32 API).

michaelstingl:
Add a note to the troubleshooting docs?

src/libsync/logger.cpp
```diff
@@ -30,10 +30,24 @@
...
 namespace {
 constexpr int CrashLogSize = 20;
+
+bool isDebuggerPresent()
+{
+#ifdef Q_OS_WIN
+    BOOL debugged;
+    if (!CheckRemoteDebuggerPresent(GetCurrentProcess(), &debugged)) {
+        const auto error = GetLastError();
+        qDebug() << "Failed to dtect debugger" << QString::fromWCharArray(_com_error(error).ErrorMessage());
```
fmoc, line 45: `detect`

Issue #8603, opened 2021-05-05
Switch to `QCommandLineParser`


fmoc:
The homebrew argument parser shall be replaced by a new [`QCommandLineParser`](https://doc.qt.io/qt-5/qcommandlineparser.html)-based setup.

~~Blocks #8574.~~

fmoc:
For the record, `showHint` also just calls `std::exit` rather than using one of the variables...

PR #8604, opened 2021-05-05
Implement `--list-languages` CLI parameter

### LINKED ISSUE #8574: Add `--list--languages`, print available languages also in the hint of `--language`
Also document the format of the language argument.

fmoc:
Closes #8574.

### DIFF
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -88,6 +88,7 @@ namespace {
             "  --logflush           : flush the log file after every write.\n"
             "  --logdebug           : also output debug-level messages in the log.\n"
             "  --language <locale>  : override UI language\n"
+            "  --list-languages     : list available translations (for use with --language)\n"
             "  --confdir <dirname>  : Use the given configuration folder.");
     }
 }
@@ -171,6 +172,7 @@ Application::Application(int &argc, char **argv)
     , _helpOnly(false)
     , _versionOnly(false)
     , _showLogWindow(false)
+    , _listAvailableTranslationsOnly(false)
     , _logExpire(0)
     , _logFlush(false)
     , _logDebug(false)
@@ -275,7 +277,7 @@ Application::Application(int &argc, char **argv)
         qCInfo(lcApplication) << "VFS suffix plugin is available";
 
     //no need to waste time;
-    if (_helpOnly || _versionOnly)
+    if (_helpOnly || _versionOnly || _listAvailableTranslationsOnly)
         return;
 
     if (_quitInstance) {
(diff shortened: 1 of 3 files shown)


src/gui/application.cpp
```diff
@@ -621,8 +622,11 @@ void Application::parseOptions(const QStringList &options)
             if (it.hasNext() && !it.peekNext().startsWith(QLatin1String("--"))) {
                 _userEnforcedLanguage = it.next();
```
TheOneRing, line 623: Please check wehter the selected language is in the list of available languages and fail if not.

fmoc: Done. It makes little sense, though. A locale like `en_US` will still lead to an English translation to be loaded even though only an `en` file is bundled. Therefore, I initially didn't want to implement this. Perhaps we can just print a warning instead?

fmoc: Actually, it looks like `en_US` leads to an English translation (that would be correct), but `nl_NL` leads to an English one as well (should be `nl`). Therefore, I think validating the input as you suggested is the right way to go.

src/gui/application.cpp
```diff
@@ -619,10 +621,21 @@ void Application::parseOptions(const QStringList &options)
             _versionOnly = true;
         } else if (option == QLatin1String("--language")) {
             if (it.hasNext() && !it.peekNext().startsWith(QLatin1String("--"))) {
-                _userEnforcedLanguage = it.next();
+                auto languageParam = it.next();
+
+                // fail if the language is unknown
+                if (!Translations::listAvailableTranslations().contains(languageParam)) {
+                    std::cerr << "Error: unknown language " << languageParam.toStdString() << std::endl;
```
TheOneRing, line 628: Hm as the errors here do not got though displayHelpText they might not be discoverable on Windows.
Since we nowadays attach to the console its not that big of an issue but...

src/gui/application.cpp
fmoc: Strike that, I had configured the language to English in the settings. `nl_NL` works fine for `--language`, even though it's not available in the list.

src/gui/application.cpp
fmoc: I can also use `displayHint` as before, which will `std::exit(...)`. However, then, I have to duplicate listing the available languages.

Why not reference `--list-languages` from this message? Then we can use `displayHint` just fine.

TheOneRing: go for it

TheOneRing: or you could define a function/lambda that you call :D

fmoc: Let's conclude that #8603 will make everything a lot better...

TheOneRing (approved):

Issue #8577, opened 2021-04-27
Implement Page Object 


swoichha:
### Description
Currently we are not implementing Page Object Model in our desktop-client tests. All the step definitions are under same file called `steps.py` and all the object names are in file `names.py`. 

So, now as more tests are being added we should implement page object model as we do for Web and Core.

For reference have a look at https://kb.froglogic.com/squish/howto/using-page-object-test-design-pattern/

To-Do:
- [x]  tst_addAccount (#8606)
- [x] tst_sharing (https://github.com/owncloud/client/pull/8634) (@swoichha )
- [x] tst_syncing(@Talank )

PR #8548, opened 2021-04-12
[tests-only] run GUI tests in CI

individual-it:

PR #8606, opened 2021-05-06
[tests-only]implement pageObject model

swoichha:
### Description

This PR is a POC for implementing Page Object Model in client test. Currently the PO is only implemented for tests related to adding accounts which is in `tst_addAccount.py`

For this implementation  the folder structure looks like this:
```
test
|
|------>gui	
	|
	|------>shared
	|	|
	|	|------>scripts
	|	|	|
	|	|	|----->helpers
	|	|	|	|
	|	|	|	|---->SetupClientHelper.py
	|	|	|	
	|	|	|------>pageObjects
	|	|	|	|	
	|	|	|	|----->AccountConnectionWizard.py
	|	|	|	
	|	|	|	
	|	|	|------->bdd_hooks.py
	|	|	|	
	|	|	|------->names.py
	|	|
	|	|
	|	|------>steps
	|		|
	|		|
	|		|----->steps.py
	|			
	|
	|------>tst_addAccount
	|
	|------>tst_sharing
	|
	|------>tst_syncing
```

- The `helper` folder contains `SetupClientHelper.py` which contains general helpers functions.
-  The `pageObjects` contains all the stepObject model. For now there is only `AccountConnectionWizard.py`
### Related Issue
https://github.com/owncloud/client/issues/8577

### DIFF
--- a/test/gui/shared/scripts/helpers/SetupClientHelper.py
+++ b/test/gui/shared/scripts/helpers/SetupClientHelper.py
@@ -1,11 +1,8 @@
-class SetupClientHelper():
-    def __int__(self):
-        pass
-    
-    def substituteInLineCodes(self, context, value):
-        from urllib.parse import urlparse
-        value = value.replace('%local_server%', context.userData['localBackendUrl'])
-        value = value.replace('%client_sync_path%', context.userData['clientSyncPath'])
-        value = value.replace('%local_server_hostname%', urlparse(context.userData['localBackendUrl']).netloc)
+from urllib.parse import urlparse
 
-        return value
+def substituteInLineCodes(context, value):
+    value = value.replace('%local_server%', context.userData['localBackendUrl'])
+    value = value.replace('%client_sync_path%', context.userData['clientSyncPath'])
+    value = value.replace('%local_server_hostname%', urlparse(context.userData['localBackendUrl']).netloc)
+
+    return value
--- a/test/gui/shared/scripts/helpers/SetupClientHelper.py
+++ b/test/gui/shared/scripts/helpers/SetupClientHelper.py
@@ -1,0 +1,11 @@
+class SetupClientHelper():
+    def __int__(self):
+        pass
+    
+    def substituteInLineCodes(self, context, value):
+        from urllib.parse import urlparse
+        value = value.replace('%local_server%', context.userData['localBackendUrl'])
+        value = value.replace('%client_sync_path%', context.userData['clientSyncPath'])
+        value = value.replace('%local_server_hostname%', urlparse(context.userData['localBackendUrl']).netloc)
+
+        return value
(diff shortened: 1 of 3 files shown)


Talank (commented):
LGTM, but we need to verify that all tests are passing.

test/gui/shared/scripts/helpers/SetupClientHelper.py
```diff
@@ -0,0 +1,11 @@
+class SetupClientHelper():
+    def __int__(self):
+        pass
+    
+    def substituteInLineCodes(self, context, value):
```
dpakach, line 5: can we just use this as a function without the class, it would be much easier to use since we don't have to create new object every time.

Talank: Or, maybe we can make that method static using `@staticmethod` decorator

kiranparajuli589: making the fn static method might be a nice solution

test/gui/shared/scripts/helpers/SetupClientHelper.py
```diff
@@ -0,0 +1,11 @@
...
+        from urllib.parse import urlparse
```
kiranparajuli589, line 6: do imports on top

test/gui/shared/scripts/helpers/SetupClientHelper.py
```diff
@@ -0,0 +1,8 @@
+from urllib.parse import urlparse
```
dpakach, line 1: this file name is wrong. please change it to appropriate name.

dpakach (commented):
lgtm 👍🏼

kiranparajuli589 (commented):
rather than exisiting comments..lgtm 👍🏼

test/gui/shared/scripts/pageObjects/AccountConnectionWizard.py
```diff
@@ -0,0 +1,35 @@
...
+
+class AccountConnectionWizard():
+    SERVER_ADDRESS_BOX = names.leUrl_OCC_PostfixLineEdit
+    NEXT_BUTTON = names.owncloudWizard_qt_passive_wizardbutton1_QPushButton
+    USERNAME_BOX = names.leUsername_QLineEdit
+    PASSWORD_BOX = names.lePassword_QLineEdit
+    SELECT_LOCAL_FOLDER = names.pbSelectLocalFolder_QPushButton
+    DIRECTORY_NAME_BOX = names.fileNameEdit_QLineEdit
+    CHOOSE_BUTTON = names.qFileDialog_Choose_QPushButton
+    CONNECT_BUTTON = names.owncloudWizard_qt_passive_wizardbutton1_QPushButton
```
individual-it, line 14: what about having the full names of the selectors just here and not in the `names` class?

individual-it (changes requested):
I suggest to merge this after #8548

individual-it (approved):

PR #8611, opened 2021-05-07
hardening setting date of public link in tests

individual-it:
doing the exact key strokes that are needed to set the date and to navigate through the fields seems to be more stable in the tests

after typing the date wait for progressindicator to disappear


swoichha (approved):
lgtm

PR #8613, opened 2021-05-07
Don't warn if we detect empty legacy credentials

TheOneRing:


dragotin (approved):
Yes, looks good to me.

Issue #8614, opened 2021-05-07
Repair failing test for file sync


swoichha:
### Description
After current redesigned for `Not Sync` tab in client the following test for checking file sync is failing
https://github.com/owncloud/client/blob/master/test/gui/tst_syncing/test.feature#L28-L50

We need to fix these tests according to the new design.

PR #8616, opened 2021-05-07
[tests-only] repair file sync failing test

swoichha:
### Description
This PR fixes failing test for file sync after the recent redesign of desktop-client.

### Related Issue
- fixes https://github.com/owncloud/client/issues/8614

### DIFF
--- a/test/gui/shared/steps/steps.py
+++ b/test/gui/shared/steps/steps.py
@@ -407,7 +407,6 @@ def step(context, filename):
             "type": "QModelIndex"
     })
     
-# {"column": 1, "container": oCC_IssuesWidget_tableView_QTableView, "row": 0, "type": "QModelIndex"}
 
 @When('the user selects the unsynced files tab')
 def step(context):
--- a/test/gui/shared/steps/steps.py
+++ b/test/gui/shared/steps/steps.py
@@ -402,18 +402,19 @@ def step(context, filename):
     waitForObject(names.settings_OCC_SettingsDialog)
     waitForObjectExists({
             "column": 1,
-            "container": names.oCC_IssuesWidget_treeWidget_QTreeWidget,
+            "container": names.oCC_IssuesWidget_tableView_QTableView,
             "text": RegularExpression(buildConflictedRegex(filename)),
             "type": "QModelIndex"
     })
+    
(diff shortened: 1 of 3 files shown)


test/gui/shared/steps/steps.py
```diff
@@ -402,18 +402,19 @@ def step(context, filename):
...
 
-
-@When('the user selects the unsynced files tab with |integer| unsynced files')
-def step(context, number):
+@When('the user selects the unsynced files tab')
+def step(context):
     # TODO: find some way to dynamically select the tab name
     # It might take some time for all files to sync except the expected number of unsynced files
     snooze(10)
-    clickTab(waitForObject(names.stack_QTabWidget), "Not Synced ({})".format(number))
```
TheOneRing, line 416: Maybe I should rather display the number again.
@michaelstingl what do you think?

michaelstingl: > Maybe I should rather display the number again.

What number? The counter in the tab-head? Yeah, good idea.

swoichha: @TheOneRing @michaelstingl For now I think we'll have to merge this PR as it is and when you change the UI again let me. As I have to modify these tests so that they can work with new UI changes.

test/gui/shared/steps/steps.py
```diff
@@ -402,18 +402,19 @@ def step(context, filename):
     waitForObject(names.settings_OCC_SettingsDialog)
     waitForObjectExists({
             "column": 1,
-            "container": names.oCC_IssuesWidget_treeWidget_QTreeWidget,
+            "container": names.oCC_IssuesWidget_tableView_QTableView,
             "text": RegularExpression(buildConflictedRegex(filename)),
             "type": "QModelIndex"
     })
+    
+# {"column": 1, "container": oCC_IssuesWidget_tableView_QTableView, "row": 0, "type": "QModelIndex"}
```
individual-it, line 410: do we need that line?

test/gui/shared/steps/steps.py
TheOneRing: I think one purpose of those tests is to detect regressions.
I'd say the missing counter is a regression.

individual-it (approved):

PR #8619, opened 2021-05-10
Remove long dead and rotten themeing code

TheOneRing:


fmoc (approved):
Looks good code wise, but I can't judge how "dead" this code is. It doesn't seem to be used anywhere.

PR #8620, opened 2021-05-10
open sharing dialog in step

individual-it:
that should fix the sharing test, otherwise it would not open the sharing dialog


swoichha (approved):
lgtm

PR #8621, opened 2021-05-10
use expected to fail system

individual-it:
for now we are expecting this step to fail because of #7459
squish has a way of defining tests(assertions) that are expected to fail, so using that and do not implement a expected to fail parser


phil-davis (commented):
This is a reasonable way to do it. The step should fail when the issue is fixed, and the developer will have to adjust the test.

Talank (commented):
looks good.

Talank (approved):

PR #8622, opened 2021-05-10
Doccument server bug

TheOneRing:


JanAckermann (changes requested):
why not set it to true?

TheOneRing:
@JanAckermann I'm now using the defaults

JanAckermann (approved):
Perfect :+1:

PR #8623, opened 2021-05-10
improve public link tests

individual-it:
mostly this hardens the exiting public link tests but also makes sure that not only day and month are changed but also year when testing the edit a public link

- check content of date input field
- send 'Return' after changing expiry date
- check first if the indicator is visible and then that it disappears
- check that all parts of the date are changed
- wait for the public link to be displayed


individual-it:
note: creating public links is done in multiple places, need to be DRYer see  #8624

test/gui/shared/steps/steps.py
```diff
@@ -484,16 +485,23 @@ def step(context, publicLinkName, resource):
     expDate = []
     for row in context.table:
         if row[0] == 'expireDate':
-            expDate = row[1].split("-")
-    expYear = expDate[0][2:]
+            expDate = datetime.datetime.strptime(row[1],'%Y-%m-%d')
+    expYear = expDate.year-2000
```
phil-davis, line 489: That's a bit sad that the app only uses 2-digit years. What happens if I want the public link to expire in 2101?

This code will have to be adjusted when we get near to 2100! :)

phil-davis (approved):

PR #8628, opened 2021-05-11
wait for all files to be synced before pause the file sync

individual-it:
pausing without waiting for all files to be synced results in a pop-up that asks if the current sync should be canceled, that breaks the next steps
as this is a GIVEN step we could add some check in an other PR to see if the sync was really paused


swoichha (approved):
LGTM

PR #8630, opened 2021-05-11
Display local file name in case clash warning

TheOneRing:
image
image


TheOneRing:
Hm wait that didn't actually achieve what I intended...

fmoc (approved):
Looks good.

PR #8632, opened 2021-05-12
Replace custom argument parsing with QCommandLineParser

### LINKED ISSUE #8603: Switch to `QCommandLineParser`

fmoc:
Closes #8603.

The big chain of `if`s could be reduced by using a map and a loop, but it works well.

I could continue to use `showHint`/`displayHelpText` for user feedback, Qt is relatively flexible there. 

Note that #8604 should be rebased once this PR has been merged.

### DIFF
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -70,31 +70,6 @@ namespace OCC {
 
 Q_LOGGING_CATEGORY(lcApplication, "gui.application", QtInfoMsg)
 
-namespace {
-
-    const QString optionsC()
-    {
-        return QStringLiteral(
-            "Options:\n"
-            "  -h --help            : show this help screen.\n"
-            "  -s --showsettings    : show the settings dialog while starting.\n"
-            "  -q --quit            : quit the running instance\n"
-            "  --logfile <filename> : write log output to file <filename>.\n"
-            "  --logfile -          : write log output to stdout.\n"
-            "  --logdir <name>      : write each sync log output in a new file\n"
-            "                         in folder <name>.\n"
-            "  --logexpire <hours>  : removes logs older than <hours> hours.\n"
-            "                         (to be used with --logdir)\n"
-            "  --logflush           : flush the log file after every write.\n"
-            "  --logdebug           : also output debug-level messages in the log.\n"
-            "  --language <locale>  : override UI language\n"
-            "  --list-languages     : list available translations (for use with --language)\n"
-            "  --confdir <dirname>  : Use the given configuration folder.");
-    }
-}
-
-// ----------------------------------------------------------------------------------
-
 bool Application::configVersionMigration()
 {
     QStringList deleteKeys, ignoreKeys;
@@ -169,10 +144,7 @@ Application::Application(int &argc, char **argv)
     : SharedTools::QtSingleApplication(Theme::instance()->appName(), argc, argv)
     , _gui(nullptr)
     , _theme(Theme::instance())
-    , _helpOnly(false)
-    , _versionOnly(false)
     , _showLogWindow(false)
-    , _listAvailableTranslationsOnly(false)
     , _logExpire(0)
     , _logFlush(false)
     , _logDebug(false)
@@ -275,10 +247,6 @@ Application::Application(int &argc, char **argv)
         qCInfo(lcApplication) << "VFS windows plugin is available";
     if (isVfsPluginAvailable(Vfs::WithSuffix))
         qCInfo(lcApplication) << "VFS suffix plugin is available";
-
-    //no need to waste time;
-    if (_helpOnly || _versionOnly || _listAvailableTranslationsOnly)
-        return;
 
     if (_quitInstance) {
         QTimer::singleShot(0, qApp, &QApplication::quit);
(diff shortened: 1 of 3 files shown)


src/gui/application.cpp
```diff
@@ -646,31 +549,106 @@ static void displayHelpText(const QString &t, std::ostream &stream = std::cout)
...
+    auto helpOption = parser.addHelpOption();
+    auto versionOption = parser.addVersionOption();
+
+    // this little snippet saves a few lines below
+    auto addOption = [&parser](const QCommandLineOption &option) {
+        parser.addOption(option);
+        return option;
+    };
+
+    auto showSettingsOption = addOption({ { "s", "showsettings" }, "Show the settings dialog while starting." });
```
TheOneRing, line 579: We should translate the description.
Not sure weather we should translate the language options though :)

src/gui/application.cpp
```diff
@@ -646,31 +549,106 @@ static void displayHelpText(const QString &t, std::ostream &stream = std::cout)
...
+    auto quitInstanceOption = addOption({ { "q", "quit" }, "Quit the running instance." });
+    auto logFileOption = addOption({ "logfile", "Write log to file (use - to write to stdout).", "filename" });
+    auto logDirOption = addOption({ "logdir", "Write each sync log output in a new file in folder.", "name" });
+    auto logExpireOption = addOption({ "logexpire", "Remove logs older than <hours> hours (to be used with --logdir).", "hours" });
+    auto logFlushOption = addOption({ "logflush", "Flush the log file after every write." });
+    auto logDebugOption = addOption({ "logdebug", "Output debug-level messages in the log." });
+    auto logWindowOption = addOption({ "logwindow", "//TODO//" });
```
TheOneRing, line 586: I think we can remove the option?
We need to update the doc of course.

src/gui/application.cpp
```diff
@@ -646,31 +549,106 @@ static void displayHelpText(const QString &t, std::ostream &stream = std::cout)
...
+    auto languageOption = addOption({ "language", "Override UI language.", "language" });
+    auto confDirOption = addOption({ "confdir", "Use the given configuration folder.", "dirname" });
+    auto debugOption = addOption({ "debug", "Enable debug mode." });
+
+    // virtual file system parameters (optional)
+    parser.addPositionalArgument("vfs file", "Virtual file system file to be opened (optional).", { "[<vfs file>]" });
```
TheOneRing, line 592: Hmm sadly I don't think the commandlineparser allows to hide options, as this is never meant to be used in a direct invocation but only by the os.

src/gui/application.cpp
```diff
@@ -646,31 +549,106 @@ static void displayHelpText(const QString &t, std::ostream &stream = std::cout)
...
+
+    auto positionalArguments = parser.positionalArguments();
+
+    // ignore any positional arguments beyond the first one
+    if (!positionalArguments.empty()) {
+        QTimer::singleShot(0, this, [this, positionalArguments] { openVirtualFile(positionalArguments.front()); });
+    }
 }
 
 void Application::showHint(const QString &errorHint)
```
TheOneRing, line 654: Hmm maybe print the hint to the console and call showHelp()

src/gui/application.cpp
```diff
@@ -562,75 +534,6 @@ void Application::slotParseMessage(const QString &msg, QObject *)
...
-            // virtual file, open it after the Folder were created (if the app is not terminated)
-            QTimer::singleShot(0, this, [this, option] { openVirtualFile(option); });
-        } else {
-            showHint(QStringLiteral("Unrecognized option '%1'").arg(option));
-        }
-    }
-}
 
 // Helpers for displaying messages. Note that there is probably no console on Windows.
 static void displayHelpText(const QString &t, std::ostream &stream = std::cout)
```
TheOneRing, line 539: Directly call parser.showHelp()

TheOneRing (changes requested):
Looks good so far :)

src/gui/application.cpp
fmoc: The translations are loaded _after_ the options are parsed. There is no point in translating any of these. To make translations any useful, we'd have to implement some mechanism to reload the translations in case `--language` is set.

Alternatively, one could eliminate that parameter and replace it with an environment variable.

src/gui/application.cpp
fmoc: I don't think there's a point in hiding it, really. How many people will ever look at this help anyway? It's for advanced users. We could extend the description if you want to.

src/gui/application.cpp
fmoc: Well, yes or no? I don't mind having this method, but I don't mind removing it either.

src/gui/application.cpp
```diff
@@ -646,31 +549,106 @@ static void displayHelpText(const QString &t, std::ostream &stream = std::cout)
...
-           << _theme->version() << endl;
+    QCommandLineParser parser;
+
+    QString descriptionText;
+    QTextStream descriptionTextStream(&descriptionText);
 
-    stream << QLatin1String("File synchronisation desktop utility.") << endl
-           << endl
-           << optionsC();
+    descriptionTextStream << _theme->appName() << QStringLiteral(" version ") << _theme->version() << endl;
```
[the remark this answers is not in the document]
fmoc: See below.

src/gui/application.cpp
fmoc: I intentionally didn't use this because of that comment. Do you want to change the UX on Windows as well?

src/gui/application.cpp
TheOneRing: ouch your right. 
Lets keep it untranslated for now.

src/gui/application.cpp
TheOneRing: Recent changes allow us to print to the console on Windows.
Also QCommandLineParser shows the popup itself afaik.

fmoc:
The argument parser has not been translated before. We decided to fix this in a future PR by calling `setupTranslations` more than once. I'll update all new strings so they are included in the translations.

src/gui/application.cpp
fmoc: Removed the option.

src/gui/application.cpp
fmoc: Hiding also doesn't seem to be possible with the standard Qt parser.

src/gui/application.cpp
```diff
@@ -661,46 +538,122 @@ static void displayHelpText(const QString &t, std::ostream &stream = std::cout)
...
+    }
 }
 
 void Application::showHint(const QString &errorHint)
 {
     QString out;
     QTextStream hint(&out);
     hint << errorHint << endl
-         << "Try '" << QFileInfo(QCoreApplication::applicationFilePath()).fileName() << " --help' for more information" << endl;
+         << tr("Try \"%1\" --help' for more information").arg(QFileInfo(QCoreApplication::applicationFilePath()).fileName()) << endl;
```
TheOneRing, line 656: no need for that stream

src/gui/application.cpp
fmoc: Removed the function in the latest commit, as discussed recently.

src/gui/application.cpp
fmoc: We're using `QCommandLineParser` directly now where applicable. As we can't use it to show our own hints, I had to keep `displayHelpText`, though. But `showHint` could be removed.

src/gui/application.cpp
fmoc: The latest commit removes the function anyway.

src/gui/application.cpp
```diff
@@ -661,60 +541,113 @@ static void displayHelpText(const QString &t, std::ostream &stream = std::cout)
...
-           << _theme->version() << endl;
+    QCommandLineParser parser;
 
-    stream << QLatin1String("File synchronisation desktop utility.") << endl
-           << endl
-           << optionsC();
+    QString descriptionText;
+    QTextStream descriptionTextStream(&descriptionText);
+
+    descriptionTextStream << tr("%1 version %2\r\nFile synchronization desktop utility.").arg(_theme->appName()).arg(_theme->version()) << endl;
```
TheOneRing, line 551: use .arg(arg1,arg2)

src/gui/application.cpp
```diff
@@ -661,60 +541,113 @@ static void displayHelpText(const QString &t, std::ostream &stream = std::cout)
...
+    auto logDebugOption = addOption({ "logdebug", tr("Output debug-level messages in the log.") });
+    auto languageOption = addOption({ "language", tr("Override UI language."), "language" });
+    auto listLanguagesOption = addOption({ "list-languages", tr("Override UI language.") });
+    auto confDirOption = addOption({ "confdir", tr("Use the given configuration folder."), "dirname" });
+    auto debugOption = addOption({ "debug", tr("Enable debug mode.") });
+
+    // virtual file system parameters (optional)
+    parser.addPositionalArgument("vfs file", tr("Virtual file system file to be opened (optional)."), { tr("[<vfs file>]") });
+
+    parser.process(*this);
```
TheOneRing, line 585: You need to conditionally check the passed arguments parameter.
This function is also invoked by ipc when a second instance is created.

TheOneRing: I'd recommend to not pass arguments in the default case but only when invoked from the ipc message.

TheOneRing (changes requested):

src/gui/application.cpp
```diff
@@ -232,6 +203,9 @@ Application::Application(int &argc, char **argv)
         }
     }
 
+    // needed during commandline options parsing
+    setApplicationVersion(_theme->versionSwitchOutput());
```
TheOneRing, line 207: The version is translatable but I guess we don't care...

TheOneRing (approved):

src/gui/application.cpp
fmoc: I considered data coming from the theme to be translated already...

TheOneRing: it is but if you  call setApplicationVersion before we set up the translator :D

fmoc: Will take this issue into account in my upcoming PR that fixes the translation of the commandline options.

PR #8634, opened 2021-05-13
[tests-only] implement POM for sharing 

swoichha:
### Description
This PR  implementing Page Object Model in client test for sharing and also make public link creation test methods DRY.

### Related Issue
- https://github.com/owncloud/client/issues/8577


Talank (commented):
Looks good, but there seem to be some conflicts. Please rebase.

test/gui/shared/scripts/pageObjects/AccountConnectionWizard.py
```diff
@@ -4,14 +4,55 @@
...
+        "type": "OCC::PostfixLineEdit",
+        "visible": 1,
+    }
+    NEXT_BUTTON = {
+        "name": "__qt__passive_wizardbutton1",
+        "type": "QPushButton",
+        "visible": 1,
+        "window": names.owncloudWizard_OCC_OwncloudWizard,
+    }
+    USERNAME_BOX = {
```
dpakach, line 19: I think you can split these changes to a different PR  since this is not sharing related

test/gui/shared/steps/steps.py
```diff
@@ -285,17 +194,9 @@ def step(context, receiver, resource, permissions):
...
-    test.compare(
-        waitForObjectExists(names.scrollArea_permissionsEdit_QCheckBox).checked,
-        ('edit' in permissionsList),
-    )
-    test.compare(
-        waitForObjectExists(names.scrollArea_permissionShare_QCheckBox).checked,
-        ('share' in permissionsList),
-    )
+    shareItem = SharingDialog()
+    snooze(10)
```
dpakach, line 198: why is there sleep here?

test/gui/shared/steps/steps.py
```diff
@@ -542,122 +436,62 @@ def step(context, receiver, resource):
...
-    test.compare(
-        str(
-            waitForObjectExists(
-                names.sharingDialog_The_file_can_not_be_shared_because_it_was_shared_without_sharing_permission_QLabel
-            ).text
-        ),
-        fileShareContext,
-    )
+    shareItem = SharingDialog()
+    shareItem.checkErrorDisplayed(fileShareContext)
```
dpakach, line 440: I think we can use the same basic rule we use on other UI tests (eg. on web). 
- do not use assertions in page objects
- page objects only return state and assertion is done in the step definitions.

test/gui/shared/steps/steps.py
```diff
@@ -542,122 +436,62 @@ def step(context, receiver, resource):
...
-    resource = sanitizePath(substituteInLineCodes(context, resource))
-    openPublicLinkDialog(context, resource)
+def getRadioObjectForPermssion(permissions):
     radioObjectName = ''
     if permissions == 'Download / View':
         radioObjectName = names.oCC_ShareLinkWidget_radio_readOnly_QRadioButton
     elif permissions == 'Download / View / Edit':
         radioObjectName = names.oCC_ShareLinkWidget_radio_readWrite_QRadioButton
     elif permissions == 'Upload only (File Drop)':
         radioObjectName = names.oCC_ShareLinkWidget_radio_uploadOnly_QRadioButton
```
dpakach, line 478: Maybe this can be done in page objects. since it's related to the selectors it makes more sense to do it there.

test/gui/shared/steps/steps.py
```diff
@@ -666,21 +500,13 @@ def step(context, permissions, resource):
...
-            findObject(names.linkShares_0_0_QModelIndex).displayText == "Public link"
-        )
+    openSharingDialog(context, resource)
+    publicLinkDialog = PublicLinkDialog()
+    publicLinkDialog.openPublicLinkDialog()
+    radioObjectName = getRadioObjectForPermssion(permissions)
+
+    publicLinkDialog.createPublicLink(
+        context, resource, password, radioObjectName, permissions
     )
```
dpakach, line 510: instead of radio object name pass permissions to the functions and get the name inside the page objects.

dpakach (changes requested):

test/gui/shared/scripts/pageObjects/AccountConnectionWizard.py
swoichha: Fixed

test/gui/shared/steps/steps.py
swoichha: That was added during checking for objects in AUI but now I have removed it.

test/gui/shared/steps/steps.py

test/gui/shared/steps/steps.py

test/gui/shared/steps/steps.py

test/gui/shared/steps/steps.py
```diff
@@ -448,7 +450,7 @@ def step(context, resource):
     openSharingDialog(context, resource)
     publicLinkDialog = PublicLinkDialog()
     publicLinkDialog.openPublicLinkDialog()
-    publicLinkDialog.createPublicLink(context, resource, '', '', '')
+    publicLinkDialog.createPublicLink(context, resource, '', '')
```
dpakach, line 453: there must be a way to pass default args in python so you don't have to give empty strings here.

test/gui/shared/steps/steps.py
```diff
@@ -542,12 +435,9 @@ def step(context, receiver, resource):
...
 @Then('the error text "|any|" should be displayed in the sharing dialog')
 def step(context, fileShareContext):
+    shareItem = SharingDialog()
     test.compare(
-        str(
-            waitForObjectExists(
-                names.sharingDialog_The_file_can_not_be_shared_because_it_was_shared_without_sharing_permission_QLabel
-            ).text
-        ),
+        str(waitForObjectExists(shareItem.ERROR_SHOWN_ON_SHARING_DIALOG).text),
```
dpakach, line 440: is there a way to get if the object is visible or not on the page in squish. If that is possible you can make a method that returns true/false whether an element is visible or not and check the assertion based on that.
since we are using page objects now I think we should avoid calling the squish API on the page (such as `waitForObjectExists` from the step definitions)

test/gui/shared/steps/steps.py
```diff
@@ -697,7 +505,9 @@ def createPublicShare(context, resource, role):
     else:
         raise Exception("No such role found for resource")
 
-    openPublicLinkDialog(context, resource)
+    openSharingDialog(context, resource)
+    publicLinkDialog = PublicLinkDialog()
+    publicLinkDialog.openPublicLinkDialog()
     clickButton(waitForObject(radioObjectName))
     clickButton(waitForObject(names.oCC_ShareLinkWidget_createShareButton_QPushButton))
     waitFor(
```
dpakach, line 513: I think you should move these to a function inside the page object.

test/gui/shared/scripts/pageObjects/SharingDialog.py
```diff
@@ -0,0 +1,95 @@
...
+        ):
+            squish.clickButton(squish.waitForObject(self.EDIT_PERMISSIONS_CHECKBOX))
+        if ('share' in permissionsList and shareChecked == False) or (
+            'share' not in permissionsList and shareChecked == True
+        ):
+            squish.clickButton(squish.waitForObject(self.SHARE_PERMISSIONS_CHECKBOX))
+
+        squish.clickButton(squish.waitForObject(self.SHARING_DIALOG_CLOSE_BUTTON))
+
+    def isCollaboratorsListed(self, receiver, permissionsList):
```
dpakach, line 83: IMO function isCollaboratorsListed should just return True/False based on weather they are listed or not. It should not do assertions itself and throw errors.

test/gui/shared/steps/steps.py
dpakach: Reason https://github.com/owncloud/client/pull/8634#discussion_r637715523

test/gui/shared/steps/steps.py
swoichha: I think `waitForObjectExists` is the only way to do this.

test/gui/shared/steps/steps.py
```diff
@@ -557,154 +456,68 @@ def step(context, fileShareContext):
...
-        radioObjectName = names.oCC_ShareLinkWidget_radio_readOnly_QRadioButton
+        radioObjectName = publicLinkDialog.READ_ONLY_RADIO_BUTTON
     elif role == 'Editor':
-        radioObjectName = names.oCC_ShareLinkWidget_radio_readWrite_QRadioButton
+        radioObjectName = publicLinkDialog.READ_WRITE_RADIO_BUTTON
     elif role == 'Contributor':
-        radioObjectName = names.oCC_ShareLinkWidget_radio_uploadOnly_QRadioButton
+        radioObjectName = publicLinkDialog.UPLOAD_ONLY_RADIO_BUTTON
     else:
         raise Exception("No such role found for resource")
```
dpakach, line 518: this part can go in the page object

test/gui/shared/steps/steps.py
dpakach: there is a function `findObject` which throws LookupError if object is not found.  you can use try, except in the function,

test/gui/shared/steps/steps.py
```diff
@@ -542,169 +444,62 @@ def step(context, receiver, resource):
 
 @Then('the error text "|any|" should be displayed in the sharing dialog')
 def step(context, fileShareContext):
+    shareItem = SharingDialog()
+    errorText = shareItem.getErrorText(shareItem.ERROR_SHOWN_ON_SHARING_DIALOG)
```
dpakach, line 448: you don't have to pass the arg here since this variable is available in the class itself

swoichha: Yes you are right. I did this so that I could you it for some other steps as well that didn't work. Will change it now.

dpakach (commented):
looks good 👍🏼

dpakach (approved):

individual-it (approved):

PR #8635, opened 2021-05-14
[tests-only]Add init in bdd hooks

swoichha:
### Description
This PR adds init and sends a http `POST` request in `OnScenarioStart`. This change in the hooks is done as https://github.com/owncloud/owncloud-test-middleware/pull/34 has been merged now.

### DIFF
--- a/test/gui/shared/scripts/bdd_hooks.py
+++ b/test/gui/shared/scripts/bdd_hooks.py
@@ -17,6 +17,7 @@
 # manual for a complete reference of the available API.
 import shutil
 import urllib.request
+import os
 
 @OnScenarioStart
 def hook(context):
@@ -52,7 +53,7 @@ def hook(context):
         context.userData['clientConfigFile']='-'
         
     req = urllib.request.Request(
-        context.userData['middlewareUrl'] + 'init',
+        os.path.join(context.userData['middlewareUrl'], 'init'),
         headers={"Content-Type": "application/json"}, method='POST'
     )
     try:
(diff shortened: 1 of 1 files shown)


test/gui/shared/scripts/bdd_hooks.py
```diff
@@ -49,6 +50,18 @@ def hook(context):
 
     if context.userData['clientConfigFile'] == '':
         context.userData['clientConfigFile']='-'
+        
+    req = urllib.request.Request(
+        context.userData['middlewareUrl'] + 'init',
```
dpakach, line 55: can you use function to join paths such as `path.join()` here. it will throw error if middlewareUrl doesn't end with `/`

test/gui/shared/scripts/bdd_hooks.py
```diff
@@ -49,6 +51,18 @@ def hook(context):
 
     if context.userData['clientConfigFile'] == '':
         context.userData['clientConfigFile']='-'
+        
+    req = urllib.request.Request(
+        os.path.join(context.userData['middlewareUrl'], 'init',''),
```
dpakach, line 56: why is there empty string in the end?

swoichha: So that it would end with `/`

dpakach: why do we need to have the URL end with `/`
also you can just replace `init` with `init/` if you want that

dpakach (commented):
LGTM 👍🏼

individual-it (approved):

Issue #8631, opened 2021-05-12
Setup linting mechanism for test


Talank:
When we create PRs in the client, there are some formatting errors around the code, e.g. Tabs vs. Spaces or extra spaces at the end. So we need to set up some linting mechanism. For example by using `pylint` or some other tool like that.

PR #8636, opened 2021-05-17
[test-only] Implement `black` to achieve uniform code formatting 

SwikritiT:
## Black
When we create PRs in the client, there are some formatting errors around the code, e.g. Tabs vs. Spaces or extra spaces at the end. This PR addresses this issue by implementing a python code formatter called `Black` , thereby hoping to achieve uniformity across the coding style. `Black` is a automatic code formatting tool for python. `Black` reformats entire files in place. The coding style used by `Black` can be viewed as a strict subset of `PEP 8` . 

Here's the documentation of black where you can find further information regarding this tool https://black.readthedocs.io/en/stable/getting_started.html

## Set-up Locally
Before creating PR we can run `black` command locally which will format the code according to its standard. 
1 . Install black: `pip3 install black`
2. To see which files will be formatted we can run `black --check [path of the file/directory you want to format]` (we can use `.` as path if we want to format all the files)
3. To see what changes will be made `black --diff [path]`
4. This step will automatically re-format the file `black [path]` 

One downside of this tool is that the style configuration options are limited. Style configuration can be done in `pyproject.toml` file under `[tools.black]` section. 

## CI Implementation 
A pipline `guitestformat` has been created. It pull's the docker image of `black` and the command ` black --check --diff . ` is run inside path `/drone/src/test/gui` to check if any file needs re-formatting. If any file needs re-formatting the CI will fail and show us which file needs reformatting and what reformatting can be done, otherwise it'll be green. 

A failing CI and its output looks like this https://drone.owncloud.com/owncloud/client/6906/2/2

Fixes: https://github.com/owncloud/client/issues/8631


kiranparajuli589 (commented):
lgtm 👍🏼

swoichha (commented):
LGTM :+1:

.idea/workspace.xml
```diff
@@ -0,0 +1,99 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Talank: I don't think we need this file `.idea/workspace.xml` to be merged into master. 🤔

SwikritiT: Yes, that wasn't in my .gitignore so got commited accidentally :sweat_smile: . Removed it on my latest commit .

.drone.star
```diff
@@ -341,6 +343,25 @@ def build_client_docs(ctx):
...
+        "type": "docker",
+        "name": "guitestformat",
+        "steps": [
+            {
+                "name": "black",
+                "image": "cytopia/black",
+                "pull": "always",
+                "commands": [
+                    'cd /drone/src/test/gui',
+                    'black --check --diff .',
```
Talank, line 358: It is better to include the steps to check the lining locally in some documentation as well.
Everything else looks good to me 👍🏽

pyproject.toml
```diff
@@ -0,0 +1,17 @@
+# Example configuration for Black.
```
individual-it, line 1: this is not an example anymore

pyproject.toml
```diff
@@ -0,0 +1,17 @@
...
+
+[tool.black]
+line-length = 88
+target-version = ['py36', 'py37', 'py38']
+include = '\.pyi?$'
+skip-string-normalization = true
+extend-exclude = '''
+# A regex preceded with ^/ will apply only to files and directories
+# in the root of the project.
+^/foo.py  # exclude a file named foo.py in the root of the project (in addition to the defaults)
```
individual-it, line 16: we don't have files called `foo.py` so no need to exclude them, but maybe we want to exclude `names.py` here

test/gui/shared/scripts/names.py
```diff
@@ -1,3 +1,4 @@
+# fmt: off
```
individual-it, line 1: could add the file alternatively to the exclude list in the config

individual-it (changes requested):
good job! I just have small comments

I would love to have `make` commands here as we have in other repos, but that can be done in a separate PR

individual-it:
issue opened to create a make command for linting: #8642 (no need to solve it in this PR)

individual-it (approved):

Issue #8624, opened 2021-05-10
DRY public link creation test methods 


individual-it:
There are multiple steps and functions that create public links, they should be consolidated into one, maybe that can be part of #8577

PR #8639, opened 2021-05-18
[tests-only] DRY public link creation

swoichha:
## Description
This PR adds new method `openShareDialog` for opening sharing dialog box and make creation of public link creation DRY.

## Related Issue
- fixes https://github.com/owncloud/client/issues/8624

### DIFF
--- a/test/gui/shared/steps/steps.py
+++ b/test/gui/shared/steps/steps.py
@@ -219,9 +219,7 @@ def executeStepThroughMiddleware(context, step):
 
 @When('the user adds "|any|" as collaborator of resource "|any|" with permissions "|any|" using the client-UI')
 def step(context, receiver, resource, permissions):
-    resource = sanitizePath(substituteInLineCodes(context, resource))
-    waitFor(lambda: isFileSynced(resource), context.userData['clientSyncTimeout'] * 1000)
-    waitFor(lambda: shareResource(resource), context.userData['clientSyncTimeout'] * 1000)
+    openSharingDialog(context, resource, '')
 
     mouseClick(waitForObject(names.sharingDialogUG_shareeLineEdit_QLineEdit), 0, 0, Qt.NoModifier, Qt.LeftButton)
     type(waitForObject(names.sharingDialogUG_shareeLineEdit_QLineEdit), receiver)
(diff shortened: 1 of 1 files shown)


test/gui/shared/steps/steps.py
```diff
@@ -391,15 +389,23 @@ def step(context):
...
 
 
-def openPublicLinkDialog(context, resource):
+def openSharingDialog(context, resource, fileType='file'):
     resource = sanitizePath(substituteInLineCodes(context, resource))
-    waitFor(lambda: isFileSynced(resource), context.userData['clientSyncTimeout'] * 1000)
+    
+    if fileType == 'folder':
+        waitFor(lambda: isFolderSynced(resource), context.userData['clientSyncTimeout'] * 1000)
+    else:
```
individual-it, line 397:
```suggestion
    else if fileType == 'file':
```

`else` should through an error, because of invalid data. Otherwise the programmer might do a mistake and set it to `Folder` and it would check for a file

test/gui/shared/steps/steps.py
```diff
@@ -391,15 +389,23 @@ def step(context):
     clickTab(waitForObject(names.stack_QTabWidget), "Not Synced")
 
 
-def openPublicLinkDialog(context, resource):
+def openSharingDialog(context, resource, fileType='file'):
```
individual-it, line 392:
```suggestion
def openSharingDialog(context, resource, itemType='file'):
```

fileType = folder sounds strange even strictly speaking a folder is a type of a file on the filesystem, maybe call it `itemType`?

test/gui/shared/steps/steps.py
```diff
@@ -391,15 +389,23 @@ def step(context):
...
     waitFor(lambda: shareResource(resource), context.userData['clientSyncTimeout'] * 1000)
+    
+def openPublicLinkDialog(context, resource, fileType):
+    openSharingDialog(context, resource, fileType)
     mouseClick(waitForObject(names.qt_tabwidget_tabbar_Public_Links_TabItem), 0, 0, Qt.NoModifier, Qt.LeftButton)
 
 @When('the user opens the public links dialog of "|any|" using the client-UI')
 def step(context, resource):
-    openPublicLinkDialog(context,resource)
+    openPublicLinkDialog(context,resource,'')
```
individual-it, line 408:
```suggestion
    openPublicLinkDialog(context,resource)
```
or 
`openPublicLinkDialog(context,resource, 'file')`

same in other places

individual-it (changes requested):

test/gui/shared/steps/steps.py
swoichha: fixed

test/gui/shared/steps/steps.py
swoichha: fixed

test/gui/shared/steps/steps.py
swoichha: fixed

individual-it (approved):

Issue #8642, opened 2021-05-19
create a `make` command for UI test linting


individual-it:
after #8636 is merged it would be nice to have some `make` command for the linting as there are in other repos, that would make it easier to remember what to type to do linting

PR #8643, opened 2021-05-20
[tests-only] Add makefile for UI test formatting

SwikritiT:
This PR adds Makefile for UI test formatting 
fixes https://github.com/owncloud/client/issues/8642

### DIFF
--- a/test/gui/Makefile
+++ b/test/gui/Makefile
@@ -1,0 +1,27 @@
+# Main Makefile for formatting python code for gui test
+#
+# Requirements to run make here:
+# - black
+#
+# Installing black can be done via https://black.readthedocs.io/en/stable/getting_started.html
+# configuration for black can be found inside pyproject.toml file under [tools.black] section.
+#
+# Usage:
+# make
+#	-auto format all the python files inside `test/gui`
+# make check
+#	-check which files will be reformatted
+# make diff
+#	-see what changes will be done
+#
+
+black:
+	black .
+
+.PHONY: diffFormat
+diffFormat:
+	black --diff .
+
+.PHONY: checkFormat
+checkFormat:
+	black --check .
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -0,0 +1,26 @@
...
+#    - black
+#
+# Installing black can be done via https://black.readthedocs.io/en/stable/getting_started.html
+# configuration for black can be found inside pyproject.toml file under [tools.black] section.
+#
+# Usage:
+#
+#   make - auto format all the python files inside test/gui
+#   make check - check which files will be reformatted
+#	make diff - see what changes will be done
```
Talank, line 13: Why is this line of comment has more space than other before the text? 🤔

Talank (commented):
Since this is a Makefile for formatting python code for gui test, is it necessary to make this makefile in the root of the client? Can we not shift it in test/gui? Or maybe rename the commands such as  `make checkFormat`, `make diffFormat` 🤔

Makefile
SwikritiT: probably due to ide , fixed it

swoichha (commented):
LGTM

individual-it (commented):
LGTM, but I would love to know if @TheOneRing is happy with that extra Makefile

TheOneRing:
I'm relative ok with it as its not in any src dir but it does not do what I would expect from a make file 😅 
As it is possible to pass files to make (-f) maybe use a more fitting name.
In the end, is there anything gained in introducing another scripting language to  a mostly python code base?
Why not just a format-code.py

individual-it:
the idea was just to make it easier to remember what to type in that specific repo, in other repos all the test related commands, including source quality checks are in some `make` command. So I thought it would be easier to type `make` and see what it can than to go and search the docs for what the command is for that particular repo

TheOneRing:
Then go for make

individual-it (approved):

Issue #8641, opened 2021-05-19
get rid of steps that sets up a client with given polling interval


individual-it:
There is a step that sets up the client with a given polling interval: `Given user "Brian" has set up a client with default settings and polling interval "5000"`

I think we should have the shortest possible polling interval by default because we want that the tests run as quick as possible. In future is there are cases where we explicitly want to have a longer interval we can add such a step again, but default should be 5000ms (shortest possible)

PR #8646, opened 2021-05-20
[tests-only] Set polling interval by default

SwikritiT:
This PR removes polling interval from test step and sets it to 5000ms by default 

fixes https://github.com/owncloud/client/issues/8641


individual-it (changes requested):
I think that should move into `setUpClient`
because now `@Given('user "|any|" has set up a client with default settings and password "|any|"')` will still setup the client with the long interval

individual-it (approved):

Issue #8272, opened 2020-12-02
Virtual Files on a Windows-PC that has Folder-Synchronization disabled


BurninLeo:
The "virtual files" feature of OwnCloud works great, especially if there are a lot of files on the server, and you only need a few of them. First of all, thanks for that!

I recently tried the experimental "virtual files" feature on a PC that has the Windows feature "Offline File Synchronization" disabled. Company stuff, because that has caused trouble in the past when working with network drives....

This apparently affects how the "virtual files" of OwnCloud as displayed. As the client explicitly asks for feedback on the experimental feature, here we go :)

### Expected behaviour

The files shall appear in the Windows Explorer as if they where there, with the original filename (`filename.extension`).

### Actual behaviour

The replacement-files are displayed as `filename.extension.owncloud`, e.g., `Sample.pdf.owncloud`.

When right-clicking the file, and choosing ownCloud -> Download File from the contect menu, the file is retrieved correctly, and can be used. But, it is not auto-retrieved upon the first use.

### Steps to reproduce
1. Disable the Window's Offline File Synchronization
2. Connect to a folder, using the experimental virtual files feature
3. Open the folder (or a subfolder of it) in the Windows Explorer

### Server configuration

Operating system: Ubuntu 20.04 LTS

Web server: Nginx

Database: MySQL

PHP version: 7.x

ownCloud version: not relevant for this issue

Storage backend (external storage): Filesystem

### Client configuration

Client version: 2.5.4

Operating system: Windows 10.0.18362

OS language: German

Installation path of client: default

michaelstingl:
Please test with the latest 2.7.2 version:
https://owncloud.com/desktop-app/

Maybe you need to setup a new account.

BurninLeo:
After updating to the latest version, owncloud crashed - and kept crashing after reboot. Of course, not owncloud per se - after deleting the owncloud settings folder, and reinstalling owncloud it started. But as soon as I tried to setup the sync folder with virtual files, it crashed, again.

After deleting the settings folder, and creating the "traditional" folder sync, everything worked fine. Therefore, I think we can assume that the "virtual files" feature does not yet work well with disabled "Offline File Synchronization".

michaelstingl:
Did you submit the crash report? Please post the crash report ID. Thanks!

BurninLeo:
[...]

PR #8649, opened 2021-05-20
Fix network drive detection

TheOneRing:
We can't rely on the file system detection because Windows is reporting the underlying file system.

Fixes: #8272

### DIFF
--- a/src/common/vfs.cpp
+++ b/src/common/vfs.cpp
@@ -68,12 +68,17 @@ Result<bool, QString> Vfs::checkAvailability(const QString &path)
     const auto mode = bestAvailableVfsMode();
 #ifdef Q_OS_WIN
     if (mode == Mode::WindowsCfApi) {
-        if (QDir(QDir(path).canonicalPath()).isRoot()) {
+        const auto info = QFileInfo(path);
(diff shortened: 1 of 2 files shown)


src/common/vfs.cpp
```diff
@@ -68,13 +68,18 @@ Result<bool, QString> Vfs::checkAvailability(const QString &path)
...
+        const auto info = QFileInfo(path);
+        if (QDir(info.canonicalPath()).isRoot()) {
             return tr("The Virtual filesystem feature does not support a drive as sync root");
         }
-        const auto fs = FileSystem::fileSystemForPath(path);
+        const auto fs = FileSystem::fileSystemForPath(info.absoluteFilePath());
         if (fs != QLatin1String("NTFS")) {
             return tr("The Virtual filesystem feature requires a NTFS file system, %1 is using %2").arg(path, fs);
         }
+        const auto type = GetDriveTypeW(reinterpret_cast<const wchar_t *>(QDir::toNativeSeparators(info.absoluteFilePath().mid(0, 3)).utf16()));
```
fmoc, line 79: Is this cast really safe? https://stackoverflow.com/a/23124860 suggests it might be...

fmoc (approved):

PR #8655, opened 2021-05-25
[Tests-Only]POM implementation in the syncing tests

Talank:
### Description
This PR  implementing the Page Object Model in the client test for syncing.

### Related Issue
- https://github.com/owncloud/client/issues/8577


test/gui/shared/scripts/pageObjects/Activity.py
```diff
@@ -0,0 +1,30 @@
+import names
+import squish
+from objectmaphelper import RegularExpression
+from helpers.FilesHelper import buildConflictedRegex
+
+
+class Activity:
+    SUBTAB = {
```
swoichha, line 8: Can you rename it to something else. It is a bit hard to understand which tab/button is this.

Talank: The name of the tab depends upon the tab name passed here `squish.clickTab(squish.waitForObject(self.SUBTAB), tabName)` So, it is like a generic selector, not a specific one.

Talank: The name of the tab depends upon the tab name passed here squish.clickTab(squish.waitForObject(self.SUBTAB), tabName) So, it is like a generic selector, not a specific one.

swoichha (commented):
Looks good :+1:

individual-it (approved):

Issue #8652, opened 2021-05-21
Refactor steps with password in them


swoichha:
## Description

There are some tests that passes password value from gherkin steps, e.g. https://github.com/owncloud/client/blob/a64b638468455a6e2edf185578d7c3be3a888d9b/test/gui/tst_sharing/test.feature#L37

So, we need to refactor these type of steps and use password from a global default user values and use that. The deafult values are defined here https://github.com/owncloud/client/blob/a64b638468455a6e2edf185578d7c3be3a888d9b/test/gui/shared/steps/steps.py#L33-L49

PR #8656, opened 2021-05-25
[tests-only] refactor steps with password in them

swoichha:
## Description
This PR refactors all the steps with password in them

## Related Issue
- fixes https://github.com/owncloud/client/issues/8652

### DIFF
--- a/test/gui/shared/steps/steps.py
+++ b/test/gui/shared/steps/steps.py
@@ -87,9 +87,21 @@ def step(context, displayname, host):
     )
 
 
-@Given('user "|any|" has set up a client with default settings and password "|any|"')
-def step(context, username, password):
-    setUpClient(context, username, confFilePath)
+def getDisplayname(username):
+    if username in defaultUsers.keys():
+        return defaultUsers[username]['displayname']
+
+
+def getPasswordForUser(username):
+    if username in defaultUsers.keys():
+        return defaultUsers[username]['password']
+
+
+@Given('user "|any|" has set up a client with default settings')
+def step(context, username):
+    password = getPasswordForUser(username)
+    displayName = getDisplayname(username)
+    setUpClient(context, username, displayName, confFilePath)
     enterUserPassword = EnterPassword()
     enterUserPassword.enterPassword(password)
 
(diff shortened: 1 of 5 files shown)


test/gui/shared/steps/steps.py
```diff
@@ -87,9 +87,18 @@ def step(context, displayname, host):
...
+    userFirstname = userId[0]
+    if userFirstname in defaultUsers.keys():
+        return defaultUsers[userFirstname]['password']
+
+
+@Given('user "|any|" has set up a client with default settings')
+def step(context, username):
+    password = getPasswordForUser(username)
     setUpClient(context, username, confFilePath)
+    snooze(10)
```
dpakach, line 101: why is this required?

test/gui/shared/steps/steps.py
```diff
@@ -87,9 +87,18 @@ def step(context, displayname, host):
     )
 
 
-@Given('user "|any|" has set up a client with default settings and password "|any|"')
-def step(context, username, password):
+def getPasswordForUser(username):
+    userId = username.split()
+    userFirstname = userId[0]
```
dpakach, line 92: I dont think this is required
this is equivalent to `userFirstname = username`

dpakach (changes requested):

test/gui/shared/steps/steps.py
swoichha: fixed

test/gui/shared/steps/steps.py
swoichha: It is for username which has both first and last name like `Alice Hansen` used here https://github.com/owncloud/client/blob/808284d75bb23a85de3c8068051995dbbf8e18d2/test/gui/tst_addAccount/test.feature#L23

dpakach: That's the display name of the user. the username cannot have spaces I think it would be better to either use only username or only displayname in these kinds of steps.

swoichha: https://github.com/owncloud/client/pull/8656/files#diff-821cad5db62830778c291de069b5182ff9bcdd90951faa2cec9d223e8efbcc20R91 will split the display-name and will only be using the first-name which is also the username

individual-it (approved):

dpakach (approved):

Issue #8651, opened 2021-05-21
Add more tests for desktop-client


swoichha:
## Description
Currently we have tests for adding new account, syncing, and sharing. There are still more tests that can be added. Some of them are mentioned below. 

## To-Do:
- [x] Login/Logout users (@swoichha )
- [x] Remove user(@swoichha )
- [x] Share files/folder with group(@Talank  )
- [x] Edit shared file/folder(rename received share item name/change content) (@swoichha )
- [x] Edit share permissions(@swoichha)
- [x] As a collaborator, share to a group to whom the file is already shared by author (@Talank )

swoichha:
CC: @michaelstingl @TheOneRing

swoichha:
## Regression Testplan

FROM: https://testlink.owncloud.team/linkto.php?tprojectPrefix=desktop&item=testcase&id=desktop-158





* [ ] Fresh install on **Win10** using the unbranded ownCloud client MSI installer.
     * [ ] The MSI details page has
         * Verified publisher: ownCloud GmbH
         * certificate expires > 1y
     * [ ] Settings tab, About has correct versions:
         * ownCloud client version
         * Qt 5.12.10
         * OpenSSL 1.1.1i
* [ ] Visually check all tabs
     * [ ] Tabs look correct
* [ ] VFS is enabled per default. 
     * [ ] Files are virtual (right click menu shows grayed out "free up local space")
* [ ] Sync a file to the server
     * [ ] File has been synced
* [ ] Check the overlay icons
     * [ ] Win10: No overlay icons if using VFS. Uses status column in explorer. Overlay icons are shown without VFS
* [ ] Share a file with another user on the same server
     * [ ] File has been shared successfully
* [ ] Create a public link
     * [ ] Public link has been created and can be opened in a browser
* [ ] disable / re-enable VFS
     * [ ] VFS has been disabled / enabled
     * [ ] explorer status column changes
 ----
* [ ] Install ownCloud desktop client on latest **macOS** using .pkg installer
     * [ ] Settings tab, About has correct versions:
         * ownCloud client version
         * Qt 5.12.10
         * OpenSSL 1.1.1i
* [ ] Visually check all tabs
     * [ ] Tabs look correct
 * [ ] Sync a file to the server
     * [ ] File has been synced
 * [ ] Check the overlay icons
     * [ ] Overlay icons are present
 * [ ] Share a file with another user on the same server
     * [ ] File has been shared successfully
 * [ ] Create a public link
[...]

PR #8663, opened 2021-05-27
[tests-only]logout user from client

swoichha:
## Description
This PR adds tests for:
- logout user from client

## Related Issue
- https://github.com/owncloud/client/issues/8651

### DIFF
--- a/test/gui/shared/scripts/pageObjects/Toolbar.py
+++ b/test/gui/shared/scripts/pageObjects/Toolbar.py
@@ -9,6 +9,50 @@ class Toolbar:
         "visible": 1,
         "window": names.settings_OCC_SettingsDialog,
     }
+    ACCOUNT_BUTTON = {
+        "container": names.settings_stack_QStackedWidget,
+        "name": "_accountToolbox",
+        "type": "QToolButton",
+        "visible": 1,
+    }
+    ACCOUNT_MENU = {
+        "type": "QMenu",
+        "unnamed": 1,
+        "visible": 1,
+        "window": names.settings_OCC_SettingsDialog,
+    }
+    SIGNED_OUT_TEXT_BAR = {
+        "container": names.settings_stack_QStackedWidget,
+        "name": "connectLabel",
+        "type": "QLabel",
+        "visible": 1,
+    }
 
     def clickActivity(self):
         squish.clickButton(squish.waitForObject(self.ACTIVITY_BUTTON))
+
+    def userLogout(self):
+        squish.sendEvent(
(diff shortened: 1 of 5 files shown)


test/gui/shared/steps/steps.py
```diff
@@ -535,3 +532,75 @@ def step(context, resource):
...
+    toolbar = Toolbar()
+    toolbar.userLogsIn()
+    password = getPasswordForUser(username)
+    enterUserPassword = EnterPassword()
+    enterUserPassword.enterPassword(password)
+
+
+@Then('user "|any|" should be connect to the client-UI')
+def step(context, username):
+    snooze(5)
```
Talank, line 605: It would be better to add a comment about why we need this snooze, and finding an alternative solution is a TODO.

test/gui/shared/scripts/pageObjects/Toolbar.py
```diff
@@ -0,0 +1,59 @@
...
+    def clickActivity(self):
+        squish.clickButton(squish.waitForObject(self.ACTIVITY_BUTTON))
+
+    def userLogout(self):
+        squish.sendEvent(
+            "QMouseEvent",
+            squish.waitForObject(self.ACCOUNT_BUTTON),
+            squish.QEvent.MouseButtonPress,
+            28,
+            10,
```
Talank, line 41:
```suggestion
            0,
            0,
```
Why 28,10? I think you can press 0,0 instead. 🤔

swoichha: fixed

test/gui/shared/steps/steps.py
```diff
@@ -535,3 +532,77 @@ def step(context, resource):
...
+    isUserSignedOut(context, username)
+
+
+@Given('user "|any|" has logged out of the client-UI')
+def step(context, username):
+    waitFor(
+        lambda: isFolderSynced(context.userData['clientSyncPath']),
+        context.userData['clientSyncTimeout'] * 1000,
+    )
+    snooze(5)
```
Talank, line 588: is there any specific reason for this snooze? Everything else looks good to me.

swoichha: It was added to see if folder was synced or not. Now there is no need for it so I have removed it.

Talank (commented):
LGTM 🥇

individual-it (commented):
code looks good, but all GUI tests fail :-(

individual-it (approved):

PR #8664, opened 2021-05-27
Fix a mac the socket api crash

TheOneRing:


jnweiger:
@michaelstingl this calls for a 2.8.2 and delays ownbrander deployment. Do you agree?

michaelstingl:
> @michaelstingl this calls for a 2.8.2 and delays ownbrander deployment. Do you agree?

yes, understood 👍

TheOneRing:
The crash reports we get are all on mac, so its probably related to the special rpc implementation on mac....

src/gui/socketapi/socketapi.cpp
```diff
@@ -479,7 +476,9 @@ void SocketApi::slotUpdateFolderView(Folder *f)
 
 void SocketApi::broadcastMessage(const QString &msg, bool doWait)
 {
-    for (const auto &listener : qAsConst(_listeners)) {
+    // operate on a copy of the list to prevent concurrency issues
+    const auto clients = _listeners.values();
```
fmoc, line 480: The Qt docs don't suggest that this method is thread-safe. I think there is no way to get around using a mutex.

fmoc (changes requested):

src/gui/socketapi/socketapi.cpp
```diff
@@ -296,15 +296,21 @@ void SocketApi::slotNewConnection()
...
         return;
     }
     qCInfo(lcSocketApi) << "New connection" << socket;
-    connect(socket, &QIODevice::readyRead, this, &SocketApi::slotReadSocket);
-    connect(socket, SIGNAL(disconnected()), this, SLOT(onLostConnection()));
-    connect(socket, &QObject::destroyed, this, &SocketApi::slotSocketDestroyed);
+    connect(socket, &SocketApiSocket::readyRead, this, &SocketApi::slotReadSocket);
+    connect(socket, &SocketApiSocket::disconnected, this, [socket] {
+        qCInfo(lcSocketApi) << "Lost connection " << socket;
+        // will trigger destroyed in the next execution of the main loop
```
fmoc, line 308: I would really appreciate a comment that indicates that this fixes a synchronization issue.

fmoc (approved):

src/gui/socketapi/socketapi.cpp
TheOneRing: the double remove was stupid in the first place

PR #8670, opened 2021-05-31
Fix migration of settings

TheOneRing:
Don't initialize a settings object just to check whether the file exists.

Fixes: https://github.com/owncloud/enterprise/issues/4597

This is intended to fix one issue, further work is planned in https://github.com/owncloud/client/issues/6213


michaelstingl:
@SamuAlfageme FYI

dschmidt (approved):

PR #8682, opened 2021-06-04
[tests-only] remove connection

swoichha:
## Description
This PR adds tests for removing connection after a user has been added to the desktop-client

## Related Issue
- https://github.com/owncloud/client/issues/8651

### DIFF
--- a/test/gui/shared/scripts/names.py
+++ b/test/gui/shared/scripts/names.py
@@ -3,7 +3,6 @@
 from objectmaphelper import *
 
 settings_OCC_SettingsDialog = {"name": "Settings", "type": "OCC::SettingsDialog", "visible": 1}
-settings_settingsdialog_toolbutton_Add_account_QToolButton = {"name": "settingsdialog_toolbutton_Add account", "type": "QToolButton", "visible": 1, "window": settings_OCC_SettingsDialog}
 owncloudWizard_OCC_OwncloudWizard = {"name": "owncloudWizard", "type": "OCC::OwncloudWizard", "visible": 1}
 owncloudWizard_label_2_QLabel = {"name": "label_2", "type": "QLabel", "visible": 1, "window": owncloudWizard_OCC_OwncloudWizard}
 leUrl_OCC_PostfixLineEdit = {"buddy": owncloudWizard_label_2_QLabel, "name": "leUrl", "type": "OCC::PostfixLineEdit", "visible": 1}
@@ -24,7 +23,6 @@ settings_settingsdialog_toolbutton_admin_localhost_QToolButton = {"name": "setti
 settings_stack_QStackedWidget = {"name": "stack", "type": "QStackedWidget", "visible": 1, "window": settings_OCC_SettingsDialog}
 stack_accountToolbox_QToolButton = {"container": settings_stack_QStackedWidget, "name": "_accountToolbox", "type": "QToolButton", "visible": 1}
 settings_QMenu = {"type": "QMenu", "unnamed": 1, "visible": 1, "window": settings_OCC_SettingsDialog}
-stack_Remove_connection_QPushButton = {"container": settings_stack_QStackedWidget, "text": "Remove connection", "type": "QPushButton", "unnamed": 1, "visible": 1}
 owncloudWizard_usernameLabel_QLabel = {"name": "usernameLabel", "type": "QLabel", "visible": 1, "window": owncloudWizard_OCC_OwncloudWizard}
 owncloudWizard_passwordLabel_QLabel = {"name": "passwordLabel", "type": "QLabel", "visible": 1, "window": owncloudWizard_OCC_OwncloudWizard}
 qFileDialog_fileNameLabel_QLabel = {"name": "fileNameLabel", "type": "QLabel", "visible": 1, "window": qFileDialog_QFileDialog}
--- a/test/gui/shared/scripts/pageObjects/AccountStatus.py
+++ b/test/gui/shared/scripts/pageObjects/AccountStatus.py
@@ -1,0 +1,48 @@
+import names
+import squish
+
+
+class AccountStatus:
+    ACCOUNT_BUTTON = {
+        "container": names.settings_stack_QStackedWidget,
+        "name": "_accountToolbox",
+        "type": "QToolButton",
+        "visible": 1,
+    }
+    ACCOUNT_MENU = {
+        "type": "QMenu",
+        "unnamed": 1,
+        "visible": 1,
+        "window": names.settings_OCC_SettingsDialog,
+    }
+    SIGNED_OUT_TEXT_BAR = {
+        "container": names.settings_stack_QStackedWidget,
+        "name": "connectLabel",
+        "type": "QLabel",
+        "visible": 1,
+    }
+
+    REMOVE_CONNECTION_BUTTON = {
+        "container": names.settings_stack_QStackedWidget,
+        "text": "Remove connection",
+        "type": "QPushButton",
+        "unnamed": 1,
+        "visible": 1,
+    }
+
+    def accountAction(self, action):
+        squish.sendEvent(
+            "QMouseEvent",
+            squish.waitForObject(self.ACCOUNT_BUTTON),
+            squish.QEvent.MouseButtonPress,
+            0,
+            0,
+            squish.Qt.LeftButton,
+            0,
+            0,
+        )
+        squish.activateItem(squish.waitForObjectItem(self.ACCOUNT_MENU, action))
+
+    def removeConnection(self):
+        self.accountAction("Remove")
+        squish.clickButton(squish.waitForObject(self.REMOVE_CONNECTION_BUTTON))
--- a/test/gui/shared/scripts/pageObjects/Toolbar.py
+++ b/test/gui/shared/scripts/pageObjects/Toolbar.py
@@ -9,50 +9,26 @@ class Toolbar:
         "visible": 1,
         "window": names.settings_OCC_SettingsDialog,
     }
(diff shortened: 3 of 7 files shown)


test/gui/shared/scripts/pageObjects/AccountStatus.py
```diff
@@ -0,0 +1,70 @@
...
+        squish.sendEvent(
+            "QMouseEvent",
+            squish.waitForObject(self.ACCOUNT_BUTTON),
+            squish.QEvent.MouseButtonPress,
+            0,
+            0,
+            squish.Qt.LeftButton,
+            0,
+        )
+        squish.activateItem(squish.waitForObjectItem(self.ACCOUNT_MENU, "Log in"))
```
Talank, line 56: I think you can tale the 2nd argument like `action` or something and combine these 2 methods to one. Because these 2 methods do exactly the same thing except for selecting either "login" or "log out".

Talank (changes requested):

test/gui/tst_removeAccount/test.feature
```diff
@@ -0,0 +1,25 @@
...
+    Given user "Alice" has been created on the server with default attributes and without skeleton files
+    And user "Brian" has been created on the server with default attributes and without skeleton files
+    And user "Alice" has set up a client with default settings
+    And the user has added an account with
+        | server      | %local_server%     |
+        | user        | Brian              |
+        | password    | AaBb2Cc3Dd4        |
+        | localfolder | %client_sync_path% |
+    When the user removes the account
+	Then an account should be displayed with the displayname Alice Hansen and host %local_server_hostname%
```
Talank, line 18: The spacing seems a little distorted

test/gui/shared/scripts/pageObjects/AccountStatus.py
swoichha: Done

test/gui/tst_removeAccount/test.feature
swoichha: fixed

test/gui/shared/scripts/pageObjects/AccountStatus.py
```diff
@@ -0,0 +1,58 @@
...
+            "QMouseEvent",
+            squish.waitForObject(self.ACCOUNT_BUTTON),
+            squish.QEvent.MouseButtonPress,
+            0,
+            0,
+            squish.Qt.LeftButton,
+            0,
+            0,
+        )
+        squish.activateItem(squish.waitForObjectItem(names.settings_QMenu, "Remove"))
```
Talank, line 57: instead of this, you can call loginLogoutAction("Remove"). And it would be nice if you rename the function `loginLogoutAction`, it sounds very specific. May be `accountAction` or something like that. 🤔

swoichha: Yes, that can be done but we still need to have `removeUser` method as it does more then just clicking on "Remove". I have refactored this method and made it more DRY.

Talank (commented):
LGTM

test/gui/shared/steps/steps.py
```diff
@@ -67,6 +67,16 @@ def hook(context):
         pass
 
 
+@Given('the user has added an account with')
+def step(context):
+    clickButton(
+        waitForObject(names.settings_settingsdialog_toolbutton_Add_account_QToolButton)
```
individual-it, line 73: shouldn't that be part of the page object?

test/gui/shared/steps/steps.py
```diff
@@ -588,3 +598,39 @@ def step(context, username):
...
+    host = substituteInLineCodes(context, host)
+    test.compare(
+        str(
+            waitForObjectExists(
+                {
+                    "name": "settingsdialog_toolbutton_" + displayname + "@" + host,
+                    "type": "QToolButton",
+                    "visible": 1,
+                }
+            ).text
```
individual-it, line 625: move into a page object if possible

test/gui/tst_removeAccount/test.feature
```diff
@@ -0,0 +1,25 @@
...
+  Scenario: remove an account
+    Given user "Alice" has been created on the server with default attributes and without skeleton files
+    And user "Brian" has been created on the server with default attributes and without skeleton files
+    And user "Alice" has set up a client with default settings
+    And the user has added an account with
+        | server      | %local_server%     |
+        | user        | Brian              |
+        | password    | AaBb2Cc3Dd4        |
+        | localfolder | %client_sync_path% |
+    When the user removes the account
```
individual-it, line 17: which account is to be removed? You have set up two

test/gui/shared/scripts/pageObjects/AccountStatus.py
```diff
@@ -0,0 +1,48 @@
...
+            squish.QEvent.MouseButtonPress,
+            0,
+            0,
+            squish.Qt.LeftButton,
+            0,
+            0,
+        )
+        squish.activateItem(squish.waitForObjectItem(self.ACCOUNT_MENU, action))
+
+    def removeUser(self):
```
individual-it, line 46: I'm astonished about how we remove a user through the desktop client ;-)
Maybe call it `removeConnection` or similar ?

individual-it (changes requested):

test/gui/tst_removeAccount/test.feature
swoichha: This will remove the account that was added later on. In this case the user that will be removed is `Brian`

test/gui/shared/steps/steps.py
swoichha: Fixed

test/gui/shared/steps/steps.py

test/gui/shared/scripts/pageObjects/AccountStatus.py
swoichha: I have now renamed it to `removeConnection`.

test/gui/shared/scripts/names.py
```diff
@@ -98,3 +96,4 @@
 oCC_ShareLinkWidget_checkBox_expire_QProgressIndicator = {"aboveWidget": oCC_ShareLinkWidget_lineEdit_password_QLineEdit, "container": qt_tabwidget_stackedwidget_OCC_ShareLinkWidget_OCC_ShareLinkWidget, "leftWidget": oCC_ShareLinkWidget_checkBox_expire_QCheckBox, "type": "QProgressIndicator", "unnamed": 1, "visible": 1}
 oCC_IssuesWidget_tableView_QTableView = {"container": qt_tabwidget_stackedwidget_OCC_IssuesWidget_OCC_IssuesWidget, "name": "_tableView", "type": "QTableView", "visible": 1}
 o_tableView_0_1_QModelIndex = {"column": 1, "container": oCC_IssuesWidget_tableView_QTableView, "row": 0, "type": "QModelIndex"}
+owncloudWizard_QWidget = {"type": "QWidget", "unnamed": 1, "visible": 1, "window": owncloudWizard_OCC_OwncloudWizard}
```
individual-it, line 99: seems to be unused

test/gui/shared/scripts/pageObjects/Toolbar.py
```diff
@@ -9,50 +10,29 @@ class Toolbar:
...
-        squish.sendEvent(
-            "QMouseEvent",
-            squish.waitForObject(self.ACCOUNT_BUTTON),
-            squish.QEvent.MouseButtonPress,
-            0,
-            0,
-            squish.Qt.LeftButton,
-            0,
-            0,
+    def isAccountDisplyed(self, displayname, host):
```
individual-it, line 26: better to get the text here and return it, the assertion should happen in the step

also there is a typo in `Displyed`

test/gui/shared/steps/steps.py
```diff
@@ -516,16 +524,16 @@ def step(context, resource, role):
 
 @When('the user logs out of the client-UI')
```
individual-it, line 525: just seeing now, the "logout" is of-course also per account, so we should define out of which account (if multiple are configured) the users logsout. Can be a separate PR

test/gui/shared/steps/steps.py
```diff
@@ -588,3 +596,30 @@ def step(context, username):
     # It might take some time for all files to sync and connect to ther server
     snooze(5)
     isUserSignedIn(context, username)
+
+
+@When("the user removes the connection")
```
individual-it, line 601: Which connection? If multiple are configured

maybe add `the last created` or `the currently selected` to the sentence.

individual-it (changes requested):

test/gui/tst_removeAccount/test.feature
individual-it: That need to be specified in the step (for less astonishment ;-) )

test/gui/shared/scripts/names.py
swoichha: removed

test/gui/shared/scripts/pageObjects/Toolbar.py
swoichha: fixed

test/gui/shared/steps/steps.py
swoichha: Yes, you're right about this. I will make a new PR for that.

test/gui/shared/steps/steps.py
swoichha: I have refactored this step to `@When('the user removes the connection for user "|any|" and host |any|')` so that we can remove connection for any user we want.

individual-it (approved):

PR #8683, opened 2021-06-05
Destroy UploadDevice instance in case it has not been opened in ReadOnly mode

alexeylysenko:
Memory leak occurs when new UploadDevice hasn't been opened in ReadOnly mode.


TheOneRing (approved):
Thanks for your contribution

PR #8685, opened 2021-06-05
Create JsonApiJob on a stack to avoid memory leaks

alexeylysenko:


TheOneRing:
Hi @alexeylysenko thanks for  your contribution.
All job objects are deleted automatically once they complete so I don't see a leak here.
What I find more concerning is the use of loop.exec() here which can cause issues.
I'll try to make the code more async and remove the need for a blocking loop.

alexeylysenko:
Hi @TheOneRing,
You're right, now I see deleteLater() call in AbstractNetworkJob.

PR #8687, opened 2021-06-05
Use qEnvironmentVariableIsEmpty instead of qgetenv().isEmpty()

alexeylysenko:
https://doc.qt.io/qt-5/qtglobal.html#qEnvironmentVariableIsEmpty

It's potentially much faster, and can't throw exceptions.


TheOneRing (approved):
Thanks for your contribution

Issue #8677, opened 2021-06-02
[QA] get information related to user through middleware


swoichha:
## Description 

Currently we get data like display-name, password, email, etc. related to a user using global default user values like here
https://github.com/owncloud/client/blob/2abdd8c8e373dbe9c8e9ab5292f14888c456ef34/test/gui/shared/steps/steps.py#L35-L51

But after https://github.com/owncloud/owncloud-test-middleware/pull/35 has been merged we can now information related to user through middleware using `state` endpoint of middleware.

## To-Do:
- [ ] Use `state` endpoint of middleware to request information related to created users 
- [ ] remove this constant https://github.com/owncloud/client/blob/2abdd8c8e373dbe9c8e9ab5292f14888c456ef34/test/gui/shared/steps/steps.py#L35

For reference also see: https://github.com/owncloud/owncloud-test-middleware/issues/42#issuecomment-852005740

PR #8691, opened 2021-06-07
[Tests-Only] get userinformation through middleware

swoichha:
## Description
This PR adds a function called `getMiddlewareState` which will get user information from middleware using `state` endpoint of middleware. Previously we were using a global variable `defaultUsers` which now is not needed and thus removed.

## Related Issue
- Fixes https://github.com/owncloud/client/issues/8677

### DIFF
--- a/test/gui/shared/steps/steps.py
+++ b/test/gui/shared/steps/steps.py
@@ -33,20 +33,22 @@ confdir = '/tmp/bdd-tests-owncloud-client/'
 confFilePath = confdir + 'owncloud.cfg'
 socketConnect = None
 
-usersDataFromMiddleware = None
+stateDataFromMiddleware = None
 
 
-@functools.lru_cache(maxsize=None)
 def getTestStateFromMiddleware(context):
-    res = requests.get(
-        os.path.join(context.userData['middlewareUrl'], 'state'),
-        headers={"Content-Type": "application/json"},
-    )
-    try:
-        global usersDataFromMiddleware
-        usersDataFromMiddleware = res.json()
-    except ValueError:
-        raise Exception("Could not get created users information from middleware")
+    global stateDataFromMiddleware
+    if stateDataFromMiddleware is None:
+        res = requests.get(
+            os.path.join(context.userData['middlewareUrl'], 'state'),
+            headers={"Content-Type": "application/json"},
+        )
+        try:
+            stateDataFromMiddleware = res.json()
+        except ValueError:
+            raise Exception("Could not get created users information from middleware")
+
+    return stateDataFromMiddleware
 
 
 @OnScenarioStart
@@ -59,11 +61,6 @@ def hook(context):
         os.remove(confFilePath)
     except:
         pass
-
-
-@OnScenarioEnd
-def hook(context):
-    getTestStateFromMiddleware.cache_clear()
 
 
 @Given('the user has added an account with')
@@ -100,10 +97,12 @@ def step(context, displayname, host):
 
 
 def getDisplaynameForUser(context, username):
+    usersDataFromMiddleware = getTestStateFromMiddleware(context)
     return usersDataFromMiddleware['created_users'][username]['displayname']
 
 
 def getPasswordForUser(context, username):
+    usersDataFromMiddleware = getTestStateFromMiddleware(context)
     return usersDataFromMiddleware['created_users'][username]['password']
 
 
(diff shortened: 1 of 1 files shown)


test/gui/shared/steps/steps.py
```diff
@@ -36,23 +36,24 @@
...
-        'displayname': 'Carol King',
-        'password': passwords['alt3'],
-        'email': 'carol@example.org',
-    },
-}
+defaultUsers = ''
+middlewareState = False
+
+
+def sendRequestToMiddlewareState(context):
```
kiranparajuli589, line 43:
```suggestion
def sendRequestToGetMiddlewareState(context):
```

test/gui/shared/steps/steps.py
```diff
@@ -36,23 +36,24 @@
...
+    res = urllib.request.urlopen(
+        os.path.join(context.userData['middlewareUrl'], 'state'),
+    )
+
+    try:
+        global defaultUsers
+        defaultUsers = res.read()
+        defaultUsers = json.loads(defaultUsers)
+        print(defaultUsers)
```
kiranparajuli589, line 52: debug line?

test/gui/shared/steps/steps.py
```diff
@@ -100,20 +101,25 @@ def step(context, displayname, host):
...
 
-def getDisplayname(username):
-    if username in defaultUsers.keys():
-        return defaultUsers[username]['displayname']
+def checkMiddlewareState(context):
+    if not middlewareState:
+        sendRequestToMiddlewareState(context)
+
+
+def getDisplayname(context, username):
```
kiranparajuli589, line 109:
```suggestion
def getDisplaynameForUser(context, username):
```

kiranparajuli589 (changes requested):
if the when steps can change the user/group details, remember to fetch the middleware endpoint again to get the latest state.

test/gui/shared/steps/steps.py
```diff
@@ -36,23 +36,24 @@
...
-        'password': passwords['alt2'],
-        'email': 'brian@example.org',
-    },
-    'Carol': {
-        'displayname': 'Carol King',
-        'password': passwords['alt3'],
-        'email': 'carol@example.org',
-    },
-}
+defaultUsers = ''
```
kiranparajuli589, line 39: set this to null

test/gui/shared/steps/steps.py
```diff
@@ -100,20 +101,25 @@ def step(context, displayname, host):
     )
 
 
-def getDisplayname(username):
-    if username in defaultUsers.keys():
-        return defaultUsers[username]['displayname']
+def checkMiddlewareState(context):
+    if not middlewareState:
```
kiranparajuli589, line 105:
```suggestion
    if not defaultUsers:
```

Talank (commented):
LGTM other than already existing reviews.

test/gui/shared/steps/steps.py
swoichha: Yes :sweat_smile:

test/gui/shared/steps/steps.py

test/gui/shared/steps/steps.py

test/gui/shared/steps/steps.py

test/gui/shared/steps/steps.py

kiranparajuli589 (commented):
lgtm

kiranparajuli589 (approved):

test/gui/shared/steps/steps.py
```diff
@@ -30,29 +31,19 @@
...
-    'Carol': {
-        'displayname': 'Carol King',
-        'password': passwords['alt3'],
-        'email': 'carol@example.org',
-    },
-}
+defaultUsers = None
+
+
+def sendRequestToGetMiddlewareState(context):
```
individual-it, line 37:
```suggestion
def getTestStateFromMiddleware(context):
```

test/gui/shared/steps/steps.py
```diff
@@ -30,29 +31,19 @@
...
+    res = requests.get(
+        os.path.join(context.userData['middlewareUrl'], 'state'),
+        headers={"Content-Type": "application/json"},
+    )
+    try:
+        global defaultUsers
```
individual-it, line 43: is the middleware returning "default" users or the really created users? If there are the users that have been created during the test runs, we maybe should not cache them here, because they might change

individual-it (changes requested):

test/gui/shared/steps/steps.py
swoichha: It returns created users. In cases where users are created during test runs like in When step we need to fetch the middleware endpoint again.

test/gui/shared/steps/steps.py

individual-it (approved):

individual-it (approved):

PR #8695, opened 2021-06-08
[tests-only] Group share tests added

Talank:
The following test scenarios have been added:
- Collaborator should not see to whom a file is shared.
- Group sharing
- User (non-author) can not share to a group to which the file is already shared

### Related issues:
- part of https://github.com/owncloud/client/issues/8651

### DIFF
--- a/test/gui/shared/steps/steps.py
+++ b/test/gui/shared/steps/steps.py
@@ -205,10 +205,31 @@ def step(context, receiver, resource, permissions):
     shareItem.addCollaborator(receiver, permissions)
 
 
+@When(
+    'the user adds group "|any|" as collaborator of resource "|any|" with permissions "|any|" using the client-UI'
+)
+def step(context, receiver, resource, permissions):
+    openSharingDialog(context, resource)
+    shareItem = SharingDialog()
+    shareItem.addCollaborator(receiver, permissions, True)
+
+
 @Then(
     'user "|any|" should be listed in the collaborators list for file "|any|" with permissions "|any|" on the client-UI'
 )
 def step(context, receiver, resource, permissions):
+    collaboratorShouldBeListed(context, receiver, resource, permissions)
+
+
+@Then(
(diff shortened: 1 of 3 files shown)


test/gui/shared/steps/steps.py
```diff
@@ -646,3 +672,24 @@ def step(context):
 def step(context):
     for tabName in context.table:
         test.vp(tabName[0])
+
+
+@Then('the error "|any|" should be displayed')
```
swoichha, line 677: We can also mention where this error is to be displayed like you have done in line 466

test/gui/shared/steps/steps.py
```diff
@@ -646,3 +672,24 @@ def step(context):
...
+def step(context, errorMessage):
+    sharingDialog = SharingDialog()
+    test.compare(sharingDialog.getErrorText(), errorMessage)
+
+
+@When("the user closes the sharing dialog")
+def step(context):
+    clickButton(waitForObject(names.sharingDialog_Close_QPushButton))
```
swoichha, line 685: This has not been used anywhere. So, you can remove this.

Talank: nice find. Thanks

test/gui/shared/steps/steps.py
Talank: Technically, it looks for an error label that can be found using the selector `{"name": "errorLabel", "type": "QLabel", "visible": 1}`. So, I think it should work for any error in general. 🤔 And IMO mentioning where the error is (for example `... should be displayed in sharing dialog`) brings more confusion.

swoichha: If you think this way is better then that's fine. I just mentioned it because this step seems a bit generic.

Talank:
Need to run after scenario to cleanup the state from middleware before merging this PR. So, blocked until https://github.com/owncloud/client/pull/8792 is merged.

swoichha (approved):
LGTM

PR #8696, opened 2021-06-08
Fix deadlock introduced in 8fcc9adcc41f2b46a74b9c0d806d2f4721081f8a

TheOneRing:
Microsoft doc:
    Do not call GdiplusStartup or GdiplusShutdown in DllMain or in any function that is called by DllMain


dschmidt (approved):
🙈

PR #8699, opened 2021-06-08
Refactor the cmd client

TheOneRing:
Remove the use of a event loop before the main loop is even started

I'm sorry that this got bigger than planned...


src/cmd/cmd.cpp
```diff
@@ -526,47 +519,43 @@ int main(int argc, char **argv)
...
+    ctx.account->setUrl(ctx.credentialFreeUrl);
+    ctx.account->setSslErrorHandler(sslErrorHandler);
 
     // Perform a call to get the capabilities.
-    QEventLoop loop;
-    JsonApiJob *job = new JsonApiJob(account, QLatin1String("ocs/v1.php/cloud/capabilities"));
-    QObject::connect(job, &JsonApiJob::jsonReceived, [&](const QJsonDocument &json) {
+    auto capabilitiesJob = new JsonApiJob(ctx.account, QLatin1String("ocs/v1.php/cloud/capabilities"));
+    QObject::connect(capabilitiesJob, &JsonApiJob::jsonReceived, qApp, [capabilitiesJob, ctx](const QJsonDocument &json) {
         auto caps = json.object().value("ocs").toObject().value("data").toObject().value("capabilities").toObject();
```
alexeylysenko, line 538: Shall the content of the received json document be checked before parsing started? I mean if the document structure is malformed there is no sense to parse it and set unrelevant values into account instance. The existing code won't crash application, but wouldn't it be better to highlight that something went wrong on an early stage.

TheOneRing: The check is done in https://github.com/owncloud/client/blob/master/src/libsync/networkjobs.cpp#L793 but it doesn't look that sane.

I think in case of an error we should emit another signal 😞 ...

alexeylysenko: Yeah, QJsonParseError just checks whether json content is valid. I meant a bit different situation, when, for example, there is no "ocs" value inside root object or even root is array. https://doc.qt.io/qt-5/qjsondocument.html#isObject

TheOneRing: So proper handling of json errors will come in another review ....

dschmidt (approved):

PR #8700, opened 2021-06-08
Removed unnecessary temporary containers during iteration over QMap values

alexeylysenko:
With QMap and QHash, foreach accesses the value component of the (key, value) pairs automatically, so you should not call values() on the container (it would generate an unnecessary copy, see below).

https://doc.qt.io/qt-5/containers.html#foreach

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -823,7 +823,7 @@ void AccountSettings::slotAccountStateChanged()
         QUrl safeUrl(account->url());
         safeUrl.setPassword(QString()); // Remove the password from the URL to avoid showing it in the UI
         FolderMan *folderMan = FolderMan::instance();
-        for (Folder *folder : folderMan->map().values()) {
+        for (Folder *folder : folderMan->map()) {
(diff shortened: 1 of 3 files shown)


TheOneRing:
Hi that's looking good :)

Are you using clazy?

For the loops I'd love if you could also port them to c++11 loops instead of the foreach macro provided by Qt.
With the c++ loops its usually recommended to use qAsConst() where possible to prevent accidental detaching.

TheOneRing:
I just saw https://github.com/owncloud/client/pull/8702/files :D
So foreach is the same define as Q_FOREACH

alexeylysenko:
> Hi that's looking good :)
> 
> Are you using clazy?

Not yet, I have just checked the most common issues that might occur within Qt application. I can check the project with clazy the next days if it's not been done yet (not promise, still have a tough schedule).

TheOneRing (approved):

alexeylysenko:
Would it be possible to merge this PR, so I could prepare the new one based on this change?

TheOneRing:
Sry thought my approval is enough :)

PR #8702, opened 2021-06-08
Removed obsolete Q_FOREACH macro

alexeylysenko:
Note: Since Qt 5.7, the use of this macro is discouraged. It will be removed in a future version of Qt. Please use C++11 range-for, possibly with qAsConst(), as needed.

https://doc.qt.io/qt-5/qtglobal.html#Q_FOREACH

### DIFF
--- a/src/gui/accountmanager.cpp
+++ b/src/gui/accountmanager.cpp
@@ -255,10 +255,8 @@ void AccountManager::saveAccountHelper(Account *acc, QSettings &settings, bool s
             acc->_credentials->persist();
         }
 
-        auto i = acc->_settingsMap.constBegin();
-        while (i != acc->_settingsMap.constEnd()) {
-            settings.setValue(i.key(), i.value());
-            ++i;
+        for (auto it = acc->_settingsMap.constBegin(); it != acc->_settingsMap.constEnd(); ++it) {
+            settings.setValue(it.key(), it.value());
         }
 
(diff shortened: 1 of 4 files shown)


src/gui/accountmanager.cpp
```diff
@@ -254,8 +254,11 @@ void AccountManager::saveAccountHelper(Account *acc, QSettings &settings, bool s
             // re-persisting them)
             acc->_credentials->persist();
         }
-        Q_FOREACH (QString key, acc->_settingsMap.keys()) {
-            settings.setValue(key, acc->_settingsMap.value(key));
+
+        auto i = acc->_settingsMap.constBegin();
+        while (i != acc->_settingsMap.constEnd()) {
```
TheOneRing, line 259: I'd prefer `for(auto it = acc->_settingsMap.constBegin(); it != acc->_settingsMap.constEnd(); ++it)`

TheOneRing (changes requested):
nitpick :)

src/gui/accountmanager.cpp
alexeylysenko: Done! :)

TheOneRing (approved):
Thanks!

alexeylysenko:
Would it be possible to merge this PR, so I could prepare the new one based on this change?

PR #8704, opened 2021-06-08
Replaced obsolete foreach-loops with for-loops

alexeylysenko:
As requested here https://github.com/owncloud/client/pull/8700#issuecomment-856833445 foreach-loops have been replaced with for-loops.
However, the issue with unnecessary temporary containers is still covered here https://github.com/owncloud/client/pull/8700


src/common/ownsql.cpp
```diff
@@ -190,7 +190,9 @@ QString SqlDatabase::error() const
 void SqlDatabase::close()
 {
     if (_db) {
-        for (auto *q : qAsConst(_queries)) {
+        // we need a copy because query is removed from a set in finish()
+        const auto queries = _queries;
```
TheOneRing, line 194: how did you spot this one?

src/gui/activitywidget.cpp
```diff
@@ -588,8 +588,8 @@ void ActivitySettings::slotRefresh(AccountState *ptr)
 
 void ActivitySettings::slotRegularNotificationCheck()
 {
-    AccountManager *am = AccountManager::instance();
-    foreach (AccountStatePtr a, am->accounts()) {
+    const auto accounts = AccountManager::instance()->accounts();
```
TheOneRing, line 591: Its not obvious that in that case accounts  is a const reference so it looks like an explicit copy for no reason.
Any reason for that, besides an unrequired limit of 80 chars :)

src/common/ownsql.cpp
alexeylysenko: The tests were failed with the first commit. And after brief investigation I found this code
```
void SqlQuery::finish()
{
    ...
    if (_sqldb) {
        _sqldb->_queries.remove(this);
    }
}
```

src/gui/activitywidget.cpp
alexeylysenko: You're completely right, I pushed the fix.

TheOneRing:
Btw are you interested in joining our rocket chat talk.owncloud.com/

alexeylysenko:
> Btw are you interested in joining our rocket chat talk.owncloud.com/

Yes, I am. Btw, are you able to restart CI job on this pull request? It has failed with 
```/bin/sh: 35: cd: can't cd to docs/```

TheOneRing:
I can but after the recent merges I'd recommend to rebase and force push :)

alexeylysenko:
I did merge from master

PR #8709, opened 2021-06-09
Update translations in GitHub actions

fmoc:
Right now, translations are updated on Drone. This PR moves the task to GitHub actions.

### DIFF
--- a/.github/workflows/translate.yml
+++ b/.github/workflows/translate.yml
@@ -1,0 +1,49 @@
+name: "Update translations"
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 3 * * *"
+
+jobs:
+  update-translations:
+    runs-on: ubuntu-latest
+
+    env:
+      TX_TOKEN: ${{ secrets.TX_TOKEN }}
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: recursive
+
+      - name: l10n-read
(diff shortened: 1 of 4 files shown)


.github/workflows/translate.yml
```diff
@@ -0,0 +1,46 @@
...
+        run: docker run --rm -it -e TX_TOKEN -v (readlink -f .):/ws rabits/qt:5.12-desktop bash -c "cd /ws/translations && make l10n-read"
+
+      - name: l10n-push
+        run: cd translations && make l10n-push
+
+      - name: l10n-pull
+        run: cd translations && make l10n-pull
+
+      - name: l10n-write
+        run: docker run --rm -it -e TX_TOKEN -v (readlink -f .):/ws owncloudci/transifex:latest bash -c "cd /ws/translations && make l10n-write"
```
TheOneRing, line 30: `TX_TOKEN=${{ secrets.TX_TOKEN }}`

TheOneRing: And maybe we can also just kill the stupid make file and call the stuff from here?

fmoc: I injected the secret as an environment variable once for the entire job, like in the Drone job. I just need to forward the value here. I'm not sure the value isn't used elsewhere as well.

TheOneRing: @xoxys any idea?

xoxys: No idea

TheOneRing (approved):

Issue #8705, opened 2021-06-09
[QA] Next steps of automated desktop testing


Talank:
For next steps in test automation in the desktop client, we need to do the following tasks:
- [x] Create docs that clarify how to run/write tests locally.
- [x] Automating Smoke tests should be the priority on writing new tests, https://github.com/owncloud/QA/blob/master/Desktop/Regression_Test_Plan_Patch_Release.md
- [x] Pin latest core release.

Talank:
There is already a list of tests to add. https://github.com/owncloud/client/issues/8651

PR #8719, opened 2021-06-11
[Tests-Only]Run smokeTest on CI

swoichha:
## Description
Currently we run all GUI tests in same pipeline name `GUI-tests`. This is a POC PR for running smokeTest in CI.

 Here, we have separated GUI tests run into two pipeline `GUI-tests-@smokeTest` where all tests with tag `@smokeTest`  will run and `GUI-tests-~@smokeTest` where other remaining tests will run. These pipelines does not run parallelly. First `GUI-tests-@smokeTest` will run and on passing  `GUI-tests-~@smokeTest` will run. 

All the normal tests depends on smokeTest. This way we also will be achieving early failure.

## Related Issue
- https://github.com/owncloud/client/issues/8705

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -50,8 +50,8 @@ def main(ctx):
             "Ninja",
             trigger = build_trigger,
         ),
-        gui_tests(ctx, trigger = build_trigger, filterTags = ["smokeTest"]),
-        gui_tests(ctx, trigger = build_trigger, depends_on = ["GUI-tests-smokeTest"]),
+        gui_tests(ctx, trigger = build_trigger, filterTags = ["@smokeTest"]),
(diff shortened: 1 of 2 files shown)


Talank (commented):
LGTM

individual-it (commented):
I don't understand what we are trying to achieve here.
Currently all UI tests run in CI, now this will create an extra pipeline to run tests that are marked as smoketests again. Why?

swoichha:
> I don't understand what we are trying to achieve here.
This PR is a POC for running smoke tests in CI.

> Currently all UI tests run in CI, now this will create an extra pipeline to run tests that are marked as smoketests again. Why?
Yes, there will be two pipeline. Need to fix that and run smoke test only in the new pipeline and remaining other tests on the older pipeline.

individual-it (approved):

PR #8725, opened 2021-06-14
[Tests-Only] Visually check all tabs

swoichha:
## Description
This PR adds test to check if all the tabs in toolbar section is visible by property verification point creation and suing squish api `test.vp` 

## Related Issue
 - https://github.com/owncloud/client/issues/8705

Also see: https://github.com/owncloud/client/issues/8651#issuecomment-857373491

### DIFF
--- a/test/gui/shared/steps/steps.py
+++ b/test/gui/shared/steps/steps.py
@@ -640,3 +640,9 @@ def step(context):
         'Ser&ver Address',
     )
     waitForObject(AccountConnectionWizard.SERVER_ADDRESS_BOX)
+
+
+@Then("the following tabs in the toolbar should match the default baseline")
+def step(context):
+    for tabName in context.table:
+        test.vp(tabName[0])
--- a/test/gui/tst_checkAlltabs/verificationPoints/Activity
+++ b/test/gui/tst_checkAlltabs/verificationPoints/Activity
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<VerificationPoint type="ObjectProperties" version="4">
+<Description/>
+<Verification object="settings_settingsdialog_toolbutton_Activity_QToolButton" property="pos.x" type="int" valueType="Value"><![CDATA[195]]>
+<base64EncodedValue>MTk1</base64EncodedValue>
+</Verification>
+<Verification object="settings_settingsdialog_toolbutton_Activity_QToolButton" property="pos.y" type="int" valueType="Value"><![CDATA[0]]>
+<base64EncodedValue>MA==</base64EncodedValue>
+</Verification>
+<Verification object="settings_settingsdialog_toolbutton_Activity_QToolButton" property="text" type="QString" valueType="Value"><![CDATA[Activity]]>
+<base64EncodedValue>QWN0aXZpdHk=</base64EncodedValue>
+</Verification>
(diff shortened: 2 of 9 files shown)


test/gui/tst_checkAlltabs/verificationPoints/Activity
```diff
@@ -0,0 +1,16 @@
...
+<Verification object="settings_settingsdialog_toolbutton_Activity_QToolButton" property="pos.x" type="int" valueType="Value"><![CDATA[195]]>
+<base64EncodedValue>MTk1</base64EncodedValue>
+</Verification>
+<Verification object="settings_settingsdialog_toolbutton_Activity_QToolButton" property="pos.y" type="int" valueType="Value"><![CDATA[0]]>
+<base64EncodedValue>MA==</base64EncodedValue>
+</Verification>
+<Verification object="settings_settingsdialog_toolbutton_Activity_QToolButton" property="text" type="QString" valueType="Value"><![CDATA[Activity]]>
+<base64EncodedValue>QWN0aXZpdHk=</base64EncodedValue>
+</Verification>
+<Verification object="settings_settingsdialog_toolbutton_Activity_QToolButton" property="visible" type="bool" valueType="Value"><![CDATA[true]]>
```
Talank, line 13: is there any way to get these objects from page objects instead of `names.py`?  🤔

swoichha: I don't think we can do that. They are auto generated when recording steps using verification point just like it is done here https://github.com/owncloud/client/blob/5211ff80696190e635fd9f7d77ddaf0074aaa569/test/gui/shared/verificationPoints/publicLinkExpirationProgressIndicatorInvisible#L4

Talank (commented):
LGTM

test/gui/shared/steps/steps.py
```diff
@@ -640,3 +640,11 @@ def step(context):
         'Ser&ver Address',
     )
     waitForObject(AccountConnectionWizard.SERVER_ADDRESS_BOX)
+
+
+@Then("all tabs in the toolbar should match the default baseline")
```
individual-it, line 645: could we make the tab names as parameter? So that every can be tested by itself

individual-it (changes requested):
CI fails with some linting issue, not sure if its related

swoichha:
> CI fails with some linting issue, not sure if its related

It has nothing to so with this PR. `test-clang-format` has been failing in CI for 2 days now.

test/gui/shared/steps/steps.py

test/gui/shared/steps/steps.py
```diff
@@ -640,3 +640,11 @@ def step(context):
         'Ser&ver Address',
     )
     waitForObject(AccountConnectionWizard.SERVER_ADDRESS_BOX)
+
+
+@Then("the following tabs in the toolbar should match the default baseline")
+def step(context):
+
+    for tabName in context.table:
+        print(tabName)
```
individual-it, line 649: do we need this print?

individual-it (changes requested):

test/gui/shared/steps/steps.py
swoichha: No we don't need that it was added just for debug purpose. Now it's removed.

individual-it (approved):
please squash on merge

PR #8726, opened 2021-06-14
A blocking connection validator has a higher priority than a non bloc…

TheOneRing:
…king


dschmidt:
The commit message sounds more like a bug report than a commit message. Reading it I don't know what the previous and the after state is.

dschmidt (approved):
ok, I guess 🙈

PR #8727, opened 2021-06-14
Remove possible qt containers detach

alexeylysenko:
This PR is based on #8704 with fixes to avoid possible detach in for-loops

### DIFF
--- a/src/gui/openfilemanager.cpp
+++ b/src/gui/openfilemanager.cpp
@@ -63,7 +63,7 @@ static QString findDefaultFileManager()
     QFileInfo fi;
     const auto &dirs = xdgDataDirs();
     for (const auto &dir : dirs) {
-        const QStringList subdirs{QStringLiteral("/applications/"), QStringLiteral("/applications/kde4/")};
+        const QStringList subdirs { QStringLiteral("/applications/"), QStringLiteral("/applications/kde4/") };
         for (const auto &subdir : subdirs) {
             fi.setFile(dir + subdir + fileName);
             if (fi.exists()) {
--- a/src/gui/openfilemanager.cpp
+++ b/src/gui/openfilemanager.cpp
@@ -62,8 +62,8 @@ static QString findDefaultFileManager()
 
     QFileInfo fi;
     const auto &dirs = xdgDataDirs();
-    const QStringList subdirs{"/applications/", "/applications/kde4/"};
     for (const auto &dir : dirs) {
+        const QStringList subdirs{QStringLiteral("/applications/"), QStringLiteral("/applications/kde4/")};
         for (const auto &subdir : subdirs) {
             fi.setFile(dir + subdir + fileName);
             if (fi.exists()) {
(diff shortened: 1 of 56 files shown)


src/gui/openfilemanager.cpp
```diff
@@ -61,12 +61,10 @@ static QString findDefaultFileManager()
...
 
     QFileInfo fi;
-    QStringList dirs = xdgDataDirs();
-    QStringList subdirs;
-    subdirs << "/applications/"
-            << "/applications/kde4/";
-    foreach (QString dir, dirs) {
-        foreach (QString subdir, subdirs) {
+    const auto &dirs = xdgDataDirs();
+    const QStringList subdirs{"/applications/", "/applications/kde4/"};
```
TheOneRing, line 65: Can you move the list directly to the loop.
Also when touching strings can you  pls convert them to QStringLiteral

src/gui/servernotificationhandler.cpp
```diff
@@ -65,7 +65,7 @@ void ServerNotificationHandler::slotNotificationsReceived(const QJsonDocument &j
         return;
     }
 
-    auto notifies = json.object().value("ocs").toObject().value("data").toArray();
+    const auto &notifies = json.object().value("ocs").toObject().value("data").toArray();
```
TheOneRing, line 68: QStringLiteral

src/libsync/discovery.cpp
```diff
@@ -242,17 +242,15 @@ bool ProcessDirectoryJob::handleExcluded(const QString &path, const QString &loc
             if (item->_file.endsWith(QLatin1Char('.'))) {
                 item->_errorString = tr("File names ending with a period are not supported on this file system.");
             } else {
-                char invalid = '\0';
-                for (const char x : QByteArrayLiteral("\\:?*\"<>|")) {
-                    if (item->_file.contains(QLatin1Char(x))) {
-                        invalid = x;
+                static const QString unsupportedCharacters = QStringLiteral("\\:?*\"<>|");
```
TheOneRing, line 245: the string literal is already static https://code.woboq.org/qt5/qtbase/src/corelib/text/qstringliteral.h.html#_M/QStringLiteral

TheOneRing (changes requested):

TheOneRing:
Some nitpicks the rest looks ok I guess.
Could you rebase your review to remove the merge commits from your history?

TheOneRing:
And see https://github.com/owncloud/client/pull/8737/commits/8d04d6b1721240b11303b55c5cfbc705d5b0daf6
I'm currently trying to enable the github actions on pull requests so lets see whether we will have some after your rebase.

src/gui/servernotificationhandler.cpp
alexeylysenko: QLatin1String would be better there

src/gui/openfilemanager.cpp
alexeylysenko: Done. I didn't touch strings because the PR is about detaching qt containers in range-based for loop, so I didn't want to mix the things up. If you want so I can add it, no problem.

PR #8731, opened 2021-06-15
ci: replace unmaintained alpine step image

xoxys:


TheOneRing:
yay another funny merge that I'll get

TheOneRing:
Narf why did I merge that, changelog is only generated on master...

xoxys:
Not true this part runs on pulls as well

TheOneRing:
ah yes your right

PR #8739, opened 2021-06-16
[Tests-Only] Run GUI tests on nightly

swoichha:
### Description
Currently we do not run GUI tests on nightly. So, this PR will run all the GUI tests against the `daily-master-qa` server and  run PRs against a stable server tarball `10.7.0`.

After this PR gets merged we need to change the server version from `10.7.0` to latest stable release when new stable version of core is released.

PR #8744, opened 2021-06-16
[Tests-Only] Revert changes done in PR-8735

swoichha:
### Description
After https://github.com/owncloud/client/pull/8739 has been merged GUI will run on latest stable server tarball. So, we can now undo all the changes done in PR https://github.com/owncloud/client/pull/8735.


phil-davis (approved):

phil-davis:
And the "23:59:59" change in server core has been reverted today. So the `daily-master-qa` server core tarball that is built tonight should behave nicely. I expect that the nightly CI will pass against `daily-master-qa`

PR #8747, opened 2021-06-16
Replace non-const QHash::find with const version

alexeylysenko:


src/cmd/netrcparser.cpp
```diff
@@ -97,14 +97,10 @@ bool NetrcParser::parse()
...
 
-NetrcParser::LoginPair NetrcParser::find(const QString &machine)
+NetrcParser::LoginPair NetrcParser::find(const QString &machine) const
 {
-    QHash<QString, LoginPair>::const_iterator it = _entries.find(machine);
-    if (it != _entries.end()) {
-        return *it;
-    } else {
-        return _default;
-    }
```
TheOneRing, line 107: How about just https://doc.qt.io/qt-5/qhash.html#value-1

alexeylysenko: Right, I was concentrated on non-const find and non-const iterator. 
It's pretty the same underhood, but your variant is more readable.

PR #8750, opened 2021-06-17
Fix a bug preventing sync root on drive roots

TheOneRing:
Fixes: https://github.com/owncloud/enterprise/issues/4639

### DIFF
--- a/src/common/vfs.cpp
+++ b/src/common/vfs.cpp
@@ -69,7 +69,7 @@ Result<bool, QString> Vfs::checkAvailability(const QString &path)
 #ifdef Q_OS_WIN
     if (mode == Mode::WindowsCfApi) {
         const auto info = QFileInfo(path);
-        if (QDir(info.canonicalPath()).isRoot()) {
+        if (QDir(info.canonicalFilePath()).isRoot()) {
             return tr("The Virtual filesystem feature does not support a drive as sync root");
         }
         const auto fs = FileSystem::fileSystemForPath(info.absoluteFilePath());
--- a/changelog/unreleased/4639
+++ b/changelog/unreleased/4639
@@ -1,0 +1,6 @@
+Bugfix: VFS support for folders in the drive root
+
+We fixed a bug where it was not possible to use a folder in 
+C:\ as sync folder.
+
+https://github.com/owncloud/enterprise/issues/4639


src/common/vfs.cpp
```diff
@@ -69,7 +69,7 @@ Result<bool, QString> Vfs::checkAvailability(const QString &path)
 #ifdef Q_OS_WIN
     if (mode == Mode::WindowsCfApi) {
         const auto info = QFileInfo(path);
-        if (QDir(info.canonicalPath()).isRoot()) {
+        if (QDir(info.canonicalFilePath()).isRoot()) {
```
alexeylysenko, line 72: What about 
```bool QFileInfo::isRoot() const```?

Comment (TheOneRing) [reply src/common/vfs.cpp:72]: We need to call canonicalFilePath as explained in https://doc.qt.io/qt-5/qdir.html#isRoot 
Else C:\\..\\ is not reported as a root drive.
I don't think it will make much difference whether we create a temporary QDir or a temporary QFileInfo

Comment (alexeylysenko) [reply src/common/vfs.cpp:72]: Can't you just 
```
if (info.isRoot()) {
...
}```
without constructing any temporary object?

alexeylysenko: bool QFileInfo::isRoot() const

Returns true if the object points to a directory or **to a symbolic link to a directory**, and that directory is the root directory; otherwise returns false.

So I assume QFileInfo::isRoot() would cover the case regarding canonicalFilePath

TheOneRing:
```
int main(int argc, char *argv[])
{
    QApplication app(argc, argv);

    auto foo = [](const QString &s){
        QFileInfo info(s);
        qDebug() << s << info.canonicalFilePath();
        qDebug() << info.isRoot();
        qDebug() << QDir(info.canonicalFilePath()).isRoot();
    };
    foo("C:\\");
    foo("C:\\..\\");
    foo("C:");
    return app.exec();
}
```


```
"C:\\" "C:/"
true
true
"C:\\..\\" "C:/"
false
true
"C:" "C:/Users/hanna/Documents/build-untitled-ownCloud-Debug"
false
false
```

alexeylysenko: Thanks for the test example. Then I think this seems to be a Qt bug. At least based on documentation I would expect a different behavior.

alexeylysenko: FYI https://bugreports.qt.io/browse/QTBUG-94582

dschmidt (approved):

PR #8767, opened 2021-06-21
Fix build warnings

alexeylysenko:
Fixed the warning messages appeared during build except following:
1) Q_DECL_DEPRECATED_X("Use uuid") QString id() const;
    Not sure whether I can simply replace id() to uuid().
2) Qt deprecation warnings
    If there is a plan to move to Qt6 it would be very useful to fix all warnings in Qt 5.15 to make a migration smoother - can provide a separate PR for that.

### DIFF
--- a/src/gui/generalsettings.cpp
+++ b/src/gui/generalsettings.cpp
@@ -51,7 +51,9 @@ GeneralSettings::GeneralSettings(QWidget *parent)
 
     connect(_ui->desktopNotificationsCheckBox, &QAbstractButton::toggled,
         this, &GeneralSettings::slotToggleOptionalDesktopNotifications);
+#ifdef Q_OS_WIN
     connect(_ui->showInExplorerNavigationPaneCheckBox, &QAbstractButton::toggled, this, &GeneralSettings::slotShowInExplorerNavigationPane);
+#endif
 
     reloadConfig();
     loadMiscSettings();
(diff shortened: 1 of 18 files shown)


TheOneRing:
At the moment we still rely on Qt 5.12 so...

TheOneRing:
But as we don't plan to quit business we certainly plan to switch to Qt6 some day

test/testcredentialmanager.cpp
```diff
@@ -81,7 +81,7 @@ private Q_SLOTS:
                 });
             });
         });
-        QTest::qWaitFor([this] { return _finished; });
+        QVERIFY(QTest::qWaitFor([this] { return _finished; }));
```
alexeylysenko, line 84: @TheOneRing could you please comment on that? I expect the test execution has to wait until job will be finished, right? Or I missed something...

src/gui/generalsettings.cpp
```diff
@@ -302,15 +304,15 @@ void GeneralSettings::slotToggleOptionalDesktopNotifications(bool enable)
     cfgFile.setOptionalDesktopNotifications(enable);
 }
 
+#ifdef Q_OS_WIN
 void GeneralSettings::slotShowInExplorerNavigationPane(bool checked)
```
TheOneRing, line 308: and in the header?

test/testcredentialmanager.cpp
TheOneRing: Yes the idea is to wait for the test to finish.
As we have the skip condition on linux the wait might time out here.

TheOneRing:
> 
> 
> Fixed the warning messages appeared during build except following:
> 
>     1. Q_DECL_DEPRECATED_X("Use uuid") QString id() const;
>        Not sure whether I can simply replace id() to uuid().
> 

Thats why some are still around :)
>     2. Qt deprecation warnings
>        If there is a plan to move to Qt6 it would be very useful to fix all warnings in Qt 5.15 to make a migration smoother - can provide a separate PR for that.

Issue #8766, opened 2021-06-21
[QA] Renamed folder never propagates to windows client, when one file was locked


jnweiger:
Seen while testing #8763 with 2.8.3.4442

* Connect a windows client to demo.owncloud.org using winVFS.
* Open Documents/Example.odt with LibreOffice writer (no changes needed, just leave it open)
* Use the server WebUI to rename the folder Documents to Dada.
* The client correctly shows an error:
![image](https://user-images.githubusercontent.com/1108546/122809716-c3806680-d2ce-11eb-8a24-afce2246bdac.png)
* Quit LibreOffice, wait.

```
06-21 11:25:17:576 [ info sync.accessmanager ]: 6 "PROPFIND" "https://demo.owncloud.org/remote.php/dav/files/demo/" has X-Request-ID "97a5f0d6-c429-445f-8535-934a05887a4d"
06-21 11:25:17:577 [ info sync.networkjob ]:    OCC::PropfindJob created for "https://demo.owncloud.org" + "/" "OCC::QuotaInfo"
06-21 11:25:18:063 [ info sync.networkjob.propfind ]:   PROPFIND of QUrl("https://demo.owncloud.org/remote.php/dav/files/demo/") FINISHED WITH STATUS "OK"
06-21 11:25:28:998 [ info sync.accessmanager ]: 6 "PROPFIND" "https://demo.owncloud.org/remote.php/dav/files/demo/" has X-Request-ID "8eca5506-09a3-4c9b-b64f-f36adb4c967b"
06-21 11:25:28:999 [ info sync.networkjob ]:    OCC::PropfindJob created for "https://demo.owncloud.org" + "/" "OCC::ConnectionValidator"
06-21 11:25:29:149 [ info sync.networkjob.propfind ]:   PROPFIND of QUrl("https://demo.owncloud.org/remote.php/dav/files/demo/") FINISHED WITH STATUS "OK"
06-21 11:25:47:267 [ info gui.folder ]: Trying to check "https://demo.owncloud.org/remote.php/dav/files/demo/" for changes via ETag check. (time since last sync: 42 s)
06-21 11:25:47:268 [ info sync.accessmanager ]: 6 "PROPFIND" "https://demo.owncloud.org/remote.php/dav/files/demo/" has X-Request-ID "2e9716b2-9abe-45d5-9522-d44b217ab1fc"
[...]

PR #8768, opened 2021-06-22
Check wehter file is locked before we move

TheOneRing:


TheOneRing:
In regards of https://github.com/owncloud/client/issues/8766 I need to try to reproduce it first

src/common/filesystembase.cpp
```diff
@@ -133,6 +133,11 @@ bool FileSystem::rename(const QString &originFileName,
     bool success = false;
     QString error;
 #ifdef Q_OS_WIN
+    if (FileSystem::isFileLocked(originFileName, FileSystem::LockMode::Exclusive)) {
+        *errorString = QCoreApplication::tr("FileSystem", "Can't rename %1, the file is currently in use").arg(originFileName);
```
dragotin, line 137: Attention: This is unchecked, `errorString` could be a nullptr. Other places use a dummy object in case `errorString` is nullptr. You should rather set the variable `error`

src/common/filesystembase.cpp
```diff
@@ -191,6 +196,11 @@ bool FileSystem::uncheckedRenameReplace(const QString &originFileName,
     }
 
 #else //Q_OS_WIN
+    if (FileSystem::isFileLocked(originFileName, FileSystem::LockMode::Exclusive)) {
+        *errorString = QCoreApplication::tr("FileSystem", "Can't rename %1, the file is currently in use").arg(originFileName);
```
dragotin, line 200: Same here, in this function `errorString` is used completely unchecked.

src/libsync/owncloudpropagator.cpp
```diff
@@ -691,15 +691,16 @@ bool OwncloudPropagator::createConflict(const SyncFileItemPtr &item,
...
     emit touchedFile(conflictFilePath);
 
+    // If the file is locked, we want to retry this sync when it
+    // becomes available again.
+    if (FileSystem::isFileLocked(fn, FileSystem::LockMode::Exclusive)) {
+        emit seenLockedFile(fn, FileSystem::LockMode::Exclusive);
+        if (error)
+            *error = tr("File %1 is currently in use").arg(fn);
+        return false;
+    }
```
dragotin, line 701: before this new block, `touchedFile()` is emitted. However, if the file is locked, we do not touch anything, so the emits should probably be moved below.

src/libsync/propagatedownload.cpp
```diff
@@ -940,20 +947,20 @@ void PropagateDownloadFile::downloadFinished()
             return;
         }
     }
+    // If the file is locked, we want to retry this sync when it
+    // becomes available again
+    if (FileSystem::isFileLocked(fn, FileSystem::LockMode::Exclusive)) {
+        emit propagator()->seenLockedFile(fn, FileSystem::LockMode::Exclusive);
+        done(SyncFileItem::SoftError, tr("%1 is currently in use").arg(fn));
+        return;
+    }
```
dragotin, line 956: I wonder who/what removes the tmp file of the download that is stopped here. But there are already other checks that interrupt the download, so that is probably handled elsewhere. Otherwise that would leave tmp files behind, no?

dragotin (changes requested):
Looks good.

src/common/filesystembase.cpp
TheOneRing: I added an assert, as the function required an error parameter from the start

src/libsync/propagatedownload.cpp
TheOneRing: I'd doubt it but that something for a different review :D

dragotin (approved):
nice.

PR #8772, opened 2021-06-23
Allow to override the server url with $OWNCLOUD_OVERRIDE_SERVER_URL

TheOneRing:
Depends on: https://github.com/owncloud/client/pull/8771

### DIFF
--- a/src/libsync/theme.cpp
+++ b/src/libsync/theme.cpp
@@ -14,6 +14,7 @@
 
 #include "theme.h"
 #include "config.h"
+#include "common/depreaction.h"
 #include "common/utility.h"
 #include "version.h"
 #include "configfile.h"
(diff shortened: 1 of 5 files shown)


TheOneRing:
@SamuAlfageme :)

src/libsync/theme.h
```diff
@@ -160,9 +160,16 @@ class OWNCLOUDSYNC_EXPORT Theme : public QObject
      * Setting a value here will pre-define the server url.
      *
      * The respective UI controls will be disabled
+     * Deprecated: Use overrideServerUrlV2 as it allows overrides
      */
+    Q_DECL_DEPRECATED_X("Use overrideServerUrlV2")
     virtual QString overrideServerUrl() const;
```
fmoc, line 166: What's the incentive to keep this function rather than changing it directly?

src/libsync/theme.cpp
```diff
@@ -318,6 +319,17 @@ QString Theme::overrideServerUrl() const
     return QString();
 }
 
+QString Theme::overrideServerUrlV2() const
+{
+    static auto serverOverride = qEnvironmentVariable("OWNCLOUD_OVERRIDE_SERVER_URL");
```
fmoc, line 324: Can be `const`.

src/libsync/theme.h
TheOneRing: The reason is that the auto generated themes define this function...

fmoc: I see.

I suppose the `V2` suffix is the reason for this confusion. The name should give an idea what the method does. This one is just a shallow wrapper that allows overrides via an environment variable, so I guess that fact should somehow end up in the method name.

PR #8774, opened 2021-06-23
CI: Run gui tests on Debug builds to enable asserts and backrtraces

TheOneRing:


.drone.star
```diff
@@ -227,7 +227,7 @@ def gui_tests(ctx, trigger = {}, depends_on = [], filterTags = [], version = "da
                  setupServerAndApp(2) +
                  fixPermissions() +
                  owncloudLog() +
-                 build_client(ctx, "gcc", "g++", "Release", "Unix Makefiles", "make", build_dir) +
+                 build_client(ctx, "gcc", "g++", "Debug", "Ninja", "ninja", build_dir) +
```
fmoc, line 230: Also changes to the ninja build system, which I'm not sure is installed in the containers. Changing only to debug build mode should work, though.

individual-it (approved):

Issue #7589, opened 2019-11-13
Do Discovery and Propagation in parallel.


ogoffart:
Right now, the sync algorithm first do a discovery step to see what changes over the whole sync folder, then propagates files.
As a result, it might take a very long time before even starting to download or upload the first file, and restarting the sync has to do the discovery step from scratch again.
In order to solve the problem, we could do the propagation of downloading / uploading files, as long as we've discovered that folder.

## How:

There might be different approach. But I guess the easier would be to merge the  discovery's `ProcessDirectoryJob` and the `PropagateDirectory` into a single operation. I believe this might be the easier.
They could be kept separate and still run in parallel, but then it might be harder to synchronize.
Note that in any case, all the removes would still be done at the end.
Currently, only the removal of directory are done at the end, but under the new approach, we would also have to perform the file removal at the end.

## Problems:

Some features will stop working. Notably:

- The progress indication: It will be impossible to give a good progress of the sync since we have no idea how much is to be done in total.

## Non-problems

Something that will **not** be a problem is the detection of moves/rename. We do not need to have the whole sync tree in memory for that. The new discovery algorithm already does move detection quite well without that.

## Alternative

We may choose not to do that at all and cache the result of the discovery in a new table: https://github.com/owncloud/client/issues/2976
This would solve the problem that restarting the first sync before the discovery is finished restart from zero.
This will however not solve the memory usage problem which will be solved by the first approach, and it will still (i think) be slower that the paralelization.

mmattel:
> The progress indication

Is it possible to dynamically not only change the current value but also change the target value (means the 100% reference value)? If this is possible, then you have a full dynamic progressbar, feeded by two dynamic streams. The current processed items and the taget number which changes based on parallel running gathering. If described, this would not be confusing.

ogoffart:
> Is it possible to dynamically not only change the current value but also change the target value 

Yes, that's possible, but that means that the progress might actually go backward.
(Start at 0%,  then reach 50% then back to 25% then 85% then 10% then 90% and finally 100%,  as files are discovered and transferred)

mmattel:
What about to write during syncing something like x of y items to process?
Both values will be dynamically written.
![image](https://user-images.githubusercontent.com/3321281/68785652-cc16b280-063e-11ea-872c-84c965533f02.png)

michaelstingl:
> Yes, that's possible, but that means that the progress might actually go backward.
[...]

Issue #3932, opened 2015-10-08
Do not show disconnected icon when `Retry-After` is sent with a 503


butonic:
### Expected behaviour

When the server responds with a `503 Service Unavailable` message that contains a `Retry-After: 120` header the client should silently wait 120 seconds before trying the same request again.
### Actual behaviour

On a 503 the client changes the icon to represent a disconnected state. The Retry-After header is ignored.
### Motivation

I am looking into integration with backup systems that will move old files off to a slower medium. The file may be available only after a few minutes. I would like to tell the client to just try again after a while without showing the disconnected icon to users, making them worry something is broken when actually miracles happen in the background to make them happy.

dragotin:
@butonic are you able to create a client logfile of this case? That would help.

ogoffart:
This would be a reply to what kind of command?  PROPFIND? GET? PUT? something else?

butonic:
@ogoffart a GET request for a single file. Nothing else.

butonic:
@dragotin is this the relevant part?
```

10-08 17:18:44:063 0x1e3a1e0 Sync state changed for folder  "ownCloud" :  "Sync Running" 
10-08 17:18:44:063 0x1e3a1e0 virtual void OCC::PropagateDownloadFileQNAM::start() "Neue Textdatei.txt" 0 
10-08 17:18:44:063 0x1e3a1e0 "INSERT OR REPLACE INTO downloadinfo (path, tmpfile, etag, errorcount) VALUES ( ?1 , ?2, ?3, ?4 )" "Neue Textdatei.txt" ".Neue Textdatei.txt.~7f56db30" "561689194b57b" 0 
10-08 17:18:44:064 0x1e3a1e0 void OCC::SyncJournalDb::commitInternal(const QString&, bool) Transaction commit  "download file start" and starting new transaction 
10-08 17:18:44:064 0x1e3a1e0 virtual void OCC::GETFileJob::start() OCC::BandwidthManager(0x2faa838) false false 
10-08 17:18:44:064 0x1e3a1e0 !!! OCC::GETFileJob created for "http://localhost/core-stable81" + "/Neue Textdatei.txt" 
10-08 17:18:44:069 0x1e3a1e0 * Discarded as is hidden! "/home/jfd/sync/stable81/.Neue Textdatei.txt.~7f56db30" 
[...]

Issue #2976, opened 2015-03-17
[Brainstorming] Discovery cache


guruz:
What about caching discovery results in the journal?

Normally this should not be needed as we're always syncing all files/dirs and storing the proper ETag in the journal.
However in practice I think there's often issues like fatal sync errors, network timeouts etc which then on the next sync mean a complete re-discovery..

Something to chat about next week.

@ogoffart @dragotin @ckamm

MorrisJobke:
Currently we noticed some problems with randomly broken connections to an LDAP server. So the user backend can't properly hand back the home directory. To avoid random guessing what the correct home directory could be we hand back an exception (owncloud/core#15332). In this case the sync client also could see such responses (that will cause a HTTP 500). If this happens during discovery the whole discovery process is aborted and starts from scratch.(and maybe never finished because of this) Is it possible to just repeat the last (failed) request (because it is likely that the error was just temporarily caused) and then resume the discovery from that point? We maybe should improve the error message that is handed back to say "this was just a temporary error and could be fixed with the next request". (503 with Retry-After header)

That would make the client a bit more robust for these cases.

guruz:
@ogoffart @ckamm @jturcotte @PVince81 @DeepDiver1975 Now that we're thinking of doing the [recursive/infinity PROPFIND](https://github.com/owncloud/core/issues/14531) I'm wondering if it makes sense to think about this again.

The infinity PROPFIND **could** be implemented by having it fill a local cache. 
map (remote_dirName+ETag)->contents 

The issue was originaly about any discovery issue. It would also help with flaky network connections, basically everytime when the discovery would abort.

This should be implemented in a way to not fill so much memory again. (e.g. sqlite3?)

ogoffart:
that cache already exists.  It's called the database (sync journal)

guruz:
Please read the bug above. This is about failed syncs and people complaining about the long repetive re-discovery

guruz:
...which means it could be flushed after a successful sync.

guruz:
Similar how it was implemented before: https://github.com/owncloud/client/commit/980c1760735a5d780558d9938c91199cb320b581

dragotin:
@ogoffart 💟
@guruz isn't it true that we only do real propfinds into a tree if the ETag of the directory has changed, and read from the client journal instead if not?

IIRC what we discussed long time ago was to keep the entire tree in memory to not have to rebuild the whole tree again, maybe that should be discussed again.

[...]

PR #8777, opened 2021-06-24
Poll interval from capabilities

dragotin:
With this, admins can configure the remote poll interval via the capabilities of the server.

```
% curl -u admin:admin 'https://demo.owncloud.com/ocs/v1.php/cloud/capabilities?format=json' \
| jq | grep -i -e  poll
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1954  100  1954    0     0   4897      0 --:--:-- --:--:-- --:--:--  4897
          "pollinterval": 60,
```

If the client has no custom interval in the [client config file](https://doc.owncloud.com/desktop/advanced_usage/configuration_file.html) and if the capability is different than the default, the remote poll interval is changed. 

Note: To use 60 seconds, the capability needs to contain the string "60!" for the reason that 60 was the default in the capas, but using that would change the behaviour of all clients which needs to be avoided.

### DIFF
--- a/changelog/unreleased/8780
+++ b/changelog/unreleased/8780
@@ -3,6 +3,7 @@ Enhancement: Consider a remote poll interval coming with the server capabilities
 This way, admins can configure the remote sync poll interval of clients through
 the capabilities settings of the server. Note that the setting in the server
 capabilities needs to be done in milliseconds. Default is 30 seconds.
+
 https://github.com/owncloud/client/issues/5947
 https://github.com/owncloud/client/issues/8780
 https://github.com/owncloud/client/pull/8777
--- a/changelog/unreleased/8780
+++ b/changelog/unreleased/8780
@@ -1,4 +1,4 @@
-Enhancement: Consider a remote poll interval coming with the server capabilities.
+Enhancement: Consider a remote poll interval coming with the server capabilities
 
 This way, admins can configure the remote sync poll interval of clients through
 the capabilities settings of the server. Note that the setting in the server
--- a/changelog/unreleased/8780
+++ b/changelog/unreleased/8780
@@ -1,0 +1,8 @@
+Enhancement: Consider a remote poll interval coming with the server capabilities.
+
+This way, admins can configure the remote sync poll interval of clients through
+the capabilities settings of the server. Note that the setting in the server
+capabilities needs to be done in milliseconds. Default is 30 seconds.
+https://github.com/owncloud/client/issues/5947
+https://github.com/owncloud/client/issues/8780
+https://github.com/owncloud/client/pull/8777
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -308,8 +308,8 @@ bool Folder::dueToSync() const
     } else {
         // read the account capability - specified in seconds. If there, use it.
         int pta = accountState()->account()->capabilities().remotePollInterval();
-        if (pta > 5) {
-            polltime = std::chrono::milliseconds(pta * 1000);
+        if (pta > 4999) {
+            polltime = std::chrono::milliseconds(pta);
         }
     }
     if (msecSinceLastSync() < polltime) {
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -295,24 +295,13 @@ bool Folder::dueToSync() const
 
     ConfigFile cfg;
     // the default poll time of 30 seconds as it had been in the client forever.
-    auto polltime = cfg.DefaultRemotePollInterval;
+    // Now with https://github.com/owncloud/client/pull/8777 also the server capabilities are considered.
+    const auto pta = std::chrono::milliseconds(accountState()->account()->capabilities().remotePollInterval());
+    const auto polltime = cfg.remotePollInterval(pta);
 
-    // ... and an user configured poll interval that might exist
-    auto polltimeCfg = cfg.remotePollInterval();
-
-    if (polltimeCfg != polltime) {
-        // there is a non default entry in the config. Highest prio
-        polltime = polltimeCfg;
-    } else {
-        // read the account capability - specified in seconds. If there, use it.
-        int pta = accountState()->account()->capabilities().remotePollInterval();
-        if (pta > 4999) {
-            polltime = std::chrono::milliseconds(pta);
-        }
-    }
     // we add half a second here as estimated duration of the last etag job. That way the wished duration
     // is met more accurate - which appears to look better in the access log.
-    auto timeSinceLastSync = std::chrono::milliseconds(500+_timeSinceLastEtagCheckDone.elapsed());
+    const auto timeSinceLastSync = std::chrono::milliseconds(500+_timeSinceLastEtagCheckDone.elapsed());
     qCInfo(lcFolder) << "dueToSync:" << alias() << timeSinceLastSync.count() << " < " << polltime.count();
     if (timeSinceLastSync >= polltime) {
         return true;
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -296,7 +296,7 @@ bool Folder::dueToSync() const
     ConfigFile cfg;
     // the default poll time of 30 seconds as it had been in the client forever.
     // Now with https://github.com/owncloud/client/pull/8777 also the server capabilities are considered.
-    const auto pta = std::chrono::milliseconds(accountState()->account()->capabilities().remotePollInterval());
+    const auto pta = accountState()->account()->capabilities().remotePollInterval();
     const auto polltime = cfg.remotePollInterval(pta);
 
     const auto timeSinceLastSync = std::chrono::milliseconds(_timeSinceLastEtagCheckDone.elapsed());
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -288,10 +288,7 @@ bool Folder::canSync() const
 bool Folder::dueToSync() const
 {
     // conditions taken from previous folderman implementation
-    if (isSyncRunning() ||
-            etagJob()   ||
-            isBusy()    ||
-            !canSync()) {
+    if (isSyncRunning() || etagJob() || isBusy() || !canSync()) {
         return false;
     }
 
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -299,9 +299,7 @@ bool Folder::dueToSync() const
     const auto pta = std::chrono::milliseconds(accountState()->account()->capabilities().remotePollInterval());
     const auto polltime = cfg.remotePollInterval(pta);
 
-    // we add half a second here as estimated duration of the last etag job. That way the wished duration
-    // is met more accurate - which appears to look better in the access log.
-    const auto timeSinceLastSync = std::chrono::milliseconds(500 + _timeSinceLastEtagCheckDone.elapsed());
+    const auto timeSinceLastSync = std::chrono::milliseconds(_timeSinceLastEtagCheckDone.elapsed());
     qCInfo(lcFolder) << "dueToSync:" << alias() << timeSinceLastSync.count() << " < " << polltime.count();
     if (timeSinceLastSync >= polltime) {
         return true;
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -285,6 +285,39 @@ bool Folder::canSync() const
     return !syncPaused() && accountState()->isConnected();
 }
 
+bool Folder::dueToSync() const
+{
+    // conditions taken from previous folderman implementation
+    if (isSyncRunning() ||
+            etagJob()   ||
+            isBusy()    ||
+            !canSync()) {
+        return false;
+    }
+
+    ConfigFile cfg;
+    // the default poll time of 30 seconds as it had been in the client forever.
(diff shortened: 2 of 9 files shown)


src/gui/folderman.cpp
```diff
@@ -797,29 +797,19 @@ void FolderMan::slotStartScheduledFolderSync()
...
         if (_disabledFolders.contains(f)) {
             continue;
         }
-        if (f->etagJob() || f->isBusy() || !f->canSync()) {
-            continue;
-        }
-        if (f->msecSinceLastSync() < polltime) {
-            continue;
+        if (f->dueToSync()) {
+            QMetaObject::invokeMethod(f, "slotRunEtagJob", Qt::QueuedConnection);
```
TheOneRing, line 811: Please use `&Folder::slotRunEtagJob`

src/libsync/capabilities.cpp
```diff
@@ -103,6 +103,26 @@ int Capabilities::defaultPermissions() const
...
 
+int Capabilities::remotePollInterval() const
+{
+    // The default capability is 60, but the clients use 30. If we returned 60 here, the clients
+    // would all jump to 60 seconds, which would be an unepected change from their default.
+    // We only return 60 if the capability was set to 60! to indicate that this is really wanted.
+    // Otherwise 60 is treatet as default.
+    int interval {-1};
+    const QString val = _capabilities.value(QStringLiteral("core")).toMap().value(QStringLiteral("pollinterval")).toString();
+    if (val == QStringLiteral("60!")) {
```
TheOneRing, line 114: Michaels suggestion sounded good :)

michaelstingl:
> Note: To use 60 seconds, the capability needs to contain the string "60!" for the reason that 60 was the default in the capas, but using that would change the behaviour of all clients which needs to be avoided.

Client config file uses milliseconds (default 30000). Better use milliseconds here too. My proposal would be, to ignore server parameters < 100.

michaelstingl:
> With this, admins can configure the remote poll interval via the capabilities of the server.

Needs docs issue.

michaelstingl:
What are the weights of the different config ways? Who will win?

config file > server capability > client default (30000)

felix-schwarz:
A few thoughts:
1) switching to milliseconds could effectively break old/other clients that interpret this value as seconds and use it. Not sure if there are any, though.
2) `60!` would mean a type change for `core.pollinterval` in JSON (think `capabilities?format=json`), from number to string, at least for this case.
3) If we are to change the type from number to string anyway, maybe we could also add a string value (i.e. `client`) to signal to clients they should use their default values.
4) making this a new entry in capabilities rather than "reforming" the old one is probably the most compatible / cleanest approach. Unless, of course, this setting also needs to be adjustable in already released server versions.

michaelstingl:
> 1. Not sure if there are any, though.

We assume there are none.

> 3\. If we are to change the type from number to string anyway, maybe we could also add a string value (i.e. `client`) to signal to clients they should use their default values.

Or use different capabilities for different clients?

- `pollinterval_desktop`
- `pollinterval_ios`

> 4\. making this a new entry in capabilities rather than "reforming" the old one is probably the most compatible / cleaner approach. Unless, of course, this setting also needs to be adjustable in already released server versions.

See proposal for  3\.

felix-schwarz:
For different intervals for different clients, I'd suggest a new `pollinterval**s**` capability, which could provide default and specific values like this:

```json
"pollintervals" : {
	"default" : 60000,
	"ios" : 10000,
	"android" : 10000,
	"desktop" : 30000
}
```

I don't know what poll interval the Android client uses right now, but if it is 10 seconds, too, `ios` and `android` could also be simplified as `mobile`.

src/gui/folder.cpp
```diff
@@ -282,6 +282,39 @@ bool Folder::canSync() const
...
+            !canSync()) {
+        return false;
+    }
+
+    ConfigFile cfg;
+    // the default poll time of 30 seconds as it had been in the client forever.
+    auto polltime = std::chrono::milliseconds(cfg.DefaultRemotePollInterval);
+
+    // ... and an user configured poll interval that might exist
+    auto polltimeCfg = cfg.remotePollInterval();
```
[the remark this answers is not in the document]
dragotin: There is no DEFAULT_REMOTE_POLL_INTERVAL any more, I removed the define. A wise man told me that "in modern C++ we don't use defines" ;-)

dragotin:
Nice suggestions to change the capabilities to have one for each client, but not in scope of this PR which is about using what is already there.
About the _60!_ - I agree, it is not nice. The easiest would be to just ignore the setting "60" - which would make it impossible to set exactly 60 seconds, and admins would have to use 59 or 61. How's that?

michaelstingl:
> just ignore the setting "60" - which would make it impossible to set exactly 60 seconds, and admins would have to use 59 or 61.

Same quirk as 60! I'd say

src/gui/folderman.cpp
dragotin: done

src/libsync/capabilities.cpp
dragotin: also done. Values under 5000 are not changing the default now. 5 seconds is the minimum sync interval since ever.

src/libsync/configfile.h
```diff
@@ -42,6 +42,7 @@ class OWNCLOUDSYNC_EXPORT ConfigFile
     static QString configPath();
     static QString configFile();
     static bool exists();
+    const int DefaultRemotePollInterval { 30000 }; // default remote poll time in milliseconds
```
alexeylysenko, line 45: Would it make sense to make 
`const std::chrono::milliseconds DefaultRemotePollInterval{30000};` instead of int?
There are 2 places where DefaultRemotePollInterval is used and both of them create std::chrono::milliseconds instance.

NannaBarz:
[Link to Portal Case](https://portal.owncloud.com/apps/customer_portal/staff#/cases/5003r00001PBj5U)

butonic:
Can we use the proper [Retry-After](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) HTTP header for this? Several reasons:
* The poll interval should be dynamically adjusted by the server, depending on load, client type (mobile clients should poll less often than desktop clients), user (premium users might be allowed to poll more often ...). A single property in the capabilities is not the right place to do that. Clients should take into account the Retry-After header and at least wait for that period of time, if not longer.
* The header can be used to prevent a sync DoS by clients when they come online in the morning by varying the poll duration.
* How often the client can poll the server actually depends on the folder that is polled.
* A retry-after with a 429 (Too many requests) is the proper way to respond to the client when https://github.com/owncloud/client/issues/7589 creates too many requests. Also dynamically, depending on server load and other factors.
* When we implement push notifications a single change could trigger thousands of clients to start a sync run. Then we also need to be able to dynamically tell some cclients to come back later.
* The server should be able to dynamically back off (properly behaving) clients to reduce overall load. To protect itself.
* I actually would like to get rid of that an additional capabilities call ... so please don't add more things to the capabilities

Related:
https://github.com/owncloud/client/issues/3932
https://github.com/owncloud/client/issues/2976
https://github.com/owncloud/product/issues/21

dragotin:
Usually this header comes along with the error message 503 and 429. It does not have to, however, the client does a normal PROPFIND - and your proposal would as a result mean that all PROPFINDs return a Retry-After header. Nothing I would fancy.

Some other comments:
- Currently the server has no code to adopt the header according to load etc as far as I know.
- About request Dos: The clients care for this already today, not starting at the same time. 

I agree that it might be a nice feature that the server can influence the clients behaviour in this regards. Given that this is not there today, I like to leave that completely out of scope of this PR, because this uses
- a capability that is already here
- gives existing installations the chance to do something **at all** , which is an improvement immediately.

Btw, there is no additional capabilities call because of this. The clients read the capabilities anyways already today, and the capability is already there.

So I suggest to move your proposal into an enhancement. My personal opinion is that the  effort does not really pay off because the clients are first class citizens that need to be served *fast*. Adding for users unpredictable delays in detection of changes is just raising confusion IMO.

src/gui/folder.cpp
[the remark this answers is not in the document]
TheOneRing: But why do you export it :)
The function already handles the DefaultRemotePollInterval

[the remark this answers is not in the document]
dragotin: Yes, but I need to detect if the local configured time is different from the default time, because that would have precedence even over the time coming in from the capabilities. That is why I need it.

src/libsync/configfile.h
dragotin: Thanks, done that!

src/gui/folder.cpp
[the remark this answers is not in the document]
TheOneRing: Please pass `capabilities().remotePollInterval()` as a default to `cfg.remotePollInterval(capabilities().remotePollInterval())`
That way we can handle the default case directly in the config without the need to duplicate the minimal interval etc

src/libsync/configfile.h
```diff
@@ -42,7 +42,7 @@ class OWNCLOUDSYNC_EXPORT ConfigFile
     static QString configPath();
     static QString configFile();
     static bool exists();
-
+    const std::chrono::milliseconds DefaultRemotePollInterval { 30000 }; // default remote poll time in milliseconds
```
TheOneRing, line 45: Now there is really no need to make the default public available :D

Also I guess we should use seconds more often instead of milliseconds

TheOneRing (changes requested):

src/libsync/configfile.h
```diff
@@ -209,6 +209,7 @@ class OWNCLOUDSYNC_EXPORT ConfigFile
 
 private:
     typedef QSharedPointer<AbstractCredentials> SharedCreds;
+    const std::chrono::seconds DefaultRemotePollInterval { 30 }; // default remote poll time in milliseconds
```
TheOneRing, line 212: I'd kept that in the .cpp but I guess I won't be able to motivate you :D

src/libsync/configfile.cpp
```diff
@@ -433,10 +432,17 @@ chrono::milliseconds ConfigFile::remotePollInterval(const QString &connection) c
     QSettings settings(configFile(), QSettings::IniFormat);
     settings.beginGroup(con);
 
-    auto defaultPollInterval = chrono::milliseconds(DEFAULT_REMOTE_POLL_INTERVAL);
+    auto defaultPollInterval { DefaultRemotePollInterval };
```
TheOneRing, line 435: no need for that var

src/libsync/capabilities.h
```diff
@@ -75,6 +75,12 @@ class OWNCLOUDSYNC_EXPORT Capabilities
     bool sharePublicLinkEnforceExpireDate() const;
     bool sharePublicLinkMultiple() const;
     bool shareResharing() const;
+    /** Remote Poll interval.
+     *
+     *  returns the requested poll interval to be used by the client in milliseconds.
```
TheOneRing, line 80: Comment outdated

src/libsync/capabilities.cpp
```diff
@@ -103,6 +103,20 @@ int Capabilities::defaultPermissions() const
     return _fileSharingCapabilities.value(QStringLiteral("default_permissions"), 1).toInt();
 }
 
+std::chrono::seconds Capabilities::remotePollInterval() const
+{
+    // The default of the capability is 60, but the clients use 30.
+    // Values below 5000 are not allowed and fall back to the default.
+    std::chrono::seconds interval(0);
```
TheOneRing, line 110: Where do we check for the 0 seconds interval?

src/libsync/configfile.cpp
dragotin: Hm, please check the code again. I think it is.

src/libsync/configfile.h
dragotin: Sure u a.

src/libsync/capabilities.cpp
dragotin: We don't explicitely but it is implicitely done by checking for `< seconds(5)`.

src/libsync/configfile.cpp
TheOneRing: ups

src/libsync/capabilities.cpp
TheOneRing: Ok I mean in that case you can skip the check for 4999 and ok here and just return std::milliseconds(tmpInt)

changelog/unreleased/8780
```diff
@@ -0,0 +1,9 @@
+Enhancement: Consider a remote poll interval coming with the server capabilities
+
+This way, admins can configure the remote sync poll interval of clients through
+the capabilities settings of the server. Note that the setting in the server
+capabilities needs to be done in milliseconds. Default is 30 seconds.
```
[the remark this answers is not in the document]
mmattel: To make the text more clear I propose the following:

The default client value is 30 seconds, if there is no change in the server sided default polling interval settings of 60 seconds. Any change on the server sided config polling interval setting, which must be made in milliseconds and has a higher value than 5000, is treated as a new client polling interval value.

PR #8779, opened 2021-06-24
Improve handling of Folder removes

TheOneRing:
As Folder objects where directly removed without any heads up the handling of removed Folders was quite complicated.
The change  allow to directly work on a folder object instead of needing to look up the folder by its path every time it is used and then checking whether the object is still valid.

### DIFF
--- a/src/gui/folderman.cpp
+++ b/src/gui/folderman.cpp
@@ -95,21 +95,14 @@ FolderMan::~FolderMan()
     _instance = nullptr;
 }
 
-const OCC::Folder::Map &FolderMan::map() const
+const QMap<QString, Folder *> &FolderMan::map() const
 {
     return _folderMap;
 }
 
-QList<Folder *> FolderMan::list() const
-{
-    return _folderMap.values();
-}
-
 void FolderMan::unloadFolder(Folder *f)
 {
-    if (!f) {
-        return;
-    }
+    Q_ASSERT(f);
 
     _socketApi->slotUnregisterPath(f->alias());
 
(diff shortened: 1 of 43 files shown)


src/gui/folderman.cpp
```diff
@@ -95,21 +95,14 @@ FolderMan::~FolderMan()
...
-{
-    return _folderMap.values();
-}
-
 void FolderMan::unloadFolder(Folder *f)
 {
-    if (!f) {
-        return;
-    }
+    Q_ASSERT(f);
```
dschmidt, line 105: This is less defensive than the old code. 
The new one asserts in debug builds which is good, but it could crash *somewhere* in production

TheOneRing: As I don't use the old unload anymore this should be fixed.

erikjv (approved):
LGTM, but I'm not sure if I have the full context.

src/gui/issueswidget.cpp
```diff
@@ -104,6 +102,15 @@ IssuesWidget::IssuesWidget(QWidget *parent)
     _ui->_conflictHelp->setText(
         tr("There were conflicts. <a href=\"%1\">Check the documentation on how to resolve them.</a>")
             .arg(Theme::instance()->conflictHelpUrl()));
+
+    connect(FolderMan::instance(), &FolderMan::folderRemoved, this, [this](Folder *f) {
+        _model->remove_if([f](const ProtocolItem &item) {
+            if (item.folder() == f) {
```
alexeylysenko, line 108: Can be a one line:
`return item.folder() == f`

src/gui/folderman.cpp
```diff
@@ -129,28 +122,19 @@ void FolderMan::unloadFolder(Folder *f)
...
-
     // clear the list of existing folders.
-    Folder::MapIterator i(_folderMap);
-    while (i.hasNext()) {
-        i.next();
-        Folder *f = i.value();
-        unloadFolder(f);
-        delete f;
-        cnt++;
+    const auto folders = std::move(_folderMap);
```
alexeylysenko, line 128: Why don't you use qAsConst that clearly shows your intention?

src/gui/protocolwidget.cpp
```diff
@@ -64,8 +64,14 @@ ProtocolWidget::ProtocolWidget(QWidget *parent)
         showHeaderContextMenu(header, _sortModel);
     });
 
-
-    _ui->_headerLabel->setText(tr("Local sync protocol"));
+    connect(FolderMan::instance(), &FolderMan::folderRemoved, this, [this](Folder *f) {
+        _model->remove_if([f](const ProtocolItem &item) {
+            if (item.folder() == f) {
```
alexeylysenko, line 69: The same as here https://github.com/owncloud/client/pull/8779/files#r668087565

src/gui/folderman.cpp
TheOneRing: I clear the _folderMap with that call and unregister it.
The old code remove the folder in each iteration and made sure the list was empty afterwards...

src/gui/issueswidget.cpp
TheOneRing: indeed

PR #8782, opened 2021-06-25
Fix case issue in folder management

TheOneRing:


src/common/utility.cpp
```diff
@@ -263,10 +263,12 @@ void Utility::usleep(int usec)
 }
 
 // This can be overriden from the tests
-OCSYNC_EXPORT bool fsCasePreserving_override = []()-> bool {
-    QByteArray env = qgetenv("OWNCLOUD_TEST_CASE_PRESERVING");
-    if (!env.isEmpty())
-        return env.toInt();
+OCSYNC_EXPORT bool fsCasePreserving_override = []() -> bool {
+    static bool ok = false;
+    static int env = qEnvironmentVariableIntValue("OWNCLOUD_TEST_CASE_PRESERVING", &ok);
```
fmoc, line 268: A lot of implicit casting between `int` and `bool` is going on here. How about making the comparisons explicit, so it's easy to understand what's going on here?

PR #8783, opened 2021-06-25
Replace string comparison with FileSystem utils

TheOneRing:
Needs more work,,,

### DIFF
--- a/src/common/filesystembase.cpp
+++ b/src/common/filesystembase.cpp
@@ -535,13 +535,15 @@ QString FileSystem::pathtoUNC(const QString &_str)
 }
 #endif
 
-bool FileSystem::isChildPathOf(QStringView child, QStringView parent)
+bool FileSystem::isChildPathOf(const QString &child, const QString &parent)
 {
-    static const auto Casing = Utility::fsCasePreserving() ? Qt::CaseInsensitive : Qt::CaseSensitive;
-    // ignore additional / in the assert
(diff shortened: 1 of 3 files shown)


src/common/utility.cpp
```diff
@@ -277,15 +279,18 @@ bool Utility::fsCasePreserving()
...
 bool Utility::fileNamesEqual(const QString &fn1, const QString &fn2)
 {
-    const QDir fd1(fn1);
-    const QDir fd2(fn2);
-
+    const QFileInfo fd1(fn1);
+    const QFileInfo fd2(fn2);
+    qDebug() << fd1.canonicalFilePath() << fd2.canonicalFilePath();
     // Attention: If the path does not exist, canonicalPath returns ""
     // ONLY use this function with existing pathes.
```
fmoc, line 286: I think the comment should be reworded a bit, now that this is actually checked (which generally is a good idea; comments in the implementation aren't really read by users).

src/common/utility.cpp
```diff
@@ -277,15 +279,18 @@ bool Utility::fsCasePreserving()
...
+    Q_ASSERT(fd1.exists());
+    Q_ASSERT(fd2.exists());
+    Q_ASSERT(fd1.canonicalFilePath() == fn1);
```
fmoc, line 289: You are calculating these canonical paths below anyway.

TheOneRing: ITS A WIP
and the idea is that I can remove the cannonical path one day

TheOneRing: I just created it to fix finish it one day

PR #8786, opened 2021-06-28
Prevent implementations of Result<bool,..>

TheOneRing:


src/common/result.h
```diff
@@ -35,6 +35,7 @@ class Result
         : _result(std::move(value))
         , _isError(false)
     {
+        static_assert(!std::is_same<T, bool>::value, "Bool is not supported as this class overrides the bool operator");
```
dschmidt, line 38: I would probably prefer to not enable the ctor for types that will trigger this assert ... 
@TheOneRing said she prefers it this way because this triggers a more explicit error and might be easier to debug. 🤷🏻

dschmidt (approved):

dschmidt:
I hope we won't need to have a "tristate" result one day that needs error, true and false :-D

PR #8789, opened 2021-06-28
Improved string replacement in a more efficient way

alexeylysenko:


src/gui/updater/ocupdater.cpp
```diff
@@ -208,10 +208,10 @@ void OCUpdater::slotStartInstaller()
         };
 
         QString msiLogFile = ConfigFile::configPath() + "msi.log";
-        QString command = QString("&{msiexec /norestart /passive /i '%1' /L*V '%2'| Out-Null ; &'%3'}")
```
TheOneRing, line 211: Please don't use QString with auto.
We have enabled the QStringBuilder and together with auto this can lead to crashes.
As there is no concatenation here that doesn't matter but better try to stick to const QString

alexeylysenko: Done

TheOneRing (approved):

PR #8790, opened 2021-06-28
Use const iterators

alexeylysenko:
There were mixed const and non-const iterators.

It is generally good practice to use const_iterator on a non-const container as well, unless you need to change the one through the iterator. Const iterators are slightly faster, and can improve code readability.

### DIFF
--- a/src/libsync/propagatedownload.cpp
+++ b/src/libsync/propagatedownload.cpp
@@ -105,7 +105,7 @@ void GETFileJob::start()
     }
 
     QNetworkRequest req;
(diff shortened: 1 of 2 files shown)


TheOneRing:
Hi what's mixed here :D could you clarify the title?

alexeylysenko:
> Hi what's mixed here :D could you clarify the title?

I've updated PR description. Hope it's clear now.

TheOneRing:
Could you just update the commit message to say "Use const iterator"

alexeylysenko:
> Could you just update the commit message to say "Use const iterator"

Done

TheOneRing (approved):

PR #8792, opened 2021-06-29
[Tests-Only]Called cleanup in afterScenario

Talank:
Calling cleanup methods after the end of each scenario was missing in the client test. So, this PR fixes that.


swoichha (commented):
lgtm

Talank:
For some reason, the user `Brian` is not available in the list of created users. I think, it has been set to `none` at some point (in given steps that is executed via middleware).

Talank:
Created new issue so that it can be fixed some day. Closing this PR for now.

PR #8800, opened 2021-07-02
Fix assert in makeDavUrl

TheOneRing:


fmoc (approved):

fmoc (approved):
Indeed, this is the true fix for the issue.

Issue #8798, opened 2021-07-02
Return boolean in OC_ASSERT and handle the result


TheOneRing:
Our macro OC_ASSERT  behaves line a normal ASSERT in debug builds and only prints an error in production builds.
As the macro is used to validate a state it should return a boolean in production builds as without handling the condition the client will most probably die ungracefully otherwise.

dragotin:
I am not sure. If you "enhance" OC_ASSERT like that, maybe it is better to solve the situation in the concrete code directly and handle the critical situation. For me that sounds like too much for an assert. Not sure.

TheOneRing:
Yes but we use it in a lot of places that never should be reached. Those should be replaced by a real assert I guess as they should just never happen.

fmoc:
Perhaps you should then add a new construct rather than changing the meaning of a debug macro in order to not confuse future readers of the code. I could image adding a macro that just shows an error message and exits (for critical issues), for instance.

The existing `OC_*` macros should probably also log whenever they fail in production, maybe even on a warning/critical level.

TheOneRing:
Thats what the existing macro does, it logs the assert message as critical.
But in most cases we just continue to execute the code which absolutely does not make any sense.

fmoc:
If you really want to tackle that issue, you will have to change that macro into some sort of "crash-with-nice-message". But there is a risk that users, in special cases, will be bombarded with error messages. I'm not sure whether that's better from a UX point of view, even if it's desirable from a developer point of view. Another risk is that developers might tend not to use the macro at all any more then in order not to bug the users.

I'm all in for "crash-with-nice-message", but you can expect at least a week of either fixing the root cause or handling the asserted condition more gracefully. It will be worth the effort, though, because in the end it will result in a better product, and users might not end up in unexpected states any more without noticing, causing undefined behaviour of the client. You can then also distinguish better between true edge cases, "things that need to be handled properly", and (likely) some false positives.

I don't think that just returning the result of an asserted condition makes any sense, though. That would make it opt-in rather than opt-out. In that case, you can also just ignore the macro's existence and implement a normal `if` directly.

As a side note, in some cases I've encountered so far, the macro is used wrongly anyway. One should assert conditions that can be expected to never happen other than edge cases, causing one to not implement specific handling for these, but instead making sure that they will be detected. For that reason, I don't mind having asserts enabled in production builds. But I'm also into Design by Contract, which is heavily based on asserts.

TheOneRing:
[...]

PR #8802, opened 2021-07-02
OC_ENSURE

TheOneRing:
This is an alternative to OC_ASSERT which is most of the time used wrong.

See the discussion in https://github.com/owncloud/client/issues/8798

### DIFF
--- a/src/common/asserts.h
+++ b/src/common/asserts.h
@@ -44,6 +44,15 @@
     } else {                                                                                                                \
     }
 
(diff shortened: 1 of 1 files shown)


dragotin (approved):
Yes, well.. I have to admit that I do not find these kind of chains of complex C++ very intuitive. 

The benefit over 
```
if ( Q_UNLIKELY(!f)) {
   qDebug() << "Folder is not defined. Big problem!";
   return;
}
```
is not really obvious to me.

TheOneRing:
It's an ssert in debug builds, so devs and our squish tests are able to run into it.
Crashes get fixed warnings tend to be ignored.

PR #8803, opened 2021-07-05
Windows: Retry convertToPlaceholder for 30s on new dir

TheOneRing:


src/libsync/propagateremotemkdir.cpp
```diff
@@ -131,15 +134,57 @@ void PropagateRemoteMkdir::success()
     itemCopy._etag.clear();
 
     // save the file id already so we can detect rename or remove
```
dragotin, line 136: This comment does not really look useful.

src/libsync/propagateremotemkdir.cpp
```diff
@@ -131,15 +134,57 @@ void PropagateRemoteMkdir::success()
...
+                done(SyncFileItem::Success);
+            } else {
+                done(SyncFileItem::SoftError, result->error());
+            }
+        });
+
+        watcher->setFuture(QtConcurrent::run([p = propagator(), itemCopy] {
+            // Try to update the meta data with a 30s timeout
+            const auto start = std::chrono::steady_clock::now();
+            while ((std::chrono::steady_clock::now() - start) < UpdateMetaDataRetyTimeOut) {
```
dragotin, line 170: I like `QElapsedTimer` for these kind of things.

src/libsync/propagateremotemkdir.cpp
```diff
@@ -131,15 +134,57 @@ void PropagateRemoteMkdir::success()
...
+                QThread::sleep(1);
```
dragotin, line 171: Any reason why you sleep a second before trying the first time?

dragotin (approved):
Ok, minor comments. Except the sleep before trying the first time.

src/libsync/propagateremotemkdir.cpp
TheOneRing: Hm but then we have to either convert units or use std::chrono for the conversion and in the end we don't gain much  besides more involved components.

src/libsync/propagateremotemkdir.cpp
TheOneRing: The idea is that we really just tried that right before so no need to retry.

Issue #8804, opened 2021-07-05
`emit aboutToRemoveAllFiles` causes owncloudcmd to hang wo/ prompt


SamuAlfageme:
I was just hit by this after discovery:

```
07-05 12:24:44:910 [ info sync.networkjob.lscol ]:	LSCOL of QUrl("https://<acct>@<server>/remote.php/webdav/<path>/abc") FINISHED WITH STATUS "OK"
07-05 12:24:44:911 [ info sync.discovery ]:	Processing "abc/abc" | valid: true/true/false | mtime: 1625479485/1625479485/0 | size: 0/0/0 | etag: "28b7e:1625479485.077"//"" | checksum: ""//"" | perm: 0x8b1058//0x7f1cb9635998 | fileid: "ZW9zaG9tZS1pMDE6MTY2Nzgy"//"" | inode: 3301694746119/3301694746119/ | type: CSyncEnums::ItemTypeDirectory/CSyncEnums::ItemTypeDirectory/CSyncEnums::ItemTypeFile
07-05 12:24:44:911 [ info sync.discovery ]:	Discovered "abc/abc" CSyncEnums::CSYNC_INSTRUCTION_REMOVE OCC::SyncFileItem::Down CSyncEnums::ItemTypeDirectory
07-05 12:24:44:914 [ info sync.discovery ]:	STARTING "abc/abc" OCC::ProcessDirectoryJob::ParentDontExist "abc/abc" OCC::ProcessDirectoryJob::NormalQuery
07-05 12:24:44:922 [ info sync.engine ]:	#### Discovery end ####################################################  150 ms
07-05 12:24:44:924 [ info sync.engine ]:	All the files are going to be changed, asking the user
```

https://github.com/owncloud/client/blob/9117275c67ca311152b4f16749c6fd32c3d40fe9/src/libsync/syncengine.cpp#L754

... the command line client indefinitely hangs when this condition is met.

michaelstingl:
@SamuAlfageme run with `--non-interactive` as a workaround?

TheOneRing:
Its not handled

TheOneRing:
Fixed in https://github.com/owncloud/client/pull/8805

PR #8805, opened 2021-07-05
Cmd: Implement aboutToRemoveAllFiles

TheOneRing:
Fixes: #8804

### DIFF
--- a/src/cmd/cmd.cpp
+++ b/src/cmd/cmd.cpp
@@ -167,9 +167,9 @@ void sync(const SyncCTX &ctx)
             }
         }
     });
-    QObject::connect(engine, &SyncEngine::aboutToRemoveAllFiles, engine, [ctx](OCC::SyncFileItem::Direction dir, std::function<void(bool)> callback) {
+    QObject::connect(engine, &SyncEngine::aboutToRemoveAllFiles, engine, [ctx](OCC::SyncFileItem::Direction dir, std::function<void(bool)> abort) {
         if (!ctx.options.interactive) {
-            callback(false);
+            abort(false);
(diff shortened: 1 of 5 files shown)


src/cmd/cmd.cpp
```diff
@@ -167,11 +167,40 @@ void sync(const SyncCTX &ctx)
             }
         }
     });
+    QObject::connect(engine, &SyncEngine::aboutToRemoveAllFiles, engine, [ctx](OCC::SyncFileItem::Direction dir, std::function<void(bool)> callback) {
```
dschmidt, line 170: Can we rename `callback` to `cancel` or `cancelCallback` or `stop` or something like that?

My brain has a hard time parsing what true and false arguments mean with `callback`...

src/cmd/cmd.cpp
```diff
@@ -167,11 +167,40 @@ void sync(const SyncCTX &ctx)
...
+                                                    : "All the files in your local sync folder '%1' were deleted. These deletes will be "
+                                                      "synchronized with your server, making such files unavailable unless restored.\n"
+                                                      "Are you sure you want to sync those actions with the server?\n"
+                                                      "If this was an accident and you decide to keep your files, they will be re-synced from the server.")
+                      << std::endl;
+            std::string s;
+            while (true) {
+                std::cout << "Remove all files?[y,n]";
+                std::getline(std::cin, s);
+                if (s.empty() || s == "y") {
```
dschmidt, line 188: As we are talking about removing files, should we maybe not accept empty as `y`?

dschmidt (approved):

PR #8808, opened 2021-07-12
Only issue a soft error if we fail to convert to placeholder for 30s

TheOneRing:


src/libsync/propagateremotemkdir.cpp
```diff
@@ -158,11 +158,18 @@ void PropagateRemoteMkdir::success()
...
-                Q_ASSERT(result->get() == Vfs::ConvertToPlaceholderResult::Ok);
-                done(SyncFileItem::Success);
-            } else {
-                done(SyncFileItem::FatalError, result->error());
+                switch (result->get()) {
+                case Vfs::ConvertToPlaceholderResult::Ok:
+                    done(SyncFileItem::Success);
+                    return;
+                case Vfs::ConvertToPlaceholderResult::Locked:
+                    done(SyncFileItem::SoftError, tr("UpdateMetadata Timed out"));
```
dragotin, line 166: we could go for a more friendly string, such as "Set file status failed due to file lock." maybe? Is the error displayed together with the file name?

src/libsync/propagateremotemkdir.cpp
```diff
@@ -158,11 +158,18 @@ void PropagateRemoteMkdir::success()
         connect(watcher, &QFutureWatcherBase::finished, this, [watcher, this] {
             const auto *result = watcher->result().get();
             if (*result) {
```
dragotin, line 160: Check on `result != nullptr` ?

dragotin (approved):

src/libsync/propagateremotemkdir.cpp
TheOneRing: the connect is only triggered if I return something and I really don't intend to return nulll :D

src/libsync/propagateremotemkdir.cpp
TheOneRing: yes and I'll change the description

PR #8810, opened 2021-07-12
Replaces obsolete QLinkedList with std::list

alexeylysenko:


src/libsync/bandwidthmanager.cpp
```diff
@@ -375,18 +377,18 @@ void BandwidthManager::switchingTimerExpired()
 
 void BandwidthManager::absoluteLimitTimerExpired()
 {
-    if (usingAbsoluteUploadLimit() && _absoluteUploadDeviceList.count() > 0) {
-        qint64 quotaPerDevice = _currentUploadLimit / qMax(1, _absoluteUploadDeviceList.count());
-        qCDebug(lcBandwidthManager) << quotaPerDevice << _absoluteUploadDeviceList.count() << _currentUploadLimit;
+    if (usingAbsoluteUploadLimit() && !_absoluteUploadDeviceList.empty()) {
+        qint64 quotaPerDevice = _currentUploadLimit / qMax(size_t(1), _absoluteUploadDeviceList.size());
```
TheOneRing, line 381: qMax<size_t>?

src/libsync/bandwidthmanager.cpp
```diff
@@ -375,18 +377,18 @@ void BandwidthManager::switchingTimerExpired()
...
             device->giveBandwidthQuota(quotaPerDevice);
             qCDebug(lcBandwidthManager) << "Gave " << quotaPerDevice / 1024.0 << " kB to" << device;
         }
     }
-    if (usingAbsoluteDownloadLimit() && _downloadJobList.count() > 0) {
-        qint64 quotaPerJob = _currentDownloadLimit / qMax(1, _downloadJobList.count());
-        qCDebug(lcBandwidthManager) << quotaPerJob << _downloadJobList.count() << _currentDownloadLimit;
-        for (auto *j : qAsConst(_downloadJobList)) {
+    if (usingAbsoluteDownloadLimit() && !_downloadJobList.empty()) {
+        qint64 quotaPerJob = _currentDownloadLimit / qMax(size_t(1), _downloadJobList.size());
```
TheOneRing, line 389: same

src/libsync/bandwidthmanager.cpp
fmoc: It should use the largest size to infer type `T`, which should make it use the `size_t` returned by `size()`. https://en.cppreference.com/w/cpp/container/list/size

alexeylysenko: @TheOneRing I prefer the case when qMax tries to deduce the type on its own. It has one advantage from my pov: if you specify the template type `qMax<T>` would try to convert all arguments to T implicitly. That might hide some issues. For example, right now in master qMax deals with int. After my change the second parameter is no longer int, but size_t. `qMax<int>()` would compile just fine here, but `qMax()` won't. The build failure attracts attention and allows to double check comparison logic.

TheOneRing: Well you changed the type T yourself already with the c style cast to size_t.
So its basically `qMax<size_t>(1`vs`qMax(static_cast<size_t>(1)`.
If you insist on the automatic deduction, would it work out of the box if you exchange the two parameters and put the size_t parameter first?
https://code.woboq.org/qt5/qtbase/src/corelib/global/qglobal.h.html#_Z4qMaxRKT_S1_

Btw: `qMax<int>(static_cast<size_t>(1)` should still warn as the parameter is still cast implicitly to int.

erikjv: Why not remove the `qMax` completely? I mean, the `if` has `!_absoluteUploadDeviceList.empty()` meaning, it can never be less than 1. Or did I miss something?

src/libsync/bandwidthmanager.cpp
erikjv: Yes, same here: why use `qMax` when you know the list is not empty?

src/libsync/bandwidthmanager.cpp
alexeylysenko: @erikjv makes sense for me. I have removed qMax() at all.
@TheOneRing @fmoc what's your opinion?

TheOneRing: Looks good :)

TheOneRing (approved):

PR #8814, opened 2021-07-14
[tests-only] overwrite content of shared file

swoichha:
### Description
This PR adds gui test for overwrite content of a file shared by sharer

### Related Issue
https://github.com/owncloud/client/issues/8651

### DIFF
--- a/test/gui/shared/steps/steps.py
+++ b/test/gui/shared/steps/steps.py
@@ -254,6 +254,7 @@ def step(context):
         lambda: isFolderSynced(context.userData['clientSyncPath']),
         context.userData['clientSyncTimeout'] * 1000,
     )
+    snooze(20)
 
 
 @When('the user waits for file "|any|" to be synced')
@@ -724,3 +725,11 @@ def step(context, resource, group):
 
     sharingDialog = SharingDialog()
     sharingDialog.selectCollaborator(group, True)
+
+    
+@When('the user overwrites the file "|any|" with content "|any|"')
+def step(context, resource, content):
+    waitForFileToBeSynced(context, resource)
+    resource = context.userData['clientSyncPath'] + resource
+    with open(resource, 'wt') as file:
+        file.write(content)
--- a/test/gui/shared/steps/steps.py
+++ b/test/gui/shared/steps/steps.py
@@ -114,6 +114,7 @@ def step(context, username):
     enterUserPassword = EnterPassword()
     enterUserPassword.enterPassword(password)
 
+
 @Given('the user has started the client')
 def step(context):
     startClient(context)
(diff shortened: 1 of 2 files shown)


test/gui/shared/steps/steps.py
```diff
@@ -724,3 +735,19 @@ def step(context, resource, group):
 
     sharingDialog = SharingDialog()
     sharingDialog.selectCollaborator(group, True)
+
+
+@When('the user overwrites the file "|any|" with content "|any|"')
+def step(context, resource, content):
+    print("before overwrite")
```
Talank, line 742: remove the print

test/gui/shared/steps/steps.py
```diff
@@ -724,3 +735,19 @@ def step(context, resource, group):
...
+    waitForFileToBeSynced(context, resource)
+    waitForFolderToBeSynced(context, '/')
+
+    snooze(3)
+
+    f = open(context.userData['clientSyncPath'] + resource, "w")
+    f.write(content)
+    f.close()
+
+    print("after overwrite")
```
Talank, line 752: remove the print

test/gui/shared/steps/steps.py
```diff
@@ -724,3 +735,19 @@ def step(context, resource, group):
 
     sharingDialog = SharingDialog()
     sharingDialog.selectCollaborator(group, True)
+
+
+@When('the user overwrites the file "|any|" with content "|any|"')
+def step(context, resource, content):
+    print("before overwrite")
+    waitForFileToBeSynced(context, resource)
+    waitForFolderToBeSynced(context, '/')
```
Talank, line 744: why wait for folder `/` to be synced? 🤔

test/gui/shared/steps/steps.py
```diff
@@ -724,3 +735,19 @@ def step(context, resource, group):
...
+
+    snooze(3)
```
Talank, line 746: can you write a comment above the `snooze(3)` that explains why this static wait is necessary?

test/gui/shared/steps/steps.py
swoichha: this is to sync everything in root

test/gui/shared/steps/steps.py
swoichha: That's a good idea. I'll add a comment and push it.

test/gui/shared/steps/steps.py
swoichha: it is for the log purpose that would be visible in CI as well

test/gui/shared/steps/steps.py
individual-it: the problem is described in https://github.com/owncloud/client/issues/8832

It is an artificial delay set in the client. See also https://github.com/owncloud/client/blob/9117275c67ca311152b4f16749c6fd32c3d40fe9/src/libsync/syncengine.cpp#L61-L74

test/gui/shared/steps/steps.py
Talank: ok, then can you write a different message something like `starting file overwrite` and `file has been successfully overwritten`?

test/gui/shared/steps/steps.py
individual-it: Please mention #8832 in the comment, that gives more background

test/gui/shared/steps/steps.py
```diff
@@ -724,3 +735,23 @@ def step(context, resource, group):
...
+    # The client does not see the change although the changes have already been made thus we are having a race condition
+    # So for now we add 3sec static wait
+
+    snooze(3)
+
+    f = open(context.userData['clientSyncPath'] + resource, "w")
+    f.write(content)
+    f.close()
+
+    print("file has been overwritten")
```
individual-it, line 756: Do we need that debug line?

test/gui/shared/steps/steps.py
```diff
@@ -724,3 +735,23 @@ def step(context, resource, group):
 
     sharingDialog = SharingDialog()
     sharingDialog.selectCollaborator(group, True)
+
+
+@When('the user overwrites the file "|any|" with content "|any|"')
+def step(context, resource, content):
+    print("starting file overwrite")
```
individual-it, line 742: Is this debug line needed?

individual-it (changes requested):

test/gui/shared/steps/steps.py
swoichha: It's for log purpose in CI

individual-it (approved):

PR #8818, opened 2021-07-15
Clang-format: Sort header

TheOneRing:


.clang-format
```diff
@@ -61,3 +61,4 @@ ForEachMacros:   [ foreach, Q_FOREACH, BOOST_FOREACH, forever, Q_FOREVER, QBENCH
 MaxEmptyLinesToKeep: 2
 KeepEmptyLinesAtTheStartOfBlocks: false
 
+SortIncludes: true
```
fmoc, line 64: According to the docs, this should be either `Never`, `CaseInsensitive` or `CaseSensitive`. I'd suggest to use the latter.

TheOneRing: Yes but it would require clang 12 or something like that...

fmoc (approved):

PR #8820, opened 2021-07-15
Properly handle metadata updates

TheOneRing:
The first commit can be reviewed in https://github.com/owncloud/client/pull/8830

### DIFF
--- a/src/libsync/discovery.cpp
+++ b/src/libsync/discovery.cpp
@@ -13,20 +13,23 @@
  */
 
 #include "discovery.h"
+#include "common/checksums.h"
 #include "common/syncjournaldb.h"
+#include "csync.h"
+#include "csync_exclude.h"
+#include "owncloudpropagator.h"
+#include "syncengine.h"
 #include "syncfileitem.h"
-#include <QDebug>
+#include "vio/csync_vio_local.h"
+
 #include <algorithm>
 #include <set>
-#include <QTextCodec>
-#include "vio/csync_vio_local.h"
-#include <QFileInfo>
-#include <QFile>
-#include <QThreadPool>
-#include "common/checksums.h"
-#include "csync_exclude.h"
-#include "csync.h"
 
+#include <QDebug>
+#include <QFile>
+#include <QFileInfo>
+#include <QTextCodec>
+#include <QThreadPool>
 
 namespace OCC {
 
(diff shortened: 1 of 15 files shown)


src/libsync/discovery.cpp
```diff
@@ -1474,9 +1476,6 @@ void ProcessDirectoryJob::startAsyncLocalQuery()
         if (_serverQueryDone)
             this->process();
     });
-
-    QThreadPool *pool = QThreadPool::globalInstance();
-    pool->start(localJob); // QThreadPool takes ownership
```
dragotin, line 1479: Was this removed intentionally?

dragotin (approved):
LGTM, one question.

src/libsync/owncloudpropagator.cpp
```diff
@@ -1112,4 +1130,43 @@ QString OwncloudPropagator::remotePath() const
...
+            FileSystem::setFileReadOnlyWeak(filePath, isReadOnly);
+            if (_item->_checksumHeader.isEmpty()) {
+                _item->_checksumHeader = prev._checksumHeader;
+            }
+            _item->_serverHasIgnoredFiles |= prev._serverHasIgnoredFiles;
+        }
+    }
+    const auto result = propagator()->updateMetadata(*_item);
+    if (!result) {
+        done(SyncFileItem::SoftError, tr("Could not update file : %1").arg(result.error()));
```
dragotin, line 1168: In this case, the file is possibly treated by `setFileReadOnlyWeak` already, but the subsequent metadata update failed. Is it neccessary here to set the read-only flag back to be conform with the previous database entry?

dragotin (commented):
I like this one, but one question.

src/libsync/discovery.cpp
```diff
@@ -1047,23 +1050,20 @@ void ProcessDirectoryJob::processFileConflict(const SyncFileItemPtr &item, Proce
     if (up._valid && up._contentChecksum == serverEntry.checksumHeader) {
         // Solve the conflict into an upload, or nothing
         item->_instruction = up._modtime == localEntry.modtime && up._size == localEntry.size
-            ? CSYNC_INSTRUCTION_NONE : CSYNC_INSTRUCTION_SYNC;
+            ? CSYNC_INSTRUCTION_UPDATE_METADATA
+            : CSYNC_INSTRUCTION_SYNC;
```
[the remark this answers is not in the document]
fmoc: This line looks like a `clang-format` victim.

src/libsync/discovery.cpp
```diff
@@ -1047,23 +1050,20 @@ void ProcessDirectoryJob::processFileConflict(const SyncFileItemPtr &item, Proce
     if (up._valid && up._contentChecksum == serverEntry.checksumHeader) {
         // Solve the conflict into an upload, or nothing
```
fmoc, line 1051: Comment is probably outdated.

src/libsync/owncloudpropagator.cpp
TheOneRing: Hmm I assume having file permissions that differ from the db don't really hurt that much.
Also the step should be retried so once it succeeded we should be back in sync.

PR #8829, opened 2021-07-19
use QCOMPARE

TheOneRing:
if the test fails now we see why....


TheOneRing:
I changed the test later on so this was no longer useful

fmoc (approved):

PR #8833, opened 2021-07-21
Tests: It is possible that we got two itemComplete signals, handle it

TheOneRing:


TheOneRing:
Turned out that multiple items where indeed a bug

PR #8836, opened 2021-07-21
Exclude placeholder

TheOneRing:


changelog/unreleased/8836
```diff
@@ -0,0 +1,8 @@
+Bugfix: Always exclude .ofwncloud files
```
fmoc, line 1:
```suggestion
Bugfix: Always exclude .owncloud files
```

src/gui/issueswidget.cpp
```diff
@@ -126,7 +139,12 @@ void IssuesWidget::slotProgressInfo(const QString &folder, const ProgressInfo &p
         const auto &engine = f->syncEngine();
         const auto style = engine.lastLocalDiscoveryStyle();
         _model->remove_if([&](const ProtocolItem &item) {
-            if (item.folder()->path() != folder) {
+            if (item.folder() != f) {
+                return false;
+            }
+            if (item.direction() == SyncFileItem::None) {
```
fmoc, line 145: Might be easier to
```cpp
return item.folder() == f && item.direction() != SyncFileItem::None;

TheOneRing: As we still have a TODO here I'd like to keep it.
(Currently we get one warning per sync run..., without skipping them the issues will get removed after the discovery which doesn't make much sense)

dragotin: For what it is worth, I love the verbose version without the `&&` because I think it is easier to read and understand.

dragotin (approved):
LGTM

Issue #8822, opened 2021-07-16
Bundle resources using qrc to ease deployment


TheOneRing:
Deployment of additional resources is not easy, and can cause a lot of issues, e.g., because of differences between the target platforms. Some of these issues are eased by Qt. However, it might be beneficial to ship as many resources as possible using Qt's built-in resource system called [qrc](https://doc.qt.io/qt-5/resources.html).

Recently, we noticed that translations are not loaded correctly in the AppImages we build. This is due to a limitation of `QStandardPaths` which is documented in the [AppImage best practices](https://docs.appimage.org/reference/best-practices.html). This issue makes the binary non-relocatable, which is important to AppImages. We would have to implement a workaround specific to Linux to resolve this issue.

We already use qrc files to bundle resources in the binaries for theme files (cf. https://github.com/owncloud/client/blob/master/cmake/modules/OCGenerateTheme.cmake#L35, https://github.com/owncloud/client/blob/master/src/gui/main.cpp#L48).

We should bundle at least the translations using qrc as well, and then simplify the loading code accordingly.

fmoc:
@TheOneRing are you aware of other resources which we could bundle with qrc that we haven't thought of yet?

TheOneRing:
The exclude list but and the ownbrander2 theme, but not yet :)

TheOneRing:
At the moment I'd say we bundle everything that needs bundling.

PR #8838, opened 2021-07-23
Bundle translations using Qt resources system

fmoc:
Contributes to #8822.

This PR replaces the existing translations deployment in "guessable" directories. They are now bundled in the binary using the [Qt resources system](https://doc.qt.io/qt-5/resources.html). This is going to be a lot more reliable and also faster than the old system. The only disadvantage is that the binaries will become a little larger.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,7 +31,7 @@ include(CTest)
 
 
 include(OCConfigPluginDir)
-include(OCGenerateTheme)
+include(OCBundleResources)
 
 if (EXISTS "${PROJECT_SOURCE_DIR}/branding")
     set(OEM_THEME_DIR "${PROJECT_SOURCE_DIR}/branding" CACHE STRING "The directory containing a custom theme")
@@ -167,11 +167,6 @@ if( WIN32 )
     add_definitions(-DWINVER=0x0600)
     add_definitions(-DNTDDI_VERSION=0x0A000003)
 endif()
-
-# Handle Translations, pick all client_* files from trans directory.
-file( GLOB TRANS_FILES ${CMAKE_SOURCE_DIR}/translations/client_*.ts)
-set(TRANSLATIONS ${TRANS_FILES})
-
 
 file(GLOB_RECURSE OWNCLOUD_ICONS "${OEM_THEME_DIR}/theme/colored/*-${APPLICATION_ICON_NAME}-icon.png")
 if (NOT OWNCLOUD_ICONS)
--- a/cmake/modules/OCBundleResources.cmake
+++ b/cmake/modules/OCBundleResources.cmake
@@ -9,19 +9,19 @@ function(__addIcon THEME ICON_NAME)
     endif()
 
     set(icon "theme/${_ICON_SRC_PATH}/${ICON_NAME}.svg")
-    set(iconAlias "theme/${THEME}/${ICON_NAME}.svg")
+    set(iconAlias "${APPLICATION_SHORTNAME}/theme/${THEME}/${ICON_NAME}.svg")
     if (EXISTS ${OEM_THEME_DIR}/${icon})
         file(APPEND "${QRC}" "<file alias=\"${iconAlias}\">${OEM_THEME_DIR}/${icon}</file>\n")
     else()
         set(icon "theme/${_ICON_SRC_PATH}/${ICON_NAME}.png")
-        set(iconAlias "theme/${THEME}/${ICON_NAME}.png")
+        set(iconAlias "${APPLICATION_SHORTNAME}/theme/${THEME}/${ICON_NAME}.png")
         if (EXISTS ${OEM_THEME_DIR}/${icon})
             file(APPEND "${QRC}" "<file alias=\"${iconAlias}\">${OEM_THEME_DIR}/${icon}</file>\n")
         else()
             set(SIZES "16;22;32;48;64;128;256;512;1024")
             foreach(size ${SIZES})
                 set(icon "theme/${_ICON_SRC_PATH}/${ICON_NAME}-${size}.png")
-                set(iconAlias "theme/${THEME}/${ICON_NAME}-${size}.png")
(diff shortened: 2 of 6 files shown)


cmake/modules/OCGenerateResources.cmake
```diff
@@ -67,3 +77,30 @@ function(generate_legacy_icons theme_dir OUT)
     endforeach()
     set(${OUT} ${OWNCLOUD_ICONS} PARENT_SCOPE)
 endfunction()
+
+function(generate_resources QRC_PATH PREFIX FILES)
+    __write_qrc_file_header(${QRC_PATH} l10n)
```
TheOneRing, line 82: A general generate_resources  and a hardcoded l10n?

fmoc: I'm rewriting this file at the moment to eliminate the remaining duplicate code. I'm not satisfied with the way the icon files are handled at the moment, and have a better, more generic solution in mind which I implement at the moment. The functions I added use too many parameters...

I'll fix the hardcoded `l10n` value, thanks for the hint.

src/gui/CMakeLists.txt
```diff
@@ -204,8 +204,9 @@ target_link_libraries(owncloud owncloudCore)
 
 find_package(Qt5LinguistTools)
 if(Qt5LinguistTools_FOUND)
-    qt5_add_translation(client_I18N ${TRANSLATIONS})
-    target_sources(owncloud PRIVATE ${client_I18N})
+    qt5_add_translation(client_compiled_translations ${TRANSLATIONS})
+    target_sources(owncloud PRIVATE ${client_compiled_translations})
+    add_resources_to_target(owncloud translations.qrc translations "${client_compiled_translations}")
```
TheOneRing, line 209: a qrc should be part of the source files

src/gui/CMakeLists.txt
```diff
@@ -238,14 +239,14 @@ if(NOT APPLE)
         endforeach(_file)
     endif()
 
-    install(FILES ${client_I18N} DESTINATION ${DATA_INSTALL_DIR}/${APPLICATION_NAME}/i18n)
+    install(FILES ${client_compiled_translations} DESTINATION ${DATA_INSTALL_DIR}/${APPLICATION_NAME}/i18n)
```
TheOneRing, line 242: We don't have anything to install as the resources are compiled into the binary?

src/gui/CMakeLists.txt
fmoc: That's done by `target_sources` in the `add_resources_to_target` function The icons function also uses `target_sources`. Instead of "returning" the name of the generated file, I thought it's easier to avoid return parameters, but instead just pass the target name and let the function figure out the rest.

src/gui/CMakeLists.txt
fmoc: Right, thanks.

src/gui/CMakeLists.txt
```diff
@@ -238,14 +239,12 @@ if(NOT APPLE)
...
 
-    install(FILES ${client_I18N} DESTINATION ${DATA_INSTALL_DIR}/${APPLICATION_NAME}/i18n)
-
     # we may not add MACOSX_BUNDLE here, if not building one
 else()
     target_sources(owncloud PRIVATE ${OWNCLOUD_BUNDLED_RESOURCES})
 
     set_source_files_properties(
-      ${client_I18N}
+      ${client_compiled_translations}
```
TheOneRing, line 247: we also don't deploy them :) (install them in the .app bundle)

TheOneRing (approved):

CMakeLists.txt
```diff
@@ -169,8 +169,7 @@ if( WIN32 )
 endif()
 
 # Handle Translations, pick all client_* files from trans directory.
-file( GLOB TRANS_FILES ${CMAKE_SOURCE_DIR}/translations/client_*.ts)
-set(TRANSLATIONS ${TRANS_FILES})
+file(GLOB TRANSLATIONS ${CMAKE_SOURCE_DIR}/translations/client_*.ts)
```
TheOneRing, line 172: Hmm better move the glob to where its used.

cmake/modules/OCBundleResources.cmake
```diff
@@ -0,0 +1,166 @@
...
+                    )
+                endif()
+            endforeach()
+        endif()
+    endif()
+endfunction()
+
+function(__write_qrc_file_header QRC_PATH FILES_PREFIX)
+    file(WRITE ${QRC_PATH} "<RCC>\n")
+    file(APPEND ${QRC_PATH} "    <qresource prefix=\"/client/${APPLICATION_SHORTNAME}\">\n")
```
TheOneRing, line 65: Hmm the files are bot available through the alias and the actual path.
Maybe we can make the actual path make look less official?
image

src/gui/translations.cpp
```diff
@@ -42,20 +42,11 @@ namespace Translations {
...
-            return devTrPath.absolutePath();
-        }
-#ifdef Q_OS_MAC
-        const auto translationDir = QStringLiteral("Translations");
-#else
-        const auto translationDir = QStringLiteral("i18n");
-#endif
-        return QStandardPaths::locate(QStandardPaths::AppDataLocation, translationDir, QStandardPaths::LocateDirectory);
+        QDir dir(":/client/" APPLICATION_SHORTNAME "/translations/");
+        Q_ASSERT(dir.exists());
```
TheOneRing, line 48: Translations are optional and at the moment the squish tests are not building them
```09:07:21:300 AUT stdout: 07-27 09:07:20:390 [ fatal default ]:	ASSERT: "dir.exists()" in file ../src/gui/translations.cpp, line 48

cmake/modules/OCBundleResources.cmake
fmoc: Looks like Qt creator's rendering is suboptimal there. Gammaray shows the correct paths, as expected.

I removed the shortname because translations are not brand-specific.

Issue #8811, opened 2021-07-13
Handle BigIP F5 session hangup


TheOneRing:
When a F5 session times out, open connections are closed with a 502.

Properly handle those as soft error in uploads and downloads.

PR #8844, opened 2021-07-27
Treat a 502 as SoftError and retry the sync

TheOneRing:
Fixes: #8811

### DIFF
--- a/src/libsync/owncloudpropagator_p.h
+++ b/src/libsync/owncloudpropagator_p.h
@@ -56,7 +56,23 @@ inline SyncFileItem::Status classifyError(QNetworkReply::NetworkError nerror,
         return SyncFileItem::FatalError;
     }
 
-    if (httpCode == 503) {
+    switch (httpCode) {
+    case 423:
+        // "Locked"
+        // Should be temporary.
+        Q_FALLTHROUGH();
+    case 502:
+        // "Bad Gateway"
+        // Should be temporary.
+        if (anotherSyncNeeded != nullptr) {
+            *anotherSyncNeeded = true;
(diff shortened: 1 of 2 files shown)


src/libsync/owncloudpropagator_p.h
```diff
@@ -56,7 +56,23 @@ inline SyncFileItem::Status classifyError(QNetworkReply::NetworkError nerror,
...
+        // "Bad Gateway"
+        // Should be temporary.
+        if (anotherSyncNeeded != nullptr) {
+            *anotherSyncNeeded = true;
+        }
+        Q_FALLTHROUGH();
+    case 412:
+        // "Precondition Failed"
+        // Happens when the e-tag has changed
+        return SyncFileItem::SoftError;
```
dragotin, line 74: Hm, I do not like this way of implementation. I had to look up what Q_FALLTHROUGH exactly does to understand...
Why not `if (httpCode == 423 || httpCode == 502 || httpCode == 412)` ? Would be clearer for me.

dragotin (approved):

src/libsync/owncloudpropagator_p.h
fmoc: The switch-case allows the developer to add important documentary comments in a way that is much easier to read. Above an `if` as you suggest, you'd have to repeat all the numbers for instance to have an association between the number and the comment. That would be redundant.

`Q_FALLTHROUGH` basically just silences a compiler warning by explicitly stating "this is intended to fall through". I personally don't mind this at all. It's a lot better than having redundant, nested `if`s.

PR #8848, opened 2021-07-27
Add the account context menu to the account buttons

TheOneRing:
image

### DIFF
--- a/changelog/unreleased/8848
+++ b/changelog/unreleased/8848
@@ -1,0 +1,5 @@
+Enhancement: The account menu is now available at on the Avatar button
+
+We made the Account menu available on the Avatar button.
+
+https://github.com/owncloud/client/pull/8848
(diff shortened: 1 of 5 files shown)


michaelstingl (approved):
yeah, works for me 👍

dragotin:
Tbh, I dislike. Is that in addition to the (also not optimal) combo box? If the plan is to remove the combo, which would make sense, I think the account avatar "buttons" would need some visual indication that there is an menu to open, and in that case it should not be right click.

More like this (example): 
ksnip_20210728-151941

TheOneRing:
I just happend to click on the Avatar as I kind of expected options to be available especially if you have multiple.
Yes they are basically buttons but the way we use them is not how buttons are used.

erikjv:
I have to say that I'd never guess that right-clicking on one of those toolbar buttons would actually do something. I would just copy the ... button that the folder has and use it instead of the account combo.

TheOneRing:
You are all right, I guess we should kill the old  button and move the menu to the account button?

fmoc:
Yep, sounds like a good idea.

Should the PR target 2.9 already?

PR #8853, opened 2021-07-29
Prevent sync root sharing

TheOneRing:

### DIFF
--- a/src/gui/folderman.cpp
+++ b/src/gui/folderman.cpp
@@ -1403,24 +1403,21 @@ QString FolderMan::checkPathValidityForNewFolder(const QString &path, const QUrl
     }
 
     // check if the local directory isn't used yet in another ownCloud sync
-    Qt::CaseSensitivity cs = Qt::CaseSensitive;
-    if (Utility::fsCasePreserving()) {
-        cs = Qt::CaseInsensitive;
-    }
+    const auto cs = Utility::fsCaseSensitivity();
 
     const QString userDir = QDir::cleanPath(canonicalPath(path)) + '/';
     for (auto i = _folderMap.constBegin(); i != _folderMap.constEnd(); ++i) {
         Folder *f = static_cast<Folder *>(i.value());
-        QString folderDir = QDir::cleanPath(canonicalPath(f->path())) + '/';
+        const QString folderDir = QDir::cleanPath(canonicalPath(f->path())) + '/';
 
-        bool differentPaths = QString::compare(folderDir, userDir, cs) != 0;
-        if (differentPaths && folderDir.startsWith(userDir, cs)) {
+        const bool differentPaths = QString::compare(folderDir, userDir, cs) != 0;
+        if (differentPaths && FileSystem::isChildPathOf(folderDir, userDir)) {
             return tr("The local folder %1 already contains a folder used in a folder sync connection. "
                       "Please pick another one!")
                 .arg(QDir::toNativeSeparators(path));
         }
 
-        if (differentPaths && userDir.startsWith(folderDir, cs)) {
+        if (differentPaths && FileSystem::isChildPathOf(userDir, folderDir)) {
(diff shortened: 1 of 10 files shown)


TheOneRing:
The first two commits are available in separate reviews.

src/gui/folderman.cpp
```diff
@@ -1371,9 +1371,12 @@ static QString checkPathValidityRecursive(const QString &path)
     Utility::NtfsPermissionLookupRAII ntfs_perm;
 #endif
     const QFileInfo selFile(path);
+    if (QFileInfo::exists(selFile.filePath() + QStringLiteral("/.owncloudsync.log"))) {
```
TheOneRing, line 1374: We could try to provide a different warning in this case as in theory using an old sync dir is a valid scenario...

dragotin: I would rather check on the syncjournal db file, as that is going to be removed in `Folder::wipeForRemoval`.

TheOneRing: Thx! ❤️

TheOneRing:
The last commit replaces the partly md5 sum in the snyc db name with ".sync_journal.db"

src/gui/folderman.cpp
```diff
@@ -1394,90 +1397,62 @@ static QString canonicalPath(const QString &path)
...
     for (auto i = _folderMap.constBegin(); i != _folderMap.constEnd(); ++i) {
         Folder *f = static_cast<Folder *>(i.value());
-        QString folderDir = QDir::cleanPath(canonicalPath(f->path())) + '/';
+        const QString folderDir = QDir::cleanPath(canonicalPath(f->path())) + '/';
 
-        bool differentPaths = QString::compare(folderDir, userDir, cs) != 0;
-        if (differentPaths && folderDir.startsWith(userDir, cs)) {
+        if (QString::compare(folderDir, userDir, cs) == 0) {
+            return tr("There is already a sync from the server to this local folder. "
+                      "Please pick another local folder!");
```
dragotin, line 1412: The check checks for sync connection from any server, right? So maybe a better string would be something like `This local folder is already used for another sync. Please pick...`

dragotin (approved):
Minor comments.

src/gui/folderman.cpp
```diff
@@ -1394,90 +1397,62 @@ static QString canonicalPath(const QString &path)
...
-    if (Utility::fsCasePreserving()) {
-        cs = Qt::CaseInsensitive;
-    }
+    const auto cs = Utility::fsCaseSensitivity();
 
     const QString userDir = QDir::cleanPath(canonicalPath(path)) + '/';
     for (auto i = _folderMap.constBegin(); i != _folderMap.constEnd(); ++i) {
         Folder *f = static_cast<Folder *>(i.value());
-        QString folderDir = QDir::cleanPath(canonicalPath(f->path())) + '/';
+        const QString folderDir = QDir::cleanPath(canonicalPath(f->path())) + '/';
```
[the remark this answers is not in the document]
TheOneRing: I started to work on something like that change kind of exploded, now I'm targeting 2.10 with it if I find time to finish it one day. https://github.com/owncloud/client/pull/8783

src/gui/folderman.cpp
```diff
@@ -1355,9 +1355,12 @@ static QString checkPathValidityRecursive(const QString &path)
     Utility::NtfsPermissionLookupRAII ntfs_perm;
 #endif
     const QFileInfo selFile(path);
+    if (!selFile.dir().entryList({ QStringLiteral(".sync_*.db") }, QDir::Hidden | QDir::Files).isEmpty()) {
```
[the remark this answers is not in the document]
TheOneRing: Hmm I'd prefer to keep * with the change to .sync_journal.db .
Maybe one day we want the branding name in there or something similar and the check would fail for no reason.

PR #8854, opened 2021-07-30
Wizard: Display sync root issues as dialog

TheOneRing:
image

The commit also removes some unused labels and uses a stack widget instead of hiding/showing widgets during runtime which messed up the layout.

Old:
image

### DIFF
--- a/src/gui/wizard/owncloudadvancedsetuppage.cpp
+++ b/src/gui/wizard/owncloudadvancedsetuppage.cpp
@@ -53,9 +53,7 @@ OwncloudAdvancedSetupPage::OwncloudAdvancedSetupPage()
 
     registerField(QLatin1String("OCSyncFromScratch"), _ui.cbSyncFromScratch);
 
-    _ui.errorScrollContents->layout()->addWidget(_progressIndi);
     stopSpinner();
-    setupCustomization();
 
     connect(_ui.pbSelectLocalFolder, &QAbstractButton::clicked, this, &OwncloudAdvancedSetupPage::slotSelectFolder);
     setButtonText(QWizard::NextButton, tr("Connect..."));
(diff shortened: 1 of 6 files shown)


michaelstingl (approved):
Yeah, LGTM

src/gui/wizard/owncloudadvancedsetuppage.cpp
```diff
@@ -196,9 +180,9 @@ void OwncloudAdvancedSetupPage::updateStatus()
         if (dirNotEmpty) {
             t += tr("<p><small><strong>Warning:</strong> The local folder is not empty. "
```
fmoc, line 181: Isn't this text kind of redundant to the new widget?

I think as a user, I wouldn't be sure to know whether both are related. Also, I would have to ask myself, "what is this resolution they're talking about?", it doesn't really get clear from the text what it wants me, the user, to do.

src/gui/wizard/owncloudadvancedsetuppage.cpp
```diff
@@ -196,9 +180,9 @@ void OwncloudAdvancedSetupPage::updateStatus()
...
                     "Pick a resolution!</small></p>");
-            _ui.resolutionWidget->setVisible(dirNotEmpty);
+            _ui.resolutionStackedWidget->setCurrentIndex(1);
```
fmoc, line 183: Is there a specific advantage of using the index over the visibility? It seems a bit counterintuitive to me to do it like that. Please add a small comment that explains why this solution is better to avoid confusion in the future.

TheOneRing: If you hide and show a widget it messes with the layout, first its size is 0 than  not.
The stackedwidget always has a size and is properly layd out.

src/gui/wizard/owncloudadvancedsetuppage.cpp
TheOneRing: This is the default behaviour if you select a non empty dir, the message box only shows up if you try to use a forbidden dir.
But yes the messing popping up below the choices is 🤣

TheOneRing:
Update 2:
image

PR #8857, opened 2021-08-02
Ensure the tests are properly set up

TheOneRing:
Q_COREAPP_STARTUP_FUNCTION doesn't work reliably in a static lib

### DIFF
--- a/test/owncloud_add_test.cmake
+++ b/test/owncloud_add_test.cmake
@@ -6,12 +6,12 @@ function(owncloud_add_test test_class)
     set(OWNCLOUD_TEST_CLASS ${test_class})
     string(TOLOWER "${OWNCLOUD_TEST_CLASS}" OWNCLOUD_TEST_CLASS_LOWERCASE)
 
-    ecm_add_test(test${OWNCLOUD_TEST_CLASS_LOWERCASE}.cpp ${ARGN}
+    ecm_add_test(test${OWNCLOUD_TEST_CLASS_LOWERCASE}.cpp
+        ${ARGN}
         TEST_NAME "${OWNCLOUD_TEST_CLASS}Test"
         LINK_LIBRARIES
-        owncloudCore syncenginetestutils Qt5::Test
+        owncloudCore syncenginetestutils testutilsloader Qt5::Test
     )
-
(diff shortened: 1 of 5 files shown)


dschmidt (changes requested):
as testutilsloader.cpp is now compiled into every test and testutils have their own CMakeLists.txt file, to me testutilsloader.cpp should not be in the testutils folder.

apart from that, LGTM

TheOneRing:
> 
> 
> as testutilsloader.cpp is now compiled into every test and testutils have their own CMakeLists.txt file, to me testutilsloader.cpp should not be in the testutils folder.
> 
> apart from that, LGTM

Hm but its part of the test utils and not a test itself, I'm trying to not fill the root folder by way more files than required....

dschmidt:
I don't think that's really an issue, all tests start with "test" prefix anyway. 
In doubt create an additional folder, but don't mix the lib and externally compiled in files, that's very confusing imho.

test/owncloud_add_test.cmake
```diff
@@ -6,7 +6,7 @@ function(owncloud_add_test test_class)
     set(OWNCLOUD_TEST_CLASS ${test_class})
     string(TOLOWER "${OWNCLOUD_TEST_CLASS}" OWNCLOUD_TEST_CLASS_LOWERCASE)
 
-    ecm_add_test(test${OWNCLOUD_TEST_CLASS_LOWERCASE}.cpp ${ARGN}
+    ecm_add_test(test${OWNCLOUD_TEST_CLASS_LOWERCASE}.cpp ${PROJECT_SOURCE_DIR}/test/testutils/testutilsloader.cpp ${ARGN}
```
fmoc, line 9: Please add a comment that you cannot put this file into a library because that breaks that Qt macro so we don't forget in the future.

TheOneRing:
It is really part of the testutils and not part of the test..
@fmoc suggested https://cmake.org/cmake/help/latest/command/add_library.html#object-libraries but that's not supported on the platforms we support....
I'm currently try to not drop all src we got in to one single folder...

fmoc:
If you prefer some sort of CMake library that is then added as a dependency to the target, there are [object libraries](https://cmake.org/cmake/help/latest/command/add_library.html#object-libraries) which are relatively similar to other types of libraries within CMake. They are relatively new, though, so we'd need to bump the CMake minimum required version.

I don't really find this confusing, to be honest, as long as the reason why it's done like this from now on is documented in the relevant places within the code.

TheOneRing:
Happy?

Issue #2432, opened 2014-10-30
Tray status icon and client status icon differ


jnfrmarks:
### Steps to reproduce

1. Upgrading from 1.6.3 (or maybe 1.6.4) to 1.7
2. The status icon on the tray shows a green check mark
3. If I open up the client, the status icon is blue for syncing

### Expected behaviour

Status icons are the same

### Actual behaviour

Status icons are different

![p196](https://cloud.githubusercontent.com/assets/8480343/4851340/94845f02-6071-11e4-940f-bc0018bef411.PNG)

### Client configuration
**Operating system**:

Windows

strugee:
It's worth noting that should the green checks be changed (instead of the blue sync icon), in addition to the status bar icon, we'd also need to change the icon on the "account" tab on the left of the client settings.

strugee:
> I think that is intentional because @jancborchardt didn't want us to disturb the user when he doesnt have the window open.

How is the blue icon "disturbing the user"? It's an ambient indicator and doesn't really get in anyone's way. And I'm worried that some users won't think to check the settings before assuming that the client isn't doing anything.

jancborchardt:
@strugee the blue arrow icon stands for »syncing«. If there is nothing syncing, but only checking if there’s something to sync, the icon should not show. If it shows when you know that you didn’t change anything, it would lead to confusion.

dragotin:
Sorry, confusion! What we agreed on was: For the discovery phase, we show the blue sync icon along the "discovering... " message in the settings dialog, becasue that is only opened for users who wanna know the details. 
The tray icon will not change to the sync icon for the discovery phase to not to disturb the user. That is exactly what the screenshot shows: The tray is green, but the one in the settings dialog shows activity. 

I'd suggest to close this as "works as designed" by @jancborchardt .

strugee:
> @strugee the blue arrow icon stands for »syncing«. If there is nothing syncing, but only checking if there’s something to sync, the icon should not show. If it shows when you know that you didn’t change anything, it would lead to confusion.

The problem with this approach is that if the discovery phase takes a long time, and you have files to sync, then it will appear as if the client is doing nothing even though you know that your files haven't been uploaded. Very frustrating unless you know what's going on. (Also, I'd argue that "syncing" with the server would include the discovery phase, where the client isn't necessarily uploading or downloading files, but instead is synchronizing state. FWIW I believe both Dropbox and Google Drive show syncing indications upon startup.)

[...]

PR #8858, opened 2021-08-02
Align sync status and icon

TheOneRing:
With this change we will display the problem state if we have blacklisted errors.
This change also aligns the icon in the tray icon with the icon displayed for the folder state.

image

### DIFF
--- a/src/gui/folderman.h
+++ b/src/gui/folderman.h
@@ -43,9 +43,13 @@ class LockWatcher;
 class TrayOverallStatusResult
 {
 public:
-    SyncResult::Status overallStatus;
-    bool hasUnresolvedConflicts;
     QDateTime lastSyncDone;
+
+    void addResult(const SyncResult &result);
+    const SyncResult &overallStatus() const;
+
+private:
+    SyncResult _overallStatus;
 };
 
 /**
@@ -145,7 +149,7 @@ public:
     bool startFromScratch(const QString &);
 
     /// Produce text for use in the tray tooltip
-    static QString trayTooltipStatusString(SyncResult::Status syncStatus, bool hasUnresolvedConflicts, bool paused);
+    static QString trayTooltipStatusString(const SyncResult &result, bool paused);
 
     /**
      * Compute status summarizing multiple folders
--- a/src/libsync/theme.cpp
+++ b/src/libsync/theme.cpp
@@ -477,33 +477,41 @@ bool Theme::aboutShowCopyright() const
 #ifndef TOKEN_AUTH_ONLY
 QIcon Theme::syncStateIcon(SyncResult::Status status, bool sysTray, bool sysTrayMenuVisible) const
 {
-    // FIXME: Mind the size!
+    return syncStateIcon(SyncResult { status }, sysTray, sysTrayMenuVisible);
+}
+QIcon Theme::syncStateIcon(const SyncResult &result, bool sysTray, bool sysTrayMenuVisible) const
+{
     QString statusIcon;
 
-    switch (status) {
-    case SyncResult::Undefined:
-        // this can happen if no sync connections are configured.
-        statusIcon = QStringLiteral("state-information");
-        break;
+    switch (result.status()) {
     case SyncResult::NotYetStarted:
+        Q_FALLTHROUGH();
     case SyncResult::SyncRunning:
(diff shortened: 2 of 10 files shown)


TheOneRing:
image

michaelstingl (commented):
Changes the "works as designed" behaviour explained in https://github.com/owncloud/client/issues/2432

michaelstingl (approved):
I think this is a good change 👍

src/gui/folderman.h
```diff
@@ -43,9 +43,14 @@ class LockWatcher;
 class TrayOverallStatusResult
 {
 public:
-    SyncResult::Status overallStatus;
-    bool hasUnresolvedConflicts;
     QDateTime lastSyncDone;
+
+    void addResult(const SyncResult &result);
+    void setStatus(SyncResult::Status status);
```
fmoc, line 49: Do we really need this setter? I think you could set the overall status within `addResult` by remembering the "worst" status you've seen so far, right?

src/gui/folderman.h
```diff
@@ -145,7 +150,7 @@ class FolderMan : public QObject
     bool startFromScratch(const QString &);
 
     /// Produce text for use in the tray tooltip
-    static QString trayTooltipStatusString(SyncResult::Status syncStatus, bool hasUnresolvedConflicts, bool paused);
+    static QString trayTooltipStatusString(const SyncResult &syncStatus, bool paused);
```
fmoc, line 153: Just like above, the variable should be called `*result`.

src/libsync/theme.cpp
```diff
@@ -477,33 +477,41 @@ bool Theme::aboutShowCopyright() const
...
+        Q_FALLTHROUGH();
     case SyncResult::Paused:
         statusIcon = QStringLiteral("state-pause");
         break;
     case SyncResult::SyncPrepare:
+        Q_FALLTHROUGH();
     case SyncResult::Success:
+        if (status.hasUnresolvedConflicts()) {
+            return syncStateIcon(SyncResult { SyncResult::Problem }, sysTray, sysTrayMenuVisible);
+        }
```
fmoc, line 502: We need to do something about the formatter here. The placing of these braces is misleading. I needed to read it a second time to see that it's constructing a new object.

src/libsync/theme.h
```diff
@@ -101,7 +101,9 @@ class OWNCLOUDSYNC_EXPORT Theme : public QObject
     /**
       * get an sync state icon
       */
-    virtual QIcon syncStateIcon(SyncResult::Status, bool sysTray = false, bool sysTrayMenuVisible = false) const;
+    QIcon syncStateIcon(const SyncResult &status, bool sysTray = false, bool sysTrayMenuVisible = false) const;
```
fmoc, line 104: Consider renaming the first param to `result`.

fmoc (changes requested):

fmoc (approved):

avrhack:
I think this is a massive mistake - I have loads of deliberately ignored files and hidden files which I choose not to sync, so the client always shows the icon instead of success which is completely misleading. This should not happen where the 'errors' are warnings caused by specific user choices that are known about.

TheOneRing:
@avrhack Please see the discussion in https://github.com/owncloud/client/issues/8970 hidden files won't display the i icon.

avrhack:
> @avrhack Please see the discussion in #8970 hidden files won't display the i icon.

Thanks Hannah - I think that discussion just confirms my point; this change is a huge step backwards.

You say hidden files won't display the icon but all my errors are either hidden or ignored files and the icon DOES show so I'm afraid I have to respectively disagree with you. The icon should not be showing in this circumstance but it is.

I've had to revert to 2.8.2 as 2.9.0 is now unusable for me with this 'enhancement' and from that discussion others feel the same.

s1mm0n:
Also this seems to affect VFS.

All Virtual Files will be shown as "ignored" with the yellow icon.

MacOS 11.5.1 - Client 2.9.0 Build 5150

image

TheOneRing:
Can you check whether there is a file2.jpeg.owncloud on your server? Stuff like that used to happen in the past.
If not please open a new issue and don't comment on pull requests.

PR #8860, opened 2021-08-02
Fix config data migration from 2.8 to 2.9

### LINKED ISSUE #8824: Regression use of legacy settings location
Testpilot cloud now falls back to the legacy location C:\Users\hanna\AppData\Roaming\testpilotcloud....

Introduced in https://github.com/owncloud/client/pull/8670

The migration in https://github.com/owncloud/client/blob/master/src/gui/application.cpp#L171 was broken since the first day as 
https://github.com/owncloud/client/blob/master/src/libsync/configfile.cpp#L112 was setting the legacy name each time the constructor is called.

TODO:
- Remove all calls to setApplicationName besides the main one in Application.
- Implement an additional migration path to the short name location.

fmoc:
Closes #8824.

### DIFF
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -66,6 +66,76 @@
 
 class QSocket;
 
+namespace {
+
+void migrateConfigFile(const QCoreApplication *app)
+{
+    using namespace OCC;
+
+    if (!ConfigFile::exists()) {
+        // check whether an old config location must be migrated
+        // we support multiple locations from old versions
+        // these are worked on in-order to upgrade from version to version
+        // the algorithm is the same for all these locations, thus we can use a loop
+        QStringList configLocationsToMigrate;
+
+        {
+            // note: this change is temporary to allow using QDesktopServices etc. to determine the paths
+            // the application name was changed to
+            auto scopeGuard = qScopeGuard([&app, oldApplicationName = app->applicationName()] {
+                // reset to original value
+                app->setApplicationName(oldApplicationName);
+            });
+
+            QCoreApplication::setApplicationName(Theme::instance()->appNameGUI());
+
+            // location used in versions <= 2.4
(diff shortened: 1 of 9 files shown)


src/gui/application.cpp
```diff
@@ -66,6 +66,72 @@
...
+    if (!ConfigFile::exists()) {
+        // check whether an old config location must be migrated
+        // we support multiple locations from old versions
+        // these are worked on in-order to upgrade from version to version
+        // the algorithm is the same for all these locations, thus we can use a loop
+        QStringList configLocationsToMigrate;
+
+        // note: this change is temporary to allow using QDesktopServices etc. to determine the paths
+        // the application name was changed to
+        const auto oldApplicationName = app->applicationName();
```
TheOneRing, line 84: Maybe a inner scope with a QScopeGuard  to reset the app name

src/gui/application.cpp
```diff
@@ -66,6 +66,72 @@
...
+                auto confDir = ConfigFile::configPath();
+                if (confDir.endsWith('/'))
+                    confDir.chop(1); // macOS 10.11.x does not like trailing slash for rename/move.
+                qCInfo(lcApplication) << "Migrating old config from" << oldDir << "to" << confDir;
+
+                if (!QFile::rename(oldDir, confDir)) {
+                    qCWarning(lcApplication) << "Failed to move the old config directory to its new location (" << oldDir << "to" << confDir << ")";
+
+                    // Try to move the files one by one
+                    if (QFileInfo(confDir).isDir() || QDir().mkdir(confDir)) {
```
TheOneRing, line 113: While touching this, maybe https://doc.qt.io/qt-5/qdir.html#entryInfoList-1 instead of manually constructing the path.

`QDir().mkdir(confDir)` will try to create `confDir` in the current working dir 😲 I'd say that never worked.

TheOneRing (approved):

PR #8873, opened 2021-08-09
Set up Git pre-commit hook for clang-format

fmoc:
I saw these hooks in other projects, and I think they are pretty useful. Right now only CMake's "Unix" platforms (i.e., Linux, BSD, ...) seem to be supported.

I did not want to bump the ECM version requirement, thus the hook is only enabled if ECM happens to be recent enough. Otherwise, a warning is printed.


fmoc:
@TheOneRing I guess this PR should target `2.9`, right?

TheOneRing:
> 
> 
> @TheOneRing I guess this PR should target `2.9`, right?

yes

TheOneRing:
then ship it

TheOneRing (approved):

PR #8874, opened 2021-08-09
Fix window size when scaled

TheOneRing:


TheOneRing:
```
14:48:48:801 AUT stdout: 08-09 14:48:47:896 [ info sync.propagator ]:	Completed propagation of "Photos/Teotihuacan.jpg" by OCC::PropagateDownloadFile(0x55de5de69a40) with status OCC::SyncFileItem::Success
2021-08-09T14:49:07	ERROR     	/drone/src/test/gui/shared/steps/steps.py:760: Script Error	LookupError: Object '{type='QMenu' unnamed='1' visible='1' window={name='Settings' type='OCC::SettingsDialog' visible='1'}}' not found. Could not match properties:\n    visible for object name: '{type='QMenu' unnamed='1' visible='1' window=:settings_OCC_SettingsDialog}'
```
I assume as we have no screen attached the resulting size is 0 or something like that...

src/gui/settingsdialog.cpp
```diff
@@ -267,7 +265,20 @@ SettingsDialog::~SettingsDialog()
 
 QSize SettingsDialog::minimumSizeHint() const
 {
-    return {800, 500};
+    const auto screen = windowHandle() ? windowHandle()->screen() : QApplication::screenAt(QCursor::pos());
+    const auto geometry = screen->availableSize();
+    const QSize min { 800, 600 };
+    const QSize currentMin { static_cast<int>(geometry.width() * 0.75), static_cast<int>(geometry.height() * 0.75) };
+    if (currentMin.width() < min.width() || currentMin.height() < min.height()) {
```
fmoc, line 272: This line is confusing. Why use `currentMin` if it's smaller than `min`? Shouldn't it be the other way round, i.e., one should use the larger of both?

fmoc (approved):

PR #8879, opened 2021-08-10
Remove dead wizard page

TheOneRing:
We now also get a finish button 🥳 
image

That we now display the finish button and not the next button breaks the squish tests.
@Talank @swoichha


src/gui/wizard/owncloudadvancedsetuppage.cpp
```diff
@@ -194,7 +194,7 @@ void OwncloudAdvancedSetupPage::stopSpinner()
 
 int OwncloudAdvancedSetupPage::nextId() const
 {
-    return WizardCommon::Page_Result;
+    return -1;
```
fmoc, line 197:
```suggestion
    // tells the caller that this is the last dialog page
    return -1;
```

src/gui/wizard/owncloudwizard.cpp
```diff
@@ -161,13 +152,17 @@ void OwncloudWizard::successfulStep()
...
 
     case WizardCommon::Page_ServerSetup:
-    case WizardCommon::Page_Result:
         qCWarning(lcWizard, "Should not happen at this stage.");
         break;
     }
 
     ownCloudGui::raiseDialog(this);
-    next();
+    if (nextId() == -1) {
```
fmoc, line 160: Perhaps

```suggestion
    if (nextId() < -1) {
```

Otherwise, it might make sense to just use `ssize_t` instead of `int`.

fmoc (approved):

src/gui/wizard/owncloudwizard.cpp
TheOneRing: <= I assume :D

TheOneRing: But only -1 is a valid page if you consider the doc

PR #8887, opened 2021-08-11
Properly handle return value

TheOneRing:


src/libsync/owncloudpropagator.cpp
```diff
@@ -752,20 +752,14 @@ QString OwncloudPropagator::adjustRenamedPath(const QString &original) const
...
 Result<Vfs::ConvertToPlaceholderResult, QString> OwncloudPropagator::updatePlaceholder(const SyncFileItem &item, const QString &fileName, const QString &replacesFile)
 {
-    const auto result = syncOptions()._vfs->updateMetadata(item, fileName, replacesFile);
-    if (!result) {
-        return result.error();
-    } else if (*result == Vfs::ConvertToPlaceholderResult::Locked) {
-        return Vfs::ConvertToPlaceholderResult::Locked;
-    }
-    return Vfs::ConvertToPlaceholderResult::Ok;
+    return syncOptions()._vfs->updateMetadata(item, fileName, replacesFile);
```
dschmidt, line 755: 😍

src/libsync/propagatedownload.cpp
```diff
@@ -927,6 +927,9 @@ void PropagateDownloadFile::downloadFinished()
         if (!result) {
             done(SyncFileItem::NormalError, result.error());
             return;
+        } else if (result.get() == Vfs::ConvertToPlaceholderResult::Locked) {
+            done(SyncFileItem::SoftError, tr("The file %1 is currently in use").arg(_item->_file));
+            return;
         }
```
dschmidt, line 933: Can we somehow enforce checking this better? 
So easy to miss one of the cases (null or != Ok).. :-\

dschmidt (approved):

src/libsync/propagatedownload.cpp
TheOneRing: Good question.
The old solution for something like that would have been an error parameter.
But I'm really not sure that would enhance the situation...

Issue #8897, opened 2021-08-13
[QA] Adjust existing GUI tests for 2.9


swoichha:
### Description
As `2.9` is in-progress some of the existing tests are failing due to change in UI elements.

For example:
test for adding new account fails because it can not find the account connect button as it has been changed.

### To-Do:
Fix the connect button object to make it work as per the new design and code changes.

TheOneRing:
Before fixing it, could you specify the breaking change, it can always be a bug :)

swoichha:
After https://github.com/owncloud/client/pull/8879 has been merged. Now we have Finish button instead of Connect. Also, the object name for the button has been changed from `"name": "__qt__passive_wizardbutton1"` to `"name": "qt_wizard_finish"`. So, we need to fix that.

swoichha:
@TheOneRing  Another thing that I found while fixing this issue was that we cannot use same `client_sync_path`. Previously when we added two users to client then they both had same sync path i.e. `/tmp/client-bdd` but now we need to have separate sync folder else the test fails with error like this

![error](https://user-images.githubusercontent.com/17021686/129344960-e9cb5724-e948-4f04-874a-80828f415ded.jpeg)

Is this suppose to happen or is it an error?

TheOneRing:
> 
> 
> @TheOneRing Another thing that I found while fixing this issue was that we cannot use same `client_sync_path`. Previously when we added two users to client then they both had same sync path i.e. `/tmp/client-bdd` but now we need to have separate sync folder else the test fails with error like this
> 
> ![error](https://user-images.githubusercontent.com/17021686/129344960-e9cb5724-e948-4f04-874a-80828f415ded.jpeg)
> 
> Is this suppose to happen or is it an error?

Sync root sharing was removed as it could cause data loss when virtual files where involved.

TheOneRing:
> 
> 
> After #8879 has been merged. Now we have Finish button instead of Connect. Also, the object name for the button has been changed from `"name": "__qt__passive_wizardbutton1"` to `"name": "qt_wizard_finish"`. So, we need to fix that.

Yes it is now using the finish button as I removed a dead page that was available after the advanced setup page.
The finish button how ever should still have "Connect" as text.

swoichha:
> Sync root sharing was removed as it could cause data loss when virtual files where involved.
Okay. I'll make some changes so that it will use two different directory for syncing two different users.

swoichha:
[...]

PR #8898, opened 2021-08-13
Fix add accounts

swoichha:
### Description
This PR changes object name for connect button which we use while adding a new account.  Also, change have two sync directories for two different users. 

### Related Issue
- Fixes https://github.com/owncloud/client/issues/8897


test/gui/shared/scripts/bdd_hooks.py
```diff
@@ -48,6 +51,8 @@ def hook(context):
         context.userData['localBackendUrl'] = 'https://localhost:9200'
     if context.userData['clientSyncPath'] == '':
         context.userData['clientSyncPath'] = '/tmp/client-bdd/'
+    if context.userData['clientSyncPathUser'] == '':
+        context.userData['clientSyncPath'] = '/tmp/client-bdd-user/'
```
Talank, line 55: may be you should rename the folders to `client-bdd-user1` and `client-bdd-user2` 🤔

individual-it, line 55: I don't understand what that will help, it will still be the same folder that will be used in different tests.
Also it will be very confusing `clientSyncPath` or `clientSyncPathUser` 
Also line 54 is using `clientSyncPathUser` but 55 `clientSyncPath`
can we maybe have the user-name in the path name?

swoichha: Currently different user should have different sync path. So, for 2nd user we need a new `clientSyncPath` for now I have called it `clientSyncPathUser2` so that it will be less confusing.

swoichha:
Need to refactor tests to use `client_sync_path_user1` and `client_sync_path_user2` instead of `client_sync_path`

test/gui/shared/scripts/helpers/SetupClientHelper.py
```diff
@@ -28,6 +33,7 @@ def getClientDetails(context):
         elif row[0] == 'localfolder':
             localfolder = row[1]
         try:
+            print(localfolder)
```
Talank, line 36: debug line?

test/gui/shared/steps/steps.py
```diff
@@ -442,6 +443,7 @@ def openSharingDialog(context, resource, itemType='file'):
 @When('the user opens the public links dialog of "|any|" using the client-UI')
 def step(context, resource):
     resource = sanitizePath(substituteInLineCodes(context, resource))
+    print(resource)
```
Talank, line 446: I don't think this print is necessary 🤔

Talank (commented):
Clean up some print statements, everything else looks fine

test/gui/shared/steps/steps.py
```diff
@@ -421,6 +421,7 @@ def step(context):
 
 def openSharingDialog(context, resource, itemType='file'):
     resource = sanitizePath(substituteInLineCodes(context, resource))
+    print(resource)
```
Talank, line 424: remove this one as well.

Talank (approved):
lgtm

phil-davis (approved):

Issue #8910, opened 2021-08-18
[QA] Resizing of the columns in 'Activity' views doesn't work properly


gabi18:
Seen with 2.9.0beta1 (build 4915) on openSUSE 15.2

The resize behaviour of the in 'Server Activity', 'Sync Protocol' and 'Not Synced' tab is strange.
Changing the width of the last column 'Time' isn't possible directly but only when changing the size of the first column.

Size of 'Time' column doesn't change:
![time_column](https://user-images.githubusercontent.com/75983861/129852348-583af3f0-6d5a-4597-856b-ad46058eb9f7.png)

Width of 'Time' changes when changing 'Action'
![action_column](https://user-images.githubusercontent.com/75983861/129852666-0c882758-4217-4dbc-ac78-8e8036ddf915.png)

There was already an issue about column size before:
https://github.com/owncloud/client/issues/8660

michaelstingl:
You can resize the time-column, but it's not intuitive:
![CleanShot 2021-08-18 at 09 08 05](https://user-images.githubusercontent.com/214010/129853673-30586712-4231-461d-8f31-5d13f9084afe.gif)

gabi18:
I haven't tried that. Indeed, I'm also able to resize 'Time'.

TheOneRing:
The code responsible for the behaviour can be fund in https://github.com/owncloud/client/blob/master/src/gui/models/expandingheaderview.cpp#L33

erikjv:
What should the behavior be? That if you resize one column, it doesn't change the sizes of the other columns?

michaelstingl:
Maybe move the separator line between Account and Time, both should change, but all other keep their width.

michaelstingl:
macOS Finder cheats with an empty column on the right:

![CleanShot 2021-08-18 at 13 01 23](https://user-images.githubusercontent.com/214010/129887183-3fde2ed4-cd21-4b69-8e19-0484099b2918.gif)

jnweiger:
> What should the behavior be? That if you resize one column, it doesn't change the sizes of the other columns?

* The behaviour as show in @michaelstingl's post looks good.
* In general: dragging a column separator should change the width of the column directly left to it. Columns on the right side move in position, but do not change size. Nothing else changes.
* The last column is hard to change, if you cannot grab the column separator to its right. That may need special handling.

On Linux
* I see multiple columns changing size, when I drag one separator, that is cunter-intuitive.
[...]

PR #8914, opened 2021-08-18
Use a more intuitive resize behaviour

TheOneRing:
Fixes: #8910


erikjv (approved):

jnweiger:
Confirmed. This fixes #8910 -- That was quick!

One slight UX glitch:
* Enlarge the e.g. File column, so that a scrollbar appears at the bottom
* Scroll all the way to the right, so that the last column Time is again fully visible.
* Further enlarge a column (e.g. File again) 
* The Time Column  moves out of view, but the horizontal scrollbar stays at the right edge.
Expected behaviour: Scrollbar shrinks and has now space on both sides.

PR #8917, opened 2021-08-18
Refactor reserved name code

TheOneRing:
~The code also checked whether the strings started with COM8. and returned true in that cas.
While COM8 is reserved COM8.png is ok.~

@michaelstingl maybe only for 2.10?

### DIFF
--- a/src/csync/csync_exclude.cpp
+++ b/src/csync/csync_exclude.cpp
@@ -39,6 +39,26 @@
 #include <QFile>
 #include <QDir>
 
+namespace {
+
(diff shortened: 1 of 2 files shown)


dschmidt:
simplify?

michaelstingl:
> @michaelstingl maybe only for 2.10?

yes, sufficient

dschmidt:
Is this covered by tests?
otherwise should be easy enough to test them now while at it

TheOneRing:
So turns out COM8.png is still illegal on Windows 10

erikjv (approved):

Issue #8938, opened 2021-08-24
[QA] Sharing dialog (often) shows no thumbnails


jnweiger:
Seen with owncloud-client 2.9.0 beta3 on Linux connected to server 10.8.0

* also install owncloud-client-nemo
* wait until the sync is done. 
* navigate to the sync folder with nemo, right click several files, open the owncloud -> sharing dialog

In the examples shown here, the portugal image does not show a thumbnail. Actually, when opening the dialog, system default thumbnail is shown briefly, then disappears.
The bicycle image has its thumbnail shown here. (server or client generated, not sure....)
![image](https://user-images.githubusercontent.com/1108546/130603061-48747ac3-9bad-4f36-b673-32df9a03929f.png)
![image](https://user-images.githubusercontent.com/1108546/130603083-d9578c0f-1804-416b-9c92-bbfdf248409b.png)

michaelstingl:
Log available?

TheOneRing:
Core replies with an jpg instead of an png

bishwasojha:
sharing in nemo , Image thumbnail is shown as follows 

## In Linux

![Screenshot from 2022-11-09 11-45-15](https://user-images.githubusercontent.com/33860671/200751606-4719b028-5a59-4628-a3be-9c0b3e999133.png)
![Screenshot from 2022-11-09 11-51-32](https://user-images.githubusercontent.com/33860671/200752486-bb5afc96-b124-4787-ab56-16617d64a67a.png)
![Screenshot from 2022-11-09 11-42-56](https://user-images.githubusercontent.com/33860671/200751642-b5e0cc60-cd60-4aac-8202-83c535be6c65.png)

## In Windows

![Screenshot from 2022-11-09 11-42-03](https://user-images.githubusercontent.com/33860671/200751645-ddd92962-e09d-4a2f-81df-c7c8c67ab098.png)
![Screenshot from 2022-11-09 11-52-58](https://user-images.githubusercontent.com/33860671/200752734-bc4126f9-b285-4310-9676-3a404b86e2b7.png)

In Linux TESTED with ownCloud version
```
ownCloud 3.0.0.9005-beta1 [6b12ec](https://github.com/owncloud/client/commit/6b12ecdb4236d3d485ed57ec0bc2feac72b8f340)
Libraries Qt 5.15.2, OpenSSL 3.0.2 15 Mar 2022
Using virtual files plugin: suffix
ubuntu-5.15.0-52-generic
```


[...]

PR #8939, opened 2021-08-24
Sharedialog preview be more resilient

TheOneRing:
The api currently does respond with non png icons, determine the image type based on the content.
Don't override the system icon if we got a null icon.

Fixes: #8938


jnweiger:
Would that also accept an SVG thumbnail with some facy embedded JavaScript exploit?

TheOneRing:
QPixmap only accepts images no svgs.

dragotin (approved):
Nice!

PR #8942, opened 2021-08-24
Skip https check if oauthOverrideAuthUrl is used

TheOneRing:


src/libsync/creds/oauth.cpp
```diff
@@ -514,6 +514,12 @@ bool isUrlValid(const QUrl &url)
 {
     qCDebug(lcOauth()) << "Checking URL for validity:" << url;
 
+    // we have hardcoded the oauthOverrideAuth
+    const auto overrideUrl = Theme::instance()->oauthOverrideAuthUrl();
+    if (!overrideUrl.first.isEmpty()) {
+        return QUrl::fromUserInput(overrideUrl.first).matches(url, QUrl::RemoveQuery);
```
orgaralf, line 520: I guess the overrideUrl is a list and you take the first entry and match it to the type "url" ... now if that is oc-auth:// is that recognized as URL?

orgaralf (approved):
not sure, if oc-auth:// qualifies as "url" in the QUrl check, but I guess you know :-)

Issue #8932, opened 2021-08-23
Disrepancy regarding Virtual Files Suffix: CLI and Version Window


sahero-86:
![Auswahl_082](https://user-images.githubusercontent.com/80453811/130448919-7ad6aaaf-6631-4b74-93da-0ae54e78ca1e.png)

### Expected behaviour
The word suffix should be visible in both instances - CLI and in the Sync Client Version Tab. (see screenshot). Both are using the same Build Number.
### Actual behaviour
In the CLI the virtual files plugin is set to "off", however the sync process takes longer to complete. (I installed the sync client via CLI and wasn't asked about using the virtual files plug-in, so I would assume it to be labeled "off". While checking the sync process, we stumbled across the discrepancy in the name values.)
@jnweiger observed the same behaviour in Beta 2.

### Steps to reproduce
1. 
2. 
3. 

### Server configuration
Operating system:

Web server: 

Database:

PHP version:

ownCloud version:

Storage backend (external storage):

### Client configuration
Client version:

Operating system:

OS language:

Qt version used by client package (Linux only, see also Settings dialog):

Client package (From ownCloud or distro) (Linux only):

Installation path of client:

### Logs

Please use Gist (https://gist.github.com/) or a similar code paster for longer
logs.

```Template for output < 10 lines```

1. Client logfile: Output of `owncloud --logwindow` or `owncloud --logfile log.txt`
(On Windows using `cmd.exe`, you might need to first `cd` into the ownCloud directory)
(See also https://doc.owncloud.org/desktop/troubleshooting.html#log-files )

2. Web server error log:

3. Server logfile: ownCloud log (data/owncloud.log):

**Comment by TheOneRing** [comment]
We don't support virtual files in the cmd client and thus no plugins are loaded.

**Comment by michaelstingl** [comment]
@TheOneRing this is `owncloud`, not `owncloudcmd` ? 🤔

**Comment by jnweiger** [comment]
Confirmed here too with both owncloudcmd and owncloud:
```
owncloud --version
ownCloud ownCloud 2.9.0beta3 (build 5018) Aug 23 2021 15:02:54
https://github.com/owncloud/client/commit/405790d408e03af1dbd77d64178af384579039d4
Libraries Qt 5.12.10, OpenSSL 1.1.1f  31 Mar 2020
Using virtual files plugin: off
linuxmint-5.4.0-81-generic
```

But testpilotcloud-2.9.0-beta3.5019.AppImage shows `Using virtual files plugin: suffix`

PR #8943, opened 2021-08-24
Set OC_PLUGIN_DIR before we set the about text

TheOneRing:
Fixes: https://github.com/owncloud/client/issues/8932


dragotin (approved):
LGTM

Issue #8923, opened 2021-08-19
[QA] appimage update notification mentions the "system's update tool"


jnweiger:
Seen in testpilotcloud-2.9.0-beta1.4914.AppImage

![image](https://user-images.githubusercontent.com/1108546/130060779-c8093bfa-d97b-4e57-9d59-6f63dd2381f7.png)

This suggestion does not work. Appimages cannot e updated with the system's update tool.

Expected behaviour: 
* "... Please download the new appimage manually" (if possible mention the URL!) or
* best: have a builtin updater in the appimage

fmoc:
Closed by #9376.

fmoc:
The production update server has not been updated yet. It will be updated with the final release of 3.0.0.

individual-it:
@fmoc so this cannot really be tested, I will put this int `no test needed` 
CC @HanaGemela

michaelstingl:
With beta1 and rc's , production update server should be updated, but only for testpilotcloud builds. Vanilla updates after final release. But QA team has it's own updater server, to tests independently from the production one.

fmoc:
@individual-it testpilotcloud should work indeed. And you can set up your own update server, too, if you like to. I use https://github.com/fmoc/minimal-update-server/.

saw-jan:
Update is there. Update pops up at the startup as well

![Screenshot from 2022-11-17 16-20-33](https://user-images.githubusercontent.com/52366632/202424027-09040702-1df6-441b-a568-8ff9bac0f0c8.png)

saw-jan:
There is no update from RC2 to RC3

CC @TheOneRing @fmoc

TheOneRing:
rc3 was not yet pushed to the updater

saw-jan:
> rc3 was not yet pushed to the updater

It works now!

PR #8966, opened 2021-08-27
Improve "update available" message in AppImages

fmoc:
Contributes to #8923. Not the nicest solution, but I didn't want to add another entry to the enum.

### DIFF
--- a/src/gui/updater/ocupdater.cpp
+++ b/src/gui/updater/ocupdater.cpp
@@ -153,6 +153,13 @@ QString OCUpdater::statusString() const
     case DownloadTimedOut:
         return tr("Could not check for new updates.");
     case UpdateOnlyAvailableThroughSystem:
+#ifdef Q_OS_LINUX
+        // https://docs.appimage.org/packaging-guide/environment-variables.html
+        // TODO: update once AppImageUpdate has been implemented
+        if (qEnvironmentVariableIsSet("APPIMAGE")) {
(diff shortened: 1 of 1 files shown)


src/gui/updater/ocupdater.cpp
```diff
@@ -153,6 +153,13 @@ QString OCUpdater::statusString() const
     case DownloadTimedOut:
         return tr("Could not check for new updates.");
     case UpdateOnlyAvailableThroughSystem:
+#ifdef Q_OS_LINUX
+        // https://docs.appimage.org/packaging-guide/environment-variables.html
+        // TODO: update once AppImageUpdate has been implemented
+        if (qgetenv("APPIMAGE") != nullptr) {
```
TheOneRing, line 159: qEnvIsSet

src/gui/updater/ocupdater.cpp
```diff
@@ -153,6 +153,13 @@ QString OCUpdater::statusString() const
     case DownloadTimedOut:
         return tr("Could not check for new updates.");
     case UpdateOnlyAvailableThroughSystem:
+#ifdef Q_OS_LINUX
+        // https://docs.appimage.org/packaging-guide/environment-variables.html
+        // TODO: update once AppImageUpdate has been implemented
+        if (qEnvironmentVariableIsSet("APPIMAGE")) {
+            return tr("New %1 available. Please click <a href='%2'>here</a> to download the new AppImage manually.").arg(updateVersion).arg(_updateInfo.web());
```
TheOneRing, line 160: Just saw it now, don't use arg().arg() but arg(1,2,3). Use QString::number() for numbers as the normal overload is 💩

PR #8972, opened 2021-09-02
[Tests-Only][CI] pin middleware commit id for running tests in CI

dpakach:
pin middleware commit id for GUI tests in CI

### DIFF
--- a/.drone.env
+++ b/.drone.env
@@ -1,0 +1,4 @@
+# To run the GUI tests we use owncloud-test-middleware(https://github.com/owncloud/owncloud-test-middleware)
+# Set the commit id to use the correct version of the middleware
+
+MIDDLEWARE_COMMITID=7adeb56a999a70aa6108cbbcc70dc37071cbc1f9
--- a/.drone.star
+++ b/.drone.star
@@ -489,8 +489,10 @@ def testMiddleware():
             "BACKEND_HOST": "http://owncloud",
         },
         "commands": [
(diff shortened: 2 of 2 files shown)


phil-davis (commented):
Having this in `.drone.star` caused hassles in the past with PRs that were a bit out-of-date from `master`. The out-of-date PR "sees" only the `.drone.star` from its own branch.

Can we do the same as in other repos - have `.drone.env` ?

phil-davis (approved):

swoichha (commented):
lgtm

Talank (commented):
The changes look good to me. However, there seem to be a lot of CI failures. I have noticed that several PRs in client are failing that have nothing to do with the changes in the PR. Need to discuss how to fix it.

swoichha:
> The changes look good to me. However, there seem to be a lot of CI failures. I have noticed that several PRs in client are failing that have nothing to do with the changes in the PR. Need to discuss how to fix it.

The error that keeps happening in CI for this PR is `2021-09-02T09:03:41	ERROR     	/usr/lib/python3.8/urllib/request.py:1353: Script Error	<urlopen error [Errno -2] Name or service not known>`. It occurs before creating usesr on the server like `Given user "Alice" has been created on the server with default attributes and without skeleton files` which is done using middleware. So, IMO we need to look into how we create users in middleware. It can be a starting point.

PR #8978, opened 2021-09-06
Warn on unsupported legacy configuration

TheOneRing:
image

### DIFF
--- a/src/gui/folderman.cpp
+++ b/src/gui/folderman.cpp
@@ -36,8 +36,16 @@
 #include <QSet>
 #include <QNetworkProxy>
 
-static const char versionC[] = "version";
-static const int maxFoldersVersion = 1;
+namespace {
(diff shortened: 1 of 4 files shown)


src/gui/folderman.cpp
```diff
@@ -36,8 +36,16 @@
 #include <QSet>
 #include <QNetworkProxy>
 
-static const char versionC[] = "version";
-static const int maxFoldersVersion = 1;
+namespace {
+const char versionC[] = "version";
+const int maxFoldersVersion = 1;
+
+int numberOfSnycJurnals(const QString &path)
```
erikjv, line 43: An o missing?

erikjv (approved):

src/gui/folderman.cpp
dschmidt: And Snyc 👯

Issue #8806, opened 2021-07-06
Using the language selector to select an RTL language on mac does not cause the ui to be reversed


TheOneRing:
![image](https://user-images.githubusercontent.com/200626/124591641-d2cdeb00-de5c-11eb-82a4-34cf5364fbe2.png)

TheOneRing:
Maybe because we call the translations code AFTER we initialized QApplication?

fmoc:
I think we just need to call `setLayoutDirection` in `setupTranslations`, once we can tell whether we have an RTL language or not. https://doc.qt.io/qt-5/qguiapplication.html#layoutDirection-prop

fmoc:
I cannot reproduce this issue. You didn't reveal which locale you were testing against. I selected a random right-to-left-looking language, and without a code change the layout was changed:

![grafik](https://user-images.githubusercontent.com/80399010/125012905-7801e280-e05a-11eb-9ee4-54080f74d206.png)

I can also control the layout direction using `setLayoutDirection`. We could add the following snippet to `setupTranslations` nevertheless, if you want to:

```cpp
            // use a right-to-left layout for such languages
            QApplication::setLayoutDirection(QLocale(lang).textDirection());
```

TheOneRing:
I don't think we should call it directly this should be handled by the translations system.
The issue seems to only occur on mac.
I could imagin it being caused by us setting up the translations in the QApplication constructor....
@erikjv

erikjv:
Hmm, Qt has this in `QGuiApplicationPrivate::init()`:
```C++
    if (layout_direction == Qt::LayoutDirectionAuto || force_reverse)
        QGuiApplication::setLayoutDirection(qt_detectRTLLanguage() ? Qt::RightToLeft : Qt::LeftToRight);
```
That in turn is called from the `QApplication` constructor.

That said, I assume you're not running with the `-reverse` cmdline option, right? And your translations have the following translated:
```C++
        (QGuiApplication::tr("QT_LAYOUT_DIRECTION",
                         "Translate this string to the string 'LTR' in left-to-right"
                         " languages or to 'RTL' in right-to-left languages (such as Hebrew"
                         " and Arabic) to get proper widget layout.") == QLatin1String("RTL"));
```
Correct?

gabi18:
Remaining problem with systray menu?

[...]

PR #8981, opened 2021-09-07
Explicitly set the layout direction if changing the locale

TheOneRing:
Can you see side effects of the patch?
The issue does not occur with Qt 5.15 so maybe I should ifdef it
Fixes: #8806

### DIFF
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -758,6 +758,10 @@ void Application::setupTranslations()
                 QLocale newLocale(lang);
                 qCDebug(lcApplication) << "language" << lang << "was enforced, changing default locale to" << newLocale;
                 QLocale::setDefault(newLocale);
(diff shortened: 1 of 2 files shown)


src/gui/application.cpp
```diff
@@ -758,6 +758,8 @@ void Application::setupTranslations()
                 QLocale newLocale(lang);
                 qCDebug(lcApplication) << "language" << lang << "was enforced, changing default locale to" << newLocale;
                 QLocale::setDefault(newLocale);
+                // setting the layout direction directly only appears to be needed on mac
```
erikjv, line 761: IF this doesn't occur on 5.15, then I'd put it in an `if (QT_VERSION < QT_VERSION_CHECK(5, 15, 0))` statement.

erikjv (approved):

Issue #8899, opened 2021-08-13
[CI] Run all GUI test in a single pipeline


swoichha:
### Description
Currently, we run GUI smoke test in one pipeline and remaining tests in another pipeline. GUI tests is about 25-30 mins. Once the smoke test pipeline is completed then only the other tests will start to run. While doing so each pipeline builds client separately.

We need to find a way to save a significant amount of build time so that the client is build only once. 

### Possible-Fix:
- Maybe we can combine smoke and non-smoke tests in a single pipeline

TheOneRing:
I assume combining them in a single pipeline and just fail hard if the smoke tests fail would be perfect.

swoichha:
Fixed in https://github.com/owncloud/client/pull/8987 so closing this issue.

Issue #8971, opened 2021-09-01
[QA] Automation test for Desktop Client 2.9


swoichha:
### Automation test for Desktop Client

[Test Plan Template](https://github.com/owncloud/client/issues/8870)  This issue has a list of Testplan Release Desktop 2.9

### Background
While doing manual testing for Desktop 2.9 we came across many tests that can be automated. In this issue are the list of test that can and should be automated.

 ----
* [x]  1. Install/Update - Connect - Login @swoichha 
-> https://github.com/owncloud/client/issues/8971#issuecomment-909884138
* [x] 2. Folders @Talank
->https://github.com/owncloud/client/issues/8971#issuecomment-909884257
* [ ] 3. Files   @Talank
-> https://github.com/owncloud/client/issues/8971#issuecomment-909884348
* [x] 4. ~~Move files and folders~~  @swoichha 
-> https://github.com/owncloud/client/issues/8971#issuecomment-909884563
* [x] 5. Edit Files @swoichha 
-> https://github.com/owncloud/client/issues/8971#issuecomment-909884684
* [x] 6. Delete Files and Folders  @swoichha 
-> https://github.com/owncloud/client/issues/8971#issuecomment-909884844
* [x] 7. Sync process  
-> https://github.com/owncloud/client/issues/8971#issuecomment-909884918
* [x] 8. ~~Without connection~~ @swoichha 
-> https://github.com/owncloud/client/issues/8971#issuecomment-909884992
* [x] 9. Sharing 
* [x] 9.1 Share link @swoichha 
-> https://github.com/owncloud/client/issues/8971#issuecomment-909885070
* [x] 9.2 Share with Users and Groups @saw-jan
-> https://github.com/owncloud/client/issues/8971#issuecomment-909885166
* [x] 10. Selective_Sync @saw-jan 
* [x] 10.1 Choose what to sync from the setup window 
-> https://github.com/owncloud/client/issues/8971#issuecomment-909885292
* [x] 10.2 "Skip folders configuration"  
* [x] 10.3 "Add Folder..." @swoichha 
-> https://github.com/owncloud/client/issues/8971#issuecomment-909885541

swoichha:
### 1. Install/Update - Connect - Login
Existing squish tests concerning this topic: tst_addAccount (also tests multiple accounts), tst_loginLogout, tst_removeAccountConnection

* For this test set, choose a random platform win/lin/mac, unless test items specify otherwise.
* Record which platform was chosen: ...

[...]

PR #8987, opened 2021-09-08
[tests-only] add tests for connect-login

swoichha:
### Description
Added automated tests for Connect - Login user.
Also in this PR we are running all tests in a single pipeline as suggested in https://github.com/owncloud/client/issues/8899

### Related Issues
- Part of https://github.com/owncloud/client/issues/8971#issuecomment-909884138
- Fixes https://github.com/owncloud/client/issues/8899


.drone.star
```diff
@@ -41,8 +45,8 @@ def main(ctx):
             "Ninja",
             trigger = build_trigger,
         ),
-        gui_tests(ctx, trigger = build_trigger, filterTags = ["@smokeTest"], version = "latest"),
-        gui_tests(ctx, trigger = build_trigger, depends_on = ["GUI-tests-@smokeTest"], filterTags = ["~@smokeTest"], version = "latest"),
+        gui_tests(ctx, trigger = build_trigger, version = "latest"),
+        #gui_tests(ctx, trigger = build_trigger, filterTags = ["@sslCertTest"], version = "latest"),
```
Talank, line 49: Why did you comment this job? Do we need it when all the tests are run in earlier job?

swoichha: No we don't. I'll remove this in next commit.

.drone.star
```diff
@@ -224,6 +226,7 @@ def gui_tests(ctx, trigger = {}, depends_on = [], filterTags = [], version = "da
                              "CLIENT_REPO": "/drone/src/",
                              "MIDDLEWARE_URL": "http://testmiddleware:3000/",
                              "BACKEND_HOST": "http://owncloud/",
+                             "SECURE_BACKEND_HOST": "https://owncloud/",
```
Talank, line 229: do we need to update this documentation for an additional environment? https://doc.owncloud.com/desktop/2.9/appendices/guitest.html#run-tests-using-docker

test/gui/shared/scripts/pageObjects/AccountConnectionWizard.py
```diff
@@ -25,34 +25,57 @@ class AccountConnectionWizard:
         "visible": 1,
         "window": names.error_QMessageBox,
     }
+    URL_LOCKED = {
+        "name": "urlLabel",
+        "type": "QLabel",
+        "visible": 1,
+        "window": names.owncloudWizard_OCC_OwncloudWizard,
+    }
```
saw-jan, line 33: `unlocked` and `locked` icon has the same name

test/gui/shared/scripts/pageObjects/AccountConnectionWizard.py
```diff
@@ -25,34 +25,57 @@ class AccountConnectionWizard:
...
+        )
         squish.clickButton(squish.waitForObject(self.CHOOSE_BUTTON))
         test.compare(
             str(squish.waitForObjectExists(self.SELECT_LOCAL_FOLDER).text),
-            self.sanitizeFolderPath(localfolder),
+            self.sanitizeFolderPath(clientDetails['localfolder']),
         )
         squish.clickButton(squish.waitForObject(self.FINISH_BUTTON))
+
+    def isUrlLocked(self):
```
saw-jan, line 80: this function seems to check the presence of an element. So better to rename the function to something else.
`isUrlLocked` is kind of misleading

test/gui/shared/steps/steps.py
```diff
@@ -806,3 +826,21 @@ def step(context):
...
+@When('the user accepts the certificate')
+def step(context):
+    clickButton(waitForObject(names.oCC_SslErrorDialog_cbTrustConnect_QCheckBox))
+    clickButton(waitForObject(names.oCC_SslErrorDialog_OK_QPushButton))
+
+
+@Then('the lock shown should be closed')
+def step(context):
+    newAccount = AccountConnectionWizard()
+    newAccount.isUrlLocked()
```
saw-jan, line 840: good to have assertion here

test/gui/tst_addAccount/test.feature
```diff
@@ -5,26 +5,45 @@ Feature: adding accounts
...
+
+    Scenario: Adding account with wrong credentials
+        Given the user has started the client
+        When the user adds the account with wrong credentials
+            | server   | %local_server% |
+            | user     | Alice          |
+            | password | 12345          |
+        Then error "The provided credentials are not correct" should be displayed
+
+    @sslCertTest
```
saw-jan, line 41: I think this tag can be removed.

.drone.star
swoichha: I guess we should

test/gui/shared/scripts/pageObjects/AccountConnectionWizard.py
swoichha: Since the element is same for both locked and unlocked icon I have used a vp to check for a locked icon.

test/gui/tst_addAccount/test.feature
```diff
@@ -5,26 +5,45 @@ Feature: adding accounts
...
+        When the user adds the account with wrong credentials
+            | server   | %local_server% |
+            | user     | Alice          |
+            | password | 12345          |
+        Then error "The provided credentials are not correct" should be displayed
+
+
+    Scenario: Adding account with self signed certificate for the first time
+        Given the user has started the client
+        When the user adds an account with the following secure server address
```
saw-jan, line 44: we can reuse it for unsecure server too.
```suggestion
        When the user adds an account with the following server address
```

swoichha: For unsecure server we generally enter other values like username and password along with the url and then click on next. But this step just enters the url so I think it is better this way.

saw-jan: Agreed!! But we might need a step only to add server url (no matter secure or unsecure) and then click next like here.
So if you have a step like `the user adds the following server address` (enter url and hit next) then this can be reused.

swoichha: Yes we can do that but I think it is not needed here in this PR maybe we can make an issue to refactor that

saw-jan (commented):
LGTM

Talank (approved):
LGTM

PR #9017, opened 2021-09-13
Fix crash on missing sync root

TheOneRing:
The change means that vfs is not initialised so I need to implement a dummy vfs or better add checks for every use of vfs()....

Fixes: #9016


erikjv (commented):
Looks ok to me.

erikjv (approved):

PR #9018, opened 2021-09-14
[tests-only] Delete and gitignore local config.ini

saw-jan:
gitignoring local `config.ini` of `tests/gui` and added `config.ini` for drone-ci.
As of now, the configs set through environment variables have the first priority, and then the configs in config.ini

[DISCUSSION]
- Should we only use configs from `config.ini` and not from `env`? (locally and CI)


test/gui/config.sample.ini
```diff
@@ -1,6 +1,7 @@
 [DEFAULT]
-CLIENT_SYNC_PATH=
-BACKEND_HOST=
-CLIENT_SYNC_TIMEOUT=
-MIDDLEWARE_URL=
-CLIENT_LOG_FILE=
+CLIENT_SYNC_PATH_USER1=/tmp/client-bdd-user1/
```
individual-it, line 2: no need to set this values in the sample ini file, those are anyway the defaults
the ini file should also work if no setting is set, because they are taken from `gui/shared/scripts/bdd_hooks.py `

test/gui/shared/scripts/bdd_hooks.py
```diff
@@ -18,14 +18,23 @@
...
 def hook(context):
     from configparser import ConfigParser
 
     cfg = ConfigParser()
-    cfg.read('../config.ini')
+
+    if os.environ.get('CI') in ['True', 'true']:
+        cfg.read('../drone/config.ini')
+    else:
+        cfg.read('../config.ini')
```
individual-it, line 33: don't do this here, if you want to use a specific ini file for drone, just copy it to `test/gui/config.ini` using the drone.star file

individual-it (changes requested):
I think overall I would leave the configuration in the env variables when using drone, then its all in one place.

test/gui/shared/scripts/bdd_hooks.py
```diff
@@ -64,7 +65,7 @@ def hook(context):
     if context.userData['clientSyncTimeout'] == '':
         context.userData['clientSyncTimeout'] = 60
     else:
-        context.userData['clientSyncTimeout'] = int(
+        context.userData['clientSyncTimeout'] = builtins.int(
```
saw-jan, line 68: Squish's `int` will throw an error if `CLIENT_SYNC_TIMEOUT` is set in config.ini
```
No matching 'int(str)' overload found:
```

individual-it (approved):

swoichha (approved):
lgtm

Talank:
e2e test is failing because of the following error:
```json
{
"Error Details": "LookupError: Lost connection to AUT",
"Feature": "remove account connection",
"Feature File": "tst_removeAccountConnection",
"Scenario": "remove an account connection",
"Test Step": "Then an account should be displayed with the displayname Alice Hansen and host %local_server_hostname%"
}
```
I think we have seen this error before 🤔

swoichha:
> e2e test is failing because of the following error:
> 
> ```json
> {
> "Error Details": "LookupError: Lost connection to AUT",
> "Feature": "remove account connection",
> "Feature File": "tst_removeAccountConnection",
> "Scenario": "remove an account connection",
> "Test Step": "Then an account should be displayed with the displayname Alice Hansen and host %local_server_hostname%"
> }
> ```
> 
> I think we have seen this error before thinking

We discussed this yesterday. It is cause by AUT crash.

PR #9020, opened 2021-09-14
[Tests-Only]Fixed GUI test log parser

Talank:
This fixes GUI test log parser to display the log from middleware steps fails as well.

### DIFF
--- a/test/gui/TestLogParser.py
+++ b/test/gui/TestLogParser.py
@@ -17,33 +17,53 @@ def traverse_loop(data):
             for scenario in feature['tests']:
 
                 # If the scenario is not skipped, then loop through all the steps in the scenario
-                if "tests" not in scenario:
-                    break
-                for test_step in scenario['tests']:
+                if "tests" in scenario:
 
-                    # If the test step fails then it contains further "tests" object
-                    # So loop through all the errors in the test step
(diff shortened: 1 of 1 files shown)


saw-jan (commented):
LGTM

phil-davis (approved):

Talank:
It looks like the log file does not contain the error details in case of run time errors. So, I will display the entire error object in that case. If we find another appropriate method to display the error precisely then we can make changes to that. https://drone.owncloud.com/owncloud/client/8944/6/14 
Screenshot from 2021-09-15 10-27-29
In the above case, the parser fails to find the error since it can not find the keyword `detail` in the error object.

test/gui/TestLogParser.py
```diff
@@ -56,7 +76,7 @@ def traverse_loop(data):
 
 # Remove duplicate nodes, if exists
 # This step is neccessary because sometimes the data in failing_tests_raw is redundent.
```
phil-davis, line 78:
```suggestion
# This step is necessary because sometimes the data in failing_tests_raw is redundant.
```

phil-davis (approved):
LGTM. If you need to rebase/change this PR again, then you can fix the typos in the existing comment.

Issue #9000, opened 2021-09-10
Allow to filter issue table by issue type


TheOneRing:
Add the ability to filter by issue type to the context menu of the issue type column header.

TheOneRing:
https://github.com/owncloud/client/pull/9010 ads the status to the table.
Filtering is still missing.

erikjv:
As far as I can see, `OCC::Models::displayFilterDialog` is used in two places. Do both need filtering by status?

The context menu is the same no matter which column you click, so I assume we need to show both the account filter, and the status filter. Also both filters should be usable independent of each other (e.g. "show all excluded for admin@")?

Or should there be only 1 active filter, depending on which column you click, that resets the filter for the other column?

The filtering can be done by a subclass of `QSortFilterProxyModel`, which asks for the `UnderlyingDataRole` of the item, and filters based on that.

TheOneRing:
I think the filter only makes sense on the issue widget, the column is hidden on the protocol view.
It might come handy to allow to filter by account and status?
Whether we want to allow multiple accounts/states as filter, depends on the complexity of the solution.

PR #9023, opened 2021-09-14
Allow to filter issue table by issue type

erikjv:
Fixes: https://github.com/owncloud/client/issues/9000

### DIFF
--- a/src/gui/activitywidget.cpp
+++ b/src/gui/activitywidget.cpp
@@ -69,6 +69,10 @@ ActivityWidget::ActivityWidget(QWidget *parent)
     header->setSectionResizeMode(QHeaderView::Interactive);
     header->setSortIndicator(static_cast<int>(ActivityListModel::ActivityRole::PointInTime), Qt::DescendingOrder);
 
+    connect(_ui->_filterButton, &QAbstractButton::clicked, this, [this] {
+        ProtocolWidget::showFilterMenu(_ui->_filterButton, _sortModel);
+    });
+
     _ui->_notifyLabel->hide();
     _ui->_notifyScroll->hide();
 
@@ -102,9 +106,8 @@ ActivityWidget::ActivityWidget(QWidget *parent)
     connect(_ui->_activityList, &QListView::customContextMenuRequested, this, &ActivityWidget::slotItemContextMenu);
     header->setContextMenuPolicy(Qt::CustomContextMenu);
     connect(header, &QListView::customContextMenuRequested, header, [header, this] {
-        auto menu = Models::displayFilterDialog(AccountManager::instance()->accountNames(), _sortModel, static_cast<int>(ActivityListModel::ActivityRole::Account), Qt::DisplayRole, this);
-        menu->addSeparator();
-        menu->addAction(tr("Reset column sizes"), this, [header] { header->resizeColumns(true); });
+        auto menu = ProtocolWidget::showFilterMenu(header, _sortModel);
+        header->addResetActionToMenu(menu);
     });
 
     connect(&_removeTimer, &QTimer::timeout, this, &ActivityWidget::slotCheckToCleanWidgets);
--- a/src/gui/issueswidget.cpp
+++ b/src/gui/issueswidget.cpp
@@ -24,7 +24,6 @@
 #include "folderman.h"
 #include "syncfileitem.h"
 #include "folder.h"
-#include "models/expandingheaderview.h"
 #include "models/models.h"
 #include "openfilemanager.h"
 #include "protocolwidget.h"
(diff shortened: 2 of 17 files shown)


erikjv:
**PLEASE TEST BEFORE MERGING** :-)

erikjv:
Oh and please change the target branch, it should be 2.9. I cannot find a way to do that...

fmoc (changes requested):
I failed to choose the right checkbox...

Please fix the test runs.

erikjv:
Note: the first patch is the same as https://github.com/owncloud/client/pull/9033 : I had to cherry-pick that into this branch so I can put the "real" patch on top of it.

TheOneRing:
```
09-17 12:41:52:557 [ warning default ]:	QSortFilterProxyModel: index from wrong model passed to mapToSource
09-17 12:41:52:557 [ fatal default ]:	ASSERT: "!"QSortFilterProxyModel: index from wrong model passed to mapToSource"" in file C:\_\17a9f6ae\qtbase-everywhere-src-5.12.10\src\corelib\itemmodels\qsortfilterproxymodel.cpp, line 548

File: src/gui/issueswidget.cpp [hunk 1]
```diff
@@ -50,6 +50,59 @@ bool persistsUntilLocalDiscovery(const OCC::ProtocolItem &data)
...
+        if (_filter.isEmpty()) {
+            return true;
+        }
+
+        QModelIndex idx = sourceModel()->index(sourceRow, filterKeyColumn(), sourceParent);
+        return _filter.contains(static_cast<SyncFileItem::Status>(data(idx, filterRole()).toInt()));
+    }
+
+private:
+    QSet<SyncFileItem::Status> _filter;
```
TheOneRing, line 85: Could you see an issue in converting the enum to QFlags? It would simplify the filter by a lot.

erikjv: A comment at the enum states that it fits in 4 bits, so I thought that turning that into flags would break something...

TheOneRing: Ah yes the 4bit optimisation ......

erikjv: So should I change it or not?

TheOneRing:
Looks and works nice 
image

Please remove success as we will never have an item with status success in the issue tab

src/gui/activitywidget.cpp
```diff
@@ -69,6 +69,10 @@ ActivityWidget::ActivityWidget(QWidget *parent)
     header->setSectionResizeMode(QHeaderView::Interactive);
     header->setSortIndicator(static_cast<int>(ActivityListModel::ActivityRole::PointInTime), Qt::DescendingOrder);
 
+    connect(_ui->_filterButton, &QAbstractButton::clicked, this, [this] {
```
TheOneRing, line 72: Hmm please move to  a lambda/function and unify with the context menu

src/gui/issueswidget.cpp
```diff
@@ -50,6 +50,56 @@ bool persistsUntilLocalDiscovery(const OCC::ProtocolItem &data)
...
+    {
+        if (_filter != newFilter) {
+            _filter = newFilter;
+            invalidateFilter();
+        }
+    }
+
+    void resetFilter()
+    {
+        setFilter({ false, true, true, true, true, true, true, true, true, true, true });
```
TheOneRing, line 83:
```
.fill(true);
````
?

File: src/gui/issueswidget.cpp [hunk 4]
```diff
@@ -200,4 +261,72 @@ void IssuesWidget::slotItemContextMenu()
     ProtocolWidget::showContextMenu(this, _model, rows);
```
TheOneRing, line 261: I can't comment line 259 but its still using the sort model instead of the filter model.
Still crashing when right clicking an issue.
```
09-28 13:32:31:040 [ warning default ]:	QSortFilterProxyModel: index from wrong model passed to mapToSource
09-28 13:32:31:040 [ fatal default ]:	ASSERT: "!"QSortFilterProxyModel: index from wrong model passed to mapToSource"" in file C:\_\17a9f6ae\qtbase-everywhere-src-5.12.10\src\corelib\itemmodels\qsortfilterproxymodel.cpp, line 548

File: src/gui/issueswidget.cpp [hunk 5]
```diff
@@ -200,4 +261,72 @@ void IssuesWidget::slotItemContextMenu()
...
+
+    // Add action to reset all filters at once:
+    menu->addAction(QCoreApplication::translate("OCC::Models", "Reset Filters"), [statusFilterGroup, this, menu]() {
+        for (QAction *action : statusFilterGroup->actions()) {
+            action->setChecked(true);
+        }
+        _statusSortModel->resetFilter();
+
+        for (QAction *action : menu->actions()) {
+            if (action->objectName() == Models::NoFilterObjectName) {
```
TheOneRing, line 324: Not optimal but I can't come up by something better atm, can you add a todo?

also
````
QObject::findChild
````

File: src/gui/protocolwidget.cpp [hunk 6]
```diff
@@ -77,11 +83,13 @@ ProtocolWidget::~ProtocolWidget()
     delete _ui;
 }
 
-void ProtocolWidget::showHeaderContextMenu(ExpandingHeaderView *header, QSortFilterProxyModel *model)
+void ProtocolWidget::showHeaderContextMenu(ExpandingHeaderView *header)
 {
```
TheOneRing, line 87: Please re add the filter menu, make sure to not duplicate the code.

src/libsync/syncfileitem.h
```diff
@@ -94,6 +94,7 @@ class OWNCLOUDSYNC_EXPORT SyncFileItem
         Excluded
     };
     Q_ENUM(Status)
+    static constexpr int StatusCount = Excluded + 1;
```
TheOneRing, line 97: I prefer to have the count in the enum it self. Yes that means we would need to touch the switch statements where the status is used. This would also allow us to drop cases where default: is define and ensure we handle all cases.

TheOneRing (changes requested):

TheOneRing (approved):
Yay thx

PR #9025, opened 2021-09-15
Fix build with libcloudproviders 2.9

### LINKED ISSUE #8996: 2.9.0 build with libcloudproviders is broken
### Expected behaviour
It builds

### Actual behaviour
```
/build/owncloud-client/src/client/src/gui/libcloudproviders/libcloudproviders.cpp:85:52: error: ‘class OCC::FolderMan’ has no member named ‘list’
   85 |     const auto newFolders = FolderMan::instance()->list();
      |                                                    ^~~~
```

`FolderMan::list()` was removed in commit 3f5163a4f30cf5978ffd7ee363e0d81065eb29c0

dragotin:
This will fix #8996 

This PR is not yet mergeable because the build of the test FolderManTest fails if libcloudproviders is enabled. There needs to be some cmake magic to add the libcloudproviders.cpp to the source list of the build of the test binary (I guess).


TheOneRing:
@dschmidt, @fmoc  could you add the requirements for libcloud providers to our ci...
And when you are on it, can you also install the lnguist?

fmoc:
I don't see any special dependences in https://gitlab.gnome.org/World/libcloudproviders/-/blob/master/meson.build. The libc minimum dependency could be an issue.

I'm looking into adding the linguist binary to the CI.

TheOneRing (changes requested):
That looks a bit invasive, I'll have a look we we need to explicitly compile that code in....

erikjv:
I can run/test this on Ubuntu with Gnome, but could this please be rebased onto the current 2.9 branch?

The reason is that the 2.9 version that this patch is based on, has a couple of bugs that are now fixed. But both rebasing this patch, or cherry-picking the fixes on top of this patch, leads to merge conflicts. And the unpatched/"unfixed" version is kinda broken.

TheOneRing:
Merged in https://github.com/owncloud/client/pull/9088

PR #9033, opened 2021-09-16
Work around compiler bug in old GCC versions

erikjv:
The previous use of template specialisation triggered a compiler bug on
Debian 9 and CentOS 7, which use gcc 5.4 or earlier. As there is just
one specialisation, the work-around is to change the function to be a
static member function of the wrapping class of the enum.

See also: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=56480

### DIFF
--- a/src/common/utility.h
+++ b/src/common/utility.h
@@ -273,12 +273,6 @@ OCSYNC_EXPORT Q_DECLARE_LOGGING_CATEGORY(lcUtility)
     {
         return QMetaEnum::fromType<E>().valueToKeys(value);
     }
-    template <class E = void>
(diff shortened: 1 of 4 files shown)


fmoc (commented):
Well, I think there's a reason this was initially declared as a template. Of course, you can just declare a regular function, which will then avoid the compiler bug.

@TheOneRing, is this okay for you?

TheOneRing:
I don't like it but well, lets merge it and revert it on 2.10.
In 2.10 we should try to get a newer compiler.

TheOneRing (approved):

PR #9036, opened 2021-09-16
Fix some compiler warnings

erikjv:
- comparison of integers of different signs
- usused parameter
- loop variable creates a copy
- unused variable (due to #ifdef)


src/csync/csync_exclude.cpp
```diff
@@ -123,12 +123,12 @@ OCSYNC_EXPORT bool csync_is_windows_reserved_word(const QStringRef &filename)
         return a.size() < b.size();
     }));
     for (const auto &word : win_device_names) {
-        if (word.size() > len_filename) {
+        if (size_t(word.size()) > len_filename) {
```
TheOneRing, line 126: Please use static casts

TheOneRing (approved):

PR #9038, opened 2021-09-16
Used make_unique instead of direct unique_ptr construction

alexeylysenko:
Use make_unique to create an object that isn’t shared (at least not yet), unless you need a custom deleter or are adopting a raw pointer from elsewhere.

### DIFF
--- a/src/libsync/configfile.cpp
+++ b/src/libsync/configfile.cpp
@@ -846,9 +846,9 @@ void ConfigFile::setClientVersionString(const QString &version)
     settings.setValue(clientVersionC(), version);
 }
 
-std::unique_ptr<QSettings> ConfigFile::settingsWithGroup(const QString &group, QObject *parent)
+std::unique_ptr<QSettings> ConfigFile::settingsWithGroup(const QString &group)
 {
-    std::unique_ptr<QSettings> settings(new QSettings(ConfigFile::configFile(), QSettings::IniFormat, parent));
+    auto settings = std::make_unique<QSettings>(ConfigFile::configFile(), QSettings::IniFormat);
     settings->beginGroup(group);
     return settings;
 }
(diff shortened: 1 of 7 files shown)


src/libsync/configfile.cpp
```diff
@@ -846,9 +846,9 @@ void ConfigFile::setClientVersionString(const QString &version)
     settings.setValue(clientVersionC(), version);
 }
 
-std::unique_ptr<QSettings> ConfigFile::settingsWithGroup(const QString &group, QObject *parent)
+std::unique_ptr<QSettings> ConfigFile::settingsWithGroup(const QString &group)
 {
-    std::unique_ptr<QSettings> settings(new QSettings(ConfigFile::configFile(), QSettings::IniFormat, parent));
```
erikjv, line 851: Why did you remove the `parent` argument?

TheOneRing: I just wanted to ask the same thing

erikjv: Whoops, that parent might result in a double delete when the parent gets destroyed before the unique ptr's constructor is run.

TheOneRing: hmm right.

I hope we properly use the function and don't keep a pointer in some member...

erikjv: I'll check that later this morning.

erikjv: We're fine, and it's good to get rid of the parent argument.

TheOneRing (approved):

src/libsync/configfile.cpp
alexeylysenko: > I just wanted to ask the same thing

There was a mixed approach of object ownership: both parent and unique_ptr - this shouldn't be the case, they aren't interoperable. Double-delete problem will occur.

PR #9042, opened 2021-09-17
Revert "Work around compiler bug in old GCC versions"

TheOneRing:
We rather update the compiler on debian 9

This reverts commit 4acec7bfdeeb798cdc7f32d6286dc7de07e17ef1.

@fmoc you had a look into updating those?

### DIFF
--- a/src/common/utility.h
+++ b/src/common/utility.h
@@ -273,6 +273,12 @@ OCSYNC_EXPORT Q_DECLARE_LOGGING_CATEGORY(lcUtility)
     {
         return QMetaEnum::fromType<E>().valueToKeys(value);
     }
+    template <class E = void>
+    QString enumDisplayName(E)
+    {
+        static_assert(std::is_same<E, void>::value, "Not implemented");
+        Q_UNREACHABLE();
+    }
 }
 /** @} */ // \addtogroup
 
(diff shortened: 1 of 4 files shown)


fmoc:
As mentioned previously, updating on CentOS is relatively easy. Red Hat provides [devtoolset](https://www.softwarecollections.org/en/scls/rhscl/devtoolset-8/), which uses some very special compiler hacks to avoid the need of having a newer libc installed as well.

There is, AFAIK, nothing like this available for Debian. In the past, I tried to just build my own compiler, but this has always bumped the built binaries' libc dependency. Unlike CentOS with devtoolset, the binaries did not just run on systems that do not have devtoolset installed.

TheOneRing:
@michaelstingl can we drop debian 9 for 2.10 and focus on appimages for stuff like that?
debian 9 is eol in June 2022 I think.

For 2.10 I plan some more modernisation and old compilers might cause us more painn...

michaelstingl:
> @michaelstingl can we drop debian 9 for 2.10 and focus on appimages for stuff like that?
> debian 9 is eol in June 2022 I think.

Sure, if we support 10 & 11, we can deprecate 9. We don't need to wait for bistro eol, we have our own policy.

PR #9049, opened 2021-09-17
Small cleanup in socket api

TheOneRing:


src/gui/socketapi/socketapi.cpp
```diff
@@ -450,17 +450,30 @@ void SocketApi::slotUpdateFolderView(Folder *f)
...
+        } break;
+        case OCC::SyncResult::Undefined:
+            Q_FALLTHROUGH();
+        case OCC::SyncResult::NotYetStarted:
+            Q_FALLTHROUGH();
+        case OCC::SyncResult::SyncRunning:
+            Q_FALLTHROUGH();
+        case OCC::SyncResult::SyncAbortRequested:
             qCDebug(lcSocketApi) << "Not sending UPDATE_VIEW for" << f->alias() << "because status() is" << f->syncResult().status();
         }
```
erikjv, line 478: if you checked all cases, maybe a `default: Q_UNREACHABE();` ?

erikjv (approved):

src/gui/socketapi/socketapi.cpp
TheOneRing: I omit the default on purpose to get compiler warnings

TheOneRing: we should rather make unhandled cases Werror

PR #9058, opened 2021-09-20
Allow up to 20000 sync issues

TheOneRing:
People might have a suboptimal setup with lots of ignored files.
The historic maximum we displayed was 2000 issues, we now allow 20000, while we keep the old value for the protocol view.

The underlying data structure was selected with the protocol widget in mind, so many continual inserts.

The issue widget however resets the model on every single sync, deletes all issues associated with an account, to then fill in the new issues. We should try to create a more clever approach....

### DIFF
--- a/src/gui/protocolwidget.cpp
+++ b/src/gui/protocolwidget.cpp
@@ -48,7 +48,7 @@ ProtocolWidget::ProtocolWidget(QWidget *parent)
 
     connect(_ui->_tableView, &QTreeWidget::customContextMenuRequested, this, &ProtocolWidget::slotItemContextMenu);
 
-    _model = new ProtocolItemModel(this);
+    _model = new ProtocolItemModel(2000, false, this);
     _sortModel = new QSortFilterProxyModel(this);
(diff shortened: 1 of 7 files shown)


src/gui/protocolwidget.cpp
```diff
@@ -48,7 +48,7 @@ ProtocolWidget::ProtocolWidget(QWidget *parent)
 
     connect(_ui->_tableView, &QTreeWidget::customContextMenuRequested, this, &ProtocolWidget::slotItemContextMenu);
 
-    _model = new ProtocolItemModel(this);
+    _model = new ProtocolItemModel(2000, false, this);
```
erikjv, line 51: Shouldn't this be 20000 (so 1 extra zero)?

TheOneRing: In the protocol the last 2k should be enough with the sync issues however...

erikjv (approved):

src/gui/protocolwidget.cpp
erikjv: 👍

PR #9077, opened 2021-09-23
[tests-only] add tests for edit files

swoichha:
### Description
Added automated tests for editing file content.

### Related Issues
- Part of https://github.com/owncloud/client/issues/8971#issuecomment-909884684


test/gui/tst_editFiles/test.feature
```diff
@@ -0,0 +1,22 @@
...
+    So that I can modify and change file data
+
+
+    Background:
+        Given user "Alice" has been created on the server with default attributes and without skeleton files
+
+
+    Scenario Outline: Modify orignal content of a file
+		Given user "Alice" has uploaded file with content "ownCloud test text file 0" to "<filename>" on the server
+        And user "Alice" has set up a client with default settings
```
Talank, line 14: Can you fix the indentation?

test/gui/tst_editFiles/test.feature
```diff
@@ -0,0 +1,22 @@
+Feature: edit files
+
+    As a user
+    I want to be able edit the file content
```
saw-jan, line 4: :smiley: 
```suggestion
    I want to be able to edit the file content
```

Talank (commented):
LGTM

saw-jan (commented):
LGTM

Talank (approved):

test/gui/tst_editFiles/test.feature
```diff
@@ -0,0 +1,22 @@
...
+        Given user "Alice" has uploaded file with content "ownCloud test text file 0" to "<filename>" on the server
+        And user "Alice" has set up a client with default settings
+        When the user overwrites the file "<filename>" with content "overwrite ownCloud test text file"
+        Then as "Alice" the file "<filename>" on the server should have the content "overwrite ownCloud test text file"
+        Examples:
+            | filename     |
+            | textfile.txt |
+            | textfile.doc |
+            | textfile.xls |
+            | textfile.pdf |
```
individual-it, line 22: I don't think we need all these examples, they cost a lot of time and don't add value in my opinion.
It would be more interesting to see if special character names are uploaded correctly

individual-it (changes requested):

test/gui/tst_editFiles/test.feature
swoichha: All these examples has been taken from https://github.com/owncloud/client/issues/8971#issuecomment-909884684. Should I just keep the `txt` one and add a special character name so that we will only have 2 examples?

individual-it: to replicate what is tested in https://github.com/owncloud/client/issues/8971#issuecomment-909884684 we would need to upload real doc/xls/pdf files, then overwrite them with other real doc/xls/pdf files and compare the content by downloading it again and checking the content. Or we could check the checksums.
currently there are all text-files just with different file name extensions.
I would suggest to focus on special characters in the name rather than different content of the file.
@jnweiger @TheOneRing @gabi18 what do you thing? how important is it to test the different file-types?

Issue #8920, opened 2021-08-19
[QA] Link in modified "About" window does not work


jnweiger:
Seen in testpilotcloud-2.9.0-beta1.4914.AppImage on Linux Mint 20.1
* Open Settings, -> click About
* A window with version number and some text opens.
* In case the theme has the modified about option, there is a badly formatted URL
![image](https://user-images.githubusercontent.com/1108546/130056291-9544ae31-d404-4db5-aff9-939a3a0b4bd5.png)

----

The testpilot theme has

![image](https://user-images.githubusercontent.com/1108546/130056123-be35ed8b-8152-49f1-88c7-938c9b9b4b71.png)

![image](https://user-images.githubusercontent.com/1108546/130056094-98417595-7bea-49bc-8177-786501852c5c.png)

The theme.json file corresponding to these settings has:
![image](https://user-images.githubusercontent.com/1108546/130056066-73e1442b-c47b-4c0d-9962-72363226ee32.png)

gabi18:
Confirmed fixed on Windows 10 with client 2.9.1rc1

PR #9084, opened 2021-09-23
No longer try to concatenate defines

TheOneRing:
Requires regeneration of the theme
Fixes: #8920


src/gui/updater/updater.cpp
```diff
@@ -165,7 +164,7 @@ qint64 Updater::Helper::stringVersionToInt(const QString &version)
 
 QString Updater::clientVersion()
 {
-    return QString::fromLatin1(MIRALL_STRINGIFY(MIRALL_VERSION_FULL));
+    return QStringLiteral("%1.%2.%3").arg(QString::number(MIRALL_VERSION_MAJOR), QString::number(MIRALL_VERSION_MINOR), QString::number(MIRALL_VERSION_PATCH));
```
dschmidt, line 167: Ah, this *intentionally* removes the `BUILD_NUMBER`, right? Can you add a comment why it's not part of the version anymore or why it should not be readded?

dschmidt (approved):

src/gui/updater/updater.cpp
TheOneRing: I did not 😱

dschmidt (approved):

Issue #9093, opened 2021-09-27
Failed to dehydrate when read only flag is set


m-emelchenkov:
### Expected behaviour
Files should be dehydrated without issues.

### Actual behaviour
ownCloud client puts files in "Not synced" tab with error "Failed to dehydrate". The file left not deleted on file system.

### Steps to reproduce
1. Create any file (i.e. Hello.txt).
2. "Always keep on device" on it, then "Free up space" --> OK.
3. Change property of Hello.txt: set RO attribute.
4. Repeat step 2 --> error.

### Server configuration
Operating system:
Ubuntu 20.04.3

Web server: 
Official docker image.

Database:
Official docker image.

PHP version:
Official docker image.

ownCloud version:
10.8.0.4

Storage backend (external storage):
no, Docker embedded storage

### Client configuration
Client version:
2.9.0 build 5150

Operating system:
Windows 11 insider preview

OS language:
English

Installation path of client:
default

TheOneRing:
Which kind of read only?
See https://github.com/owncloud/client/issues/9070#issuecomment-924715043

The entry in the not synced tab can be right clicked, and copied, I'd need the full error message.

m-emelchenkov:
Unfortunately the error message is not very informational.
`Action         ,File                                                            ,Folder                        ,Size  ,Account             ,Time                ,
Failed to dehydrate C:/Users/Mikhail/ownCloud/Hello.txt,Hello.txt                                                       ,ownCloud                      ,      ,admin@x.x.x.x:8080,2021-09-28T11:06:10.804,
`

and from .woncloudsync.log:

`#=#=#=#=# Propagation starts 2021-09-28T08:10:15Z (last step: 97 msec , total: 97 msec) 
 | Hello.txt | SyncInstruction(CSYNC_INSTRUCTION_SYNC) | OCC::SyncFileItem::Down | | 1632816353 | d48b0e0625e347cb8666c83cff526157 | 3 | 00000721ocwv8a3d73ho | OCC::SyncFileItem::NormalError | Failed to dehydrate C:/Users/Mikhail/ownCloud/Hello.txt | 0 | 3 | 1632816353 |  | 
#=#=#=# Syncrun finished 2021-09-28T08:10:15Z (last step: 16 msec , total: 114 msec) 
`

I checked folder permissions, I even added Full Control for Users and Authenticated Users groups, just to test, and no, it doesn't change anything.

TheOneRing:
[...]

PR #9094, opened 2021-09-28
 Fix dehydration of read only files

TheOneRing:
This enables us to dehydrate read only files.
Side note the read only flag won't be re applied after the dehydration...

Please also review https://gitea.owncloud.services/client/client-plugin-vfs-win/pulls/33

Issue: https://github.com/owncloud/client/issues/9093


michaelstingl:
> Side note the read only flag won't be re applied after the dehydration...

Compare with OneDrive & iCloud for Win?

erikjv (approved):

PR #9096, opened 2021-09-28
CI: Use a docker container for centos

TheOneRing:
The change also skips two checks in order to fix running tests on docker


TheOneRing:
So far so good I'd say, now why are the unit tests failing....

erikjv (approved):

PR #9102, opened 2021-09-30
Ci: Better logging in squish test

TheOneRing:
I also tried to set `os.environ["QT_LOGGING_RULES"] = "sync.httplogger=true"`
But that seemed to have no effect on the test.
See https://doc.owncloud.com/desktop/2.9/appendices/troubleshooting.html for context.


individual-it (approved):

TheOneRing:
@Talank @swoichha regarding QT_LOGGING_RULES can we set it? an alternative would be to enable it in the cfg file.

swoichha:
> QT_LOGGING_RULES

I am not sure if it will work but we won't know for sure until we try.

TheOneRing:
Yes the way I tired it did not work, well I have no idea of squish :D

PR #9107, opened 2021-09-30
Enable c++17, this enables us to use std::filesystem

TheOneRing:


dragotin:
Hm, any idea when common distros like Ubuntu, Debian, openSUES started to ship C++ 17 enabled compilers? I see danger to break builds on older platforms.

dragotin:
Ok, it seems that the CI even builds on CentOS7 - which should be old enough. Never mind.

TheOneRing:
We require gcc9 and all platforms that won't allow us to update to a gcc9 version will need to use that centos app image.

shell_integration/windows/OCContextMenu/OCClientInterface.cpp
```diff
@@ -28,17 +28,22 @@
 #include <iterator>
 #include <unordered_set>
 
-// use std::min in gdiplus
-using namespace std;
+// gdiplus min/max
+#define min std::min
```
fmoc, line 32: Why not `using std::min` and `using std::max`? You put `using namespace std` below anyway, which has the same effect.

TheOneRing: If I use std from the beginning we get a clash in a legacy windows header with `struct byte` and std::byte or what ever crazyness.

TheOneRing: using std::min is much better thx, when handling such windows stuff I kind of start thinking in macros and defines...

fmoc (approved):

PR #9108, opened 2021-09-30
Work/fix deprecated api

erikjv:


src/common/utility.cpp
```diff
@@ -52,7 +52,13 @@
 #if defined(Q_OS_WIN)
 #include "utility_win.cpp"
 #elif defined(Q_OS_MAC)
-#include "utility_mac.cpp"
+// We cannot include utility_mac.mm, so it will be compiled separately. Below are the forward declarations.
+namespace OCC {
+void setupFavLink_private(const QString &folder);
```
TheOneRing, line 57: I guess we can declare them for all in a private header?
Maybe we can also get rid of those cpp includes...

src/common/utility_mac.mm
```diff
@@ -0,0 +1,213 @@
...
+        if (item)
+            CFRelease(item);
+    }
+    CFRelease(placesItems);
+    CFRelease(folderCFStr);
+    CFRelease(urlRef);
+}
+
+
+static Result<void, QString> writePlistToFile(NSString *plistFile, NSDictionary *plist)
```
TheOneRing, line 53: Below the declaration of the Result class is also a Error class.
Which might match the requirements better.

src/common/utility_mac.mm
```diff
@@ -0,0 +1,213 @@
...
+{
+    // Our deployment target is 10.12, but in the 10.13 SDK `writeToFile:atomically:` is deprecated. So let's future-proof this code:
+    if (@available(macOS 10.13, *)) {
+        NSError *error = nil;
+        if (![plist writeToURL:[NSURL fileURLWithPath:plistFile isDirectory:NO] error:&error]) {
+            return QString::fromNSString(error.localizedDescription);
+        }
+    } else {
```
TheOneRing, line 61: Maybe we should target 2.10 with this one, it would remove the requirement so support < 10.13

You could also switch to std::optional https://github.com/owncloud/client/pull/9107

src/common/utility_mac.mm
```diff
@@ -0,0 +1,213 @@
...
+        } else {
+            return QStringLiteral("Some error occurred, but it's old API that won't tell us what happened");
+        }
+    }
+}
+
+bool hasLaunchOnStartup_private(const QString &)
+{
+    @autoreleasepool {
+        NSFileManager *fileManager = [NSFileManager defaultManager];
```
TheOneRing, line 92: Thx those releases creeped me out

src/common/utility_mac.mm
```diff
@@ -46,84 +49,150 @@ void setupFavLink_private(const QString &folder)
...
-                    }
-                    CFRelease(itemUrlRef);
-                }
-            }
-            CFRelease(itemsArray);
-        };
-        CFRelease(loginItems);
+            // We're not an installed application, there is anoter executable mentioned in the plist file,
+            // and that executable seems to exist, so let's not touch the file. Normally we'd report this
+            // as an error, but 🤷‍
```
TheOneRing, line 190: We could at least log it :D

TheOneRing (changes requested):

TheOneRing:
As mentioned it could be the best to only apply the change to master.
(Also ci should be back)

TheOneRing:
For the mac stuff, I have to believe you :D

src/common/utility.cpp
erikjv: done

src/common/utility_mac.mm
erikjv: No, there isn't, but I'll add one.

src/common/utility_mac.mm
erikjv: done

src/common/utility_mac.mm
erikjv: done

src/common/result.h
```diff
@@ -148,6 +148,12 @@ class Result<void, Error> : public Result<detail::NoResultData, Error>
     Result() : Result(detail::NoResultData{}) {}
 };
 
+template <typename ErrTy>
+class Error : public Result<void, ErrTy>
```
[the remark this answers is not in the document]
erikjv: ugh yeah, that's bad I'm probably going back to `Result<void, QString>`, anything else is too prone to misinterpretation.

PR #9115, opened 2021-10-05
[tests-only] Bump middleware commit id for tests

saw-jan:
This PR bumps middleware commit id for tests (but not to the latest commit id)


swoichha (approved):
lgtm

PR #9119, opened 2021-10-05
Prevent downgrade of vfs folder

TheOneRing:
image

### DIFF
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -47,7 +47,18 @@
 #include <QPushButton>
 #include <QApplication>
 
-static const char versionC[] = "version";
+namespace {
+
+/*
+ * [Accounts]
+ * 1\Folders\4\version=2
+ * 1\FoldersWithPlaceholders\3\version=3
+ */
+auto versionC()
+{
+    return QStringLiteral("version");
+}
+}
 
 namespace OCC {
 
@@ -1299,9 +1310,9 @@ void FolderDefinition::save(QSettings &settings, const FolderDefinition &folder)
 
     // Ensure new vfs modes won't be attempted by older clients
     if (folder.virtualFilesMode == Vfs::WindowsCfApi) {
-        settings.setValue(QLatin1String(versionC), 3);
+        settings.setValue(versionC(), 3);
     } else {
-        settings.setValue(QLatin1String(versionC), 2);
+        settings.setValue(versionC(), 2);
     }
 
     // Happens only on Windows when the explorer integration is enabled.
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -58,6 +58,9 @@ auto versionC()
 {
     return QStringLiteral("version");
 }
+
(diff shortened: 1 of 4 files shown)


src/gui/folder.cpp
```diff
@@ -1299,9 +1310,9 @@ void FolderDefinition::save(QSettings &settings, const FolderDefinition &folder)
 
     // Ensure new vfs modes won't be attempted by older clients
     if (folder.virtualFilesMode == Vfs::WindowsCfApi) {
-        settings.setValue(QLatin1String(versionC), 3);
+        settings.setValue(versionC(), 4);
```
erikjv, line 1313: shouldn't this call `maxSettingsVersion()`?

src/gui/folderman.cpp
```diff
@@ -37,8 +37,20 @@
 #include <QNetworkProxy>
 
 namespace {
-const char versionC[] = "version";
-const int maxFoldersVersion = 1;
+/*
+ * [Accounts]
+ * 0\version=1
+ */
+auto versionC()
```
erikjv, line 44: Isn't this a duplicate of the one in `folder.cpp`?

src/gui/folderman.cpp
```diff
@@ -403,12 +415,12 @@ void FolderMan::backwardMigrationSettingsKeys(QStringList *deleteKeys, QStringLi
...
     auto processSubgroup = [&](const QString &name) {
         settings->beginGroup(name);
-        const int foldersVersion = settings->value(QLatin1String(versionC), 1).toInt();
+        const int foldersVersion = settings->value(versionC(), 1).toInt();
         if (foldersVersion <= maxFoldersVersion) {
             const auto &childGroups = settings->childGroups();
             for (const auto &folderAlias : childGroups) {
                 settings->beginGroup(folderAlias);
-                const int folderVersion = settings->value(QLatin1String(versionC), 1).toInt();
+                const int folderVersion = settings->value(versionC(), 1).toInt();
```
erikjv, line 423: `1` is another magic number. Can you make constants somewhere in 1 place (in a header file?) and add some comments to them? I've now seen 1, 2, and 4. Oh, and there was a 3, but that's now "old"?

erikjv (changes requested):

src/gui/folderman.cpp
TheOneRing: Both are version but in different section that are evaluated independently.
But yes...
However refactoring the settings stuff is stuff for a major release...

src/gui/folderman.cpp
TheOneRing: For some reason someone decided to start to count with 1, that's why we default to 1 if no version at all is set.

src/gui/folder.cpp
TheOneRing: Hmm as you can see two lines later it doesn't always need to be the max version.
But I'll extract the two versions.

erikjv (approved):

PR #9122, opened 2021-10-06
Fix changelog entry

TheOneRing:


gabi18 (commented):
9010 should not say "... and allow to sort by the error type"

gabi18:
My fault: sorting of the table is possible, mixed it up with filtering

LGTM

Issue #8758, opened 2021-06-18
[FEATURE REQUEST] respect capability for Avatar support


michaelstingl:
There’s a new capability for Avatar support:

- https://github.com/owncloud/core/pull/38722 (oC10 backend)
- https://github.com/owncloud/ocis/issues/2031 (oCIS backend)
- Web client already uses it: https://github.com/owncloud/web/pull/5178

PR #9127, opened 2021-10-06
Draft: Work/no avatars no cry

erikjv:
The server has a capability to indicate that profile pictures (avatars)
are not available. Previously, the client would try to fetch them no
matter what, and even retry when the server responded with a 404. Now
we take the server capability into account, and do not try to fetch
those when not available.

Fixes: #8758


src/gui/connectionvalidator.cpp
```diff
@@ -286,13 +286,25 @@ void ConnectionValidator::slotUserFetched(const QJsonDocument &json)
...
-#else
-    reportResult(Connected);
+    if (capabilities.isValid() && capabilities.avatarsAvailable()) {
+        AvatarJob *job = new AvatarJob(_account, _account->davUser(), 128, this);
+        job->setTimeout(20 * 1000);
+        QObject::connect(job, &AvatarJob::avatarPixmap, this, &ConnectionValidator::slotAvatarImage);
+        job->start();
+        // reportResult will be called when the avatar has been received by `slotAvatarImage`
+    } else {
+#endif
```
TheOneRing, line 304: maybe just move the opening brace out if the ifdef?

For 2.10 we might consider to remove `TOKEN_AUTH_ONLY` as far as I understand my predecessor it was hard work to build the client without ui components but I'm pretty sure its broken anyhow....

src/libsync/capabilities.cpp
```diff
@@ -222,6 +222,12 @@ bool Capabilities::versioningEnabled() const
     return _capabilities.value(QStringLiteral("files")).toMap().value(QStringLiteral("versioning")).toBool();
 }
 
+bool Capabilities::avatarsAvailable() const
+{
+    // true by default for older servers, because turning off profile pictures was introduced in later versions
+    return _fileSharingPublicCapabilities.value(QStringLiteral("profile_picture"), true).toBool();
```
TheOneRing, line 228: I think that should be `_capabilities.value("user").toMap().value(`
Please retest with demo.owncloud.com

erikjv: no, apparently it's `files_sharing` -> `user` -> `profile_picture`.

I thought it was on the same level as `expire_date`, but it turns out there are 4 places where `expire_date` occurs.

PR #9132, opened 2021-10-07
Work/filewatcher cleanups

erikjv:
When checking if there was any deprecated API used in `utility_mac.cpp`, I noticed that using `QString::fromCFString()` would simplify the code a bit. The patches after the one for the mac is to give other platforms some love too.

### DIFF
--- a/src/gui/folderwatcher_linux.cpp
+++ b/src/gui/folderwatcher_linux.cpp
@@ -45,8 +45,11 @@ FolderWatcherPrivate::FolderWatcherPrivate(FolderWatcher *p, const QString &path
 // attention: result list passed by reference!
 bool FolderWatcherPrivate::findFoldersBelow(const QDir &dir, QStringList &fullList)
 {
-    if (!(dir.exists() && dir.isReadable())) {
-        qCDebug(lcFolderWatcher) << "Non existing path coming in: " << dir.absolutePath();
+    if (!dir.exists()) {
+        qCDebug(lcFolderWatcher) << "      - non existing path coming in: " << dir.absolutePath();
+        return false;
+    } else if (!dir.isReadable()) {
+        qCDebug(lcFolderWatcher) << "      - path without read permissions coming in: " << dir.absolutePath();
         return false;
     }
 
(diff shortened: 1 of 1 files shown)


src/gui/folderwatcher_linux.cpp
```diff
@@ -124,22 +117,21 @@ void FolderWatcherPrivate::slotAddFolderRecursive(const QString &path)
...
 void FolderWatcherPrivate::slotReceivedNotification(int fd)
 {
     int len;
-    struct inotify_event *event;
-    int i;
-    int error;
     QVarLengthArray<char, 2048> buffer(2048);
 
-    do {
+    forever {
```
TheOneRing, line 132: Those Qt macros are not always optiomal, maybe just a while(true)

src/gui/folderwatcher_linux.h
```diff
@@ -37,12 +37,11 @@ class FolderWatcherPrivate : public QObject
 public:
     FolderWatcherPrivate() {}
     FolderWatcherPrivate(FolderWatcher *p, const QString &path);
-    ~FolderWatcherPrivate() override;
 
     int testWatchCount() const { return _pathToWatch.size(); }
 
     /// On linux the watcher is ready when the ctor finished.
-    bool _ready = 1;
+    constexpr bool isReady() const { return true; }
```
TheOneRing, line 44: Is there then a reason to keep isReady?

erikjv: See windows. I didn't look at the usage side yet.

src/gui/folderwatcher_win.cpp
```diff
@@ -43,24 +40,24 @@ void WatcherThread::watchChanges(size_t fileNotifyBufferSize,
         NULL);
 
     if (_directory == INVALID_HANDLE_VALUE) {
-        const auto error = GetLastError();
-        qCWarning(lcFolderWatcher) << "Failed to create handle for" << _path << ", error:" << Utility::formatWinError(error);
-        _directory = 0;
-        return;
+        qCWarning(lcFolderWatcher) << "Failed to create handle for" << _path << ", error:" << Utility::formatWinError(GetLastError());
```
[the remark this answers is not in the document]
erikjv: good point. fixed it.

PR #9135, opened 2021-10-08
Fix warnings for using deprecated Qt API

erikjv:
This patch only fixes those warnings that we can still build on top of
Qt 5.12.

### DIFF
--- a/src/common/utility.cpp
+++ b/src/common/utility.cpp
@@ -23,21 +23,20 @@
 #include "version.h"
 
 // Note:  This file must compile without QtGui
+#include <QCollator>
 #include <QCoreApplication>
-#include <QSettings>
-#include <QTextStream>
+#include <QDateTime>
 #include <QDir>
 #include <QFile>
-#include <QUrl>
-#include <QProcess>
 #include <QObject>
-#include <QThread>
-#include <QDateTime>
-#include <QSysInfo>
+#include <QProcess>
+#include <QRandomGenerator>
+#include <QSettings>
 #include <QStandardPaths>
-#include <QCollator>
 #include <QSysInfo>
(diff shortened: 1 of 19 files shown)


src/common/utility.cpp
```diff
@@ -23,47 +23,47 @@
...
+#include <cstring>
 #include <math.h>
 #include <stdarg.h>
-#include <cstring>
 
 namespace OCC {
 
 Q_LOGGING_CATEGORY(lcUtility, "sync.utility")
 
 bool Utility::writeRandomFile(const QString &fname, int size)
```
TheOneRing, line 55: Hm can you move that function to https://github.com/owncloud/client/tree/master/test/testutils ?

src/gui/openfilemanager.cpp
```diff
@@ -108,7 +108,7 @@ void showInFileManager(const QString &localPath)
             // only around the path. Use setNativeArguments to bypass this logic.
             p.setNativeArguments(nativeArgs);
 #endif
-            p.start(explorer);
+            p.start(explorer, QStringList());
```
TheOneRing, line 111: `{}`

src/libsync/abstractnetworkjob.cpp
```diff
@@ -220,8 +219,8 @@ void AbstractNetworkJob::slotFinished()
 
             resetTimeout();
             if (_requestBody) {
-                if(!_requestBody->isOpen())
-                   _requestBody->open(QIODevice::ReadOnly);
+                if (!_requestBody->isOpen())
```
TheOneRing, line 222: I know you didn't touch that line but could you convert it to if() {}

PR #9140, opened 2021-10-13
Fix re-authenticating at startup when logged out before

### LINKED ISSUE #8924: [QA] Client ask for credentials even if logged out before
Seen with 2.9.0beta1 on Win10 and openSUSE 15.2

Steps done:
- click on "Account" button and 'Log out'
- quit ownCloud client
- restart client

Result:
- the client ask for credentials (either on browser page or with a popup) 

Expected result:
- the client shouldn't ask for login credentials if the use has logged out before

erikjv:
Fixes #8924

### DIFF
--- a/.craft.shelf
+++ b/.craft.shelf
@@ -12,7 +12,7 @@ revision = f2ef3cfe9c17827e5db19c2cce06397376d8e9e0
 
 [craft/craft-core]
 version = master
-revision = 77c114917826480f294d0432f147c9e9f7d19e21
+revision = f5f203157804b968f5e248ab1aaad201c7270c5e
 
 [core/cacert]
 version = 2020-12-08


src/gui/accountstate.cpp
```diff
@@ -28,6 +28,10 @@
 #include <QTimer>
 #include <qfontmetrics.h>
 
+namespace {
+static const QLatin1String userSignedOutC("userSignedOut");
```
TheOneRing, line 32: Maybe an inline function to align with the others used?
I'm also wondering whether we should better handle it in httpcredentials along with the user name?

erikjv:
@dragotin Actually, I disagree. The other states are either `Connected`, or temporary states that should end up in the connected state. So this is the only state where a re-authentication should not happen. Naming it `signedIn` is wouldn't be correct: `AskingCredentials` is clearly not signed-in, nor are things like `ServiceUnavailable`.

I added a comment in the source to explain this, and named the setting `userExplicitlySignedOut`.

src/gui/accountstate.cpp
erikjv: I'd say no: independent of how the user authenticated before, the state is signed-out.

That said: if the user re-authenticates, can he/she give in different credentials (e.g. the username), or is the method fixed after authenticating once?

Issue #9129, opened 2021-10-07
Bug: Remove sync root folder on disabled folder removes wrong folder


TheOneRing:
https://github.com/owncloud/client/pull/9099 does not remove the folder selected.

![image](https://user-images.githubusercontent.com/200626/136386880-c3be5b9a-aecc-4bc8-b57f-9128c88ef973.png)

PR #9144, opened 2021-10-14
Work/fix remove disabled sync root folder

erikjv:
Fixes: #9129

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -247,7 +247,7 @@ void AccountSettings::slotCustomContextMenuRequested(const QPoint &pos)
     QModelIndex index = tv->indexAt(pos);
     if (!index.isValid()) {
         return;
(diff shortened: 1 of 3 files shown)


src/gui/accountsettings.cpp
```diff
@@ -160,7 +160,7 @@ AccountSettings::AccountSettings(AccountState *accountState, QWidget *parent)
     addAction(syncNowAction);
 
     QAction *syncNowWithRemoteDiscovery = new QAction(this);
-    syncNowWithRemoteDiscovery->setShortcut(QKeySequence(Qt::CTRL + Qt::Key_F6));
+    syncNowWithRemoteDiscovery->setShortcut(QKeySequence(Qt::CTRL + int(Qt::Key_F6)));
```
TheOneRing, line 163: Hmm rather that one https://doc.qt.io/qt-5/qkeysequence.html#QKeySequence-2

PR #9145, opened 2021-10-14
Fix installation of libcloudproviders files

fmoc:
This commit fixes the installation of the INI file that belongs to libcloudproviders. I suppose the intention was to use CMake's `GNUInstallDirs` due to the variable name. The variable was unset, thus the file was installed to /cloud-providers instead of share/cloud-providers, causing the RPM builds to complain.

@TheOneRing are you aware of any KF5 macros that could replace `GNUInstallDirs`? We use KF5 elsewhere, too...

### DIFF
--- a/src/gui/libcloudproviders/libcloudproviders.cmake
+++ b/src/gui/libcloudproviders/libcloudproviders.cmake
@@ -32,7 +32,7 @@ if(WITH_LIBCLOUDPROVIDERS)
     configure_file(libcloudproviders/cloud-provider.ini.in ${CMAKE_CURRENT_BINARY_DIR}/${APPLICATION_CLOUDPROVIDERS_DBUS_NAME}.ini)
     install(
         FILES ${CMAKE_CURRENT_BINARY_DIR}/${APPLICATION_CLOUDPROVIDERS_DBUS_NAME}.ini
-        DESTINATION "${DATADIR}/cloud-providers")
+        DESTINATION "${DATA_INSTALL_DIR}/cloud-providers")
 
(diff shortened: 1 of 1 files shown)


src/gui/libcloudproviders/libcloudproviders.cmake
```diff
@@ -27,11 +29,11 @@ if(WITH_LIBCLOUDPROVIDERS)
             PkgConfig::GIO
     )
     target_compile_definitions(owncloudCore PRIVATE WITH_LIBCLOUDPROVIDERS)
-    
+
     configure_file(libcloudproviders/cloud-provider.ini.in ${CMAKE_CURRENT_BINARY_DIR}/${APPLICATION_CLOUDPROVIDERS_DBUS_NAME}.ini)
     install(
         FILES ${CMAKE_CURRENT_BINARY_DIR}/${APPLICATION_CLOUDPROVIDERS_DBUS_NAME}.ini
-        DESTINATION "${DATADIR}/cloud-providers")
+        DESTINATION "${CMAKE_INSTALL_DATADIR}/cloud-providers")
```
dschmidt, line 36:
```suggestion
        DESTINATION "${DATA_INSTALL_DIR}/cloud-providers")
```

should work without `GNUInstallDirs`?

TheOneRing:
I assume that was already an issue in 2.9?
Maybe fix it there so we can announce in 2.9.2 that we fixed cloudprovides this time for real?

TheOneRing:
Please bease on 2.9

Issue #7390, opened 2019-08-14
Installer downloaded even though it doesn't exist


HanaGemela:
Windows 10
Client beta1
Online updater: client-qa-server-0.4

Steps to recreate:

1. Set a wrong download url (https://download.owncloud.com/desktop/daily/ownCloud-11.6.0.12480.12264-daily20190807.msi) on the online updater
2. Click 'Restart&Update' button in General tab in the client (new version has to be available to see this button)
3. Check the 'Update failed' dialog message

Actual result: The message says that the installer has been downloaded and it really is present in AppData\Roaming\ownCloud
Expected result: File has not been downloaded as it doesn't exist
---
TODO:
- Allow https redirects
- Download to a temporary file and only rename to the actual name if the download was successful

ckamm:
I don't understand this one. There already is error handling for catching bad downloads. Will need to try to reproduce.

TheOneRing:
Ensure the download was successful and follow https redirects, right now we don't even follow redirects to our attic server.
On a redirect we write the http response to the target file....

PR #9153, opened 2021-10-15
Improve windows update downloader 

TheOneRing:
I guess I should implement a fake server and not rely on our mirror....
Fixes: #7390

### DIFF
--- a/src/gui/updater/ocupdater.cpp
+++ b/src/gui/updater/ocupdater.cpp
@@ -175,7 +175,7 @@ QString OCUpdater::statusString() const
     }
 }
 
-int OCUpdater::downloadState() const
+OCUpdater::DownloadState OCUpdater::downloadState() const
 {
     return _state;
 }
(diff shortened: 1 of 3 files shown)


src/gui/updater/ocupdater.cpp
```diff
@@ -355,15 +360,17 @@ void NSISUpdater::versionInfoArrived(const UpdateInfo &info)
...
             showNoUrlDialog(info);
         } else {
             _targetFile = ConfigFile::configPath() + url.mid(url.lastIndexOf('/') + 1);
-            if (QFile(_targetFile).exists()) {
+            if (QFile::exists(_targetFile)) {
                 setDownloadState(DownloadComplete);
             } else {
-                QNetworkReply *reply = qnam()->get(QNetworkRequest(QUrl(url)));
+                auto request = QNetworkRequest(QUrl(url));
+                request.setAttribute(QNetworkRequest::RedirectPolicyAttribute, QNetworkRequest::NoLessSafeRedirectPolicy);
```
fmoc, line 372: Shouldn't this be the default everywhere nowadays?

TheOneRing: The default are manual redirects, with the flag I enable redirects and don't limit them.

PR #9163, opened 2021-10-21
Add prefix to legacy migration log messages

fmoc:
Makes it easier to distinguish between the regular migration code and the legacy code's logging.


TheOneRing:
Please fix the history

PR #9184, opened 2021-10-29
Work/use account state ptr everywhere

erikjv:


src/gui/owncloudsetupwizard.cpp
```diff
@@ -167,18 +167,22 @@ void OwncloudSetupWizard::slotFoundServer(const QUrl &url, const QJsonObject &in
...
             return;
         }
     }
     slotDetermineAuthType();
 }
 
+void OwncloudSetupWizard::slotUrlUpdated()
+{
+    auto account = qobject_cast<Account *>(sender());
+    disconnect(account, &Account::urlChanged, this, &OwncloudSetupWizard::slotUrlUpdated);
```
TheOneRing, line 182: I prefer the lambda for small connects like that, moving the 3 line to a function doesn't help with the readability at all.

src/gui/accountstate.cpp
```diff
@@ -443,4 +400,62 @@ std::unique_ptr<QSettings> AccountState::settings()
...
+        }
+        if (url2.path().isEmpty()) {
+            url2.setPath(QStringLiteral("/"));
+        }
+        return url1.matches(url2, QUrl::StripTrailingSlash | QUrl::NormalizePathSegments);
+    };
+
+    if (matchUrl(newUrl, account->url())) {
+        updateUrlDialog->accept();
+        delete updateUrlDialog;
```
TheOneRing, line 458: Maybe refactor acceptNewUrl to make the code accessible without without the need for a UpdateUrlDialog object.

src/gui/accountstate.cpp
```diff
@@ -443,4 +400,62 @@ std::unique_ptr<QSettings> AccountState::settings()
...
+    private:
+        AccountPtr _account;
+        const QUrl _newUrl;
+    };
+
+} // anonymous namespace
+
+void AccountState::updateUrlDialog(AccountPtr &account, const QUrl &newUrl)
+{
+    auto updateUrlDialog = new UpdateUrlDialog(account, newUrl, ocApp()->gui()->settingsDialog());
```
TheOneRing, line 442: What about https://github.com/owncloud/client/pull/9184/commits/b7d82f4539df83b59bc3d3ebbaeea332139bb25f#diff-feb647b6021a1b6be5318e112187bd9e0ed390397dfe2f39d1b1e6c52c642473L37

I think it was one of the reasons why I moved the dialog to the account state.

TheOneRing (changes requested):
Please also target master as this change doesn't fix an issue and feels to big for a patch release.

erikjv:
I removed the patch moving the dialog out of AccountState, and fixed the leak in another way.

erikjv:
The code smells that sonarcloud reports are not directly in code I changed, or are of scale/sensitivity that they should be done in separate patches.

erikjv:
About the drone GUI tests:

```
/drone/src/build-GUI-tests/bin/owncloud: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /drone/src/build-GUI-tests/bin/owncloud)
```

TheOneRing:
> `GLIBC_2.34' not found

Uh looks like we need to get a matching squish to our new base image

src/gui/accountmanager.cpp
```diff
@@ -435,13 +436,15 @@ QString AccountManager::generateFreeAccountId() const
     }
 }
 
-void AccountManager::addAccountState(AccountState *accountState)
+void AccountManager::addAccountState(AccountStatePtr accountState)
 {
+    OC_ASSERT(accountState != nullptr)
```
TheOneRing, line 441: Q_ASSERT(accountState) ?
I mean this should really not happen, and if it does we would need to handle it.

src/gui/owncloudsetupwizard.cpp
```diff
@@ -167,12 +167,15 @@ void OwncloudSetupWizard::slotFoundServer(const QUrl &url, const QJsonObject &in
...
             _ocWizard->account()->setUrl(url);
         } else {
-            auto accountState = new AccountState(_ocWizard->account());
-            connect(accountState, &AccountState::urlUpdated, this, [accountState, this] {
-                accountState->deleteLater();
-                slotDetermineAuthType();
-            });
-            accountState->updateUrlDialog(url);
+            auto accountState = AccountState::fromNewAccount(_ocWizard->account()).take(); // we manage the lifetime for this temporary AccountState ourselves (see below)
+            connect(accountState, &AccountState::urlUpdated, this, &OwncloudSetupWizard::slotDetermineAuthType);
```
TheOneRing, line 171: the connect could be moved down inside the if

TheOneRing (changes requested):
Two nitpicks the rest is praying :D

src/gui/accountmanager.cpp
erikjv: No, it cannot happen.

src/gui/owncloudsetupwizard.cpp
erikjv: done

erikjv:
What's left to be done for this MR?

TheOneRing (approved):

PR #9186, opened 2021-10-30
Work/socketapi cleanups

erikjv:
This is in preparation to move the macOS finder sync extension over to use local sockets.

### DIFF
--- a/src/gui/main.cpp
+++ b/src/gui/main.cpp
@@ -14,17 +14,11 @@
  */
 #include <QtGlobal>
 
-#include <signal.h>
-
-#ifdef Q_OS_UNIX
-#include <sys/time.h>
-#include <sys/resource.h>
-#endif
-
 #include "application.h"
-#include "theme.h"
 #include "common/utility.h"
-#include "cocoainitializer.h"
+#include "guiutility.h"
+#include "platform.h"
+#include "theme.h"
 
 #include "updater/updater.h"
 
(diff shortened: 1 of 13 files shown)


src/gui/main.cpp
```diff
@@ -48,43 +48,16 @@ int main(int argc, char **argv)
...
-#ifdef Q_OS_WIN
-    QCoreApplication::setAttribute(Qt::AA_EnableHighDpiScaling, true);
-#endif // !Q_OS_WIN
 
-#ifdef Q_OS_MAC
-    Mac::CocoaInitializer cocoaInit; // RIIA
-#endif
+    // Create a `Platform` instance so it can set-up/tear-down stuff for us, and do any
+    // initialisation that needs to be done before creating a QApplication
+    QScopedPointer<Platform> platform(Platform::create());
```
TheOneRing, line 54: A scoped pointer and a singleton?
If we really need the singleton should we rather use a qAddPostRoutine ?

src/gui/platform.cpp
```diff
@@ -12,23 +12,28 @@
...
+Platform *Platform::_instance = nullptr;
+
+Platform::Platform()
+{
+    _instance = this;
+}
+
+QString Platform::socketApiSocketPath()
+{
+    qCWarning(lcPlatform) << "An unexpected system detected, SocketApi probably won't work.";
```
TheOneRing, line 30: If we move this to utility.h we can get rid of the single pointer?

src/gui/platform_win.cpp
```diff
@@ -0,0 +1,65 @@
...
+        }
+    }
+
+    ~WinPlatform() override;
+
+    QString socketApiSocketPath() override
+    {
+        return QLatin1String("\\\\.\\pipe\\")
+            + QLatin1String("ownCloud-")
+            + QString::fromLocal8Bit(qgetenv("USERNAME"));
```
TheOneRing, line 48: qEnvironmentVariable

src/gui/socketapi/socketapi.cpp
```diff
@@ -214,73 +215,45 @@ void SocketListener::sendMessage(const QString &message, bool doWait) const
...
+    if (!Utility::isMac()) {
+        // Not on macOS: there the directory is there, and created for us by the sandboxing
+        // environment, because we belong to an App Group.
+        QFileInfo info(socketPath);
+        if (!info.dir().exists()) {
+            bool result = info.dir().mkpath(".");
+            qCDebug(lcSocketApi) << "creating" << info.dir().path() << result;
+            if (result) {
+                QFile::setPermissions(socketPath,
+                    QFile::Permissions(QFile::ReadOwner + QFile::WriteOwner + QFile::ExeOwner));
```
TheOneRing, line 237: That should use an |  interesting that this worked...

TheOneRing (approved):

PR #9188, opened 2021-11-01
[tests-only] added tests for  multiuser share

swoichha:
### Description
This PR adds test for sharing a file with multiple users(3-users)

### Related Issue
https://github.com/owncloud/client/issues/8971#issuecomment-909885070

### DIFF
--- a/test/gui/shared/steps/steps.py
+++ b/test/gui/shared/steps/steps.py
@@ -239,6 +239,21 @@ def step(context, receiver, resource, permissions):
     openSharingDialog(context, resource)
     shareItem = SharingDialog()
     shareItem.addCollaborator(receiver, permissions)
+    shareItem.closeSharingDialog()
+
+
+@When('the user adds following collaborators of resource "|any|" using the client-UI')
+def step(context, resource):
+    openSharingDialog(context, resource)
+    shareItem = SharingDialog()
+
+    # In the following loop we are trying to share resource with given permission to one user at a time given from the data table in the feature file
+    for row in context.table[1:]:
+        receiver = row[0]
+        permissions = row[1]
+        shareItem.addCollaborator(receiver, permissions)
+
+    shareItem.closeSharingDialog()
 
 
 @When(
@@ -248,6 +263,7 @@ def step(context, receiver, resource, permissions):
     openSharingDialog(context, resource)
     shareItem = SharingDialog()
     shareItem.addCollaborator(receiver, permissions, True)
+    shareItem.closeSharingDialog()
 
 
 @Then(
@@ -265,14 +281,39 @@ def step(context, receiver, resource, permissions):
     collaboratorShouldBeListed(context, receiver, resource, permissions)
 
 
-def collaboratorShouldBeListed(context, receiver, resource, permissions):
+def collaboratorShouldBeListed(
+    context, receiver, resource, permissions, receiverCount=0
+):
     resource = getResourcePath(context, resource)
     socketConnect = syncstate.SocketConnect()
     socketConnect.sendCommand("SHARE:" + resource + "\n")
     permissionsList = permissions.split(',')
-    test.compare(
-        str(waitForObjectExists(names.scrollArea_sharedWith_QLabel).text), receiver
+
+    waitForObject(
+        {
+            "container": names.sharingDialogUG_scrollArea_QScrollArea,
+            "name": "sharedWith",
+            "type": "QLabel",
+            "visible": 1,
+        }
     )
+
+    # findAllObjects: This functionfinds and returns a list of object references identified by the symbolic or real (multi-property) name objectName.
+    sharedWithObj = findAllObjects(
+        {
+            "container": names.sharingDialogUG_scrollArea_QScrollArea,
+            "name": "sharedWith",
+            "type": "QLabel",
+            "visible": 1,
+        }
+    )
+
+    #     we use sharedWithObj list from above while verifying if users are listed or not.
+    #     For this we need an index value i.e receiverCount
+    #     For 1st user in the list the index will be 0 which is receiverCount default value
+    #     For 2nd user in the list the index will be 1 and so on
+
+    test.compare(str(sharedWithObj[receiverCount].text), receiver)
     test.compare(
         waitForObjectExists(names.scrollArea_permissionsEdit_QCheckBox).checked,
         ('edit' in permissionsList),
@@ -281,6 +322,8 @@ def collaboratorShouldBeListed(context, receiver, resource, permissions):
         waitForObjectExists(names.scrollArea_permissionShare_QCheckBox).checked,
         ('share' in permissionsList),
     )
+    shareItem = SharingDialog()
+    shareItem.closeSharingDialog()
 
 
 @When('the user waits for the files to sync')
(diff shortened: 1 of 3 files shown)


test/gui/shared/steps/steps.py
```diff
@@ -239,6 +239,20 @@ def step(context, receiver, resource, permissions):
     openSharingDialog(context, resource)
     shareItem = SharingDialog()
     shareItem.addCollaborator(receiver, permissions)
+    clickButton(waitForObject(shareItem.SHARING_DIALOG_CLOSE_BUTTON))
```
Talank, line 242: why did you move it from Page object to the steps?

test/gui/shared/steps/steps.py
```diff
@@ -239,6 +239,20 @@ def step(context, receiver, resource, permissions):
...
+def step(context, resource):
+    openSharingDialog(context, resource)
+    shareItem = SharingDialog()
+
+    for row in context.table[1:]:
+        receiver = row[0]
+        permissions = row[1]
+        shareItem.addCollaborator(receiver, permissions)
+
+    clickButton(waitForObject(shareItem.SHARING_DIALOG_CLOSE_BUTTON))
```
Talank, line 255: since it is being used in many places, you should create a function, and try to reuse it.

test/gui/shared/steps/steps.py
```diff
@@ -248,6 +262,7 @@ def step(context, receiver, resource, permissions):
     openSharingDialog(context, resource)
     shareItem = SharingDialog()
     shareItem.addCollaborator(receiver, permissions, True)
+    clickButton(waitForObject(shareItem.SHARING_DIALOG_CLOSE_BUTTON))
```
Talank, line 265: same here, repeated code

test/gui/shared/steps/steps.py
```diff
@@ -281,6 +316,8 @@ def collaboratorShouldBeListed(context, receiver, resource, permissions):
         waitForObjectExists(names.scrollArea_permissionShare_QCheckBox).checked,
         ('share' in permissionsList),
     )
+    shareItem = SharingDialog()
+    clickButton(waitForObject(shareItem.SHARING_DIALOG_CLOSE_BUTTON))
```
Talank, line 320: and here

test/gui/shared/steps/steps.py
swoichha: Because we only need this in few place not for all `addCollaborator`. So I thought it would be better to have it in here.

test/gui/shared/steps/steps.py
swoichha: The reason for doing this is that we already have `SharingDialog` object so if we make a new function for this then every time we will have two object of `SharingDialog` instead fo having to do that I did this.

test/gui/shared/steps/steps.py
```diff
@@ -239,6 +239,20 @@ def step(context, receiver, resource, permissions):
...
     shareItem.addCollaborator(receiver, permissions)
+    clickButton(waitForObject(shareItem.SHARING_DIALOG_CLOSE_BUTTON))
+
+
+@When('the user adds following collaborators of resource "|any|" using the client-UI')
+def step(context, resource):
+    openSharingDialog(context, resource)
+    shareItem = SharingDialog()
+
+    for row in context.table[1:]:
```
Talank, line 250: Add a little bit of comment here

test/gui/shared/steps/steps.py
```diff
@@ -1104,3 +1141,18 @@ def step(context, resource):
...
     publicLinkDialog = PublicLinkDialog()
     publicLinkDialog.changePassword(publicLinkName, password)
+
+
+@Then(
+    'the following users should be listed in as collaborators for file "|any|" on the client-UI'
+)
+def step(context, resource):
+    receiverCount = 0
+    for row in context.table[1:]:
```
Talank, line 1151: Add a little bit of comment here as well

test/gui/shared/steps/steps.py
```diff
@@ -239,6 +239,21 @@ def step(context, receiver, resource, permissions):
...
     shareItem.addCollaborator(receiver, permissions)
+    clickButton(waitForObject(shareItem.SHARING_DIALOG_CLOSE_BUTTON))
+
+
+@When('the user adds following collaborators of resource "|any|" using the client-UI')
+def step(context, resource):
+    openSharingDialog(context, resource)
+    shareItem = SharingDialog()
+
+    #     In the following loop we are trying to share resource with given permission to one user at a time given from the data table in the feature file
```
saw-jan, line 250: extra spaces
```suggestion
    # In the following loop we are trying to share resource with given permission to one user at a time given from the data table in the feature file
```

test/gui/shared/steps/steps.py
```diff
@@ -239,6 +239,21 @@ def step(context, receiver, resource, permissions):
     openSharingDialog(context, resource)
     shareItem = SharingDialog()
     shareItem.addCollaborator(receiver, permissions)
+    clickButton(waitForObject(shareItem.SHARING_DIALOG_CLOSE_BUTTON))
```
saw-jan, line 242: we can create a function `closeSharingDialog`  in pageObject and call it here and other places when needed.

test/gui/shared/steps/steps.py
swoichha: This change was made by black

test/gui/shared/steps/steps.py
```diff
@@ -265,14 +281,39 @@ def step(context, receiver, resource, permissions):
...
+            "name": "sharedWith",
+            "type": "QLabel",
+            "visible": 1,
+        }
     )
+
+    #     we use sharedWithObj list from above while verifying if users are listed or not.
+    #     For this we need an index value i.e receiverCount
+    #     For 1st user in the list the index will be 0 which is receiverCount default value
+    #     For 2nd user in the list the index will be 1 and so on
```
saw-jan, line 314: not a big deal but could you please fromat them?
```suggestion
    # we use sharedWithObj list from above while verifying if users are listed or not.
    # For this we need an index value i.e receiverCount
    # For 1st user in the list the index will be 0 which is receiverCount default value
    # For 2nd user in the list the index will be 1 and so on
```

test/gui/shared/steps/steps.py
```diff
@@ -1104,3 +1147,20 @@ def step(context, resource):
...
     publicLinkDialog = PublicLinkDialog()
     publicLinkDialog.changePassword(publicLinkName, password)
+
+
+@Then(
+    'the following users should be listed in as collaborators for file "|any|" on the client-UI'
+)
+def step(context, resource):
+    #     Here we are trying to verify if the user added in when step are listed in the client-UI or not
+    #     We now have a variable name receiverCount which is used in collaboratorShouldBeListed function call
```
saw-jan, line 1157: same as above
```suggestion
    # Here we are trying to verify if the user added in when step are listed in the client-UI or not
    # We now have a variable name receiverCount which is used in collaboratorShouldBeListed function call
```

test/gui/shared/steps/steps.py
swoichha: @saw-jan This is also done by black

saw-jan (commented):
LGTM

Talank (approved):

Issue #8733, opened 2021-06-15
Creating public link share with expiration date doesn't work


swoichha:
### Steps to reproduce
1. Create a public link for a file with expiration date e.g. 6/25/21
2. Close the sharing dialog
3. Open the public links sharing dialog for that file again
4. Select the recently created public link 

### Expected behavior
In the sharing dialog for public link shared file with expiration date the `Set expiration date` checkbox should be checked and the expiration date field should have the date same as the date set during creation of public link(6/16/21).
![image](https://user-images.githubusercontent.com/17021686/122049333-07bcc400-ce02-11eb-82db-66352d20d333.png)

### Actual behavior
When a user creates a public link with expiration date and opens the sharing dialog in desktop-client for that file then the `Set expiration date` checkbox is unchecked and the expiration date field have public link creation date(6/16/21).
![image](https://user-images.githubusercontent.com/17021686/122049350-0c817800-ce02-11eb-8ef7-d2c9dc79e53d.png)

individual-it:
This seems only to happen with the latest master of core
CC @micbar

phil-davis:
Is this a problem introduced by the expiration time in share responses changing from 00:00:00 to 23:59:59 ?

We really need to know if that is going to break desktop clients that are "in the wild" - before branching core for 10.8 release!

@micbar @TheOneRing

individual-it:
Tomorrow we will check the traffic and see what causes it

swoichha:
## Pre-Condition
- For user admin a folder name `test-user` containing file `login.feature` should exist.

###  Latest master of core  
Creating public link  with expiration date `2021-06-25` and `Download / View` permission using following command:
- `curl -k -X POST -u admin:admin 'http://localhost/core/ocs/v2.php/apps/files_sharing/api/v1/shares' -d 'shareType=3&path=%2Ftest-user%2Flogin.feature&permissions=1&expireDate=18-06-2021'| xmllint --format -`

### Response
We get `<expiration>2021-06-18 23:59:59</expiration>` .

### Tarball 10.7.0

Creating public link  with expiration date `2021-06-25` and `Download / View` permission using following command:
- `curl -k -X POST -u admin:admin 'http://localhost/owncloud-10.7.0/owncloud/ocs/v2.php/apps/files_sharing/api/v1/shares' -d 'shareType=3&path=%2Ftest-user%2Flogin.feature&permissions=1&expireDate=18-06-2021'| xmllint --format -`

### Response
[...]

PR #9190, opened 2021-11-01
[tests-only] Added more gui tests for sharing with user and group (Part 01)

saw-jan:
### Description

### Related Issues
- Part of https://github.com/owncloud/client/issues/8971#issuecomment-909884257

### DIFF
--- a/test/gui/shared/scripts/pageObjects/PublicLinkDialog.py
+++ b/test/gui/shared/scripts/pageObjects/PublicLinkDialog.py
@@ -22,6 +22,18 @@ class PublicLinkDialog:
         "type": "QCheckBox",
         "visible": 1,
     }
+    PASSWORD_INPUT_FIELD = {
+        "container": names.qt_tabwidget_stackedwidget_OCC_ShareLinkWidget_OCC_ShareLinkWidget,
+        "name": "lineEdit_password",
+        "type": "QLineEdit",
+        "visible": 1,
+    }
+    EXPIRYDATE_CHECKBOX = {
+        "container": names.qt_tabwidget_stackedwidget_OCC_ShareLinkWidget_OCC_ShareLinkWidget,
+        "name": "checkBox_expire",
+        "type": "QCheckBox",
+        "visible": 1,
+    }
     CREATE_SHARE_BUTTON = {
         "container": names.qt_tabwidget_stackedwidget_OCC_ShareLinkWidget_OCC_ShareLinkWidget,
         "name": "createShareButton",
@@ -69,7 +81,9 @@ class PublicLinkDialog:
             squish.Qt.LeftButton,
         )
 
-    def createPublicLink(self, context, resource, password='', permissions=''):
+    def createPublicLink(
+        self, context, resource, password='', permissions='', expireDate='', linkName=''
+    ):
         radioObjectName = ''
         if permissions:
             radioObjectName = self.getRadioObjectForPermssion(permissions)
@@ -86,18 +100,13 @@ class PublicLinkDialog:
             squish.clickButton(squish.waitForObject(radioObjectName))
 
         if password:
-            squish.clickButton(squish.waitForObject(self.PASSWORD_CHECKBOX))
-            squish.mouseClick(
-                squish.waitForObject(self.PASSWORD_CHECKBOX),
-                0,
-                0,
-                squish.Qt.NoModifier,
-                squish.Qt.LeftButton,
-            )
-            squish.type(
-                squish.waitForObject(self.PASSWORD_CHECKBOX),
-                password,
-            )
+            self.setPassword(password)
+
+        if expireDate:
+            self.setExpirationDate(expireDate)
+
+        if linkName:
+            self.setLinkName(linkName)
 
         squish.clickButton(squish.waitForObject(self.CREATE_SHARE_BUTTON))
         squish.waitFor(
@@ -107,22 +116,35 @@ class PublicLinkDialog:
             )
         )
 
-    def togglesPassword(self):
+    def togglePassword(self):
         squish.clickButton(squish.waitForObject(self.PASSWORD_CHECKBOX))
 
-    def setExpirationDate(self, context, publicLinkName, resource):
-        test.compare(
-            str(squish.waitForObjectExists(self.ITEM_TO_SHARE).text),
-            resource,
+    def setPassword(self, password):
+        enabled = squish.waitForObjectExists(self.PASSWORD_CHECKBOX).checked
+        if not enabled:
+            self.togglePassword()
+
+        squish.mouseClick(
+            squish.waitForObject(self.PASSWORD_INPUT_FIELD),
+            0,
+            0,
+            squish.Qt.NoModifier,
+            squish.Qt.LeftButton,
         )
-        test.compare(
-            str(squish.waitForObjectExists(self.PUBLIC_LINK_NAME).text),
-            publicLinkName,
+        squish.type(
+            squish.waitForObject(self.PASSWORD_INPUT_FIELD),
+            password,
         )
-        expDate = []
-        for row in context.table:
-            if row[0] == 'expireDate':
-                expDate = datetime.datetime.strptime(row[1], '%Y-%m-%d')
+
+    def toggleExpirationDate(self):
+        squish.clickButton(squish.waitForObject(self.EXPIRYDATE_CHECKBOX))
+
+    def setExpirationDate(self, expireDate):
+        enabled = squish.waitForObjectExists(self.EXPIRYDATE_CHECKBOX).checked
+        if not enabled:
+            self.toggleExpirationDate()
+
+        expDate = datetime.datetime.strptime(expireDate, '%Y-%m-%d')
         expYear = expDate.year - 2000
         squish.mouseClick(
             squish.waitForObject(self.EXPIRATION_DATE_FIELD),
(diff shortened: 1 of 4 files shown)


test/gui/shared/steps/steps.py
```diff
@@ -873,6 +910,27 @@ def step(context, resource, content):
...
+@When('the user tries to overwrite the file "|any|" with content "|any|"')
+def step(context, resource, content):
+    waitForFileToBeSynced(context, resource)
+    waitForFolderToBeSynced(context, '/')
+
+    # overwriting the file immediately after it has been synced from the server seems to have some problem.
+    # The client does not see the change although the changes have already been made thus we are having a race condition
+    # So for now we add 5sec static wait
+    # an issue https://github.com/owncloud/client/issues/8832 has been created for it
+    snooze(5)
```
Talank, line 922: do we need to wait for at least 5 sec?

test/gui/shared/scripts/pageObjects/PublicLinkDialog.py
```diff
@@ -137,17 +159,45 @@ def setExpirationDate(self, context, publicLinkName, resource):
...
+        squish.nativeType("<Ctrl+Right>")
+        squish.nativeType("<Ctrl+Right>")
+        squish.nativeType(year)
+        squish.nativeType("<Ctrl+Left>")
+        squish.nativeType(day)
+        squish.nativeType("<Ctrl+Left>")
+        squish.nativeType("<Ctrl+Left>")
+        squish.nativeType(month)
+        squish.nativeType("<Return>")
+
```
Talank, line 200: Can't we do this in one line? 🤔 the documentation is `nativeType(keys);` so can you try if you can send all the keys in a single call? such as `squish.nativeType("<Ctrl+Right>", "<Ctrl+Right>", year)`. I am not sure if it works.

Also, can you add some comment here.

test/gui/shared/scripts/pageObjects/PublicLinkDialog.py
```diff
@@ -137,17 +159,45 @@ def setExpirationDate(self, context, publicLinkName, resource):
...
+        actualDate = squish.waitForObjectExists(self.EXPIRATION_DATE_FIELD).displayText
+        expectedDate = f"{expDate.month}/{expDate.day}/{expYear}"
+        if not actualDate == expectedDate:
+            # retry with workaround
+            self.setExpirationDateWithWorkaround(expYear, expDate.month, expDate.day)
+
         squish.testSettings.silentVerifications = True
         squish.waitFor(
-            lambda: (test.xvp("publicLinkExpirationProgressIndicatorInvisible"))
+            lambda: (test.vp("publicLinkExpirationProgressIndicatorInvisible"))
```
Talank, line 171: why did you do this? Is it not failing anymore? When did this happen?

saw-jan: Yeah, I forgot about that.
what is that for actually? Before the test was failing locally so I had to change that one.
Haven't tried recently though. I will

test/gui/shared/steps/steps.py
saw-jan: Just copy pasted from code from above. Have to try deleting it.

test/gui/shared/scripts/pageObjects/PublicLinkDialog.py
Talank: @swoichha any idea why and since when is it passing?

saw-jan: I found out that `xvp` skips the test steps below so that the test scenario is shown as `pass`.
And with `vp` all the test steps will get to run and the test scenario is also passing. so good to have these changes.

test/gui/shared/scripts/pageObjects/PublicLinkDialog.py
saw-jan: `nativeType` seems to accept only one argument. So cannot send multiple keys at once.

test/gui/shared/scripts/pageObjects/PublicLinkDialog.py
Talank: yeah, xvp is like `expected to fail`

test/gui/shared/scripts/pageObjects/SharingDialog.py
```diff
@@ -87,20 +100,12 @@ def selectCollaborator(self, receiver, isGroup=False):
...
-            squish.Qt.NoModifier,
-            squish.Qt.LeftButton,
-        )
-        squish.type(
-            squish.waitForObject(self.SHARE_WITH_COLLABORATOR_INPUT_FIELD),
-            receiver,
-        )
+        self.searchCollaborator(receiver)
+
+        escapedReceiverName = receiver.replace("_", "\\_").replace(".", "\\.")
```
Talank, line 105: add a comment here as well. May be include example input and output

test/gui/shared/steps/steps.py
```diff
@@ -1103,4 +1155,53 @@ def step(context, resource):
...
+    shareItem = SharingDialog()
+    shareItem.searchCollaborator(collaborator)
+
+
+@When(
+    'the user searches for collaborator with autocomplete characters "|any|" using the client-UI'
+)
+def step(context, collaborator):
+    shareItem = SharingDialog()
+    shareItem.searchCollaborator(collaborator)
```
Talank, line 1173: How come these 2 step definitions has the same logic? 🤔 And if it is the same logic, then we can reuse the steps.

saw-jan: Because when you type the full collaborator name or 3 characters, it is just the same (you search by typing)

test/gui/shared/scripts/pageObjects/PublicLinkDialog.py
swoichha: @saw-jan @Talank The reason we had `xvp` was due to this issue https://github.com/owncloud/client/issues/8733. As the checkbox use to be unchecked we did use `xvp` for that but now this issue has been fixed and closed. So, IMO replacing `xvp` with `vp` should make the test work.

test/gui/shared/steps/steps.py
Talank: You can put this logic in a function and then call the function from both of these step definitions. And if the logic should be changed someday, it is easier if we reuse the code.
```
shareItem = SharingDialog()
shareItem.searchCollaborator(collaborator)
```

test/gui/shared/scripts/pageObjects/PublicLinkDialog.py
```diff
@@ -173,24 +229,22 @@ def createPublicLinkWithRole(self, role):
...
-            0,
-            squish.Qt.NoModifier,
-            squish.Qt.LeftButton,
-        )
-        squish.type(
-            squish.waitForObject(names.oCC_ShareLinkWidget_lineEdit_password_QLineEdit),
-            password,
-        )
+    def changePassword(self, password):
+        self.setPassword(password)
```
swoichha, line 233: I  think this function should remain the way it is as we check the link name(which might be useful when there are many public links for a file/folder) and edit the existing password but `setPassword` does not check the name of public link. 

Maybe you can call `verifyPublicLinkName` function here?

test/gui/tst_sharing/test.feature
```diff
@@ -6,15 +6,32 @@ Feature: Sharing
...
+        And user "Brian Murphy" should be listed in the collaborators list for file "simple-folder" with permissions "edit,share" on the client-UI
+
+
+    Scenario: sharing file/folder with a user that has special characters as username
+        Given user "Speci@l_Name-.+" has been created on the server with default attributes and without skeleton files
+        And user "Alice" has created folder "FOLDER" on the server
+        And user "Alice" has uploaded file with content "ownCloud test text file" to "textfile.txt" on the server
+        And user "Alice" has set up a client with default settings
+        When the user opens the sharing dialog of "textfile.txt" using the client-UI
+        When the user adds "Speci@l_Name-.+" as collaborator of resource "textfile.txt" with permissions "edit,share" using the client-UI
```
Talank, line 31:
```suggestion
        When the user opens the sharing dialog of "textfile.txt" using the client-UI
        And the user adds "Speci@l_Name-.+" as collaborator of resource "textfile.txt" with permissions "edit,share" using the client-UI
```

test/gui/shared/scripts/pageObjects/PublicLinkDialog.py
saw-jan: the public link name is checked in this step `the user changes the password of public link...`
here: https://github.com/owncloud/client/blob/e4cabdb9bd1eb2323b2df3b8a923deb4c92aeac3/test/gui/shared/steps/steps.py#L1199

test/gui/tst_sharing/test.feature
```diff
@@ -72,21 +135,89 @@ Feature: Sharing
         And user "Brian" has been created on the server with default attributes and without skeleton files
         And user "Alice" has shared folder "simple-folder" on the server with user "Brian" with "all" permissions
         And user "Brian" has set up a client with default settings
-        When the user overwrites the file "simple-folder/textfile.txt" with content "overwrite ownCloud test text file"
-        Then as "Brian" the file "simple-folder/textfile.txt" on the server should have the content "overwrite ownCloud test text file"
+        When the user overwrites the file "Shares/simple-folder/textfile.txt" with content "overwrite ownCloud test text file"
+        Then as "Brian" the file "Shares/simple-folder/textfile.txt" on the server should have the content "overwrite ownCloud test text file"
```
Talank, line 139: Any reason for setting default share location to `Shares` dir? You can avoid many lines of changes if you do noe set the sharing dir.

saw-jan: I did it for consistency but now I know the context :smile:

Talank (approved):
LGTM

PR #9191, opened 2021-11-02
Properly encode version suffix

TheOneRing:


TheOneRing:
Will be solved on master only

PR #9201, opened 2021-11-09
[tests-only] add tests for public share

swoichha:
### Description
add tests for public link 
- Public link to a file and set a password with special characters
- Public link to a file and disable it after | use the sharing menu to delete the public link
- Public link to a file and set a password. Modify Password Policies and retry

### Related Issue
https://github.com/owncloud/client/issues/8971#issuecomment-909885070

### DIFF
--- a/test/gui/shared/steps/steps.py
+++ b/test/gui/shared/steps/steps.py
@@ -1082,3 +1082,25 @@ def step(context):
         test.compare(actualFolder, expectedFolder)
 
         rowIndex += 1
+
+
+@When('the user deletes the public link for file "|any|"')
+def step(context, resource):
+    openSharingDialog(context, resource)
+    publicLinkDialog = PublicLinkDialog()
+    publicLinkDialog.openPublicLinkDialog()
+
+    test.compare(
+        str(waitForObjectExists(publicLinkDialog.ITEM_TO_SHARE).text),
+        resource.replace(context.userData['currentUserSyncPath'], ''),
+    )
+    clickButton(waitForObject(names.linkShares_QToolButton_2))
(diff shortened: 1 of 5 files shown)


test/gui/tst_sharing/test.feature
```diff
@@ -131,12 +131,39 @@ Feature: Sharing
...
 
 
-    Scenario: simple sharing of a file by public link with password
+    Scenario: sharing of a file by public link and deleting the link
+        Given user "Alice" has uploaded file with content "ownCloud test text file 0" to "/textfile0.txt" on the server
+        And user "Alice" has set up a client with default settings
+        And user "Alice" has created a public link on the server with following settings
+            | path     | textfile0.txt |
+            | name     | Public-link   |
+        When the user deletes the public link for "textfile0.txt"
```
saw-jan, line 140: good to have resource type as well
```suggestion
        When the user deletes the public link for file "textfile0.txt"
```

test/gui/shared/steps/steps.py
```diff
@@ -1088,3 +1088,30 @@ def step(context):
...
+
+@When(
+    'the user changes the password of public link "|any|" to "|any|" using the client-UI'
+)
+def step(context, publicLinkName, password):
+    publicLinkDialog = PublicLinkDialog()
+    publicLinkDialog.changePassword(publicLinkName, password)
+
+
+#     mouseClick(waitForObject(names.oCC_ShareLinkWidget_lineEdit_password_QLineEdit), 373, 13, Qt.NoModifier, Qt.LeftButton)
```
saw-jan, line 1115: so can we delete this?

individual-it (approved):

PR #9202, opened 2021-11-09
Ensure we only update the clients url after the sync was stopped

TheOneRing:
When the client detects a change of the url we ask the user to accept
the change or if it was only representational change we directly accept the change.
Due to a bug the we aborted the sync only after we updated the url.
This caused the client to idle for one minute.


TheOneRing:
See https://github.com/owncloud/client/blob/2.9/src/gui/connectionvalidator.cpp#L118 
and
https://github.com/owncloud/enterprise/issues/4803#issuecomment-963565226

erikjv (approved):

PR #9211, opened 2021-11-11
macOS: allow for debug builds in xcode-build

erikjv:
Propagate the cmake build type to xcode-build too. Previously, a release
build would always be done. Now, a debug build will be done when the
cmake build type is Debug.

### DIFF
--- a/shell_integration/MacOSX/CMakeLists.txt
+++ b/shell_integration/MacOSX/CMakeLists.txt
@@ -1,13 +1,19 @@
 if(APPLE)
 set(OC_OEM_SHARE_ICNS "${CMAKE_BINARY_DIR}/src/gui/${APPLICATION_ICON_NAME}.icns")
 
+if (CMAKE_BUILD_TYPE STREQUAL "Debug")
+    set(XCODE_CONFIG "Debug")
+else()
+    set(XCODE_CONFIG "Release")
+endif()
+
 # The bundle identifier and application group need to have compatible values with the client
 # to be able to open a Mach port across the extension's sandbox boundary.
 # Pass the info through the xcodebuild command line and make sure that the project uses
 # those user-defined settings to build the plist.
 add_custom_target( mac_overlayplugin ALL
     xcodebuild -project ${CMAKE_SOURCE_DIR}/shell_integration/MacOSX/OwnCloudFinderSync/OwnCloudFinderSync.xcodeproj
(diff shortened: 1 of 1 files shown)


TheOneRing (approved):

shell_integration/MacOSX/CMakeLists.txt
```diff
@@ -1,13 +1,19 @@
 if(APPLE)
 set(OC_OEM_SHARE_ICNS "${CMAKE_BINARY_DIR}/src/gui/${APPLICATION_ICON_NAME}.icns")
 
+if (${CMAKE_BUILD_TYPE} EQUAL "Debug")
```
dschmidt, line 4:
```suggestion
if ("${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
```
or
```suggestion
if (CMAKE_BUILD_TYPE STREQUAL "Debug")
```

shell_integration/MacOSX/CMakeLists.txt
```diff
@@ -1,13 +1,19 @@
...
+endif()
+
 # The bundle identifier and application group need to have compatible values with the client
 # to be able to open a Mach port across the extension's sandbox boundary.
 # Pass the info through the xcodebuild command line and make sure that the project uses
 # those user-defined settings to build the plist.
 add_custom_target( mac_overlayplugin ALL
     xcodebuild -project ${CMAKE_SOURCE_DIR}/shell_integration/MacOSX/OwnCloudFinderSync/OwnCloudFinderSync.xcodeproj
-        -target FinderSyncExt -configuration Release "SYMROOT=${CMAKE_CURRENT_BINARY_DIR}"
+        -target FinderSyncExt -configuration ${XCODE_CONFIG} "SYMROOT=${CMAKE_CURRENT_BINARY_DIR}"
```
dschmidt, line 16:
```suggestion
        -target FinderSyncExt -configuration "${XCODE_CONFIG}" "SYMROOT=${CMAKE_CURRENT_BINARY_DIR}"
```

dschmidt (changes requested):
The first change is a must, the second doesn't matter in this case but I think it's better to consistently quote strings :)

shell_integration/MacOSX/CMakeLists.txt
erikjv: done

shell_integration/MacOSX/CMakeLists.txt
erikjv: done

dschmidt (approved):

Issue #9209, opened 2021-11-11
Team identifier missing in release builds


erikjv:
When doing a release build, the CMake variable `SOCKETAPI_TEAM_IDENTIFIER_PREFIX` is not set to the team identifier. The team identifier looks like `4AP2STM4H5 `, and the variable should include an extra period, eg: `4AP2STM4H5.`

To see the team identifier used to build the app bundle, you can use:
```sh
codesign -dvvvv owncloud.app/Contents/MacOS/owncloud | grep TeamIdentifier
TeamIdentifier=4AP2STM4H5
```

This team-id is already propagated to the FinderSyncExt, and should show up in the `com.apple.security.application-groups` array as `4AP2STM4H5.com.owncloud.desktopclient` This can be verified by running:

```sh
codesign -d --entitlements :- owncloud.app/Contents/PlugIns/FinderSyncExt.appex/Contents/MacOS/FinderSyncExt
```

TheOneRing:
Shouldn't we add the extra period to the code that needs it and not to the variable itself?

erikjv:
> Shouldn't we add the extra period to the code that needs it and not to the variable itself?

Done

dschmidt:
Enough to fix this for 2.10?

erikjv:
> Enough to fix this for 2.10?

If the cmake invocation of the CI/release-build-thingy is done, then yes.

PR #9212, opened 2021-11-12
macOS: change SOCKETAPI_TEAM_IDENTIFIER_PREFIX concatenation

erikjv:
Previously, the socket api prefix was created by concatenating
SOCKETAPI_TEAM_IDENTIFIER_PREFIX and OC_APPLICATION_REV_DOMAIN together,
so the first variable had to end in a "." to get the correct result.
This change moves adding that period from configuration to the places
where the concatenation happens.


TheOneRing:
Downside is that we now  probably require a value...  I doubt that apple is fond of leading periods.
On the other hand we will require signed builds for mac anyhow. At least when we build the shell extension or the fileprovider extension.

erikjv:
I think we should wait with merging this one until the rest of #9209 is fixed.

PR #9213, opened 2021-11-12
Fix values in sync item

### LINKED ISSUE #7348: .owncloudsync.log does not contain duration
### Expected behaviour
The .owncloudsync.log should contain the duration column.

### Actual behaviour
The column (it's the second one after the timestamp) is always empty. It used to be there, but was eliminated.

### Client configuration
Client version: 2.5.4, but also master

The .owncloudsync.log is a useful file to make examinations about the sync operation of the client for advanced users. I often use it to check the what happened. I think it is an important feature for people who appreciate transparency of their tools, especially in complex things like file sync. That is a benefit of free software IMHO, and it should be maintained carefully ;-)

The duration was removed in patches like 4aeaa67caeea952a567c9d5c036fbe7c8aeba880. I think that is the wrong way. The less information the log contains the less useful it is. Please consider to bring the full data back. Maybe there is a good way of disabling it for "normal" use?

Also, the reason for saving it because of memory usage is not longer valid because you do not build the full reconcile list any more, right @ogoffart ?

Thanks for considering!

dragotin:
Always set the request time, response value and request Id in the SyncFileItem.
    
These values are written to the .owncloudsync.log file which is useful for debugging. For some operations, some data is missing. This PR fixes that. 
Fixes: #7348

### DIFF
--- a/src/libsync/propagateuploadng.cpp
+++ b/src/libsync/propagateuploadng.cpp
@@ -222,14 +222,12 @@ void PropagateUploadFileNG::slotPropfindFinished()
 void PropagateUploadFileNG::slotPropfindFinishedWithError()
 {
     auto job = qobject_cast<LsColJob *>(sender());
-    auto httpErrorCode = job->reply()->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
-    _item->_httpErrorCode = httpErrorCode;
+    auto httpErrorCode = _item->_httpErrorCode = job->reply()->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
     _item->_responseTimeStamp = job->responseTimestamp();
     _item->_requestId = job->requestId();
     slotJobDestroyed(job); // remove it from the _jobs list
 
     QNetworkReply::NetworkError err = job->reply()->error();
-    job->reply()->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
     auto status = classifyError(err, httpErrorCode, &propagator()->_anotherSyncNeeded);
     if (status == SyncFileItem::FatalError) {
         propagator()->_activeJobList.removeOne(this);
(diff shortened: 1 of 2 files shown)


src/libsync/propagateuploadng.cpp
```diff
@@ -222,12 +222,16 @@ void PropagateUploadFileNG::slotPropfindFinished()
 void PropagateUploadFileNG::slotPropfindFinishedWithError()
 {
     auto job = qobject_cast<LsColJob *>(sender());
+    auto httpErrorCode = job->reply()->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
+    _item->_httpErrorCode = httpErrorCode;
```
TheOneRing, line 226: Nitpick: just set  _item->_httpErrorCode in one line (yes its 3 chars longer)

src/libsync/propagateuploadng.cpp
```diff
@@ -222,12 +222,16 @@ void PropagateUploadFileNG::slotPropfindFinished()
...
+    _item->_responseTimeStamp = job->responseTimestamp();
+    _item->_requestId = job->requestId();
     slotJobDestroyed(job); // remove it from the _jobs list
+
     QNetworkReply::NetworkError err = job->reply()->error();
-    auto httpErrorCode = job->reply()->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
+    job->reply()->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
```
TheOneRing, line 232: remove the now dead line?

TheOneRing (changes requested):

TheOneRing:
Offtoppic: https://github.com/owncloud/client/blob/4a3fe9e8bc0c8d288cede46fd7691f21dcac2320/src/libsync/propagateupload.h#L173 
Shouldn't we use a set? I mean I wouldn't expect the same job in there multiple times and it would reduce the complexity of the removal of the job.

src/libsync/propagateuploadng.cpp
dragotin: yes, thanks

PR #9214, opened 2021-11-12
Work/macos use local socket

erikjv:


erikjv:
Merge after #9212 is merged.

PR #9216, opened 2021-11-15
Branding option to disable the multiple db files warning

TheOneRing:
Also cache error message to prevent unnecessary listing of sync root.

### DIFF
--- a/changelog/unreleased/9216
+++ b/changelog/unreleased/9216
@@ -1,0 +1,7 @@
+Enhancement: Branding option to disable warning for multiple sync_journal.db's
+
+We added a branding option that disables the `Multiple accounts are sharing the folder` warning.
+In previous client versions a bug caused the creation of new sync journals, causing false positives in the detection.
+While this can be handled by the individual user, companies with multiple hundreds of users may opt to disable the warning.
+
+https://github.com/owncloud/client/pull/9216
--- a/src/gui/folderman.cpp
+++ b/src/gui/folderman.cpp
@@ -1535,10 +1535,17 @@ Result<void, QString> FolderMan::unsupportedConfiguration(const QString &path) c
     if (it == _unsupportedConfigurationError.end()) {
         it = _unsupportedConfigurationError.insert(path, [&]() -> Result<void, QString> {
             if (numberOfSyncJournals(path) > 1) {
-                return tr("Multiple accounts are sharing the folder %1.\n"
-                          "This configuration is know to lead to dataloss and is no longer supported.\n"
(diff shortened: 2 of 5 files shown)


src/gui/folderman.cpp
```diff
@@ -1531,13 +1531,23 @@ void FolderMan::restartApplication()
...
+    if (it == _unsupportedConfigurationError.end()) {
+        it = _unsupportedConfigurationError.insert(path, [&]() -> Result<void, QString> {
+            if (numberOfSyncJournals(path) > 1) {
+                const QString error = tr("Multiple accounts are sharing the folder %1.\n"
+                                         "This configuration is know to lead to dataloss and is no longer supported.\n"
+                                         "Please consider removing this folder from the account and adding it again.")
+                                          .arg(path);
+                qCWarning(lcFolderMan) << error;
+                if (Theme::instance()->warnOnMultipleDb()) {
+                    return error;
```
dragotin, line 1544: Not sure I like this: In the logfile, you will find the Warning/Error, but in fact the code does not return the error. That will be very confusing in debugging situations. Rather add a hint that it was disabled for the log as well.

dragotin (commented):
One comment, otherwise ok.

changelog/unreleased/9216
```diff
@@ -0,0 +1,7 @@
+Enhancement: Branding option to disable warning for multiple sync_journal.db's
+
+We added a branding option that disables the `Multiple accounts are sharing the folder`.
+In previous client version a bug caused the creation of new sync journals, causing false positives in the detection.
```
dschmidt, line 4:
```suggestion
In previous client versions a bug caused the creation of new sync journals, causing false positives in the detection.
```

changelog/unreleased/9216
```diff
@@ -0,0 +1,7 @@
...
+While this can be handle by the individual user, companies with multiple hundreds of users may opt to disable the warning.
```
dschmidt, line 5:
```suggestion
While this can be handled by the individual user, companies with multiple hundreds of users may opt to disable the warning.
```

changelog/unreleased/9216
```diff
@@ -0,0 +1,7 @@
+Enhancement: Branding option to disable warning for multiple sync_journal.db's
+
+We added a branding option that disables the `Multiple accounts are sharing the folder`.
```
dschmidt, line 3:
```suggestion
We added a branding option that disables the `Multiple accounts are sharing the folder` warning.
```

PR #9219, opened 2021-11-16
Ensure we have a window to receive the nativeEvents

### LINKED ISSUE #8979: Client not always stops when requested by Windows
Seen (again) with 2.9.0-rc2

* run any older version of the owncloud client
* Download the msi installer of an update
* click start the install.
* It announces, it needs to shutdown `owncloud` and sometimes also `explorer`
* when the user confirms. It usually fails to shut down one of the two.

image

After a second round it normally succeeds.

TheOneRing:
Fixes #8979

### DIFF
--- a/src/gui/settingsdialog.cpp
+++ b/src/gui/settingsdialog.cpp
@@ -275,6 +275,10 @@ SettingsDialog::SettingsDialog(ownCloudGui *gui, QWidget *parent)
     setMinimumSize(::minimumSizeHint(this));
 #ifdef Q_OS_MAC
     setActivationPolicy(ActivationPolicy::Accessory);
+#elif defined(Q_OS_WIN) && QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
+    // ensure we create a native window in order to receive nativeEvents
+    // by default this is only called on show
+    create();
 #endif
 }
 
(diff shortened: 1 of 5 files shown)


test/manual/windows close event/README.md
```diff
@@ -0,0 +1,11 @@
+# Simple application to simulate application termination by MSI
+
+This is a simple application that sends the Window messages `WM_QUERYENDSESSION` and `WM_ENDSESSION` to all windows of an application.
+Those messages are send by MSI when a current running application should quit.
```
dschmidt, line 4:
```suggestion
Those messages are sent by MSI when a current running application should quit.
```

src/gui/settingsdialog.cpp
```diff
@@ -275,6 +275,10 @@ SettingsDialog::SettingsDialog(ownCloudGui *gui, QWidget *parent)
     setMinimumSize(::minimumSizeHint(this));
 #ifdef Q_OS_MAC
     setActivationPolicy(ActivationPolicy::Accessory);
+#elif defined(Q_OS_WIN) && QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
+    // ensure we create a native window in order to receive nativeEvents
+    // by default this is only called on show
+    create();
```
dschmidt, line 281: 😮

dschmidt: Have you checked the behavior with Qt6?

dschmidt (approved):

src/gui/settingsdialog.cpp
TheOneRing: I just did, the create is needed there too.

dschmidt (changes requested):
Remove the version condition please as it's still required with qt6

PR #9220, opened 2021-11-17
Ensure we receive WM_ENDSESSION 

TheOneRing:


TheOneRing:
Alternative implementation to https://github.com/owncloud/client/pull/9219

dschmidt:
One hundred gazillion lines vs. one line? :-D

erikjv (approved):

TheOneRing:
> 
> 
> One hundred gazillion lines vs. one line? :-D

Yes I don't like it either but we can assume what happens in here.
I have no idea whether `create()` has any unexpected sideeffects...

dschmidt:
I guess `create()` is fine, but handling QUERYENDSESSION and ENDSESSION more correctly is what convinces me more of this version.

TheOneRing:
The fact that the ci is stuck in tests since 20min tells me that I need to have another look.
❤️ ci

TheOneRing:
Now we have an unrelated segfault in a unittest on mac

erikjv (approved):

PR #9221, opened 2021-11-17
Ci: Translations, only push if we have a change

TheOneRing:
Switch to powershell as the other scripts are using it too.
Only haggle with ssh and push if we have a change.
(I'm getting a mail every day https://github.com/owncloud/client/runs/4220100543?check_suite_focus=true)


fmoc:
The link doesn't point anywhere. Please make sure you merge this into `master`, otherwise the change won't take effect.

fmoc (approved):

PR #9226, opened 2021-11-19
Reverse order of migrations

### LINKED ISSUE #9224: Branded migration might migrate the wrong profile
Migration from 2.4 might have failed in 2.8.
This resulted in a new configuration used in 2.8.

In 2.9 we fixed the migration, but we migrate the abandoned 2.4 config to 2.9 instead of the 2.8.

This only happened with branded clients, due to the change in the config location in the update from 2.8 to 2.9.

fmoc:
Should fix #9224.

### DIFF
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -77,6 +77,8 @@ void migrateConfigFile(const QCoreApplication *app)
         // we support multiple locations from old versions
         // these are worked on in-order to upgrade from version to version
         // the algorithm is the same for all these locations, thus we can use a loop
+        // note that we try to migrate in descending order, i.e., we try to migrate from the last release, then from the release before, ...
+        // this is done in order to avoid porting old configu
         QStringList configLocationsToMigrate;
 
         {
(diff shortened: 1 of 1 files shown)


src/gui/application.cpp
```diff
@@ -72,11 +72,15 @@ void migrateConfigFile(const QCoreApplication *app)
 {
     using namespace OCC;
 
+    static const auto logPrefix = QStringLiteral("Settings migration: ");
```
TheOneRing, line 75: Maybe Q_FUNC_INFO?

src/gui/application.cpp
```diff
@@ -89,47 +93,62 @@ void migrateConfigFile(const QCoreApplication *app)
...
                         for (const auto &fileInfo : filesList) {
                             if (!QFile::rename(fileInfo.canonicalFilePath(), confDir + "/" + fileInfo.fileName())) {
-                                qCWarning(lcApplication) << "Fallback move of " << fileInfo.fileName() << "also failed";
+                                qCWarning(lcApplication) << logPrefix << "Fallback move of " << fileInfo.fileName() << "also failed";
                             }
                         }
                     }
                 } else {
+                    // migration worked!
+                    success = true;
```
TheOneRing, line 138: Both branches migrate, I guess we should break in both?

src/gui/application.cpp
fmoc: Wouldn't that create a lot more noise (because it includes technical information such as the class, the return type and the arguments) than necessary? We have not done this in the legacy settings migration either, but instead use a user-friendly string.

TheOneRing: Well the log messages are for the devs to read not the user.
It will also help to locate the origin, I mean you can grep for "Settings migration:" but `migrateConfigFile(const QCoreApplication *app)` will be more clear.

erikjv (approved):

PR #9227, opened 2021-11-19
Windows: Block system shutdown until we are properly shutdown

TheOneRing:


TheOneRing:
We now block for up to 5s, if we take longer we might get killed by the system.
https://docs.microsoft.com/en-us/windows/win32/shutdown/shutdown-changes-for-windows-vista

We could also use ShutdownBlockReasonCreate

src/gui/guiutility.cpp
```diff
@@ -38,47 +45,76 @@ Q_LOGGING_CATEGORY(lcUtility, "gui.utility", QtInfoMsg)
...
+    // We also block a system shutdown until we are properly shutdown our selfs
+    // In the unlikely case that we require more than 5s Windows will require a fullscreen message
+    // with our icon, title and the reason why we are blocking the shutdown.
+    new std::thread([] {
         WNDCLASS wc = {};
         wc.hInstance = GetModuleHandle(nullptr);
         wc.lpszClassName = L"ocWindowMessageWatcher";
+#if MIRALL_VERSION_MINOR > 9
+// TODO: for now we won't display a proper icon
+#error "Please add the QtWinExras dependency"
```
fmoc, line 68: `QtWinExtras`

src/gui/CMakeLists.txt
```diff
@@ -2,6 +2,10 @@ include(ECMAddAppIcon)
 
 find_package(Qt5 REQUIRED COMPONENTS Widgets)
 
+#if (WIN32)
```
fmoc, line 5: If this is an optional dependency, consider introducing a CMake variable allowing users to opt into adding it. Could also be used in the code to opt into setting that icon.

src/gui/guiutility.cpp
```diff
@@ -38,47 +45,76 @@ Q_LOGGING_CATEGORY(lcUtility, "gui.utility", QtInfoMsg)
...
             } else if (msg == WM_ENDSESSION) {
-                qCInfo(OCC::lcUtility) << "Received WM_ENDSESSION quitting";
-                QTimer::singleShot(0, qApp, &QApplication::quit);
+                qCDebug(OCC::lcUtility) << "Received WM_ENDSESSION quitting";
+                QMetaObject::invokeMethod(qApp, &QApplication::quit);
+                QElapsedTimer shutdownTimer;
+                if (lParam == ENDSESSION_LOGOFF) {
+                    // block the windows shutdown until we are done
+                    const QString description = QApplication::translate("Utility", "Shutting down %1").arg(Theme::instance()->appNameGUI());
+                    OC_ASSERT(ShutdownBlockReasonCreate(hwnd, reinterpret_cast<const wchar_t *>(description.utf16())));
```
fmoc, line 83: The Windows docs recommend to provide a short, reasonable explanation why the shutdown is postponed. For instance, "Closing SQL database properly".

src/gui/guiutility.cpp
```diff
@@ -38,47 +45,76 @@ Q_LOGGING_CATEGORY(lcUtility, "gui.utility", QtInfoMsg)
...
                 return 0;
             }
             return DefWindowProc(hwnd, msg, wParam, lParam);
         };
-
         OC_ASSERT(RegisterClass(&wc));
 
-        auto window = CreateWindowW(wc.lpszClassName, L"watcher", WS_OVERLAPPED, CW_USEDEFAULT, CW_USEDEFAULT, CW_USEDEFAULT, CW_USEDEFAULT, nullptr, nullptr, wc.hInstance, nullptr);
-        OC_ASSERT_X(window, Utility::formatWinError(GetLastError()).toUtf8().constData());
+        auto watcherWindow = CreateWindowW(wc.lpszClassName, reinterpret_cast<const wchar_t *>(Theme::instance()->appNameGUI().utf16()),
```
fmoc, line 97: You need the app name more than once. Might be easier to create a variable for it.

src/gui/guiutility.cpp
TheOneRing: Way to technical

src/gui/CMakeLists.txt
TheOneRing: I don't consider it optional.
Optional dependencies never solved an issue, optional Qt modules certainly never did that.

src/gui/guiutility.cpp
TheOneRing: Different scopes, and the lambda can't capture anything.

erikjv (approved):

PR #9228, opened 2021-11-19
Fix crash in PropagateRemoteMkdir

TheOneRing:
Fixes: #9170

### DIFF
--- a/src/libsync/propagateremotemkdir.cpp
+++ b/src/libsync/propagateremotemkdir.cpp
@@ -22,8 +22,11 @@
 #include <QFile>
 #include <QLoggingCategory>
 #include <QtConcurrent>
+
+using namespace std::chrono_literals;
+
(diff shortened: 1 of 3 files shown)


src/libsync/propagateremotemkdir.cpp
```diff
@@ -22,8 +22,11 @@
 #include <QFile>
 #include <QLoggingCategory>
 #include <QtConcurrent>
+
+using namespace std::chrono_literals;
+
 namespace {
-auto UpdateMetaDataRetyTimeOut = std::chrono::seconds(30);
+auto UpdateMetaDataRetyTimeOut = 30s;
```
erikjv, line 29: const or constexpr

erikjv (changes requested):

erikjv (approved):

PR #9229, opened 2021-11-19
Work/fix crash while handling ssl errors

erikjv:


src/libsync/account.cpp
```diff
@@ -421,12 +430,20 @@ void Account::slotHandleSslErrors(QPointer<QNetworkReply> reply, const QList<QSs
         }
         return {};
     };
+
+    // Call `handleErrors` NOW, BEFORE checking if the scoped `reply` pointer. The lambda might take
+    // a long time to complete: if a dialog is shown, the user could be "slow" to click it away, and
+    // the reply might have been deleted at that point. So if we'd do this call inside the if below,
+    // the object inside the reply guarded pointer could there, but might be gone *after* we finish
+    // with `ignoreSslErrors`.
```
TheOneRing, line 438: Please add a line that explains why handling the error makes sense even if the reply might already be gone.

TheOneRing (changes requested):

TheOneRing (approved):

Issue #9222, opened 2021-11-18
[gui][tests-only] Passing tests that are in `Scenario Outline` are also reported as failed whenever one or more tests fail.


saw-jan:
When the tests are in Scenario Outline, they also get reported as failed (but they actually pass) whenever some other test fails.
Example log: from https://drone.owncloud.com/owncloud/client/9689/2/14
```json
[{
"Error Details": "Exception: Step execution through test middleware failed. Error: AssertionError [ERR_ASSERTION]: Expiry Date Mismatch: [object Object]\n    at Object.assertUserLastPublicShareDetails (/drone/src/middleware/src/helpers/sharingHelper.js:431:16)\n    at runMicrotasks (<anonymous>)\n    at processTicksAndRejections (internal/process/task_queues.js:95:5)\n    at async /drone/src/middleware/src/app.js:66:7",
"Feature": "Sharing",
"Feature File": "tst_sharing",
"Scenario": "user changes the expiration date of an already existing public link using webUI",
"Test Step": "Then the fields of the last public link share response of user \"Alice\" should include on the server"
 },
{
"Error Details": "Error details not found",
"Feature": "Sharing",
"Feature File": "tst_sharing",
"Scenario": "simple sharing of folder by public link with different roles",
"Test Step": "role=Viewer, permissions=read"
},
{
"Error Details": "Error details not found",
"Feature": "Sharing",
"Feature File": "tst_sharing",
"Scenario": "change collaborator permissions of a file & folder",
"Test Step": "permissions=edit, expected-folder-permission=read, share, expected-file-permission=read, share"
}]
```
Here, the first test is the failed scenario. But the following two scenarios are actually passing. They are reported in the error log because those tests are in `Scenario Outline`.

I separated those scenario outline tests each in a separate `scenario` and ran the CI. The following is the error report:
from https://drone.owncloud.com/owncloud/client/9690/2/14
```json
[{
"Error Details": "Exception: Step execution through test middleware failed. Error: AssertionError [ERR_ASSERTION]: Expiry Date Mismatch: [object Object]\n    at Object.assertUserLastPublicShareDetails (/drone/src/middleware/src/helpers/sharingHelper.js:431:16)\n    at runMicrotasks (<anonymous>)\n    at processTicksAndRejections (internal/process/task_queues.js:95:5)\n    at async /drone/src/middleware/src/app.js:66:7",
"Feature": "Sharing",
"Feature File": "tst_sharing",
[...]

PR #9232, opened 2021-11-22
Fix TestLogParser for scenario outline tests

swoichha:
### Description
This PR fixes gui-tests log that reports passing Scenario Outline tests as failed whenever one or more tests fail.

### Related Issue
fixes https://github.com/owncloud/client/issues/9222

### DIFF
--- a/test/gui/TestLogParser.py
+++ b/test/gui/TestLogParser.py
@@ -32,29 +32,63 @@ def traverse_loop(data):
                                 if 'result' in error and error['result'] == 'PASS':
                                     continue
 
-                                # Append the information of failing tests into the list of failing tests
-                                # If the error detail is missing(occurs mainly in runtime error) then we display the entire error object.
-                                test = {
-                                    "Feature File": str(feature_file['name']),
-                                    "Feature": str(feature['name']),
-                                    "Scenario": str(scenario['name']),
-                                    "Test Step": str(test_step['name']),
-                                    "Error Details": str(error['detail'])
-                                    if ('detail' in error)
-                                    else "Error details not found",
-                                }
+                                # Again, we will have to loop through the 'tests' to get failing tests from Scenario Outlines.
+                                if "tests" in error and len(error['tests']) > 0:
+                                    for outlineStep in error['tests']:
+                                        # Append the information of failing tests into the list of failing tests
+                                        if (
+                                            'result' in outlineStep
+                                            and outlineStep['result'] == 'ERROR'
+                                        ):
+                                            failing_test = {
+                                                "Feature File": str(
+                                                    feature_file['name']
+                                                ),
+                                                "Feature": str(feature['name']),
+                                                "Scenario": str(scenario['name']),
+                                                "Example": str(test_step['name']),
+                                                "Test Step": str(error['name']),
+                                                "Error Details": str(
+                                                    outlineStep['detail']
+                                                )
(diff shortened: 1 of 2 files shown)


saw-jan:
Now the error parser is working fine for `Scenario` and `Scenario Outline` tests.
Example build: https://drone.owncloud.com/owncloud/client/9863/1/14

test/gui/TestLogParser.py
```diff
@@ -32,29 +32,60 @@ def traverse_loop(data):
...
-                                    "Feature": str(feature['name']),
-                                    "Scenario": str(scenario['name']),
-                                    "Test Step": str(test_step['name']),
-                                    "Error Details": str(error['detail'])
-                                    if ('detail' in error)
-                                    else "Error details not found",
-                                }
-
-                                failing_tests.append(test)
+                                if 'result' in error and error['result'] == 'ERROR':
```
Talank, line 61: Is it always `ERROR`? Are you sure that there are no other value such as `FAIL` instead of `ERROR`?

test/gui/tst_syncing/test.feature
```diff
@@ -157,7 +157,7 @@ Feature: Syncing files
 
 
     Scenario: Many subfolders can be synced
-        Given user "Alice" has created folder "parent" on the server 
+        Given user "Alice" has created folder "parent" on the server
```
Talank, line 160: you might not need this change in the feature file

test/gui/TestLogParser.py
saw-jan: Yeah, It's always `ERROR` whenever the step fails.

test/gui/tst_syncing/test.feature
saw-jan: Yeah, that's by formatter. It won't hurt the test so I think it's ok-ish.

test/gui/TestLogParser.py
Talank: the comment says `If the error detail is missing(occurs mainly in runtime error) then we display the entire error object.` Where are you doing this here? And how does it look like (the final parsed log)? Can you comment the screenshot or something?

saw-jan: I don't know exactly what that comment was for. Maybe it means to display `error` object rather than `Error details not found` whenever there is no `detail` in error.

saw-jan:
Example test log is attached below. If scenario outline step fails then the error details will be in the `tests` property.
[testlog.txt](https://github.com/owncloud/client/files/7648859/testlog.txt)

test/gui/TestLogParser.py
```diff
@@ -32,29 +32,61 @@ def traverse_loop(data):
...
 
     for scenario in raw_data:
-        if scenario['Scenario'] not in unique_scenarios:
+        if (
+            'Example' in scenario
+            and scenario['Example'] not in unique_scenario_outline_examples
+        ):
+            unique_scenario_outline_examples.append(scenario['Example'])
+            filtered_data.append(scenario)
+        elif 'Example' not in scenario and scenario['Scenario'] not in unique_scenarios:
```
Talank, line 89: can you add a line of comment about this?

saw-jan: Of course

Talank (commented):
LGTM

individual-it (approved):

PR #9237, opened 2021-11-23
Ensure we prooperly close the log file

TheOneRing:
Now the question remains, will this post routine be called first or https://github.com/owncloud/client/blob/master/src/gui/guiutility.cpp#L114

The log file could also be parented with qApp.

### DIFF
--- a/src/libsync/logger.cpp
+++ b/src/libsync/logger.cpp
@@ -16,6 +16,7 @@
 
 #include "config.h"
 
+#include <QCoreApplication>
 #include <QDir>
 #include <QStringList>
 #include <QtGlobal>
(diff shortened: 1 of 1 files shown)


src/libsync/logger.cpp
```diff
@@ -52,8 +53,15 @@ namespace OCC {
 
 Logger *Logger::instance()
 {
-    static Logger log;
-    return &log;
+    static auto *log = [] {
+        auto log = new Logger;
+        qAddPostRoutine([] {
```
dragotin, line 58: Can't that be avoided by having the `close()` command in the destructor of the Logger?

I always have to look up these Qt special routines, and also ... as you mention ... which qPostRoutine() is called first? Can you have multiple at all?

TheOneRing: You can have multiple, the thing is that the watcher also uses the logger.

erikjv (approved):

erikjv:
> https://github.com/owncloud/client/blob/master/src/gui/guiutility.cpp#L114

Can't you move that into the `Platform` destructor? That will be called on exit from `main()`, so before any PostRoutine is executed. That should also be before the log file is closed.

TheOneRing:
> > https://github.com/owncloud/client/blob/master/src/gui/guiutility.cpp#L114
> 
> Can't you move that into the `Platform` destructor? That will be called on exit from `main()`, so before any PostRoutine is executed. That should also be before the log file is closed.

Will do so on master

Issue #9165, opened 2021-10-22
Kill owncloudcmd if file for unsyncedfolders can't be found


Spanni26:
owncloudcmd only gives a critical warning, when the file for "--unsyncedfolders" can't be found. But it does the whole sync.

In my opinion owncloudcmd should die, if the file can't be found.

erikjv:
Fixed in 2.10

PR #9241, opened 2021-11-25
Bugfix: Do not sync when unsyncedfolders file cannot be read

erikjv:
owncloudcmd now checks if the file specified by --unsyncedfolders exists
and can be read, before starting the sync. If it does not exist, show an
error message and quit immediately.

https://github.com/owncloud/client/issues/9165

### DIFF
--- a/src/cmd/cmd.cpp
+++ b/src/cmd/cmd.cpp
@@ -237,7 +237,7 @@ public:
 #else
         tcgetattr(STDIN_FILENO, &tios);
         termios tios_new = tios;
-        tios_new.c_lflag &= ~ECHO;
+        tios_new.c_lflag &= ~static_cast<tcflag_t>(ECHO);
         tcsetattr(STDIN_FILENO, TCSANOW, &tios_new);
 #endif
     }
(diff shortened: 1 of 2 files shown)


TheOneRing:
Please target 2.10 and add a change log entry

erikjv:
Done

src/cmd/cmd.cpp
```diff
@@ -533,6 +533,19 @@ int main(int argc, char **argv)
...
     }
 
+    // Pre-flight check: verify that the file specified by --unsyncedfolders can be read by us:
+    if (!ctx.options.unsyncedfolders.isNull()) { // yes, isNull and not isEmpty because...:
+        // ... if the user entered "--unsyncedfolders ''" on the command-line, opening that will
+        // also fail
+        QFile f(ctx.options.unsyncedfolders);
+        if (!f.open(QFile::ReadOnly)) {
+            qFatal("Cannot read unsyncedfolders file '%s': %s",
+                qPrintable(ctx.options.unsyncedfolders),
```
TheOneRing, line 543: In the gui part I replaced all qPrintable calls with qUtf8Printable, which makes sense as we have a log file that is capable of printing this.
When I did the change I also replaced the qPrintable  in here, probable won't work but we should keep to one?
So maybe change the others back?

erikjv: I changed the other to `qPrintable`.

TheOneRing (approved):

PR #9243, opened 2021-11-26
Rename a folder's navigationPaneClsid to uuid

erikjv:
So we can use it in more places, not only on Windows.

### DIFF
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -537,7 +537,8 @@ void Folder::startVfs()
     }
 
     VfsSetupParams vfsParams;
-    vfsParams.filesystemPath = path();
+    vfsParams.uuid = uuid();
+    vfsParams.filesystemPath = path(); // This is ignored by the macOS VFS plug-in
     vfsParams.remotePath = remotePathTrailingSlash();
     vfsParams.account = _accountState->account();
     vfsParams.journal = &_journal;
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -1333,54 +1333,55 @@ void FolderDefinition::save(QSettings &settings, const FolderDefinition &folder)
     }
 }
 
-bool FolderDefinition::load(const QSettings &settings, const QString &alias, FolderDefinition *folder)
+FolderDefinition FolderDefinition::load(const QSettings &settings, const QString &alias)
 {
-    folder->alias = FolderMan::unescapeAlias(alias);
-    folder->localPath = settings.value(localPathC).toString();
-    folder->journalPath = settings.value(journalPathC).toString();
-    folder->targetPath = settings.value(targetPathC).toString();
-    folder->paused = settings.value(pausedC).toBool();
-    folder->ignoreHiddenFiles = settings.value(ignoreHiddenFilesC, QVariant(true)).toBool();
-    folder->isInNavigationPane = settings.value(isInNavigationPaneC, QVariant(true)).toBool();
+    FolderDefinition folder;
+    folder.alias = FolderMan::unescapeAlias(alias);
+    folder.localPath = settings.value(localPathC).toString();
+    folder.journalPath = settings.value(journalPathC).toString();
+    folder.targetPath = settings.value(targetPathC).toString();
+    folder.paused = settings.value(pausedC).toBool();
+    folder.ignoreHiddenFiles = settings.value(ignoreHiddenFilesC, QVariant(true)).toBool();
+    folder.isInNavigationPane = settings.value(isInNavigationPaneC, QVariant(true)).toBool();
 
     if (settings.contains(uuidC)) {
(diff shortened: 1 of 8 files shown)


src/gui/folder.cpp
```diff
@@ -64,6 +64,18 @@ auto versionC()
 
 constexpr int WinVfsSettingsVersion = 4;
 constexpr int SettingsVersion = 2;
+
+// Settings keys:
+const char localPathC[] = "localPath";
+const char journalPathC[] = "journalPath";
+const char targetPathC[] = "targetPath";
+const char pausedC[] = "paused";
+const char ignoreHiddenFilesC[] = "ignoreHiddenFiles";
```
TheOneRing, line 73: Could you directly make them QStrings/QLatin1Strings here?

src/gui/folder.cpp
```diff
@@ -64,6 +64,18 @@ auto versionC()
 
 constexpr int WinVfsSettingsVersion = 4;
 constexpr int SettingsVersion = 2;
```
TheOneRing, line 66: Thats the version written to the settings so you need to bump both to 5 I assume or better unify them.

erikjv: Where is the other one?

erikjv: done

src/gui/folder.cpp
erikjv: done

src/gui/folder.cpp
TheOneRing: the two lines above.

erikjv: I don't think it's needed, see the commit message.

src/gui/folder.cpp
```diff
@@ -1306,47 +1315,66 @@ void Folder::slotAboutToRemoveAllFiles(SyncFileItem::Direction dir, std::functio
...
+    }
+
+    // Sanity check:
+    if (folder->uuid.isNull()) {
+        folder->uuid = QUuid::createUuid();
+    }
 
     folder->virtualFilesMode = Vfs::Off;
-    QString vfsModeString = settings.value(QStringLiteral("virtualFilesMode")).toString();
+    QString vfsModeString = settings.value(QLatin1String(virtualFilesModeC)).toString();
```
TheOneRing, line 1369: should already be a QLatin1Srting?

src/gui/folder.cpp
```diff
@@ -1306,47 +1315,66 @@ void Folder::slotAboutToRemoveAllFiles(SyncFileItem::Direction dir, std::functio
...
     if (!vfsModeString.isEmpty()) {
         if (auto mode = Vfs::modeFromString(vfsModeString)) {
             folder->virtualFilesMode = *mode;
         } else {
             qCWarning(lcFolder) << "Unknown virtualFilesMode:" << vfsModeString << "assuming 'off'";
         }
     } else {
-        if (settings.value(QLatin1String("usePlaceholders")).toBool()) {
+        if (settings.value(QLatin1String(usePlaceholdersC)).toBool()) {
```
TheOneRing, line 1377: should already be a QLatin1Srting?

src/gui/folder.cpp
TheOneRing: I think the uiid was only used when with Windows vfs disabled, with vfs we use a different mechanism.
So you need to bump the non vfs version too.
Thats why I suggested to merge both and only use one version independent of vfs.

TheOneRing (changes requested):

src/gui/folder.cpp
erikjv: yes, done.

src/gui/folder.cpp
erikjv: yes, done.

src/gui/folder.cpp
erikjv: I merged them. Please check if I did it correctly.

TheOneRing (approved):

src/gui/folder.h
```diff
@@ -91,8 +94,9 @@ class FolderDefinition
      *            (version remains readable by 2.5.1)
      * Version 3: introduction of new windows vfs mode in 2.6.0
      * Version 4: until 2.9.1 windows vfs tried to unregister folders with a different id from windows.
+     * Version 5: renamed navigationPaneClsid to uuid in the settings for FolderDefinition
      */
-    static int maxSettingsVersion() { return 4; }
+    static int maxSettingsVersion() { return 5; }
```
TheOneRing, line 99: Hmm can we merge this one now with the one we write to the settings?

TheOneRing (approved):

TheOneRing:
@erikjv please rebase or close

TheOneRing:
@erikjv

PR #9252, opened 2021-12-02
Rename stop and unregisterFolder methods

erikjv:
In case of `stop` this makes it clear if it is a temporary state
and the connection is resumed later (e.g. when the client exits),
or if it is permanent (e.g. when the sync older is going to be
removed).

### DIFF
--- a/src/common/vfs.h
+++ b/src/common/vfs.h
@@ -142,11 +142,11 @@ public:
      */
     void start(const VfsSetupParams &params);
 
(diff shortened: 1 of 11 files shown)


TheOneRing:
@erikjv ?

TheOneRing:
@erikjv please decide on whether you want to continue with this

erikjv:
This should go in together with https://github.com/owncloud/client-desktop-vfs-win/pull/7 . I'm postponing this until after getting the tests running on windows.

PR #9254, opened 2021-12-03
reduce time waiting for password box

individual-it:
the test code waits for the password box, which is not comming most of the time, because the password is stored in the keyring after the first test.
We can save a lot of test time by reducing that waiting


Talank (approved):
looks good, it should save significant test execution time. And looking at the test run in drone (so far), I think it is ok.

PR #9263, opened 2021-12-08
add "Upload linux keys to key server"

gabi18:
After first build of a new release the extended linux keys have to be uploaded to the key server (esp. for rpm-based linux distributions)


jnweiger (approved):
sure thing.

PR #9265, opened 2021-12-08
Cover WinVFS in testcases

erikjv:


test/testsyncmove.cpp
```diff
@@ -841,21 +841,73 @@ private slots:
...
+        {
+            auto localState = fakeFolder.currentLocalState();
+            QVERIFY(localState.find("A/file") == nullptr); // check if the file is gone
+            QVERIFY(localState.find("A") == nullptr); // check if the directory is gone
+            FileInfo *file = localState.find({ "B/file" });
+            QVERIFY(file != nullptr); // check if the file exists
+            if (vfsMode != Vfs::Off) {
+                QVERIFY(file->isDehydratedPlaceholder); // check that no-one messed with the placeholder state
+            }
+            QVERIFY(localState.equals(fakeFolder.currentRemoteState(), cmpOpt));
```
TheOneRing, line 909: The benefit of a QCOMPARE here is that we had human readable errors in the log

test/testutils/syncenginetestutils.cpp
```diff
@@ -254,9 +255,38 @@ bool FileInfo::operator==(const FileInfo &other) const
...
+            && children == other.children;
+}
+
+bool FileInfo::equals(const FileInfo &other, ComparissonOption opt) const
+{
+    switch (opt) {
+    case ContentIsKing:
+        return *this == other;
+    case IgnoreContentOfDehydratedFiles:
+        if (!isDehydratedPlaceholder && !other.isDehydratedPlaceholder) {
```
TheOneRing, line 269: please also check modtime and other available meta data.

test/testsyncmove.cpp
TheOneRing: hmm would be awesome if we could wrap all this complexity behind some magic else our tests complexity will increase exponential 🤯

erikjv: I'm wondering if we should get rid of the old `==` implementation. But I don't know how many tests that'd break.

test/testutils/syncenginetestutils.cpp
erikjv: which ones, other than `lastModified`?

TheOneRing: Well you are right we only have the modtime.

TheOneRing (approved):

PR #9266, opened 2021-12-08
Fix hiding of activity tab in case app is disabled or not installed

### LINKED ISSUE #9260: Activity tab is shown when the server does not provide it
https://github.com/owncloud/enterprise/issues/4914

fmoc:
Fixes #9260.


fmoc:
Not entirely sure why the behavior on the server side has changed. @michaelstingl should we ping the server team about it?

TheOneRing (approved):

PR #9268, opened 2021-12-09
Tests: simplify equality check for state

erikjv:
Always take the hydration state into account, and always check the
mtime.


TheOneRing (changes requested):
Tests are failing

TheOneRing (approved):

Issue #9208, opened 2021-11-11
Possible endless loop when conflict file is locked


TheOneRing:
If the client tries to rename a file to its conflict file name and that file is currently locked, the client fails, still creates an error message and immediately tries again.

The result is a flickering ui and a log full of hundreds of errors.

dragotin:
Same if the a file is locked (ie. in Office) and the user clicks on "Dehydrate files" using Windows VFS. That results in endless loops trying to dehydrate the file which fails until it is unlocked.

gabi18:
Tested ownCloud-2.10.0-daily20211214.6294.x64.msi on Windows10 

Steps done:
- open 'New text file.odt' in LibreOffice
- edit and save

Result:
Client shows 'i' and info that file is locked in 'Not synced' tab
![client-2-10-libreoffice-save](https://user-images.githubusercontent.com/75983861/146154683-5f0c5972-9732-49ed-a798-db833fe76d97.png)

- 'Free up space' on right click on file 'New text file.odt'
Result: Client shows 'i' - no flickering, no endless loop of retry -> OK

- 'Free up space' on folder 'Neuer Ordner (2)'
Result: nothing changed

- Exit LibreOffice
Result: 'New text file.odt'  is synced to server, client shows green check mark -> OK

![client-2-10-libreoffice-close](https://user-images.githubusercontent.com/75983861/146155615-3bd46d9f-7272-40a8-b8b4-4d8f9505c5c2.png)

cloud icon in Explorer for fille 'New text file.odt' -> OK
Spinning icon remains for folder 'Neuer Ordner (2)', also after  'Force sync' -> NOT ok

gabi18:
Tested ownCloud-2.10.0-daily20211216.6305.x64.msi on Windows10 20H2 (Build 19042.1348)

- file opened in LibreOffice
- free up space -> OK, no endless retry (spinning icon for file and folder)
- exit LibreOffice -> folder: cloud icon, file: full file

![2 10-folder+file-ok](https://user-images.githubusercontent.com/75983861/146374053-fb2751e3-1b52-4954-89c9-ac135af69292.png)

Issue #9076, opened 2021-09-23
Internal: 4758


TheOneRing:
https://github.com/owncloud/enterprise/issues/4758

Issue #9133, opened 2021-10-08
Checksum retries on retry-list


michaelstingl:
In case downloads are discarded because of checksum mismatch, clients retry again and again and again. Retry-list standard behaviour would better, to retry downloads after increasing time intervals, up to 24h.

PR #9269, opened 2021-12-09
Handle local errors with the ignore list

TheOneRing:
Fixes: #9208, #9133

Might also fix https://github.com/owncloud/client/issues/9076

### DIFF
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -95,6 +95,9 @@ Folder::Folder(const FolderDefinition &definition,
 
     _syncResult.setFolder(_definition.alias);
 
+    // those errors should not persist over sessions
+    _journal.wipeErrorBlacklistCategory(SyncJournalErrorBlacklistRecord::Category::LocalSoftError);
+
     _engine.reset(new SyncEngine(_accountState->account(), path(), remotePath(), &_journal));
     // pass the setting if hidden files are to be ignored, will be read in csync_update
     _engine->setIgnoreHiddenFiles(_definition.ignoreHiddenFiles);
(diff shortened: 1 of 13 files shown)


src/gui/folder.cpp
```diff
@@ -95,6 +95,9 @@ Folder::Folder(const FolderDefinition &definition,
 
     _syncResult.setFolder(_definition.alias);
 
+    // those errors should not persist over sessions
+    _journal.wipeErrorBlacklistCategory(SyncJournalErrorBlacklistRecord::Category::LocalSoftError);
```
dragotin, line 99: Would somehow be cool if we had a logging if there were blacklisted errors from the previous run. That might be an important hint when debugging. No idea how difficult that is...

src/libsync/owncloudpropagator.cpp
```diff
@@ -213,24 +209,13 @@ static void blacklistUpdate(SyncJournalDb *journal, SyncFileItem &item)
...
     }
 
-    // Some soft errors might become louder on repeat occurrence
-    if (item._status == SyncFileItem::SoftError
-        && newEntry._retryCount > 1) {
-        qCWarning(lcPropagator) << "escalating soft error on " << item._file
-                                << " to normal error, " << item._httpErrorCode;
-        item._status = SyncFileItem::NormalError;
-        return;
-    }
```
dragotin, line 233: Hmm, this is going away completely? Which soft errors were ending up here, that were escalated to NormalErrors before?

dragotin (approved):
LGTM, only the second comment pls.

src/gui/folder.cpp
TheOneRing: Hm true, in general `wipeErrorBlacklist*` should log what was wiped

src/libsync/owncloudpropagator.cpp
TheOneRing: Hmm maybe I should look for _httpErrorCode that way http errors will behave as before.

dragotin (approved):
LGTM

Issue #9196, opened 2021-11-03
OAuth2: Restart local socket after an error occured


michaelstingl:
### Expected behaviour
Login attempt after "Error returned from the server: code verifier invalid" should succeed.

### Actual behaviour
Invalid redirect URI with `localhost:0` is generated

### Steps to reproduce
1. Login to demo.owncloud.com with user `demo:demo` in system browser
2. Add new server account with demo.owncloud.com to desktop client
3. "Switch User" in web UI (OAuth 2.0 app)
4. Login with user `test:test` in web UI
5. "Authorise" in web UI, to authorise desktop client
==> **Login Error** Error returned from the server: code verifier invalid
6. Go back to desktop client, click "Re-open Browser"
==> New URL contain invalid redirect URI with `localhost:0`

### Server configuration

```
curl https://demo.owncloud.com/status.php
{"installed":true,"maintenance":false,"needsDbUpgrade":false,"version":"10.8.0.4","versionstring":"10.8.0","edition":"Enterprise","productname":"ownCloud"}
```

### Client configuration
```
testpilotcloud 2.9.1 (build 5501) 42fb29 Oct 13 2021 10:57:21
Libraries Qt 5.12.11, OpenSSL 1.1.1k 25 Mar 2021
Using virtual files plugin: suffix
osx-21.1.0
```

Related: https://github.com/owncloud/oauth2/issues/309

TheOneRing:
Turned out to not be a client issue.

saw-jan:
@michaelstingl, I have tested this behavior manually as of https://github.com/owncloud/client/issues/9290#issuecomment-994566663
My findings:
- :exclamation: step 5 has the same error message `Login Error Error returned from the server: code verifier invalid` redirected to localhost:
- :heavy_check_mark: step 6 redirects to the correct server URL not to localhost

PR #9275, opened 2021-12-10
Restart oauth on error

TheOneRing:
Fixes: #9196

### DIFF
--- a/changelog/unreleased/9196
+++ b/changelog/unreleased/9196
@@ -1,0 +1,6 @@
+Bugfix: Always restart OAuth2 on error
+
+We now always restart the OAuth2 process once we got a result.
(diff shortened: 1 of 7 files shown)


erikjv (approved):

changelog/unreleased/9196
```diff
@@ -0,0 +1,6 @@
+Bugfix: Always restart OAuth2 on error
+
+We now always restart the OAuth2 process once we got a result.
```
erikjv, line 3: "We now always reset the OAuth2 status before processing the result of an authentication request."

It took me a bit to understand what the intention of this change was ;-)

TheOneRing: Its a bit late now after the release :P

How ever you can still provide a pull request for the change log item

PR #9280, opened 2021-12-13
[Tests-Only] add test for folder sync

swoichha:
### Description
This PR adds tests for Creating a new folder with space at the end and sync it.

### Related Issue
https://github.com/owncloud/client/issues/8971#issuecomment-909884918


Talank (approved):
Since we already had similar scenario, adding an example in the scenario outline will do the job.

saw-jan (approved):
LGTM

PR #9282, opened 2021-12-13
[tests-only]Add sharing test

swoichha:
### Description
This PR adds following tests:
- Share a file with user B with Can Edit permissions enabled. Login as User B and delete it
### Related Issue
https://github.com/owncloud/client/issues/8971#issuecomment-909885166

### DIFF
--- a/test/gui/tst_sharing/test.feature
+++ b/test/gui/tst_sharing/test.feature
@@ -301,3 +301,17 @@ Feature: Sharing
             | edit        | read, share                  | read, share              |
             | share       | read, update, create, delete | read,update              |
             | edit,share  | read                         | read                     |
+
+
+    Scenario: sharee deletes a file shared by sharer
+        Given the setting "shareapi_auto_accept_share" on the server of app "core" has been set to "yes"
+        And the administrator on the server has set the default folder for received shares to "Shares"
+        And user "Alice" has uploaded file with content "ownCloud test text file 0" to "textfile.txt" on the server
+        And user "Brian" has been created on the server with default attributes and without skeleton files
+        And user "Alice" has shared file "textfile.txt" on the server with user "Brian" with "all" permissions
+        And user "Brian" has set up a client with default settings
+        When the user waits for file "Shares/textfile.txt" to be synced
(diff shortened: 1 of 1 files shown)


test/gui/shared/steps/steps.py
```diff
@@ -997,6 +997,7 @@ def step(context, errorMsg):
 
 @When(r'the user deletes the (file|folder) "([^"]*)"', regexp=True)
 def step(context, itemType, resource):
+    waitForFileToBeSynced(context, resource)
```
saw-jan, line 1000: I think it will be better to use step instead of here.
```suggestion
```

test/gui/tst_sharing/test.feature
```diff
@@ -301,3 +301,16 @@ Feature: Sharing
...
+
+    Scenario: sharee deletes a file shared by sharer
+        Given the setting "shareapi_auto_accept_share" on the server of app "core" has been set to "yes"
+        And the administrator on the server has set the default folder for received shares to "Shares"
+        And user "Alice" has uploaded file with content "ownCloud test text file 0" to "textfile.txt" on the server
+        And user "Brian" has been created on the server with default attributes and without skeleton files
+        And user "Alice" has shared file "textfile.txt" on the server with user "Brian" with "all" permissions
+        And user "Brian" has set up a client with default settings
+        When the user deletes the file "Shares/textfile.txt"
+        And the user waits for the files to sync
```
saw-jan, line 314: related to above comment
```suggestion
        And the user waits for file "Shares/textfile.txt" to be synced
```

saw-jan: See if it works.

swoichha: This step won't work here as we have deleted file `Shares/textfile.txt` in line 313 but as you mentioned in the comment above we can use this to check if the file has been synced or not.

saw-jan: Ohh, Yeah!!

Talank (approved):
LGTM

saw-jan (approved):
LGTM

PR #9291, opened 2021-12-15
Remove Availability menu from the ui

TheOneRing:
@swoichha @Talank the pr removes the "Availability" entry from the context menu which causes the tests to fail.


erikjv (approved):

dschmidt (approved):
🤷🏻‍♂️

PR #9292, opened 2021-12-15
Enable VFS testing in more tests

erikjv:


erikjv:
Please check what should happen in `moveFileToDifferentFolderOnBothSides()`, because the test passes, but I have no idea what the 2 `PUT` requests contain...

TheOneRing:
Maybe metadata changes

TheOneRing:
> Please check what should happen in `moveFileToDifferentFolderOnBothSides()`, because the test passes, but I have no idea what the 2 `PUT` requests contain...

If  you understand the difference please add a short comment to explain it.

test/testutils/syncenginetestutils.cpp
```diff
@@ -920,16 +944,15 @@ QNetworkReply *FakeQNAM::createRequest(QNetworkAccessManager::Operation op, cons
     return reply;
 }
 
-FakeFolder::FakeFolder(const FileInfo &fileTemplate, OCC::Vfs::Mode vfsMode)
```
erikjv, line 923: @TheOneRing can you verify that this constructor is doing the correct thing, including wrt. the call to `switchVfs` and the surrounding code?

test/testsyncmove.cpp
```diff
@@ -188,25 +210,35 @@ private slots:
...
+        counter.reset();
 
-        // Move-and-change, causing a upload and delete
+        // Move-and-change, mtime+size, causing a upload and delete
+        QVERIFY(fakeFolder.currentLocalState().find("A/a2")->isDehydratedPlaceholder == filesAreDehydrated); // no-one touched it, so the hydration state should be the same as the initial state
         fakeFolder.localModifier().rename("A/a2", "A/a2m");
-        fakeFolder.localModifier().appendByte("A/a2m");
+        fakeFolder.localModifier().setContents("A/a2m", 'x', fakeFolder.remoteModifier().contentSize + 1);
         QVERIFY(fakeFolder.syncOnce());
+        QVERIFY(!fakeFolder.currentLocalState().find("A/a2m")->isDehydratedPlaceholder); // We overwrote all data in the file, so whatever the state was before, it is no longer dehydrated
```
erikjv, line 224: This fails for winvfs

TheOneRing:
I just changed the target to master as 2.10 is more or less closed by now.
Can you have a look at conflicts?

ownclouders:
Results for <strong>GUI-tests</strong> https://drone.owncloud.com/owncloud/client/11955/6/1

<details>


</details>

test/testchunkingng.cpp
```diff
@@ -62,25 +65,27 @@ private slots:
 
     void testFileUpload() {
         FakeFolder fakeFolder{FileInfo::A12_B12_C12_S12()};
+        fakeFolder.syncEngine().account()->setCapabilities({ { "dav", QVariantMap{ {"chunking", "1.0"} } } });
```
erikjv, line 68: @TheOneRing is this not needed anymore?

TheOneRing: yes it was removed see https://github.com/owncloud/client/blob/master/test/testutils/testutils.cpp#L35 and https://github.com/owncloud/client/blob/master/test/testutils/testutils.cpp#L35

test/testutils/syncenginetestutils.cpp
```diff
@@ -41,65 +41,123 @@ PathComponents PathComponents::subComponents() const &
...
+{
+    if (_processArguments.isEmpty()) {
+        return true;
+    }
+
+    HelperProcess helper(_rootDir, _processArguments);
+    helper.waitForStarted();
+    _processArguments.clear(); // the helper has a copy, we clean it now for the next run
+    do {
+        QThread::msleep(50);
```
TheOneRing, line 126: please use 50ms, as QThread has no overload for that  std::this_thread::sleep_for(50ms)
```suggestion
        std::this_thread::sleep_for(50ms);
```

test/testutils/syncenginetestutils.cpp
```diff
@@ -41,65 +41,123 @@ PathComponents PathComponents::subComponents() const &
...
+    // Non-classic mode:
+
+    HelperProcess p(_rootDir, _processArguments);
+    _processArguments.clear();
+
+    do {
+        // process any pending events
+        QThread::currentThread()->eventDispatcher()->processEvents(QEventLoop::AllEvents);
+        // wait a bit to get the helper started with it's work
+        QThread::msleep(50);
```
TheOneRing, line 150:
```suggestion
        std::this_thread::sleep_for(50ms);
```

test/testutils/test_helper.cpp
```diff
@@ -0,0 +1,475 @@
...
+
+    const QString _fileName;
+};
+
+Command::~Command() { }
+
+class SetMtimeCommand : public Command
+{
+public:
+    static const QString name;
```
TheOneRing, line 54: Hmm I guess we could use `constexpr std::string_view name` then the name could be inline?
(I don't like the name to be defined outside of the class).

test/testutils/test_helper.cpp
```diff
@@ -0,0 +1,475 @@
...
+        cerr << qPrintable(name) << endl;
+        QFile f(rootDir.filePath(_fileName));
+        if (!f.open(QIODevice::WriteOnly | QIODevice::Append)) {
+            cerr << "Error: cannot open file '" << qPrintable(_fileName) << "' for " << qPrintable(name) << " command: "
+                 << qPrintable(f.errorString()) << endl;
+            return false;
+        }
+        char ch = _ch;
+        if (ch == '\0') {
+            ch = f.readAll().at(0);
```
TheOneRing, line 231: f.read(1)?

TheOneRing (commented):
Could it make sense to switch from positional parsing to json in a follow up?
In my opinion the mental load for parsing the parsers might be less with json :D

test/testallfilesdeleted.cpp
```diff
@@ -71,6 +71,7 @@ private slots:
         for (auto it = children.cbegin(); it != children.cend(); ++it) {
             modifier.remove(it.key());
         }
+        QVERIFY(fakeFolder.applyLocalModificationsWithoutSync());
```
TheOneRing, line 74: That was previously done implicit? Or why is it needed now.

test/testutils/syncenginetestutils.cpp
erikjv: done

test/testutils/test_helper.cpp
erikjv: done

test/testutils/test_helper.cpp
erikjv: removed the read completely

test/testallfilesdeleted.cpp
erikjv: added a comment

PR #9296, opened 2021-12-16
Only check whether a file is locked before we start the upload

TheOneRing:
I tried to fix this before, but in the wrong uploader.

Looking at the code I assume that legacy chunking with parallel chunking was completely  broken in 2.9


erikjv (approved):

dschmidt (approved):
🤷🏻‍♂️

PR #9300, opened 2021-12-17
Fix unstable test

erikjv:


test/testallfilesdeleted.cpp
```diff
@@ -300,8 +300,10 @@ private slots:
     }
 
     void testSelectiveSyncNoPopup() {
+        auto original = FileInfo::A12_B12_C12_S12();
```
TheOneRing, line 303: I understand the original is not to be modified, so maybe const?

TheOneRing (commented):
Good catch while including mtime in the tests proofs to be a bit of a pain I'm pretty sure it improved the quality of the tests.

test/testutils/syncenginetestutils.h
```diff
@@ -628,7 +628,7 @@ inline void addFiles(QStringList &dest, const FileInfo &fi)
         for (const auto &fi : fi.children)
             addFiles(dest, fi);
     } else {
-        dest += QStringLiteral("%1 - %2 %3-bytes").arg(fi.path()).arg(fi.contentSize).arg(fi.contentChar);
+        dest += QStringLiteral("%1 - %2 %3-bytes (%4)").arg(fi.path()).arg(fi.contentSize).arg(fi.contentChar).arg(fi.lastModifiedInUtc());
```
TheOneRing, line 631: .arg(, , , , )

Also the build seems to fail with this one.

test/testallfilesdeleted.cpp
erikjv: done

test/testutils/syncenginetestutils.h
erikjv: done

TheOneRing (approved):

Issue #9240, opened 2021-11-25
squish tests fail on master, update the version of squish used on master?


TheOneRing:
```
/drone/src/build-GUI-tests/bin/owncloud: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /drone/src/build-GUI-tests/bin/owncloud)
```

_Originally posted by @erikjv in https://github.com/owncloud/client/issues/9184#issuecomment-979003582_

individual-it:
We can update squish, but I think that is more about having a that version of the library actually to start the client.
Do different versions of the client need different versions of glibc?

TheOneRing:
Should not be the case.
@fmoc ^

Talank:
Do we need to upgrade qt5 to version `5.15` to compile client? I can compile the code on 2.9 and 2.10 but not the master! getting error duting the `cmake`
```
Could not find a configuration file for package "Qt5" that is compatible
  with requested version "5.15".

  The following configuration files were considered but not accepted:

    /usr/lib/x86_64-linux-gnu/cmake/Qt5/Qt5Config.cmake, version: 5.12.8
    /lib/x86_64-linux-gnu/cmake/Qt5/Qt5Config.cmake, version: 5.12.8
```

individual-it:
CI uses Qt 5.15
![grafik](https://user-images.githubusercontent.com/2425577/143430449-7766ce09-8fa6-4156-b914-e6235a034e5b.png)

TheOneRing:
Yes master requires Qt 5.15

individual-it:
Then this becomes a bigger task, we need to use different versions of squish in master and in the other repos.
The squish docker container needs to get tags, according to the used Qt version

individual-it:
I guess we need to make similar changes like @fmoc done in https://github.com/owncloud-ci/client/pull/13/files

individual-it:
WIP PR: https://github.com/owncloud-ci/squish/pull/26

individual-it:
strange even with the chages from  https://github.com/owncloud-ci/client/pull/13/files I get Qt 5.14

individual-it:
Qt 5.15 installed by using Ubuntu 21.10 and the client compiles fine but 21.10 comes with Python 3.9 and squish only supports 3.8
```
The Squish package has no support for Python 3.9 and cannot be configured for this interpreter.
The package contains support for the following Python version(s): 2.7.x & 3.8.x
[...]

PR #9302, opened 2021-12-18
Fix the startup period at beginning, make it short again.

### LINKED ISSUE #9253: first sync takes a long time
On start the client needs a long time to get going.
After starting it take around 1min to get out of the "waiting" phase. It sends multiple quota requests in the background and then all comes green. 

git bisect shows a0bf5f9fb91804693575975155f9e15575d33d7f as the first broken commit

dragotin:
Fixes #9253

### DIFF
--- a/changelog/unreleased/9253
+++ b/changelog/unreleased/9253
@@ -1,0 +1,8 @@
+Bugfix: Make first sync start fast again. 
+
+The first sync of the folders start now directly after startup instead of  
+after a full "pausing" interval.
+
+Also better say nothing than "Waiting..." which confuses users.
+
+https://github.com/owncloud/client/issues/9253
--- a/src/gui/folderman.cpp
+++ b/src/gui/folderman.cpp
@@ -817,7 +817,7 @@ void FolderMan::startScheduledSyncSoon()
         return;
     }
 
-    seconds delay;
+    seconds delay { 2s }; // Startup, if _lastSyncFolder is still empty.
     seconds sinceLastSync;
 
     // Require a pause based on the duration of the last sync run.
@@ -829,6 +829,8 @@ void FolderMan::startScheduledSyncSoon()
         //  1min -> 12s pause
         //  1h   -> 90s pause
         delay = seconds(static_cast<int64_t>(qSqrt(duration_cast<seconds>(lastFolder->msecLastSyncDuration()).count()) / 20));
+    } else {
+        qDebug() << "Setting initial sync start delay of" << delay.count();
     }
 
     // Delays beyond one minute seem too big, particularly since there
--- a/src/gui/folderstatusmodel.cpp
+++ b/src/gui/folderstatusmodel.cpp
@@ -1071,14 +1071,10 @@ void FolderStatusModel::slotFolderSyncStateChange(Folder *f)
             if (other != f && other->isSyncRunning())
                 pos += 1;
         }
-        QString message;
-        if (pos <= 0) {
-            message = tr("Waiting...");
-        } else {
-            message = tr("Waiting for %n other folder(s)...", "", pos);
+        if (pos > 0) {
+            pi._overallSyncString = tr("Waiting for %n other folder(s)...", "", pos);
         }
         pi = SubFolderInfo::Progress();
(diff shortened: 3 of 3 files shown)


src/gui/folderman.cpp
```diff
@@ -817,7 +817,7 @@ void FolderMan::startScheduledSyncSoon()
         return;
     }
 
-    seconds delay;
+    seconds delay { 2s }; // Startup, if _lastSyncFolder is still empty.
```
TheOneRing, line 820: Nitpic: you can save that comment by moving it into the else case :D

TheOneRing: Anyhow it should default to 1s why do we need the 2s?

src/gui/folderstatusmodel.cpp
```diff
@@ -1071,14 +1071,10 @@ void FolderStatusModel::slotFolderSyncStateChange(Folder *f)
             if (other != f && other->isSyncRunning())
                 pos += 1;
         }
-        QString message;
-        if (pos <= 0) {
-            message = tr("Waiting...");
-        } else {
-            message = tr("Waiting for %n other folder(s)...", "", pos);
+        if (pos > 0) {
+            pi._overallSyncString = tr("Waiting for %n other folder(s)...", "", pos);
```
TheOneRing, line 1075: Unrelated.
We will also display a progress bar in this condition which might look even stranger without text.
image

changelog/unreleased/9253
```diff
@@ -0,0 +1,8 @@
+Bugfix: Make first sync start fast again.
```
TheOneRing, line 1: While change logs are awesome, bugfixes for regressions that where never released might bloat it a bit.
The bug was only introduced recently on master and never part of any released version.

TheOneRing (changes requested):

src/gui/folderman.cpp
TheOneRing: `delay = qBound(1s, delay - sinceLastSync, 60s);`

src/gui/folderstatusmodel.cpp
dragotin: Hmm? Have you tried with that patch? That looks different in my branch. No bar..

src/gui/folderman.cpp
dragotin: I hate uninitialized variables as they are always a source of errors, like here. 
Default can be 1s, I dont care.

changelog/unreleased/9253
dragotin: True. Sorry. Was late already.

TheOneRing: You are right the visibility of the bar is not depending on well progress but on `overallString.isEmpty()`

src/gui/folderman.cpp
```diff
@@ -817,7 +817,7 @@ void FolderMan::startScheduledSyncSoon()
         return;
     }
 
-    seconds delay;
```
TheOneRing, line 820: I guess we should initialize both with `{}` as the code remains wrong.

src/gui/folderstatusmodel.cpp
TheOneRing: You are right the visibility of the bar is not depending on well progress but on overallString.isEmpty()

individual-it:
@dragotin I've pushed changes to the test code and resolved conflicts. Hope tests will pass same as in #9333

individual-it:
I received an patched version of squish see https://github.com/owncloud-ci/squish/pull/29
as soon as that is merged we should be able to revert 0932216

individual-it:
fixes #9240

dragotin:
Is there still anything open with this one @TheOneRing ?

src/gui/folderman.cpp
TheOneRing: Please fix

TheOneRing:
> Is there still anything open with this one @TheOneRing ?

https://github.com/owncloud/client/pull/9302#discussion_r772250605

TheOneRing (approved):

Issue #9316, opened 2021-12-22
[QA] Folder deleted on server not removed locally after LibreOffice is closed


gabi18:
Seen with client 2.10 RC1

Steps done 
- open file 'OpenDocument Text (neu)' inside folder '(Neuer Ordner)' in LibreOffice, do not change
- remove the folder on the server
- exit LibreOffice
Expected: folder should be deleted locally

Result:
The folder isn't deleted, the client still shows 'i' and folder is currently in use in 'Not synced' tab

![currently in use](https://user-images.githubusercontent.com/75983861/147087335-538bad57-17dc-4d08-a042-448fe9fbaa44.PNG)

Lockhunter shows:
![explorer-lock](https://user-images.githubusercontent.com/75983861/147087755-238bdec0-a796-4e23-a723-270b711535a2.PNG)

Related to: https://github.com/owncloud/client/issues/9315 ?

Logfiles:
[20211222_1208_owncloud.log.0.gz](https://github.com/owncloud/client/files/7762303/20211222_1208_owncloud.log.0.gz)
[20211222_1208_owncloud.log.1.gz](https://github.com/owncloud/client/files/7762304/20211222_1208_owncloud.log.1.gz)
[20211222_1209_owncloud.log.0.no-delete.gz](https://github.com/owncloud/client/files/7762321/20211222_1209_owncloud.log.0.no-delete.gz)

'Force sync' doesn't help while Explorer is open, only after closing the Explorer and 'Force sync' the delete of the folder and file is done.

[20211222_1244_owncloud.log.0.gz](https://github.com/owncloud/client/files/7762419/20211222_1244_owncloud.log.0.gz)
[20211222_1244_owncloud.log.1.gz](https://github.com/owncloud/client/files/7762422/20211222_1244_owncloud.log.1.gz)
[20211222_1245_owncloud.log.0.gz](https://github.com/owncloud/client/files/7762423/20211222_1245_owncloud.log.0.gz)

gabi18:
Confirmed fixed with ownCloud-2.10.0-rc3.6417.x64.msi

Issue #9315, opened 2021-12-22
[QA] Locked file doesn't sync automatically after releasing lock in client 2.10.0 RC1


saw-jan:
After edit and save:
![placeholder01](https://user-images.githubusercontent.com/52366632/147079343-91aad2e5-1366-48b2-a7cd-11d075976e48.jpg)
After waiting few minutes:
![placeholder02](https://user-images.githubusercontent.com/52366632/147079373-2c950446-ddb9-4a2b-884d-344940904d1d.jpg)

### Expected behaviour
expected file changes to sync automatically

### Actual behaviour
client still says `file is currently in use` even though the lock is released

### Steps to reproduce
1. sync a file from server
2. open the file in LibreOffice (locks the file) and save some changes
3. close the LibreOffice (releases lock)
4. wait for client to sync the file (file will not be synced automattically)

Note: force sync will work

### Server configuration
ownCloud version: `10.9.0 (stable)`

### Client configuration
Client version: `2.10.0 RC1`
Operating system: win10

saw-jan:
Logs:
[20211222_0303_owncloud.log.0.gz](https://github.com/owncloud/client/files/7762213/20211222_0303_owncloud.log.0.gz)
[20211222_0303_owncloud.log.1.gz](https://github.com/owncloud/client/files/7762214/20211222_0303_owncloud.log.1.gz)
[20211222_0303_owncloud.log.2.gz](https://github.com/owncloud/client/files/7762215/20211222_0303_owncloud.log.2.gz)
[20211222_0303_owncloud.log.3.gz](https://github.com/owncloud/client/files/7762216/20211222_0303_owncloud.log.3.gz)
[20211222_0303_owncloud.log.4.gz](https://github.com/owncloud/client/files/7762218/20211222_0303_owncloud.log.4.gz)

saw-jan:
@TheOneRing Fix has been merged, so can we close this now?

saw-jan:
Closing this for now as the fix has been done

PR #9318, opened 2021-12-22
Fix unlock of files

TheOneRing:
Fixes: #9316, #9315


erikjv (approved):

dragotin (approved):
Strange, why did it work before in some cases/all cases for me?

TheOneRing:
> Strange, why did it work before in some cases/all cases for me?

The function taking the category as argument was only introduced last week

Issue #9303, opened 2021-12-20
Fix GUI tests related to virtual file support  


Talank:
After https://github.com/owncloud/client/pull/9291, the tests related to VFS are failing. Normally, the all tests in `tst_vfs` are failing. Also, the following 2 step implementations need to be changed.

```
@When("the user enables virtual file support")
def step(context):
    enableVFSSupport("Enable virtual file support (experimental)...")
```

AND 

```
@Given("the user has enabled virtual file support")
def step(context):
    enableVFSSupport("Enable virtual file support (experimental)...")
```

Talank:
Enable/Disable VFS must be there. It was removed by accident and has been reverted by https://github.com/owncloud/client/pull/9317. So, we should revert test for enable/disable feature. However, we do not need to remove the VRT for enable VFS.

TODO
- [ ] Include test for enable/disable VFS using the dot menu

individual-it:
@Talank #9320 got merged, close this?

Talank:
Fixed by https://github.com/owncloud/client/pull/9320.

PR #9307, opened 2021-12-20
[Tests-Only] Fixed failing GUI tests for VFS

Talank:
I have used screenshot comparison to check if VFS is enabled or not. The tolerance of 16 has been set so that it ignores difference in background color of the object to be compare.

### Related Issue
Fixes https://github.com/owncloud/client/issues/9303

### How this fix works?
Since the button to disable VFS is not available, we had to do the visual test. Also, the content test was not reliable so, I used the screenshot testing. In order to focus on the availability if the element shown in the picture, I have created a positive mask around the text `Virtual file support is enabled.` Also, threshold and tolerance of 4.5 is used because in drone, the difference is less than 4.5 and failing difference is more than 4.5 (about 5.8%).
Screenshot from 2021-12-20 15-07-24

PR #9319, opened 2021-12-27
[Tests-only]Reverted removed steps of VFS tests

Talank:
The steps that were removed in https://github.com/owncloud/client/pull/9307 has been reestablished. It is because, the removal of `Enable/Disable VFS` button was unintentional.

### Related issues
- fixes https://github.com/owncloud/client/issues/9303


Talank:
Moved into another PR https://github.com/owncloud/client/pull/9320 because drone build did not start in this PR

PR #9320, opened 2021-12-27
[Tests-Only] Reverted removed steps of VFS tests

Talank:
The steps that were removed in https://github.com/owncloud/client/pull/9307 has been reestablished. It is because, the removal of `Enable/Disable VFS` button was unintentional.

### Related issues
- fixes https://github.com/owncloud/client/issues/9303

### DIFF
--- a/test/gui/tst_vfs/test.feature
+++ b/test/gui/tst_vfs/test.feature
@@ -4,15 +4,23 @@ Feature: Enable/disable virtual file support
     I want to enable virtual file support
     So that I can synchronize virtual files with local folder
 
-    
+
     Scenario: Enable VFS
         Given user "Alice" has been created on the server with default attributes and without skeleton files
         And user "Alice" has set up a client with default settings
         When the user enables virtual file support
-        Then VFS enabled baseline image should match the default screenshot
+        Then the "Disable virtual file support..." button should be available
+        And VFS enabled baseline image should match the default screenshot
 
-    
+
     Scenario: VFS is disabled by default
         Given user "Alice" has been created on the server with default attributes and without skeleton files
         And user "Alice" has set up a client with default settings
         Then VFS enabled baseline image should not match the default screenshot
+
+    Scenario: Disable VFS
+        Given user "Alice" has been created on the server with default attributes and without skeleton files
+        And user "Alice" has set up a client with default settings
+        And the user has enabled virtual file support
+        When the user disables virtual file support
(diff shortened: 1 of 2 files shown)


test/gui/tst_sharing/test.feature
```diff
@@ -324,7 +324,7 @@ Feature: Sharing
         And the public should be able to download the file "textfile.txt" with password "pass123" from the last created public link by "Alice" on the server
 
 
-    @issue-8733
+    @skip @issue-9321
```
individual-it, line 327: why skipping this test and changing the issue number?

test/gui/tst_sharing/test.feature
```diff
@@ -354,7 +354,7 @@ Feature: Sharing
         Then as user "Alice" the folder "simple-folder" should have a public link on the server
         And the public should be able to download the folder "lorem.txt" with password "pass123" from the last created public link by "Alice" on the server
 
-    @issue-8733
+    @skip @issue-9321
```
individual-it, line 357: why skipping this test and changing the issue number?

individual-it (changes requested):

test/gui/tst_sharing/test.feature
Talank: The issue 8733 was solved, and closed. Now the test is failing with a different reason. But I shall revert this change. I skipped it in this PR just to check if everything else passes. Fixing these tests will be done in separate PR(s).

test/gui/tst_vfs/test.feature
```diff
@@ -4,15 +4,23 @@ Feature: Enable/disable virtual file support
...
     So that I can synchronize virtual files with local folder
 
-    
+
     Scenario: Enable VFS
         Given user "Alice" has been created on the server with default attributes and without skeleton files
         And user "Alice" has set up a client with default settings
         When the user enables virtual file support
-        Then VFS enabled baseline image should match the default screenshot
+        Then the "Disable virtual file support..." button should be available
```
saw-jan, line 12: It would be nice to include where the button is expected to be available.
```suggestion
        Then the "Disable virtual file support..." button should be available in (<some menu>or <location>)
```

individual-it: I think it is OK the way it is, the button might move somewhere else one day but the test-description should then still stay the same, just the implementation would change

individual-it (approved):

PR #9323, opened 2021-12-29
[tests-only] Added more gui tests for sharing with user (Part 02)

saw-jan:
### Description

### Related Issues
- Part of https://github.com/owncloud/client/issues/8971#issuecomment-909884257

### DIFF
--- a/test/gui/shared/scripts/pageObjects/PublicLinkDialog.py
+++ b/test/gui/shared/scripts/pageObjects/PublicLinkDialog.py
@@ -1,7 +1,7 @@
 import names
 import squish
 import test
-import datetime
+from datetime import datetime
 
 
 class PublicLinkDialog:
@@ -72,6 +72,9 @@ class PublicLinkDialog:
         "visible": 1,
     }
 
+    # to store current default public link expiry date
+    defaultExpiryDate = ''
+
     def openPublicLinkDialog(self):
         squish.mouseClick(
             squish.waitForObject(self.PUBLIC_LINKS_TAB),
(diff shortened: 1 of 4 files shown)


test/gui/shared/scripts/pageObjects/PublicLinkDialog.py
```diff
@@ -246,3 +266,13 @@ def verifyResource(self, resource):
             str(squish.waitForObjectExists(self.ITEM_TO_SHARE).text),
             resource,
         )
+
+    def verifyExpirationDate(self, expectedDate):
+        expectedDate = datetime.strptime(expectedDate, '%Y-%m-%d')
+        expYear = expectedDate.year - 2000
```
Talank, line 272: Oh! I didn't realize that the year is composed of only 2 digits. May be add some comment here?

test/gui/shared/steps/steps.py
```diff
@@ -336,17 +332,20 @@ def step(context):
...
-    waitForFolderToBeSynced(context, folderName)
+    resource = join(context.userData['currentUserSyncPath'], folderName)
+    waitForFolderToBeSynced(context, resource)
 
 
 @Given('the user has waited for file "|any|" to be synced')
 def step(context, fileName):
-    waitForFileToBeSynced(context, fileName)
+    resource = join(context.userData['currentUserSyncPath'], fileName)
+    waitForFileToBeSynced(context, resource)
```
Talank, line 348: The logic here is same as `when the user waits for file "|any|" to be synced`. So can you reuse the logic? (By putting the logic in a different function and then calling the function from step definition)

test/gui/shared/scripts/pageObjects/PublicLinkDialog.py
saw-jan: I didn't get you?

shell_integration/nautilus/syncstate.py
```diff
@@ -97,9 +101,11 @@ def _connectToSocketServer(self):
             self._sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
             sock_file = os.path.join(get_runtime_dir(), appname, "socket")
             try:
-                self._sock.connect(sock_file) # fails if sock_file doesn't exist
+                self._sock.connect(sock_file)  # fails if sock_file doesn't exist
```
Talank, line 104: did `black` make this change?

test/gui/shared/scripts/pageObjects/PublicLinkDialog.py
Talank: I meant, add a comment here, describing the logic.

shell_integration/nautilus/syncstate.py
saw-jan: maybe

saw-jan: I must revert them back. They are source code files.

Talank (commented):
lgtm

test/gui/tst_sharing/test.feature
```diff
@@ -41,24 +65,37 @@ Feature: Sharing
...
+        And as "Brian" folder "SampleFolder,With,$pecial?Characters" should exist on the server
+        And as "Brian" file "$ample1?.txt" should exist on the server
 
 
-    Scenario: try to share a file with a user to whom the file has already been shared
+    Scenario: try to share a file/folder with a user to whom the file has already been shared
         Given user "Brian" has been created on the server with default attributes and without skeleton files
+        And user "Alice" has created folder "SharedFolder" on the server
         And user "Alice" has uploaded file with content "ownCloud test text file" to "/textfile.txt" on the server
+        And user "Alice" has shared file "SharedFolder" on the server with user "Brian" with "all" permissions
```
individual-it, line 76:
```suggestion
        And user "Alice" has shared folder "SharedFolder" on the server with user "Brian" with "all" permissions
```

test/gui/shared/steps/steps.py
```diff
@@ -380,6 +387,11 @@ def step(context, username, foldername):
     createFolder(context, foldername, username)
 
 
+@When('user "|any|" tries to create a folder "|any|" inside the sync folder')
```
individual-it, line 390: I don't think we need an extra step here, that does the same thing as the one above?

actually the user does not only try to create a folder, the user creates a folder (on the local disk) but it might or might not get synced to the server.

Same for the steps with files

individual-it (changes requested):
I think we don't need the "tries" steps

also please split the PR into smaller ones

saw-jan:
Splitted into https://github.com/owncloud/client/pull/9388 and https://github.com/owncloud/client/pull/9389

PR #9329, opened 2022-01-04
Fix compilation with Qt 5.12.8

fmoc:
Should still compile with Qt 5.15, where some of these aliases have been deprecated and replaced by the new spelling.


TheOneRing:
Looks ok, please revert once merged to master.

fmoc:
My bad. In fact we intend to require Qt 5.15 on `master`. (As this PR shows, it wouldn't be hard to retain Qt 5.12.8 at this point, but later on we plan to move to using Qt 5.15 features.)

Although we specify Qt 5.15 as lower version requirement, CMake does not appear to make sure a compatible version is available. I'm investigating this at the moment.

Issue #9325, opened 2022-01-04
File creation GUI tests fails saying file changed during sync


Talank:
As of the error in Scenario `Syncing a file to the server`, the file is not synced where the not synced tab says that the file changed while uploading.
![Screenshot from 2021-12-28 17-01-52](https://user-images.githubusercontent.com/27364813/147561194-9ce9eddf-7cf2-411b-81b5-7efc8da6d285.png)

_Originally posted by @Talank in https://github.com/owncloud/client/issues/9281#issuecomment-1002033359_

Talank:
Note that 2 of the failing tests tagged with #9281 are because of this issue. Those tests are as follows: 
```
6 | {
7 | "Error Details": "Exception: Step execution through test middleware failed. Error: AssertionError [ERR_ASSERTION]: File should have content \"test content\", but found \"<?xml version=\"1.0\" encoding=\"utf-8\"?>\n<d:error xmlns:d=\"DAV:\" xmlns:s=\"http://sabredav.org/ns\">\n  <s:exception>Sabre\\DAV\\Exception\\NotFound</s:exception>\n  <s:message>File with name lorem-for-upload.txt could not be located</s:message>\n</d:error>\n\"\n    at fileShouldHaveContent (/drone/src/middleware/src/stepDefinitions/webdavContext.js:36:17)\n    at runMicrotasks (<anonymous>)\n    at processTicksAndRejections (internal/process/task_queues.js:95:5)\n    at async /drone/src/middleware/src/app.js:66:7",
8 | "Feature": "Syncing files",
9 | "Feature File": "tst_syncing",
10 | "Scenario": "Syncing a file to the server",
11 | "Test Step": "Then as \"Alice\" the file \"lorem-for-upload.txt\" on the server should have the content \"test content\""
12 | },
13 | {
[...]

PR #9331, opened 2022-01-05
[Tests-Only]Fixed file creation and syncing tests

Talank:
### Description
If a file fails, we mark it as ignored for 5s. If it fails again within the 25s, 525, if again 55*25...
If you want to test the "changed during upload" scenario, you should wait for the error to be resolved, or force sync.
If you don't want to test it, don't modify the src file to often and not after the sync was started.

Besides the mentioned ignore list there are other random timeouts in the client which are hard to predict.

REF: https://github.com/owncloud/client/issues/9325#issuecomment-1016179775

Hence, it looks like waiting for 5 sec in the beginning is a good solution.

### Related Issue
Fixes https://github.com/owncloud/client/issues/9325

### DIFF
--- a/test/gui/tst_syncing/test.feature
+++ b/test/gui/tst_syncing/test.feature
@@ -7,7 +7,7 @@ Feature: Syncing files
     Background:
         Given user "Alice" has been created on the server with default attributes and without skeleton files
 
-    @smokeTest @skip @issue-9281
+    @smokeTest @issue-9281
     Scenario: Syncing a file to the server
         Given user "Alice" has set up a client with default settings
         When user "Alice" creates a file "lorem-for-upload.txt" with the following content inside the sync folder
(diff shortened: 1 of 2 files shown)


test/gui/tst_syncing/test.feature
```diff
@@ -7,7 +7,7 @@ Feature: Syncing files
     Background:
         Given user "Alice" has been created on the server with default attributes and without skeleton files
 
-    @smokeTest @skip @issue-9281
```
saw-jan, line 10: Should we remove the issue tag or keep it for future references?

Talank: I am not sure. What do you suggest?

saw-jan: Let's not remove the issue tag as we do in other repos

Talank: ok, done

saw-jan (commented):
LGTM

individual-it (approved):

Issue #9327, opened 2022-01-04
make CI nightly run tests on 2.10 branch


individual-it:
- [x] nightly builds should run on the 2.10 branch
- [x] template of tests need to have a note saying to update the branch

Talank:
I have added a new cron job in drone with the name `nightly-2-10`

Talank:
The nightly build is running https://drone.owncloud.com/owncloud/client/10186

individual-it:
the reporting of the result to Rocket.Chat does not work yet

Talank:
The rocket chat always shows the result for build https://drone.owncloud.services/client/build/6443 instead of showing the result of the recent nightly builds. The issue is with all the cron jobs.

PR #9338, opened 2022-01-07
[Tests-Only]Fixed drone notification channel

Talank:
We did not receive notification in desktop-internal channel because the notification channel was set to `desktop-ci` which does not exists AFAIK. 

part of https://github.com/owncloud/client/issues/9327

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -48,6 +48,7 @@ def main(ctx):
                 "check-starlark",
                 "changelog",
                 "clang-debug-ninja",
+                "GUI-tests",
             ],
         ),
     ]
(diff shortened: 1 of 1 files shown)


TheOneRing:
As all changes from 2.10 are merged to master why do we need a pr for master?

Talank:
Because the drone.star file looked different to me. There is a separate pipeline defined for just the cron jobs (namely `cron_pipelines`) in 2.10, which seems to be missing in the master. Note that this pipeline was created 7 months ago. However, If everythings from 2.10 will be eventually merged into master then I guess we do not need to merge this PR.

PR #9339, opened 2022-01-07
Improve oauth2 authorization

mortbauer:
- fixed a small bug regarding the redirect_uri for the token_endpoint
- add option oauthPrompt with default value true to disable it
  the prompt parameter for the authorization_endpoint optionally, ory
  hydra doesn't support that option and the oauth2 flow would fail
- add option oauthBasicAuth with default value true to disable sending
   of the Authorization header for the token_endpoint, ory hydra only
   allows either client_secret_basic or client_secret_post not a mix of
   both


TheOneRing:
Hi thx for you contribution.
https://github.com/owncloud/client/pull/9339/commits/c05733e56eb3f35fccd3a9214d8164d055e941b1
Was fixed last week, however it wasn't merged to master until today.
The rest is more complex and needs discussion on our side at the moment we only support custom idps with branded builds, but I'm pretty sure we are going to support it with unbranded clients.

PR #9344, opened 2022-01-11
use https to clone submodules

individual-it:
submodules step failed with 
```
fatal: clone of 'git://github.com/dschmidt/libcrashreporter-qt.git' into submodule path '/drone/src/src/3rdparty/libcrashreporter-qt' failed
Failed to clone 'src/3rdparty/libcrashreporter-qt'. Retry scheduled
```
GitHub changed stuff for security reasons and without username we have to use HTTPS see https://github.blog/2021-09-01-improving-git-protocol-security-github/


TheOneRing:
thx

PR #9345, opened 2022-01-11
use https to clone submodules (master branch)

individual-it:
forward port of #9344


TheOneRing:
I will just merge 2.9 with 2.10

PR #9349, opened 2022-01-11
Remove jhash32

TheOneRing:


TheOneRing:
I'm wrong check_c_jhash_trials is pointless as far as I can tell.
c_jhash64 however is still used in 
https://github.com/owncloud/client/blob/bfa988779de6fc08edf905f0d3780dd3db893af2/src/common/syncjournaldb.cpp#L358

dragotin (approved):
Why should that go into 2.10? Isn't that asking for trouble?

Maybe the same algorithm is implemented in some standard lib? Or cloud be replaced by something Qt brings?

fmoc (approved):

PR #9350, opened 2022-01-11
[tests-only] wait for deleted account to disappear

### LINKED ISSUE #9348: test "remove account connection" is unstable
that test sometimes fails e.g. https://drone.owncloud.com/owncloud/client/10307/5/14

looks like `waitForObjectExists` does not return an object

individual-it:
after deleting an account wait till it disappeared from the UI, this hopefully will fix #9348
I think we are sometimes checking the UI in the middle of rearranging and that way missing elements

To make the code nicer I moved more actions into the pageobject


saw-jan (approved):
LGTM

PR #9353, opened 2022-01-11
Make sure to use the same Qt version in our CMake scripts

fmoc:
Both our own CMake code as well as third-party dependencies we use uses `find_package(Qt5 ...)` and variants to build against Qt. This PR introduces the `OC_QT_VERSION` variable in our project's root scope, and makes sure to use it in every `find_package` call.

We cannot control third-party dependency lookups this way, unfortunately. So far, this has not led to problems in our build environments, though.

The other commit implements a small change to the platform utilities which is unrelated but improves compilation speed slightly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,7 +23,10 @@ include(KDECompilerSettings NO_POLICY_SCOPE)
 
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
-find_package(Qt5 REQUIRED COMPONENTS Core)
+# global Qt version, used in all find_package calls
+set(OC_QT_VERSION 5.15)
+
+find_package(Qt5 ${OC_QT_VERSION} REQUIRED COMPONENTS Core)
 
 if(UNIT_TESTING)
     message(DEPRECATION "Setting UNIT_TESTING is deprecated please use BUILD_TESTING")
--- a/admin/osx/CMakeLists.txt
+++ b/admin/osx/CMakeLists.txt
@@ -11,7 +11,6 @@ else()
   set(MAC_INSTALLER_DO_CUSTOM_BACKGROUND "false")
 endif()
 
-find_package(Qt5 5.6 COMPONENTS Core REQUIRED)
 configure_file(create_mac.sh.cmake ${CMAKE_CURRENT_BINARY_DIR}/create_mac.sh @ONLY)
 configure_file(macosx.pkgproj.cmake ${CMAKE_CURRENT_BINARY_DIR}/macosx.pkgproj @ONLY)
 configure_file(pre_install.sh.cmake ${CMAKE_CURRENT_BINARY_DIR}/pre_install.sh @ONLY)
--- a/shell_integration/dolphin/CMakeLists.txt
+++ b/shell_integration/dolphin/CMakeLists.txt
@@ -2,10 +2,9 @@ project(dolphin-owncloud)
 
 cmake_minimum_required(VERSION 2.8.12)
 include(FeatureSummary)
-set(QT_MIN_VERSION "5.3.0")
 set(KF5_MIN_VERSION "5.16.0")
 
-find_package(Qt5 ${QT_MIN_VERSION} CONFIG REQUIRED COMPONENTS Core Network)
+find_package(Qt5 ${OC_QT_VERSION} CONFIG REQUIRED COMPONENTS Core Network)
 
 find_package(ECM 1.2.0 REQUIRED CONFIG)
 set(CMAKE_MODULE_PATH ${ECM_MODULE_PATH} ${ECM_KDE_MODULE_DIR} "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
--- a/src/common/utility.h
+++ b/src/common/utility.h
@@ -121,11 +121,11 @@ OCSYNC_EXPORT Q_DECLARE_LOGGING_CATEGORY(lcUtility)
 #endif
 
     // convenience OS detection methods
-    inline bool isWindows();
-    inline bool isMac();
-    inline bool isUnix();
-    inline bool isLinux(); // use with care
-    inline bool isBSD(); // use with care, does not match OS X
+    constexpr inline bool isWindows();
+    constexpr inline bool isMac();
+    constexpr inline bool isUnix();
+    constexpr inline bool isLinux(); // use with care
+    constexpr inline bool isBSD(); // use with care, does not match OS X
 
     OCSYNC_EXPORT QString platformName();
     // crash helper for --debug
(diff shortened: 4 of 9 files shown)


CMakeLists.txt
```diff
@@ -23,7 +23,10 @@ include(KDECompilerSettings NO_POLICY_SCOPE)
 
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
-find_package(Qt5 REQUIRED COMPONENTS Core)
+# global Qt version, used in all find_package calls
+set(OC_QT_VERSION 5.15 CACHE STRING "Qt version to build against" FORCE)
```
TheOneRing, line 27: There is no need to cache it.

CMakeLists.txt
```diff
@@ -23,7 +23,10 @@ include(KDECompilerSettings NO_POLICY_SCOPE)
...
+
+find_package(Qt5 ${OC_QT_VERSION} REQUIRED COMPONENTS Core)
```
TheOneRing, line 29: This will break the build of the dolphin plugin which requires the system Qt.

admin/osx/CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ else()
   set(MAC_INSTALLER_DO_CUSTOM_BACKGROUND "false")
 endif()
 
-find_package(Qt5 5.6 COMPONENTS Core REQUIRED)
+find_package(Qt5 ${OC_QT_VERSION} COMPONENTS Core REQUIRED)
```
TheOneRing, line 14: Do we use Qt at all in here?

shell_integration/dolphin/CMakeLists.txt
```diff
@@ -2,10 +2,9 @@ project(dolphin-owncloud)
 
 cmake_minimum_required(VERSION 2.8.12)
 include(FeatureSummary)
-set(QT_MIN_VERSION "5.3.0")
 set(KF5_MIN_VERSION "5.16.0")
 
-find_package(Qt5 ${QT_MIN_VERSION} CONFIG REQUIRED COMPONENTS Core Network)
```
TheOneRing, line 8: We need the system Qt here which could be anything.

@dschmidt why do we build the dolphin stuff as part of the main project?

src/common/utility.h
```diff
@@ -121,11 +121,11 @@ OCSYNC_EXPORT Q_DECLARE_LOGGING_CATEGORY(lcUtility)
 #endif
 
     // convenience OS detection methods
-    inline bool isWindows();
-    inline bool isMac();
-    inline bool isUnix();
-    inline bool isLinux(); // use with care
-    inline bool isBSD(); // use with care, does not match OS X
+    constexpr inline bool isWindows();
```
TheOneRing, line 124: doesn't constexpr imply inline?

src/gui/CMakeLists.txt
```diff
@@ -152,7 +152,7 @@ target_sources(owncloudCore PRIVATE
 )
 
 IF( APPLE )
-    find_package(Qt5 COMPONENTS MacExtras)
```
TheOneRing, line 155: maybe we can move all find package stuff to the top of the file 😵‍💫

TheOneRing (changes requested):

CMakeLists.txt
fmoc: Indeed. I'll revert that change.

admin/osx/CMakeLists.txt
fmoc: Doesn't look like it. Do you want me to remove it?

dragotin (changes requested):
What does that mean? Do you want to pin the Qt version? 

Please do not, that is only good for our internal build infra. Apply a patch there if you need it.
All other usecases of externals, building on whatever Qt, will likely break otherwise.
Bad idea IMHO.

src/common/utility.h
fmoc: Not necessarily. The function will be inlined as well, but `inline` also implies some definition rules according to the standard. I don't think it'll make a difference here. Nevertheless, as long as there is no reason to remove `inline`, I'd keep those specifiers.

src/gui/CMakeLists.txt
fmoc: We can do that, but then we have to replicate the platform-specific logic there as well. IMO that could be considered duplicate code.

CMakeLists.txt
fmoc: It shouldn't hurt, though. I'm not sure why I added the caching instructions. I can remove it.

admin/osx/CMakeLists.txt
TheOneRing: yes

fmoc:
It does not pin the version, it properly enforces the lower compatibility limit (as has been done before, but not quite as thoroughly as this PR does). Pinning would mean the use of `EXACT` in `find_package`, which you won't find for a good reason. Any newer version of Qt can be used just fine. The build system should, however make sure that

a) the same version is included/used every where (to avoid mixing, e.g., our own Qt 5.15 build with some Qt 5.12 from the system)
b) make sure a compatible version is available (that is, Qt 5.15 or newer)

CMakeLists.txt
fmoc: Actually, I think you want to make sure everyone uses the same Qt version. We can do a separate search with another version in the Dolphin plugin project, which can be built independently anyway, but is included by default for convenience reasons. But when building as part of the whole project, we should make sure to use the same version everywhere.

shell_integration/dolphin/CMakeLists.txt
fmoc: We don't need to look for some system Qt when building this sub-project with the remaining project IMO. This file declares a project anyway, so this can be built separately. What should be preserved is the old minimum Qt version for use in standalone builds.

dschmidt (commented):
I'm not sure the OC_QT_VERSION solves an actual problem. 
I assume that if we find 5.15 on top level we won't find other modules from other versions in the subdirectories ... 

For the shell_integration this change harmful as OC_QT_VERSION won't be defined (although it was before) when built outside of the main project.

CMakeLists.txt
TheOneRing: I think the correct solution would be to move the plugin out of the main repo.
However it would massively increase the management overhead.

dschmidt: Define "correct". Yes, it's theoretically independent ... but what do we gain from splitting it out except from overhead?

fmoc:
> I assume that if we find 5.15 on top level we won't find other modules from other versions in the subdirectories ...

CMake managed to mix libraries from craft, a PPA and my system once. Took me a while to figure out why. When looking for additional modules, it is required to also make sure the right versions are found. This PR attempts to solve this issue.

CMakeLists.txt
fmoc: Is there any advantage of building this plugin along with the rest of the binaries? It will most likely not work anyway unless built against the system Qt. We need to link the rest against the craft Qt, though. And I'm pretty sure we cannot ensure CMake uses the system Qt for this subproject only.

PR #9354, opened 2022-01-12
Improve code style 😜

TheOneRing:


TheOneRing:
Looks like g++ doesn't share my sense of humor

fmoc:
It does, starting in version 10. But we cannot bump the requirement at this point.

dragotin:
Grmbl....

PR #9358, opened 2022-01-12
Fix propagation of build flags to unrelated targets

fmoc:
This PR makes sure flags set on our own targets won't be propagated to third-party targets which do not need them. In the present case, a new third-party dependency generated a `config.h` file, but instead of that one, ownCloud's was included, causing the build to fail.

I replaced all occurrences of `add_libraries`, `include_directories` etc. (`link_libraries` was not used anywhere).


shell_integration/dolphin/CMakeLists.txt
```diff
@@ -43,6 +40,8 @@ add_library(${OWNCLOUDDOLPHINHELPER} SHARED ownclouddolphinpluginhelper.cpp)
 target_link_libraries(${OWNCLOUDDOLPHINHELPER} Qt5::Network Qt5::Gui)
 generate_export_header(${OWNCLOUDDOLPHINHELPER} BASE_NAME ownclouddolphinpluginhelper)
 install(TARGETS ${OWNCLOUDDOLPHINHELPER} LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR})
+target_include_directories(${OWNCLOUDDOLPHINHELPER} PUBLIC ${CMAKE_CURRENT_BINARY_DIR} ${PROJECT_BINARY_DIR} ${PROJECT_BINARY_DIR}/src)
```
TheOneRing, line 43: https://cmake.org/cmake/help/latest/variable/CMAKE_INCLUDE_CURRENT_DIR.html
Will be needed in the dolphin CMakeLists.txt too as it has its own project() scope.
But should be covered by https://invent.kde.org/frameworks/extra-cmake-modules/-/blob/master/kde-modules/KDECMakeSettings.cmake#L194

shell_integration/dolphin/CMakeLists.txt
```diff
@@ -43,6 +40,8 @@ add_library(${OWNCLOUDDOLPHINHELPER} SHARED ownclouddolphinpluginhelper.cpp)
...
+target_compile_definitions(${OWNCLOUDDOLPHINHELPER} PUBLIC -DQT_USE_FAST_CONCATENATION -DQT_USE_FAST_OPERATOR_PLUS)
```
TheOneRing, line 44: Private

src/csync/CMakeLists.txt
```diff
@@ -42,11 +33,49 @@ configure_file(csync_version.h.in ${CMAKE_CURRENT_BINARY_DIR}/csync_version.h)
...
+    PUBLIC -DQT_NO_URL_CAST_FROM_STRING
+    PUBLIC -DQT_NO_CAST_FROM_BYTEARRAY
+    PUBLIC -DQT_USE_QSTRINGBUILDER
+    PUBLIC -DQT_MESSAGELOGCONTEXT #enable function name and line number in debug output
+    PUBLIC -DQT_DEPRECATED_WARNINGS
+    PUBLIC -DQT_NO_FOREACH
+)
+
+if(MEM_NULL_TESTS)
+    target_compile_definitions(csync PUBLIC -DCSYNC_MEM_NULL_TESTS)
```
TheOneRing, line 56: Can be removed, isn't referenced anywhere.

src/csync/CMakeLists.txt
```diff
@@ -42,11 +33,49 @@ configure_file(csync_version.h.in ${CMAKE_CURRENT_BINARY_DIR}/csync_version.h)
...
+target_include_directories(csync
+    PUBLIC ${CMAKE_CURRENT_BINARY_DIR} ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_SOURCE_DIR}/std
+    # include root source dir in order to make src/common available to the targets
+    PUBLIC ${PROJECT_SOURCE_DIR}/src
+    # make sure auto-generated headers can be included
+    PUBLIC ${PROJECT_BINARY_DIR}
 )
 
+target_compile_definitions(csync
+    PUBLIC -DQT_NO_CAST_TO_ASCII
```
TheOneRing, line 45: Private

src/csync/CMakeLists.txt
```diff
@@ -42,11 +33,49 @@ configure_file(csync_version.h.in ${CMAKE_CURRENT_BINARY_DIR}/csync_version.h)
 
 add_library(csync SHARED ${common_SOURCES} ${csync_SRCS})
 
-target_include_directories(
-  csync
-  PUBLIC ${CMAKE_CURRENT_BINARY_DIR} ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_SOURCE_DIR}/std
+target_include_directories(csync
+    PUBLIC ${CMAKE_CURRENT_BINARY_DIR} ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_SOURCE_DIR}/std
```
TheOneRing, line 37: Should be covered by https://invent.kde.org/frameworks/extra-cmake-modules/-/blob/master/kde-modules/KDECMakeSettings.cmake#L194
https://cmake.org/cmake/help/latest/variable/CMAKE_INCLUDE_CURRENT_DIR.html

src/csync/CMakeLists.txt
```diff
@@ -42,11 +33,49 @@ configure_file(csync_version.h.in ${CMAKE_CURRENT_BINARY_DIR}/csync_version.h)
...
+    target_compile_definitions(csync PUBLIC -DCSYNC_MEM_NULL_TESTS)
+endif(MEM_NULL_TESTS)
+
+if(TOKEN_AUTH_ONLY)
+    message("Compiling with token authentication")
+    target_compile_definitions(csync PUBLIC -DTOKEN_AUTH_ONLY=1)
+endif()
+
+if(NO_MSG_HANDLER)
+    target_compile_definitions(csync PUBLIC -DNO_MSG_HANDLER=1)
```
TheOneRing, line 65: Private in libsync where it is used.

src/csync/CMakeLists.txt
```diff
@@ -42,11 +33,49 @@ configure_file(csync_version.h.in ${CMAKE_CURRENT_BINARY_DIR}/csync_version.h)
...
+    PUBLIC -DQT_NO_FOREACH
+)
+
+if(MEM_NULL_TESTS)
+    target_compile_definitions(csync PUBLIC -DCSYNC_MEM_NULL_TESTS)
+endif(MEM_NULL_TESTS)
+
+if(TOKEN_AUTH_ONLY)
+    message("Compiling with token authentication")
+    target_compile_definitions(csync PUBLIC -DTOKEN_AUTH_ONLY=1)
```
TheOneRing, line 61: I hate this flag as the feature does not exist any more, this is an imaginary no ui config which probably won't even build.

src/libsync/CMakeLists.txt
```diff
@@ -82,6 +77,12 @@ target_link_libraries(libsync PUBLIC
     csync
     Qt5::Core Qt5::Network
 )
+target_compile_definitions(libsync
+    PUBLIC -DQT_NO_CAST_TO_ASCII
```
TheOneRing, line 81: private

test/csync/CMakeLists.txt
```diff
@@ -1,15 +1,16 @@
...
-  ${CMOCKA_INCLUDE_DIR}
-)
 
-include_directories(${CHECK_INCLUDE_DIRS})
 # create test library
 add_library(torture STATIC torture.c cmdline.c)
 target_link_libraries(torture ${CMOCKA_LIBRARIES})
 set_target_properties(torture PROPERTIES C_STANDARD 99 AUTOMOC OFF)
+target_include_directories(torture
+    PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}
```
TheOneRing, line 8: https://cmake.org/cmake/help/latest/variable/CMAKE_INCLUDE_CURRENT_DIR.html

And the src dir should already be included by linking to csync.

src/csync/CMakeLists.txt
TheOneRing: One PUBLIC should be enough

TheOneRing (changes requested):

shell_integration/dolphin/CMakeLists.txt
fmoc: Let me check whether it works without this dir.

I think in Qt projects, `CMAKE_INCLUDE_CURRENT_DIR` should be set to `On` anyway.

src/csync/CMakeLists.txt
fmoc: Indeed, it's a code style thing. I'd rather move the other `PUBLIC` public flags to new lines as well.

src/csync/CMakeLists.txt
fmoc: Which of those? They used to be used in all binaries.

src/csync/CMakeLists.txt
fmoc: Should we remove it? We could use a `message(FATAL_ERROR ...`).
(In fact, not using `message(STATUS ...)` is a problem already.)

src/libsync/CMakeLists.txt
fmoc: All of them? Only this one?

shell_integration/dolphin/CMakeLists.txt
fmoc: I could only make them private when repeating this code for the other two targets.

shell_integration/dolphin/CMakeLists.txt
TheOneRing: It is through KDE settings

shell_integration/dolphin/CMakeLists.txt
TheOneRing: Please do so, setting compile flags is not duplicating code and such build flags should never be propagated to all dependent targets.

src/csync/CMakeLists.txt
TheOneRing: Some yes, but some where only set on a directory level.
QT_NO_CAST_TO_ASCII will cause all QString("foo") calls to fail which are still used in the gui code.

src/libsync/CMakeLists.txt
fmoc: We have to remove the first one altogether, the rest should be private, in order to make the build work.

src/csync/CMakeLists.txt
TheOneRing: Would require removing the ifdef from all the code, keep it for now pls.
I was just mentioning my resent...

src/libsync/CMakeLists.txt
TheOneRing: All

TheOneRing (approved):

individual-it:
@TheOneRing @fmoc 
are these changes intended?
**before this PR**
grafik

**after this PR**
grafik

TheOneRing:
Its not, looks like the application name isn't set correctly.
@fmoc pls fix

PR #9366, opened 2022-01-14
Webfinger

TheOneRing:
A branch based on v2.10.0 can be found here: https://github.com/owncloud/client/compare/v2.10.0...work/v2.10.0-webfinger?expand=1

A branch based on v2.10.0 https://github.com/owncloud/client/tree/work/v2.10.1-webfinger .
The change modifies the behavior of the wizard and therefor must not be merged.
The next major release will bring the feature as a branding option, the feature however depends on the rewrite of the Wizard.


TheOneRing:
This branch is not to be merged

PR #9375, opened 2022-01-18
Ensure to raise the ssl dialog

TheOneRing:


fmoc (approved):

TheOneRing:
Lets mark that one as untestable, it was more a theoretic discovery that it could happen, however I don't think I experienced the issue.

PR #9376, opened 2022-01-19
Built-in AppImage self-updater

fmoc:
Adds a libappimageupdate-based updater to the code.

Note that we still have to build libappimageupdate and its dependencies with craft for the build to work. The feature is off by default for Linux builds.


src/gui/updater/appimageupdater.cpp
```diff
@@ -0,0 +1,186 @@
...
+        setDownloadState(DownloadFailed);
+        return;
+    }
+
+    auto dialog = new Ui::AppImageUpdateAvailableDialog(currentVersion, newVersion, ocApp()->gui()->settingsDialog());
+
+    connect(dialog, &Ui::AppImageUpdateAvailableDialog::skipUpdateButtonClicked, this, [this]() {
+        qCInfo(lcUpdater) << "Update skipped by user";
+
+        // TODO: remember this setting
```
TheOneRing, line 159: yes do so

src/gui/updater/appimageupdater.cpp
```diff
@@ -0,0 +1,186 @@
...
+                qCInfo(lcUpdater) << "AppImage update failed";
+                setDownloadState(DownloadFailed);
+            }
+        });
+
+        setDownloadState(Downloading);
+        appImageUpdaterShim->startUpdateInBackground();
+    });
+
+    dialog->show();
```
TheOneRing, line 178: Please also call ownCloudGui::raiseDialog(dialog);

src/gui/updater/updater.cpp
```diff
@@ -72,7 +76,15 @@ QUrlQuery Updater::getQueryParams()
     Theme *theme = Theme::instance();
     QString platform = QStringLiteral("stranger");
     if (Utility::isLinux()) {
-        platform = QStringLiteral("linux");
+#ifdef WITH_APPIMAGEUPDATER
+        if (qEnvironmentVariableIsSet("APPIMAGE")) {
```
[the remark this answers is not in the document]
fmoc: Same issue in #9403. I think it might be easier to put the const string somewhere...

src/gui/updater/appimageupdater.cpp
fmoc: The wizard raises the settings dialog rather than raising the wizard. I'd implement the same semantics in here, too. Since the dialog is modal, it'll be shown above the settings dialog then.

TheOneRing: Well it might do so.
We also added some hacks in there to ensure the os will bring the wizard to the front when stuff is minimized and other applications are currently active.
However a linux implementation is currently missing, which means you will just get a desktop notification that tells you that ownCloud request your attention or something like that.

src/gui/updater/CMakeLists.txt
```diff
@@ -6,3 +6,53 @@ if(APPLE AND NOT BUILD_LIBRARIES_ONLY)
         set(OWNCLOUD_BUNDLED_RESOURCES ${OWNCLOUD_BUNDLED_RESOURCES} PARENT_SCOPE)
     endif()
 endif()
+
+target_sources(owncloudCore PRIVATE
+    ${CMAKE_CURRENT_SOURCE_DIR}/ocupdater.cpp
```
TheOneRing, line 11: We bumped the cmake deps, so you can now use a relative path.

src/gui/updater/CMakeLists.txt
```diff
@@ -6,3 +6,53 @@ if(APPLE AND NOT BUILD_LIBRARIES_ONLY)
...
+    target_link_libraries(owncloudCore PRIVATE ${SPARKLE_LIBRARY})
+endif()
+
+if(WITH_APPIMAGEUPDATER)
+    # needed for set_source_files_properties(... TARGET_DIRECTORY ...)
+    cmake_minimum_required(VERSION 3.18)
+
+    message(STATUS "Including built-in libappimageupdate based updater")
+
+    set(appimageupdater_sources
```
TheOneRing, line 35: Directly use target_source

src/gui/updater/CMakeLists.txt
```diff
@@ -6,3 +6,53 @@ if(APPLE AND NOT BUILD_LIBRARIES_ONLY)
...
+    # we have to enable exceptions for this one compilation unit
+    set_source_files_properties(${appimageupdater_sources}
+        TARGET_DIRECTORY owncloudCore
+        PROPERTIES COMPILE_OPTIONS "-fexceptions"
+    )
+
+    target_compile_definitions(owncloudCore PRIVATE WITH_APPIMAGEUPDATER)
+    target_link_libraries(owncloudCore PRIVATE libappimageupdate-qt)
+
+    find_package(Threads REQUIRED)
```
TheOneRing, line 56: https://github.com/owncloud/client/blob/master/src/gui/CMakeLists.txt#L203

src/gui/updater/appimageupdater.cpp
```diff
@@ -0,0 +1,206 @@
...
+#include <QMessageBox>
+#include <QSettings>
+#include <QTimer>
+#include <appimage/update.h>
+
+#include "appimageupdater.h"
+#include "common/version.h"
+#include "theme.h"
+#include "updater_private.h"
+// SettingsDialog is apparently forward-declared somewhere, need to include the header to make the compilation work
```
TheOneRing, line 24: that comment is a bit weird, especially as it comes before the wrong include.

src/gui/updater/appimageupdater.cpp
```diff
@@ -0,0 +1,206 @@
...
+            return false;
+        }
+    }
+
+    void startUpdateInBackground() noexcept
+    {
+        // monitor progress and log status messages
+        auto *timer = new QTimer(this);
+
+        timer->setInterval(100);
```
TheOneRing, line 96: 100ms

src/gui/updater/appimageupdater.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+        connect(timer, &QTimer::timeout, this, [=]() {
+            _logStatusMessages();
+
+            if (!_updater.isDone()) {
+                emit finished(!_updater.hasError());
+            }
+
+            timer->stop();
```
TheOneRing, line 105: We always stop after 100ms?
I was expecting some polling here?

src/gui/updater/appimageupdater.cpp
```diff
@@ -0,0 +1,206 @@
...
+    const auto &currentVersion = Version::versionWithBuildNumber();
+    const auto newVersion = QVersionNumber::fromString(info.version());
+
+    if (info.version().isEmpty() || currentVersion >= newVersion) {
+        qCInfo(lcUpdater) << "Client is on latest version!";
+        setDownloadState(UpToDate);
+        return;
+    }
+
+    QSettings settings(ConfigFile::configFile(), QSettings::IniFormat);
```
TheOneRing, line 144: If we don't have a factory for that qsettings object, please add one.
Needing to pass the format looks weird.

src/gui/updater/updater.cpp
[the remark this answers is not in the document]
TheOneRing: Still an env var here?

src/gui/updater/updater_private.h
```diff
@@ -0,0 +1,23 @@
...
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+ * for more details.
+ */
+
+#pragma once
+
+namespace OCC {
+static const char updateAvailableC[] = "Updater/updateAvailable";
```
TheOneRing, line 18: Please use a `Q.*String.*` here instead of converting 30times in the code

TheOneRing (changes requested):

src/gui/updater/CMakeLists.txt
fmoc: Only enabled when the crashreporter is active, too, which might not be the case. Do you want to suggest to merge all the checks and include the dependency in a central place? Linking the dependency into the binaries twice shouldn't hurt either, though.

src/gui/updater/appimageupdater.cpp
fmoc: `clang-format` apparently reordered the includes here and did not keep the comment with the source. I'll just dump the comment.

src/gui/updater/appimageupdater.cpp
fmoc: Indeed, this makes no sense. The `if` above is broken, too. I'll fix it.

src/gui/updater/appimageupdater.cpp
fmoc: I'd rather do that in a second PR, and replace all occurrences, wdyt?

src/gui/updater/updater.cpp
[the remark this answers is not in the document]
fmoc: I must have missed one occurrence.

src/gui/updater/CMakeLists.txt
fmoc: Not possible, since I need to set compile flags for those files.

ownclouders:
Results for <strong>GUI-tests</strong> https://drone.owncloud.com/owncloud/client/10973/5/1
GUI Logs: (https://cache.owncloud.com/public/owncloud/client/10973/guiReportUpload/index.html)
Server Logs: (https://cache.owncloud.com/public/owncloud/client/10973/guiReportUpload/serverlog.log)

src/libsync/configfile.cpp
```diff
@@ -371,6 +371,16 @@ QString ConfigFile::configFile()
     return configPath() + Theme::instance()->configFileName();
 }
 
+QSettings ConfigFile::makeQSettings()
+{
+    return { configFile(), QSettings::IniFormat };
+}
+
+QSettings *ConfigFile::makeQSettingsPtr()
```
TheOneRing, line 379: Why would we need a ptr here, and yes I know it sucks but can you split the review, changes like this can easily be reviewed on their own and it makes reviews MUCH easier.
But yes in the end you will have to rebase this one again.

src/gui/updater/CMakeLists.txt
TheOneRing: I guess we can unconditionally depend on threads.

src/gui/updater/appimageupdater.cpp
```diff
@@ -0,0 +1,202 @@
...
+    // nothing to do, update will be performed while app is running, if anything
+    return false;
+}
+
+void AppImageUpdater::versionInfoArrived(const UpdateInfo &info)
+{
+    const auto &currentVersion = Version::versionWithBuildNumber();
+    const auto newVersion = QVersionNumber::fromString(info.version());
+
+    if (info.version().isEmpty() || currentVersion >= newVersion) {
```
TheOneRing, line 138: I would expect that both evaluate to the same result (no change needed)

src/gui/updater/appimageupdater.cpp
```diff
@@ -0,0 +1,202 @@
...
+        qCInfo(lcUpdater) << "Client is on latest version!";
+        setDownloadState(UpToDate);
+        return;
+    }
+
+    auto settings = ConfigFile::makeQSettings();
+
+    const auto seenVersionString = settings.value(seenVersionC).toString();
+    if (!seenVersionString.isEmpty()) {
+        if (QVersionNumber::fromString(seenVersionString) >= QVersionNumber::fromString(info.version())) {
```
TheOneRing, line 148: Use newVersion

src/gui/updater/appimageupdater.cpp
```diff
@@ -0,0 +1,202 @@
...
+        return;
+    }
+
+    auto dialog = new Ui::AppImageUpdateAvailableDialog(currentVersion, newVersion, ocApp()->gui()->settingsDialog());
+
+    connect(dialog, &Ui::AppImageUpdateAvailableDialog::skipUpdateButtonClicked, this, [newVersion]() {
+        qCInfo(lcUpdater) << "Update" << newVersion << "skipped by user";
+
+        auto settings = ConfigFile::makeQSettings();
+        settings.setValue(seenVersionC, newVersion.toString());
```
TheOneRing, line 174: This is all very similar to the nsis code, can it be refactored?

ownclouders:
Results for <strong>GUI-tests</strong> https://drone.owncloud.com/owncloud/client/11009/5/1
GUI Logs: (https://cache.owncloud.com/public/owncloud/client/11009/guiReportUpload/index.html)
Server Logs: (https://cache.owncloud.com/public/owncloud/client/11009/guiReportUpload/serverlog.log)

src/gui/updater/appimageupdater.cpp
fmoc: I tried to stay close to the other updater code, which also performs this check. There, the message "No version information is available at the moment" is logged.

src/gui/updater/appimageupdater.cpp
fmoc: The only similarities I see are fetching and storing the seen version as well as checking whether the client is up to date or there is no version info available. I moved the former into two static utility functions within `OCUpdater`, but I think moving half of the up-to-date check into some helper doesn't solve any actual problem.

ownclouders:
Results for <strong>GUI-tests</strong> https://drone.owncloud.com/owncloud/client/11016/5/1
GUI Logs: (https://cache.owncloud.com/public/owncloud/client/11016/guiReportUpload/index.html)
Server Logs: (https://cache.owncloud.com/public/owncloud/client/11016/guiReportUpload/serverlog.log)

TheOneRing (approved):

Issue #9382, opened 2022-01-24
File modified during upload should not be handled by the local ignore list


TheOneRing:
If an upload of a file fails 3 times in a row because it was modified during sync, we add it to an ignore list.

This can easily happen for files managed by an IDE.

```
Issues         ,File                                                            ,Folder                        ,Size  ,Account             ,Time                ,Status              ,
Local file changed during sync.,untitled1/main.cpp                                              ,testpilotcloud2               ,      ,test@XXX,2022-01-24T09:13:09.848,Info                ,
Local file changed during sync.,build-untitled1-KDE-Debug/Testing/Temporary/LastTest.log        ,testpilotcloud2               ,      ,test@XXX,2022-01-24T09:13:09.837,Info                ,
File/Folder is ignored because it's hidden.,build-untitled1-KDE-Debug/.ninja_log                            ,testpilotcloud2               ,      ,test@XXX,2022-01-24T09:13:09.829,Excluded            ,
File/Folder is ignored because it's hidden.,build-untitled1-KDE-Debug/.ninja_deps                           ,testpilotcloud2               ,      ,test@XXX,2022-01-24T09:13:09.821,Excluded            ,
File/Folder is ignored because it's hidden.,build-untitled1-KDE-Debug/.cmake                                ,testpilotcloud2               ,      ,test@XXX,2022-01-24T09:13:09.814,Excluded            ,
Local file changed during sync.,build-untitled1-KDE-Debug/CMakeFiles/rules.ninja                ,testpilotcloud2               ,      ,test@XXX,2022-01-24T09:13:09.807,Info                ,
Local file changed during sync.,build-untitled1-KDE-Debug/CMakeFiles/cmake.check_cache          ,testpilotcloud2               ,      ,test@XXX,2022-01-24T09:13:09.800,Info                ,
Local file changed during sync.,build-untitled1-KDE-Debug/CMakeFiles/clean_additional.cmake     ,testpilotcloud2               ,      ,test@XXX,2022-01-24T09:13:09.794,Info                ,
Local file changed during sync.,build-untitled1-KDE-Debug/CMakeFiles/TargetDirectories.txt      ,testpilotcloud2               ,      ,test@XXX,2022-01-24T09:13:09.788,Info                ,
Local file changed during sync.,build-untitled1-KDE-Debug/CMakeFiles/CMakeOutput.log            ,testpilotcloud2               ,      ,test@XXX,2022-01-24T09:13:09.778,Info                ,
[...]

PR #9386, opened 2022-01-25
Introduce new error category that is not handled by the ignore list

TheOneRing:
Fixes: #9382

The filter menu will now be sorted by the localised string.
image

### DIFF
--- a/src/gui/issueswidget.cpp
+++ b/src/gui/issueswidget.cpp
@@ -302,14 +302,15 @@ std::function<void(void)> IssuesWidget::addStatusFilter(QMenu *menu)
     const auto initialFilter = _statusSortModel->filter();
 
     { // Add all errors under 1 action:
-        const std::array<SyncFileItem::Status, 5> ErrorStatusItems = {
+        const std::vector<SyncFileItem::Status> ErrorStatusItems = {
             SyncFileItem::Status::FatalError,
             SyncFileItem::Status::NormalError,
             SyncFileItem::Status::SoftError,
-            SyncFileItem::Status::DetailError
+            SyncFileItem::Status::DetailError,
         };
 
-        auto action = menu->addAction(SyncFileItem::statusEnumDisplayName(SyncFileItem::NormalError), [this, ErrorStatusItems](bool checked) {
+
+        auto action = menu->addAction(SyncFileItem::statusEnumDisplayName(SyncFileItem::NormalError), this, [this, ErrorStatusItems](bool checked) {
             auto currentFilter = _statusSortModel->filter();
             for (const auto &item : ErrorStatusItems) {
                 currentFilter[item] = checked;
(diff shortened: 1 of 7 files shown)


src/gui/issueswidget.cpp
```diff
@@ -320,23 +321,33 @@ std::function<void(void)> IssuesWidget::addStatusFilter(QMenu *menu)
...
+    const std::vector<SyncFileItem::Status> OtherStatusItems = {
         SyncFileItem::Status::Conflict,
         SyncFileItem::Status::FileIgnored,
         SyncFileItem::Status::Restoration,
         SyncFileItem::Status::BlacklistedError,
-        SyncFileItem::Status::Excluded
+        SyncFileItem::Status::Excluded,
+        SyncFileItem::Status::Message
     };
+    std::vector<std::pair<QString, SyncFileItem::Status>> otherStatusItems;
```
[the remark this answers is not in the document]
TheOneRing: LIke
```
 // list of OtherStatusItems with the localised name
    std::vector<std::pair<QString, SyncFileItem::Status>> otherStatusItems;
    otherStatusItems.reserve(OtherStatusItems.size());
    for (const auto &item : OtherStatusItems) {
        otherStatusItems.emplace_back(SyncFileItem::statusEnumDisplayName(item), item);
    }

File: src/libsync/localdiscoverytracker.cpp [hunk 2]
```diff
@@ -64,23 +64,48 @@ void LocalDiscoveryTracker::slotItemCompleted(const SyncFileItemPtr &item)
     //
     // For failures, we want to add the file to the list so the next sync
     // will be able to retry it.
```
TheOneRing, line 66: As this is right out of hell, please read carefully

src/libsync/localdiscoverytracker.cpp
```diff
@@ -64,23 +64,48 @@ void LocalDiscoveryTracker::slotItemCompleted(const SyncFileItemPtr &item)
...
-                      || item->_instruction == CSYNC_INSTRUCTION_UPDATE_METADATA))) {
-        if (_previousLocalDiscoveryPaths.erase(item->_file))
+
+    switch (item->_status) {
+    case SyncFileItem::NoStatus:
+        // we can't use the flags operator with CSYNC_INSTRUCTION_NONE
+        if (item->_instruction != CSYNC_INSTRUCTION_NONE && item->_instruction != CSYNC_INSTRUCTION_UPDATE_METADATA) {
+            break;
+        }
+        Q_FALLTHROUGH();
```
dragotin, line 74: This is clever, and _checkers_ of course know that macro. But not all people are Qt masters, and a comment what that macro does and how it changes the code flow would be great. As this changes the flow *significantly*.

src/libsync/propagateupload.cpp
```diff
@@ -225,26 +225,19 @@ void PropagateUploadFileCommon::slotStartUpload(const QByteArray &transmissionCh
...
-
     // But skip the file if the mtime is too close to 'now'!
     // That usually indicates a file that is still being changed
     // or not yet fully copied to the destination.
-    if (fileIsStillChanging(*_item)) {
+    _item->_modtime = FileSystem::getModTime(fullFilePath);
+    if (prevModtime != _item->_modtime || fileIsStillChanging(*_item)) {
         propagator()->_anotherSyncNeeded = true;
-        done(SyncFileItem::SoftError, tr("Local file changed during sync."));
+        done(SyncFileItem::Message, tr("Local file changed during syncing. It will be resumed."));
```
dragotin, line 240: I am not sure if "during syncing" is an accurate english term here. Maybe something like `The local file is still busy. It will be resumed.`?

dragotin (approved):
Two little comments...

src/libsync/localdiscoverytracker.cpp
TheOneRing: If I start do document the Qt functions I'll be getting no where.

src/libsync/propagateupload.cpp
TheOneRing: I merged that error with the one a few lines above they where pretty identical.
In this case I don't think we need to describe it accurate but yes during sync is more correct grammar.

src/libsync/localdiscoverytracker.cpp
dragotin: Hm, you miss my point here.

src/libsync/propagateupload.cpp
```diff
@@ -225,26 +225,19 @@ void PropagateUploadFileCommon::slotStartUpload(const QByteArray &transmissionCh
...
-
     // But skip the file if the mtime is too close to 'now'!
     // That usually indicates a file that is still being changed
     // or not yet fully copied to the destination.
-    if (fileIsStillChanging(*_item)) {
+    _item->_modtime = FileSystem::getModTime(fullFilePath);
+    if (prevModtime != _item->_modtime || fileIsStillChanging(*_item)) {
         propagator()->_anotherSyncNeeded = true;
-        done(SyncFileItem::SoftError, tr("Local file changed during sync."));
+        done(SyncFileItem::Message, tr("Local file changed during sync. It will be resumed."));
```
erikjv, line 240: resumed->retried?

Or maybe: "The next sync will retry to upload the file."

erikjv (approved):

src/libsync/propagateupload.cpp
TheOneRing: Delay until next sync?

fmoc (approved):

src/libsync/localdiscoverytracker.cpp
TheOneRing: It won't help with the readability but on master we can directly use:
https://en.cppreference.com/w/cpp/language/attributes/fallthrough

PR #9388, opened 2022-01-26
[tests-only] Added more GUI tests for sharing with user (part 02)

saw-jan:
### Description

Splitted from https://github.com/owncloud/client/pull/9323

### DIFF
--- a/test/gui/shared/scripts/bdd_hooks.py
+++ b/test/gui/shared/scripts/bdd_hooks.py
@@ -37,7 +37,7 @@ def hook(context):
     DEFAULT_CONFIG = {
         'localBackendUrl': 'https://localhost:9200/',
         'secureLocalBackendUrl': 'https://localhost:9200/',
-        'clientSyncTimeout': 60,
+        'clientSyncTimeout': 10,
         'middlewareUrl': 'http://localhost:3000/',
         'clientConfigFile': '-',
         'clientRootSyncPath': '/tmp/client-bdd/',
--- a/test/gui/shared/steps/steps.py
+++ b/test/gui/shared/steps/steps.py
@@ -157,20 +157,16 @@ def isFileSynced(fileName):
     return isItemSynced('FILE', fileName)
 
 
-def waitForFileToBeSynced(context, fileName):
+def waitForFileToBeSynced(context, filePath):
     waitFor(
-        lambda: isFileSynced(
-            sanitizePath(context.userData['currentUserSyncPath'] + fileName)
-        ),
+        lambda: isFileSynced(sanitizePath(filePath)),
         context.userData['clientSyncTimeout'] * 1000,
     )
 
 
-def waitForFolderToBeSynced(context, folderName):
+def waitForFolderToBeSynced(context, folderPath):
     waitFor(
-        lambda: isFolderSynced(
-            sanitizePath(context.userData['currentUserSyncPath'] + folderName)
-        ),
+        lambda: isFolderSynced(sanitizePath(folderPath)),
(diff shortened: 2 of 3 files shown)


test/gui/shared/scripts/bdd_hooks.py
```diff
@@ -37,7 +37,7 @@ def hook(context):
     DEFAULT_CONFIG = {
         'localBackendUrl': 'https://localhost:9200/',
         'secureLocalBackendUrl': 'https://localhost:9200/',
-        'clientSyncTimeout': 60,
+        'clientSyncTimeout': 10,
```
Talank, line 40: This PR seems to be dependent upon changes made in some other PRs (part 1 and 3). Does the order of merge should be Part1 -> part2 -> part3 ?

saw-jan: the actual order should not affect each other. one needs to be rebased. So whichever gets merged first, it should not create affect other PR

Talank (commented):
lgmt

test/gui/tst_sharing/test.feature
```diff
@@ -98,44 +143,112 @@ Feature: Sharing
...
+        And as "Alice" the file "textfile.txt" on the server should have the content "overwrite ownCloud test text file"
+        And as "Brian" the file "FOLDER/simple.txt" on the server should have the content "overwrite some content"
+        And as "Alice" the file "FOLDER/simple.txt" on the server should have the content "overwrite some content"
+        When the user adds another account with
+            | server   | %local_server% |
+            | user     | Alice          |
+            | password | 1234           |
+        And the user removes permissions "edit" for user "Brian Murphy" of resource "textfile.txt" using the client-UI
+        And the user closes the sharing dialog
+        And the user removes permissions "edit" for user "Brian Murphy" of resource "FOLDER" using the client-UI
```
individual-it, line 245: is this test about changing the sharing from the UI or about the overwriting?
Maybe we could have those steps done by the API?

test/gui/shared/steps/steps.py
```diff
@@ -920,9 +919,28 @@ def step(context, resource, group):
...
+
+
+def tryToOverwriteFile(context, resource, content):
+    waitForFileToBeSynced(context, resource)
+    waitForFolderToBeSynced(context, '/')
+
+    try:
+        overwriteFile(resource, content)
+    except:
+        pass
```
individual-it, line 935: why would that ever fail? Does the client set read-only permissions on the local file?

individual-it (changes requested):

test/gui/tst_sharing/test.feature
saw-jan: The scenario is to do `overwrite` -> `change permissions` -> `try to overwrite` using the client UI.
Any suggestions?

test/gui/shared/steps/steps.py
saw-jan: I found out that when a folder is shared with `read` permission, files inside it have `-r--r--r--` permissions but the folder itself has `drwxrwxr-x` permissions in the files system.
And when a file is shared with `read` permission, the file has `-rw-rw-r--` permissions

individual-it: same happens when you do it manually, or only in the test?

saw-jan: Yeah, it is the same but only if we share-edit-permissions first from the server and then add the user to the client.

individual-it (approved):

PR #9390, opened 2022-01-26
Fix translated icon names in desktop file

fmoc:
Follow-up for #9150, which fixed the generic icon but not the translated ones. Those translated `Icon` entries were completely redundant, so simply removing it fixes the issue and removes some bloat.

The second commit fixes an (accidentally?) translated placeholder for the `el` locale.


TheOneRing:
Please target 2.10

dschmidt:
Is this file exclusively manually edited? Then this seems to be fine

dschmidt (approved):

TheOneRing (changes requested):
Please target 2.10

TheOneRing:
> Is this file exclusively manually edited? Then this seems to be fine

We used to have automatic translations from transifex for that file,  until they stopped the support without notice.

TheOneRing (approved):

PR #9391, opened 2022-01-26
Add changelog entry for win-vfs

erikjv:


TheOneRing:
We can't change the past, please create the file in the unreleased folder.

TheOneRing (changes requested):
See above

TheOneRing (approved):

PR #9403, opened 2022-01-31
Fix AppImage autostart

fmoc:
Ideas welcome where to put that `APPIMAGE` const string.

### DIFF
--- a/src/common/utility_unix.cpp
+++ b/src/common/utility_unix.cpp
@@ -65,12 +65,27 @@ void setLaunchOnStartup_private(const QString &appName, const QString &guiName, 
             qCWarning(lcUtility) << "Could not write auto start entry" << desktopFileLocation;
             return;
         }
+
+        auto autostartApplicationPath = []() {
+            // $APPIMAGE will be set to the AppImage's path by the AppImage runtime
+            // if it is set, we can assume to be run from within an AppImage
+            // in that case, the desktop file should point to the AppImage rather than the
+            // main binary, which will be in a temporary mount point
+            static const char APPIMAGE[] = "APPIMAGE";
+
+            if (qEnvironmentVariableIsSet(APPIMAGE)) {
+                return qEnvironmentVariable(APPIMAGE);
+            }
+
+            return QCoreApplication::applicationFilePath();
+        }();
+
         QTextStream ts(&iniFile);
         ts.setCodec("UTF-8");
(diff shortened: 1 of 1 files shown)


src/common/utility_unix.cpp
```diff
@@ -65,12 +65,27 @@ void setLaunchOnStartup_private(const QString &appName, const QString &guiName,
             qCWarning(lcUtility) << "Could not write auto start entry" << desktopFileLocation;
             return;
         }
+
+        auto autostartApplicationPath = []() {
+            // $APPIMAGE will be set to the AppImage's path by the AppImage runtime
+            // if it is set, we can assume to be run from within an AppImage
+            // in that case, the desktop file should point to the AppImage rather than the
+            // main binary, which will be in a temporary mount point
+            static const char APPIMAGE[] = "APPIMAGE";
```
TheOneRing, line 74: As that env var is now used and defined in multiple places, maybe add it to Utility?

fmoc: I was thinking about something like this:

```cpp
std::optional<QString> appImageLocation();
```

Qt doesn't seem to have an equivalent of `std::optional`, `QVariant` could be used. I guess we can also define an empty string as "not running in AppImage".

TheOneRing:
```c++
QString appImageLocation(){
static QString loc = qGetEnv();
retun loc;
}
bool isAppimage() {
   return !appImageLocation().isEmpty();
}
```

I don't think std::optional would add much here.

TheOneRing:
I'd say the rebase failed somehow

src/common/utility_unix.cpp
```diff
@@ -98,6 +111,16 @@ void Utility::setLaunchOnStartup(const QString &appName, const QString &guiName,
     }
 }
 
+#ifdef Q_OS_LINUX
+QString Utility::appImageLocation() {
+    return qEnvironmentVariable("APPIMAGE");
```
TheOneRing, line 116: Not sure about the performance nowadays but in the past we cached the env var in a local static.

fmoc: Seems a bit overkill. The method is called just once.

fmoc:
The build is broken because of an unrelated craft issue. @TheOneRing mind to have a look? Could be a temporary hiccup.

PR #9410, opened 2022-02-01
Ensure the retry counter is reset

TheOneRing:


src/libsync/creds/httpcredentials.cpp
```diff
@@ -239,44 +239,54 @@ bool HttpCredentials::stillValid(QNetworkReply *reply)
...
     // don't touch _ready or the account state will start a new authentication
     // _ready = false;
 
     OAuth *oauth = new OAuth(_account, this);
-    connect(oauth, &OAuth::refreshError, this, [oauth, this](QNetworkReply::NetworkError error, const QString &errorString) {
+    connect(oauth, &OAuth::refreshError, this, [oauth, tokenRefreshRetriesCount, this](QNetworkReply::NetworkError error, const QString &errorString) {
         oauth->deleteLater();
+        int nextTry = tokenRefreshRetriesCount + 1;
         _isRenewingOAuthToken = false;
+        std::chrono::seconds timeout = {};
```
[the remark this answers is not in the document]
TheOneRing: yes but it would looks weird with just a break, I only initialise it once to prevent possible surprises.

src/libsync/creds/httpcredentials.cpp
```diff
@@ -239,44 +239,54 @@ bool HttpCredentials::stillValid(QNetworkReply *reply)
...
+            timeout = 30s;
+        }
+        if (nextTry >= TokenRefreshMaxRetries) {
+            qCWarning(lcHttpCredentials) << "Too many failed refreshs" << nextTry << "-> log out";
+            forgetSensitiveData();
+            Q_EMIT _account->invalidCredentials();
+            return;
         }
+        QTimer::singleShot(timeout, this, [nextTry, this] {
+            _isRenewingOAuthToken = false;
```
TheOneRing, line 281: I think that part was essential too as else different parts of the client could retrigger it independent of the timer.

src/libsync/creds/httpcredentials.h
```diff
@@ -73,7 +73,10 @@ class OWNCLOUDSYNC_EXPORT HttpCredentials : public AbstractCredentials
     /* If we still have a valid refresh token, try to refresh it assynchronously and emit fetched()
      * otherwise return false
      */
-    bool refreshAccessToken();
+    bool refreshAccessToken()
```
fmoc, line 76: Should likely be moved into implementation.

src/libsync/creds/httpcredentials.cpp
fmoc: Sounds reasonable.

fmoc (approved):

Issue #9399, opened 2022-01-28
[tests-only][GUI] Use `owncloud-test-middleware` from docker image


saw-jan:
Currently, we use `owncloud-test-middleware` from github repo and pin commit-id.
`owncloud-test-middleware` has docker its own docker images so like in `web` it is better to use here too.

middleware image: https://hub.docker.com/r/owncloud/owncloud-test-middleware/tags

NOTE: Implementing this might cause a lot of test failures

PR #9412, opened 2022-02-02
[Tests-only]Used middleware docker image

Talank:
Use middleware docker image instead of cloning the middleware from git.

### Related Issues
- Closes https://github.com/owncloud/client/issues/9399
- Opens https://github.com/owncloud/client/issues/9430
- Opens https://github.com/owncloud/client/issues/9439

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -5,6 +5,8 @@
 # with a specific compiler.
 #
 
+OC_TESTING_MIDDLEWARE = "owncloud/owncloud-test-middleware:1.3.1"
+
 dir = {
     "base": "/drone",
 }
@@ -232,11 +234,15 @@ def gui_tests(ctx, trigger = {}, depends_on = [], filterTags = [], version = "da
                      },
                  ] +
                  showGuiTestResult(),
-        "services": testMiddleware() +
+        "services": testMiddlewareService() +
                     owncloudService() +
                     databaseService(),
         "trigger": trigger,
         "depends_on": depends_on,
+        "volumes": [{
+            "name": "uploads",
+            "temp": {},
+        }],
     }
 
 def build_client(ctx, c_compiler, cxx_compiler, build_type, generator, build_command, build_dir):
@@ -484,23 +490,22 @@ def owncloudService():
         ],
     }]
 
-def testMiddleware():
+def testMiddlewareService():
+    environment = {
+        "BACKEND_HOST": "http://owncloud",
+        "NODE_TLS_REJECT_UNAUTHORIZED": "0",
+        "MIDDLEWARE_HOST": "testmiddleware",
+        "REMOTE_UPLOAD_DIR": "/uploads",
+    }
+
     return [{
         "name": "testmiddleware",
-        "image": "owncloudci/nodejs:14",
-        "pull": "always",
-        "environment": {
-            "MIDDLEWARE_HOST": "testmiddleware",
-            "BACKEND_HOST": "http://owncloud",
-        },
-        "commands": [
-            ". ./.drone.env",
-            "git clone https://github.com/owncloud/owncloud-test-middleware.git /drone/src/middleware",
-            "cd /drone/src/middleware",
-            "git checkout $MIDDLEWARE_COMMITID",
-            "yarn install",
-            "yarn start",
-        ],
+        "image": OC_TESTING_MIDDLEWARE,
+        "environment": environment,
+        "volumes": [{
+            "name": "uploads",
+            "path": "/uploads",
+        }],
     }]
 
 def owncloudLog():
--- a/test/gui/shared/steps/steps.py
+++ b/test/gui/shared/steps/steps.py
@@ -517,16 +517,6 @@ def step(context):
     )
     activateItem(waitForObjectItem(names.settings_QMenu, "Force sync now"))
 
-    # It is not guranteed that the object showing warning message will appear during the force sync.
-    # If the message appears, we need to click on the "Remove all files" button. Else, we should carry on with rest of the steps in test without any action.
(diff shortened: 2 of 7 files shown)


.drone.star
```diff
@@ -485,22 +487,20 @@ def owncloudService():
...
-            "git checkout $MIDDLEWARE_COMMITID",
-            "yarn install",
-            "yarn start",
-        ],
+        "name": "middleware",
+        "image": OC_TESTING_MIDDLEWARE,
+        "environment": environment,
+        "volumes": [{
+            "name": "uploads",
+            "path": "/filesForUpload",
```
kiranparajuli589, line 502:
```suggestion
            "path": "/usr/src/app/filesForUpload",
```

kiranparajuli589 (changes requested):

.drone.star
Talank: Still getting the error `can't create '/uploads/'single'quotes.txt': No such file or directory` https://drone.owncloud.com/owncloud/client/10546/5/2

.drone.star
```diff
@@ -484,23 +490,21 @@ def owncloudService():
         ],
     }]
 
-def testMiddleware():
+def startTestMiddleware():
+    environment = {
+        "BACKEND_HOST": "http://owncloud",
+        "NODE_TLS_REJECT_UNAUTHORIZED": "0",
+        "MIDDLEWARE_HOST": "testmiddleware",
+    }
```
kiranparajuli589, line 498: try putting the remote upload dir too....check if works
```suggestion
        "REMOTE_UPLOAD_DIR": "/usr/src/app/filesForUpload",
    }
```

Talank: I think I tried that already. But lets try it again.

Talank: It didn't work https://drone.owncloud.com/owncloud/client/10590/5/2

.drone.star
```diff
@@ -484,23 +490,22 @@ def owncloudService():
...
-            "cd /drone/src/middleware",
-            "git checkout $MIDDLEWARE_COMMITID",
-            "yarn install",
-            "yarn start",
-        ],
+        "image": OC_TESTING_MIDDLEWARE,
+        "environment": environment,
+        "volumes": [{
+            "name": "uploads",
+            "path": "/usr/src/app/filesForUpload",
```
kiranparajuli589, line 507:
```suggestion
            "path": "/uploads",
```

kiranparajuli589: this should make uploads dir for you, no need to create manually

.drone.star
```diff
@@ -232,11 +234,15 @@ def gui_tests(ctx, trigger = {}, depends_on = [], filterTags = [], version = "da
                      },
                  ] +
                  showGuiTestResult(),
-        "services": testMiddleware() +
+        "services": startTestMiddleware() +
```
saw-jan, line 237: Since it is used as service
```suggestion
        "services": testMiddlewareService() +
```

saw-jan (commented):
LGTM

test/gui/shared/steps/steps.py
```diff
@@ -516,6 +516,12 @@ def step(context):
         Qt.LeftButton,
     )
     activateItem(waitForObjectItem(names.settings_QMenu, "Force sync now"))
+    try:
+        clickButton(
+            waitForObject(names.remove_All_Files_Remove_all_files_QPushButton, 2000)
```
kiranparajuli589, line 521: seems a strange try-catch to me, the exception is just passed. but the then should catch errors if something goes wrong.
also, for the usage of static waits, a comment would be reasonable. will `1000` work or is `2000` enough? these types of confusion will not arise if there is already an explanation.

Talank: The `2000` is the timeout. I put it inside try-catch because it is not guranteed that the object will appear during the force sync. And if it does not appear then there is loockup error. But if we get loockup error, then we want to move on to next step. Because it implies that the message (object) did not appear.

test/gui/shared/steps/steps.py
```diff
@@ -517,6 +517,16 @@ def step(context):
     )
     activateItem(waitForObjectItem(names.settings_QMenu, "Force sync now"))
 
+    # It is not guranteed that the object showing warning message will appear during the force sync.
+    # If the message appears, we need to click on the "Remove all files" button. Else, we should carry on with rest of the steps in test without any action.
+    # It takes upto 2000 ms for the object to appear. So, I put the timeout to be 2000 ms.
```
individual-it, line 522: no you haven't ;) I see a timeout of 1000ms

individual-it (changes requested):
do we have to put both things into one PR? To me they look pretty unrelated.
Also I don't understand why we need a force-sync. If a file is deleted the client should just sync that change

test/gui/shared/steps/steps.py
Talank: I changed it for try and forgot to revert it back 🤣 Thanks for noticing.

saw-jan: Delete tests need some adjustments alongside using the latest middleware. I suspect currently we are getting the false-positive results from delete test scenarios.

Talank: I have fixed it and squashed the commits. Can you review it again?

Talank:
The file/folder delete tests are failing after the numping of middleware. The reasons are is as follows:
- Given there is only one file/folder in the sync directory of current user, after syncing from server. When we delete the file/folder then delete activity is not synced.
- This test, however, used to pass in the previously pinned version of middleware. Because, in the previous version of middleware, the file existance in the server was not compared properly, causing the false pass of the files/folders tests.

individual-it (approved):

individual-it (approved):

PR #9413, opened 2022-02-02
Handle cases where server checksums might be zero-padded

ishank011:


TheOneRing:
Thx for the pull request.
Why does reva pad the checksum, are you aware that such a server change will break all existing clients?

ishank011:
Hi @TheOneRing, this isn't done by reva. It just passes the checksum returned by the storage backend, so it's EOS that does that, and so might other file systems.

To get around this, we currently trim the padding in reva but we thought it would be good to handle such cases in the client itself.

src/common/checksums.cpp
```diff
@@ -360,8 +360,16 @@ void ValidateChecksumHeader::slotChecksumCalculated(const QByteArray &checksumTy
         emit validationFailed(tr("The checksum header contained an unknown checksum type '%1'").arg(QString::fromLatin1(_expectedChecksumType)));
         return;
     }
-    if (checksum != _expectedChecksum) {
```
TheOneRing, line 363: Hm it would be sufficient to check:
```c++
// allow leading padding
if (_expectedChecksum.endsWith(checksum)) {
```


https://doc.qt.io/qt-5/qbytearray.html#endsWith

TheOneRing (commented):
If we apply this change it will be available with the next release in a couple of month.
Only upcoming releases would be compatible with the changed server behaviour, so I'm not 100% sure this should be handled in the client.

src/common/checksums.cpp
fmoc: I think the padding prefix character should be specified, because otherwise the probability of false positive matches will be increased as far as I can see.

I'd suggest `_expectedChecksum.split('0', QString::SkipEmptyParts).join("") == checksum)`.

fmoc (commented):
I think this is a case of "sender makes it right" rather than "client makes it right". The server should a correct checksum. Padding is just unexpected.

TheOneRing:
We discussed this request and came to the conclusion that this should rather be handled in the server.
Especially with the existence of other clients in mind, which all would require similar patches.
Checksums are supposed to be 1:1 matches.

ishank011:
@TheOneRing @fmoc thanks for the reviews. Will handle this on the server side then

PR #9414, opened 2022-02-02
Add assert to assure verb is uppercase

TheOneRing:


fmoc (approved):
Makes sense, although I think you could also just forcefully convert to upper-case.

An alternative solution is to check against an allowlist.

TheOneRing:
> Makes sense, although I think you could also just forcefully convert to upper-case.
> 
> An alternative solution is to check against an allowlist.

That would just waste cpu on code that should not exist in the first place.

PR #9415, opened 2022-02-02
Only set valid icons on top-level windows

erikjv:
Also warn when an icon is null.


erikjv:
I hope this fixes https://sentry.io/organizations/owncloud/issues/2956248154/?project=79001

TheOneRing:
The question is why does it fail to load the icon, thats not supposed to happen in the first place.
And well I would not expect Qt to crash in that case.

fmoc:
I think assertions in the getters might be a better solution. There's likely a root cause to this mess...

TheOneRing:
As discussed in chat we now expect https://github.com/owncloud/client/pull/9418 to fix this.

PR #9416, opened 2022-02-02
Fix duplicate option in clang-format config

fmoc:
The option was in there twice, causing confusion for some tools that can parse `.clang-format` files.

Also, the value was invalid, see https://clang.llvm.org/docs/ClangFormatStyleOptions.html. It must be one of `Never`, `CaseSensitive` or `CaseInsensitive`. I think `CaseSensitive` is a sensible choice.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -48,11 +48,7 @@ AlwaysBreakTemplateDeclarations: true
 # clang-format does not handle that
 AllowShortFunctionsOnASingleLine: Inline
 
-# The coding style specifies some include order categories, but also tells to
-# separate categories with an empty line. It does not specify the order within
-# the categories. Since the SortInclude feature of clang-format does not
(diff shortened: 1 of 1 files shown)


TheOneRing:
> The option was in there twice, causing confusion for some tools that can parse `.clang-format` files.
> 
> Also, the value was invalid, see https://clang.llvm.org/docs/ClangFormatStyleOptions.html. It must be one of `Never`, `CaseSensitive` or `CaseInsensitive`. I think `CaseSensitive` is a sensible choice.

Hmm see the discussion from 7 month ago:
https://github.com/owncloud/client/pull/8818#discussion_r670415732

fmoc:
The option was duplicate nevertheless. I can change it back to `true` if needed.

TheOneRing:
I'll fix craft

Issue #9405, opened 2022-01-31
"Open local folder" and "Open folder in browser" are not shown in the context menu


naokton:
### Expected behaviour
"Open local folder" and "Open folder in browser" are shown in the context menu when I right click synced sub folders in ownCloud app. This was the behavior of version 2.9.2.

### Actual behaviour
"Open local folder" and "Open folder in browser" are not shown. Instead, "Remove folder sync connection" is shown.

### Steps to reproduce
1. Open ownCloud desktop application
2. Right click synced sub folder

### Server configuration
Operating system:

Web server: 

Database:

PHP version:

ownCloud version:

Storage backend (external storage):

### Client configuration
Client version:
2.10.0

Operating system:
Windows Server 2019

OS language:
English

Installation path of client:
C:\Program Files\ownCloud

### Logs

Please use Gist (https://gist.github.com/) or a similar code paster for longer
logs.

```Template for output < 10 lines```

1. Client [logfile](https://doc.owncloud.com/desktop/latest/appendices/troubleshooting.html#log-files)

2. Web server error log:

3. Server logfile: ownCloud log (data/owncloud.log):

TheOneRing:
Could you share a screenshot of what you are talking about?
Names etc can be redacted.

naokton:
Sure. This is version 2.10.0
![image](https://user-images.githubusercontent.com/996594/151829082-95c18d1c-e763-47a4-a65f-d6b147ae29f4.png)

And this is version 2.9.2
![image](https://user-images.githubusercontent.com/996594/151829346-1dd82f79-4cd0-486d-ac6a-2b1e17c8564e.png)

TheOneRing:
@naokton thx for reporting.

@erikjv can you have a look, you recently changed the model index handling.

erikjv:
Can you add a screenshot of what you get when you right-click on the cloud-with-checkmark?

naokton:
@erikjv This is the screenshot. Both version (2.10.0 and 2.9.2) show the same items.

![image](https://user-images.githubusercontent.com/996594/152233971-d80d3fde-6b32-4a48-911e-cac7155a8f44.png)

PR #9420, opened 2022-02-04
Bugfix: add open local/remote folder options

erikjv:
Add the "open local/remote folder" context menu items for non-sync-root
items back into the accounts tab in the settings dialog.

https://github.com/owncloud/client/issues/9405


src/gui/accountsettings.cpp
```diff
@@ -243,82 +243,116 @@ void AccountSettings::slotCustomContextMenuRequested(const QPoint &pos)
...
+
+    auto classification = _model->classify(index);
+    if (classification != FolderStatusModel::RootFolder && classification != FolderStatusModel::SubFolder) {
+        return;
+    }
+
+
+    // Only allow removal if the item isn't in "ready" state.
+    if (classification == FolderStatusModel::RootFolder && !_model->data(index, FolderStatusDelegate::IsReady).toBool()) {
         QMenu *menu = new QMenu(tv);
```
TheOneRing, line 261: So only the root node will get the remove action?
(Do we paint the children if the root is != ready?)

src/gui/accountsettings.cpp
```diff
@@ -243,82 +243,116 @@ void AccountSettings::slotCustomContextMenuRequested(const QPoint &pos)
...
+        QString openLocallyLabel;
+        if (Utility::isWindows()) {
+            openLocallyLabel = tr("Show in Explorer");
+        } else if (Utility::isMac()) {
+            openLocallyLabel = tr("Show in Finder");
+        } else {
+            openLocallyLabel = tr("Show if file manager");
+        }
+
+        QAction *ac = menu->addAction(openLocallyLabel);
```
TheOneRing, line 295:
```cpp
addAction(openLocallyLabel, this, []{
});
```

src/gui/accountsettings.cpp
```diff
@@ -243,82 +243,116 @@ void AccountSettings::slotCustomContextMenuRequested(const QPoint &pos)
...
         }
+    }
+
+    // Add an action to open the folder on the server in a webbrowser:
 
-        ac = menu->addAction(tr("Open folder in browser"));
-        auto info = _model->infoForIndex(index);
-        OC_ASSERT(info);
+    if (auto info = _model->infoForIndex(index)) {
+        QAction *ac = menu->addAction(tr("Open Remote Folder in Browser"));
```
TheOneRing, line 311: see `addAction(openLocallyLabel, this, []{`

src/gui/accountsettings.cpp
erikjv: > So only the root node will get the remove action?

Yes. My thinking was that it's a destructive action, and shouldn't be confused with "delete this folder". But I can change that if you think having it everywhere is more consistent.

> (Do we paint the children if the root is != ready?)

No idea, because I'm not 100% sure what this state means. When the account is singled out it is still ready, so I don't know how to get into that state.

TheOneRing: A read only folder should be != ready

src/gui/accountsettings.cpp
TheOneRing: Open in web browser maybe?

src/gui/accountsettings.cpp
```diff
@@ -243,82 +243,116 @@ void AccountSettings::slotCustomContextMenuRequested(const QPoint &pos)
...
         QString path = info->_folder->remotePathTrailingSlash();
         path += info->_path;
```
TheOneRing, line 313: for the root folder this results in //

erikjv: No, `remotePathTrailingSlash()` only appends the slash if there isn't already a trailing slash.

src/gui/accountsettings.cpp
erikjv: Hmm, I wanted to make it clear that it's the *remote* folder that's being opened. But maybe that's too obvious. What do you think?

src/gui/accountsettings.cpp
TheOneRing: yes but info->_path will be `/`

TheOneRing: I've tried :D

src/gui/accountsettings.cpp
erikjv: No, the child items are not shown, only the sync root

src/gui/accountsettings.cpp
TheOneRing: Yes but what is a remote folder :D
and we should align it with the others.

image

TheOneRing (approved):

PR #9422, opened 2022-02-07
Extract "Update URL dialog" into separate class

fmoc:


src/gui/accountstate.cpp
```diff
@@ -47,47 +47,36 @@ Q_LOGGING_CATEGORY(lcAccountState, "gui.account.state", QtInfoMsg)
...
-    auto accept = [=] {
+
+    _updateUrlDialog = new UpdateUrlDialog(
+        tr("Url update requested for %1").arg(_account->displayName()),
+        tr("The url for %1 changed from %2 to %3, do you want to accept the changed url?").arg(_account->displayName(), _account->url().toString(), newUrl.toString()),
+        _account->url(),
+        newUrl,
+        ocApp()->activeWindow());
+
+    auto cleanup = [=]() {
```
TheOneRing, line 64: Better use:
`_updateUrlDialog->setAttribute(Qt::WA_DeleteOnClose);`

src/gui/updateurldialog.cpp
```diff
@@ -0,0 +1,60 @@
+#include "updateurldialog.h"
+
+#include <QMessageBox>
+// this include allows comparing QAbstractButton* and QPushButton*
+#include <QPushButton>
+
+namespace OCC {
+
+UpdateUrlDialog::UpdateUrlDialog(const QString &messageBoxTitle, const QString &messageBoxContent, const QUrl &oldUrl, const QUrl &newUrl, QWidget *parent)
```
TheOneRing, line 9: We do we inherit QObject and not QDialog?

fmoc: Using `QWidget` allows passing the parent to the message box, which expects such a pointer.

src/gui/accountstate.cpp
fmoc: That'd only work if the class inherited `QDialog` (or maybe even `QMessageBox`). The class's name is not ideal.

I'm not sure whether inheritance makes sense over composition here. We don't intend to extend the functionality of `QMessageBox`, but use one as-is with the existing interfaces. Therefore, composition makes more sense. Arguably, it requires some more boilerplate code...

src/gui/updateurldialog.cpp
TheOneRing: I mean why do you not inherit QDialog or QMessagebox but build that wrapper around it?

fmoc: See https://github.com/owncloud/client/pull/9422/files/fb490599e3ca47b86799f6de8ddd3231a3ba1b8a#r800696137.

src/gui/updateurldialog.cpp
```diff
@@ -31,30 +30,26 @@ void UpdateUrlDialog::show()
...
-            Q_EMIT changeAccepted();
-        }
-
-        Q_EMIT finished();
-    });
-
-    _messageBox->show();
+    auto yesButton = addButton(tr("Change url permanently to %1").arg(_newUrl.toString()), QMessageBox::AcceptRole);
+    addButton(tr("Reject"), QMessageBox::RejectRole);
+    //
```
TheOneRing, line 43: ?

src/gui/accountstate.cpp
```diff
@@ -72,8 +72,8 @@ UpdateUrlDialog *AccountState::updateUrlDialog(const QUrl &newUrl)
         cleanup();
     };
 
-    connect(_updateUrlDialog, &UpdateUrlDialog::changeAccepted, this, accept);
-    connect(_updateUrlDialog, &UpdateUrlDialog::finished, this, cleanup);
+    connect(_updateUrlDialog, &UpdateUrlDialog::accepted, this, accept);
+    connect(_updateUrlDialog, &UpdateUrlDialog::done, this, cleanup);
```
TheOneRing, line 76: Is this still needed, the dialog is now supposed to delete itself?

src/gui/owncloudsetupwizard.cpp
```diff
@@ -172,7 +172,10 @@ void OwncloudSetupWizard::slotFoundServer(const QUrl &url, const QJsonObject &in
             auto accountState = AccountState::fromNewAccount(_ocWizard->account()).take(); // we manage the lifetime for this temporary AccountState ourselves (see below)
             if (auto dialog = accountState->updateUrlDialog(url)) {
                 // Dialog is show, make sure the accountState is deleted when the dialog is done.
-                connect(dialog, &UpdateUrlDialog::finished, accountState, &QObject::deleteLater);
+                connect(dialog, &UpdateUrlDialog::finished, accountState, [accountState](int result) {
+                    (void)result;
```
TheOneRing, line 176: Q_UNUSED, it the same but more verbose in a Qt way.

src/gui/updateurldialog.h
```diff
@@ -6,27 +6,19 @@
...
+class UpdateUrlDialog : public QMessageBox
 {
     Q_OBJECT
 public:
-    explicit UpdateUrlDialog(const QString &messageBoxTitle, const QString &messageBoxContent, const QUrl &oldUrl, const QUrl &newUrl, QWidget *parent = nullptr);
+    explicit UpdateUrlDialog(const QString &title, const QString &content, const QUrl &oldUrl, const QUrl &newUrl, QWidget *parent = nullptr);
 
+    // warning: this method overwrites the superclass's start method
+    // do *not* call the function via a base class pointer!
     void show();
```
TheOneRing, line 17: Thats a hack, I assume we can move all logic to the constructor and use the base implementation of show?

src/gui/updateurldialog.cpp
```diff
@@ -0,0 +1,55 @@
+#include "updateurldialog.h"
+
+#include <QMessageBox>
+#include <QTimer>
+// this include allows comparing QAbstractButton* and QPushButton*
```
TheOneRing, line 5: Does commenting includes really help?

TheOneRing (changes requested):

src/gui/updateurldialog.cpp
fmoc: Qt is full of forward declarations. `QMessageBox` declares those classes, but you can't instantiate them. I can remove the comment.

src/gui/updateurldialog.cpp
fmoc: Debug code, will be removed.

src/gui/updateurldialog.h
fmoc: How would you signalize that "immediate return" from the constructor? You need to be able to hook up a method to `accepted` before the URLs are compared. We can't use exceptions there.

I was thinking about a named constructor with a `bool` return type.

src/gui/owncloudsetupwizard.cpp
```diff
@@ -172,7 +172,10 @@ void OwncloudSetupWizard::slotFoundServer(const QUrl &url, const QJsonObject &in
             auto accountState = AccountState::fromNewAccount(_ocWizard->account()).take(); // we manage the lifetime for this temporary AccountState ourselves (see below)
             if (auto dialog = accountState->updateUrlDialog(url)) {
                 // Dialog is show, make sure the accountState is deleted when the dialog is done.
-                connect(dialog, &QDialog::finished, accountState, &QObject::deleteLater);
+                connect(dialog, &UpdateUrlDialog::finished, accountState, [accountState](int result) {
+                    Q_UNUSED(result);
+                    accountState->deleteLater();
```
TheOneRing, line 177: Do we still need the account state?

src/gui/updateurldialog.cpp
```diff
@@ -0,0 +1,48 @@
+#include "updateurldialog.h"
```
TheOneRing, line 1: License header

src/gui/updateurldialog.h
fmoc: I'll try that.

src/gui/owncloudsetupwizard.cpp
fmoc: I don't think so, given it was deleted beforehand, too.

src/gui/accountstate.cpp
```diff
@@ -47,47 +47,30 @@ Q_LOGGING_CATEGORY(lcAccountState, "gui.account.state", QtInfoMsg)
...
-QDialog *AccountState::updateUrlDialog(const QUrl &newUrl)
+UpdateUrlDialog *AccountState::updateUrlDialog(const QUrl &newUrl)
 {
     // guard to prevent multiple dialogs
     if (_updateUrlDialog) {
         return nullptr;
     }
-    auto accept = [=] {
+
+    _updateUrlDialog = new UpdateUrlDialog(
```
TheOneRing, line 57: I guess it would make sense to move most of the text to the dialog itself, only then its possible to use it elsewhere

src/gui/owncloudsetupwizard.cpp
TheOneRing: We used the account state ptr, to be able to utilise updateUrlDialog.
Now that you refactored it we can use it directly?

src/gui/accountstate.cpp
fmoc: Is the new wizard going to use the exactly same wording?

src/gui/owncloudsetupwizard.cpp
fmoc: Indeed, we can just remove the helper entirely and move the code into the new class. That way, we don't need an account state pointer.

fmoc:
Should be good now. I'll squash the branch once approved.

TheOneRing (approved):

PR #9426, opened 2022-02-08
Remove Account dependency from OAuth

fmoc:
Could not reopen #9397, hence creating a new PR. This is a preparation for the new wizard.


fmoc:
Going to squash this branch once reviewed.

src/libsync/abstractcorejob.h
```diff
@@ -0,0 +1,69 @@
...
+
+    [[nodiscard]] const QVariant &result() const;
+
+    [[nodiscard]] const QString &errorMessage() const;
+    [[nodiscard]] QNetworkReply::NetworkError networkError() const;
+
+    [[nodiscard]] bool success() const;
+
+protected:
+    void finishWithResult(const QVariant &result);
```
TheOneRing, line 27: The name looks like a signal, maybe setResult?

src/libsync/abstractcorejob.h
```diff
@@ -0,0 +1,69 @@
...
+
+    void finishWithError(const QString &errorMessage, QNetworkReply::NetworkError networkError);
```
TheOneRing, line 29: Same

src/libsync/abstractcorejob.h
```diff
@@ -0,0 +1,69 @@
...
+public:
+    explicit AbstractCoreJobFactory(QNetworkAccessManager *nam, QObject *parent = nullptr);
+    ~AbstractCoreJobFactory() override;
+
+    virtual Job *startJob(const QUrl &url) = 0;
+
+protected:
+    [[nodiscard]] QNetworkAccessManager *nam() const;
+
+    static void finishJobWithSuccess(Job *job, const QVariant &result)
```
TheOneRing, line 55: Maybe also set*?

src/libsync/creds/oauth.cpp
```diff
@@ -286,27 +294,20 @@ void OAuth::startAuthentication()
...
+
+                    connect(job, &Job::finished, this, [=]() {
+                        if (!job->success()) {
                             httpReplyAndClose(socket, QByteArrayLiteral("500 Internal Server Error"),
-                                tr("<h1>Login Error</h1><p>Failed to retrieve user info</p>").toUtf8());
+                                tr("<h1>Login Error</h1><p>%1</p>").arg(job->errorMessage()).toUtf8());
                             emit result(Error);
                         } else {
-                            const QString user = job->data().value(QStringLiteral("ocs")).toObject().value(QStringLiteral("data")).toObject().value(QStringLiteral("id")).toString();
+                            const auto user = qvariant_cast<QString>(job->result());
```
TheOneRing, line 306: result toString? I'm not sure which one feels better...

And in general I try to not use auto with QStrings as concatenation with the string builder and auto can go 💥 .
If we always stick to QString that can't happen by accident.

src/libsync/creds/oauth.cpp
```diff
@@ -557,4 +549,70 @@ void OAuth::openBrowser()
     });
 }
 
+void OAuth::dynamicRegistrationDataReceived(const QVariantMap &dynamicRegistrationData)
+{
+    // the default class (used by the wizard only) doesn't know how to handle such data, and also doesn't need to
+    // only the derived AccountBasedOAuth class must be able to handle the data
+    (void)dynamicRegistrationData;
```
TheOneRing, line 556: Q_UNUSED? or unnamed parameter?
I guess dynamicRegistrationDataReceived could just be a signal?

src/libsync/creds/oauth.cpp
```diff
@@ -557,4 +549,70 @@ void OAuth::openBrowser()
...
+}
+
+void AccountBasedOAuth::startAuthentication()
+{
+    auto credentialsJob = _account->credentialManager()->get(dynamicRegistrationDataC());
+
+    connect(credentialsJob, &CredentialJob::finished, this, [=] {
+        credentialsJob->deleteLater();
+
+        qDebug() << credentialsJob->errorString() << credentialsJob->error();
```
TheOneRing, line 573: Debug leftover

src/libsync/determineauthtypejob2.cpp
```diff
@@ -0,0 +1,49 @@
...
+{
+}
+
+DetermineAuthTypeJobFactory::~DetermineAuthTypeJobFactory() = default;
+
+Job *DetermineAuthTypeJobFactory::startJob(const QUrl &url)
+{
+    auto job = new Job;
+
+    QNetworkRequest req(Utility::concatUrlPath(url, QStringLiteral("remote.php/dav/files/")));
```
TheOneRing, line 23: TODO: which url to use with spaces

src/libsync/determineauthtypejob2.cpp
```diff
@@ -0,0 +1,49 @@
+#include "determineauthtypejob2.h"
```
TheOneRing, line 1: Please add new classes to a subfolder, we will probably still need to add the to the src from this directories CMakeLists.txt as add_target_src dosn't support relative paths, but it will give the project at least some mental sturcture.

src/libsync/determineuserjobfactory.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include "common/utility.h"
+#include "creds/httpcredentials.h"
+#include <QJsonParseError>
+#include <QNetworkReply>
+
+Q_LOGGING_CATEGORY(lcDetermineUserJob, "sync.networkjob.determineuserjob", QtInfoMsg);
+
+using namespace OCC;
+
+DetermineUserJobFactory::DetermineUserJobFactory(QNetworkAccessManager *networkAccessManager, const QString &accessToken, QObject *parent)
```
TheOneRing, line 11: That one is a bit different, its a oc json job, it always has the account url as base and we should usually use the JsonApiJob class for it to ensure proper error handling.
As we need a non account bound version only in the Oath flow, I don't think we should make this class generally available.

src/libsync/resolveurljobfactory.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include "common/utility.h"
+#include <QNetworkReply>
+
+namespace {
+Q_LOGGING_CATEGORY(lcResolveUrl, "wizard.resolveurl")
+}
+
+namespace OCC {
+
+ResolveUrlJobFactory::ResolveUrlJobFactory(QNetworkAccessManager *nam, QObject *parent)
```
TheOneRing, line 11: Only relevant in the new wizard, no need to add it to libsync?
and unrelated to the pr?

TheOneRing (changes requested):

TheOneRing:
License hears are missing

src/libsync/determineuserjobfactory.cpp
fmoc: I'll move it there. Long term goal should be to eliminate the duplicate code, though. I could imagine that we can have a shared code base for both jobs, one of which is extended with account-related functionality. We just did something similar in https://github.com/owncloud/client/pull/9422.

src/libsync/abstractcorejob.h
fmoc: It does trigger the "this job has finished" signal, though. `set*` doesn't transport that message.

We should reconsider removing the `friend` class stuff and using public setters.

src/libsync/abstractcorejob.h
fmoc: We should reconsider removing the `friend` class stuff and using public setters IMO.

src/libsync/creds/oauth.cpp
fmoc: It used to be, when the factory (which then became a job, and later was replaced by a derived class) handled all these details, and the `OAuth` class did not have to know about such details. Now, there is no reason to expose this signal to the public any more.

src/libsync/creds/oauth.cpp
fmoc: I didn't mind having this debug snippet in our code base. But I'll remove it.

src/libsync/determineauthtypejob2.cpp
fmoc: Should I add such a TODO? Please don't hesitate to provide some text.

TheOneRing (approved):

fmoc:
The CI fails sporadically on `testTimeout`, reason unknown. I cannot reproduce it locally. Any ideas?

ownclouders:
Results for <strong>GUI-tests</strong> https://drone.owncloud.com/owncloud/client/10802/5/1
:boom: The GUI tests failed.
GUI Logs: (https://cache.owncloud.com/public/owncloud/client/10802/guiReportUpload/index.html)            
Server Logs: ((https://cache.owncloud.com/public/owncloud/client/10802/guiReportUpload/serverlog.log)

PR #9435, opened 2022-02-10
Change "Start on System Startup" to "Start on Login"

erikjv:
Because, well, the client isn't started on startup, but only when
logging in.


erikjv:
Please don't merge this. Need to add a comment in the code to explain what the scenario is for doing a case sensitive compare.

TheOneRing:
Thx

Issue #9121, opened 2021-10-05
Strange logging for TUS Upload in oCIS


dragotin:
I was trying to sync up a > 500 MB file to oCIS via Linux desktop client, version  2.9.0

The server pretends that this fails due to https://github.com/owncloud/ocis/issues/1205 and the client handles the operation as failed, but in fact, the file ends up on the server. 

What makes me writing this BR is that in the client logfile I can find the following logging:

```
10-05 14:06:15:623 [ debug sync.propagator.upload.tus ] [ OCC::PropagateUploadFileTUS::startNextChunk ]:        Starting creation with upload: "/openmediavault_5.3.9-amd64.iso"
10-05 14:06:15:623 [ debug sync.propagator.upload.tus ] [ OCC::PropagateUploadFileTUS::makeCreationWithUploadJob ]:     FullPath: "/openmediavault_5.3.9-amd64.iso"
10-05 14:06:15:623 [ debug sync.propagator.upload.tus ] [ OCC::PropagateUploadFileTUS::startNextChunk ]:        Offset: 0 0 Chunk: 586153984 0
10-05 14:06:15:623 [ warning default ]: QObject::connect(QNetworkReply, OCC::UploadDevice): invalid nullptr parameter
10-05 14:06:15:623 [ warning default ]: QObject::connect(QNetworkReply, OCC::PropagateUploadFileTUS): invalid nullptr parameter
10-05 14:06:15:623 [ info sync.accessmanager ]: 4 "" "https://ocis.team.owncloud.works/remote.php/dav/files/kfreitag/" has X-Request-ID "26cd29ba-1c4a-4fce-41f3-c94a6709440a"
10-05 14:06:15:623 [ debug sync.cookiejar ]     [ OCC::CookieJar::cookiesForUrl ]:      QUrl("https://ocis.team.owncloud.works/remote.php/dav/files/kfreitag/") requests: ()
```
Note the strange warnings from `QObject::connect()`!

PR #9437, opened 2022-02-11
Fix TUS progress reporting

TheOneRing:
Fixes: https://github.com/owncloud/client/issues/9121


TheOneRing:
> > I wonder which of these changes actually fixed the issue. Do you have to start the job before connecting signals?
> 
> I need to start the job in order to have a reply ready....

Thinking about it, this is still wrong if the job gets restarted.... :sigh:

TheOneRing:
AAAAA thats wrong all over the place...

TheOneRing:
Please check whether I overlooked one
`git grep "connect.*reply"`

erikjv (approved):

src/libsync/networkjobs.h
```diff
@@ -427,16 +430,22 @@ class OWNCLOUDSYNC_EXPORT SimpleNetworkJob : public AbstractNetworkJob
 
     void start() override;
 
+    void addNewReplyHook(std::function<void(QNetworkReply *)> &&hook);
```
fmoc, line 433: Might use a typedef here, DRY.

fmoc (approved):

PR #9438, opened 2022-02-11
Simplify etag parsing

TheOneRing:


fmoc (approved):

dragotin:
Well, this PR drops the log entry that tells that the two ETags are different. Not sure if that has any relevance, but it does.

Issue #9109, opened 2021-10-01
QT_LOGGING_RULES = "sync.httplogger=true" in squish tests


TheOneRing:
Enabling http logging can help track down exotic issues happening in the squish tests

Talank:
HTTP log has been enabled in #9443

TheOneRing:
Thx

PR #9443, opened 2022-02-14
[Tests-Only]Enabled http log on GUI tests

Talank:
Set up QT logging rules using environment variable to disable socketapi log and enable http log.

### related issues
- closes https://github.com/owncloud/client/issues/9109


saw-jan (approved):
LGTM

PR #9449, opened 2022-02-16
Use a base url with all network jobs

TheOneRing:
Those changes prepare the introduction of spaces, where each folder has its own webdav url.

While the code still contains some TODOS:
1. the wizard will be replaced soon.
2. the upload socket job will just not support spaces for now.
3. For selective sync etc, will be done once we support that with spaces.

### DIFF
--- a/src/gui/folderman.cpp
+++ b/src/gui/folderman.cpp
@@ -68,8 +68,8 @@ QString makeLegacyDbName(const OCC::FolderDefinition &def, const OCC::AccountPtr
     if (legacyUrl.path() == QLatin1String("/")) {
         legacyUrl.setPath(QString());
     }
-    const QString key = QStringLiteral("%1@%2:%3").arg(account->credentials()->user(), legacyUrl.toString(), def.targetPath);
-    return OCC::SyncJournalDb::makeDbName(def.localPath, QString::fromUtf8(QCryptographicHash::hash(key.toUtf8(), QCryptographicHash::Md5).left(6).toHex()));
+    const QString key = QStringLiteral("%1@%2:%3").arg(account->credentials()->user(), legacyUrl.toString(), def.targetPath());
+    return OCC::SyncJournalDb::makeDbName(def.localPath(), QString::fromUtf8(QCryptographicHash::hash(key.toUtf8(), QCryptographicHash::Md5).left(6).toHex()));
 }
 }
 
@@ -315,8 +315,8 @@ void FolderMan::setupFoldersHelper(QSettings &settings, AccountStatePtr account,
             const auto defaultJournalPath = [&account, folderDefinition] {
                 // if we would have booth the 2.9.0 file name and the lagacy file
                 // with the md5 infix we prefer the 2.9.0 version
-                const QDir info(folderDefinition.localPath);
-                const QString defaultPath = SyncJournalDb::makeDbName(folderDefinition.localPath);
+                const QDir info(folderDefinition.localPath());
+                const QString defaultPath = SyncJournalDb::makeDbName(folderDefinition.localPath());
                 if (info.exists(defaultPath)) {
                     return defaultPath;
                 }
@@ -332,6 +332,11 @@ void FolderMan::setupFoldersHelper(QSettings &settings, AccountStatePtr account,
                 }
                 return defaultPath;
             }();
+
+            // migration: 2.10 did not specify a webdav url
+            if (folderDefinition._webDavUrl.isEmpty()) {
+                folderDefinition._webDavUrl = account->account()->davUrl();
+            }
 
             // Migration: Old settings don't have journalPath
             if (folderDefinition.journalPath.isEmpty()) {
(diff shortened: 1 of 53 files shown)


src/libsync/abstractnetworkjob.cpp
```diff
@@ -144,19 +150,26 @@ bool AbstractNetworkJob::needsRetry() const
...
-void AbstractNetworkJob::sendRequest(const QByteArray &verb, const QUrl &url,
+void AbstractNetworkJob::sendRequest(const QByteArray &verb,
     const QNetworkRequest &req, QIODevice *requestBody)
 {
     _verb = verb;
     _request = req;
-    _request.setUrl(url);
     _requestBody = requestBody;
+    // TODO:
+    if (_request.url().isEmpty()) {
```
TheOneRing, line 160: One possibility for this to happen is https://github.com/owncloud/client/pull/9449/files#diff-1b4697b262724707f8fa1a7b3fa265141cc0e97d67dc0a2a38975ebcd1ad9612R697

Not sure how to handle this properly.

src/libsync/networkjobs.cpp
```diff
@@ -636,14 +613,15 @@ bool AvatarJob::finished()
...
+
+EntityExistsJob::EntityExistsJob(AccountPtr account, const QUrl &rootUrl, const QString &path, QObject *parent)
+    : AbstractNetworkJob(account, rootUrl, path, parent)
 {
 }
 
 void EntityExistsJob::start()
 {
-    sendRequest("HEAD", makeAccountUrl(path()));
+    sendRequest("HEAD");
```
fmoc, line 624: Since the URL is now considered essential data and passed in the constructor, the keyword should likely be managed the same way. Unless there is a job which uses multiple (but then I guess there's a violation of the SRP).

src/libsync/networkjobs.cpp
```diff
@@ -748,7 +724,7 @@ void SimpleNetworkJob::start()
 {
     Q_ASSERT(!_verb.isEmpty());
     // AbstractNetworkJob will take ownership of the buffer
-    sendRequest(_verb, jobUrl(), _request, _device);
+    sendRequest(_verb, _request, _device);
```
fmoc, line 727: See my previous comment, too. I think there is no reason for the `verb` to change after the class's construction. It should be managed by the superclass.

fmoc:
I think we need to put in some more thought how to reduce the complexity of using any of those jobs. There is a lot of duplicate code now which always looks the same: pass an account, pass some kind of base URL, pass some more data, ...

I think the abstract factory pattern will be of great use when introducing spaces. We can have an "old style" concrete factory which initializes the jobs in a way suitable for oC <= 10, and have a "space-enabled" factory that takes care of initializing jobs for specific spaces. The jobs still require to know the base URL, but the callers do not need to worry about this. All they get access to is the factory, which they can use to create the required jobs. This takes some of the complexity of handling spaces from users of the jobs. Should we have to ever adapt details about the spaces' server API, we can do so in one place only. Another advantage is that if we ever have to introduce an abstract base job (e.g., because some detail about the job has changed in some newer version and we want to retain compatibility with the old one, and do not want to hack support for both into one single class but rather introduce an abstract base and two concrete implementations), we can do so easily, because the object creation is handled outside the using classes.

src/gui/folderman.cpp
```diff
@@ -333,6 +333,11 @@ void FolderMan::setupFoldersHelper(QSettings &settings, AccountStatePtr account,
                 return defaultPath;
             }();
 
+            // 2.10 did not specify a webdav url
+            if (folderDefinition._webDavUrl.isEmpty()) {
```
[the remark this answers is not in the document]
TheOneRing: It is migration code, in the location where migrate all the other stuff.

src/gui/quotainfo.cpp
```diff
@@ -98,10 +98,9 @@ void QuotaInfo::slotCheckQuota()
         _job->deleteLater();
     }
 
-    AccountPtr account = _accountState->account();
-    _job = new PropfindJob(account, quotaBaseFolder(), this);
-    _job->setProperties(QList<QByteArray>() << "quota-available-bytes"
-                                            << "quota-used-bytes");
+    const AccountPtr &account = _accountState->account();
```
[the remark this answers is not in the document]
TheOneRing: Its a shared ptr, and we play one time less with its refs.

TheOneRing:
> I think we need to put in some more thought how to reduce the complexity of using any of those jobs. There is a lot of duplicate code now which always looks the same: pass an account, pass some kind of base URL, pass some more data, ...
> 
> I think the abstract factory pattern will be of great use when introducing spaces. We can have an "old style" concrete factory which initializes the jobs in a way suitable for oC <= 10, and have a "space-enabled" factory that takes care of initializing jobs for specific spaces. The jobs still require to know the base URL, but the callers do not need to worry about this. All they get access to is the factory, which they can use to create the required jobs. This takes some of the complexity of handling spaces from users of the jobs. Should we have to ever adapt details about the spaces' server API, we can do so in one place only. Another advantage is that if we ever have to introduce an abstract base job (e.g., because some detail about the job has changed in some newer version and we want to retain compatibility with the old one, and do not want to hack support for both into one single class but rather introduce an abstract base and two concrete implementations), we can do so easily, because the object creation is handled outside the using classes.

Patches welcome

src/libsync/abstractnetworkjob.cpp
TheOneRing: done

src/gui/folderman.cpp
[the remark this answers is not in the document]
fmoc: Might want to emphasize that in the comment.

fmoc, line 337:
```suggestion
            // migration: 2.10 did not specify a webdav url
            if (folderDefinition._webDavUrl.isEmpty()) {
```

fmoc (approved):

PR #9450, opened 2022-02-16
macOS: Do case-insensitive compare in auto-launch code

erikjv:
Case-sensitive compares will give problems when a launchd plist file
exists, but the executable gets upated, and the name only has a
different case (e.g.: owncloud -> ownCloud).


TheOneRing:
Thx

Issue #9030, opened 2021-09-15
Desktop client not showing overlay icons if name has a space at the end


belth245:
### Expected behaviour
- Finder should show the expected status icon
- Finder context menu shows usual options including VFS options

### Actual behaviour
- Finder shows no icon at all
- Finder context menu is not showing the usual options

<img width="896" alt="Bildschirmfoto 2021-09-15 um 14 42 09" src="https://user-images.githubusercontent.com/8570318/133441137-f7e11566-dfe1-48b6-bceb-c96392e9e02e.png">

<img width="893" alt="Bildschirmfoto 2021-09-15 um 14 42 55" src="https://user-images.githubusercontent.com/8570318/133441168-5480aaca-3b16-4a18-bd80-4e86ac7695e2.png">

### Steps to reproduce
1. Create folder with a space at the end of the name
2. Navigate into the folder and back out again
3. Icon has vanished

Operating system:
MacOS 11.5.2

OwnCloud Client 2.9.0 (build 5150)
OwnCloud Server: 10.7.0.4 Enterprise

OS language: German / English

Most actions seem to appropriate values but for some actions the client seems to trim spaces off (i.e. folder sync status and context menu):

<img width="1253" alt="Log_01_FolderStatus" src="https://user-images.githubusercontent.com/8570318/133441254-3772df2c-8530-4efe-a4df-c7e7a3985902.png">

<img width="1186" alt="Log_02_MenuItems" src="https://user-images.githubusercontent.com/8570318/133441299-490f23c2-d348-4d82-88d2-d808ce41e9db.png">

TheOneRing:
You hit a limitation of our old socket api (We have plans to modernise it).
Be aware that files with spaces at the and also can't be synced to Windows.

PR #9452, opened 2022-02-16
SocketAPI: Do not strip trailing whitespace from paths

erikjv:
When stripped, the wrong file name is queried in the DB, and the status
badges shown in Finder will not be there.

https://github.com/owncloud/client/issues/9030


TheOneRing:
Ouch, that was me, thx for fixing.

TheOneRing:
Changlog is wrong, needs to be `Bugfix: `

TheOneRing:
And could you uses the `Fixed: #issue` syntax so that the issue is closed once the change reaches master.

TheOneRing (approved):

Issue #9238, opened 2021-11-23
Sync Protocol fails to mention source name when renaming a file


jnweiger:
Seen in 2.9.2-rc5 

* connect to demo.owncloud.org
* wait for the initial sync to succeed
* visit the Photos folder with a local file browser, rename one of the photos to `foo.jpg`
* after the next sync run, the Sync Protocol tab has
![image](https://user-images.githubusercontent.com/1108546/143054072-194c9b93-28f3-4afd-a562-d0eecd5d0d7b.png)

* The user cannot tell, what the old name of the photo was. BAD
* The new name of the file is shown twice. That is redundant.

Expected behaviour: 
* For a move, report both, old and new name. E.g.
  * Action: Moved from Photos/Portugal.jpg
  * File: Photos/foo.jpg

PR #9453, opened 2022-02-17
Bugfix: mention source file in activity tab when renaming

erikjv:
Fixed: https://github.com/owncloud/client/issues/9238

### DIFF
--- a/changelog/unreleased/9238
+++ b/changelog/unreleased/9238
@@ -1,0 +1,4 @@
+Bugfix: mention source file in activity tab when renaming
+
+https://github.com/owncloud/client/issues/9238
+https://github.com/owncloud/client/pull/9453
--- a/src/libsync/progressdispatcher.cpp
+++ b/src/libsync/progressdispatcher.cpp
@@ -45,7 +45,7 @@ QString Progress::asResultString(const SyncFileItem &item)
         return QCoreApplication::translate("progress", "Deleted");
     case CSYNC_INSTRUCTION_EVAL_RENAME:
     case CSYNC_INSTRUCTION_RENAME:
-        return QCoreApplication::translate("progress", "Moved to %1").arg(item._renameTarget);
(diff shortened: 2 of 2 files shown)


src/libsync/progressdispatcher.cpp
```diff
@@ -45,7 +45,7 @@ QString Progress::asResultString(const SyncFileItem &item)
         return QCoreApplication::translate("progress", "Deleted");
     case CSYNC_INSTRUCTION_EVAL_RENAME:
     case CSYNC_INSTRUCTION_RENAME:
-        return QCoreApplication::translate("progress", "Moved to %1").arg(item._renameTarget);
+        return QCoreApplication::translate("progress", "%1 moved to %2").arg(item._file, item._renameTarget);
```
TheOneRing, line 48: Can you grep the code base, I'm pretty sure we have duplication of similar messages.

changelog/unreleased/9238
```diff
@@ -0,0 +1,3 @@
+Bugfix: mention source file in activity tab when renaming
+
+Fixed: https://github.com/owncloud/client/issues/9238
```
TheOneRing, line 3: Please remove the Fixed: here.

TheOneRing (changes requested):

src/libsync/progressdispatcher.cpp
erikjv: grepping for "moved to" (case insensitive) only yields https://github.com/owncloud/client/blob/90708a5b02eaae871b90abde2008435d273301bf/src/gui/folder.cpp#L499

I wouldn't change either: this one is displayed in limited space, and the 2 filenames make it quite long. The one in `folder.cpp` is in present perfect tense (passive voice) instead of past tense (active voice), just like all other messages  around it.

TheOneRing: I think for another release we should review all our strings...

fmoc (approved):

PR #9456, opened 2022-02-18
Remove translations before we update them

TheOneRing:


.github/workflows/translate.yml
```diff
@@ -22,6 +22,9 @@ jobs:
           submodules: recursive
           ref: 2.8
 
+      - name: l10n-clean
+        run: bash .github/workflows/run-in-docker.sh rabits/qt:5.12-desktop "cd translations && rm *.ts"
```
fmoc, line 26: Should be integrated into the Makefile. Also, we already have a `clean` job, so there's a name clash. The Makefile also contains a `clean` target already.

changelog/unreleased/9105
```diff
@@ -0,0 +1,7 @@
+Bugfix: Remove outdated translations
+
+Due to a bug we where not removing the translations that fell below
```
dschmidt, line 3:
```suggestion
Due to a bug we were not removing the translations that fell below
```

translations/Makefile
```diff
@@ -1,23 +1,33 @@
...
 #--------------------------------------
+# pull all changes push none, intended for local testing
+.PHONY: dry
+dry:  l10n-remove-old l10n-read l10n-pull
 
+# remove outdated translations
+.PHONY: l10n-remove-old
+l10n-remove-old:
+	rm *.ts
+	git checkout client_en.ts
```
dschmidt, line 12: Ah, nice 💪🏻

PR #9465, opened 2022-02-22
Improve poll-interval comments

phil-davis:
PR #8777 made the client responsive to the "pollinterval" setting reporting  by the server in the capabilities.

The value of pollintervalis now interpreted as milliseconds.

This PR clarifies some comments. Hopefully I have made it clearer for future readers.

### DIFF
--- a/src/gui/folderman.cpp
+++ b/src/gui/folderman.cpp
@@ -36,6 +36,8 @@
 #include <QSet>
 #include <QNetworkProxy>
 
+using namespace std::chrono_literals;
+
 namespace {
 /*
  * [Accounts]
--- a/src/gui/folderman.cpp
+++ b/src/gui/folderman.cpp
@@ -145,10 +145,12 @@ FolderMan::FolderMan(QObject *parent)
 
     _socketApi.reset(new SocketApi);
 
-    // Set the remote poll interval fixed to 1 second (1000 milliseconds)
-    // That does not mean that it polls every 1 second, but it checks every 1 second
-    // if one of the folders is due to sync.
-    _etagPollTimer.setInterval(1000);
+    // Set the remote poll interval fixed to 10 seconds.
+    // That does not mean that it polls every 10 seconds, but it checks every 10 seconds
+    // if one of the folders is due to sync. This means that if the server advertises a
+    // pollinterval that is not a multiple of 10 seconds, then that pollinterval will be
+    // rounded up to the next 10 seconds in practice. 10-second granularity is acceptable.
+    _etagPollTimer.setInterval(10s);
     QObject::connect(&_etagPollTimer, &QTimer::timeout, this, &FolderMan::slotEtagPollTimerTimeout);
     _etagPollTimer.start();
 
--- a/src/gui/folderman.cpp
+++ b/src/gui/folderman.cpp
@@ -145,8 +145,8 @@ FolderMan::FolderMan(QObject *parent)
 
     _socketApi.reset(new SocketApi);
 
-    // Set the remote poll interval fixed to 10 seconds.
-    // That does not mean that it polls every 10 seconds, but it checks every 10 secs
+    // Set the remote poll interval fixed to 1 second (1000 milliseconds)
+    // That does not mean that it polls every 1 second, but it checks every 1 second
     // if one of the folders is due to sync.
     _etagPollTimer.setInterval(1000);
     QObject::connect(&_etagPollTimer, &QTimer::timeout, this, &FolderMan::slotEtagPollTimerTimeout);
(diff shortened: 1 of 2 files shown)


src/libsync/configfile.cpp
```diff
@@ -93,7 +93,7 @@ const QString moveToTrashC() { return QStringLiteral("moveToTrash"); }
 }
 
 QString ConfigFile::_confDir = QString();
-const std::chrono::seconds DefaultRemotePollInterval { 30 }; // default remote poll time in milliseconds
+const std::chrono::seconds DefaultRemotePollInterval { 30 }; // default remote poll time in seconds
```
fmoc, line 96: I think you can just get rid of "in seconds", since that's clear from the type. Should one ever change the type to, e.g., `std::chrono::milliseconds` or `std::chrono::minutes`, the comment would have to be touched again.

phil-davis: After removing "in seconds" from the comment, it ended up pretty much repeating the words that are already in the const name, which is more redundant comment crud that can get out-of-sync with the real code. So I removed the whole comment.

ownclouders:
Results for <strong>GUI-tests</strong> https://drone.owncloud.com/owncloud/client/10858/5/1
:boom: The GUI tests failed.
GUI Logs: (https://cache.owncloud.com/public/owncloud/client/10858/guiReportUpload/index.html)            
Server Logs: ((https://cache.owncloud.com/public/owncloud/client/10858/guiReportUpload/serverlog.log)

src/gui/folderman.cpp
```diff
@@ -145,8 +145,8 @@ FolderMan::FolderMan(QObject *parent)
 
     _socketApi.reset(new SocketApi);
 
-    // Set the remote poll interval fixed to 10 seconds.
-    // That does not mean that it polls every 10 seconds, but it checks every 10 secs
+    // Set the remote poll interval fixed to 1 second (1000 milliseconds)
```
TheOneRing, line 148: That pretty much looks like a bug instead, 1s polling sounds ridiculous.

TheOneRing: Looking of the change that introduces that, yes that should be 10s.
Could you instead of fixing the doc, replace the `1000` with `10s`

phil-davis: @TheOneRing you mean replace `1000` with `10000` ?

IMO the side-effect will be that if someone sets the server to advertise a `pollinterval` of 8 seconds, for example, then actually that will not be effective. This code here will only wake up every 10 seconds to see if the `pollinterval` has expired.

Maybe set it to `5000` (5 seconds)? And then the client will respect accurately a server `pollinterval` that is a multiple of 5000. (And respect less accurately a `pollinterval` like 18000)

TheOneRing: So we have a poll interval per sync folder pair, this can be multiple and we again have multiple accounts.
The poll interval here only calls a function that iterates over the sync folder pairs and looks for those in need of a fresh poll.
The whole code should be reworked as it offers lots of limitations, and we plan to have parallel syncs in the future.

I meant `_etagPollTimer.setInterval(10s);` a newer version of C++ now allows us to use seconds as type.

PS: I think a poll interval below the random value of 30s is pointless.

phil-davis: done

ownclouders:
Results for <strong>GUI-tests</strong> https://drone.owncloud.com/owncloud/client/10862/5/1
:boom: The GUI tests failed.
GUI Logs: (https://cache.owncloud.com/public/owncloud/client/10862/guiReportUpload/index.html)            
Server Logs: ((https://cache.owncloud.com/public/owncloud/client/10862/guiReportUpload/serverlog.log)

ownclouders:
Results for <strong>GUI-tests</strong> https://drone.owncloud.com/owncloud/client/10864/5/1
:boom: The GUI tests failed.
GUI Logs: (https://cache.owncloud.com/public/owncloud/client/10864/guiReportUpload/index.html)            
Server Logs: ((https://cache.owncloud.com/public/owncloud/client/10864/guiReportUpload/serverlog.log)

ownclouders:
Results for <strong>GUI-tests</strong> https://drone.owncloud.com/owncloud/client/10865/5/1
:boom: The GUI tests failed.
GUI Logs: (https://cache.owncloud.com/public/owncloud/client/10865/guiReportUpload/index.html)            
Server Logs: ((https://cache.owncloud.com/public/owncloud/client/10865/guiReportUpload/serverlog.log)

TheOneRing:
I added a missing namespace.

phil-davis:
IMO this is ready for review. Someone please review, approve and it can be merged.

TheOneRing:
thx

PR #9466, opened 2022-02-22
Fix timeouts in OAuth

fmoc:
Forgot to add proper handling in my previous PR, #9426. This reliably fixes the `OAuthTest` (tested with a sample size of 1000 against the previous implementation, where about 9% of the test runs failed).

### DIFF
--- a/src/libsync/abstractcorejob.h
+++ b/src/libsync/abstractcorejob.h
@@ -18,6 +18,7 @@
 #include <QNetworkAccessManager>
 #include <QNetworkReply>
 
+#include "abstractnetworkjob.h"
 #include "owncloudlib.h"
 
 namespace OCC {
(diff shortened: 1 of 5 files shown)


src/libsync/abstractcorejob.h
```diff
@@ -122,6 +123,20 @@ class OWNCLOUDSYNC_EXPORT AbstractCoreJobFactory : public QObject
      */
     static void setJobError(CoreJob *job, const QString &errorMessage, const QNetworkReply::NetworkError networkError);
 
+    /**
+     * Factory to create QNetworkRequests with properly set timeout.
+     */
+    template <typename... Params>
+    static OWNCLOUDSYNC_EXPORT QNetworkRequest makeRequest(Params... params)
```
TheOneRing, line 130: hmmm I'm tempted to suggest subclassing and enforcing the use of the new class in the factory but...

TheOneRing:
Rebasing on master could have an influence on the result of the test.

src/libsync/abstractcorejob.h
fmoc: As long as one just needs a "configured" object, subclassing should not be used in my opinion. Sure, this is not ideal, but the long term goal is to simplify the job infrastructure and eliminate duplicate structures anyway.

ownclouders:
Results for <strong>GUI-tests</strong> https://drone.owncloud.com/owncloud/client/10868/5/1
:boom: The GUI tests failed.
GUI Logs: (https://cache.owncloud.com/public/owncloud/client/10868/guiReportUpload/index.html)            
Server Logs: ((https://cache.owncloud.com/public/owncloud/client/10868/guiReportUpload/serverlog.log)

Issue #9367, opened 2022-01-17
crash on remove account


Talank:
We have seen this error message before, and haven't seen it in a while. However, we need to investigate why it is failing. Since no changes has been merged, and it failed in only one of the nightly build during the weekend, **it clearly is a flaky test.** 

The error details filtered by our log parser is as follows: 
```json
{
        "Error Details": "LookupError: Lost connection to AUT",
        "Feature": "remove account connection",
        "Feature File": "tst_removeAccountConnection",
        "Scenario": "remove an account connection",
        "Test Step": "Then an account should be displayed with the displayname Alice Hansen and host %local_server_hostname%"
}
```

Related drone run (failing): https://drone.owncloud.com/owncloud/client/10363/3/14

individual-it:
from the logs:
`02:23:39:023 Info: AUT process '/drone/src/build-GUI-tests/bin/owncloud OWNCLOUD_CORE_DUMP=1 -s --logfile - --logdebug --logflush --confdir /tmp/bdd-tests-owncloud-client/' (PID 5279) finished with exit status -1: Received signal number 11.`

TheOneRing:
Yay a crash, shouldn't we get a backtrace?

Talank:
Yup. It was a crash. AUT coredump is still not printed in the log because of https://github.com/owncloud/client/issues/9259. I will look into the fixing apport problem so that this unavailability of AUT crash log does not happen again.
![Screenshot from 2022-01-19 09-52-11](https://user-images.githubusercontent.com/27364813/150062282-ba7967e7-c5b1-473f-9d74-61b81b6de52b.png)

individual-it:
happened again today https://drone.owncloud.com/owncloud/client/10411/3/14

Talank:
Failed again today https://drone.owncloud.com/owncloud/client/10514/3/14

individual-it:
@Talank Any idea what happens here and what we could do?

Talank:
Well, we do have the squish backtrace in the log. But clearly it is an AUT crash causing the connection lost error. We can start investigating from the log that we already have, but I highly doubt that there will be any other outcomes from that investigation. The backtracking works perfectly in local test runs so we can try to reproduce it locally, which might give us some additional idea (if we can reproduce it).

Talank:
@TheOneRing I reproduced the issue locally. Here is thefull  backtrace in this file 
[...]

PR #9475, opened 2022-02-28
Fix shared pointer deletion before the slot was invoked

TheOneRing:
The fix involves some guesses.

imho: we should not use shared pointers at all and rather call deleteLater but the code is what it is.

Fixes: #9367


fmoc (approved):

TheOneRing:
> Looks sensible to me. I guess the actual fix was using `copy` in https://github.com/owncloud/client/pull/9475/files#diff-e3cf730b014ea3f7217fa08347f929d739130e6b21f6d32c3eaf824880d04e6eR394?

Yes

Issue #9425, opened 2022-02-08
[QA] Account Filter for Server Activity does not do anything


jnweiger:
Seen on 2.10.0rc3 (build 6436) in Linux Mint

* connect multiple servers.
* wait for server activity messages to come
* view Activity -> Server Activity
* click Filter, three accounts are shown there. Choose one.
* Still Activity messages from all accounts are shown. BAD
 
![image](https://user-images.githubusercontent.com/1108546/152970817-396c8fab-222b-409a-aff8-4f9c09b21eba.png)

The screenhot shows the situation after re-opening the Filter dropdown. The Account column shows accounts not filtered for.

The available filters in the the other tabs (Activity -> Sync Protocol tab and in Activity -> Not Synced) work fine.

jnweiger:
@enbrnz @gabi18 fyi

gabi18:
Also seen with ownCloud 2.10.0 (build 6519) on Windows 10 20H2

jnweiger:
UX improvement ideas:

* Replace "no filter" with "All" (avoid a brain-erors due to double negatives)
* When a filter is active, replace the Label "Filter" on the drop down button with something that indicates that a filter is active.

PR #9481, opened 2022-03-02
Fix account filtering in Server Activity tab

erikjv:
Fixes: https://github.com/owncloud/client/issues/9425

### DIFF
--- a/changelog/unreleased/9425
+++ b/changelog/unreleased/9425
@@ -1,0 +1,3 @@
+Bugfix: Fix Account Filter for Server Activity tab
+
+https://github.com/owncloud/client/pull/9481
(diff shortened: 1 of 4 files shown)


changelog/unreleased/9425
```diff
@@ -0,0 +1,6 @@
+Bugfix: Fix Account Filter for Server Activity tab
+
+This fixes the "bug" part of the reported issue. The UI/UX improvements
+are done separately.
```
TheOneRing, line 4: I'm not sure the second sentence should end up in the change of.

fmoc: I'd also rather add PRs that implement UI changes to the existing changelog item instead of announcing them in this one.

TheOneRing (approved):

PR #9487, opened 2022-03-07
Fix broken context menu entry

fmoc:


TheOneRing (approved):

gabi18:
Better have lower case "browser" here: "Show in Browser" to be consistent with other menu entries

dschmidt (approved):

PR #9489, opened 2022-03-07
  CheckServerJob: clear cookies under more circumstances 

TheOneRing:


src/libsync/networkjobs.cpp
```diff
@@ -430,10 +431,19 @@ void CheckServerJob::start()
     req.setAttribute(QNetworkRequest::RedirectPolicyAttribute, QNetworkRequest::NoLessSafeRedirectPolicy);
     req.setRawHeader(QByteArrayLiteral("OC-Connection-Validator"), QByteArrayLiteral("desktop"));
     req.setMaximumRedirectsAllowed(_maxRedirectsAllowed);
+
+    if (_clearCookies && Theme::instance()->connectionValidatorClearCookies()) {
```
fmoc, line 435: Since you're not running that check elsewhere, does it make sense here? Perhaps the variable should be called `_requestClearCookies` to signalize it won't necessarily clear the cookies.

fmoc (approved):

src/libsync/networkjobs.cpp
TheOneRing: The feature is opt in and we don't want to mess with other providers.

PR #9492, opened 2022-03-08
Only log after the logger is set up

### LINKED ISSUE #9486: Only log after the logger is set up
https://github.com/owncloud/client/blob/e8afc2b14c6841765a8316d314bf8319192f4533/src/gui/application.cpp#L253

Also compare to https://github.com/owncloud/client/blob/96d6dd3ae6e67effcea558acce6ef5dea5198f5d/src/gui/application.cpp#L234-L238

(We fixed the same issue twice)

fmoc:
Fixes #9486

### DIFF
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -248,19 +248,10 @@ Application::Application(int &argc, char **argv)
     setApplicationVersion(_theme->versionSwitchOutput());
 
 #if defined(OC_PLUGIN_DIR) && defined(Q_OS_LINUX)
+    // logged below, must be done once logger has been set up
     const QString extraPluginDir = QDir(QApplication::applicationDirPath()).filePath(QStringLiteral(OC_PLUGIN_DIR));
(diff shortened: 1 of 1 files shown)


src/gui/application.cpp
```diff
@@ -282,6 +273,19 @@ Application::Application(int &argc, char **argv)
     setupLogging();
     setupTranslations();
 
+#if defined(OC_PLUGIN_DIR) && defined(Q_OS_LINUX)
+    qCInfo(lcApplication) << "Added extra plugin search path:" << extraPluginDir;
```
TheOneRing, line 277: We could get rid of the extra ifdef if we just log the `libraryPath()`

fmoc: Right. You mentioned that before.

TheOneRing (approved):

TheOneRing:
I assume the crash in the unit test to be unrelated.

PR #9500, opened 2022-03-09
Lock upload during finalization

TheOneRing:
| Client| Server| State|
|---|---|---|
| A | A |Same version on both sides
| B | A | Versions differ, client starts upload
| B | A | During the upload, B gets locked
| B | B | B is uploaded, Local B is locked so we don't enter the state to the db
| C | B | The local db was not updated, we don't know about B on the server, local file gets updated and becomes C
| C | B | We have C locally and B on the server -> clash

### DIFF
--- a/changelog/unreleased/enterprise-5052
+++ b/changelog/unreleased/enterprise-5052
@@ -1,0 +1,6 @@
+Bugfix: Don't publish upload if we can't finish the transaction in the client
+
+When a file gets locked during an upload we aborted after the upload finished on the server.
+Resulting in a divergence of the local and remote state which could lead to conflicts.
+
+https://github.com/owncloud/enterprise/issues/5052
--- a/src/common/utility_win.cpp
+++ b/src/common/utility_win.cpp
@@ -319,4 +319,31 @@ Utility::NtfsPermissionLookupRAII::~NtfsPermissionLookupRAII()
     qt_ntfs_permission_lookup--;
 }
 
+
+Utility::Handle::Handle(HANDLE h, std::function<void(HANDLE)> &&close)
+    : _handle(h)
+    , _close(std::move(close))
+{
+}
+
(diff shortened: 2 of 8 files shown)


src/common/utility_win.cpp
```diff
@@ -319,4 +319,31 @@ Utility::NtfsPermissionLookupRAII::~NtfsPermissionLookupRAII()
     qt_ntfs_permission_lookup--;
 }
 
+
+Utility::Handle::Handle(HANDLE h, std::function<void(HANDLE)> &&close)
+    : _handle(h)
```
dragotin, line 324: is the close function used anywhere so far? If not I'd rather not have it. You're only using the default, right?

dragotin (commented):
This patch means that any call to `isFileLocked()` creates a lock on the file, which is released automatically. That function is called often, are you sure there are no side effects?

src/common/utility_win.cpp
TheOneRing: In the vfs plugin, the handle class is just moved to the main repo.

TheOneRing:
> This patch means that any call to `isFileLocked()` creates a lock on the file, which is released automatically. That function is called often, are you sure there are no side effects?

We did the same before, just explicitly called CloseHandle before.

dragotin (approved):
ok

changelog/unreleased/enterprise-5052
```diff
@@ -0,0 +1,6 @@
+Bugfix: Don't publish upload if we can't finish the transaction in the client
+
+When a file ges locked during an upload we aborted after the upload finished on the server.
```
dschmidt, line 3:
```suggestion
When a file gets locked during an upload we aborted after the upload finished on the server.
```

dschmidt (approved):
looks reasonable ... but I'm not familiar with part of the code base

PR #9502, opened 2022-03-09
Include locale in language dropdown

### LINKED ISSUE #9460: Provide informal german translations
The community is providing informal German translations for years.

We however don't even build them.

ISO allows the `-informal` suffix and the Qt supports it as well.

# TODO: 
## in the client:
1. Display the lang code in the language selector to allow multiple "German translations"

image

fmoc:
Fixes #9460

screenshot_2022-03-09_18-21-07
.

### DIFF
--- a/src/gui/generalsettings.cpp
+++ b/src/gui/generalsettings.cpp
@@ -34,11 +34,11 @@
 #include "config.h"
 #include "translations.h"
 
-#include <QNetworkProxy>
 #include <QDir>
-#include <QScopedValueRollback>
 #include <QMessageBox>
+#include <QNetworkProxy>
 #include <QOperatingSystemVersion>
+#include <QScopedValueRollback>
 
 namespace OCC {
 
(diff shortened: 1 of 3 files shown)


src/gui/generalsettings.cpp
```diff
@@ -282,10 +283,15 @@ void GeneralSettings::saveMiscSettings()
...
         // therefore, we can store them directly in the config file
         // in future versions, we will likely display nice names instead of locales to improve the user experience
-        const auto pickedLanguageName = _ui->languageDropdown->itemText(pickedLanguageIndex);
-        const auto pickedLanguageLocale = localesToLanguageNamesMap.key(pickedLanguageName);
 
-        cfgFile.setUiLanguage(pickedLanguageLocale);
+        QRegularExpression regex(".+\\s+\\(([^)]+)\\)");
+        Q_ASSERT(regex.isValid());
+
+        auto match = regex.match(_ui->languageDropdown->itemText(pickedLanguageIndex));
```
TheOneRing, line 290: Comboboxes also provide https://doc.qt.io/qt-5/qcombobox.html#itemData which could hold the file name or raw language etc, might be more elegant than the regex.

fmoc: Good hint. I was looking for something like this before, actually. Must have missed it.

TheOneRing (approved):

PR #9503, opened 2022-03-09
Use factories to construct QSettings objects

fmoc:
Extracted from #9376, as requested by @TheOneRing.

Fun fact: I accidentally cherry-picked the commit on 2.10 first and then spent a lot of time trying to figure out why it wouldn't build. Turns out the issue was that the old branch uses `-std=gnu++14` vs. `-std=gnu++17 -fsyntax-only` on `master` (not sure about the latter, though). In C++14, return-by-value in the `makeQSettings` factory calls the copy constructor for some reason I don't know, which `QSettings` disabled, and I found no way to use move-construction instead.

The pointer factory is there for `ProxyAuthHandler`, by the way. I didn't want to rewrite that class. Perhaps we can avoid `makeQSettingsPtr` by creating a new object on the heap and then move-assigning the old one to the one on the heap?

### DIFF
--- a/src/gui/proxyauthhandler.cpp
+++ b/src/gui/proxyauthhandler.cpp
@@ -42,7 +42,7 @@ ProxyAuthHandler::ProxyAuthHandler()
     _dialog = new ProxyAuthDialog();
 
     _configFile.reset(new ConfigFile);
-    _settings.reset(ConfigFile::makeQSettingsPtr());
+    _settings.reset(new QSettings(ConfigFile::makeQSettings()));
     _settings->beginGroup(QLatin1String("Proxy"));
     _settings->beginGroup(QLatin1String("Credentials"));
 }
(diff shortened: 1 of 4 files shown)


fmoc:
The suggested workaround actually seems to work, something I didn't necessarily expect. Please have a detailed look, though, we want to be sure that the initialization works.

Long-term, we should open an issue to rework `ProxyAuthenticationHandler` so it does not need a pointer on the `QSettings` object. I haven't checked what's necessary (perhaps using a pointer was a workaround to having no copy constructor?).

ownclouders:
Results for <strong>GUI-tests</strong> https://drone.owncloud.com/owncloud/client/10997/5/1
GUI Logs: (https://cache.owncloud.com/public/owncloud/client/10997/guiReportUpload/index.html)
Server Logs: (https://cache.owncloud.com/public/owncloud/client/10997/guiReportUpload/serverlog.log)

ownclouders:
Results for <strong>GUI-tests</strong> https://drone.owncloud.com/owncloud/client/10998/5/1
GUI Logs: (https://cache.owncloud.com/public/owncloud/client/10998/guiReportUpload/index.html)
Server Logs: (https://cache.owncloud.com/public/owncloud/client/10998/guiReportUpload/serverlog.log)

TheOneRing:
makeQSettings feels a bit to generic but ...
LGTM

TheOneRing (approved):

Issue #9336, opened 2022-01-06
When password protecting a previously created public link displays password policy error, the checkbox will be unchecked


saw-jan:
### Steps to reproduce
1. Enable `Password Policy` app on the server 
2. Set minimum password characters to `8` for public links (from Settings -> Admin -> Security)
3. From desktop-client, open public link sharing dialog for a file and create a public link share
6. Edit previously created public link
7. Enable (check) `Password protect` and enter password `1234` 
8. Click `Set password` 
(throws error `The password is too short...` and the checkbox will get unchecked but password filed can be edited)

### Expected behavior
expected `Password protect` checkbox to be checked

### Actual behavior
`Password protect` checkbox will get unchecked

![pp01](https://user-images.githubusercontent.com/52366632/148367078-c5794ac6-256f-4659-91ca-e1b3405e9603.png)
![pp02](https://user-images.githubusercontent.com/52366632/148367101-d5d4e4a3-d7bc-4bce-aa41-fd197d3f4091.png)

### Client configuration
Client version: `2.10.0rc3.6436`
Operating system: `win10x64`

fmoc:
I tested the dialog a bit. Seems like keeping the checkbox checked and the password field filled with the old value is what we do when creating new shares, too. The only difference there is that we do not (have to?) disable the close button there, since the link just hasn't been created.

I'm actually in favor with moving the editing controls into a secondary dialog. IMO this increases the usability, since with a modal dialog one cannot change to another entry in the list for instance while the other one's changes are sent to the server (IMO a "save"/"commit" button is missing anyway for editing, changes should be done in a "transactional" way). What do you think?

For a short-term fix, I'm evaluating the first option which will be easier to implement. But long-term, I'd like to rework the share dialog.

PR #9508, opened 2022-03-11
Leave password line edit enabled on errors

fmoc:
Contributes to #9336.

New behavior:

screenshot_2022-03-11_10-53-21

I think we shouldn't disable the close button, because that'll bloat the (already relatively complex) code within the dialog, and requires the user to uncheck the checkbox first (in that case, we'd actually have to disable the close button generally when information is missing to make the UX consistent). Just removing the password and asking for a new one in the line edit combined with displaying the error message gives the user sufficient feedback to continue. Keeping the password makes little sense either, we should not store secrets for longer than necessary.

Upon success, a generic placeholder will be inserted in the line edit, as before.

### DIFF
--- a/src/gui/sharelinkwidget.cpp
+++ b/src/gui/sharelinkwidget.cpp
@@ -363,14 +363,13 @@ void ShareLinkWidget::slotShareSelectionChanged()
             _ui->lineEdit_password->setEnabled(false);
         }
     } else if (!selectionUnchanged) {
+        // we used to disable the checkbox (and thus the line edit) when setting a password failed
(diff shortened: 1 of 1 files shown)


fmoc:
The format is enforced by the picker. For instance, with the `en` locale, a short format `mm/dd/YY` is used. You cannot enter invalid characters, I haven't managed to generate an error message.

(By the way, I don't like abbreviated years in date specifications... perhaps we should change that in a future release.)

TheOneRing (approved):

PR #9513, opened 2022-03-11
Fix filter menus

erikjv:
- replaced "No filter" option text with "All", to avoid the "No filter
  is not enabled" situation
- replace the "Filter" label on the button with "1 Filter"/"2 Filters"
  when a filter is active, so a user can immediately see that without
  having to open the filter pop-up

### DIFF
--- a/changelog/unreleased/9513
+++ b/changelog/unreleased/9513
@@ -1,0 +1,7 @@
+Bugfix: improve filter pop-up menu and button
+    
+- replaced "No filter" option text with "All", to avoid the "No filter is not enabled" situation
+- replace the "Filter" label on the button with "1 Filter"/"2 Filters" when a filter is active, so a user can immediately see that without having to open the filter pop-up
+
+https://github.com/owncloud/client/issues/9425
+https://github.com/owncloud/client/pull/9513
--- a/src/gui/issueswidget.cpp
+++ b/src/gui/issueswidget.cpp
@@ -15,23 +15,24 @@
 #include <QtGui>
 #include <QtWidgets>
 
-#include "issueswidget.h"
+#include "account.h"
+#include "accountmanager.h"
+#include "accountstate.h"
+#include "common/syncjournalfilerecord.h"
+#include "commonstrings.h"
 #include "configfile.h"
-#include "syncresult.h"
-#include "syncengine.h"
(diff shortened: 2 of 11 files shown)


src/gui/issueswidget.cpp
```diff
@@ -281,6 +295,16 @@ void IssuesWidget::slotItemCompleted(const QString &folder, const SyncFileItemPt
     _model->addProtocolItem(ProtocolItem { folder, item });
 }
 
+void IssuesWidget::filterDidChange()
+{
+    int filterCount = _statusSortModel->filterCount();
```
TheOneRing, line 300: I can't really follow what why count here and why we increase it if the regexp is empty.
Could you add a comment?

changelog/unreleased/9513
```diff
@@ -0,0 +1,5 @@
+Bugfix: improve filter pop-up menu and button
+    
+- replaced "No filter" option text with "All", to avoid the "No filter is not enabled" situation
+- replace the "Filter" label on the button with "1 Filter"/"2 Filters" when a filter is active, so a user can immediately see that without having to open the filter pop-up
+
```
TheOneRing, line 5: Add a url to this pr

changelog/unreleased/9513
```diff
@@ -0,0 +1,7 @@
...
+https://github.com/owncloud/client/issues/9425
```
TheOneRing, line 6: Maybe merge the entry with 9425?

TheOneRing (changes requested):
As discussed

erikjv:
sigh. something weird happened with this MR. Should be fine now.

src/gui/protocolwidget.cpp
```diff
@@ -94,18 +95,28 @@ ProtocolWidget::~ProtocolWidget()
...
     QTimer::singleShot(0, menu, [menu] {
         menu->popup(QCursor::pos());
     });
     return menu;
 }
 
+QString ProtocolWidget::filterButtonText(int filterCount)
+{
+    if (filterCount == 0) {
+        return tr("Filter");
```
TheOneRing, line 112: Hmm https://doc.qt.io/qt-5/i18n-source-translation.html#handling-plurals

Issue #9479, opened 2022-03-02
Show failed test in nightly builds


Talank:
After https://github.com/owncloud/client/pull/9459 we have proper provision to comment the link of test report, when the test fails and report exists. However, after we disabled the use of `gui test result parser`, debugging fails in nightly builds has been difficult. Hence there is a need for appropriate method to show failed tests result (or link to log) in nightly builds.

Possible solutions could be (but not limited to):
- In the notification of nightly builds, use a template which contains the link of test result (like that we show via GitHub comment in PRs)

saw-jan:
Implementation PR merged

PR #9541, opened 2022-03-23
[tests-only] Notify with logs whenever there is a nightly fail

saw-jan:
Uses the custom template for build notification

Fixes https://github.com/owncloud/client/issues/9479


saw-jan:
should be in 2.10 branch

PR #9551, opened 2022-03-24
Ensure the folder definition is initialised

TheOneRing:


fmoc (approved):
GitHub rendered `folderman.cpp`'s diff in a weird way. Looks good!

PR #9554, opened 2022-03-25
Draft: Add test for open-during-upload

erikjv:
https://github.com/owncloud/enterprise/issues/5052


erikjv:
From what I understand from the issue, this test should "do it". However, I tried it against a version of 2.10 before the fix went in, and it does _not_ fail... Any ideas?

TheOneRing (changes requested):
Does not fail with 2.10.0

PR #9565, opened 2022-04-05
Move FolderWizard to module

TheOneRing:


src/gui/folderwizard/folderwizard.cpp
```diff
@@ -13,90 +13,102 @@
...
-Q_LOGGING_CATEGORY(lcFolderWizard, "gui.folderwizard", QtInfoMsg)
-
-QString FormatWarningsWizardPage::formatWarnings(const QStringList &warnings, bool isError) const
+namespace {
+QString formatWarnings(const QStringList &warnings, bool isError = false)
 {
     QString ret;
     if (warnings.count() == 1) {
-        ret = isError ? tr("<b>Error:</b> %1").arg(warnings.first()) : tr("<b>Warning:</b> %1").arg(warnings.first());
+        ret = isError ? QApplication::translate("FolderWizard", "<b>Error:</b> %1").arg(warnings.first()) : QApplication::translate("FolderWizard", "<b>Warning:</b> %1").arg(warnings.first());
```
fmoc, line 54: Doesn't this change the context of these messages, i.e., invalidate old translations?

src/gui/folderwizard/folderwizard.h
```diff
@@ -15,17 +15,19 @@
...
 
-#include "folder.h"
 #include "accountfwd.h"
 
-#include "ui_folderwizardsourcepage.h"
-#include "ui_folderwizardtargetpage.h"
+#include "gui/folder.h"
 
 class QCheckBox;
+class QTreeWidgetItem;
```
fmoc, line 27: Do we need these forward declarations? I guess we could just `#include` these classes.

fmoc (approved):

PR #9606, opened 2022-04-26
Style wizard

fmoc:
Contributes to #9249.

Requires #9608.

Screenshots using testpilotcloud theme:

<details>

screenshot_2022-04-26_08-55-42

screenshot_2022-04-26_08-55-58

screenshot_2022-04-26_08-56-13

screenshot_2022-04-26_08-56-23
screenshot_2022-04-26_08-56-31

screenshot_2022-04-26_08-56-39

screenshot_2022-04-26_08-57-11

</details>

Screenshots using standard theme:

<details>

screenshot_2022-04-26_08-58-25
screenshot_2022-04-26_08-58-34
screenshot_2022-04-26_08-59-09

</details>

PR #9608, opened 2022-04-26
Use folder wizard for manual sync config

fmoc:
Extracted from #9606. Fixed the mistake mentioned in the reviews there, and further removed some now-obsolete code.

### DIFF
--- a/src/gui/newwizard/pages/accountconfiguredwizardpage.ui
+++ b/src/gui/newwizard/pages/accountconfiguredwizardpage.ui
@@ -146,12 +146,12 @@
              </widget>
             </item>
             <item>
-             <widget class="QRadioButton" name="selectiveSyncRadioButton">
+             <widget class="QRadioButton" name="configureSyncManuallyRadioButton">
               <property name="toolTip">
-               <string>You can configure which directories to download from the server in a dialog once the wizard has finished.</string>
+               <string>After completing this wizard, you can set up folder synchronization manually.</string>
               </property>
               <property name="text">
-               <string>Choose what to download</string>
+               <string>Configure synchronization manually</string>
               </property>
              </widget>
             </item>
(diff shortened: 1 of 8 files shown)


src/gui/newwizard/pages/accountconfiguredwizardpage.ui
```diff
@@ -13,6 +13,9 @@
   <property name="windowTitle">
    <string>Form</string>
   </property>
+  <property name="styleSheet">
+   <string notr="true">-</string>
```
TheOneRing, line 17: Does that mean, unstyled or whats the meaning.

src/gui/owncloudgui.cpp
```diff
@@ -51,61 +53,51 @@ namespace {
...
-                                addFolder(driveLocalFolder.absolutePath(), {}, QUrl::fromEncoded(d.getRoot().getWebDavUrl().toUtf8()), d.getName());
-                            }
-                            finalize();
-                        }
-                    }
-                });
+    // saving once after adding makes sure the account is stored in the config in a working state
+    // this is needed to ensure a consistent state in the config file upon unexpected terminations of the client
+    // (for instance, when running from a debugger and stopping the process from there)
+    AccountManager::instance()->save(true);
```
TheOneRing, line 61: I guess that should happen in runNewAccountWizard after the connection validator?

TheOneRing (changes requested):

src/gui/owncloudgui.cpp
fmoc: Perhaps it makes sense to save the account even before running the validator. What if client crashes in the validator step? The account might be stored in an inconsistent state then, too.

TheOneRing (approved):

PR #9621, opened 2022-05-02
Use HTTPS if no scheme has been provided in wizard

fmoc:


src/gui/newwizard/setupwizardcontroller.cpp
```diff
@@ -101,8 +101,10 @@ void SetupWizardController::nextStep(std::optional<PageIndex> currentPage, std::
 
                 // fix scheme if necessary
                 // using HTTPS as a default is a really good idea nowadays, users can still enter http:// explicitly if they wish to
-                if (QUrl(userProvidedUrl).isRelative() && !userProvidedUrl.isEmpty()) {
-                    userProvidedUrl.prepend(QStringLiteral("https://"));
+                if (!userProvidedUrl.startsWith(QStringLiteral("http://")) && !userProvidedUrl.startsWith(QStringLiteral("https://"))) {
+                    static const QString defaultScheme(QStringLiteral("https://"));
```
TheOneRing, line 105: Move to anonymous namespace.

Remove the qDebug/promise to implement a logging category in an upcoming pr.

What was wrong with https://github.com/owncloud/client/blob/ca7f788b1ca68ad3ea8062c3f3e537793df89b93/src/gui/wizard/owncloudsetuppage.cpp#L123

now you will convert `foo://bar` to `https://foo://bar`.
Does Qt detect `foo:8080` wrong?

src/gui/newwizard/setupwizardcontroller.cpp
```diff
@@ -33,6 +33,8 @@ QString initLocalFolder()
     return OCC::FolderMan::instance()->findGoodPathForNewSyncFolder(localFolder);
 }
 
+const QString defaultUrlScheme = QStringLiteral("https://");
```
TheOneRing, line 36: I'd move constants to the top of the anonymous namespace.
Historically the constants at least for strings where followed the schema `fooC`.
The C might have meant CString or constant, I don't care but we should come up with a standard for constants.

I'd suggest to either keep the C or to start Uppercase?

PR #9626, opened 2022-05-03
[Tests-Only]Removed GUI test retry option as it is errorsome in Squish

Talank:
Overall test result shows the test has failed even if there is any pass in retry situation. Hence, the `retry` function is not working as expected, and there is no point on using it because it raises more confusions. So, I removed that option here.


saw-jan (approved):
LGTM

PR #9635, opened 2022-05-04
Introduce per account default folder

TheOneRing:


src/gui/newwizard/setupwizardaccountbuilder.h
```diff
@@ -97,6 +97,8 @@ class SetupWizardAccountBuilder
      */
     bool hasValidCredentials() const;
 
+    QString userName() const;
```
fmoc, line 100: Is there a long-term plan to return the display name here? `userName` seems ambiguous since it currently returns the DAV user name, but is supposed to return the display name in the future.

src/libsync/account.h
```diff
@@ -94,6 +94,14 @@ class OWNCLOUDSYNC_EXPORT Account : public QObject
     QString davUser() const;
     void setDavUser(const QString &newDavUser);
 
+    /***
+     * With OC 10 this is the equivalent to the sync root.
+     * With ocis and spaces this will be the default folder containing all spaces.
+     */
+    QString defaultSyncRoot() const;
+    void setDefaultSyncRoot(const QString &syncRoot);
```
fmoc, line 102: Should we document that setting an empty string sets `hasDefaultSyncRoot` to false?

src/libsync/account.cpp
```diff
@@ -515,4 +515,26 @@ void Account::setServerVersion(const QString &version)
...
 }
 
+QString Account::defaultSyncRoot() const
+{
+    Q_ASSERT(!_defaultSyncRoot.isEmpty());
+    return _defaultSyncRoot;
+}
+bool Account::hasDefaultSyncRoot() const
+{
+    return !_defaultSyncRoot.isEmpty();
```
fmoc, line 525: What's the motivation of having this extra function? Right now, one could just check whether the returned value is empty, too. We've done this elsewhere, too. In case there is a reason why you designed it like this, I'd appreciate some comments.

src/libsync/account.cpp
```diff
@@ -515,4 +515,26 @@ void Account::setServerVersion(const QString &version)
     emit serverVersionChanged(this, oldServerVersion, version);
 }
 
+QString Account::defaultSyncRoot() const
+{
+    Q_ASSERT(!_defaultSyncRoot.isEmpty());
```
fmoc, line 520: Perhaps we should check whether the directory exists here, too. It might have been removed in the meantime.

src/gui/owncloudgui.cpp
```diff
@@ -70,17 +70,15 @@ void setUpInitialSyncFolder(AccountStatePtr accountStatePtr, const QString &loca
...
-        QObject::connect(drive, &OCC::GraphApi::Drives::finishedSignal, [accountStatePtr, localFolder, drive, addFolder, finalize] {
+        QObject::connect(drive, &OCC::GraphApi::Drives::finishedSignal, [accountStatePtr, drive, addFolder, finalize] {
             if (drive->parseError().error == QJsonParseError::NoError) {
                 const auto &drives = drive->drives();
                 if (!drives.isEmpty()) {
-                    const QDir localDir(localFolder);
-                    localDir.mkdir(".");
-                    FileSystem::setFolderMinimumPermissions(localFolder);
-                    Utility::setupFavLink(localFolder);
+                    const QDir localDir(accountStatePtr->account()->defaultSyncRoot());
```
fmoc, line 77: As far as I can see, there is no guarantee that the directory has not been removed in the meantime. Do we need to handle this case?

TheOneRing: We reach that code directly after the account was setup but well I won't hurt.

src/libsync/account.cpp
```diff
@@ -515,4 +515,25 @@ void Account::setServerVersion(const QString &version)
...
+    return _defaultSyncRoot;
+}
+bool Account::hasDefaultSyncRoot() const
+{
+    return !_defaultSyncRoot.isEmpty();
+}
+
+void Account::setDefaultSyncRoot(const QString &syncRoot)
+{
+    Q_ASSERT(_defaultSyncRoot.isEmpty());
```
fmoc, line 530: Maybe this should be logged as well using `OC_ASSERT`?

src/libsync/account.h
```diff
@@ -94,6 +94,24 @@ class OWNCLOUDSYNC_EXPORT Account : public QObject
...
+    QString defaultSyncRoot() const;
+
+    /***
+     * Whether we have defaultSyncRoot defined.
+     */
+    bool hasDefaultSyncRoot() const;
+
+    /***
+     * Set defaultSyncRoot and creates the path on the filesystem.
+     * Setting an empty string will have no effect.
```
fmoc, line 111: Now I'm confused. Is there a reason to ever call this with an empty string?

TheOneRing: Loading a legacy account.

TheOneRing: Loading an account without a defaultSyncRoot

src/libsync/account.cpp
TheOneRing: I think the user should take care.

src/libsync/account.cpp
TheOneRing: See the comment about the asser.

fmoc (approved):

PR #9642, opened 2022-05-05
Handle custom trusted CA certificates in our access manager

fmoc:
Preparation for an upcoming PR that implements a TLS error dialog in the wizard.

This PR extends the existing `AccessManager` subclass of `QNetworkAccessManager` with TLS certificates management. Unfortunately, the default Qt class does not support this.

This PR replaces all the hacks that altered the global state in order to make requests to servers using (previously untrusted) certificates work.

### DIFF
--- a/src/libsync/accessmanager.cpp
+++ b/src/libsync/accessmanager.cpp
@@ -78,9 +78,13 @@ QNetworkReply *AccessManager::createRequest(QNetworkAccessManager::Operation op,
         newRequest.setAttribute(QNetworkRequest::Http2AllowedAttribute, http2EnabledEnv);
     }
 
-    auto sslConfiguration = newRequest.sslConfiguration();
-    sslConfiguration.addCaCertificates({ _customTrustedCaCertificates.begin(), _customTrustedCaCertificates.end() });
-    newRequest.setSslConfiguration(sslConfiguration);
+    // for some reason, passing an empty list causes the default chain to be removed
+    // this behavior does not match the documentation
+    if (!_customTrustedCaCertificates.isEmpty()) {
+        auto sslConfiguration = newRequest.sslConfiguration();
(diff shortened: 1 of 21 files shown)


src/libsync/accessmanager.cpp
```diff
@@ -78,9 +78,41 @@ QNetworkReply *AccessManager::createRequest(QNetworkAccessManager::Operation op,
         newRequest.setAttribute(QNetworkRequest::Http2AllowedAttribute, http2EnabledEnv);
     }
 
+    auto sslConfiguration = newRequest.sslConfiguration();
```
TheOneRing, line 81: Move sslConfiguration into the if

src/libsync/accessmanager.cpp
```diff
@@ -78,9 +78,41 @@ QNetworkReply *AccessManager::createRequest(QNetworkAccessManager::Operation op,
...
+}
+
+void AccessManager::setCustomTrustedCaCertificates(const QSet<QSslCertificate> &certificates)
+{
+    _customTrustedCaCertificates = certificates;
+}
+
+void AccessManager::addCustomTrustedCaCertificates(const QList<QSslCertificate> &certificates)
+{
+    std::for_each(certificates.begin(), certificates.end(), [this](const QSslCertificate &certificate) {
```
TheOneRing, line 108: { _customTrustedCaCertificates.begin(), _customTrustedCaCertificates.end() } ?

TheOneRing (changes requested):

src/libsync/accessmanager.cpp
fmoc: I'm not sure what you mean.

(I tried `std::copy`, but the Qt containers don't work with STL algorithms.)

TheOneRing:
Looks great, lets find out how it will blow up.

PR #9646, opened 2022-05-06
Improve spaces handling

TheOneRing:


src/common/utility_win.cpp
```diff
@@ -55,29 +55,8 @@ const QString systemThemesC()
 
 namespace OCC {
 
-void Utility::setupFavLink(const QString &folder)
+void Utility::setupFavLink(const QString &)
```
fmoc, line 58: Perhaps we should mark this as `deprecated`? I mean the Linux version has been broken for several years as well (not saying that that one should necessarily be discarded, but so far it just didn't do anything anyway).

src/gui/folderman.h
```diff
@@ -85,6 +85,7 @@ class FolderMan : public QObject
     Q_OBJECT
 public:
     static QString suggestSyncFolder(const QUrl &server, const QString &displayName);
+    static bool prepareFolder(const QString &folder);
```
fmoc, line 88: Do you ever intend to handle errors externally? Maybe `void` with some logging is more suitable?

fmoc (approved):

PR #9652, opened 2022-05-09
Make sure at least Vfs::Off plugin is built

fmoc:
Helps during development when just building the main owncloud binary target.


fmoc:
Will be replaced by #9639.

PR #9657, opened 2022-05-10
Remove Account:sslConfiguration

TheOneRing:


src/libsync/accessmanager.cpp
```diff
@@ -80,11 +80,15 @@ QNetworkReply *AccessManager::createRequest(QNetworkAccessManager::Operation op,
 
     // for some reason, passing an empty list causes the default chain to be removed
     // this behavior does not match the documentation
```
fmoc, line 82: This comment only makes sense right above the `if`.

fmoc (changes requested):
Great improvement. Just one minor issue with a comment.

src/libsync/accessmanager.cpp
TheOneRing: fixed in follow up

PR #9703, opened 2022-05-19
Add build job with libcloudproviders enabled

### LINKED ISSUE #9594: Test libcloudproviders integration using GitHub actions/craft
As seen recently (05b2dddf4e02e19bb2d80ccc197399fc70c17b90), we currently do not perform any integration testing with libcloudproviders enabled. We should update our GitHub actions workflow to build both with and without libcloudproviders (i.e., setting `WITH_LIBCLOUDPROVIDERS` explicitly to `ON` and `OFF` respectively, otherwise CMake would enable libcloudproviders automatically if it can be found).

fmoc:
Fixes #9594.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -74,8 +74,6 @@ jobs:
         if ($IsWindows) {
             & "${env:GITHUB_WORKSPACE}/.github/workflows/.craft.ps1" -c dev-utils/nsis
         } elseif($IsLinux) {
-            sudo apt-get update -y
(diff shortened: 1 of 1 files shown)


TheOneRing:
I'm against adding multiple builds per platform, sure you can argue that we should test all possible combinations of platforms and configurations, but in the end its just costing energy.
And to build the whole stuff again for a difference of 10 line of code?....

fmoc:
I'd then forcibly require libcloudproviders. Untested configurations are the root of all evil.

TheOneRing:
We could enable it by default in https://github.com/owncloud/client/blob/master/.craft.ini#L68 or in the blueprint.

fmoc:
... or maybe prevent builds without it in CMake. Anyway, this PR is ready. Please don't forget to update the branch protection.

Issue #9691, opened 2022-05-18
[QA] 3.0.0 alpa1 client: adding an account takes long time


gabi18:
Seen with ownCloud-3.0.0-alpha1.7623-vfs.x64.GPO.msi on Windows10 21H2 VirtualBox VM.

Adding a new account to demo.owncloud.org takes more than 30 seconds (the spinning icon is shown).

![add-account](https://user-images.githubusercontent.com/75983861/169022790-4f97779a-13b1-4fd7-bcc8-31338cfd88f9.png)

```
05-18 11:52:36:523 [ debug sync.networkjob ]    [ OCC::AbstractNetworkJob::slotFinished ]:      Network job finished OCC::AvatarJob("https://demo.owncloud.org/remote.php/dav/avatars/demo/128.png","GET", "786582ea-0e2a-4558-97c5-fce9dceb2615", "786582ea-0e2a-4558-97c5-fce9dceb2615", "Error transferring https://demo.owncloud.org/remote.php/dav/avatars/demo/128.png - server replied: Not Found")
05-18 11:53:06:187 [ info sync.networkjob ]:    Created OCC::PropfindJob("https://demo.owncloud.org/remote.php/dav/files/demo/","PROPFIND", "a0a1c056-3ba6-4c40-9452-ed8fcb7b2406", "a0a1c056-3ba6-4c40-9452-ed8fcb7b2406") for OCC::QuotaInfo(0x273f71652f8)
```
[connect-pause.log.gz](https://github.com/owncloud/client/files/8715598/connect-pause.log.gz)

TheOneRing:
If you think an issue is not fixed please reopen it...

gabi18:
I cannot reproduce the issue yet. Adding the first account to the client usually is fast (while it seems that accounts on oCIS take longer).
When adding more accounts it gets slower. The logging below shows a strange behaviour: I have added an account connecting to ocis.owncloud.com, the time gap is for demo.owncloud.com.

I will retest if basic auth/oauth makes a difference.

```
05-23 14:08:54:254 [ info gui.folder.manager ]: <========== Sync finished for folder [ "ownCloud - demo@demo.owncloud.com2" ] of account [ "demo@demo.owncloud.com" ] with remote [ "https://demo.owncloud.com/remote.php/dav/files/demo/" ]

[...]

PR #9727, opened 2022-05-24
Fix scheduling of newly added folders

TheOneRing:
Fixes: #9691


src/gui/folderman.cpp
```diff
@@ -1495,4 +1497,12 @@ bool FolderMan::prepareFolder(const QString &folder)
     return true;
 }
 
+void OCC::FolderMan::reenableSyncAfterFolderAdded()
+{
+    // The vfs folder might be set up async, schedule it after enough time to setup
```
fmoc, line 1502: A timer is not really a guarantee... is there any way we can ensure that this is being executed in the right order?

fmoc (approved):

PR #9730, opened 2022-05-24
Use ocis productversion

TheOneRing:


src/gui/application.cpp
```diff
@@ -396,21 +396,8 @@ Application::~Application()
     AccountManager::instance()->shutdown();
 }
 
-void Application::slotAccountStateRemoved(AccountStatePtr accountState) const
+void Application::slotAccountStateRemoved() const
 {
-    if (_gui) {
```
TheOneRing, line 401: @fmoc @erikjv @dschmidt @dragotin as far as I can tell we don't need to disconnect those as once the emitting object is gone they can no longer be triggered....

src/gui/application.cpp
```diff
@@ -425,13 +412,17 @@ void Application::slotAccountStateAdded(AccountStatePtr accountState) const
     connect(accountState.data(), &AccountState::stateChanged,
         _gui.data(), &ownCloudGui::slotAccountStateChanged);
     connect(accountState->account().data(), &Account::serverVersionChanged,
-        _gui.data(), &ownCloudGui::slotTrayMessageIfServerUnsupported);
+        _gui.data(), [account = accountState->account().data(), this] {
+            _gui->slotTrayMessageIfServerUnsupported(account);
+        });
```
dragotin, line 417: I do not get why this implementation is better than the old one.

dragotin (approved):

src/gui/application.cpp
TheOneRing: I had to touch this connect as I changed the signature of the signal.
the old emit serverVersionChanged(this).
It is highly uncommon for Qt signal to pass this with the emit.

fmoc (changes requested):

src/cmd/cmd.cpp
```diff
@@ -560,8 +560,10 @@ int main(int argc, char **argv)
                 auto caps = capabilitiesJob->data().value("ocs").toObject().value("data").toObject().value("capabilities").toObject();
                 qDebug() << "Server capabilities" << caps;
                 ctx.account->setCapabilities(caps.toVariantMap());
-                ctx.account->setServerInfo(caps["core"].toObject()["status"].toObject()["version"].toString(),
-                    caps["core"].toObject()["status"].toObject()["productname"].toString());
+
+                if (ctx.account->serverVersionUnsupported()) {
+                    qFatal("Error unsupported server");
```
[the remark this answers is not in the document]
TheOneRing: Hmm lets not support unknown servers with the not well maintained cmd client.

fmoc (approved):

PR #9732, opened 2022-05-24
Change the way the continous log files are created

TheOneRing:
We now only rotate the logs when the client is started or the log gets bigger than 100mb

An inititial sync with cloud.owncloud.com creates archives of 4-7mb .
Keeping only 5 log files is however not enough for servers like that.

### DIFF
--- a/changelog/unreleased/9731
+++ b/changelog/unreleased/9731
@@ -1,0 +1,9 @@
+Enhancement: Create continuus log files
+
+Previously, when logging was enabled, we started a new log file for every sync.
+This worked quite well if you sync a single account and a single folder.
+With spaces however we have a multitude of sync folders, which resulted in hundreds of tiny log files.
+
+Now, as soon as a log file's size exceeds 100 MiB, a new log file is started, and the old one is moved and compressed.
+
+https://github.com/owncloud/client/issues/9731
--- a/changelog/unreleased/9731
+++ b/changelog/unreleased/9731
@@ -1,9 +1,10 @@
-Enhancement: Create continuus log files
+Enhancement: Create continuous log files
 
 Previously, when logging was enabled, we started a new log file for every sync.
 This worked quite well if you sync a single account and a single folder.
 With spaces however we have a multitude of sync folders, which resulted in hundreds of tiny log files.
 
 Now, as soon as a log file's size exceeds 100 MiB, a new log file is started, and the old one is moved and compressed.
+The option to delete log files older than 4h was replaced by an option to keep a number of log files.
 
 https://github.com/owncloud/client/issues/9731
--- a/src/libsync/logger.cpp
+++ b/src/libsync/logger.cpp
@@ -13,14 +13,14 @@
  */
 
 #include "logger.h"
-
 #include "config.h"
+#include "theme.h"
 
 #include <QCoreApplication>
 #include <QDir>
 #include <QStringList>
+#include <QtConcurrent>
 #include <QtGlobal>
-#include <qmetaobject.h>
 
 #include <iostream>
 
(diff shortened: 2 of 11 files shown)


ownclouders:
Results for <strong>GUI-tests</strong> https://drone.owncloud.com/owncloud/client/11902/6/1

<details>


</details>

src/libsync/logger.cpp
```diff
@@ -272,44 +282,49 @@ void Logger::enterNextLogFile()
...
+        // set the creation time to now
+        _logFile.setFileTime(now, QFileDevice::FileTime::FileBirthTime);
+
+        QtConcurrent::run([now, previousLog, dir, logExpire = _logExpire] {
+            // Expire old log files and deal with conflicts
+            const auto &files = dir.entryList(QStringList(QStringLiteral("*" APPLICATION_SHORTNAME "-*.log.gz")),
+                QDir::Files, QDir::Name);
+            for (const auto &s : files) {
+                if (logExpire.count() > 0) {
+                    std::chrono::seconds expireSeconds(logExpire);
```
TheOneRing, line 311: Now that the amount of log files is quite low, deleting everything older than 4h feels dangerous, we might only have one or two in that time range.

I'd rather go for the more common approach of keeping up to N logs?

changelog/unreleased/9731
```diff
@@ -0,0 +1,9 @@
+Enhancement: Create continuus log files
+
+Previously, when logging was enabled, we started a new log file for every sync.
+This worked quite well if you sync a single account and a single folder.
+With spaces however we have a multitude of sync folders, which resulted in hundreds of tiny log files.
+
+Now new log files are started once a file becomes bigger than 100mb, the file is then move and compressed.
```
fmoc, line 7:
```suggestion
Now, as soon as a log file's size exceeds 100 MiB, a new log file is started, and the old one is moved and compressed.
```

src/libsync/logger.cpp
```diff
@@ -33,7 +33,8 @@
 #endif
 
 namespace {
-constexpr int CrashLogSize = 20;
+constexpr int crashLogSizeC = 20;
+constexpr int maxLogSizeC = 1024 * 1024 * 100; // 100mb
```
fmoc, line 37:
```suggestion
constexpr int maxLogSizeC = 1024 * 1024 * 100; // 100 MiB
```

src/libsync/logger.cpp
```diff
@@ -117,7 +118,33 @@ void Logger::doLog(QtMsgType type, const QMessageLogContext &ctx, const QString
...
+{
+    bool openSucceeded = false;
+    if (name == QLatin1Char('-')) {
+        attacheToConsole();
+        setLogFlush(true);
+        openSucceeded = _logFile.open(stdout, QIODevice::WriteOnly);
+    } else {
+        _logFile.setFileName(name);
+        openSucceeded = _logFile.open(QIODevice::WriteOnly);
+    }
```
fmoc, line 139: Might be easier to just move that into a lambda (like it is done with a lot of variable initializations).

src/libsync/logger.cpp
fmoc: Sounds like a good plan. However, 100 MiB of logs is a lot of data. Maybe lower the number to 10 MiB and keep more files? Concatenating them is trivial.

fmoc (approved):
Code looks good, so I'll approve this PR. But we should discuss the file size and log retention.

fmoc (approved):

fmoc (changes requested):
Wrong button. Please see my comments.

fmoc (approved):
Looks good!

Issue #9733, opened 2022-05-25
Skip all failing GUI tests in master until they are fixed to comply with new changes


Talank:
Skip the following test suites if they do not pass. Fixing all test suites in one PR will be impossible so in each upcoming work to fix the tests, unskip only the passing tests

- [ ] tst_addAccount
- [ ] tst_checkAlltabs
- [ ] tst_deletFilesFolders
- [ ] tst_editFiles
- [ ] tst_loginLogout
- [ ] tst_removeAccountConnection
- [ ] tst_sharing
- [ ] tst_syncing
- [ ] tst_vfs

Talank:
All tests were skipped and almost all tests except some in `sts_syncing` has been unskipped.

PR #9734, opened 2022-05-25
[Tests-Only]Skipped all failing GUI tests

Talank:
### Description
Currently, most of the tests are failing either because of the changed GUI components or because they have not been maintained for a long time. So all tests has been skipped, which will be shall be unskipped in upcoming PRs when they are fixed.

### Issue
- fixes https://github.com/owncloud/client/issues/9733

### DIFF
--- a/test/gui/tst_addAccount/test.feature
+++ b/test/gui/tst_addAccount/test.feature
@@ -1,3 +1,4 @@
+@skip @issue-9733
 Feature: adding accounts
 
     As a user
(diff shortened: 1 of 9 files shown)


test/gui/tst_addAccount/test.feature
```diff
@@ -1,3 +1,4 @@
+@skip @issue-9733
```
saw-jan, line 1: So we abandon the 2.10 branch from now?

individual-it (approved):

test/gui/tst_addAccount/test.feature
Talank: We do not abandon it completely. It is of less priority from now on.

PR #9742, opened 2022-05-25
Ask user to reauthenticate against OAuth2 server using modal dialog

### LINKED ISSUE #9559: Add "Copy to clipboard" action to "Re-open Browser" button on account screen
CleanShot 2022-04-04 at 09 27 03 | CleanShot 2022-04-04 at 09 28 30
-- | --

Could it be added there as well?

fmoc:
Fixes #9559.

screenshot_2022-07-04_22-28-49

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -39,22 +39,24 @@
 
 #include <math.h>
 
+#include <QAction>
+#include <QClipboard>
 #include <QDesktopServices>
 #include <QDir>
+#include <QIcon>
+#include <QKeySequence>
 #include <QListWidgetItem>
 #include <QMessageBox>
-#include <QAction>
-#include <QVBoxLayout>
-#include <QTreeView>
-#include <QKeySequence>
-#include <QIcon>
-#include <QVariant>
 #include <QToolTip>
-#include <qstringlistmodel.h>
+#include <QTreeView>
+#include <QVBoxLayout>
+#include <QVariant>
 #include <qpropertyanimation.h>
+#include <qstringlistmodel.h>
 
 #include "account.h"
 #include "askexperimentalvirtualfilesfeaturemessagebox.h"
+#include "askforoauthlogindialog.h"
 
 namespace OCC {
 
@@ -186,11 +188,6 @@ AccountSettings::AccountSettings(AccountStatePtr accountState, QWidget *parent)
 
     connect(&_quotaInfo, &QuotaInfo::quotaUpdated,
         this, &AccountSettings::slotUpdateQuota);
-
-    ui->openBrowserButton->setVisible(false);
-    connect(ui->openBrowserButton, &QToolButton::clicked, this, [this]{
-        qobject_cast<HttpCredentialsGui *>(_accountState->account()->credentials())->openBrowser();
-    });
 }
 
 
@@ -797,7 +794,9 @@ void AccountSettings::slotAccountStateChanged()
                 errors << tr("The server version %1 is unsupported! Proceed at your own risk.").arg(account->capabilities().status().versionString());
             }
             showConnectionLabel(tr("Connected to %1.").arg(serverWithUser), errors);
-            ui->openBrowserButton->setVisible(false);
+            if (_askForOAuthLoginDialog != nullptr) {
+                _askForOAuthLoginDialog->accept();
+            }
             break;
         }
         case AccountState::ServiceUnavailable:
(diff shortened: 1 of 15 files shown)


TheOneRing:
image

Please fix, we also discussed to display a modal dialog requesting the users attention and explaining the workflow.

src/gui/accountsettings.cpp
```diff
@@ -797,7 +794,9 @@ void AccountSettings::slotAccountStateChanged()
                 errors << tr("The server version %1 is unsupported! Proceed at your own risk.").arg(account->capabilities().status().versionString());
             }
             showConnectionLabel(tr("Connected to %1.").arg(serverWithUser), errors);
-            ui->openBrowserButton->setVisible(false);
+            if (_askForOAuthLoginDialog != nullptr) {
+                _askForOAuthLoginDialog->accept();
```
TheOneRing, line 798: Why accept?

src/gui/accountsettings.cpp
```diff
@@ -812,10 +811,33 @@ void AccountSettings::slotAccountStateChanged()
...
+                qCDebug(lcAccountSettings) << "showing modal dialog asking user to log in again via OAuth2";
+
+                _askForOAuthLoginDialog = new AskForOAuthLoginDialog(_accountState->account(), this);
+
+                // make sure to clean up the memory and to null the QPointer once finished
+                _askForOAuthLoginDialog->setAttribute(Qt::WA_DeleteOnClose);
+
+                connect(
+                    cred, &HttpCredentialsGui::authorisationLinkChanged,
+                    this, [this]() {
```
TheOneRing, line 828: Lambda not needed

TheOneRing (changes requested):

TheOneRing:
The open browser button also doesn't work

src/gui/accountsettings.cpp
fmoc: There's no particular reason. We can also just `hide()` or `close()` it if you like to. In my opinion, this makes sense to some extent, as logging in in the browser is similar to clicking a submit button in a basic auth form.

TheOneRing:
Clicking on reopen has an effect however.
I guess you need to start the oauth flow not only create it.
Also take care of error handling, what happens if the server is not available.

fmoc:
Cannot reproduce your reported issues. Please describe how you got there (e.g., logging out explicitly, restarting the app, etc.).

fmoc:
I'll try the scenario "server disappears". This kind of reminds me of fixing the mess that is `authorisationLink[Async]`...

src/gui/accountsettings.cpp
TheOneRing: Accept and close will trigger the corresponding signals to be emitted.
Anyway you should add a comment that we are hiding the dialog here.
Maybe deleteLater would be correct I'm not exactly sure.

src/gui/accountsettings.cpp
```diff
@@ -812,10 +811,31 @@ void AccountSettings::slotAccountStateChanged()
...
+                }
+
+                qCDebug(lcAccountSettings) << "showing modal dialog asking user to log in again via OAuth2";
+
+                _askForOAuthLoginDialog = new AskForOAuthLoginDialog(_accountState->account(), this);
+
+                // make sure to clean up the memory and to null the QPointer once finished
+                _askForOAuthLoginDialog->setAttribute(Qt::WA_DeleteOnClose);
+
+                connect(
```
TheOneRing, line 826: Formated a bit weird but probably due to clang format

src/gui/askforoauthlogindialog.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+namespace OCC {
+
+AskForOAuthLoginDialog::AskForOAuthLoginDialog(AccountPtr accountPtr, QWidget *parent)
+    : QDialog(parent)
+    , _ui(new ::Ui::AskForOAuthLoginDialog)
+{
+    _ui->setupUi(this);
+
+    _ui->label->setText(tr("The account %1 has been logged out by the server.\n\nPlease use your browser to log into %2.").arg(accountPtr->displayName(), Theme::instance()->appNameGUI()));
```
TheOneRing, line 31: To "login to"

src/gui/askforoauthlogindialog.cpp
```diff
@@ -35,6 +35,15 @@ AskForOAuthLoginDialog::AskForOAuthLoginDialog(AccountPtr accountPtr, QWidget *p
 
     setFixedSize(this->sizeHint());
 
+    auto creds = qobject_cast<HttpCredentialsGui *>(accountPtr->credentials());
```
TheOneRing, line 38: move down to its usage

src/gui/askforoauthlogindialog.cpp
```diff
@@ -35,6 +35,15 @@ AskForOAuthLoginDialog::AskForOAuthLoginDialog(AccountPtr accountPtr, QWidget *p
...
+
+    _ui->copyUrlToClipboardButton->setEnabled(false);
```
TheOneRing, line 40: Maybe disable in the ui file?

src/libsync/creds/oauth.cpp
```diff
@@ -497,16 +497,9 @@ QUrl OAuth::authorisationLink() const
...
-}
 
-void OAuth::authorisationLinkAsync(std::function<void (const QUrl &)> callback) const
-{
-    if (_wellKnownFinished) {
-        callback(authorisationLink());
-    } else {
-        connect(this, &OAuth::authorisationLinkChanged, callback);
-    }
+    Q_ASSERT(!url.isEmpty());
```
TheOneRing, line 501: How would that be possible?

TheOneRing: We rather need to check for _wellKnownFinished ?

src/gui/askforoauthlogindialog.cpp
fmoc: I totally forgot about an assertion I wanted to add there. Will fix that.

src/libsync/creds/oauth.cpp
fmoc: I did want to have some assertion in there at least. We can increase the checks' granularity of course.

src/gui/accountsettings.cpp
TheOneRing: if hide deletes the dialog :D

src/gui/askforoauthlogindialog.cpp
fmoc: I used `log into` everywhere else. I assume we have to adapt those occurrences as well then.

TheOneRing: Hmm yes  but we should discuss it in chat maybe we can find a native speaker

PR #9755, opened 2022-06-01
Add link to app password configuration 

### LINKED ISSUE #9754: Inform about app passwords in new wizard
See https://github.com/owncloud/client/issues/9249#issuecomment-1142256146 for context.

We need to:

- [ ] inform users that an app password may be used instead of their main password
- [ ] provide a direct link to the server

@TheOneRing can we just show these for every basic auth workflow, or do we have to make those conditional (e.g., enable only when a specific capability is available)?

fmoc:
Fixes #9754.

The first commit fixes a minor issue introduced in #9746.

screenshot_2022-06-01_10-03-37


tbsbdr:
1st approach is better. I can provide more nitty-gritty refinements next week.

dragotin:
Hm, I fear people mix up with the "Forgot my password" link and click-without-reading. 

Also - users will not know what an "app password" exactly is. Any link to help?

TheOneRing:
Sadly branded clients don't provide doc and until now we don't have a solution for that.

fmoc:
Just had a quick conversation with @tbsbdr discussing this problem.

TL;DR: the text should be moved above the username/password form, below the "Enter your credentials" label.

Some background: the old wizard placed it right above the form as well:

screenshot_2022-06-15_16-18-07

There are two good reasons to move the app password link label into the proposed position:

a) existing users are used to finding the link above the form
b) typically, "forgot password" links are just below the password line edit on most websites and in many apps (@dragotin also suggested this in https://github.com/owncloud/client/pull/9755#issuecomment-1145668157)

screenshot_2022-06-15_16-22-50

fmoc:
I'll push my original proposal for now so it can be tested. I'd like to hear @TheOneRing and @dragotin's opinion.

src/gui/newwizard/pages/basiccredentialssetupwizardpage.cpp
```diff
@@ -24,30 +24,37 @@ BasicCredentialsSetupWizardPage::BasicCredentialsSetupWizardPage(const QUrl &ser
...
-    case Theme::UserIDEmail:
-        _ui->usernameLabel->setText(tr("E-mail address"));
-        break;
-    case Theme::UserIDCustom:
-        _ui->usernameLabel->setText(Theme::instance()->customUserID());
-        break;
-    default:
-        Q_UNREACHABLE();
-    }
+    const QString usernameLabelText = []() {
```
TheOneRing, line 36: enumToStirng

fmoc: Also, how to handle translations?

TheOneRing: yes in the theme.
QCoreApplication::translate("enumName","foo")

src/libsync/theme.cpp
```diff
@@ -686,5 +686,19 @@ bool Theme::allowDuplicatedFolderSyncPair() const
     return true;
 }
 
+template <>
+OWNCLOUDSYNC_EXPORT QString Utility::enumToDisplayName(Theme::UserIDType userIdType)
+{
+    switch (userIdType) {
+    case Theme::UserIDUserName:
+        return QCoreApplication::tr("Username");
+    case Theme::UserIDEmail:
```
TheOneRing, line 695: https://doc.qt.io/qt-5/qcoreapplication.html#translate

TheOneRing (changes requested):

PR #9760, opened 2022-06-01
Allow to use our logger code in external libraries

TheOneRing:
I removed the install target for probably not working headers and added the logger.h file instead.
Ideally we would also provide a cmake export target but at the moment I would not recommend anybody to use the logger besides for testing purposes.

### DIFF
--- a/src/libsync/CMakeLists.txt
+++ b/src/libsync/CMakeLists.txt
@@ -69,30 +69,14 @@ endif()
 
 # These headers are installed for libowncloudsync to be used by 3rd party apps
 set(owncloudsync_HEADERS
-    account.h
-    syncengine.h
-    configfile.h
-    networkjobs.h
-    progressdispatcher.h
-    syncfileitem.h
-    syncresult.h
+    ${CMAKE_CURRENT_BINARY_DIR}/owncloudlib.h
+    logger.h
 )
 
-set(creds_HEADERS
-    creds/abstractcredentials.h
-    creds/httpcredentials.h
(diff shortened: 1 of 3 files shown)


ownclouders:
Results for <strong>GUI-tests</strong> https://drone.owncloud.com/owncloud/client/11978/6/1

<details>


</details>

src/libsync/CMakeLists.txt
```diff
@@ -69,31 +69,15 @@ endif()
...
+    ${CMAKE_CURRENT_BINARY_DIR}/owncloudlib.h
+    logger.h
 )
 
-set(creds_HEADERS
-    creds/abstractcredentials.h
-    creds/httpcredentials.h
+INSTALL(
+    FILES ${owncloudsync_HEADERS}
+    DESTINATION ${INCLUDE_INSTALL_DIR}/${synclib_NAME}
```
fmoc, line 78: You should probably add these to the target, then use `install(TARGETS owncloudsync [PUBLIC|PRIVATE]_HEADER DESTINATION <...>)`.

fmoc (approved):
Added a suggestion, but I think since this PR just moves some code around, there's no need to perform this refactoring right now. We use plenty legacy APIs in CMake anyway.

src/libsync/CMakeLists.txt
TheOneRing: As far as I understand PUBLIC_HEADER is a mac os framework thingy.

Issue #9763, opened 2022-06-02
fix `notification` pipeline that is failing in nightly build


sushmita56:
The `notification` pipeline in `master` is failing for the last two days, that's why notification of nightly build is not popping up in our `desktop internal` channel. So we have to fix it.

```
latest: Pulling from plugins/slack
Digest: sha256:57fb90fd174908c0f5be58fd11b5bf1c420807c64a934c4a346a9b257b6495ba
Status: Downloaded newer image for plugins/slack:latest
2022/06/02 01:22:39 {"success":false,"error":"error-running-script"}
```

sushmita56:
The image `plugins/slack` was last updated 4 months ago so we can say there is not any problem with the image. The problem while running the script is because of the `template` settings of the image. The `notifications-build` pipeline works fine after removing the `template`.

Talank:
This issue was first observed in https://drone.owncloud.com/owncloud/client/11970

Talank:
The notification passed today https://drone.owncloud.com/owncloud/client/12040/4/3

sushmita56:
The `notification` pipeline in CI is passing for the last seven days. So, I think this PR can be closed for now. It will be reopened if the same issue appears again.

PR #9766, opened 2022-06-02
check notification

sushmita56:
Related issue: 
#9763


sushmita56:
reference: https://github.com/owncloud/client/issues/9763#issuecomment-1153511012

PR #9771, opened 2022-06-03
[Tests-Only]Unskipped sharing test with example table

Talank:
The test is passing without any change. https://drone.owncloud.com/owncloud/client/12018/6/14

However, while running locally, squish is not being able to properly parse the example table.


saw-jan (approved):
Noice!

PR #9779, opened 2022-06-07
Split up controller logic into multiple classes

fmoc:

PR #9801, opened 2022-06-15
Improve wizard logging

### LINKED ISSUE #9664: Use QLoggingCategory in new Wizard

fmoc:
Fixes #9664.

I'd like to extend the logging further, but this makes sense only after merging #9779.

### DIFF
--- a/src/gui/newwizard/logging.cpp
+++ b/src/gui/newwizard/logging.cpp
@@ -1,3 +1,0 @@
-#include "logging.h"
-
-Q_LOGGING_CATEGORY(lcWizard, "wizard")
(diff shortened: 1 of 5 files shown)


src/gui/newwizard/logging.cpp
```diff
@@ -0,0 +1,3 @@
+#include "logging.h"
```
TheOneRing, line 1: I guess we should have more than one category and thus we won't need the logging.h/cpp

TheOneRing: As the category is only used in the .cpp then we won't need to make them available in the header with Q_DECLARE_LOGGING_CATEGORY

TheOneRing: We could also look on https://api.kde.org/ecm/module/ECMQtDeclareLoggingCategory.html

fmoc: We can use more specific categories.

TheOneRing (approved):

PR #9804, opened 2022-06-17
Provide cmake install targets and install filesystembase.h

TheOneRing:


src/csync/CMakeLists.txt
```diff
@@ -25,10 +25,13 @@ configure_file(csync_version.h.in ${CMAKE_CURRENT_BINARY_DIR}/csync_version.h)
...
 add_library(csync SHARED ${common_SOURCES} ${csync_SRCS})
 
+add_library(ownCloud::CSync ALIAS csync)
+set_target_properties(csync PROPERTIES EXPORT_NAME CSync)
+
 target_include_directories(csync
     PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/std
     # include root source dir in order to make src/common available to the targets
-    PUBLIC ${PROJECT_SOURCE_DIR}/src
+    PUBLIC $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src>
```
fmoc, line 34: I think you should include a line like `$<INSTALL_INTERFACE:include>`, too.

src/csync/CMakeLists.txt
```diff
@@ -91,7 +96,16 @@ set_target_properties(
       SOVERSION
         "0"
 )
-INSTALL(TARGETS csync ${KDE_INSTALL_TARGETS_DEFAULT_ARGS})
+
+INSTALL(
+    FILES
+    ${CMAKE_CURRENT_BINARY_DIR}/ocsynclib.h
```
fmoc, line 102: Indentation is off.

src/csync/CMakeLists.txt
```diff
@@ -91,7 +96,16 @@ set_target_properties(
...
+        ../common/filesystembase.h
+        ../common/utility.h
+    DESTINATION ${INCLUDE_INSTALL_DIR}/ownCloud/common
```
fmoc, line 105: Code style wise, the argument should be moved to a new line.

src/libsync/CMakeLists.txt
```diff
@@ -68,22 +68,26 @@ else()
...
+    DESTINATION ${INCLUDE_INSTALL_DIR}/ownCloud/libsync
 )
 
 add_library(libsync SHARED ${libsync_SRCS})
-
-target_link_libraries(libsync PUBLIC
-    csync
-    Qt5::Core Qt5::Network
+add_library(ownCloud::SyncCore ALIAS libsync)
+set_target_properties(libsync PROPERTIES EXPORT_NAME SyncCore)
```
fmoc, line 81: Perhaps just rename `libsync` entirely?

src/libsync/CMakeLists.txt
```diff
@@ -96,7 +100,7 @@ target_compile_definitions(libsync
             QT_NO_FOREACH
 )
 
-target_link_libraries(libsync PUBLIC OpenAPI::LibreGraphAPI)
+target_link_libraries(libsync PUBLIC $<BUILD_INTERFACE:OpenAPI::LibreGraphAPI>)
```
fmoc, line 103: See above, you probably should add an install interface include.

fmoc (changes requested):

src/libsync/CMakeLists.txt
TheOneRing: We can now use ownCloud::SyncCore in our code

src/libsync/CMakeLists.txt
TheOneRing: Its not ready for use yet, and the 3 headers  we install don't depend on it.

TheOneRing: Might be a bit hacky, its basically a PRIVATE lib and I was to lazy to explicitly link the ui to it..

src/libsync/CMakeLists.txt
fmoc: What I mean is, we probably should just rename directories etc. in our code base, too, and use one name everywhere.

fmoc: If this is the plan, a TODO should be left above this code.

fmoc (commented):
Let's just consider this a "first start" then.

fmoc (approved):

Issue #9082, opened 2021-09-23
Add cmake option to disable auto updater


TheOneRing:
Distros might not want to have our auto updater enabled, add option to disable it.

bermeitinger-b:
In the meantime, you can set the environment variable `OCC_UPDATE_URL=.` (or anything else) to disable the update check.

PR #9814, opened 2022-06-22
CMake: Implement WITH_AUTO_UPDATER

TheOneRing:
Fixes: #9082

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -104,6 +104,9 @@ option(BUILD_LIBRARIES_ONLY "BUILD_LIBRARIES_ONLY" OFF)
 # build the GUI component, when disabled only owncloudcmd is built
 option(BUILD_GUI "BUILD_GUI" ON)
 
+# build the auto updater component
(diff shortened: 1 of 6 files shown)


fmoc (approved):

dragotin:
This needs to be documented in the client docs, there is a paragraph about building the client. @mmattel 

Are there, apart from this switch, ohters that should be documented there?

TheOneRing:
They are already kind of documented in code.
https://github.com/owncloud/client/blob/3ae537b8c4dcfc9c935799baf2f6d855f679bddd/CMakeLists.txt#L80-L105

dragotin:
> They are already kind of documented in code.

yes, and with a cmake-guru-level above 8 people will find that ;-)

TheOneRing:
Those settings are available in the cmake gui and in most IDE's

TheOneRing:
I don't really have a problem with it, its just an other part of the documentation that can be outdated quite fast

Issue #8839, opened 2021-07-23
Move resources bundled with qrc into shared library


fmoc:
Qt's resources system bundles files directly inside binaries. While this makes handling resources more reliable, it also adds bloat when the same files are embedded in different binaries.

Qt itself provides a way to bundle resources into separate files and then load these during runtime. However, there is no easy built-in way to load these. If we can bundle the resources into a shared library instead, the linker will take care of loading the resources into memory (bundling shared libraries is a problem we solved already). The method is described in https://falsinsoft.blogspot.com/2018/02/qt-load-resources-from-static-or-shared.html.

Since we are in the process of moving resources previously deployed as files into qrc (see #8822), we need to think about the bloat. We should consider implementing this shared library approach.

PR #9813, opened 2022-06-22
Improve oauth error handling

TheOneRing:
Depends on: https://github.com/owncloud/client/pull/9815
With the ownCloud 10 app we used to be redirected to a theme "you can now close the browser window" page.
With oidc we will display our own page which looks kind of unprofessional.
This change starts theming the web page we display.

Old:
image

New:
image

@fmoc thx for the css :D

PR #9815, opened 2022-06-22
Move resources to a shared libary

TheOneRing:
This allows us to use resources in the cmd app and the unit tests.

This became relevant with https://github.com/owncloud/client/pull/9813

Fixes: #8839


src/resources/loadresources.cpp
```diff
@@ -0,0 +1,38 @@
+#include "loadresources.h"
+
+#include <qglobal.h>
+
+using namespace OCC;
+
+void static load_rc()
+{
+    Q_INIT_RESOURCE(ownCloudResources_translations);
```
fmoc, line 9: According to the Qt docs, we should not need any of these calls when using a shared library, it should be handled by Qt itself. I assume you have tested this without these calls? I do not have any experience moving Qt resources into a dedicated shared library, so I'm curious.

https://doc.qt.io/qt-6/resources.html

src/resources/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+
+target_link_libraries(ownCloudResources PUBLIC Qt5::Core)
+
+set_target_properties(ownCloudResources PROPERTIES
+    OUTPUT_NAME "${APPLICATION_EXECUTABLE}Resources"
+    AUTOUIC ON
+    AUTORCC ON
+)
+
+GENERATE_EXPORT_HEADER(ownCloudResources
```
fmoc, line 19:
```suggestion
generate_export_header(ownCloudResources
```

src/resources/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+        EXPORT_MACRO_NAME OWNCLOUDRESOURCES_EXPORT
+        EXPORT_FILE_NAME owncloudresources.h
+        STATIC_DEFINE OWNCLOUD_BUILT_AS_STATIC
```
fmoc, line 22: This define is never used.

src/gui/CMakeLists.txt
```diff
@@ -186,28 +186,8 @@ set_target_properties(owncloud PROPERTIES
...
-        PREFIX translations
-        FILES "${client_compiled_translations}"
-    )
-endif()
-
-#TODO Move resources files
-target_sources(owncloud PRIVATE
-    ${PROJECT_SOURCE_DIR}/client.qrc
-    ${PROJECT_SOURCE_DIR}/core_theme.qrc)
+target_link_libraries(owncloud owncloudCore ownCloudResources)
```
fmoc, line 189: Why not link the resources into the core library?

src/resources/loadresources.cpp
TheOneRing: Hm I haven't tested it, however we would still need to export a and use symbols from this lib.

src/resources/CMakeLists.txt
TheOneRing: Its used in the generated export header.

src/gui/main.cpp
```diff
@@ -41,8 +42,8 @@ void warnSystray()
 
 int main(int argc, char **argv)
 {
-    Q_INIT_RESOURCE(client);
-
+    // load the resources
+    const OCC::ResourcesLoader resource;
```
fmoc, line 46: This variable is called `resources` in the test utils. You might want to keep the naming consistent.

fmoc (approved):

Issue #9249, opened 2021-11-30
Rewrite the Wizard


TheOneRing:
# Missing
- [x] App passwords

Its time to rewrite redesign the wizard.

In https://github.com/owncloud/client/issues/5030 we got  a proposal design for the wizard.
Lacking in the proposal are oauth and webfinger.

# Webfing:
Either an additional page after the server is selected or  an input field on the initial server page.

# Fixed url
If the url is branded we currently directly continue to the login, using oauth or basic auth.
If no account is present yet we will open the wizard on start, which with oauth will directly open the browser with out any guidance of the user.
- Provide a "Start" button and a description on what will happen after the wizard process stats.

# Advances setup
The selection of the default folder, the syncing mechanism, selective sync, or manual sync are all advanced features.
- I'd suggest to a "Show advanced settings" checkbox on all pages, only if the the option is selected the advanced page is displayed, else we end after authentication. (What to do when the default folder already exists)
  - There will only be one advanced page with settings, the option won't enable additional settings on the other pages. 
- Customers already requested to disable the advanced step, in this case we would just hide that checkbox

# TODO
- How to display errors

---
Related:
- https://github.com/owncloud/client/issues/5064
- https://github.com/owncloud/client/issues/5030

TheOneRing:
@tbsbdr While looking deeper I had the idea of using setup page in between the pages, this page then would display a spinner/error.

tbsbdr:
Not sure if I understood you - what problem would a in-between page solve?

TheOneRing:
Current todos:
- VFS
- Changing target dir
- Integrate webfinger https://github.com/owncloud/client/pull/9366
- Selective sync?

fmoc:
What needs to be done regarding app passwords? They work just fine already, it seems.

TheOneRing:
> What needs to be done regarding app passwords? They work just fine already, it seems.

We used to display a url to the app password settings in oc10 and mention the existance of app passwords.

fmoc:
Looks like we finished all the TODOs. Closing therefore.

HanaGemela:
Test:
Regression test that no features are missing in comparison with 2.11 client

HanaGemela:
Some features missing - will be fixed in https://github.com/owncloud/client/issues/10266

PR #9841, opened 2022-06-30
Implement Webfinger protocol in setup wizard

fmoc:
Based on #9779. Uses the protocol implementation from https://github.com/owncloud/client/pull/9366. Contributes to #9249.

Opened as a draft because this requires some more cleanup and the error handling could be a little more user friendly.

Tested against the customer's production system.

### DIFF
--- a/src/gui/newwizard/states/webfingersetupwizardstate.cpp
+++ b/src/gui/newwizard/states/webfingersetupwizardstate.cpp
@@ -1,0 +1,75 @@
+/*
+ * Copyright (C) Fabian Müller <fmueller@owncloud.com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+ * for more details.
+ */
+
+#include "webfingersetupwizardstate.h"
+#include "creds/webfinger.h"
+#include "determineauthtypejobfactory.h"
+#include "pages/webfingersetupwizardpage.h"
+
+namespace OCC::Wizard {
+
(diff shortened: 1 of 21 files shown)


ownclouders:
Results for <strong>GUI-tests</strong> https://drone.owncloud.com/owncloud/client/12273/6/1

<details>


</details>

src/gui/newwizard/states/webfingersetupwizardstate.cpp
```diff
@@ -0,0 +1,63 @@
...
+}
+
+void WebFingerSetupWizardState::evaluatePage()
+{
+    auto *webFingerSetupWizardPage = qobject_cast<WebFingerSetupWizardPage *>(_page);
+    OC_ASSERT(webFingerSetupWizardPage != nullptr);
+
+    const QString resource = QStringLiteral("acct:%1").arg(webFingerSetupWizardPage->username());
+
+    //    _context->accountBuilder().setWebFingerData(QStringLiteral("acct:bla@example.org"), QUrl(QStringLiteral("https://oauth.example.org")));
```
TheOneRing, line 34: cleanup?

TheOneRing (changes requested):
Webfinger is supposed to be optional and depending on a branding parameter.

TheOneRing (changes requested):
`07-08 12:55:22:576 [ fatal default ]:	ENSURE: "webFinger->href().isEmpty()" in file /Users/hannah/ownbuild/master/downloads/`

fmoc:
I'll handle that error more gracefully.

TheOneRing:
The user name is the user name and should not contain the webfinger api prefix.
image

TheOneRing (changes requested):
⬆️

fmoc:
You must be using an old state. This has been fixed. https://github.com/owncloud/client/pull/9841/files#diff-fb7667184e5f26f7b04f378cf8d0971deb76e2ffe35f18d138d8eb103494aabbR61

TheOneRing:
> You must be using an old state. This has been fixed. https://github.com/owncloud/client/pull/9841/files#diff-fb7667184e5f26f7b04f378cf8d0971deb76e2ffe35f18d138d8eb103494aabbR61

You are right.

However please set the dav password when we initialize the oauth workflow, this ensures that the id entered by the user is passed to the idp/oauth2 and the user is not asked to enter the id again.

Issue #9826, opened 2022-06-28
[QA] make GUI test result easy to read


Talank:
### Describe the QA issue

It would be handy if we could print the test id (or test scenario name) in the server log of GUI test so that it would be easy to link the GUI test result (in html format) to the related server log 

I think we can get the test scenario name from the test context in squish. And logging that on server log should solve our problem for now.

PR #9853, opened 2022-07-04
[Tests-only] Run `append scenario title to serverlog` code  only in CI

amrita-shrestha:
### Description
This PR 
- refactor appending scenario title to serverlog.log  so that it runs only in CI 

### Related Issue:
- https://github.com/owncloud/client/issues/9826

### DIFF
--- a/test/gui/shared/scripts/bdd_hooks.py
+++ b/test/gui/shared/scripts/bdd_hooks.py
@@ -53,15 +53,16 @@ def hook(context):
     }
 
     # log tests scenario title on serverlog file
-    guiTestReportDir = os.environ.get("GUI_TEST_REPORT_DIR")
-    f = open(guiTestReportDir + "/serverlog.log", "a")
-    f.write(
-        str((datetime.now()).strftime("%H:%M:%S:%f"))
-        + "\tBDD Scenario: "
-        + context._data["title"]
-        + "\n"
-    )
-    f.close()
+    if os.getenv('CI'):
+        guiTestReportDir = os.environ.get("GUI_TEST_REPORT_DIR")
+        f = open(guiTestReportDir + "/serverlog.log", "a")
+        f.write(
+            str((datetime.now()).strftime("%H:%M:%S:%f"))
+            + "\tBDD Scenario: "
+            + context._data["title"]
+            + "\n"
+        )
+        f.close()
 
     # read configs from environment variables
     context.userData = {}


test/gui/shared/scripts/bdd_hooks.py
```diff
@@ -127,10 +128,14 @@ def hook(context):
         filename = context._data["title"].replace(" ", "_") + ".png"
         directory = os.environ["GUI_TEST_REPORT_DIR"] + "/screenshots"
 
-        if not os.path.exists(directory):
+        while not (os.path.exists(directory)):
             os.makedirs(directory)
```
saw-jan, line 132: You might get infinite loop (in rare case). `if` is good to go.
```suggestion
        if not os.path.exists(directory):
            os.makedirs(directory)
```

sushmita56 (commented):
run `black test/gui` command and push again

test/gui/shared/scripts/bdd_hooks.py
```diff
@@ -127,10 +128,14 @@ def hook(context):
...
 
-        pb.savev(os.path.join(directory, filename), "png", [], [])
+        try:
+            pb.savev(os.path.join(directory, filename), "png", [], [])
+        except:
+            pb.savev(os.path.join(directory, filename), "png", [], [])
```
saw-jan, line 137: ignore the screenshot file if it doesn't exists.
```suggestion
        try:
            pb.savev(os.path.join(directory, filename), "png", [], [])
        except:
            pass
```

saw-jan (commented):
LGTM

sushmita56 (commented):
LGTM

test/gui/shared/scripts/bdd_hooks.py
```diff
@@ -53,15 +53,16 @@ def hook(context):
...
-    guiTestReportDir = os.environ.get("GUI_TEST_REPORT_DIR")
-    f = open(guiTestReportDir + "/serverlog.log", "a")
-    f.write(
-        str((datetime.now()).strftime("%H:%M:%S:%f"))
-        + "\tBDD Scenario: "
-        + context._data["title"]
-        + "\n"
-    )
-    f.close()
+    if os.getenv('CI'):
```
individual-it, line 56: don't we need the server log when running the tests locally?

test/gui/shared/scripts/bdd_hooks.py
```diff
@@ -130,7 +131,10 @@ def hook(context):
         if not os.path.exists(directory):
             os.makedirs(directory)
 
-        pb.savev(os.path.join(directory, filename), "png", [], [])
+        try:
+            pb.savev(os.path.join(directory, filename), "png", [], [])
+        except:
+            pass
```
individual-it, line 137: why is that upload failing, and if it happens the developer will be puzzled by not seeing the screenshot.
I think we should not just cover the issue but at lease report that the upload failed

individual-it (changes requested):

test/gui/shared/scripts/bdd_hooks.py
saw-jan, line 137: @amrita-shrestha 
There seems to be another issue with the screenshot. So I would recommend not to include fixes for screenshot here but instead create a new issue and investigate the issue and make fix in another PR.

saw-jan (changes requested):
@sushmita56 Please, link the issue in this comment as well

test/gui/shared/scripts/bdd_hooks.py
amrita-shrestha: ok

test/gui/shared/scripts/bdd_hooks.py
amrita-shrestha: Locally `server log` can be viewed in the terminal. In my view, we don't need this while running tests locally.

test/gui/shared/scripts/bdd_hooks.py
amrita-shrestha: For the screenshot, I will create the next PR.

individual-it (approved):

saw-jan (approved):
LGTM

PR #9865, opened 2022-07-06
Fix potential crash when the log stream is destroyed

TheOneRing:


changelog/unreleased/9864
```diff
@@ -0,0 +1,3 @@
+Bugfix: We fixed a potential crash
```
fmoc, line 1: Seems a bit short.

fmoc (approved):

Issue #9879, opened 2022-07-07
[QA] `Scenario: Sync all is selected by default` started to fail


sushmita56:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the QA issue

The scenario ` Sync all is selected by default` failed on nightly build[July 7] on step `And the user "Alice" clicks on the next button in sync connection wizard` with the following error

```
RuntimeError: Error in type() invocation: Object '{container={name='groupBox' type='QGroupBox' visible='1' window={type='OCC::FolderWizard' unnamed='1' visible='1' windowTitle='Add Folder Sync Connection'}} name='localFolderLineEdit' type='QLineEdit' visible='1'}' not found. Could not match properties:
    container for object name: '{container={name='groupBox' type='QGroupBox' visible='1' window=:add_Folder_Sync_Connection_OCC_FolderWizard} name='localFolderLineEdit' type='QLineEdit' visible='1'}'
    visible for object name: '{name='groupBox' type='QGroupBox' visible='1' window=:add_Folder_Sync_Connection_OCC_FolderWizard}'
```
And we observed the same behavior locally too.

### Log:
[GUI log](https://cache.owncloud.com/public/owncloud/client/12394/guiReportUpload/index.html)

### Screenshots
![Screenshot from 2022-07-07 11-57-38](https://user-images.githubusercontent.com/66173400/177704653-013c6f29-1fa9-47aa-bca1-45eab8af9c0f.png)

sushmita56:
The same crash is observed in today's nightly build too.
https://cache.owncloud.com/public/owncloud/client/12417/guiReportUpload/index.html
![Screenshot from 2022-07-08 09-42-17](https://user-images.githubusercontent.com/66173400/177914105-62ae3c34-0244-4363-b649-feea4d6335e1.png)

PR #9889, opened 2022-07-08
Refactor selector of choose local sync

amrita-shrestha:
### Description:
This PR fixes the fail in `  And the user "Alice" clicks on the next button in sync connection wizard` step
due to selector change

### Related issue:
[Scenario: Sync all is selected by default started to fail ](https://github.com/owncloud/client/issues/9879)

### DIFF
--- a/test/gui/shared/scripts/pageObjects/AccountConnectionWizard.py
+++ b/test/gui/shared/scripts/pageObjects/AccountConnectionWizard.py
@@ -53,10 +53,10 @@ class AccountConnectionWizard:
         "window": names.owncloudWizard_OCC_OwncloudWizard,
     }
     CHOOSE_LOCAL_SYNC_FOLDER = {
-        "container": names.add_Folder_Sync_Connection_groupBox_QGroupBox,
         "name": "localFolderLineEdit",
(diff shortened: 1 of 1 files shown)


test/gui/shared/scripts/pageObjects/AccountConnectionWizard.py
```diff
@@ -52,11 +52,11 @@ class AccountConnectionWizard:
         "visible": 1,
         "window": names.owncloudWizard_OCC_OwncloudWizard,
     }
-    CHOOSE_LOCAL_SYNC_FOLDER = {
-        "container": names.add_Folder_Sync_Connection_groupBox_QGroupBox,
+    Add_Folder_Sync_Connection_localFolderLineEdit_QLineEdit = {
```
sushmita56, line 55:
```suggestion
   CHOOSE_LOCAL_SYNC_FOLDER = {
```
I think it would be better to use the same name ` CHOOSE_LOCAL_SYNC_FOLDER` as the selector assigned to it is used to choose the local sync directory.

sushmita56 (approved):

Issue #9725, opened 2022-05-24
SyncConflictTest is unreliable (Timeout) 


TheOneRing:
14 - SyncConflictTest (Timeout) 
I attached the ci logs.
[logs_10214.zip](https://github.com/owncloud/client/files/8761163/logs_10214.zip)

PR #9926, opened 2022-07-15
Ensure we remove a child job from before we trigger the next

TheOneRing:
A delete job that triggered slotComputeContentChecksum could reach slotComputeTransmissionChecksum before the delete job was removed from _jobs.
The upload job would then not start the first chunk upload and be stalled for ever.

Fixes: #9725

### DIFF
--- a/src/libsync/abstractnetworkjob.cpp
+++ b/src/libsync/abstractnetworkjob.cpp
@@ -221,6 +221,7 @@ void AbstractNetworkJob::slotFinished()
 
     bool discard = finished();
     if (discard) {
+        Q_EMIT abstractJobFinished();
         qCDebug(lcNetworkJob) << "Network job finished" << this;
         deleteLater();
     }
--- a/src/libsync/abstractnetworkjob.cpp
+++ b/src/libsync/abstractnetworkjob.cpp
@@ -218,13 +218,11 @@ void AbstractNetworkJob::slotFinished()
             qCWarning(lcNetworkJob) << "Don't retry:" << _reply->url();
         }
     }
-
-    bool discard = finished();
-    if (discard) {
-        Q_EMIT abstractJobFinished();
-        qCDebug(lcNetworkJob) << "Network job finished" << this;
-        deleteLater();
-    }
+    Q_EMIT aboutToFinishSignal(AbstractNetworkJob::QPrivateSignal());
+    finished();
+    Q_EMIT finishedSignal(AbstractNetworkJob::QPrivateSignal());
+    qCDebug(lcNetworkJob) << "Network job finished" << this;
+    deleteLater();
 }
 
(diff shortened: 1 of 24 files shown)


ownclouders:
Results for <strong>GUI-tests</strong> https://drone.owncloud.com/owncloud/client/12514/6/1

<details>


</details>

src/libsync/abstractnetworkjob.cpp
```diff
@@ -218,12 +218,11 @@ void AbstractNetworkJob::slotFinished()
             qCWarning(lcNetworkJob) << "Don't retry:" << _reply->url();
         }
     }
-
-    bool discard = finished();
-    if (discard) {
-        qCDebug(lcNetworkJob) << "Network job finished" << this;
-        deleteLater();
-    }
+    Q_EMIT aboutToFinishedSignal(AbstractNetworkJob::QPrivateSignal());
```
fmoc, line 221: I wonder whether this wants to be called `aboutToFinishedSignal`.

src/libsync/bandwidthmanager.h
```diff
@@ -87,14 +87,14 @@ public slots:
     qint64 _relativeUploadLimitProgressAtMeasuringRestart;
     qint64 _currentUploadLimit;
 
-    std::list<GETJob *> _downloadJobList;
+    std::list<GETFileJob *> _downloadJobList;
```
fmoc, line 90: Maybe we should call the class `GetFileJob` to use the naming pattern used by the other classes.

src/libsync/propagatedownload.cpp
```diff
@@ -110,12 +117,23 @@ void GETFileJob::start()
     AbstractNetworkJob::start();
 }
 
+void GETFileJob::finished()
+{
+    if (_bandwidthManager) {
+        _bandwidthManager->unregisterDownloadJob(this);
+    }
+    if (reply()->bytesAvailable() && _httpOk) {
+        // we where throttled, write out the remaining data
```
fmoc, line 126:
```suggestion
        // we were throttled, write out the remaining data
```

src/libsync/propagateupload.cpp
```diff
@@ -89,17 +89,14 @@ void PUTFileJob::start()
     AbstractNetworkJob::start();
 }
 
-bool PUTFileJob::finished()
+void PUTFileJob::finished()
```
fmoc, line 92: See above. This should probably be renamed to `PutFileJob`, maybe in a follow-up PR.

fmoc (changes requested):

src/libsync/bandwidthmanager.h
TheOneRing: The class had that name for 10 years

fmoc (approved):

PR #9944, opened 2022-07-21
Cleanup checksum handling

TheOneRing:
Looks like this change requires gcc > 10

### DIFF
--- a/src/common/checksumalgorithms.h
+++ b/src/common/checksumalgorithms.h
@@ -28,6 +28,7 @@ namespace CheckSums {
         SHA1 = QCryptographicHash::Sha1,
         MD5 = QCryptographicHash::Md5,
         ADLER32 = 100,
+        DUMMY_FOR_TESTS,
         Error
     };
 
@@ -44,6 +45,8 @@ namespace CheckSums {
             return "MD5";
         case Algorithm::ADLER32:
             return "ADLER32";
+        case Algorithm::DUMMY_FOR_TESTS:
+            return "DUMMY_FOR_TESTS";
         case Algorithm::Error:
             break;
         }
@@ -67,7 +70,8 @@ namespace CheckSums {
         pair(Algorithm::SHA256),
         pair(Algorithm::SHA1),
         pair(Algorithm::MD5),
-        pair(Algorithm::ADLER32)
+        pair(Algorithm::ADLER32),
+        pair(Algorithm::DUMMY_FOR_TESTS)
     };
 
     inline Algorithm fromName(std::string_view s)
--- a/src/common/checksumalgorithms.h
+++ b/src/common/checksumalgorithms.h
@@ -81,15 +81,5 @@ namespace CheckSums {
         });
         return it != All.end() ? it->first : Algorithm::Error;
     }
-
-    inline QByteArray calcCryptoHash(QIODevice *device, QCryptographicHash::Algorithm algo)
-    {
-        QCryptographicHash crypto(algo);
-        if (crypto.addData(device)) {
-            return crypto.result().toHex();
-        }
-        return {};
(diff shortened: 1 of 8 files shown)


ownclouders:
Results for <strong>GUI-tests</strong> https://drone.owncloud.com/owncloud/client/12581/6/1

<details>


</details>

src/common/checksumalgorithms.h
```diff
@@ -0,0 +1,95 @@
...
+
+    inline Algorithm fromName(std::string_view s)
+    {
+        auto it = std::find_if(All.begin(), All.end(), [s](const auto &it) {
+            return it.second == s;
+        });
+        return it != All.end() ? it->first : Algorithm::Error;
+    }
+
+    inline QByteArray calcCryptoHash(QIODevice *device, QCryptographicHash::Algorithm algo)
```
fmoc, line 85: This could (and probably should) be a template, same goes for `toString`.

src/common/checksumalgorithms.h
```diff
@@ -0,0 +1,95 @@
...
+#pragma once
+
+#include <QCryptographicHash>
+#include <QString>
+
+namespace OCC {
+namespace CheckSums {
+    // Sorted by priority
+    enum class Algorithm {
+        SHA3_256 = QCryptographicHash::Sha3_256,
```
fmoc, line 27: Does the sorting have any practical effect? You're assigning the values from Qt, so we are using their sorting (except for the last 3 entries at least).

Edit: I see, priorities are handled below. I guess the comment is just stale. Perhaps you can clarify that this is just sorted for the reader here, and the real priorizing is done below.

src/common/checksums.cpp
```diff
@@ -143,14 +123,19 @@ QByteArray findBestChecksum(const QByteArray &_checksums)
...
-    const auto checksums = QString::fromUtf8(_checksums);
-    int i = 0;
+
     // The order of the searches here defines the preference ordering.
-    if (-1 != (i = checksums.indexOf(QLatin1String("SHA3-256:"), 0, Qt::CaseInsensitive))
-        || -1 != (i = checksums.indexOf(QLatin1String("SHA256:"), 0, Qt::CaseInsensitive))
-        || -1 != (i = checksums.indexOf(QLatin1String("SHA1:"), 0, Qt::CaseInsensitive))
-        || -1 != (i = checksums.indexOf(QLatin1String("MD5:"), 0, Qt::CaseInsensitive))
-        || -1 != (i = checksums.indexOf(QLatin1String("ADLER32:"), 0, Qt::CaseInsensitive))) {
+    // usually not a good idea to use toUpper on a byte array but we only care for the id which has no special chars
```
fmoc, line 128: Sounds like you really want to have an assertion there. Assumptions need to be verified.

src/common/checksumalgorithms.h
fmoc, line 85: Why not use your own enum here?

fmoc (changes requested):

src/common/checksumalgorithms.h
```diff
@@ -0,0 +1,95 @@
...
+            return "SHA256";
+        case Algorithm::SHA1:
+            return "SHA1";
+        case Algorithm::MD5:
+            return "MD5";
+        case Algorithm::ADLER32:
+            return "ADLER32";
+        case Algorithm::DUMMY_FOR_TESTS:
+            return "DUMMY_FOR_TESTS";
+        case Algorithm::Error:
```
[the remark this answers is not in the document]
TheOneRing: As soon as I use default I'll no longer get compiler warnings about missing values.

fmoc (approved):

PR #9948, opened 2022-07-22
[GUI-tests] Open only one socket connection per test scenario

saw-jan:
In the GUI tests, the function `hasSyncStatus` opens a new socket connection every time it is called. The main issue is with `waitForFileOrFolderToHaveSyncStatus` function which calls `hasSyncStatus` function rapidly until the condition is met. So, there will be many socket connection requests in a single test scenario (which may or may not be an issue for the Unix socket server).

IMO, I think we must open only a single socket connection per test scenario. So, in this PR,  I have refactored the test code to only create a new socket connection if there is none and reuse the existing connection until a test scenario finishes and close the connection afterward. This way, we will open only one socket connection per scenario.

I tried to reuse a single connection for a whole test suite but that was not possible. I think the connection will be different every time we add a user to the client (I have no idea).

### DIFF
--- a/test/gui/shared/scripts/bdd_hooks.py
+++ b/test/gui/shared/scripts/bdd_hooks.py
@@ -151,6 +151,12 @@ def scenarioFailed():
 
 @OnScenarioEnd
 def hook(context):
+    # close socket connection
+    global socketConnect
+    if socketConnect:
+        socketConnect.connected = False
+        socketConnect._sock.close()
+
     # Currently, this workaround is needed because we cannot find out a way to determine the pass/fail status of currently running test scenario.
     # And, resultCount("errors")  and resultCount("fails") return the total number of error/failed test scenarios of a test suite.
     global previousFailResultCount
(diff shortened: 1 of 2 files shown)


TheOneRing:
The socket api I designed for long running connections.
The windows explroer for examples connect to the socket stays connected until either the client(the explorer) or server (the desktop client) are closed.
So I agree that you probably should keep the socket open.

ownclouders:
Results for <strong>GUI-tests</strong> https://drone.owncloud.com/owncloud/client/12590/6/1

:boom: The GUI tests failed.

GUI Logs: https://cache.owncloud.com/public/owncloud/client/12590/guiReportUpload/index.html

<details>

Server Logs: https://cache.owncloud.com/public/owncloud/client/12590/guiReportUpload/serverlog.log

</details>

saw-jan:
> The socket api I designed for long running connections. The windows explroer for examples connect to the socket stays connected until either the client(the explorer) or server (the desktop client) are closed. So I agree that you probably should keep the socket open.

If the connection will get closed after closing the desktop client then I guess we won't be able to reuse a single connection for a whole test-suite. That's why I was getting trouble using single connection for a test-suite.

saw-jan:
I don't think the test failures with the following screen are because of the test code.

logging_out

saw-jan:
Blocked until https://github.com/owncloud-ci/squish/issues/34

amrita-shrestha (approved):
Like you mentioned above using a single socket for a test suite has a problem then this changes LGTM

PR #9950, opened 2022-07-22
Enable spaces only on new accounts

TheOneRing:
Depends on https://github.com/owncloud/client/pull/9887


ownclouders:
Results for <strong>GUI-tests</strong> https://drone.owncloud.com/owncloud/client/12593/6/1

:boom: The GUI tests failed.

GUI Logs: https://cache.owncloud.com/public/owncloud/client/12593/guiReportUpload/index.html

<details>

Server Logs: https://cache.owncloud.com/public/owncloud/client/12593/guiReportUpload/serverlog.log

Screenshots:
- [Sync_all_is_selected_by_default.png](https://cache.owncloud.com/public/owncloud/client/12593/guiReportUpload/screenshots/Sync_all_is_selected_by_default.png)
- [Sync_only_one_folder_from_the_server.png](https://cache.owncloud.com/public/owncloud/client/12593/guiReportUpload/screenshots/Sync_only_one_folder_from_the_server.png)

</details>

src/gui/folderman.cpp
```diff
@@ -1456,6 +1452,11 @@ Folder *FolderMan::addFolderFromWizard(AccountStatePtr accountStatePtr, const QS
     return newFolder;
 }
 
+Folder *FolderMan::addFolderFromWizardResult(AccountStatePtr accountStatePtr, const FolderWizard::Result &config)
```
fmoc, line 1455: The name is ambiguous since we have two wizards.

fmoc (changes requested):

src/gui/folderman.cpp
TheOneRing: And both functions can be called from anywhere

fmoc: The problem is that the result you pass is not a "wizard result", it is a "folder wizard result". In case you want to generalize the result class, it should be moved to another namespace.

fmoc: P.S.: We should try to always distinguish properly between folder wizard and setup wizard to avoid any confusion within the code base.

TheOneRing: Not sure whether the now ridiculously long name helps but here it is.

fmoc (approved):

PR #9956, opened 2022-07-26
Don't abort sync to dehydrate files

TheOneRing:
Notes for testing:
- Lots of files online
- Initial sync
- While the sync is running, double click on a placeholder already created
  - The file should be downloaded in reasonable time
  - The sync progress bar should behave sensible
  -  No warnings are expected
  - The initial sync continues

### DIFF
--- a/changelog/unreleased/9956
+++ b/changelog/unreleased/9956
@@ -1,0 +1,8 @@
+Enhancement: Don't abort sync if a user requests a file
+
+Previously we aborted any running sync if a user requested a file that was not yet available locally.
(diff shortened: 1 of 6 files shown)


dragotin (approved):
This PR does not look too dangerous because it "just" removes some stuff. The "QA Question" however is if everything is still working of course... so well, approve here, but still it is a dangerous thing ;-)

jnweiger:
The description hydrates a file. I asume the subject line is wrong. Right?

TheOneRing:
Yes looks like I massed that up

PR #9966, opened 2022-07-28
Don't set the pin state of the root folder but assume unspecified

TheOneRing:
For suffix vfs the code returned always local by default and thus needed the setRootPinState.
Obviously that wrong with anything but he suffix plugin.

For testing:
- Nothing for Windows VFS
- Suffix VFS: test how pinning behaves for child items of pinned directories.
- What happens on enable disable VFS.
- Adding a folder with files

The expected behavior is:
- Remote new remote files are online only by default.
- Existing local files stay available
- Pin states persist.


fmoc (approved):

jnweiger:
Not sure about the expected behaviour. This is what I get: https://github.com/owncloud/client/issues/10035

PR #9987, opened 2022-08-01
Use existing helper to display directory in file browser

### LINKED ISSUE #9750: [QA] AppImage not terminated on 'Quit'
Seen with ownCloud-3.0.0-daily20220530.7712.AppImage on openSUSE 15.3

Steps to reproduce:
- start the AppImage and add an account
- choose 'Show in file manager' from dot menu to start "dolphin"
- 'Quit' the client

Result: it seems that the client is terminated correctly but the the AppImage is still running

appimage-running

Only after closing "dolphin" the AppImage process is stopped.

fmoc:
Should fix #9750 (which I have not been able to reproduce so far).

This change should reduce the cases in which we fall back to `QDesktopServices` to open a directory in the user's file manager on Linux. The helper in use should usually detach (orphan) the process from the parent process and therefore the AppImage runtime can exit just fine when the client is terminated.

Not sure if we need a changelog item, this possibly fixes an issue present in 2.10. However, we considered the AppImage support experimental there.

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -57,6 +57,7 @@
 #include "account.h"
 #include "askexperimentalvirtualfilesfeaturemessagebox.h"
 #include "askforoauthlogindialog.h"
+#include "openfilemanager.h"
 
 namespace OCC {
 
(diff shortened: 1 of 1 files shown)


ownclouders:
Results for <strong>GUI-tests</strong> https://drone.owncloud.com/owncloud/client/12714/6/1

:boom: The GUI tests failed.

GUI Logs: https://cache.owncloud.com/public/owncloud/client/12714/guiReportUpload/index.html

<details>

Server Logs: https://cache.owncloud.com/public/owncloud/client/12714/guiReportUpload/serverlog.log

Screenshots:
- [Sync_all_is_selected_by_default.png](https://cache.owncloud.com/public/owncloud/client/12714/guiReportUpload/screenshots/Sync_all_is_selected_by_default.png)
- [Sync_only_one_folder_from_the_server.png](https://cache.owncloud.com/public/owncloud/client/12714/guiReportUpload/screenshots/Sync_only_one_folder_from_the_server.png)

</details>

TheOneRing:
As far as I can tell this doesn't solve the issue it just adds an other level of complexification.

TheOneRing (changes requested):
nope

fmoc:
@TheOneRing checked Qt's internal workflow, and this won't solve the problem. I  Fun fact: we found out we can replace the internal workflow of that helper with a much simpler D-Bus based solution even. A PR is on the way.

PR #9993, opened 2022-08-02
Support overrideServerUrlV2 branding in wizard

fmoc:
One of the last remaining branding options to be implemented in the new wizard.

I wasn't sure where to put `getNavigationEntries`, feedback appreciated. Opening as a draft to make sure we discuss this first.

### DIFF
--- a/src/gui/newwizard/enums.cpp
+++ b/src/gui/newwizard/enums.cpp
@@ -14,6 +14,7 @@
 
 // just a stub so the MOC file can be included somewhere
 #include "moc_enums.cpp"
+#include "theme.h"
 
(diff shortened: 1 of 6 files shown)


TheOneRing:
Maybe we should still start with a welcome page, even so the the url is fixed it provides more context to the user?
image

fmoc:
Ah right, that is what we intended to do initially, right?

fmoc:
screenshot_2022-08-03_17-14-40

Looks much better. I had to revert most of my changes, though, but this looks and feels much better, both UX and code wise.

TheOneRing (approved):

PR #10000, opened 2022-08-04
Fix OAuth2 refresh

### LINKED ISSUE #9975: OIDC dynamic registation fails as we don't restore _dynamicRegistrationData from keychain
When we refreshAuthentication is called the _dynamicRegistrationData  was not fetched from the keychain.
This results in a new registration.
With the now new client id, the refresh token is invalid.

Tested with https://ocis.team.owncloud.works/
Requires https://github.com/owncloud/client/pull/9974

fmoc:
Moves the relevant code into our account-based subclass.

Fixes #9975.

### DIFF
--- a/src/libsync/creds/oauth.cpp
+++ b/src/libsync/creds/oauth.cpp
@@ -534,7 +534,22 @@ AccountBasedOAuth::AccountBasedOAuth(AccountPtr account, QObject *parent)
 
 void AccountBasedOAuth::startAuthentication()
 {
-    updateCredentialsAndRun([this]() {
+    auto credentialsJob = _account->credentialManager()->get(dynamicRegistrationDataC());
+
+    connect(credentialsJob, &CredentialJob::finished, this, [=] {
+        credentialsJob->deleteLater();
+
+        const auto data = [credentialsJob]() -> QVariantMap {
+            if (credentialsJob->data().isValid()) {
+                return credentialsJob->data().value<QVariantMap>();
+            } else {
+                qCCritical(lcOauth) << "Failed to read client id" << credentialsJob->errorString();
+                return {};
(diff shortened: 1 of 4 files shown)


src/libsync/creds/oauth.cpp
```diff
@@ -636,4 +571,89 @@ void AccountBasedOAuth::fetchWellKnown()
...
+                return credentialsJob->data().value<QVariantMap>();
+            } else {
+                qCCritical(lcOauth) << "Failed to read client id" << credentialsJob->errorString();
+                return {};
+            }
+        }();
+
+        _dynamicRegistrationData = data;
+
+        doRefresh();
```
TheOneRing, line 655: Does using the lambda doRefresh actually solve a problem?

I'm kind of getting the impression that one linear spaghetti style function here would be easier to parse than what we  got here with  doRefresh and the parse lambda.

fmoc: I tried to keep the original function encapsulated. It's already way too complex. But I also don't mind refactoring this, if you wish.

src/libsync/creds/oauth.cpp
```diff
@@ -636,4 +571,89 @@ void AccountBasedOAuth::fetchWellKnown()
...
+            } else {
+                refresh();
+            }
+        });
+        fetchWellKnown();
+    };
+
+    auto credentialsJob = _account->credentialManager()->get(dynamicRegistrationDataC());
+
+    connect(credentialsJob, &CredentialJob::finished, this, [=] {
```
TheOneRing, line 641: We got  100% duplication with startAuthentication, please refactor

TheOneRing (approved):
Looks good

Issue #10006, opened 2022-08-08
[Wizard] Sync directory selector freezes 3.0 client


michaelstingl:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the bug

Clicking the folder picker icon in the advanced configuration, freezes the client.

### Expected behavior

System folder picker should show up.

### Steps to reproduce the issue

1. Start client after fresh installation
2. Enter server URL
3. Login
4. Open advanced configuration
5. Click sync directory selector

### Screenshots

![CleanShot 2022-08-08 at 11 35 34@2x](https://user-images.githubusercontent.com/214010/183388092-c7b6c6a6-883a-4469-9938-084c1cb97298.png)

### Client version number

testpilotcloud 3.0.0.8280-daily20220808 2d3ba7 Aug 8 2022 02:13:56
Libraries Qt 5.15.5, OpenSSL 1.1.1q 5 Jul 2022
Using virtual files plugin: suffix
osx-21.6.0

(running on macOS Monterey 12.5 (21G72))

fmoc:
Can reproduce this behavior with build 8281. There is no way to exit this deadlock, apparently. I checked an older build (8209) as well, same behavior, so #9994 is not to blame (which was my first guess). The client does not provide any log messages.

I also tried creating the directory or using an existing directory to make sure the dialog won't be displayed because of this, but these also don't seem to be the cause. I'll set up a development build. I suspect that the issue is within the Qt macOS platform code.

Note that before #9994 was merged, the wizard could be closed via the `X` button, but the remaining app has frozen as well.

TheOneRing:
My guess, the folder picker is stacked below the wizard.

fmoc:
Should be fixed by https://github.com/owncloud/client/pull/10008, right?

TheOneRing:
yes

TheOneRing:
And it will be closed once merged

fmoc:
Now it should be (note: commit messages won't trigger auto-close).

TheOneRing:
Commit messages will do so

TheOneRing:
https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

PR #10008, opened 2022-08-09
Fix dialog raising

TheOneRing:
Fixes: https://github.com/owncloud/client/issues/10006

### DIFF
--- a/src/gui/ignorelisteditor.cpp
+++ b/src/gui/ignorelisteditor.cpp
@@ -34,7 +34,7 @@ static const int skippedLinesRole = Qt::UserRole;
 static const int isGlobalRole = Qt::UserRole + 1;
 
 IgnoreListEditor::IgnoreListEditor(QWidget *parent)
-    : QDialog(parent, Qt::Sheet)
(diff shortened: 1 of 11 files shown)


src/gui/ignorelisteditor.cpp
```diff
@@ -34,7 +34,7 @@ static const int skippedLinesRole = Qt::UserRole;
 static const int isGlobalRole = Qt::UserRole + 1;
 
 IgnoreListEditor::IgnoreListEditor(QWidget *parent)
-    : QDialog(parent, Qt::Sheet)
+    : QDialog(parent)
```
fmoc, line 37: Should we remove `Qt::Sheet` from the other locations as well, now that this is handled differently in that utility method?

fmoc (changes requested):

fmoc (approved):
History should be cleaned up a bit. Since this removes the `QSheet` from the constructors as suggested in https://github.com/owncloud/client/pull/9994#issuecomment-1205765854, that should probably be a separate commit in `master`.

PR #10011, opened 2022-08-09
Rewrite and unify "login required" dialogs

fmoc:
In the future, we may want to re-use the new widget classes introduced in this PR in the wizard, too, to provide a unified UX and avoid code duplication.


src/gui/creds/httpcredentialsgui.cpp
```diff
@@ -13,10 +13,12 @@
  * for more details.
  */
 
-#include "application.h"
+#include "creds/httpcredentialsgui.h"
 #include "account.h"
+#include "application.h"
 #include "common/asserts.h"
-#include "creds/httpcredentialsgui.h"
+#include "httploginrequiredwidget.h"
```
TheOneRing, line 20: Please use the full path gui/loginreq...

I'd like to add a bit more structure to the includes, at least for new code.

src/gui/creds/httpcredentialsgui.cpp
```diff
@@ -113,39 +115,15 @@ void HttpCredentialsGui::asyncAuthResult(OAuth::Result r, const QString &user,
...
-    dialog->setWindowTitle(tr("Enter Password"));
-    dialog->setLabelText(msg);
-    dialog->setTextValue(_previousPassword);
-    dialog->setTextEchoMode(QLineEdit::Password);
-    if (QLabel *dialogLabel = dialog->findChild<QLabel *>()) {
-        dialogLabel->setOpenExternalLinks(true);
-        dialogLabel->setTextFormat(Qt::RichText);
-    }
+    connect(dialog, &LoginRequiredDialog::finished, ocApp()->gui()->settingsDialog(), [this, dialog, contentWidget](const int result) {
+        contentWidget->deleteLater();
```
TheOneRing, line 122: contentWidget should be handled by parenting?

src/gui/loginrequireddialog/loginrequireddialog.cpp
```diff
@@ -0,0 +1,60 @@
...
+    for (auto [button, role] : contentWidget->buttons()) {
+        _ui->rightButtonBox->addButton(button, role);
+    }
+
+    // using a stacked widget appears to work better than a plain widget
+    // we do this in the setup wizard as well
+    _ui->contentWidget->addWidget(contentWidget);
+    _ui->contentWidget->setCurrentWidget(contentWidget);
+
+    setModal(true);
```
TheOneRing, line 51: Already handled in Utility::setModal(this);
(It also turned out that some combinations of some setModal calls can be too modal, so please rely on the magic in Utility::setModal(this);)

src/gui/loginrequireddialog/oauthloginrequiredwidget.cpp
```diff
@@ -0,0 +1,69 @@
...
+* WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+* or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+* for more details.
+*/
+
+#include "oauthloginrequiredwidget.h"
+#include "ui_oauthloginrequiredwidget.h"
+#include "gui/application.h"
+#include "gui/guiutility.h"
+#include "../creds/httpcredentialsgui.h"
```
TheOneRing, line 19: "gui/creds..

TheOneRing (changes requested):

src/gui/creds/httpcredentialsgui.cpp
fmoc: Right now the content widget has to be passed to the constructor (which makes sense, since such an object is required for the parent to work). Therefore, I have two... workarounds in mind:

- call `contentWidget->setParent` after setting up the parent
- use a setter to pass `contentWidget` to the dialog

TheOneRing: The parent of contentWidget will already be the LoginRequiredDialog as the widget will be added to one of its layouts.
Widgets take the ownership of their children.

src/gui/loginrequireddialog/oauthloginrequiredwidget.cpp
```diff
@@ -0,0 +1,69 @@
...
+* WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+* or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+* for more details.
+*/
+
+#include "oauthloginrequiredwidget.h"
+#include "gui/application.h"
+#include "gui/creds/httpcredentialsgui.h"
+#include "gui/guiutility.h"
+#include "ui_oauthloginrequiredwidget.h"
```
TheOneRing, line 19: I know that its nitpicking at that level but please move the ui include below oauthloginrequiredwidget and then add an empty line.

src/gui/creds/httpcredentialsgui.cpp
fmoc: Nevermind then.

src/gui/loginrequireddialog/httploginrequiredwidget.cpp
```diff
@@ -0,0 +1,58 @@
...
+* This program is distributed in the hope that it will be useful, but
+* WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+* or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+* for more details.
+*/
+
+#include "httploginrequiredwidget.h"
+#include "creds/httpcredentials.h"
+#include "gui/guiutility.h"
+#include "ui_httploginrequiredwidget.h"
```
TheOneRing, line 18: order

src/gui/loginrequireddialog/loginrequireddialog.cpp
```diff
@@ -0,0 +1,59 @@
...
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+ * for more details.
+ */
+
+#include "loginrequireddialog.h"
+#include "../creds/httpcredentialsgui.h"
```
TheOneRing, line 16: ?

TheOneRing (changes requested):

TheOneRing (approved):

PR #10013, opened 2022-08-09
Use correctly themed icons

fmoc:
I kept the icons within the Qt Designer files because it helps visualize changes.


TheOneRing (changes requested):
Please use QStringLiteral

TheOneRing (approved):

PR #10015, opened 2022-08-10
Ci: Don't package debug symbols

TheOneRing:
Packaging and uploading the symbols takes ages.
Our nightlies have an integrated crash reporter and local full debug builds are rather easy with craft.


fmoc:
Mind to explain what causes the slowdown?

TheOneRing:
Packaging 600mb+ takes some time.

fmoc:
Is it the compression that is problematic, or is it extracting the files from the binaries? All our CI is running on SSDs, so `tar`ing without compressing should be relatively fast.

fmoc (approved):
Still don't know for sure what exactly causes the slowdown, but I guess this is fine.

PR #10016, opened 2022-08-11
Create correct implementation of access manager

TheOneRing:


src/gui/connectionvalidator.cpp
```diff
@@ -97,7 +97,7 @@ void ConnectionValidator::systemProxyLookupDone(const QNetworkProxy &proxy)
 void ConnectionValidator::slotCheckServerAndAuth()
 {
     // in order to receive all ssl erorrs we need a fresh QNam
-    auto nam = new AccessManager(this);
```
fmoc, line 100: Should we make the constructor `protected` and make the credentials a friend class? It's not a great solution, but it'll force the use of the factory in the future.

fmoc (approved):

src/gui/connectionvalidator.cpp
TheOneRing: I guess its totally ok to be able to create an AccessManager.
Here however we need to use what the credentials provide.
Actually we only have one implementation of AccessManager, however the unit tests implement it differently.
I'd suggest that we get rid of createAM and find a different solution for the unit tests.

PR #10024, opened 2022-08-11
Apply strict strings to all our code

TheOneRing:


cmake/modules/OCApplyCommonSettings.cmake
```diff
@@ -0,0 +1,12 @@
+function(apply_common_settings targetNamme)
+    target_compile_definitions(${targetNamme}
+        PRIVATE QT_NO_CAST_TO_ASCII
```
fmoc, line 3: 4 spaces indentation is the default. Might also be easier if you just repeat `PRIVATE` on every line.

cmake/modules/OCApplyCommonSettings.cmake
```diff
@@ -0,0 +1,12 @@
+function(apply_common_settings targetNamme)
```
fmoc, line 1: I'd include `target` in the name. Also, you may want to just check if the target exists (sure, the command itself will likely fail, too, but this makes debugging easier).

Also, I'd prefer to have this refactoring on a separate commit.

src/gui/creds/httpcredentialsgui.cpp
```diff
@@ -136,11 +136,8 @@ void HttpCredentialsGui::showDialog()
 
 QString HttpCredentialsGui::requestAppPasswordText(const Account *account)
 {
-    auto baseUrl = account->url().toString();
-    if (baseUrl.endsWith('/'))
-        baseUrl.chop(1);
     return tr("<a href=\"%1\">Click here</a> to request an app password from the web interface.")
-        .arg(Utility::concatUrlPath(baseUrl, QStringLiteral("/index.php/settings/personal?sectionid=security#apppasswords")).toString());
+        .arg(Utility::concatUrlPath(account->url(), QStringLiteral("/index.php/settings/personal?sectionid=security#apppasswords")).toString());
```
fmoc, line 140: Not covered by the commit message, should probably be its own commit.

fmoc (changes requested):

fmoc (approved):

PR #10026, opened 2022-08-16
Allow user to retry OAuth2/OIDC login

fmoc:
screenshot_2022-08-16_10-48-33
screenshot_2022-08-16_10-49-05
screenshot_2022-08-16_10-49-12

I'm sure this could and should be polished. We are going to work on these widgets soon again when integrating them into the wizard, which is a good chance to improve the UX.

### DIFF
--- a/src/gui/creds/httpcredentialsgui.cpp
+++ b/src/gui/creds/httpcredentialsgui.cpp
@@ -159,9 +159,6 @@ void HttpCredentialsGui::restartOAuth()
         this, &HttpCredentialsGui::asyncAuthResult);
     connect(_asyncAuth.data(), &OAuth::destroyed,
         this, &HttpCredentialsGui::authorisationLinkChanged);
-    connect(_asyncAuth.data(), &AccountBasedOAuth::refreshError, this, [this]() {
-        // FIXME
-    });
     _asyncAuth->startAuthentication();
     emit authorisationLinkChanged();
 }
--- a/src/gui/creds/httpcredentialsgui.cpp
+++ b/src/gui/creds/httpcredentialsgui.cpp
@@ -58,24 +58,15 @@ void HttpCredentialsGui::askFromUser()
 
 void HttpCredentialsGui::askFromUserAsync()
 {
-    const auto updateOAuth = [this] {
-        _asyncAuth.reset(new AccountBasedOAuth(_account->sharedFromThis(), this));
-        connect(_asyncAuth.data(), &OAuth::result,
-            this, &HttpCredentialsGui::asyncAuthResult);
-        connect(_asyncAuth.data(), &OAuth::destroyed,
-            this, &HttpCredentialsGui::authorisationLinkChanged);
-        _asyncAuth->startAuthentication();
-        emit authorisationLinkChanged();
-    };
     if (isUsingOAuth()) {
-        updateOAuth();
+        restartOAuth();
     } else {
         // First, we will check what kind of auth we need.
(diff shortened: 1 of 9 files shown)


TheOneRing:
Maybe we should move the icon to the top left?
Especially when the error is displayed it looks like its not aligned to anything?

fmoc:
WIP in an upcoming PR. This just adds the retry functionality.

src/gui/creds/httpcredentialsgui.h
```diff
@@ -64,7 +68,7 @@ private slots:
     void authorisationLinkChanged();
 
 private:
-    QScopedPointer<OAuth, QScopedPointerObjectDeleteLater<OAuth>> _asyncAuth;
+    QScopedPointer<AccountBasedOAuth, QScopedPointerObjectDeleteLater<OAuth>> _asyncAuth;
```
TheOneRing, line 71: Wrong class in the deleter

TheOneRing (changes requested):
then merge

src/gui/loginrequireddialog/oauthloginrequiredwidget.cpp
```diff
@@ -67,4 +85,25 @@ QList<QPair<QAbstractButton *, QDialogButtonBox::ButtonRole>> OAuthLoginRequired
     };
 }
 
+void OAuthLoginRequiredWidget::showErrorLabelAndRetryButton()
+{
+    _ui->errorFrame->show();
+
+    // apparently, we need to run this in the main event loop since the buttons are owned by the layout within the login required dialog
+    QTimer::singleShot(0, [this]() {
```
TheOneRing, line 93: @erikjv  can you have a look with @fmoc why this hack is needed?
I guess there is a deeper issue here.

TheOneRing (changes requested):

TheOneRing:
Sadly the change is so outdated that I can't take over the development...

src/gui/loginrequireddialog/oauthloginrequiredwidget.cpp
erikjv: What is the order in which this method is called, and when `buttons()` is called? I mean, in the `buttons()` method, the ownership of the widget is transferred to the caller. So I'm wondering if this method is called before the layout adds the buttons to itself, and somehow invalidate its state.

fmoc:
Need to rebase this PR on https://github.com/owncloud/client/pull/10176, which fixes a couple of problems.

src/gui/creds/httpcredentialsgui.cpp
```diff
@@ -161,4 +152,18 @@ QUrl HttpCredentialsGui::authorisationLink() const
...
 
+void HttpCredentialsGui::restartOAuth()
+{
+    _asyncAuth.reset(new AccountBasedOAuth(_account->sharedFromThis(), this));
+    connect(_asyncAuth.data(), &OAuth::result,
+        this, &HttpCredentialsGui::asyncAuthResult);
+    connect(_asyncAuth.data(), &OAuth::destroyed,
+        this, &HttpCredentialsGui::authorisationLinkChanged);
+    connect(_asyncAuth.data(), &AccountBasedOAuth::refreshError, this, [this]() {
+        // FIXME
```
TheOneRing, line 163: ?

fmoc:
screenshot_2022-10-11_17-31-30

Had to start over, but I managed to bring it into a usable state. Still investigating why the error label uses this much vertical space (Qt won't let me shrink it vertically during runtime at this point).

TheOneRing (approved):

Issue #9790, opened 2022-06-13
Server not available causes error


dragotin:
### Describe the bug

When starting the client with a configured account that is not reachable at the moment (for example an ocis server on localhost which is not started), the client reacts with an error message:

![image](https://user-images.githubusercontent.com/1070214/173434515-de14c917-f669-4548-a3c3-187340c8f863.png)

### Expected behavior

Just put the account offline.

### Client version number

Version 3.0.0 git (commit 8b664899481242b2c453e74d8fa128af958e096e)

### Desktop environment (Linux only)

Linux, KDE, openSUSE Tumbleweed

bishwasojha:
- When connection to the server is lost , the client does not show any error message but it shows that the account is offline 
- When server comes back online , the client connects automatically. 
### Screenshot in Linux

![Screenshot from 2022-11-09 12-04-36](https://user-images.githubusercontent.com/33860671/200781530-2b1dd3de-7ac5-4471-bf66-8aab94fff353.png)

### same in windows 10 too

ownCloud version
```
ownCloud 3.0.0.9005-beta1 [6b12ec](https://github.com/owncloud/client/commit/6b12ecdb4236d3d485ed57ec0bc2feac72b8f340)
Libraries Qt 5.15.2, OpenSSL 3.0.2 15 Mar 2022
Using virtual files plugin: suffix
ubuntu-5.15.0-52-generic
```

PR #10028, opened 2022-08-16
Don't display error if server is unavailable

TheOneRing:
Are there errors we really should display?

Fixes: #9790

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -831,9 +831,9 @@ void AccountSettings::slotAccountStateChanged()
             break;
         }
         case AccountState::NetworkError:
+            // don't display the error to the user, https://github.com/owncloud/client/issues/9790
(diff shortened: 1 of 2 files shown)


src/gui/accountsettings.cpp
```diff
@@ -832,8 +832,7 @@ void AccountSettings::slotAccountStateChanged()
         }
         case AccountState::NetworkError:
             showConnectionLabel(tr("No connection to %1.")
-                                    .arg(server),
-                _accountState->connectionErrors());
+                                    .arg(server));
```
fmoc, line 835: How about a comment that explains that we explicitly don't want to display an error in this case?

TheOneRing: Errors are logged independent of the ui

fmoc (approved):

PR #10029, opened 2022-08-16
Schedule a new folder when wen we remove the current running folder

TheOneRing:


src/gui/folderman.cpp
```diff
@@ -1206,21 +1206,14 @@ void FolderMan::removeFolder(Folder *f)
     f->removeFromSettings();
 
     unloadFolder(f);
-    if (currentlyRunning) {
```
TheOneRing, line 1209: This code was probably meant to ensure we upload everything.  However it might come from the days where there was just one folder.

TheOneRing: Hmmm
https://github.com/owncloud/client/commit/c1045af500d18c52129d2a81f9ac51ec7ca92c8b

TheOneRing: Well I guess it really can be removed.

fmoc (approved):

PR #10036, opened 2022-08-17
Fix checksums

TheOneRing:


test/testchecksumvalidator.cpp
```diff
@@ -33,11 +33,12 @@ using namespace OCC::Utility;
 
     public slots:
 
-    void slotUpValidated(const QByteArray& type, const QByteArray& checksum) {
-         qDebug() << "Checksum: " << checksum;
-         QVERIFY(_expected == checksum );
-         QVERIFY(_expectedType == type );
-    }
+        void slotUpValidated(CheckSums::Algorithm type, const QByteArray &checksum)
```
TheOneRing, line 36: The whole file is formatted wrong

src/common/checksums.h
```diff
@@ -48,15 +48,35 @@ class SyncJournalDb;
  */
 OCSYNC_EXPORT QByteArray findBestChecksum(const QByteArray &checksums);
 
+class OCSYNC_EXPORT ChecksumHeader
+{
+public:
+    /// Parses a checksum header
```
fmoc, line 54: This comment isn't all that helpful. It's been there before, though, so nevermind.

src/common/checksumalgorithms.h
```diff
@@ -29,15 +31,19 @@
...
+
     enum class Algorithm {
         SHA3_256 = QCryptographicHash::Sha3_256,
         SHA256 = QCryptographicHash::Sha256,
         SHA1 = QCryptographicHash::Sha1,
         MD5 = QCryptographicHash::Md5,
         ADLER32 = 100,
         DUMMY_FOR_TESTS,
-        Error
+        NONE,
```
fmoc, line 43: Since this is the default value, it should probably be moved up to the top and assigned `0`.

fmoc (changes requested):

fmoc (approved):

PR #10039, opened 2022-08-18
Don't display image url in spaces browser

TheOneRing:
Fixes: #10030


src/gui/spaces/spacesmodel.cpp
```diff
@@ -14,6 +14,8 @@
 #include "spacesmodel.h"
 
 #include "common/utility.h"
+// TOD: move models out from core
```
fmoc, line 17:
```suggestion
// TODO: move models out from core
```

fmoc (changes requested):

fmoc (approved):

PR #10042, opened 2022-08-19
Remove build time from about

TheOneRing:
We have the date in the build number, no need for the unpopular __DATE__ macro.


michaelstingl (approved):
Works for me 👍 

Needs to be adjusted in the branding template too?

TheOneRing:
> 

No change needed

fmoc (approved):
Another plus: this is a great step ahead regarding reproducible builds, which might be interesting for us some day.

PR #10058, opened 2022-08-24
Fix expansion of selective sync tree

fmoc:
Before:
screenshot_2022-08-24_14-47-12

After:
screenshot_2022-08-24_14-49-07

### DIFF
--- a/changelog/unreleased/10058
+++ b/changelog/unreleased/10058
@@ -1,0 +1,6 @@
+Bugfix: Fix configuration of selective sync from account settings
+
+We removed a superfluous check which caused the directory tree in the account settings window to no longer expand beyond the first level.
+The problem was introduced in 8d0dd36d2.
+
+https://github.com/owncloud/client/pull/10058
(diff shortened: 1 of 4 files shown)


ownclouders:
Results for <strong>GUI-tests</strong> https://drone.owncloud.com/owncloud/client/12933/6/1

<details>


</details>

erikjv (approved):

fmoc:
We decided to have another look.

fmoc:
I had another look and I think this new fix is significantly better than my last attempt. Now there is a check to make sure invalid `QVariant`s won't slip through the check.

dragotin (changes requested):
The whole patch seems bogus since the isReady flag is always true. Why don't you remove the isReady flag completely? There is one more usage in `AccountSettings::slotCustomContextMenuRequested` which can be cleaned up/removed as well.

dragotin:
Currently, whenever you use `data( index, FolderStatusDelegate::IsReady)` you have access to the folder object as well. With that you could ask if the folder object is ready, which is what you're really interested in. Pulling that through the model makes that more complex in my opinion. 
So again, my recommendation is:  Drop the isReady from the model data and use `folder->isReady()` directly.

That removes alternative code pathes that actually do the same thing.

fmoc:
Makes sense, totally agree, thanks!

erikjv (approved):
Nice!

dragotin (approved):
LGTM

PR #10062, opened 2022-08-25
[not-to-merge][tests-only] Run ctests on fedora based container

saw-jan:


saw-jan:
:heavy_check_mark:  ctests can pass on the fedora36 based image

PR #10066, opened 2022-08-27
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,7 +7,7 @@ on:
       - '[0-9]+.*'
   pull_request:
   
-permissions: {} # none
+permissions: {}
 
 defaults:
   run:
@@ -17,7 +17,8 @@ jobs:
 # ------------------------------------------------------------------------------------------------------------------------------------------
   build:
     permissions:
-      contents: read # actions/upload-artifact doesn't need contents: write
+      # actions/upload-artifact doesn't need contents: write
+      contents: read
     strategy:
         matrix:
             include:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,6 +6,8 @@ on:
       - master
       - '[0-9]+.*'
   pull_request:
+  
+permissions: {} # none
 
 defaults:
   run:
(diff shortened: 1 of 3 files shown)


fmoc:
First of all, the `permissions` key only applies to the `GITHUB_TOKEN` and therefore limits the access to the GitHub API. Therefore, changes to this key can prevent, e.g., unintentional writes, providing another "safety net". However, it doesn't really close any attack vectors, since third parties can always open PRs and change the values. Only in combination with GitHub's approval mechanism for first time contributors to run workflows on GitHub actions could reasonably protect the repository, but one could argue that the additional `permissions` doesn't provide any security gain over this, since in any case, changes must be reviewed before approving workflow runs.

There are use cases in which this change can help prevent security issues, though. For instance, tools running as part of the pipeline (including random third-party GitHub actions) cannot gain access to a token capable of writing while running as part of our workflows on our main branches, for instance. PR tokens are more limited than regular ones anyway

Right now, I cannot really tell whether the claims made in https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions apply. GitHub's docs aren't clear on this, and the document does not discuss PRs. However, as said, this can help in certain cases, and that justifies the inclusion of this change. I'll post a review asap.

fmoc (changes requested):
No same-line comments, please. They're the root of all evil. Please move them above.

There is no need to document `none`, this is actually pretty obvious, and also the docs quickly reveal that information. What's good, however, is documenting why you need the `write` access, so please keep those comments.

sashashura:
Done, please take a look.

fmoc (approved):
Thanks!

PR #10068, opened 2022-08-29
[not-to-merge] Run ctests on ubuntu 22.04 based container

saw-jan:


saw-jan:
Ctests work on ubuntu 22.04

PR #10075, opened 2022-09-01
Compatibility with older python3-nautilus

nteodosio:
Quoting from [devhelp](https://tracker.debian.org/pkg/devhelp) (sorry, couldn't find a web document to link to), migrating Nautilus section:

>    The get_file_items, get_file_items_full, get_background_items a get_background_items_full methods of Nautilus.MenuProvider no longer take the window argument. Remove it from your implementations.
    If you need to keep supporting older versions of Nautilus, you can use variadic arguments:
```
  def get_file_items(self, *args):
      # `args` will be `[files: List[Nautilus.FileInfo]]` in Nautilus 4.0 API,
      # and `[window: Gtk.Widget, files: List[Nautilus.FileInfo]]` in Nautilus 3.0 API.
      files = args[-1]
```

### DIFF
--- a/shell_integration/nautilus/syncstate.py
+++ b/shell_integration/nautilus/syncstate.py
@@ -200,9 +200,13 @@ class MenuExtension_ownCloud(GObject.GObject, Nautilus.MenuProvider):
                 break
         return (topLevelFolder, internalFile)
 
-    def get_file_items(self, window, files):
+    # args in Nautilus 4.0: [files: List[Nautilus.FileInfo]]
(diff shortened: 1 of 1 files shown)


fmoc (changes requested):
Please document this workaround in the code base as well, so we won't forget in the future why the files are queried from the packed parameters.

fmoc:
Your change looks good, but I'd like to have the reasoning documented in the code as well.

fmoc (approved):
Thanks!

jbicha:
Here's the weblink for the devhelp documentation:

https://gnome.pages.gitlab.gnome.org/nautilus-python/nautilus-python-migrating-to-4.html

Issue #9866, opened 2022-07-06
[QA] Scenarios failing with the same error `KeyError: 'Brian'`


sushmita56:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the QA issue

The following four scenarios from `tst_sharing` suite 
```
Scenario: User (non-author) can not share to a group to which the file/folder is already shared
Scenario: sharee edits content of files shared by sharer
Scenario: reshare a file/folder
Scenario: try to reshare a file/folder shared without share permission
```
are failing locally with the following error while running the test from the terminal but it passes while running using the squish ide.
 ```
2022-07-06T11:37:03	ERROR     	/home/sushmita/www/client/test/gui/shared/steps/steps.py:112: Script Error	KeyError: 'Brian'
```

### Steps to reproduce the issue

NA

### Screenshots

![Screenshot from 2022-07-06 13-49-40](https://user-images.githubusercontent.com/66173400/177501367-951817f8-8101-4ed1-afa3-95ab0d033a5c.png)

saw-jan:
I was not able to reproduce this one.

amrita-shrestha:
This could be the reason: `missing Brain data in the dictionary that contain user information from middleware`.

Closing this issue because can't reproduce locally.

saw-jan:
This error happens to be a legit one. We can reproduce this by putting a scenario that creates only one user at the very first. Because if the first scenario only creates user Alice then in the second scenario where Alice and Brian are created, KeyError will be thrown.

PR #10078, opened 2022-09-02
[gui-tests] Update/Add users in the global user store

saw-jan:
We want to have the current scenario's state of the server not the state stored in global by the previous scenario. Storing state in global causes this issue #9866 if the scenarios are to be shuffled. Because if the first scenario only creates user Alice then in the second scenario where Alice and Brian are created, `KeyError` will be thrown if we want to add user Brain to desktop client.

UPDATE: user information is stored in a global variable and information of new user is added to the list

### Related issues
Fixes https://github.com/owncloud/client/issues/9866

### DIFF
--- a/test/gui/shared/steps/steps.py
+++ b/test/gui/shared/steps/steps.py
@@ -35,7 +35,6 @@ confdir = '/tmp/bdd-tests-owncloud-client/'
 confFilePath = confdir + 'owncloud.cfg'
 socketConnect = None
 
-stateDataFromMiddleware = None
 
 # File syncing in client has the following status
 SYNC_STATUS = {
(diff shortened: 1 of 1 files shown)


ownclouders:
Results for <strong>GUI-tests</strong> https://drone.owncloud.com/owncloud/client/13002/6/1

:boom: The GUI tests failed.

GUI Logs: https://cache.owncloud.com/public/owncloud/client/13002/guiReportUpload/index.html

<details>

Server Logs: https://cache.owncloud.com/public/owncloud/client/13002/guiReportUpload/serverlog.log

Screenshots:
- [Delete_a_folder.png](https://cache.owncloud.com/public/owncloud/client/13002/guiReportUpload/screenshots/Delete_a_folder.png)

</details>

amrita-shrestha (commented):
LGTM, But some questions instead of requesting user data from middleware everytime cant we store the data in a global variable, if the user data doesn't exist in a global variable variable then only requesting to middleware

saw-jan:
> LGTM, But some questions instead of requesting user data from middleware everytime cant we store the data in a global variable, if the user data doesn't exist in a global variable variable then only requesting to middleware

If we are using only user information then that's a good idea. I will have look and implement it

test/gui/shared/steps/steps.py
```diff
@@ -102,14 +102,24 @@ def step(context, displayname, host):
     )
 
 
+def getUserInfo(context, username, attribute):
+    # add and update users to the global createdUsers dict if not already there
+    # so that we don't have to request for user information in every scenario
+    # but instead get user information from the global dict
+    global createdUsers
+    if createdUsers and username in createdUsers:
```
amrita-shrestha, line 110: why if createdUsers...? Can't we do 
```suggestion
    if username in createdUsers:
```

saw-jan: Previously, `createdUsers` was intialized with `None` but now it is initialized with `{}` so we can do it.

amrita-shrestha (approved):
lgtm

PR #10089, opened 2022-09-07
Always disable VFS when experimental features are disabled

### LINKED ISSUE #9781: [QA] 3.0.0 master branded client enables Linux experimental virtual files
### Pre-submission Checks

### Describe the bug

It is possible to enable experimental Linux virtual files with 3.0.0 branded Linux client.
If 'forceVirtualFilesOption' is set, the client (here BtR client) enables VFS per default for a new account.

### Expected behavior

For branded Linux clients (except 'testpilotcloud') it should not be possible to enable the experimental VFS support. 

_No response_

### Screenshots

BtR-VFS

### Logs

_No response_

### Client version number

3.0.0 master from git 08.06.2022

openSUSE 15.3

_No response_

_No response_

### Server information

_No response_

### Additional context

_No response_

fmoc:
Fixes #9781.

### DIFF
--- a/src/gui/newwizard/pages/accountconfiguredwizardpage.cpp
+++ b/src/gui/newwizard/pages/accountconfiguredwizardpage.cpp
@@ -37,7 +37,7 @@ AccountConfiguredWizardPage::AccountConfiguredWizardPage(const QString &defaultS
         _ui->advancedConfigGroupBox->setVisible(false);
     }
 
-    if (!Theme::instance()->showVirtualFilesOption()) {
+    if (!Theme::instance()->enableExperimentalFeatures() || !Theme::instance()->showVirtualFilesOption()) {
         _ui->useVfsRadioButton->setVisible(false);
         vfsIsAvailable = false;
     }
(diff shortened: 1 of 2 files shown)


src/gui/newwizard/pages/accountconfiguredwizardpage.cpp
```diff
@@ -37,7 +37,7 @@ AccountConfiguredWizardPage::AccountConfiguredWizardPage(const QString &defaultS
         _ui->advancedConfigGroupBox->setVisible(false);
     }
 
-    if (!Theme::instance()->showVirtualFilesOption()) {
+    if (!Theme::instance()->enableExperimentalFeatures() || !Theme::instance()->showVirtualFilesOption()) {
```
TheOneRing, line 40: Only suffix vfs is considert experimental see `vfsModeIsExperimental`.

dragotin, line 40: Tbh, the `enableExperimentalFeatures()` should be moved into the `showVirtualFilesOption()`. Otherwise there are again two places where that is implemented, and the `showVirtualFilesOption` is exactly what it is about, or?

dragotin (changes requested):

src/gui/newwizard/pages/accountconfiguredwizardpage.cpp
fmoc: This is the case already. Currently, however, `forceVirtualFilesOption` is evaluated later than `showVirtualFilesOption`. "force" seems higher prio than "show". We can change this, though.

PR #10096, opened 2022-09-08
Suffix VFS: only apply local modtime changes on download

erikjv:
... and not when uploading.


fmoc (commented):
Looks good. @TheOneRing should have a look, too, though.

TheOneRing (approved):

fmoc (approved):

PR #10099, opened 2022-09-09
Tests: Clean up FileModifier API

erikjv:
- make the default values into static constexpr member fields
- remove unused method `modifyByte`
- don't rely on on-disk info like file sizes (so it can be used in dehydrated VFS modes)

### DIFF
--- a/test/testblacklist.cpp
+++ b/test/testblacklist.cpp
@@ -79,8 +79,9 @@ private slots:
             QVERIFY(entry._ignoreDuration > 0);
             QCOMPARE(entry._requestId, reqId);
 
-            if (remote)
+            if (remote) {
                 QCOMPARE(journalRecord(fakeFolder, "A")._etag, initialEtag);
+            }
         }
         cleanup();
 
@@ -131,6 +132,10 @@ private slots:
                 QCOMPARE(journalRecord(fakeFolder, "A")._etag, initialEtag);
         }
         cleanup();
+
+        // Try to make sure that the modifications below do are not within the same second as the
+        // previous sync attempt was done, otherwise the changes go undetected.
+        QThread::sleep(1);
 
         // When the file changes a retry happens immediately
         modifier.appendByte(testFileName);
--- a/test/testchunkingng.cpp
+++ b/test/testchunkingng.cpp
@@ -15,6 +15,22 @@ using namespace std::chrono_literals;
 using namespace OCC;
 
 namespace {
+
+constexpr unsigned long long operator"" _kb(unsigned long long sz)
+{
+    return sz * 1024;
+}
+
+constexpr unsigned long long operator"" _mb(unsigned long long sz)
+{
+    return operator"" _kb(sz) * 1024;
+}
+
+constexpr unsigned long long operator"" _gb(unsigned long long sz)
+{
+    return operator"" _mb(sz) * 1024;
+}
+
 /* Upload a 1/3 of a file of given size.
  * fakeFolder needs to be synchronized */
 void partialUpload(FakeFolder &fakeFolder, const QString &name, qint64 size)
@@ -52,7 +68,8 @@ void setChunkSize(SyncEngine &engine, qint64 size)
     options._minChunkSize = size;
     engine.setSyncOptions(options);
 }
-}
+
+} // anonymous namespace
 
 class TestChunkingNG : public QObject
 {
@@ -367,9 +384,9 @@ private slots:
         QCOMPARE(fakeFolder.uploadState().children.count(), 1);
         auto chunkingId = fakeFolder.uploadState().children.first().name;
 
-
-        fakeFolder.localModifier().setContents(QStringLiteral("A/a0"), 'B');
-        fakeFolder.localModifier().appendByte(QStringLiteral("A/a0"));
+        const auto a0size = fakeFolder.currentLocalState().find("A/a0")->contentSize;
+        fakeFolder.localModifier().setContents(QStringLiteral("A/a0"), a0size, 'B');
+        fakeFolder.localModifier().appendByte(QStringLiteral("A/a0"), 'B');
 
         QVERIFY(fakeFolder.syncOnce());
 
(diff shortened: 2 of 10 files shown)


test/testchunkingng.cpp
```diff
@@ -461,17 +463,17 @@ private slots:
     void testModifyLocalFileWhileUploading() {
 
         FakeFolder fakeFolder{FileInfo::A12_B12_C12_S12()};
-        const int size = 10 * 1000 * 1000; // 100 MB
+        const int size = 10 * 1000 * 1000; // 10 MB
```
TheOneRing, line 466: We should really introduce file size literals.

test/testchunkingng.cpp
```diff
@@ -600,7 +602,7 @@ private slots:
     // Test uploading large files (2.5GiB)
     void testVeryBigFiles() {
         FakeFolder fakeFolder{FileInfo::A12_B12_C12_S12()};
-        const qint64 size = 2.5 * 1024 * 1024 * 1024; // 2.5 GiB
+        const qint64 size = 1.5 * 1024 * 1024 * 1024; // 1.5 GiB
```
TheOneRing, line 605: Still bug but not a ssd stress test? Or what was the motivation.

test/testpermissions.cpp
```diff
@@ -313,15 +313,15 @@ private slots:
...
+        fakeFolder.localModifier().setContents(QStringLiteral("readonlyDirectory_PERM_M_/cannotBeModified_PERM_DVN_.data"), FileModifier::DefaultFileSize, 's');
         //do the sync
         applyPermissionsFromName(fakeFolder.remoteModifier());
         QVERIFY(fakeFolder.syncOnce());
         assertCsyncJournalOk(fakeFolder.syncJournal());
 
         QThread::sleep(1); // make sure changes have different mtime
         editReadOnly(QStringLiteral("readonlyDirectory_PERM_M_/cannotBeModified_PERM_DVN_.data"));
-        fakeFolder.localModifier().setContents(QStringLiteral("readonlyDirectory_PERM_M_/cannotBeModified_PERM_DVN_.data"), 'd');
+        fakeFolder.localModifier().setContents("readonlyDirectory_PERM_M_/cannotBeModified_PERM_DVN_.data", FileModifier::DefaultFileSize, 'd');
```
TheOneRing, line 324: Why is this no longer a string literal?

test/testutils/syncenginetestutils.cpp
```diff
@@ -152,21 +141,25 @@ void FileInfo::remove(const QString &relativePath)
...
+    file->contentSize = newSize;
+    if (!file->isDehydratedPlaceholder) {
+        file->fileSize = newSize;
+    }
 }
 
 void FileInfo::appendByte(const QString &relativePath, char contentChar)
 {
-    Q_UNUSED(contentChar);
+    Q_UNUSED(contentChar)
```
TheOneRing, line 162: Why is it part of the function signature in the first place?

test/testchunkingng.cpp
```diff
@@ -140,7 +141,7 @@ private slots:
         fakeFolder.setServerOverride([&](QNetworkAccessManager::Operation op, const QNetworkRequest &request, QIODevice *) -> QNetworkReply * {
             if (op == QNetworkAccessManager::PutOperation) {
                 // Test that we properly resuming, not resending the first chunk
-                Q_ASSERT(request.rawHeader("OC-Chunk-Offset").toLongLong() >= firstChunk.contentSize);
+                Q_ASSERT(request.rawHeader("OC-Chunk-Offset").toULongLong() >= firstChunk.contentSize);
```
fmoc, line 144: Are we sure this value is sent unsigned by the server?

test/testchunkingng.cpp
```diff
@@ -373,9 +374,9 @@ private slots:
         QCOMPARE(fakeFolder.uploadState().children.count(), 1);
         auto chunkingId = fakeFolder.uploadState().children.first().name;
 
-
-        fakeFolder.localModifier().setContents(QStringLiteral("A/a0"), 'B');
-        fakeFolder.localModifier().appendByte(QStringLiteral("A/a0"));
+        auto a0size = fakeFolder.currentLocalState().find("A/a0")->contentSize;
```
fmoc, line 377: Could be `const` as far as I can see.

test/testchunkingng.cpp
fmoc: It's a good idea in general to explain why this size is ideal, since right now it looks like an arbitrary choice.

test/testutils/syncenginetestutils.cpp
```diff
@@ -1106,9 +1090,10 @@ void FakeFolder::fromDisk(QDir &dir, FileInfo &templateFi)
             fi.setLastModified(diskChild.lastModified());
             if (fi.isDehydratedPlaceholder) {
                 fi.contentChar = '\0';
+                fi.contentSize = 0;
             } else {
                 QFile f { diskChild.filePath() };
-                f.open(QFile::ReadOnly);
+                OC_ENFORCE(f.open(QFile::ReadOnly));
```
fmoc, line 1096: A regular assert should do. There is no point in building unit tests in release mode anyway, all our asserts in the business code would never run.

fmoc (dismissed):

test/testutils/syncenginetestutils.cpp
erikjv: because the function signature of the method in the parent class has it (`FileModifier::appendByte`). The storage on the server side only stores 1 byte, with the assumption that files consist of `contentSize` bytes, all the same, so storing 1 `contentChar` is enough.

I'm not sure what to think of this "design", it's there. See also the FIXMEs in the class definition.

test/testchunkingng.cpp
erikjv: I'd hope so... I could assert it if needed...

test/testutils/syncenginetestutils.cpp
erikjv: We're doing a release build, and asserts are off.

test/testblacklist.cpp
```diff
@@ -132,6 +133,9 @@ private slots:
         }
         cleanup();
 
+        // Make sure that the modifications below do are not within the same second as the previous sync attempt was done, otherwise the changes go undetected.
```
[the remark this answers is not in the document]
erikjv: done in the last force-push

erikjv:
I think I addressed all comments in the last version of this PR.

test/testchunkingng.cpp
TheOneRing: The header is set by our code and its a `qint64` https://github.com/owncloud/client/blob/e93cda468efd09ffe5f0b36f4cb1d27a57f586bf/src/libsync/propagateupload.h#L343-L343

TheOneRing: https://github.com/owncloud/client/pull/10129

test/testpermissions.cpp
erikjv: done

test/testchunkingng.cpp
erikjv: added comment (2.5G > INT_MAX)

test/testchunkingng.cpp
erikjv: rebased on top of that PR, so now it's gone

test/testchunkingng.cpp
```diff
@@ -15,6 +15,22 @@ using namespace std::chrono_literals;
...
+{
+    return sz * 1024;
+}
+
+constexpr unsigned long long operator"" _mb(unsigned long long sz)
+{
+    return operator"" _kb(sz) * 1024;
+}
+
+constexpr unsigned long long operator"" _gb(unsigned long long sz)
```
TheOneRing, line 29: ❤️

TheOneRing (approved):

PR #10103, opened 2022-09-12
Handle propfind replies with status 425 as valid.

dragotin:
With the async upload some entries might have a status code 425 which is a valid code.


dragotin:
There might be more places where changes are needed.

PR #10114, opened 2022-09-14
Run unit tests as non-root user, second attempt

fmoc:


fmoc:
The issue from the last attempt has been resolved successfully.

TheOneRing (changes requested):
Tests are still not executed, the whole step fails

PR #10117, opened 2022-09-15
Ci: Run clazy standalone

TheOneRing:


DeepDiver1975:
clazy is executed - but job does not fail .... needs adjustments ...

DeepDiver1975:
Maybe use https://github.com/marketplace/actions/clazy-action ....

PR #10119, opened 2022-09-16
Work/more launch on startup

erikjv:


src/common/utility_mac.mm
```diff
@@ -61,6 +61,25 @@
 static Result<void, QString> writePlistToFile(NSString *plistFile, NSDictionary *plist)
 {
     NSError *error = nil;
+
+    // Check if the directory exists. On a fresh installation for example, the directory does not
+    // exist, so writing the plist file below will fail.
+    QDir plistDir = QFileInfo(QString::fromNSString(plistFile)).dir();
```
TheOneRing, line 67: Why the extra step through QFileInfo?

src/common/utility_unix.cpp
```diff
@@ -111,6 +111,10 @@ void Utility::setLaunchOnStartup(const QString &appName, const QString &guiName,
     }
 }
 
+void Utility::migrateLaunchOnStartup()
```
TheOneRing, line 114: Do we need to make this function publically available? 
If not we won't need to implement it on the other platfomrs.

src/common/utility_mac.mm
erikjv: To get the directory from the file. The input is something like `<homedir>/Library/LaunchAgents/foo.plist`, and I want to check if the `LaunchAgents` dir exists. Is there a better way to do that?

src/common/utility_unix.cpp
erikjv: Hmm, I'll check if I can put this in the `Platform` start-up code.

erikjv: The best I can do is to call the code from `Platform::create()` (that would hide it for non-mac platforms), and call that as follows:
```c++
auto platform = Platform::create(APPLICATION_REV_DOMAIN);
```
That is because the migration code is in `src/common`, and doesn't have the app domain available. 

What's your opinion?

src/common/utility_mac.mm
TheOneRing: Ah yes

src/common/utility_unix.cpp
TheOneRing: Sounds like a good solution.

erikjv: @TheOneRing fixed

src/common/utility.h
```diff
@@ -68,9 +69,19 @@ OCSYNC_EXPORT Q_DECLARE_LOGGING_CATEGORY(lcUtility)
...
+    /**
+     * @brief Return if launch on startup is enabled for the current user.
+     * @param appName the name of the application (because of branding)
+     */
     OCSYNC_EXPORT bool hasLaunchOnStartup(const QString &appName);
     OCSYNC_EXPORT void setLaunchOnStartup(const QString &appName, const QString &guiName, bool launch);
 
+#ifdef Q_OS_MAC
+    /// Migrate launch on startup (only used on macOS).
+    OCSYNC_EXPORT void migrateLaunchOnStartup(const QString &appDomain);
```
TheOneRing, line 82: Can be moved to the mac platform I assume?

erikjv: Will do that

erikjv:
**For testing:** install an older version of the client (e.g. 2.10), and turn auto-start on. Then install/migrate to 3.0, turn auto-start off, and check that the client won't auto-start at login.

PR #10137, opened 2022-09-23
Normalise spaces sync folder name before creation

TheOneRing:
image
```
ls "C:\Users\hanna\ownCloud - Katherine Johnson@ocis.owncloud.test15"

    Directory: C:\Users\hanna\ownCloud - Katherine Johnson@ocis.owncloud.test15

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
la--s          23/09/2022    11:32                émgW_;____時as
la--s          23/09/2022    11:28                Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo
                                                  consequat. Duis aute
la--s          23/09/2022    11:30                Lost in space
```


ownclouders:
Results for <strong>GUI-tests</strong> https://drone.owncloud.com/owncloud/client/13225/6/1

<details>


</details>

src/common/filesystembase.cpp
```diff
@@ -550,6 +560,21 @@ bool FileSystem::isChildPathOf(const QString &child, const QString &parent)
...
+QString OCSYNC_EXPORT FileSystem::createPortableFileName(const QFileInfo &path, int reservedSize)
+{
+    // remove leading and trailing whitespace
+    QString tmp = path.fileName().trimmed();
+    // limit size to fileNameMaxC
+    tmp.resize(std::min(tmp.size(), fileNameMaxC - reservedSize));
+    // remove eventual trailing whitespace after the resize
+    tmp = tmp.trimmed();
+    // replace non portable characters
+    for (auto c : IllegalFilenameCharsWindows) {
```
fmoc, line 573: What about other OSes?

Also, I think it makes more sense to use an allowlist than an excludelist approach. Is that a viable option?

fmoc (changes requested):

src/common/filesystembase.cpp
TheOneRing: The idea of the function is to create a portable file name, which means its the same on all platforms.

fmoc: If it applies to all platforms, you may want to change the name of that variable.

TheOneRing: The var is a list of characters that are illegal on Windows, it is used in different locations.
In this function it is used to create a portable file name.

PR #10176, opened 2022-10-10
Don't create redundant content widgets

fmoc:
The widgets are pre-initialized by the .ui file already, there is no need for initializing them manually.

### DIFF
--- a/src/gui/loginrequireddialog/loginrequireddialog.cpp
+++ b/src/gui/loginrequireddialog/loginrequireddialog.cpp
@@ -24,7 +24,7 @@
 
 namespace OCC {
 
-LoginRequiredDialog::LoginRequiredDialog(AbstractLoginWidget *contentWidget, QWidget *parent)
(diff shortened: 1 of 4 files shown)


src/gui/loginrequireddialog/loginrequireddialog.cpp
```diff
@@ -41,8 +41,16 @@ LoginRequiredDialog::LoginRequiredDialog(AbstractLoginWidget *contentWidget, QWi
 
     // using a stacked widget appears to work better than a plain widget
     // we do this in the setup wizard as well
-    _ui->contentWidget->addWidget(contentWidget);
-    _ui->contentWidget->setCurrentWidget(contentWidget);
+    _ui->contentWidget->setCurrentWidget([this, mode]() -> QWidget * {
```
TheOneRing, line 44: bit late but you can use the enum as index for a stack widget.

TheAssassin: Hmm, how would I make sure the stacked widget index and the enum are in sync?

PR #10181, opened 2022-10-12
Merge PropfindJob with LsColJob

TheOneRing:


src/gui/quotainfo.cpp
```diff
@@ -99,14 +99,14 @@ void QuotaInfo::slotCheckQuota()
...
+    _job = new PropfindJob(account, account->davUrl(), quotaBaseFolder(), 0, this);
     _job->setProperties({ QByteArrayLiteral("quota-available-bytes"), QByteArrayLiteral("quota-used-bytes") });
-    connect(_job.data(), &PropfindJob::result, this, &QuotaInfo::slotUpdateLastQuota);
+    connect(_job.data(), &PropfindJob::directoryListingIterated, this, &QuotaInfo::slotUpdateLastQuota);
     connect(_job.data(), &AbstractNetworkJob::networkError, this, &QuotaInfo::slotRequestFailed);
     _job->start();
 }
 
-void QuotaInfo::slotUpdateLastQuota(const QMap<QString, QString> &result)
+void QuotaInfo::slotUpdateLastQuota(const QString &, const QMap<QString, QString> &result)
```
fmoc, line 109: If the string isn't used, why add it to the signature?

src/gui/sharedialog.h
```diff
@@ -52,7 +52,7 @@ class ShareDialog : public QDialog
     QSize minimumSizeHint() const override;
 
 private slots:
-    void slotPropfindReceived(const QMap<QString, QString> &result);
+    void slotPropfindReceived(const QString &, const QMap<QString, QString> &result);
```
fmoc, line 55: Wouldn't it be better style to adapt the connections?

Alternatively we might add a second signature with a FIXME that calls the original one.

src/libsync/networkjobs.cpp
```diff
@@ -294,73 +290,35 @@ bool LsColXMLParser::parse(const QByteArray &xml, QHash<QString, qint64> *sizes,
...
-LsColJob::LsColJob(AccountPtr account, const QUrl &url, const QString &path, QObject *parent)
+PropfindJob::PropfindJob(AccountPtr account, const QUrl &url, const QString &path, int depth, QObject *parent)
     : AbstractNetworkJob(account, url, path, parent)
+    , _depth(depth)
 {
     // Always have a higher priority than the propagator because we use this from the UI
     // and really want this to be done first (no matter what internal scheduling QNAM uses).
     // Also possibly useful for avoiding false timeouts.
     setPriority(QNetworkRequest::HighPriority);
+    Q_ASSERT(depth == 0 || depth == 1);
```
fmoc, line 301: This was introduced in 3b3531f77753d1c72909d057e0f79058c0044ae7, but I don't see why it's important. You should probably add a comment.

src/libsync/networkjobs.h
```diff
@@ -75,11 +75,11 @@ class OWNCLOUDSYNC_EXPORT LsColXMLParser : public QObject
     void finishedWithoutError();
 };
 
-class OWNCLOUDSYNC_EXPORT LsColJob : public AbstractNetworkJob
+class OWNCLOUDSYNC_EXPORT PropfindJob : public AbstractNetworkJob
 {
     Q_OBJECT
 public:
-    explicit LsColJob(AccountPtr account, const QUrl &url, const QString &path, QObject *parent = nullptr);
+    explicit PropfindJob(AccountPtr account, const QUrl &url, const QString &path, int depth = 1, QObject *parent = nullptr);
```
fmoc, line 82: Since we allow only two values, it might be easier (and safer) to introduce one or two named constructors.

fmoc (changes requested):

src/gui/quotainfo.cpp
TheOneRing: Because thats how connects work

src/gui/sharedialog.h
TheOneRing: An additional signal would need to be emitted.
How would you change the connect?

src/gui/quotainfo.cpp
fmoc: https://github.com/owncloud/client/pull/10181#discussion_r993405244

ownclouders:
Results for <strong>GUI-tests</strong> https://drone.owncloud.com/owncloud/client/13389/6/1

<details>


</details>

src/gui/sharedialog.h
fmoc: There's the option of connecting a lambda which calls the slot correctly. But that's less efficient.

fmoc (approved):
Approving. Please make sure it builds, though.

PR #10199, opened 2022-10-19
 Implement a test for chunking ng fail due to a lock

TheOneRing:
The unit test is supposed to fail

PR #10200, opened 2022-10-19
Include saved uploads in partial sync

TheOneRing:


TheOneRing:
The unit test was implemented in https://github.com/owncloud/client/pull/10199

src/common/syncjournaldb.cpp
```diff
@@ -1407,31 +1407,63 @@ SyncJournalDb::UploadInfo SyncJournalDb::getUploadInfo(const QString &file)
 {
     QMutexLocker locker(&_mutex);
 
+
+    if (!checkConnect()) {
+        return {};
```
fmoc, line 1412: Should we log this case? It looks like this case should not occur really.

src/libsync/syncengine.h
```diff
@@ -300,6 +300,8 @@ private slots:
     /** The kind of local discovery the last sync run used */
     LocalDiscoveryStyle _lastLocalDiscoveryStyle = LocalDiscoveryStyle::FilesystemOnly;
     LocalDiscoveryStyle _localDiscoveryStyle = LocalDiscoveryStyle::FilesystemOnly;
+
+    // requires to be ordered
```
fmoc, line 304:
```suggestion
    // must be ordered
```

fmoc (changes requested):

fmoc (approved):

PR #10208, opened 2022-10-21
[gui-tests] Remove unnecessary waits

saw-jan:
This PR has the following changes:
- Remove waits where there is no need
- Wisely wait for some actions to finish (if needed)

Previously, there were waits before any action (`create`, `delete`, `rename`, and `move` file/folder). The wait was needed but only before the first action so there was no need for other waits after the first wait. Those waits accumulate and cause test cases to run for a long. Thus, this PR has refactored the test code so that there is a wait only before performing the first sync operation and other operations can be performed without waits.

Similarly, we have explicitly waited for `minSyncTimeout` i.e. `5 secs` for the AUT to be terminated thus causing unnecessary waits and long tests run. In this PR, I have checked the availability of the AUT process, determined whether it is running or not, and waited for its termination. This prevents the use of unnecessary waits.

### Motivation
Reduce tests runtime

### Results
Before: (https://drone.owncloud.com/owncloud/client/13483)
Screenshot from 2022-10-24 11-17-20

After:
Screenshot from 2022-10-24 11-17-11

### DIFF
--- a/test/gui/shared/scripts/bdd_hooks.py
+++ b/test/gui/shared/scripts/bdd_hooks.py
@@ -158,6 +158,25 @@ def scenarioFailed():
     )
 
 
+def isAppKilled(pid):
+    if os.path.isdir('/proc/{}'.format(pid)):
+        # process is still running
+        # wait 100ms before checking again
+        snooze(0.1)
+        return False
+    return True
+
+
+def waitUntilAppIsKilled(context, pid=0):
+    timeout = context.userData['minSyncTimeout'] * 1000
+    killed = waitFor(
+        lambda: isAppKilled(pid),
+         timeout,
+    )
+    if not killed:
+        test.log("Application was not terminated within {} milliseconds".format(timeout))
+
+
 @OnScenarioEnd
 def hook(context):
     global socketConnect, socket_messages, waitedAfterSync, previousFailResultCount, previousErrorResultCount, waitedAfterSync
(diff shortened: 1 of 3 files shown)


test/gui/shared/scripts/bdd_hooks.py
```diff
@@ -30,6 +30,10 @@
 # socket messages
 socket_messages = []
 
+# whether waited for context.userData['touchTimeout'] seconds or not
```
saw-jan, line 33:
```suggestion
```

ownclouders:
Results for <strong>GUI-tests</strong> https://drone.owncloud.com/owncloud/client/13482/6/1

:boom: The GUI tests failed.

GUI Logs: https://cache.owncloud.com/public/owncloud/client/13482/guiReportUpload/index.html

<details>

Server Logs: https://cache.owncloud.com/public/owncloud/client/13482/guiReportUpload/serverlog.log

</details>

test/gui/shared/scripts/bdd_hooks.py
```diff
@@ -152,20 +158,40 @@ def scenarioFailed():
     )
 
 
+def isAppKilled(pid):
+    if os.path.isdir('/proc/{}'.format(pid)):
```
saw-jan, line 162: Running processes will have a folder created with its pid in this location

individual-it (approved):

PR #10235, opened 2022-11-04
Work/remove unused window titles

erikjv:


TheOneRing:
Hmm could that still be handy during development? Should we maybe just flag them as untranslatable?

erikjv:
Personally I don't see a use, but I can be persuaded otherwise :-)

PR #10237, opened 2022-11-07
Fix two years old off by one

TheOneRing:


dragotin (approved):
I would have an opinion about usage of ?:-operators and their effect on clean code, but maybe not on Monday morning ;-)

src/libsync/theme.cpp
```diff
@@ -410,7 +410,7 @@ QString Theme::aboutVersions(Theme::VersionFormat format) const
         Q_UNREACHABLE();
     }();
     const QString qtVersion = QString::fromUtf8(qVersion());
-    const QString qtVersionString = (QLatin1String(QT_VERSION_STR) == qtVersion ? qtVersion : QCoreApplication::translate("ownCloudTheme::qtVer", "%1 (Built against Qt %1)").arg(qtVersion, QStringLiteral(QT_VERSION_STR)));
+    const QString qtVersionString = (QLatin1String(QT_VERSION_STR) == qtVersion ? qtVersion : QCoreApplication::translate("ownCloudTheme::qtVer", "%1 (Built against Qt %2)").arg(qtVersion, QStringLiteral(QT_VERSION_STR)));
```
fmoc, line 413: This issue might have been introduced because this line is way too long and contains too much logic. Looks fine to me, but I'd like to advocate extracting the `if`.

fmoc (approved):

erikjv (approved):

PR #10240, opened 2022-11-07
Hide disabled spaces in spaces browser

### LINKED ISSUE #10173: [Spaces] Exclude disabled Spaces from initial sync
### Pre-submission Checks

### Feature description

Spaces can be "disabled". At least at initial sync, we should exclude them. (no concept for Spaces that get disabled or deleted later yet)

### Proposed solution

Exclude disabled Spaces from initial sync.

_No response_

### Additional context

_No response_

fmoc:
Fixes #10173.


TheOneRing:
The issue is about not adding disabled spaces. Your change is certainly an improvement, but it missed the point.

TheOneRing (changes requested):
^

fmoc:
Yeah, I got this wrong. Please approve the PR if it is fine, though (and maybe merge it right away). I'll send a follow-up PR.

TheOneRing:
Hmm I think we should filter before we reach the model?

TheOneRing (changes requested):
I think we should filter before we reach the model?

TheOneRing:
> I think we should filter before we reach the model?

I mean in the drives class, that weay you can use the same code when actually handling the issue.

fmoc:
Ok, I'll go for that solution next then.

fmoc:
May be merged only after #10244, therefore converted to draft.

src/gui/models/models.h
```diff
@@ -28,7 +28,8 @@ namespace Models {
 
     enum DataRoles {
         UnderlyingDataRole = Qt::UserRole + 100,
-        StringFormatWidthRole // The width for a cvs formated column
+        StringFormatWidthRole, // The width for a cvs formatted column
+        FilterRole,
```
TheOneRing, line 32: // role must return bool

src/gui/spaces/spacesbrowser.cpp
```diff
@@ -29,15 +29,31 @@
...
 SpacesBrowser::SpacesBrowser(QWidget *parent)
     : QWidget(parent)
     , ui(new Ui::SpacesBrowser)
 {
     ui->setupUi(this);
     _model = new SpacesModel(this);
 
-    auto *sortModel = new OCC::Models::WeightedQSortFilterProxyModel(this);
-    sortModel->setSourceModel(_model);
+    auto *filterModel = new FilteringProxyModel(this);
```
TheOneRing, line 52: setFilterRole
setFilterKeyColumn

src/gui/spaces/spacesbrowser.cpp
```diff
@@ -29,15 +29,31 @@
 
 using namespace OCC::Spaces;
 
+class FilteringProxyModel : public QSortFilterProxyModel
+{
+    using QSortFilterProxyModel::QSortFilterProxyModel;
+
+protected:
```
TheOneRing, line 36: // accepting boolean for the filter role

TheOneRing (approved):

Issue #10232, opened 2022-11-04
[QA] ''You're all set!'' window can be closed without any warning


HanaGemela:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the QA issue

User wants to add a new account but after closing the ''You're all set!'' window, the account hasn't been added

### Steps to reproduce the issue

1. Click `Add account` in settings
2. Add server address
3. Login in WebUI
4. Click the Cross to close `You're all set` window

### Screenshots

<img width="719" alt="Screenshot 2022-11-04 at 11 46 33" src="https://user-images.githubusercontent.com/49001702/199956114-3d194a26-5f6d-4a68-bb1f-0d1ee9071072.png">

### Expected behavior

Either:
There's a pop-up with warning that the actions will not be saved and user is prompted to confirm his decision 

or: 
The cross behaves like the Finish button and the account is saved like the 'You're all set" description suggest

@michaelstingl please decide

### Actual behavior

The windows gets closed without a warning and the account hasn't been added

jnweiger:
Reproduced on 3.0.0-beta.1

The issue is not about having or hiding a close button. It is about a false expectation induced by "You are all set", when the user actually has not set up anything, unless he also presses the "Finish" button.

I suggest to change the text to "You are all set. Press 'Finish' to start syncing." :see_no_evil:

jnweiger:
![grafik](https://user-images.githubusercontent.com/1108546/201330880-8387adcd-bdec-4482-9be1-6b3f823c06fe.png)

While working in the Advanced configuration, it is obvious, that the "You are all set!" info is premature.

TheOneRing:
Not a bug.

jnweiger:
Problems in the UX design are by definition "Works as designed" -- so, I understand, that they cannot qualify as "bug" here.
That is unfortunate, as I see no other way to address them. If we just close here, it will be forgotten and won't see improvement, right? @michaelstingl do you have an idea, how to report such issues properly?

PR #10242, opened 2022-11-07
Ask before closing the wizard

TheOneRing:
Fixes: #10232

### DIFF
--- a/src/gui/newwizard/setupwizardwindow.cpp
+++ b/src/gui/newwizard/setupwizardwindow.cpp
@@ -8,6 +8,7 @@
 #include "theme.h"
 
 #include <QLabel>
+#include <QMessageBox>
 #include <QStyleFactory>
 
 using namespace std::chrono_literals;
@@ -35,6 +36,7 @@ SetupWizardWindow::SetupWizardWindow(SettingsDialog *parent)
     , _ui(new ::Ui::SetupWizardWindow)
 {
     Utility::setModal(this);
+    setWindowFlag(Qt::WindowCloseButtonHint, false);
 
     _ui->setupUi(this);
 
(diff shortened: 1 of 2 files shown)


src/gui/newwizard/setupwizardwindow.cpp
```diff
@@ -149,6 +151,18 @@ void SetupWizardWindow::slotStartTransition()
     slotHideErrorMessageWidget();
 }
 
+void SetupWizardWindow::reject()
+{
+    auto messageBox = new QMessageBox(QMessageBox::Warning, tr("Cancel Setup"), tr("Do you really want to cancel the account setup?"), QMessageBox::Yes | QMessageBox::No, ocApp()->gui()->settingsDialog());
```
fmoc, line 156: This message doesn't make a lot of sense to me terminology wise. We consider this the "setup wizard".

src/gui/newwizard/setupwizardwindow.cpp
```diff
@@ -35,6 +36,7 @@ SetupWizardWindow::SetupWizardWindow(SettingsDialog *parent)
     , _ui(new ::Ui::SetupWizardWindow)
 {
     Utility::setModal(this);
+    setWindowFlag(Qt::WindowCloseButtonHint, false);
```
fmoc, line 39: Why do you need this? Closing the dialog via the regular exit button added by the window manager also triggers `reject`.

Apparently, on macOS one needs to combine this with `WindowSystemMenuHint`.

fmoc (changes requested):
I think we need to complement the "you're all set" message with a "press finish to actually finish" kind of message.

src/gui/newwizard/setupwizardwindow.cpp
TheOneRing: Yes but the wizard does set up an account?

src/gui/newwizard/setupwizardwindow.cpp
TheOneRing: This disables the X on Windows, I think we give better guidance if we force the user to use the close button?

TheOneRing:
> I think we need to complement the "you're all set" message with a "press finish to actually finish" kind of message.

I expect 99% of the users will figure it out and for the rest we have the cancel dialog.

src/gui/newwizard/setupwizardwindow.cpp
fmoc: You mean the cancel button? I'm not sure... I like to use either option. But I'm a dev.

> Apparently, on macOS one needs to combine this with WindowSystemMenuHint.

My bad, I guess with these sheets one doesn't have to care about this anyway.

Issue #10184, opened 2022-10-13
Update icon location in desktop.ini on folder initialisation


TheOneRing:
The icon path is only set when the folder is created an not updated.
This is problematic when the client location (exe contains the icon) changes.

PR #10245, opened 2022-11-07
Update the folder icon on every run

TheOneRing:
Fixes: #10184

### DIFF
--- a/changelog/unreleased/10184
+++ b/changelog/unreleased/10184
@@ -1,0 +1,6 @@
+Change: Windows: Update the folder icon on every start
+
+The ownCloud installation path might have changed and ths the Desktop.ini pointed to the wrong path.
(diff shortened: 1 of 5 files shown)


changelog/unreleased/10184
```diff
@@ -0,0 +1,6 @@
+Change: Windows: Update the folder icon on every start
+
+The ownCloud installation path might have changed and ths the Desktop.ini pointed to the wrong path.
```
fmoc, line 3:
```suggestion
The ownCloud installation path might have changed, causing the desktop.ini to point at the wrong path.
```

fmoc (approved):

fmoc (changes requested):
See above.

fmoc (approved):

PR #10276, opened 2022-11-11
Fix retry for uploads

TheOneRing:


src/libsync/creds/httpcredentials.cpp
```diff
@@ -233,10 +233,13 @@ void HttpCredentials::fetchFromKeychainHelper()
 
 bool HttpCredentials::stillValid(QNetworkReply *reply)
 {
-    // The function is called in order to determine whether we need to ask the user for a password
-    // if we are using oaut we already started a refresh in slotAuthentication.
-    // If the refresh fails we will emit authenticationFailed ourself.
     if (isUsingOAuth()) {
+        // The function is called in order to determine whether we need to ask the user for a password
+        // if we are using oaut we already started a refresh in slotAuthentication, at least in theory, ensure the auth is started.
```
fmoc, line 238:
```suggestion
        // if we are using OAuth, we already started a refresh in slotAuthentication, at least in theory, ensure the auth is started.
```

I'd leave a TODO that this is more or less guesswork.

src/libsync/creds/httpcredentials.cpp
```diff
@@ -233,10 +233,13 @@ void HttpCredentials::fetchFromKeychainHelper()
...
+        // If the refresh fails we will emit authenticationFailed ourself.
```
fmoc, line 239:
```suggestion
        // If the refresh fails, we are going to emit authenticationFailed ourselves
```

fmoc (changes requested):

fmoc (approved):

PR #10279, opened 2022-11-14
macOS: Fix restart-after-quit

erikjv:
Instead of using a dictionary for the value of KeepAlive, setting it to @NO will keep launchd from relaunching the client.


fmoc (approved):
Looks good to me code wise.

Issue #10288, opened 2022-11-16
Sqlite log broken


TheOneRing:
The new sqlite query estimation contains too many entries and appears to long

PR #10290, opened 2022-11-16
Fix sqlite log

TheOneRing:
Clear the debug string, don't append to previous values.

Fixes: #10288


fmoc (approved):
Seems like a really, really good idea.

Issue #10260, opened 2022-11-10
Squish tests are passing OWNCLOUD_CORE_DUMP=1 as an arg


TheOneRing:
The squish tests are passing OWNCLOUD_CORE_DUMP=1 as a command line argument which is not supported.
OWNCLOUD_CORE_DUMP is an env var.

PR #10297, opened 2022-11-18
[gui-tests] Remove unsupported owncloud arg

saw-jan:
### Related issues
Closes https://github.com/owncloud/client/issues/10260


.drone.star
```diff
@@ -225,8 +226,13 @@ def gui_test_pipeline(ctx, trigger = {}, filterTags = [], server_version = oc10_
...
 
-def build_client(c_compiler, cxx_compiler, build_type, generator, build_command, image = OC_CI_CLIENT):
-    cmake_options = '-G"%s" -DCMAKE_C_COMPILER="%s" -DCMAKE_CXX_COMPILER="%s" -DCMAKE_BUILD_TYPE="%s" -DBUILD_TESTING=1 -DWITH_LIBCLOUDPROVIDERS=ON' % (generator, c_compiler, cxx_compiler, build_type)
+def build_client(c_compiler, cxx_compiler, build_type, generator, build_command, image = OC_CI_CLIENT, ctest = True):
+    cmake_options = '-G"%s" -DCMAKE_C_COMPILER="%s" -DCMAKE_CXX_COMPILER="%s" -DCMAKE_BUILD_TYPE="%s" -DWITH_LIBCLOUDPROVIDERS=ON' % (generator, c_compiler, cxx_compiler, build_type)
+
+    if ctest:
+        cmake_options += " -DBUILD_TESTING=1"
+    else:
+        cmake_options += " -DBUILD_TESTING=0"
```
saw-jan, line 235: not building ctest files make the build process a bit faster. And I think those test files are not required to build while testing the GUI

CC @TheOneRing

.drone.star
```diff
@@ -285,6 +291,7 @@ def gui_tests(squish_parameters = "", server_type = "oc10"):
             "SQUISH_PARAMETERS": squish_parameters,
             "STACKTRACE_FILE": "%s/stacktrace.log" % dir["guiTestReport"],
             "PLAYWRIGHT_BROWSERS_PATH": "%s/.playwright" % dir["base"],
+            "OWNCLOUD_CORE_DUMP": 1,
```
saw-jan, line 294: @TheOneRing Can I use `OWNCLOUD_CORE_DUMP` as env like this?

.drone.star
TheOneRing: Sure, good idea.
Nitpick: please use ON OFF.

fmoc: Right, as long as we don't run the tests anyway, there is no reason to build them. They're tested on GitHub actions.

.drone.star
TheOneRing: Yes it is an env var and this should work.

fmoc: The value seems alright, even though env vars are usually strings. As long as it runs (I assume Python just implicitly converts this to `str` where necessary) this is fine.

saw-jan: yeah, env values will be treated as strings

TheOneRing (approved):

saw-jan:
> I think we should target 3.0

Changed the target to `3.0` branch

individual-it (approved):

TheOneRing (approved):

PR #10301, opened 2022-11-22
Fix OAuth retry

### LINKED ISSUE #10300: Cannot close OAuth login dialog after retry
### Pre-submission Checks

### Describe the bug

When using the retry feature to retry after an error, the dialog is not closed when logging in successfully.

Note that failed logins correctly lead to a new error message as expected. This only affects successful logins after at least one retry.

### Expected behavior

The dialog should close automatically when logged in successfully.

1. Connect to account using OAuth/OIDC
1. Log out of accout
1. Log in again
1. In the dialog, make sure you need to retry (e.g., by supplying incomplete credentials on the server)
1. Log in with the correct username

### Screenshots

_No response_

### Logs

_No response_

### Client version number

Current 3.0 branch

_No response_

_No response_

_No response_

### Server information

_No response_

### Additional context

Follow-up from #9772.

fmoc:
In one of the slots for asked(), we need to be able to distinguish between "error" and "retry" situations.

I'm sure there's a more elegant way, but I don't see it currently.

Also cleans up a superfluous variable in HttpCredentials.

Fixes #10300.

### DIFF
--- a/src/gui/accountstate.cpp
+++ b/src/gui/accountstate.cpp
@@ -494,7 +494,7 @@ void AccountState::slotCredentialsAsked(AbstractCredentials *credentials)
     case AbstractCredentials::ReadyState::Error:
         setState(SignedOut);
         return;
-    default:
+    case AbstractCredentials::ReadyState::Ready:
         break;
     }
 
--- a/src/gui/accountstate.cpp
+++ b/src/gui/accountstate.cpp
@@ -454,7 +454,7 @@ void AccountState::slotInvalidCredentials()
         qCInfo(lcAccountState) << "Invalid credentials for" << _account->url().toString();
 
         _waitingForNewCredentials = true;
-        if (account()->credentials()->ready()) {
+        if (account()->credentials()->ready() != AbstractCredentials::ReadyState::Ready) {
             account()->credentials()->invalidateToken();
         }
         if (auto creds = qobject_cast<HttpCredentials *>(account()->credentials())) {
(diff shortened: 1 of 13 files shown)


ownclouders:
Results for <strong>GUI-tests-oc10</strong> https://drone.owncloud.com/owncloud/client/13743/6/1

<details>


</details>

src/libsync/creds/abstractcredentials.h
```diff
@@ -33,6 +33,13 @@ class OWNCLOUDSYNC_EXPORT AbstractCredentials : public QObject
     Q_OBJECT
 
 public:
+    enum class ReadyState {
+        Ready,
+        Retry,
```
TheOneRing, line 38: Maybe Error and FatalError?
Retry implies that we already decided to rety, but thats a user choice.

src/gui/accountstate.cpp
```diff
@@ -488,10 +488,14 @@ void AccountState::slotCredentialsAsked(AbstractCredentials *credentials)
...
 
-    if (!credentials->ready()) {
-        // User canceled the connection or did not give a password
+    switch (credentials->ready()) {
+    case AbstractCredentials::ReadyState::Retry:
+        return;
+    case AbstractCredentials::ReadyState::Error:
         setState(SignedOut);
         return;
+    default:
```
TheOneRing, line 497: Don't add default case. We wan't the compiler to warn when a new state is added.

src/libsync/creds/abstractcredentials.h
fmoc: This differentiation already shows how bad the naming is, and also highlights that this change doesn't really make sense in this context. I mean, what kind of error can be so fatal that we can't restart the `OAuth` process and actually need to sign out the user?

TheOneRing: True I mean with basic out you should also be able to retry

fmoc:
Superseded by #10305.

PR #10306, opened 2022-11-22
Handle RemoteHostClosedError as retry reason

TheOneRing:
Fixes: #5066


TheOneRing:
See: https://github.com/owncloud/ocis/issues/5066#issuecomment-1326317727

PR #10309, opened 2022-11-23
Fix encoding of Gtk bookmarks

### LINKED ISSUE #10280: Escape Gtk 3.0 bookmarks correctly
### Pre-submission Checks

### Describe the bug

Currently, our Gtk 3.0 bookmark generation does not escape new entries correctly. This causes a variety of issues, as documented in #9752. On MATE, this even prevents the entries from being displayed at all.

### Expected behavior

We should URL-escape the entries we add.

Run any of the distros mentioned in #9752.

### Screenshots

_No response_

### Logs

_No response_

### Client version number

3.0.0-beta1/2 or any recent daily.

_No response_

_No response_

_No response_

### Server information

_No response_

### Additional context

Follow-up from #9752.

fmoc:
This should fix #10280. I could not find any other encoding issues other than spaces while testing on a variety of distributions an desktops. 

Unfortunately, there is little to no documentation on how this file should really be structured. We have not seen any issues other than with spaces in paths so far, though. We can add additional escaping in the future if needed.

### DIFF
--- a/src/common/utility_unix.cpp
+++ b/src/common/utility_unix.cpp
@@ -32,7 +32,7 @@ void Utility::setupFavLink(const QString &folder)
 {
     // Nautilus: add to ~/.gtk-bookmarks
     QFile gtkBookmarks(QDir::homePath() + QLatin1String("/.config/gtk-3.0/bookmarks"));
-    QByteArray folderUrl = "file://" + QUrl(folder).url(QUrl::ComponentFormattingOption::EncodeSpaces).toUtf8();
+    QByteArray folderUrl = QUrl::fromLocalFile(folder).url(QUrl::ComponentFormattingOption::EncodeSpaces).toUtf8();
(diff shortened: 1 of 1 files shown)


src/common/utility_unix.cpp
```diff
@@ -32,7 +32,7 @@ void Utility::setupFavLink(const QString &folder)
 {
     // Nautilus: add to ~/.gtk-bookmarks
     QFile gtkBookmarks(QDir::homePath() + QLatin1String("/.config/gtk-3.0/bookmarks"));
-    QByteArray folderUrl = QUrl::fromLocalFile(folder).url(QUrl::ComponentFormattingOption::EncodeSpaces).toUtf8();
+    QByteArray folderUrl = QUrl::fromLocalFile(folder).toEncoded(QUrl::ComponentFormattingOption::EncodeSpaces);
```
TheOneRing, line 35: const?

TheOneRing (approved):

PR #10310, opened 2022-11-23
Log parseables

dragotin:
Changes the log format to:
```
22-11-23 18:38:34:438 [ info sync.httplogger ]: "5dca3f29-99a7-488b-a35d-8113302e4ba8: Response: PROPFIND 207 (duration(0h, 0min, 0s, 214ms)=214ms) ht...
```
Note 
1. The year at the beginning of the line to make the time stamp standard compatible and parseable
2. The duration in milliseconds appended to the "verbose" duration for humans...

### DIFF
--- a/src/libsync/httplogger.cpp
+++ b/src/libsync/httplogger.cpp
@@ -61,7 +61,8 @@ void logHttp(const QByteArray &verb, const QString &url, const QByteArray &id, c
         if (reply->attribute(QNetworkRequest::HttpPipeliningWasUsedAttribute).toBool()) {
             stream << "Piplined,";
         }
-        stream << duration << ")";
+        const auto durationInMs = std::chrono::duration_cast<std::chrono::milliseconds>(duration);
+        stream << durationInMs.count() << "ms)";
     }
     stream << " " << url << " Header: { ";
     for (const auto &it : header) {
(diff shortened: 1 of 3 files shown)


src/libsync/httplogger.cpp
```diff
@@ -61,7 +61,8 @@ void logHttp(const QByteArray &verb, const QString &url, const QByteArray &id, c
         if (reply->attribute(QNetworkRequest::HttpPipeliningWasUsedAttribute).toBool()) {
             stream << "Piplined,";
         }
-        stream << duration << ")";
+        const auto int_ms = std::chrono::duration_cast<std::chrono::milliseconds>(duration);
+        stream << duration << "=" << int_ms.count() << "ms)";
```
fmoc, line 65:
```suggestion
        const auto durationInMs = std::chrono::duration_cast<std::chrono::milliseconds>(duration);
        stream << durationInMs.count() << "ms)";
```

Logging the duration once is more than sufficient.

fmoc (changes requested):

dragotin:
Yes, I think logging it once is sufficient, but @TheOneRing wanted to have the human readable version as well when she first checked it...

dragotin:
Thanks for pulling that to 3.0 !

PR #10323, opened 2022-11-28
Fix login retry compatibility with oC10

fmoc:
Fixes a compatibility issue with ownCloud 10. Context can be found in https://github.com/owncloud/client/issues/10300#issuecomment-1329161751.


TheOneRing:
Not a client bug, thise change would break the proper reauth flow.

PR #10339, opened 2022-12-07
Unify version string again

TheOneRing:


VERSION.cmake
```diff
@@ -17,7 +17,10 @@ set( MIRALL_VERSION "${MIRALL_VERSION_MAJOR}.${MIRALL_VERSION_MINOR}.${MIRALL_VE
 # Version with Build ID. Used in the installer
 set( MIRALL_VERSION_FULL "${MIRALL_VERSION}.${MIRALL_VERSION_BUILD}" )
 
-set( MIRALL_VERSION_STRING "${MIRALL_VERSION}${MIRALL_VERSION_SUFFIX}" )
+set( MIRALL_VERSION_STRING "${MIRALL_VERSION_FULL}" )
```
fmoc, line 20: I'd like to see a comment that mentions that we use this string for packaging, too. This way, changing or removing the variable won't result in unexpected problems within our CI infra.

fmoc (changes requested):

VERSION.cmake
fmoc: This variable doesn't quite work, since below the build number is appended another time. We either need to remove that or introduce another variable to either store the value that is assigned after this `if` clause separately for packaging, or by adding another variable that contains the build number in brackets. I think since we encode the build number into the version string already, we can just remove lines 25 ff.

fmoc (approved):
Should be good now.

HanaGemela:
@fmoc please provide steps to test

fmoc:
You can check whether the version in the "about" dialog follows the desired format of `<major>.<minor>.<patch>[.<build-number>][<suffix>]`, where parts in `[]` are optional. Our current daily build does it right: `3.1.0.9307-daily20221209` (`9307` is the build number, the `-daily...` part is the suffix)

HanaGemela:
Passed in 3.1.0.9389-daily20221215 on mac

PR #10349, opened 2022-12-12
Configure tx minimum percentage in config file

fmoc:


fmoc:
Let's merge this now so the nightly cron job can run at least once. We can then check tomorrow whether this configuration change works as intended.

Issue #10353, opened 2022-12-16
[QA] Refactoring Squish GUI tests


saw-jan:
### Describe the QA issue

More automated tests are likely to be added in the coming days. So, in order to make that easy and clean, here are some refactoring and code clean up to do:

- [x] Avoid using Squish `names` in the step def file. Instead, use methods from page objects
- [x] Avoid using assertion/comparison in the When steps
- [x] Avoid using assertion/comparison in the page objects
- [x] :information_source: Check if step definitions can be splitted
- [ ] :information_source: Check if multiple feature files can be created in a test suite (folder)

saw-jan:
> - [ ] :information_source: Check if multiple feature files can be created in a test suite (folder)

Multiple feature files can be created within a test folder but the Squish IDE will have trouble running the tests wisely. For example, while running one test scenario using IDE, the test scenario from another file also runs.
But this is not the case while running all test scenario at once and also with CLI

saw-jan:
If the scenarios grow large, the feature file can be split into different test case folders. And also the `tst_` folders can be moved into `features` folder but this requires some adjustment in the [squish](https://github.com/owncloud-ci/squish) image as well.

Todo might be:
- move all `tst_` folders inside the `features` folder
```
test/gui/
    - features/
        - tst_a/
        - tst_b/
        - ...
```
- And allow squish image to accept an env to test suite folder (e.g. `TEST_SUITE=<client-repo>/test/gui/features`)  and run test with
```bash
~/squish/bin/squishrunner --testsuite ${TEST_SUITE} ${SQUISH_PARAMETERS} --exitCodeOnFail 1
```

saw-jan:
Also, we might want to discuss if we want to rename test case and scenarios according to the regression test plan (https://github.com/owncloud/QA/blob/master/Desktop/Regression_Test_Plan_Minor_Release.md)

saw-jan:
This is not possible with the Squish IDE. So closing it here

PR #10356, opened 2022-12-19
[gui-tests] Avoid use of names in step defs but use methods from page objects (part 1)

saw-jan:
We want to avoid using `names` (UI component selector) directly in the step def files. This will help us to maintain the selectors in the page objects. Using page object methods for UI actions is the preferred way and it helps to maintain the DRY code

Part of https://github.com/owncloud/client/issues/10353


amrita-shrestha:
@saw-jan bro please adds some description:smiley_cat:

amrita-shrestha (approved):
seems good to me but how this helpful at that in description so that in future others understand this changes reason

PR #10364, opened 2022-12-20
Enable next button when pasting data using context menu, too

### LINKED ISSUE #10321: [QA] Next button not enabled after pasting the password
### Pre-submission Checks

### Describe the QA issue

Seen with ownCloud 3.0.0.9173-daily20221125

When connecting a new account the new wizard UI is used to query username and password.
While either username or password is blank, the Next button is grayed out. OK
I use a password manager, that provides my password into the paste buffer.
After pasting the password, the Next button remains disabled. BAD.

grafik

Workaround:
Type a space, and a backspace, then the button becomes enabled.

grafik

.

### Screenshots

.

### Expected behavior

Pasting should also enable the button.

### Actual behavior

It isn't.

fmoc:
This removes the now-obsolete handling via the event filter, which doesn't need to run that often anyway.

Closes #10321.

### DIFF
--- a/src/gui/newwizard/setupwizardwindow.cpp
+++ b/src/gui/newwizard/setupwizardwindow.cpp
@@ -130,6 +130,8 @@ void SetupWizardWindow::displayPage(AbstractSetupWizardPage *page, SetupWizardSt
     // bring to front if necessary
     ownCloudGui::raiseDialog(this);
 
+    connect(_currentPage, &AbstractSetupWizardPage::contentChanged, this, &SetupWizardWindow::slotUpdateNextButton);
+
     QTimer::singleShot(0, [this]() {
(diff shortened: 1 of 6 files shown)


src/gui/newwizard/setupwizardwindow.cpp
```diff
@@ -244,7 +236,9 @@ SetupWizardWindow::~SetupWizardWindow() noexcept
 
 void SetupWizardWindow::slotMoveToNextPage()
 {
-    slotStartTransition();
-    Q_EMIT nextButtonClicked();
+    if (_currentPage->validateInput()) {
```
TheOneRing, line 239: Chat: this was supposed to be an assert

TheOneRing (changes requested):
^

Issue #10294, opened 2022-11-17
[3.0] client starts in "No connection configured" state


michaelstingl:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the bug

My 3.0-rc1 client starts in "No connection configured" state. Re-start doesn't fix it. Also window doesn't open and doesn't come to the foreground automatically. So I'm not aware, it stalled.

### Expected behavior

Client should re-connect automatically, or notify the user and present an option to let the user do something. "No connection configured" is just wrong.

### Screenshots

![CleanShot 2022-11-17 at 11 24 37@2x](https://user-images.githubusercontent.com/214010/202426499-db0aff9b-29e7-4212-aa8a-47020f08a4e5.png)

### Logs

- [`no-connection-configured.log`](https://cloud.owncloud.com/index.php/f/5925020)

### Client version number

```
% /Applications/testpilotcloud.app/Contents/MacOS/testpilotcloud --version
testpilotcloud testpilotcloud 3.0.0.9116-rc1
3419c6
Libraries Qt 5.15.5, OpenSSL 1.1.1q  5 Jul 2022
Using virtual files plugin: suffix
osx-22.1.0
michaelstingl@Michaels-MacBook-Air ~ % 

```

### Server information

https://ocis.ocis-keycloak.released.owncloud.works/

individual-it:
Just happened to me after stopping and starting ocis


[ownCloud (Kopie).log](https://github.com/owncloud/client/files/10111374/ownCloud.Kopie.log)

TheOneRing:
> Just happened to me after stopping and starting ocis
> 
> [ownCloud (Kopie).log](https://github.com/owncloud/client/files/10111374/ownCloud.Kopie.log)

Looks like you just had a log rotation before that, could you also upload the previous log?
But anyhow, it appears that we don't properly handle unreachable servers or at least need a reconnect button or something like that.

individual-it:
I thought that was before the log rotate, but I will upload again if it happens again

gass:
This happened to me today when updateing to 3.0

HanaGemela:
Steps to test: https://github.com/owncloud/client/pull/10370#issuecomment-1371978375

[...]

PR #10370, opened 2022-12-21
Add an option to trigger a reconnect

TheOneRing:
image

Fixes: #10294

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -218,6 +218,11 @@ void AccountSettings::createAccountToolbox()
     connect(_toggleSignInOutAction, &QAction::triggered, this, &AccountSettings::slotToggleSignInState);
     menu->addAction(_toggleSignInOutAction);
 
+    _toggleReconnect = menu->addAction(tr("Reconnect"));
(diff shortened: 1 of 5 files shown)


src/gui/accountsettings.cpp
```diff
@@ -916,8 +921,9 @@ void AccountSettings::slotAccountStateChanged()
     // sync user interface buttons.
     refreshSelectiveSyncStatus();
 
-    /* set the correct label for the Account toolbox button */
     if (_accountState) {
+        _toggleReconnect->setVisible(!_accountState->isConnected());
```
fmoc, line 925: Please keep it visible at all times but disable it when it's not needed.

fmoc (changes requested):
I'd move the button below the log out one.

fmoc (approved):

TheOneRing:
Testing:
- Start the client
- Interrupt the internet connection
- Force sync or wait 1 minute
- Client should display timeouts and the reconnect button in the accounts menu is enabled
- Restore the connection
- Use reconnect button or wait 5-10 minutes.

Issue #10346, opened 2022-12-09
[QA] Folder with log files can't be opened


HanaGemela:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the QA issue

Folder with log files can't be opened when it's path is ./ownCloud-logdir

### Steps to reproduce the issue

1. Install ownCloud-3.0.0.9215-x86_64.pkg
2. Update to ownCloud-3.1.0-daily20221209.9307-x86_64.pkg

### Screenshots

<img width="616" alt="Screenshot 2022-12-09 at 09 42 43" src="https://user-images.githubusercontent.com/49001702/206661457-741b4aa3-94e9-44d1-8822-b3e157637a08.png">

### Expected behavior

Path is always set correctly to e.g. /private/var/folders/18/sy9lzt9x7j1_29cr9qq5kz680000gn/T/ownCloud-logdir
Folder can be opened 

### Actual behavior

Folder can't be opened but restart helps

TheOneRing:
I assume you called the client with a special argument?

HanaGemela:
Nothing special has been done

TheOneRing:
Then I don't know where the log file path is coming from....

HanaGemela:
<img width="602" alt="Snímek obrazovky 2022-12-23 v 8 35 26" src="https://user-images.githubusercontent.com/49001702/209293118-9838e40d-bb44-48e1-b87a-59441b9bc9f5.png">

Failed in ownCloud 3.1.0.9460-daily20221223 mac

TheOneRing:
How do you start the client, from a terminal?

HanaGemela:
I search with Spotlight, then it's running, I do the update and it just continues running

HanaGemela:
> Ensure that we have a proper environment when restarting the client f… #10375

Not sure if this went into the last available daily build so it hasn't been tested yet

HanaGemela:
Passed in ownCloud 3.1.0.9748-daily20230113 mac

PR #10375, opened 2022-12-23
Ensure that we have a proper environment when restarting the client f…

TheOneRing:
…from the installer

Instead of directly launching the binary we ask the system to start the registered service for us.

This is probably only possible on upgrades from 3.0+ as we used a different autostart mechanism before.

Fixes: #10346

### DIFF
--- a/admin/osx/post_install.sh.cmake
+++ b/admin/osx/post_install.sh.cmake
@@ -20,7 +20,7 @@ fi
 
 if [[ -f "${INSTALLER_TEMP}/OC_RESTART_NEEDED" ]]; then
     if [[ "${COMMAND_LINE_INSTALL}" = "" ]]; then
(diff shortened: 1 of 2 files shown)


erikjv (approved):

TheOneRing:
Testing:
- Install 3.0 upgrade to 3.1. The client should restart after the update and logs properly work.
- Install 3.0, disable autostart, upgrade to 3.1. Does the client restart once the installer is done?
- Install 2.11 upgrade to 3.1. The client does not automatically restart.

HanaGemela:
@TheOneRing what is the correct answer to this one? Install 3.0, disable autostart, upgrade to 3.1. Does the client restart once the installer is done?

ownCloud 3.1.0.9784-daily20230115 does restart

HanaGemela:
Upgrade from 2.11 to ownCloud 3.1.0.9784-daily20230115 also restarted automatically

HanaGemela:
Install 3.0 upgrade to ownCloud 3.1.0.9784-daily20230115. The client doesn't restart

PR #10387, opened 2023-01-04
Fix opening the browser

TheOneRing:
Due to a bug opening in browser only worked when logs where active


fmoc (approved):

TheOneRing:
Testing:
- 3.0 Disable logs (start the client without --logfile and disable logging in the settings), connect to a Koko server, right click a .odt, open in web... nothing will happen.
- With logs enabled the browser opens.
- 3.1 Independent of logging the browser will open.

HanaGemela:
Passed in ownCloud 3.1.0.9748-daily20230113 mac

PR #10402, opened 2023-01-16
[gui-tests] Added upload 3gb file test

saw-jan:


saw-jan:
@HanaGemela, will it be necessary to add this test case?
Because we already have a similar test case but for a `1GB` file. The reasons I am hesitating to add this test case are that the large file might not always be synced within 60 seconds period and might lead to flaky failure. And this drastically increases the CI build time.

https://github.com/owncloud/client/blob/9c6056da886cffd04c6f1c1f2a996e18bd76af6d/test/gui/tst_syncing/test.feature#L328-L332

HanaGemela:
@TheOneRing Do you see any value in testing both 1G and 3G file?

TheOneRing:
I don't see much point in uploading files bigger than lets say 110 mb.
In the current implementation 100mb is usually the maximum for a chunk.
For oc10 chunking this can be controlled by the setting `[General]maxChunkSize` for ocis its a server capability 
```
          "tus_support": {
            "version": "1.0.0",
            "resumable": "1.0.0",
            "extension": "creation,creation-with-upload",
            "max_chunk_size": 100000000,
            "http_method_override": ""
          },
```

Everything bigger than that is supposed to behave the same.

saw-jan:
^ So, closing here

PR #10416, opened 2023-01-19
Fix log description

TheOneRing:


michaelstingl (approved):
Late string change not perfect, but I could live with the untranslated until the next version.

PR #10431, opened 2023-01-26
[POC][gui-tests] Tests/gen custom report

saw-jan:
Custom report generator for QA

Notify on (to specified channel/individual):
- push to master
- push to rc branch
- tags/release
- nightly

Open tasks:
- [ ] include client version
- [ ] include server version

### Motivation
From https://github.com/owncloud/QA/issues/794
> The report for public needs to be readable for all people therefore if possible:
    - pdf format
    - no Time data table
    - no Logs
    - can the name of the feature be more user friendly? Currently it's called e.g. tst_addAccount
    - can we do something with the structure? Could it be a nice table with e.g. 2 columns Scenario and the result? How it is currently will be a nightmare to read when it gets longer

### Comments
IMO or what I was thinking was nightlies will notify the Squish generated report with verbose test logs only on failure otherwise just the CI passed notification, and anyone who wants to only check the minimal test report will get the notification with reports both on CI pass and fail.

We may not want to publish squish report frequently (except on failure-we need them for debugging) because they take up MBs of storage. But we can generate minial report and publish to remote storage as it is only of some KB size.


saw-jan:
Custom Reports:
oC10: https://cache.owncloud.com/public/owncloud/client/14128/oc10/guiReportUpload/minimal-report.html
oCIS: https://cache.owncloud.com/public/owncloud/client/14128/ocis/guiReportUpload/minimal-report.html

test/gui/drone/generate_report.js
```diff
@@ -0,0 +1,183 @@
...
+  if (!fs.existsSync(resultFile)) {
+    throw new Error('Report file not found: ' + resultFile)
+  }
+  if (!fs.lstatSync(resultFile).isFile()) {
+    throw new Error('Provided path is not a file: ' + resultFile)
+  }
+  if (path.extname(resultFile) !== '.js') {
+    // TODO: parse results from the file content
+    throw new Error('Provided path is not a JS file: ' + resultFile)
+  }
```
saw-jan, line 171: can be removed
```suggestion
```

saw-jan:
Not required as of now

PR #10445, opened 2023-01-31
Work/move add button out of folder status model

erikjv:


erikjv:
There is still an open question about "space" vs. "Space", so do not merge yet.

michaelstingl:
> There is still an open question about "space" vs. "Space", so do not merge yet.

@tbsbdr @EParzefall you had different opinions. Is this written down somewhere?

/cc @owncloud/pm

TheOneRing:
Screenshot pls :)

EParzefall:
@tbsbdr and I discussed this. Since space isn't really product or proper name, only a feature (though a really cool one), we'll keep it lower case in regular sentences. Only where headline style is appropriate, upper case is appropriate. 
Quote: proper noun: a noun that names a particular person, place, or thing
Source: https://www.merriam-webster.com/dictionary/proper%20name

src/gui/accountsettings.cpp
```diff
@@ -196,17 +196,38 @@ AccountSettings::AccountSettings(const AccountStatePtr &accountState, QWidget *p
     connect(_accountState.data(), &AccountState::stateChanged, this, &AccountSettings::slotAccountStateChanged);
     slotAccountStateChanged();
 
+    connect(ui->addButton, &QPushButton::clicked, this, &AccountSettings::slotAddFolder);
+    connect(_accountState.data(), &AccountState::stateChanged, this, [this](AccountState::State state) {
+        ui->addButton->setEnabled(state == AccountState::Connected);
+
+        if (state == AccountState::Connected) {
+            if (_accountState->supportsSpaces()) {
+                ui->addButton->setToolTip(tr("Click this button to add a Space."));
```
EParzefall, line 205:
```suggestion
                ui->addButton->setToolTip(tr("Click this button to add a space."));
```

EParzefall (approved):

src/gui/accountsettings.cpp
TheOneRing: Independent of that should space be translatable? If not we should remove it from the tr().
Else we might end up with forms of "Welträume"....

erikjv: My opinion is that we actually _should_ treat it as a proper noun, for that reason.

But if not used as a proper noun, don't leave it out of the translatable string. That would be very awkward for both developers as well as translators. I would go with a translation guide explaining the specific reason not to translate it.

EParzefall: > Independent of that should space be translatable? If not we should remove it from the tr(). Else we might end up with forms of "Welträume"....

Let's go with "Galaxien". :-) Gut question, but for now I'd leave it translatable. I don't think the decision to keep "Space(s)" in German is final yet.

erikjv:
Before:

Screenshot 2023-02-02 at 14 11 20

After:

Screenshot 2023-02-02 at 14 10 19

Note: I changed "Add Folder Sync Connection" to just "Add Folder". I'm not really sure what the two words "Sync Connection" add to the meaning of the button.

src/gui/accountsettings.cpp
EParzefall: > My opinion is that we actually _should_ treat it as a proper noun, for that reason.

We don't have one particular space but many. Take a look at Confluence, they don't use capitalization for their spaces either, not in regular sentences.

src/gui/folderstatusmodel.cpp
```diff
@@ -513,9 +489,9 @@ QModelIndex FolderStatusModel::parent(const QModelIndex &child) const
     }
     switch (classify(child)) {
     case RootFolder:
-    case AddButton:
         return QModelIndex();
     case SubFolder:
+        Q_FALLTHROUGH();
```
TheOneRing, line 494: nitpick.
As we require c++17 we should start using `[[fallthrough]];`

TheOneRing:
Looks like a massive improvement.
When ready to review please remove the lable and mark as ready for review.

src/gui/folderstatusmodel.cpp
erikjv: done

TheOneRing (approved):

PR #10476, opened 2023-02-07
Fix duplicate acct: in Webfinger resource ID

fmoc:
The WebFinger class expects a resource ID already. It's the caller's job to assemble the `acct:` URL, not WebFinger's.

Fixes #10471.

Issue #10471, opened 2023-02-07
Double acct in Webfinger request


michaelstingl:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the QA issue

Double acct in Webfinger request

### Steps to reproduce the issue

https://github.com/owncloud/enterprise/issues/5566

### Screenshots

.

michaelstingl:
To verify, check HTTP request in the log:

```
23-01-26 09:13:28:828 [ info sync.httplogger ]:	"a2534a92-1717-4cdd-9074-3c1e06d8a408: 
Request: GET https://lookup.sciebo.de/.well-known/webfinger?resource=acct%3Aacct%3Adsieben%40example.com&rel=http://webfinger.owncloud/rel/server-instance 
Header: { Content-Type: application/x-www-form-urlencoded, User-Agent: Mozilla/5.0 (Macintosh) mirall/3.1.0.9886 (owncloudazure, osx-21.6.0 ClientArchitecture: x86_64 OsArchitecture: x86_64), Accept: */*, X-Request-ID: a2534a92-1717-4cdd-9074-3c1e06d8a408, Original-Request-ID: a2534a92-1717-4cdd-9074-3c1e06d8a408, } 
Data: []"
```

- 2x `acct%3Aacct%3A` ⚠️
- 1x `acct%3A` ✅

HanaGemela:
`23-02-10 19:45:24:727 [ info sync.httplogger ]:	"a34c650e-c907-417d-9942-81cf44660968: Request: GET https://lookup.sciebo.de/.well-known/webfinger?resource=acct%3Adsieben%40example.com&rel=http://webfinger.owncloud/rel/server-instance Header: { Content-Type: application/x-www-form-urlencoded, User-Agent: Mozilla/5.0 (Macintosh) mirall/3.2.0.10094-rc1 (qa, osx-22.3.0 ClientArchitecture: x86_64 OsArchitecture: x86_64), Accept: */*, X-Request-ID: a34c650e-c907-417d-9942-81cf44660968, Original-Request-ID: a34c650e-c907-417d-9942-81cf44660968, } Data: []"`

Passed in qa 3.2.0.10094-rc1 mac

PR #10478, opened 2023-02-07
Fix duplicate acct: in Webfinger resource ID

fmoc:
The `WebFinger` class expects a resource ID already. It's the caller's job to assemble the `acct:` URL, not `WebFinger`'s.


fmoc:
Fixed version of #10476.

TheOneRing (changes requested):
Don't remove `toPercentEncoding`

TheOneRing (approved):

saw-jan:
Tested in https://github.com/owncloud/client/issues/10471#issuecomment-1426202976

PR #10490, opened 2023-02-08
Don't disable crashreporter on arm

TheOneRing:


dschmidt (approved):
if it works, :shipit:

PR #10493, opened 2023-02-09
Improve wizard pages' names

### LINKED ISSUE #10474: [WebFinger] WebFinger not a good user-facing term
The step shouldn't be called WebFinger. This is only a technical term.

CleanShot 2023-02-07 at 11 16 36@2x

```
branded 3.1.0.10011 e99d0f
Libraries Qt 5.15.5, OpenSSL 1.1.1q 5 Jul 2022
Using virtual files plugin: suffix
osx-22.3.0
```

@tbsbdr What would be a good term? (English + German)

Instance Lookup? Server Lookup?

fmoc:
Fixes #10474.


saw-jan:
Tested. See https://github.com/owncloud/client/issues/10474#issuecomment-1426175550

Issue #10519, opened 2023-02-20
Server-provided theming


fmoc:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Feature description

Modern oCIS servers provide one or more theme configurations. Each of them provides a `theme.json` API endpoint that provides basic information (e.g., instance name), color themes and icons (via relative URLs). The JSON file provides theming not only for the web client, but also for the native desktop and mobile clients

We should make use of this and use the data to enrich the user experience in the wizard and the client overall.

### Proposed solution

We should query the theme endpoint and display logos, texts and colors in the wizard. Some examples:

- server theme logo could be displayed in the bottom half of the wizard window next to the instance title (replacing the URL)
- server theme color could be used as the background color
- the logo can further be displayed instead of the user icon placeholder in the top navigation toolbar in the account settings

The information we are most interested in is the server logo at this point.

We will need to cache the downloaded metadata in some way to support offline use cases. The JSON contents could be stored in the account settings. Resources like images need to be cached on disk. We should be able to use the URLs as unique keys. Possibly, [`QNetworkDiskCache`](https://doc.qt.io/qt-5/qnetworkdiskcache.html) can be used for this.

### Additional context

See also:
- https://github.com/owncloud/web/issues/8424
- https://github.com/owncloud/web/issues/8425

fmoc:
Example JSON: https://ocis.ocis-traefik.latest.owncloud.works/themes/owncloud/theme.json

dragotin:
The theme endpoint exists today.

How do we takle the issue with

1. branded installers
2. offline operation 
3. themed server URL

?

I think the json should be delivered with the clients, and stored on the clients. With the theme endpoint, it can be updated.

Still we need some kind of "ownBrander" to create the initial branding json and hardcode the URL.

PR #10526, opened 2023-02-22
Cache specific network resources

fmoc:
This pull request adds a local network cache to every account. Using this cache, we can store resources locally to reduce the load on the server and also implement better offline support.

By default, network jobs' target resources are not stored in the cache and we always download the latest version. Therefore, currently, one needs to enable caching on a per-job basis. This might change in the future, since we depend on the server-side caching policy anyway.

Qt supports the common headers. In fact, Qt won't cache resources with `Cache-Control: ... must-revalidate` anyway (we had to check Qt's code base for this). We may have to either implement our own caching logic outside of `QNetworkAccessManager` or adapt the server policy.

Contributes to https://github.com/owncloud/client/issues/10519.

### DIFF
--- a/src/libsync/abstractnetworkjob.cpp
+++ b/src/libsync/abstractnetworkjob.cpp
@@ -155,20 +155,33 @@ void AbstractNetworkJob::sendRequest(const QByteArray &verb,
     const QNetworkRequest &req, QIODevice *requestBody)
 {
     _verb = verb;
+
     _request = req;
+    _request.setAttribute(QNetworkRequest::CacheSaveControlAttribute, _storeInCache);
+
+    if (_cacheLoadControl.has_value()) {
+        _request.setAttribute(QNetworkRequest::CacheLoadControlAttribute, _cacheLoadControl.value());
+    }
+
     _requestBody = requestBody;
+
     Q_ASSERT(_request.url().isEmpty() || _request.url() == url());
     Q_ASSERT(_request.transferTimeout() == 0 || _request.transferTimeout() == duration_cast<milliseconds>(_timeout).count());
+
     _request.setUrl(url());
     _request.setPriority(_priority);
     _request.setTransferTimeout(duration_cast<milliseconds>(_timeout).count());
+
     if (!isAuthenticationJob() && _account->jobQueue()->enqueue(this)) {
         return;
     }
+
     auto reply = _account->sendRawRequest(verb, _request.url(), _request, requestBody);
+
     if (_requestBody) {
         _requestBody->setParent(this);
     }
+
     adoptRequest(reply);
 }
 
@@ -391,6 +404,16 @@ QNetworkRequest::Priority AbstractNetworkJob::priority() const
 int AbstractNetworkJob::httpStatusCode() const
 {
     return reply()->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
+}
+
+void AbstractNetworkJob::setStoreInCache(bool storeInCache)
+{
+    _storeInCache = storeInCache;
+}
+
+void AbstractNetworkJob::setCacheLoadControl(QNetworkRequest::CacheLoadControl cacheLoadControl)
+{
+    _cacheLoadControl = cacheLoadControl;
 }
 
 } // namespace OCC
--- a/src/libsync/abstractnetworkjob.h
+++ b/src/libsync/abstractnetworkjob.h
@@ -32,6 +32,7 @@
 #include <QUrlQuery>
 
 #include <chrono>
+#include <optional>
 
 class QUrl;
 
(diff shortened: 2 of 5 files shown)


src/libsync/account.cpp
```diff
@@ -160,6 +162,12 @@ void Account::setCredentials(AbstractCredentials *cred)
 
     _am = _credentials->createAM();
 
+    const QString cacheLocation = QStringLiteral("%1/accounts/%2/network/").arg(QStandardPaths::writableLocation(QStandardPaths::CacheLocation), _uuid.toString());
```
TheOneRing, line 165: Can you add an example.
Is the app name included?

src/libsync/account.cpp
```diff
@@ -160,6 +162,12 @@ void Account::setCredentials(AbstractCredentials *cred)
...
+    qCDebug(lcAccount) << "Cache location for account" << this << "set to" << cacheLocation;
+    _networkCache->setCacheDirectory(cacheLocation);
+    _networkCache->setMaximumCacheSize(10 * 1024 * 1024);
```
TheOneRing, line 168: can you use our literals?

src/libsync/account.cpp
```diff
@@ -160,6 +162,12 @@ void Account::setCredentials(AbstractCredentials *cred)
...
+    _am->setCache(_networkCache);
```
TheOneRing, line 169: The nam takes ownershipt of the cache, so after a nam reset the cache will be null.

src/libsync/networkjobs.cpp
```diff
@@ -346,6 +346,8 @@ const QHash<QString, qint64> &PropfindJob::sizes() const
 AvatarJob::AvatarJob(AccountPtr account, const QString &userId, int size, QObject *parent)
     : AbstractNetworkJob(account, account->url(), QStringLiteral("remote.php/dav/avatars/%1/%2.png").arg(userId, QString::number(size)), parent)
 {
+    setStoreInCache(true);
```
TheOneRing, line 349: Requiring two calls looks like a good source for bugs?

src/libsync/account.cpp
fmoc: `~/.cache/ownCloud/accounts/<uuid>/network` on Linux.

src/libsync/account.cpp
fmoc: I want to get rid of our custom setting completely. We added it yesterday just for testing.

ownclouders:
Results for <strong>GUI-tests-oc10</strong> https://drone.owncloud.com/owncloud/client/14380/6/1

<details>


</details>

ownclouders:
Results for <strong>GUI-tests-ocis</strong> https://drone.owncloud.com/owncloud/client/14380/7/1

<details>


</details>

src/libsync/networkjobs.cpp
fmoc: I don't think just a subset of all possible permutations of these flags will suit our needs. We need the flexibility, though. The future network jobs added to download the server theme and according resources need to load both explicitly from the cache only(!) as well as download resources from the network (while storing this fresh response in the cache if successful). We can implement this by creating a the job twice and changing the caching policies accordingly.

I documented that both functions are related to one another. Also, the one existing example of cached resources, user avatars, uses `PreferCache` with `true` for instance. The approach attempting downloading and using the cache in parallel mentioned above will combine `AlwaysCache` with `false` and `AlwaysNetwork` with `true`.

michaelstingl:
@fmoc please open backend issue

src/libsync/abstractnetworkjob.cpp
```diff
@@ -155,20 +155,30 @@ void AbstractNetworkJob::sendRequest(const QByteArray &verb,
     const QNetworkRequest &req, QIODevice *requestBody)
 {
     _verb = verb;
+
     _request = req;
+    _request.setAttribute(QNetworkRequest::CacheSaveControlAttribute, _storeInCache);
+    _request.setAttribute(QNetworkRequest::CacheLoadControlAttribute, _cacheLoadControl.value());
```
TheOneRing, line 161: if(_cacheLoadControl)?

TheOneRing (changes requested):

src/libsync/abstractnetworkjob.h
```diff
@@ -130,6 +131,23 @@ class OWNCLOUDSYNC_EXPORT AbstractNetworkJob : public QObject
...
+     */
+    void setStoreInCache(bool storeInCache);
+
+    /**
+     * Configure whether to load data from the cache.
+     * By default, all resources are always fetched from the server, the cache is not being used. Qt normally would use the PreferNetwork policy (see the Qt docs for more information).
+     * See also \ref setStoreInCache
+     * @param cacheLoadControl a policy suitable for the current job
+     */
+    void setCacheLoadControl(std::optional<QNetworkRequest::CacheLoadControl> cacheLoadControl);
```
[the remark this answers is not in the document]
fmoc: You may want to clear the state again.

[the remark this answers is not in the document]
TheOneRing: I really don't think we need to expose that.
And in which scenario would you want to unset it?

TheOneRing (approved):

PR #10527, opened 2023-02-22
Don't persist empty passwd

TheOneRing:


fmoc (commented):
Maybe we should reduce the number of cases in which the password can be empty at all? I'm not sure it's a desired or needed state of the client. We could and probably should make that a class invariant.

TheOneRing:
The password can basically only be empty if we are logged out.
However we can't check whether we are logged out here as that state is managed by the accountstate....

fmoc (approved):

PR #10532, opened 2023-02-24
Qt::AA_DisableWindowContextHelpButton

TheOneRing:


src/libsync/platform_win.cpp
```diff
@@ -29,11 +31,20 @@ WinPlatform::~WinPlatform()
...
+    // The Windows style still has pixelated elements with Qt 5.6,
+    // it's recommended to use the Fusion style in this case, even
+    // though it looks slightly less native. Check here after the
+    // QApplication was constructed, but before any QWidget is
+    // constructed.
+    if (auto guiApp = qobject_cast<QGuiApplication *>(application)) {
+        if (guiApp->devicePixelRatio() > 1) {
+            QApplication::setStyle(QStringLiteral("fusion"));
+        }
+    }
```
fmoc, line 47: Not a fan of this. We should probably just set fusion all the time. But on the other hand, I don't really care.

Issue #9270, opened 2021-12-09
Text and icon in tray menu stays wrong


sfadschm:
The windows client tray icon sometimes shows the "idle" state, while the client is still syncing:

![145189629-5d1ccbf9-ce60-4a0a-a466-625d96bb2630](https://user-images.githubusercontent.com/40514119/145452774-e431c949-aa52-4323-9ff8-5f53c3e014c2.png)

This was observed with 2.9.1, I will try to observe and replicate with 2.9.2 from today.

Note: The screenshot was taken during a large upload action with a slow internet connection.
I only use sync folders, not the full sync feature (as this was important info in a previous issue).

Ref.: #7365.

### Expected behaviour
The tray icon should show the "syncing" state while syncing.

### Actual behaviour
It occasionally show the "idle/synced" state.

### Steps to reproduce
1. Install owncloud and add sync folders.
2. Upload a bunch of files.
3. Watch the tray icon.

### Client configuration
Client version: 2.9.1

Operating system: Windows 10 (up-to-date)

OS language: German

Installation path of client: default

sfadschm:
Bump.

Salipa-Gurung:
Tested with `ownCloud version 4.0.0.10468-daily20230321`
oCIS Server: [ocis.ocis-traefik.latest.owncloud.works](https://ocis.ocis-traefik.latest.owncloud.works/)
OS: Windows 10 Pro
| ![Screenshot from 2023-03-21 14-36-27](https://user-images.githubusercontent.com/53458341/226558484-e86dc0a3-ae63-4a18-9ff6-dd581adb2c98.png) | ![Screenshot from 2023-03-21 14-36-48](https://user-images.githubusercontent.com/53458341/226558542-87dfb23d-8d0b-4874-9002-e1d712939e23.png) |
|--|--|
|multiple file sync| single file sync |
| Tray icon mouse over shows "Waiting to start syncing."  | Tray icon mouse over shows "Sync is running." :heavy_check_mark: |

TheOneRing:
Well the icon however is correct, the text not so much

Salipa-Gurung:
Tested with `ownCloud version 4.0.0.10542-daily20230331`
oCIS Server: [ocis.ocis-traefik.latest.owncloud.works](https://ocis.ocis-traefik.latest.owncloud.works/)
OS: Windows 10 Pro
| Uploading all files at once | Uploading files consecutively |
|--|--|
[...]

PR #10536, opened 2023-02-24
Fix overall status computation

TheOneRing:
Please read carefully, I'm not sure what the whole logic here was meant to do (It`s my code).
Fixes: #9270

### DIFF
--- a/src/gui/folderman.cpp
+++ b/src/gui/folderman.cpp
@@ -85,42 +85,12 @@ void TrayOverallStatusResult::addResult(Folder *f)
         lastSyncDone = time;
     }
 
-    const auto status = f->syncPaused() ? SyncResult::Paused : f->syncResult().status();
-    switch (status) {
-    case SyncResult::Paused:
-        Q_FALLTHROUGH();
-    case SyncResult::SyncAbortRequested:
-        // Problem has a high enum value but real problems and errors
-        // take precedence
-        if (_overallStatus.status() < SyncResult::Success) {
-            _overallStatus.setStatus(status);
-        }
-        break;
-    case SyncResult::Success:
-        Q_FALLTHROUGH();
-    case SyncResult::NotYetStarted:
-        Q_FALLTHROUGH();
-    case SyncResult::SyncPrepare:
-        Q_FALLTHROUGH();
-    case SyncResult::SyncRunning:
-        if (_overallStatus.status() < SyncResult::Problem) {
-            _overallStatus.setStatus(status);
-        }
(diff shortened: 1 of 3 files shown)


src/gui/folderman.cpp
```diff
@@ -85,42 +85,12 @@ void TrayOverallStatusResult::addResult(Folder *f)
...
-        Q_FALLTHROUGH();
-    case SyncResult::Error:
-        Q_FALLTHROUGH();
-    case SyncResult::SetupError:
-        if (_overallStatus.status() < status) {
-            _overallStatus.setStatus(status);
-        }
-        break;
+    auto status = f->syncPaused() ? SyncResult::Paused : f->syncResult().status();
+    if (SyncResult::Undefined) {
```
fmoc, line 89: You probably want to compare against `status`.

fmoc (changes requested):
I guess the new logic is reasonable. I found a problem, please fix. I'll re-review afterwards.

dragotin (commented):
The fix is nice, but why and how does that solve the problem described in the Bugreport?

src/gui/folderman.cpp
TheOneRing: wow ... 🙈

TheOneRing:
Before status done was taking precedence over status in progress

src/gui/folderman.cpp
```diff
@@ -85,42 +85,12 @@ void TrayOverallStatusResult::addResult(Folder *f)
...
-        if (_overallStatus.status() < status) {
-            _overallStatus.setStatus(status);
-        }
-        break;
+    auto status = f->syncPaused() ? SyncResult::Paused : f->syncResult().status();
+    if (status == SyncResult::Undefined) {
+        status = SyncResult::Problem;
+    }
+    if (status > _overallStatus.status()) {
+        _overallStatus.setStatus(status);
```
dragotin, line 93: This somehow looks as if the status can never go to "lower" values any more, because of the if statement. In the code before it could. Or is it reset elsewhere again?

dragotin (approved):

src/gui/folderman.cpp
TheOneRing: The commutation happens on a temporary so we always start with undefined.

fmoc: `Undefined` is currently the smallest value. The order is documented above the definition. This should be fine.

fmoc (approved):

Issue #10495, opened 2023-02-09
"No connection configured" after device sleep, network change etc


michaelstingl:
I used the `3.1.0.9784-daily20230115` in production setup, with multiple accounts, on my MacBook Air M1. This means, I usually never reboot the device, only close the lid, and roam between different Wifi or offline areas.

After a while, I got "No connection configured" in one of the accounts, and it doesn't recover.

Suspicious findings in the logs:

```
23-01-25 17:19:40:351 [ info sync.httplogger ]:	"b18bd792-379e-413f-b654-154b2b3988c5: 
Response: PROPFIND 0 (Error: Temporary network failure.,326938ms) https://cloud.owncloud.com/remote.php/webdav/ 
Header: { } 
Data: []"

23-01-25 17:19:40:380 [ info sync.httplogger ]:	"1fb9ce1b-438a-4bb3-8f1a-2cfeeaa1c9b7: 
Response: PROPFIND 0 (Error: Network unreachable,4ms) https://cloud.owncloud.com/remote.php/dav/files/mstingl/ 
Header: { } 
Data: []"

23-01-25 17:19:44:467 [ info sync.httplogger ]:	"9515a772-7e31-4842-b0a9-a1a19ff37ac3: 
Response: PROPFIND 0 (Error: Network unreachable,3ms) https://cloud.owncloud.com/remote.php/dav/files/mstingl/ 
Header: { } 
Data: []"

23-01-25 18:36:18:385 [ info sync.httplogger ]:	"817f94a7-1b17-4415-87b7-8e6cd997bc9c: 
Response: POST 0 (Error: Operation canceled,2450657ms) https://cloud.owncloud.com/index.php/apps/oauth2/api/v1/token 
Header: { } 
Data: []"
```

I uploaded the full logs here:
https://cloud.owncloud.com/index.php/f/5955168

For me, it feels like the refresh token got lost accidentally. Client should only discard the refresh token when the token endpoint clearly says, it's invalid.

TheOneRing:
Looks like your refresh token expired.
```

[...]

Issue #10494, opened 2023-02-09
[4.0] [OIDC] "No connection configured" after application access was removed in Keycloak


michaelstingl:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the bug

When 4.0 client starts after application access was removed in Keycloak, it shows misleading "No connection configured" message.

### Expected behavior

Client should start with proper login prompt after token refresh failure.

### Steps to reproduce the issue

1. Connect 4.0 desktop sync client to https://ocis.team.owncloud.works ✅ 
2. Restart 4.0 desktop sync client  ✅ (`testpilotcloud-0209_21.58.39.341.log.gz`)
3. Quit 4.0 desktop sync client
4. Remove application access in Keycloak (https://keycloak.team.owncloud.works/auth/realms/oCIS/account/#/applications)
5. Launch 4.0 desktop sync client  ➡️ "No connection configured" ⚠️ (`testpilotcloud-0209_22.00.18.892.log.gz`)
6. Quit 4.0 desktop sync client
7. Launch 4.0 desktop sync client  ➡️ Prompts for login ✅ (`testpilotcloud-0209_22.01.15.586.log.gz`)

## Desktop Client

![CleanShot 2023-02-09 at 21 59 15](https://user-images.githubusercontent.com/214010/217941842-9ff4c043-86d6-4409-b85e-be48d369b4a0.png) | ![CleanShot 2023-02-09 at 22 00 33](https://user-images.githubusercontent.com/214010/217941769-7868c154-afd9-40e7-8ca0-4f6a7752a897.png) | ![CleanShot 2023-02-09 at 22 01 35](https://user-images.githubusercontent.com/214010/217941756-e600d5d4-e2a5-4c3f-9367-91b43ad3814e.png)
-- | -- | --
(2) Restart 4.0 desktop sync client  ✅ | (5) Launch 4.0 desktop sync client  ➡️ "No connection configured" ⚠️ | (7) Launch 4.0 desktop sync client  ➡️ Prompts for login ✅

## Keycloak

[...]

PR #10553, opened 2023-03-01
Emit proper signal on oauth error

TheOneRing:
Fixes: #10494
Fixes: #10495

### DIFF
--- a/src/libsync/account.h
+++ b/src/libsync/account.h
@@ -196,13 +196,19 @@ public:
 
     GraphApi::SpacesManager *spacesManager() const { return _spacesManager; }
 
+    /**
+     * We encountered an authentication error.
+     */
+    void invalidCredentialsEncountered();
+
 public slots:
     /// Used when forgetting credentials
     void clearAMCache();
 
 signals:
-    /// Triggered by handleInvalidCredentials()
-    void invalidCredentials();
+    /// Triggered by invalidCredentialsEncountered()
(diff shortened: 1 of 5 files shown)


src/libsync/account.h
```diff
@@ -191,13 +191,19 @@ class OWNCLOUDSYNC_EXPORT Account : public QObject
 
     CredentialManager *credentialManager() const;
 
+    /**
```
TheOneRing, line 194: Not sure this function in combination with the private signal really prevents anyone from using it wrong again.

src/libsync/creds/httpcredentials.cpp
```diff
@@ -328,8 +328,9 @@ bool HttpCredentials::refreshAccessTokenInternal(int tokenRefreshRetriesCount)
         if (refreshToken.isEmpty()) {
             // an error occured, log out
             forgetSensitiveData();
-            Q_EMIT _account->invalidCredentials();
             Q_EMIT authenticationFailed();
+            Q_EMIT fetched();
+            // Q_EMIT _account->invalidCredentials();
```
fmoc, line 333:
```suggestion
```
Should be cleaned up.

fmoc (changes requested):

src/libsync/account.h
fmoc: Well, it certainly won't hurt either.

fmoc (approved):

PR #10554, opened 2023-03-01
Work/sparkle errors

### LINKED ISSUE #10531: Sparkle updater errors are not logged sufficiently
### Pre-submission Checks

### Feature description

When using a self-hosted HTTP-only update server, the Sparkle updater reports an error:

image

One would assume that the updater now terminates, but it seems to continue:

image

So far, so good. Now, when we click `Install update`:

image

My guess is that since the URL provided by the updater is also HTTP and not HTTPS, the updater refuses to continue. Unfortunately, we can't tell for sure. The regular Sparkle updater log, `~/Library/Logs/SparkleUpdaterLog.log` is not being used by our own updater. This is likely due to the fact that we use the Objective-C API, which provides some handlers to handle errors manually. We do use some of these handlers, but this seems insufficient.

Long story short: we need better error logging for the Sparkle updater. Currently, it is impossible to tell for sure why the scenario described above failed.

### Proposed solution

Looking at Sparkle's [updater API docs](https://sparkle-project.org/documentation/api-reference/Protocols/SPUUpdaterDelegate.html), I noticed we only override a few of the handlers. I could imagine we need to override more of these (basically all of them which are passed an error). 

Maybe we can enable writing a log file to our config directory, or maybe there is a callback interface that can provide Sparkle's log messages to us. If possible, we should increase the verbosity of the update server.

### Additional context

I am going to deploy the config used to produce this bug on our demo update server later. I'll provide a URL then.

Another Sparkle issue (which is not really related, but could be worked on, too, while we're at it anyway): https://github.com/owncloud/client/issues/7592

erikjv:
Fixes #10531.

### DIFF
--- a/src/gui/updater/sparkleupdater_mac.mm
+++ b/src/gui/updater/sparkleupdater_mac.mm
@@ -59,7 +59,13 @@
 - (void) updater:(SUUpdater *)updater didAbortWithError:(NSError *)error
 {
     Q_UNUSED(updater)
-    qCDebug(OCC::lcUpdater) << error.description;
+
+    // Warning for end-user:
+    qCWarning(OCC::lcUpdater) << error.localizedDescription;
+
+    // Full error, log it for when we need to debug problems (this will hold the full error object,
+    // including possible nested `NSError`s):
+    qCInfo(OCC::lcUpdater) << "Full error:" << error.description;
 }
 
 - (void)updater:(SUUpdater *)updater didFinishLoadingAppcast:(SUAppcast *)appcast
@@ -68,7 +74,6 @@
     Q_UNUSED(appcast)
     qCDebug(OCC::lcUpdater) << "";
 }
-
 
 @end
 


src/gui/updater/sparkleupdater_mac.mm
```diff
@@ -59,7 +59,8 @@ - (void)updater:(SUUpdater *)updater willInstallUpdate:(SUAppcastItem *)update
 - (void) updater:(SUUpdater *)updater didAbortWithError:(NSError *)error
 {
     Q_UNUSED(updater)
-    qCDebug(OCC::lcUpdater) << error.description;
+    qCWarning(OCC::lcUpdater) << error.localizedDescription; // user facing error
+    qCInfo(OCC::lcUpdater) << "Full error description:" << error.description;
```
fmoc, line 63: `qCInfo` is user facing, too, isn't it?

fmoc (changes requested):

src/gui/updater/sparkleupdater_mac.mm
```diff
@@ -59,7 +59,8 @@ - (void)updater:(SUUpdater *)updater willInstallUpdate:(SUAppcastItem *)update
 - (void) updater:(SUUpdater *)updater didAbortWithError:(NSError *)error
 {
     Q_UNUSED(updater)
-    qCDebug(OCC::lcUpdater) << error.description;
+    qCWarning(OCC::lcUpdater) << error.localizedDescription; // user facing error
```
fmoc, line 62:
```suggestion
    qCWarning(OCC::lcUpdater) << error.localizedDescription;
```

fmoc (approved):

PR #10563, opened 2023-03-03
Ci: cancel previous jobs

TheOneRing:


.github/workflows/main.yml
```diff
@@ -8,6 +8,10 @@ on:
   pull_request:
   workflow_dispatch:
 
+concurrency: 
+  group: ${{ github.head_ref }}
```
fmoc, line 12: Fallback recommended, see https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-a-fallback-value.

fmoc (approved):

PR #10566, opened 2023-03-03
Remove support for credentials encoded in urls

TheOneRing:
This removes support for credentials stored in the url like `demo:demo@demo.owncloud.org`.
The 2.11 client allowed to enter such a url in the wizard, however the credentials where properly stored.
So there is no need to keep this hack around.

Do we need a changelog entry?


fmoc (approved):
I guess we should add a changelog entry, since this is a breaking change.

michaelstingl (approved):

Issue #9311, opened 2021-12-21
Folder move not propagated to server when any error occurs during the sync


gabi18:
Seen with client ownCloud-2.10.0-rc1.6315.x64.msi on Windows 10 20H2 

Steps to reproduce:

- On server: create a subfolder EmptyFolder inside folder Test-Data, create 2 files having a file name clash
Result: the client shows red error message "...file name clash..." for a while then shows the  'i' icon, 'Not synced' tab shows the file clash. The Explorer icon of file NeuesFile.txt and folder Test-Data remains spinning. 

![filename-clash](https://user-images.githubusercontent.com/75983861/146906154-cff1bf9b-94a9-42b5-b774-3dfdc34d64b7.png)

- move EmptyFolder to a new location (outside snyc root)
Result: the folder is moved locally but is not deleted on server

![client-2-10_filename-clash](https://user-images.githubusercontent.com/75983861/146910922-be7f4f37-702d-4a74-a051-c7821d5596ec.png)

```
...
20211220_1822_owncloud.log.0:12-20 18:23:43:182 [ warning sync.vfs.win ]:       void __cdecl OCC::VfsWin::fileStatusChanged(const class QString &,class OCC::SyncFileStatus) : couldn't CfSetInSyncState for "C:/Users/gabi/ownCloud26/Test-Data/EmptyFolder"  status  OCC::SyncFileStatus::StatusSync : "WindowsError: ffffffff80070178: Die Datei ist keine Clouddatei."
20211220_1822_owncloud.log.0:12-20 18:23:45:205 [ debug sync.localdiscoverytracker ]    [ OCC::LocalDiscoveryTracker::startSyncPartialDiscovery ]:      partial discovery with paths:  ("Test-Data", "Test-Data/123344888889999999999", "Test-Data/123344888889999999999/Neues Textdokument.txt", "Test-Data/EmptyFolder")
20211220_1822_owncloud.log.0:12-20 18:23:45:429 [ info sync.discovery ]:        Processing "Test-Data/EmptyFolder" | valid: false/true/false | mtime: 0/1640014730/0 | size: 0/0/0 | etag: ""//"" | checksum: ""//"" | perm: ""//"" | fileid: ""//"" | inode: 0/621914/ | type: CSyncEnums::ItemTypeSkip/CSyncEnums::ItemTypeDirectory/CSyncEnums::ItemTypeFile
[...]

PR #10582, opened 2023-03-08
Implement unit test for #9311

TheOneRing:


test/testsyncengine.cpp
```diff
@@ -906,6 +906,108 @@ private slots:
...
+        QVERIFY(fakeFolder.currentLocalState().find(lockedFileName));
+
+        // the unaffected download happened
+        QVERIFY(fakeFolder.currentLocalState().find(QStringLiteral("B/bb")));
+
+        // the remove did not happen as we had a case clash
+        QVERIFY(fakeFolder.currentLocalState().find(QStringLiteral("A")));
+
+        // S was deleted independent of the clas of A/aa and A/aA
+        QVERIFY(!fakeFolder.currentLocalState().find(QStringLiteral("S")));
```
TheOneRing, line 1008: Why is S gone in this case

TheOneRing: Relevant code 
https://github.com/owncloud/client/blob/bfa988779de6fc08edf905f0d3780dd3db893af2/src/libsync/owncloudpropagator.cpp#L1159-L1163

TheOneRing: Implemented in:
https://github.com/owncloud/client/commit/1e652e12b5027afa11e7a0443ec4b10669d4d79e

>Note that this means that if there are errors in subJobs the
>dirDeletionJobs won't get executed.

TheOneRing:
Merged with solution

Issue #10546, opened 2023-03-01
Log mac energy saver events


TheOneRing:
MacOS wakes up a macbook air periodically for a short moment.
If we trigger a token refresh during one of the wakeups, the computer might have returned to sleep before we received an answer.

- Log the events to better understand issues related to periodic wakeups.
- As we might want to block traffic during such a wakeup, design it  with the possibility to emit signals.

Related:
https://ladydebug.com/blog/2020/05/21/programmatically-capture-energy-saver-event-on-mac/

michaelstingl:
Maybe avoid performing tasks during wake from [Power Nap](https://support.apple.com/en-gb/guide/mac-help/mh40773/13.0/mac/13.0) ? (Can be disabled on Intel Macs, always active in Apple Silicon)

PR #10589, opened 2023-03-09
mac: Add listener and logging for system power notifications

erikjv:
Fixes: #10546


src/libsync/platform_mac.mm
```diff
@@ -39,20 +44,115 @@ - (BOOL)applicationShouldHandleReopen:(NSApplication *)sender hasVisibleWindows:
...
     }
     return YES;
 }
 
 @end
 
+namespace {
+
+class PowerNotificationsListener
+{
```
TheOneRing, line 57: Should we mention the source of inspiration?

src/libsync/platform_mac.mm
```diff
@@ -39,20 +44,115 @@ - (BOOL)applicationShouldHandleReopen:(NSApplication *)sender hasVisibleWindows:
...
+            qCInfo(OCC::lcPlatform) << "System power message: system will power on.";
+            break;
+
+        case kIOMessageSystemHasPoweredOn:
+            /* Announces that the system and its devices have woken up. */
+            qCInfo(OCC::lcPlatform) << "System power message: system has powered on.";
+            break;
+
+        default:
+            /* Not a system sleep and wake notification. */
```
TheOneRing, line 134: I guess we should handle all cases and not define default to make clang tell us about unhanded cases.
Additionally we should not use printf

TheOneRing (changes requested):
^

src/libsync/platform_mac.mm
erikjv: There are more messages, some even unused. See https://opensource.apple.com/source/xnu/xnu-2050.48.11/iokit/IOKit/IOMessage.h.auto.html . I think a break here is in order, also because Apple might add new ones. Also, take a look if we want to receive/log other events.

src/libsync/platform_mac.mm
erikjv: done

src/libsync/platform_mac.mm
```diff
@@ -39,20 +44,115 @@ - (BOOL)applicationShouldHandleReopen:(NSApplication *)sender hasVisibleWindows:
...
+        // add the notification port to the application runloop
+        CFRunLoopAddSource(CFRunLoopGetCurrent(), IONotificationPortGetRunLoopSource(notifyPortRef), kCFRunLoopCommonModes);
+    }
+
+private:
+    static void sleepWakeupCallBack(void *refParam, io_service_t service, natural_t messageType, void *messageArgument)
+    {
+        Q_UNUSED(service)
+
+        auto thiz = static_cast<PowerNotificationsListener *>(refParam);
```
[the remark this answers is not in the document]
erikjv: done

src/libsync/platform_mac.mm
erikjv: (printfs have been removed)

TheOneRing: Ok I expected it to be a enum, with defines no compiler in the world will be able to warn for new values.
Can we then log the unhandled cases to debug?

TheOneRing (approved):

Issue #10502, opened 2023-02-14
[Spaces] Button "Add Folder" instead of "Add Space"


michaelstingl:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the bug

When connected to oCIS, "Add Space" button should be shown below the sync folder list.

### Expected behavior

"Add Folder" button is shown below the sync folder list.

### Steps to reproduce the issue

1. Setup 4.0 client for `einstein:relativity` with https://ocis.ocis-traefik.latest.owncloud.works/

### Screenshots

![CleanShot 2023-02-14 at 11 04 06@2x](https://user-images.githubusercontent.com/214010/218706054-3acafec5-a0e8-4e10-951f-7af558163b45.png) | ![CleanShot 2023-02-14 at 11 08 35@2x](https://user-images.githubusercontent.com/214010/218706079-7994bc51-0434-4dac-bbf7-869dbc2b56c1.png)
-- | --
After account setup ⚠️ | After restart ✅ 

### Logs

- [testpilotcloud-0214_11.01.35.588.log.gz](https://github.com/owncloud/client/files/10731287/testpilotcloud-0214_11.01.35.588.log.gz)
- [testpilotcloud-0214_11.05.55.705.log.gz](https://github.com/owncloud/client/files/10731288/testpilotcloud-0214_11.05.55.705.log.gz)

### Client version number

```
testpilotcloud 4.0.0.10138-daily20230213 f56b10
Libraries Qt 5.15.8, OpenSSL 1.1.1t 7 Feb 2023
Using virtual files plugin: suffix
osx-22.2.0
```

### Server information

```
ownCloud Web UI 7.0.0-rc.12 
Infinite Scale 2.0.0+2b0457362 Community 
```

### Additional context

After restart, it looks correct.

PR #10591, opened 2023-03-14
Fix button label in account settings page

erikjv:
After adding an oCIS account, the "add" button on the account settings page will now correctly show "Add Space" instead of "Add Folder".

Fixes: #10502


TheOneRing (approved):

TheOneRing:
As the affected version wasn't released, do we need the changelog entry?

fmoc:
I don't think so. Nobody would ever see the difference. As long as you don't need to update an existing changelog entry, there's nothing to do here.

fmoc (approved):

PR #10593, opened 2023-03-14
Fix: Set the expected download content length in the job properly.

dragotin:
Remove a member that was keeping the actual content length.


TheOneRing (changes requested):
It appears we have to use all 3 https://github.com/owncloud/client-desktop-vfs-win/blob/d99bbd14ad50dd43d8983d4ee43651a4ac7cbad9/vfs_win.cpp#L294-L294

dragotin:
Closing in favor of #10607

Issue #10594, opened 2023-03-15
[QA] Quota update is too slow


HanaGemela:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the QA issue

Quota update is too slow
mac 
ownCloud 4.0.0.10373-daily20230310

### Steps to reproduce the issue

1. Set quota 
2. Upload the first file
3. Check used quota

### Screenshots

<img width="798" alt="Screenshot 2023-03-15 at 10 52 40" src="https://user-images.githubusercontent.com/49001702/225275737-810ea9b8-2233-4ee7-85ed-5bb02b842e81.png">

### Expected behavior

Quota is updated straightaway

### Actual behavior

It takes too long to update the quota

TheOneRing:
Should not take longer than 30s for the quota to update which in my opinion should be quite ok

TheOneRing:
Can you try with https://github.com/owncloud/client/pull/10600

HanaGemela:
This could also be a solution for https://github.com/owncloud/client/issues/10453#issuecomment-1471471063

HanaGemela:
I think this has fixed 10453. 
Quota update can take up to 30s cc @michaelstingl - please reopen if this is not sufficient 
Tested with ownCloud 4.0.0.10422-daily20230316, mac

PR #10600, opened 2023-03-15
Ensure delegate is redrawn when spaces are refreshed

TheOneRing:
Fixes: #10594


src/libsync/graphapi/spacesmanager.cpp
```diff
@@ -57,6 +57,7 @@ void SpacesManager::refresh()
             }
             _drivesMap = std::move(drivesMap);
         }
+        Q_EMIT done();
```
fmoc, line 60: `refreshDone` perhaps? `done` is too generic.

fmoc (changes requested):

fmoc (approved):

PR #10610, opened 2023-03-17
Enforce C++ version requirement

fmoc:
https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD_REQUIRED.html


saw-jan:
Needs no manual testing. Moving to Done
CC @fmoc

Issue #10539, opened 2023-02-03
Replace https://doc.owncloud.org/ help url with doc.owncloud.com/


jnweiger:
Seen with desktop client 3.0 and server 10.11.0

* Connect client to server, wait for initial sync.
* Change one file simultaneously on server and client. wait for sync.
* A conflict error occurs. OK
![grafik](https://user-images.githubusercontent.com/1108546/216598913-bf6852ba-0b0a-491d-8d1b-fa7363f87dd3.png)

* Click on the yellow link in the client. the client shows the not synced details. OK
![grafik](https://user-images.githubusercontent.com/1108546/216600216-aeb33c35-868d-4e32-a2f3-165edc53821d.png)

* Click the link at the bottom.
* A browser window opens for https://doc.owncloud.org/desktop/3.0/conflicts.html
* But this is immediately redirected to https://doc.owncloud.com/docs/next/ where we have a short section about client 3.0, but nothing about conflict resolution.

michaelstingl:
Vanilla app needs to be fixed in [`theme.cpp`](https://github.com/owncloud/client/blob/master/src/libsync/theme.cpp). 4.0?

HanaGemela:
https://doc.owncloud.com/desktop/4.0/conflicts.html in ownCloud 4.0.0.10490-daily20230324
Passed - the url has been changed

PR #10631, opened 2023-03-23
Replace owncloud.org urls with owncloud.com

TheOneRing:
Fixes: #10539

### DIFF
--- a/src/libsync/theme.cpp
+++ b/src/libsync/theme.cpp
@@ -284,7 +284,7 @@ QString Theme::defaultServerFolder() const
 
 QString Theme::helpUrl() const
 {
-    return QStringLiteral("https://doc.owncloud.org/desktop/%1.%2/").arg(OCC::Version::version().majorVersion()).arg(OCC::Version::version().microVersion());
+    return QStringLiteral("https://doc.owncloud.com/desktop/%1.%2/").arg(OCC::Version::version().majorVersion()).arg(OCC::Version::version().microVersion());
 }
 
 QString Theme::conflictHelpUrl() const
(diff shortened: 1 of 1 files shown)


src/libsync/theme.cpp
```diff
@@ -284,7 +284,7 @@ QString Theme::defaultServerFolder() const
 
 QString Theme::helpUrl() const
 {
-    return QStringLiteral("https://doc.owncloud.org/desktop/%1.%2/").arg(OCC::Version::version().majorVersion()).arg(OCC::Version::version().microVersion());
+    return QStringLiteral("https://doc.owncloud.com/desktop/%1.%2/").arg(OCC::Version::version().majorVersion()).arg(OCC::Version::version().microVersion());
```
mmattel, line 287: Why not writing: `https://doc.owncloud.com/desktop/latest/`
This would resolve in any case to a valid address.
Imagine one keeps the app longer than the actual 2 releases. Such a case would end up in a 404 while using latest would always resolve correct.
This is also true when using `https://doc.owncloud.com/desktop/latest/installing.html` which would finally reslove to `https://doc.owncloud.com/desktop/3.2/installing.html`

TheOneRing: The intention here is clear, if you have a 3 client, don't show the 4 doc.
I guess a better solution would be to not serve a 404 but redirect to latest.

ownclouders:
Results for <strong>GUI-tests-oc10</strong> https://drone.owncloud.com/owncloud/client/14733/6/1

:boom: The GUI tests failed.

GUI Logs: https://cache.owncloud.com/public/owncloud/client/14733/oc10/guiReportUpload/index.html

<details>

Server Logs: https://cache.owncloud.com/public/owncloud/client/14733/oc10/guiReportUpload/serverlog.log

Screenshots:
- [share_file_and_folder_to_a_group.png](https://cache.owncloud.com/public/owncloud/client/14733/oc10/guiReportUpload/screenshots/share_file_and_folder_to_a_group.png)

</details>

Issue #10254, opened 2022-11-09
Mac: 'Remove all files dialog' selected button has the wrong color


HanaGemela:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the QA issue

Option to 'Delete all' is highlighted. More user friendly would be to highlight the option to 'Keep files'
mac client 3.0.0 beta 1
ocis 2.0.0 rc1

### Steps to reproduce the issue

1. Create a Space on the server with some files
2. Sync the Space to the client
3. Rename the Space folder in the Finder
4. Create a new folder with the name of Space in the Finder

### Screenshots

<img width="820" alt="Screenshot 2022-11-09 at 12 59 14" src="https://user-images.githubusercontent.com/49001702/200828788-765ef25a-6d92-4d15-9a28-d310928ac78d.png">

### Expected behavior

Option 'Keep files' is highlighted 

### Actual behavior

Option 'Remove all files' is highlighted

fmoc:
Moved to 4.0 because of string changes. We want to create a proper dialog. We could use a stacked widget to deal with the different strings, for instance.

HanaGemela:
Failed ownCloud 4.0.0.10542-daily20230331, mac

HanaGemela:
The 'Keeps files' is dark grey. The dialog is not active so the button is not blue also after clicking it

TheOneRing:
Apparently the selection was correct the whole time, setting the default action explicitly however doesn't hurt.
The issue here is that on mac the window appears not properly activated and thus the default action is drawn in the inactive state "grey" instead of "blue".

HanaGemela:
Passed in ownCloud 5.0.0.11999-daily20230901 mac

<img width="951" alt="Screenshot 2023-09-01 at 09 52 15" src="https://github.com/owncloud/client/assets/49001702/5707f9f7-0270-4f7d-bdf1-f38604a31851">

PR #10646, opened 2023-03-29
DeleteAllFiles: Default to keep all files

TheOneRing:
Fixes: #10254

### DIFF
--- a/changelog/unreleased/10254
+++ b/changelog/unreleased/10254
@@ -1,0 +1,3 @@
+Bugfix: Select keep files as default action in all files removed dialog
+
+https://github.com/owncloud/client/issues/10254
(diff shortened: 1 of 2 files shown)


changelog/unreleased/10254
```diff
@@ -0,0 +1,3 @@
+Bugfix: Select keep files as default action in all files removed dialog
```
fmoc, line 1:
```suggestion
Bugfix: Pre-select keep files as default action in all files removed dialogs
```
```suggestion
Bugfix: Pre-select "keep files" as default action in all "files removed" dialogs
```

fmoc (changes requested):

changelog/unreleased/10254
TheOneRing: Why plural?

PR #10647, opened 2023-03-30
Introduce new Space class which offers additional fuctions on OAIDrive

TheOneRing:


src/libsync/graphapi/space.cpp
```diff
@@ -0,0 +1,74 @@
...
+    } else if (_drive.getId() == sharesIdC) {
+        // don't call it ShareJail
+        return tr("Shares");
+    }
+    return _drive.getName();
+}
+
+uint32_t Space::priority() const
+{
+    if (_drive.getDriveType() == personalC) {
```
fmoc, line 62: This logic should be documented.

fmoc (changes requested):

src/libsync/graphapi/space.cpp
TheOneRing: The function is documented

fmoc (approved):

Issue #10657, opened 2023-04-04
[QA] Brandable button remains selected after clicking


HanaGemela:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the QA issue

Brandable button remains selected after clicking
qa 4.0.0.10572-daily20230404, mac

https://gitea.owncloud.services/ownbrander/themes/commit/893b7217bbd7e50675e61f5b1eaac826af89a0db
and
https://gitea.owncloud.services/ownbrander/themes/commit/90f1a9c8989c548804e983f3a5a85ef350b2b650

### Steps to reproduce the issue

1. Click on the button

### Screenshots

<img width="796" alt="Screenshot 2023-04-04 at 12 47 47" src="https://user-images.githubusercontent.com/49001702/229770401-59a89a7b-f79f-421b-afe9-9e70be03090e.png">

### Expected behavior

Button is blue only while being clicked on

### Actual behavior

Button remains clicked/blue/selected (until a user clicks on it again)

HanaGemela:
Same on Windows

PR #10659, opened 2023-04-04
Url button must be non checkable

TheOneRing:
Fixes: #10657


fmoc (approved):
Shouldn't that be the default, actually?

TheOneRing:
> Shouldn't that be the default, actually?

It is for a toolbutton but not for our ToolButtonAction class

PR #10705, opened 2023-04-14
Add assert to validate that the file really did not change

TheOneRing:


src/gui/folder.cpp
```diff
@@ -751,6 +754,25 @@ void Folder::slotWatchedPathsChanged(const QSet<QString> &paths, ChangeReason re
             }
             if (spurious) {
                 qCInfo(lcFolder) << "Ignoring spurious notification for file" << relativePath;
+                Q_ASSERT([&] {
+                    if (record._fileSize < 1_mb) {
```
fmoc, line 758: Is this some performance optimization? If yes, you should add a comment.

fmoc (changes requested):

fmoc (approved):

PR #10706, opened 2023-04-14
Speed up mime detection

TheOneRing:


fmoc (approved):
We could also check for an existing suffix of the file.

Issue #10709, opened 2023-04-18
Update channel dialog shown on first run/after update


fmoc:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the bug

After #10627, the update channel dialog may be shown immediately after launching the client:

![screenshot_2023-04-18_10-28-26](https://user-images.githubusercontent.com/80399010/232719575-577f516a-7248-465f-b13b-987a69b41149.png)

Apparently, this window is even shown before the parent is initialized.

We assume the config file value is either invalid, empty or not set, hence the index lookup in the new code returns `-1`, a value that is not treated properly at the moment.

### Expected behavior

When an invalid value is found (i.e., `-1` is returned by the index lookup), we should fall back to the default value.

### Client version number

4.0.0 beta 1

fmoc:
Cannot reproduce on Linux. The code actually makes sure to fall back to `stable`. I can't tell right now how @michaelstingl managed to produce this problem. 

@michaelstingl could you check your configuration file for anomalies, please?

michaelstingl:
> @michaelstingl could you check your configuration file for anomalies, please?

No `updateChannel` in the `owncloud.cfg`. I uploaded it here:
- https://cloud.owncloud.com/index.php/f/5991565

HanaGemela:
Please test @michaelstingl

PR #10713, opened 2023-04-18
Connect currentIndexChanged after the data was populated

TheOneRing:
Fixes: #10709

### DIFF
--- a/src/gui/generalsettings.cpp
+++ b/src/gui/generalsettings.cpp
@@ -60,7 +60,6 @@ GeneralSettings::GeneralSettings(QWidget *parent)
 
     reloadConfig();
     loadMiscSettings();
-    slotUpdateInfo();
 
     // misc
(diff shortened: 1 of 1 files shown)


src/gui/generalsettings.cpp
```diff
@@ -206,16 +206,20 @@ void GeneralSettings::slotUpdateInfo()
...
 {
 #ifdef WITH_AUTO_UPDATER
-    QString channel = index == 0 ? QStringLiteral("stable") : QStringLiteral("beta");
+    QString channel;
+    if (index < 0) {
+        // invalid index reset to stable
+        channel = QStringLiteral("stable");
+    } else {
+        channel = _ui->updateChannel->itemData(index).toString();
+    }
```
fmoc, line 222: Do we want to handle overflows as well (i.e., `index > [...]->size()`)?

Issue #10721, opened 2023-04-19
[QA] Cannot add a new space 


HanaGemela:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the QA issue

I've added a new space in web ui and now been waiting already 20 minutes till it'll be shown in 'Add a space' dialog...

ownCloud 4.0.0.10697-daily20230419
mac
https://ocis.ocis-traefik.latest.owncloud.works and https://ocis.ocis-keycloak.latest.owncloud.works

Possibly related to https://github.com/owncloud/client/issues/10718

### Steps to reproduce the issue

1. Create a new space in web ui
2. Add a space in the client using the Add space button

### Screenshots

No screenshot

### Expected behavior

New space is listed in the 'Add space' dialog

### Actual behavior

New space is not listed in the 'Add space' dialog. Only Personal and Shares are shown

michaelstingl:
@HanaGemela please share logs (HTTP included)

HanaGemela:
Logs here https://cloud.owncloud.com/index.php/f/5991897

I can recreate with just one account and let me know how long should I wait for the new space to arrive. It used to be displayed immediately though

HanaGemela:
I've added a new log with one file and no waiting. 
Also realised that the space are only added after quitting and restarting

TheOneRing:
Please reproduce with a single account

HanaGemela:
@TheOneRing done

TheOneRing:
thx got an idea

HanaGemela:
Still broken in ownCloud 4.0.0.10703-daily20230419

HanaGemela:
Still broken in ownCloud 4.0.0.10718-beta2, mac

TheOneRing:
This time for real I hope

HanaGemela:
Fixed in ownCloud 4.0.0.10729-daily20230421

PR #10741, opened 2023-04-21
Connect legacy signal

TheOneRing:
This fixes spaces listing after a relogin
Fixes: #10721


src/libsync/graphapi/spacesmanager.cpp
```diff
@@ -43,6 +43,8 @@ SpacesManager::SpacesManager(Account *parent)
 
     connect(_refreshTimer, &QTimer::timeout, this, &SpacesManager::refresh);
     connect(_account, &Account::credentialsFetched, this, &SpacesManager::refresh);
+    // legacy signal which is going to be removed in 5.0
```
fmoc, line 46:
```suggestion
    // TODO: legacy signal which is going to be removed in 5.0
```

Makes this a little more visible in common IDEs.

TheOneRing: I have already a pr for that :D

fmoc (approved):

PR #10742, opened 2023-04-21
Cleanup autoupdate setup

TheOneRing:
Fixes: #10709

### DIFF
--- a/src/gui/generalsettings.cpp
+++ b/src/gui/generalsettings.cpp
@@ -52,8 +52,7 @@ GeneralSettings::GeneralSettings(QWidget *parent)
 {
     _ui->setupUi(this);
 
-    connect(_ui->desktopNotificationsCheckBox, &QAbstractButton::toggled,
-        this, &GeneralSettings::slotToggleOptionalDesktopNotifications);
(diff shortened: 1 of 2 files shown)


src/gui/generalsettings.cpp
```diff
@@ -131,8 +128,32 @@ GeneralSettings::GeneralSettings(QWidget *parent)
     // that way, when we intend to reset to the original selection when the dialog, we can look up the config file's stored value in the data model
     _ui->updateChannel->addItem(tr("stable"), QStringLiteral("stable"));
     _ui->updateChannel->addItem(tr("beta"), QStringLiteral("beta"));
-    connect(_ui->updateChannel, static_cast<void (QComboBox::*)(int)>(&QComboBox::currentIndexChanged), this, &GeneralSettings::slotUpdateChannelChanged);
-    slotUpdateInfo();
+
+    if (ConfigFile().skipUpdateCheck() || !Updater::instance()) {
```
erikjv, line 132: I think you got the condition inverted? In the removed code the function returns if this condition is true...

erikjv (changes requested):

erikjv (approved):

PR #10743, opened 2023-04-21
Remove unused includes

erikjv:


fmoc (changes requested):
Looks good, but CI fails, please fix.

src/gui/clientproxy.h
```diff
@@ -20,9 +20,8 @@
 #include <QRunnable>
 #include <QUrl>
 
-#include <csync.h>
 #include "common/utility.h"
-#include "owncloudlib.h"
+#include <csync.h>
```
TheOneRing, line 24: better not use a system include

src/gui/networksettings.cpp
```diff
@@ -15,16 +15,16 @@
...
 #include "configfile.h"
 #include "folderman.h"
-#include "accountmanager.h"
+#include "theme.h"
 
+#include <QList>
 #include <QNetworkProxy>
 #include <QString>
-#include <QList>
+#include <qevent.h>
```
TheOneRing, line 27: please have a look

TheOneRing (changes requested):

src/gui/networksettings.cpp
erikjv: This is on purpose, otherwise QtCore's `QEvent.h` is included, and that does not have one of the gui events.

TheOneRing: oO
Then please add a comment

src/gui/clientproxy.h
erikjv: done

src/gui/networksettings.cpp
erikjv: done

TheOneRing (approved):

PR #10787, opened 2023-05-02
libre-graph-api-cpp-qt-client static

TheOneRing:
Ecm set the symbols visibility to hidden, libre-graph-api-cpp-qt-client has no idea about visibility and does not explicitly set it. The result is a binary without symbols.

Force libre-graph-api-cpp-qt-client to build static libs.


fmoc (changes requested):
This seems like a really bad workaround... Let's try to find a better way.

If needed, I'd rather have a dependency specific build flag for this.

TheOneRing:
Feel free to fix upstream.

fmoc:
Obsoleted by #10790.

PR #10791, opened 2023-05-02
Replace QMacAutoReleasePool

TheOneRing:
QMacAutoReleasePool appears to be undocumented.
And I was unable to use it in Qt6.


TheOneRing:
@erikjv this is more or less the code you moved two years ago to platform.
BTW: What does it actually do...

erikjv (approved):

Issue #10461, opened 2023-02-03
Make sure app name etc. can only be queried from theme


fmoc:
We should define `APPLICATION_NAME` and friends only on the theme code and call `Theme::instance()->property()` everywhere.

See e.g., https://github.com/owncloud/client/pull/10458/commits/d1f3a827208481ed37c315eb35327c5ab22b56bb.

saw-jan:
Is this only the code refactoring? Or can be tested in some way?
CC @TheOneRing

PR #10812, opened 2023-05-05
Use theme methods instead of macros for application name

erikjv:
And also for the application icon name. There is one left in the common sources, but those cannot depend on libsync, so that's left alone.

### DIFF
--- a/src/gui/systray.cpp
+++ b/src/gui/systray.cpp
@@ -14,7 +14,6 @@
 
 #include "systray.h"
 #include "theme.h"
-#include "config.h"
 
 #ifdef USE_FDO_NOTIFICATIONS
 #include <QDBusConnection>
--- a/src/gui/systray.cpp
+++ b/src/gui/systray.cpp
@@ -75,9 +75,10 @@ void Systray::showMessage(const QString &title, const QString &message, const QI
 #else
 #ifdef USE_FDO_NOTIFICATIONS
     if (QDBusInterface(NOTIFICATIONS_SERVICE_C(), NOTIFICATIONS_PATH_C(), NOTIFICATIONS_IFACE_C()).isValid()) {
-        QList<QVariant> args = QList<QVariant>() << Theme::instance()->appNameGUI() << quint32(0) << QStringLiteral(APPLICATION_ICON_NAME)
-                                                 << title << message << QStringList() << QVariantMap() << qint32(-1);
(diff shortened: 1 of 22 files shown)


TheOneRing:
I assume there is are a few more. In general you could check to not include config.h which should only be needed in one or two locations the rest should rely on theme.h

erikjv:
Note: `guiutility_mac.mm` is still a problem

erikjv:
> Note: `guiutility_mac.mm` is still a problem

I left two uses of defines in there, as they are mac specific.

TheOneRing:
```
owncloud-client [work/get-app-name-from-theme ≡ +3 ~0 -0 !]> git grep "config.h"
src/common/plugin.cpp:#include "config.h"
src/csync/CMakeLists.txt:configure_file(config.h.in ${CMAKE_CURRENT_BINARY_DIR}/config.h)
src/csync/csync_exclude.cpp:#include "config.h"
src/gui/folderwatcher_linux.cpp:#include "config.h"
src/gui/guiutility_mac.mm:#include "config.h"
src/gui/libcloudproviders/libcloudproviders.cpp:#include "config.h"
src/gui/lockwatcher.h:#include "config.h"
src/gui/main.cpp:#include "csync/config.h"
src/gui/socketapi/socketapi.h:#include "config.h"
src/gui/updater/updater.cpp:#include "config.h"
src/libsync/configfile.cpp:#include "config.h"
src/libsync/filesystem.h:#include "config.h"
src/libsync/theme.cpp:#include "config.h"
test/testsyncvirtualfiles.cpp:#include "config.h"
test/testutility.cpp:#include "config.h"

File: src/gui/systray.cpp [hunk 1]
```diff
@@ -75,8 +74,8 @@ void Systray::showMessage(const QString &title, const QString &message, const QI
 #else
 #ifdef USE_FDO_NOTIFICATIONS
     if (QDBusInterface(NOTIFICATIONS_SERVICE_C(), NOTIFICATIONS_PATH_C(), NOTIFICATIONS_IFACE_C()).isValid()) {
-        QList<QVariant> args = QList<QVariant>() << Theme::instance()->appNameGUI() << quint32(0) << QStringLiteral(APPLICATION_ICON_NAME)
-                                                 << title << message << QStringList() << QVariantMap() << qint32(-1);
+        QList<QVariant> args = QList<QVariant>() << Theme::instance()->appNameGUI() << quint32(0) << Theme::instance()->applicationIcon() << title << message
```
TheOneRing, line 77: This used to be a QString, we now send a QIcon, I don't think that will work.

TheOneRing (changes requested):

erikjv:
This fixes https://github.com/owncloud/client/issues/10461

TheOneRing (approved):

PR #10891, opened 2023-05-30
Display the wizard again on first launch

TheOneRing:


fmoc (approved):
Not sure where this was broken... is this a new problem both introduced and fixed in 5.0?

PR #10896, opened 2023-05-31
Fix crash reporter check when variable is not defined

fmoc:


fmoc:
I tested this. Only newer versions of CMake complain. Fixed.

CMakeLists.txt
```diff
@@ -46,7 +46,8 @@ if(NOT WITH_CRASHREPORTER)
     message(STATUS "Build of crashreporter disabled.")
 else()
     # obviously, a crash reported without the URL set is pretty pointless and won't work
-    if(CRASHREPORTER_SUBMIT_URL STREQUAL "")
+    # note: to make the string comparison work, we need to have CMake evaluate the variable using ${...}
+    if("${CRASHREPORTER_SUBMIT_URL}" STREQUAL "")
```
TheOneRing, line 50:
```suggestion
    if(NOT DEFINED CRASHREPORTER_SUBMIT_URL)
```

TheOneRing (changes requested):

CMakeLists.txt
fmoc: This opens the case of "variable is set but empty" again. What's the point?

We could combine both checks:

```suggestion
    # note: to make the string comparison work, we need to have CMake evaluate the variable using ${...}
    if(NOT DEFINED CRASHREPORTER_SUBMIT_URL OR CRASHREPORTER_SUBMIT_URL STREQUAL "")
```

TheOneRing: Why would we set the variable to empty? And why do we need to document cmake behavior?

fmoc: > Why would we set the variable to empty?

If you rerun CMake to clear a variable, you will likely `-DTHIS_VAR=`, which just sets it to an empty string. It is widely accepted and therefore expected behavior to have empty strings be handled like undefined variables, especially in shell scripts and related contexts.

> And why do we need to document cmake behavior?

We want to make sure future readers know *why* we evaluate this variable in this context. We want to prevent future readers from making the same mistakes. CMake's behavior is not quite obvious here, therefore we should describe our thoughts.

TheOneRing: > > Why would we set the variable to empty?
> 
> If you rerun CMake to clear a variable, you will likely `-DTHIS_VAR=`, which just sets it to an empty string. It is widely accepted and therefore expected behavior to have empty strings be handled like undefined variables, especially in shell scripts and related contexts.
> 
> > And why do we need to document cmake behavior?
> 
> We want to make sure future readers know _why_ we evaluate this variable in this context. We want to prevent future readers from making the same mistakes. CMake's behavior is not quite obvious here, therefore we should describe our thoughts.

You would use `-UTHIS_VAR` but what scenario are you trying to fix?
I'd vote for `if (NOT DEFINED` and if you want to be pedantic you can add a static_assert to the crash reporter.

TheOneRing: > > > Why would we set the variable to empty?
> > 
> > 
> > If you rerun CMake to clear a variable, you will likely `-DTHIS_VAR=`, which just sets it to an empty string. It is widely accepted and therefore expected behavior to have empty strings be handled like undefined variables, especially in shell scripts and related contexts.
> > > And why do we need to document cmake behavior?
> > 
> > 
> > We want to make sure future readers know _why_ we evaluate this variable in this context. We want to prevent future readers from making the same mistakes. CMake's behavior is not quite obvious here, therefore we should describe our thoughts.
> 
> You would use `-UTHIS_VAR` but what scenario are you trying to fix? I'd vote for `if (NOT DEFINED` and if you want to be pedantic you can add a static_assert to the crash reporter.

I mean you are suggesting to implement url validation in cmake...

fmoc: The entire purpose of this discussion is to make a check weaker. I don't get it. Anyway. I'll merge your suggestion so you will  approve this PR.

Issue #10568, opened 2023-03-03
[QA] hard to read grammer when a conflict happens


jnweiger:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the QA issue

* Connect client to server, 
* Change a file on the server, 
* Quickly also change the same file locally.
* A conflict occurs. OK.
* Two messages are show in the 'Not Synced' tab. The second one is hard to read:
![grafik](https://user-images.githubusercontent.com/1108546/222809027-6a1cdf13-3f4a-4fa2-b2ea-59d262e5ae8d.png)
----
* "Server version downloaded, copied changed local file into conflict file." The other message says "local copy renamed" to describe the same.

Suggest to replace the text "copied changed local file" with "renamed local copy" 

### Steps to reproduce the issue

.

### Screenshots

.

### Expected behavior

.

### Actual behavior

.

saw-jan:
:heavy_check_mark: Tested with `ownCloud 5.0.0.11203-daily20230608 078f59`

![Screenshot from 2023-06-08 14-46-01](https://github.com/owncloud/client/assets/52366632/8519b971-30a5-47ae-bd40-ec689d7691c7)

PR #10901, opened 2023-05-31
Fix hard to read error message

erikjv:
A verb-used-as-adjective after a verb is very confusing: "copied changed local file", and problematic as it was used when a sync *conflict* occurred. This now reads "renamed local copy".

Fixes: #10568

### DIFF
--- a/src/libsync/progressdispatcher.cpp
+++ b/src/libsync/progressdispatcher.cpp
@@ -40,7 +40,7 @@ QString Progress::asResultString(const SyncFileItem &item)
             return QCoreApplication::translate("progress", "Uploaded");
         }
     case CSYNC_INSTRUCTION_CONFLICT:
-        return QCoreApplication::translate("progress", "Server version downloaded, copied changed local file into conflict file");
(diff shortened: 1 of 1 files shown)


src/libsync/progressdispatcher.cpp
```diff
@@ -40,7 +40,7 @@ QString Progress::asResultString(const SyncFileItem &item)
             return QCoreApplication::translate("progress", "Uploaded");
         }
     case CSYNC_INSTRUCTION_CONFLICT:
-        return QCoreApplication::translate("progress", "Server version downloaded, copied changed local file into conflict file");
+        return QCoreApplication::translate("progress", "Server version downloaded, renamed local copy into conflict file");
```
fmoc, line 43:
```suggestion
        return QCoreApplication::translate("progress", "Server version downloaded, local copy backed up as conflict file");
```

erikjv: done but slightly different: "local copy was backed up as conflict file"

fmoc (approved):

saw-jan:
Tested: https://github.com/owncloud/client/issues/10568#issuecomment-1582179382

Issue #10788, opened 2023-05-02
[Win] Remove legacy setting for sidebar entries


michaelstingl:
https://central.owncloud.org/t/synced-folders-still-show-in-win10-explorer-pane-despite-unchecking-show-option-in-owncloud-client/42096

@TheOneRing @HanaGemela 4.x or 5.0 ?
---

The setting has no effect on windows vfs, its only relevant for vfs off mode.
As Windows vfs is the default, we decided to remove the legacy setting which is confusing to the user.
This means with vfs off we will always create a sidebar entry.
![image](https://github.com/owncloud/client/assets/200626/f419f1ae-5297-4641-af44-1b80b0477f63)

saw-jan:
Tested with `ownCloud 5.0.0.11384-daily20230705 1c0613` on `win10`

![Screenshot from 2023-07-05 11-24-44](https://github.com/owncloud/client/assets/52366632/e1c77738-9c1c-499d-800f-04d4d1c98a6c)

PR #10910, opened 2023-06-05
[Win] Remove legacy setting for sidebar entries

erikjv:
We get this for free from the CfApi, and enabling this with winvfs results in duplicate entries.

Fixes: #10788


TheOneRing:
Actually the issue was just about removing the setting from the ui.
The issue there is that we offer a setting that allows to disable the sidebar entry, but it has only an effect on the legacy sidebar entry we use when vfs is off, with vfs on it had no effect at all.

I'll discuss with @michaelstingl what we do about the vfs off scenario, in the meantime I update the issue which lacks any description.

TheOneRing:
Please update the commit description and add a changeelog entry that we decided to drop support for sidebar entries with non vfs setups.

TheOneRing (changes requested):
^

TheOneRing:
Hmm can you apply clang format, the suggested change doesn't sound unreasonable.

TheOneRing:
Please also check whether https://github.com/owncloud/client/blob/9da79ecab40e7803c8aa7014d29f5abd3e8eb56e/src/common/utility.h#L254-L259 is still used, if not please remove.

TheOneRing:
@erikjv looks good, can you add a changelog entry.
`Change: Remove...`

Issue #10789, opened 2023-05-02
Create config file backup with EVERY update


michaelstingl:
I think currently only version change creates a new backup copy of the config file. I'd prefer a new backup copy with EVERY update (daily, alpha, beta, rc). As I side effect, I could see my full update history.

![CleanShot 2023-05-02 at 14 00 58@2x](https://user-images.githubusercontent.com/214010/235691748-6079420e-7f78-4c4f-9c0a-1d737e12501a.png)

fmoc:
DoR: append build number to config backups' version number (e.g., `4.0.0.10005`) and back up when build number changes, too. This should catch all the RCs, alphas, betas etc.

saw-jan:
Different config backups per build number.

![Screenshot from 2023-07-03 16-01-46](https://github.com/owncloud/client/assets/52366632/c746740b-3411-40fd-b549-41a8cd017b7e)

michaelstingl:
> @michaelstingl can you attach a config file for where it doesn't work?

Uploaded here:
- https://cloud.owncloud.com/f/6088317

michaelstingl:
I removed the `owncloud.cfg.backup_20221201_132437_2.11.1 (build 8438)`, then performed upgrades from
`ownCloud-5.0.0.12137-beta.3-arm64.pkg`
to
`ownCloud-5.0.0.12140-daily20230916-arm64.pkg`
and
`ownCloud-5.0.0.12148-daily20230917-arm64.pkg`,
but there are no new entries in the config folder

Salipa-Gurung:
Config file backup is created in every update :heavy_check_mark: 

![Screenshot from 2023-09-25 15-47-19](https://github.com/owncloud/client/assets/53458341/06c6ee0b-48ac-48f2-b221-5cfb95f83159)

Salipa-Gurung:
- Started desktop client `ownCloud-2.11.1.8438.AppImage` and then started client `ownCloud-5.0.0.12137-beta.3-x86_64.AppImage` 

- removed `'owncloud.cfg.backup_20230925_161714_2.11.1 (build 8438)'`
- upgraded client from `ownCloud-5.0.0.12137-beta.3-x86_64.AppImage` to `ownCloud-5.0.0.12195-daily20230924-x86_64.AppImage`, backup config file `owncloud.cfg.backup_20230925_161850_5.0.0.12137` was created.
[...]

PR #10916, opened 2023-06-09
Create config file backup with every update

erikjv:
The version number stored in the config file now includes the build number. So when the build number changes between RCs, alphas, betas etc, there will be a backup made of the "old" configuration.

Fixes: #10789

### DIFF
--- a/src/gui/application.cpp
+++ b/src/gui/application.cpp
@@ -75,7 +75,7 @@ bool Application::configVersionMigration()
 
     // Did the client version change?
     // (The client version is adjusted further down)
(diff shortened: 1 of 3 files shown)


fmoc:
After some testing, the only problem I see with this PR is that the build number is 0 in local/development builds. We need to make sure that that doesn't cause problems for distributions packaging our code and for developers testing their code locally with both own builds and dailies.

fmoc (approved):
Seems to work. I asked @TheOneRing to test this, too.

TheOneRing (approved):

saw-jan:
Tested in https://github.com/owncloud/client/issues/10789#issuecomment-1617822366

Issue #10443, opened 2023-01-31
[QA] Can't add an account in GUI when 'Show multi-account or disconnect' is not checked


HanaGemela:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the QA issue

Can't add an account in GUI when 'Show multi-account or disconnect' is not checked
Client 3.1.0

### Steps to reproduce the issue

1. Build a client with 'Show multi-account or disconnect' not checked
2. Open the client/wizard
3. Click Cancel 
4. Click Yes in 'Do you really want to cancel the account setup?'

### Screenshots

<img width="802" alt="Screenshot 2023-01-31 at 14 12 20" src="https://user-images.githubusercontent.com/49001702/215769164-841f98d7-1295-4cc4-b7e7-f31e600508bc.png">

### Expected behavior

User can add a new account in the GUI

### Actual behavior

No option to add an account in GUI

michaelstingl:
Workaround: Menu bar entry or restart

fmoc:
See https://github.com/owncloud/client/blob/master/src/libsync/theme.cpp, `multiAccount`.

HanaGemela:
Passed in btr 5.0.0.11351-daily20230629

Issue #270, opened 2013-01-25
Client does not sync changes made by other users


wexi79:
We have a shared folder that is accessed by multiple users. The folder syncs ok on the server between different users but the client only syncs changes made by the user itself.

Changes made by users are uploaded to the server but not distributed to other users' clients.

Client version 1.2.0 on Linux (Kubuntu 12.04) and Windows 7
Server version 5.0 pre alpha (4.91.2) on Ubuntu 12.04

PR #10918, opened 2023-06-12
Show Add Account button for single account clients

erikjv:
Specifically, show it when the wizard was aborted and no account was created. Previously, the button was not shown, and the user could not add an account after aborting the wizard.

Fixes: #10443

### DIFF
--- a/src/gui/settingsdialog.cpp
+++ b/src/gui/settingsdialog.cpp
@@ -199,15 +199,13 @@ SettingsDialog::SettingsDialog(ownCloudGui *gui, QWidget *parent)
     _actionGroup->setExclusive(true);
 
 
-    if (Theme::instance()->multiAccount()) {
-        _addAccountAction = new ToolButtonAction(QStringLiteral("plus-solid"), tr("Add account"), this);
-        _addAccountAction->setCheckable(false);
-        connect(_addAccountAction, &QAction::triggered, this, []{
-            // don't directly connect here, ocApp might not be defined yet
-            ocApp()->gui()->runNewAccountWizard();
-        });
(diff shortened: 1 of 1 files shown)


src/gui/settingsdialog.cpp
```diff
@@ -199,15 +199,13 @@ SettingsDialog::SettingsDialog(ownCloudGui *gui, QWidget *parent)
...
-        });
-        _ui->toolBar->addAction(_addAccountAction);
-    }
+    _addAccountAction = new ToolButtonAction(QStringLiteral("plus-solid"), tr("Add account"), this);
+    _addAccountAction->setCheckable(false);
+    connect(_addAccountAction, &QAction::triggered, this, [] {
+        // don't directly connect here, ocApp might not be defined yet
+        ocApp()->gui()->runNewAccountWizard();
+    });
+    _ui->toolBar->addAction(_addAccountAction);
```
fmoc, line 208: Don't we need a check here whether an account has been added already? In that case, we don't need to display the button right away.

fmoc (changes requested):

src/gui/settingsdialog.cpp
erikjv: I could do that, but in line \#270 (still in the constructor) all accounts are added, so it will remove that button before the dialog is shown. Yes, it's a small bit of overhead, but less logic to think about later ;-)

erikjv: I could also add a comment explaining this. So, whatever you like.

fmoc: I guess that's fine then. Thanks for the explanation.

fmoc (approved):

PR #10920, opened 2023-06-13
Remove code for touchedByUs

TheOneRing:
Remove code that relied on random timeouts to decide whether a file was modified by us or another process


src/libsync/owncloudpropagator.cpp
```diff
@@ -766,8 +766,6 @@ bool OwncloudPropagator::createConflict(const SyncFileItemPtr &item,
         return false;
     }
 
-    emit touchedFile(fn);
-    emit touchedFile(conflictFilePath);
```
TheOneRing, line 770: As we no longer announce the creation of the conflict file, we can no longer misuse the signal for the unit test.

fmoc (approved):

HanaGemela:
@TheOneRing anything to test here?

Issue #10884, opened 2023-05-26
[Activity] Persist Filter Settings


michaelstingl:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Feature description

Desktop sync client got powerful features for users to filter the activity protocol according to their needs. Only limitation is, filter settings weren't persisted, so users need to set filters again whenever they restart the client.

### Proposed solution

To improver user experience, filter settings should be persisted across client restarts.

### Additional context

![CleanShot 2023-05-26 at 17 45 32@2x](https://github.com/owncloud/client/assets/214010/27d0ce7f-f1f8-419a-9384-71558a1d1691)

```
testpilotcloud 5.0.0.11081-daily20230526 79bd0c
Libraries Qt 6.4.3, OpenSSL 1.1.1t 7 Feb 2023
Using virtual files plugin: suffix
macos-22.4.0
```

TheOneRing:
Only persist the category filter for the sync issue table

michaelstingl:
![CleanShot 2023-06-19 at 14 29 43](https://github.com/owncloud/client/assets/214010/1b757c46-ab44-4049-852d-ad70ac64f27a) | ![CleanShot 2023-06-19 at 14 30 08](https://github.com/owncloud/client/assets/214010/5fc0cc4e-91d9-4535-9c34-db912a5a076d) | ![CleanShot 2023-06-19 at 14 30 31](https://github.com/owncloud/client/assets/214010/1b98ea44-81a8-4676-aa7c-a4fd3f196daf)
-- | -- | --
Filter indicator correct ✅ | Empty filter indicator after restart ⚠️ | Filter indicator weird after change 💥 

```
testpilotcloud 5.0.0.11270-daily20230619 7c55d9
Libraries Qt 6.4.3, OpenSSL 1.1.1t 7 Feb 2023
Using virtual files plugin: suffix
macos-22.5.0
```

@erikjv could you have another look?

erikjv:
Two problems here:
1. the filter button text was not set correctly
2. the `SyncFileItem::Status::FilenameReserved` enum value was not handled (it was added recently by @TheOneRing)

S-Panta:
[...]

PR #10928, opened 2023-06-14
Persist filter settings for Not Synced tab

erikjv:
Fixes: #10884


src/libsync/syncfileitem.cpp
erikjv, line 1: @TheOneRing Could you give your opinion on this?

TheOneRing (commented):
Looks good to me.
Please add a change log entry.

TheOneRing (approved):

saw-jan:
Tested in https://github.com/owncloud/client/issues/10884#issuecomment-1619554110

PR #10929, opened 2023-06-14
Rewrite how we schedule syncs

TheOneRing:
With this change we need to update the documentation as we dropped OWNCLOUD_FULL_LOCAL_DISCOVERY_INTERVAL

### DIFF
--- a/src/gui/scheduling/syncscheduler.cpp
+++ b/src/gui/scheduling/syncscheduler.cpp
@@ -16,6 +16,9 @@
 
 #include "gui/folderman.h"
 #include "gui/scheduling/etagwatcher.h"
+#include "libsync/configfile.h"
+
+using namespace std::chrono_literals;
 
 using namespace OCC;
 
@@ -95,6 +98,19 @@ SyncScheduler::SyncScheduler(FolderMan *parent)
     , _queue(new FolderPriorityQueue)
 {
     new ETagWatcher(parent, this);
+
+    // Normal syncs are performed incremental but when fullLocalDiscoveryInterval times out
+    // a complete local discovery is performed.
+    // This timer here triggers a sync independent of etag changes on the server.
+    auto *fullLocalDiscoveryTimer = new QTimer(this);
+    fullLocalDiscoveryTimer->setInterval(ConfigFile().fullLocalDiscoveryInterval() + 2min);
+    connect(fullLocalDiscoveryTimer, &QTimer::timeout, this, [parent, this] {
+        for (auto *f : parent->folders()) {
+            if (f->isReady() && f->accountState()->state() == AccountState::State::Connected) {
+                enqueueFolder(f);
+            }
+        }
+    });
 }
 
 


src/gui/scheduling/syncscheduler.h
```diff
@@ -0,0 +1,85 @@
...
+    bool _run = false;
+    EtagWatcher *_etagManager;
+
+    QPointer<Folder> _currentSync;
+
+    struct QueueInfo
+    {
+        QueueInfo(Folder *f, Priority p);
+
+        // can be deleted
```
fmoc, line 72: When, where, why?

src/gui/scheduling/syncscheduler.h
```diff
@@ -0,0 +1,85 @@
...
+        QPointer<Folder> folder;
+        // raw pointer for lookup in _scheduledFolders
+        Folder *rawFolder;
```
fmoc, line 75: Shouldn't this be equivalent to `folder.data()`? If not, please add a few more words explaining this (apparent, but not necessarily existing) redundancy.

src/gui/scheduling/syncscheduler.h
```diff
@@ -0,0 +1,85 @@
...
+
+class FolderMan;
+
+class SyncScheduler : public QObject
+{
+    Q_OBJECT
+public:
+    enum class Priority : uint8_t {
+        // Normal sync triggered by etag change or something similar
+        None,
```
fmoc, line 36: Perhaps "Low" might suit better here?

src/gui/scheduling/syncscheduler.h
```diff
@@ -0,0 +1,85 @@
...
+
+        // Related to a user action
+        Medium,
+
+        // Usually triggered by a user (ForceSync)
+        Urgent
```
fmoc, line 42: Low, Medium, High might make more sense.

src/gui/scheduling/syncscheduler.h
```diff
@@ -0,0 +1,85 @@
...
+
+    void enqueueFolder(Folder *folder, Priority priority = Priority::None);
+
+    void start()
+    {
+        _run = true;
+        startNext();
+    }
+
+    void stop() { _run = false; }
```
fmoc, line 55: The way I understand this, it's more of a "stop request on next occasion". Could we better reflect this in the naming?

src/gui/scheduling/syncscheduler.cpp
```diff
@@ -0,0 +1,101 @@
...
+            _queue = std::move(out);
+        }
+    }
+    startNext();
+}
+
+void SyncScheduler::startNext()
+{
+    if (!_run || _currentSync) {
+        return;
```
fmoc, line 75: We should probably add an assert here, and make sure that `enqueueFolder` and the lambda below respect `_run` accordingly.

src/gui/scheduling/syncscheduler.h
```diff
@@ -0,0 +1,85 @@
...
+
+    bool isSyncRunning() const { return _currentSync; }
+
+
+private:
+    void startNext();
+
+    bool _run = false;
```
fmoc, line 63: Should be `_running`.

src/gui/scheduling/syncscheduler.h
```diff
@@ -0,0 +1,85 @@
...
+
+    void start()
+    {
+        _run = true;
+        startNext();
+    }
+
+    void stop() { _run = false; }
+
+    bool isSyncRunning() const { return _currentSync; }
```
fmoc, line 57: I'm not entirely sure how `_currentSync` relates to `_run`, given the current code and commentary. Please clarify.

src/gui/scheduling/syncscheduler.cpp
```diff
@@ -0,0 +1,101 @@
...
+    Q_ASSERT(folder->isReady());
+    Q_ASSERT(folder->canSync());
+    auto old = _scheduledFolders.find(folder);
+    if (old == _scheduledFolders.cend()) {
+        // the folder is not yet scheduled
+        _queue.emplace(folder, priority);
+    } else {
+        _scheduledFolders.insert({folder, priority});
+        // if the new priority is higher we need to rebuild the queue
+        if (priority > old->second) {
```
fmoc, line 54: This code should never have landed here. We need a standalone class for the raw priority queue.

Please remind me why we couldn't use `std::priority_queue`. (This could go into a comment by the way.)

src/gui/scheduling/etagwatcher.h
```diff
@@ -0,0 +1,53 @@
...
+#include <QObject>
+
+#include <unordered_map>
+
+namespace OCC {
+
+class FolderMan;
+class Folder;
+
+class EtagWatcher : public QObject
```
fmoc, line 28: Naming wise this should maybe be renamed to `ETagWatcher` (same goes for `ETagInfo` and the methods). Not really important, though.

src/gui/scheduling/etagwatcher.cpp
```diff
@@ -0,0 +1,104 @@
...
+constexpr auto pollTimeoutC = 30s;
+}
+
+EtagWatcher::EtagWatcher(FolderMan *folderMan, QObject *parent)
+    : QObject(parent)
+    , _folderMan(folderMan)
+{
+    connect(folderMan, &FolderMan::folderListChanged, this, [this] {
+        decltype(_lastEtagJob) intersection;
+        for (auto *f : _folderMan->folders()) {
```
fmoc, line 38: Please add some explanatory comment what the purpose of this intersection is and why you have to manage it yourself here.

src/gui/folder.cpp
```diff
@@ -1000,17 +932,9 @@ void Folder::startSync()
...
-            interval = std::chrono::milliseconds(env.toLongLong());
-        }
-        return interval;
-    }();
-    bool hasDoneFullLocalDiscovery = _timeSinceLastFullLocalDiscovery.isValid();
-    bool periodicFullLocalDiscoveryNow =
-        fullLocalDiscoveryInterval.count() >= 0 // negative means we don't require periodic full runs
+    const std::chrono::milliseconds fullLocalDiscoveryInterval = ConfigFile().fullLocalDiscoveryInterval();
+    const bool hasDoneFullLocalDiscovery = _timeSinceLastFullLocalDiscovery.isValid();
+    const bool periodicFullLocalDiscoveryNow = fullLocalDiscoveryInterval.count() >= 0 // negative means we don't require periodic full runs
```
fmoc, line 937: Please avoid inline comments. In this case especially, the line has become quite long.

src/gui/folder.cpp
```diff
@@ -763,7 +695,7 @@ void Folder::slotWatchedPathsChanged(const QSet<QString> &paths, ChangeReason re
 
         // Also schedule this folder for a sync, but only after some delay:
         // The sync will not upload files that were changed too recently.
-        scheduleThisFolderSoon();
+        QTimer::singleShot(SyncEngine::minimumFileAgeForUpload, this, [this] { FolderMan::instance()->scheduler()->enqueueFolder(this); });
```
fmoc, line 698: Could these single shot timers accumulate over time? We recently saw an issue where Windows's timer limit was reached.

src/gui/folder.cpp
```diff
@@ -641,8 +569,12 @@ void Folder::startVfs()
         _vfs->fileStatusChanged(stateDbFile + QStringLiteral("-wal"), SyncFileStatus::StatusExcluded);
         _vfs->fileStatusChanged(stateDbFile + QStringLiteral("-shm"), SyncFileStatus::StatusExcluded);
         _engine->setSyncOptions(loadSyncOptions());
+        registerFolderWatcher();
         _vfsIsReady = true;
-        slotScheduleThisFolder();
+        if (canSync()) {
```
fmoc, line 574: I think a comment that explains why we can safely ignore the "cannot sync" condition here would help readers understand the code better.

fmoc (changes requested):

src/gui/scheduling/syncscheduler.h
TheOneRing: Hm it stops scheduling, as its the stop function on the scheduler.
Its not called abortCurrentSync?

src/gui/scheduling/syncscheduler.cpp
TheOneRing: We do use std::priority_queue but a priority queue but we use a pair and could queue the same folder multiple times.

src/gui/scheduling/etagwatcher.cpp
TheOneRing: Failed to parse the question.
We maintain a list of folders we can either add a folderAdded signal and subscribe it along with folderRemoved or we take the change event.

src/gui/folder.cpp
TheOneRing: this is old code but well I can format it to your liking

src/gui/folder.cpp
TheOneRing: its a 3ms timer, which should not exist longer than 3ms

src/gui/scheduling/syncscheduler.cpp
```diff
@@ -0,0 +1,164 @@
...
+        Folder *rawFolder;
+        SyncScheduler::Priority priority;
+
+        friend bool operator<(const Element &lhs, const Element &rhs) { return lhs.priority < rhs.priority; }
+    };
+
+    // the actual queue
+    std::priority_queue<Element> _queue;
+    // helper container to ensure we don't enqueue a Folder multiple times
+    std::unordered_map<Folder *, SyncScheduler::Priority> _scheduledFolders;
```
fmoc, line 93: Might be more efficient to store a reference to the element in the queue as the value instead of the raw priority. Querying the priority is still possible, but it could speed up the find calls when looking for the duplicate in the priority queue.

src/gui/scheduling/syncscheduler.cpp
```diff
@@ -0,0 +1,164 @@
...
+    Q_ASSERT(folder->isReady());
+    Q_ASSERT(folder->canSync());
+    _queue->enqueueFolder(folder, priority);
+    startNext();
+}
+
+void SyncScheduler::startNext()
+{
+    Q_ASSERT(!_currentSync);
+    if (!_runing || _currentSync) {
```
fmoc, line 133: Typo: running.

fmoc (changes requested):

src/gui/scheduling/syncscheduler.cpp
TheOneRing: Is copy pasting the condition really an improvement?

src/gui/folder.cpp
TheOneRing: Please be constructive and suggest something, documenting the function canSync here is definitely the wrong place.

src/gui/scheduling/syncscheduler.cpp
TheOneRing: Are you sure this is possible without making assumptions about the implementation of priority_queue

TheOneRing: Also this is not  a hot path.

src/gui/scheduling/syncscheduler.cpp
fmoc: I think the code has changed in the meantime, and the code looks good at the moment.

fmoc (approved):

Issue #4693, opened 2016-04-17
[Linux] [Ubuntu 16.04] No tray icon in Xenial


stephen-d-hill:
### Expected behaviour
To see icon

### Actual behaviour
No icon

### Steps to reproduce
Install owncloud client

### Client configuration
Client version: 2.1.1+dfsg-1ubuntu1

Operating system: Xubuntu 16.04

OS language: English UK

guruz:
What do you see instead?
Can you run the client as ```owncloud --logfile -``` and paste the output?
Which Qt version?

stephen-d-hill:
The owncloud main window works fine, just no notification icon.
I am unsure how to find out my QT version.

logfile - 
[owncloud_logfile.txt](https://github.com/owncloud/client/files/223431/owncloud_logfile.txt)

guruz:
Thanks, the logfile mentioned it.. ```Using Qt 5.5.1```

Is there nothing before that in the log?

@ckamm @danimo maybe you know more about those tray issues..

stephen-d-hill:
Used --logfile instead of --logwindow this time - 
[owncloud_logfile1.txt](https://github.com/owncloud/client/files/223508/owncloud_logfile1.txt)

guruz:
Maybe this in Qt 5.6 helps..

```
commit f156c33c2739d84b97cdedf6ae9568b9cea728d5
Author: Dmitry Shachnev <mitya57@gmail.com>
Date:   Fri Apr 1 21:28:00 2016 +0200

    dbustray: Implement better detection of indicator-application
    
    We need to do the icon cache trick all desktops using indicator-application,
    these are not limited to Unity. For example, the default Xubuntu and Lubuntu
    desktops use indicator-application too.
    
    Without this, tray icons will be improperly shown on these desktops.
    
    Change-Id: Id397bbe9b594152d7c3a29c36c853e928af7dde4
    Reviewed-by: Shawn Rutledge <shawn.rutledge@theqtcompany.com>
```

guruz:
Some more changes from that Qt branch.. FYI @ckamm 

```
commit a4fac65938fdee74c07d34727920fb74087f82c6
Author: Dmitry Shachnev <mitya57@gmail.com>
Date:   Sat Feb 6 14:42:39 2016 +0300

    dbustray: Support late registering of tray icon menu
    
    If a menu is added to the tray icon after the icon itself has been registered,
    we need to register the menu properly.
    
[...]

PR #10943, opened 2023-06-19
Remove unused code

erikjv:
With Qt6 the `appmenu-qt5` is never available, so remove the codition and the else-clause.

### DIFF
--- a/src/gui/main.cpp
+++ b/src/gui/main.cpp
@@ -297,44 +297,7 @@ int main(int argc, char **argv)
             return 1;
         }
 #endif
-        // TODO: still needed? Move to platform
-        // We can't call isSystemTrayAvailable with appmenu-qt5 begause it hides the systemtray
-        // (issue #4693)
-        if (qgetenv("QT_QPA_PLATFORMTHEME") != "appmenu-qt5") {
-            if (!QSystemTrayIcon::isSystemTrayAvailable()) {
-                // If the systemtray is not there, we will wait one second for it to maybe start
-                // (eg boot time) then we show the settings dialog if there is still no systemtray.
-                // On XFCE however, we show a message box with explainaition how to install a systemtray.
-                qCInfo(lcMain) << "System tray is not available, waiting...";
-                Utility::sleep(1);
 
-                auto desktopSession = qgetenv("XDG_CURRENT_DESKTOP").toLower();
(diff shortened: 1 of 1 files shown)


src/gui/main.cpp
```diff
@@ -297,60 +297,40 @@ int main(int argc, char **argv)
...
-                    qCInfo(lcMain) << "System tray still not available, showing window and trying again later";
-                    QTimer::singleShot(10s, ocApp, &Application::tryTrayAgain);
-                } else {
-                    qCInfo(lcMain) << "System tray still not available, but assuming it's fine on 'ubuntu' desktop";
+        if (!QSystemTrayIcon::isSystemTrayAvailable()) {
+            // If the systemtray is not there, we will wait one second for it to maybe start
+            // (eg boot time) then we show the settings dialog if there is still no systemtray.
+            // On XFCE however, we show a message box with explainaition how to install a systemtray.
+            qCInfo(lcMain) << "System tray is not available, waiting...";
+            Utility::sleep(1);
```
TheOneRing, line 306: We really should make this async or remove it, we are blocking the main thread here...

src/gui/main.cpp
```diff
@@ -297,60 +297,40 @@ int main(int argc, char **argv)
...
+                        warnSystray();
+                        break;
+                    }
+                    Utility::sleep(1);
                 }
             }
-        }
-    } else {
-        // if the application is already running, notify it.
-        qCInfo(lcMain) << "Already running, exiting...";
```
TheOneRing, line 340: Why was the single instance ipc removed?

src/gui/main.cpp
TheOneRing: I mean they are talking of qt4 builds in the original issue ... https://github.com/owncloud/client/issues/4693

TheOneRing: Related:  https://github.com/owncloud/client/blob/f0d13707c36fc6e21a4f94394ee690bb9906c171/src/gui/owncloudgui.cpp#L578 all those hacks make me sick

src/gui/main.cpp
erikjv: done

HanaGemela:
@erikjv please provide steps to test or move to Done is there is nothing to do from QA

erikjv:
Nothing to test.

PR #10944, opened 2023-06-19
Fix misleading message

erikjv:
Some servers do not provide activities at all. In that case, the label should not read "have activities enabled", because that implies that they could be provided, but are disabled. Which is misleading.

### DIFF
--- a/src/gui/activitywidget.cpp
+++ b/src/gui/activitywidget.cpp
@@ -155,7 +155,7 @@ void ActivityWidget::showLabels()
     t.clear();
     QSetIterator<QString> i(_accountsWithoutActivities);
     while (i.hasNext()) {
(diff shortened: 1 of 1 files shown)


src/gui/activitywidget.cpp
```diff
@@ -155,7 +155,7 @@ void ActivityWidget::showLabels()
     t.clear();
     QSetIterator<QString> i(_accountsWithoutActivities);
     while (i.hasNext()) {
-        t.append(tr("<br/>Account %1 does not have activities enabled.").arg(i.next()));
+        t.append(tr("<br/>Account %1 does not provide activities.").arg(i.next()));
```
TheOneRing, line 158: Hmm its probably more about the server than about the account.
Maybe just
```suggestion
        t.append(tr("<br/>%1 does not provide activities.").arg(i.next()));
```

fmoc (approved):

Issue #10763, opened 2023-04-26
Do not set stylesheets globally in application setup code


fmoc:
See, e.g., `application.cpp`. We can't even tell for sure why these stylesheets are set and what they were supposed to improve. They likely can be removed.

HanaGemela:
@fmoc please provide steps to test

Issue #4197, opened 2015-11-25
Credentials Dialog too wide 


dragotin:
The dialog that asks for the password that comes up if the keychain could not be opened looks ugly because it is very wide. The error message is not folded.

Code: `HttpCredentialsGui::askFromUserAsync()`, _fetchErrorString needs to have newlines added.

Dianafg76:
@ogoffart I tested this issue and on my windows is working OK
Desktop v ownCloud-2.1.1.5722-nightly20151223-setup.exe
Server v {"installed":true,"maintenance":false,"version":"8.2.2.2","versionstring":"8.2.2","edition":"Enterprise"}

PR #10945, opened 2023-06-19
Do not set stylesheets globally in application setup code

erikjv:
This was introduced in #4197, but as far as we can tell, it is not needed anymore.

Fixes: #10763


saw-jan:
Issue was marked as Done: https://github.com/owncloud/client/issues/10763

CC @HanaGemela @erikjv

TheOneRing:
I guess you have to believe us.

saw-jan:
> I guess you have to believe us.

Then I will move this as well to Done

Issue #10950, opened 2023-06-21
[BUG] TUS -H 'Upload-Metadata: mtime missing


michaelstingl:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the bug

@glensc reported:

> I’ve started using OCIS 3.0 and using the latest macOS desktop client, all my files show upload time for file date. Full sync hasn’t completed yet so perhaps such is synced after, I’m not sure, though I think if it were syncing date it would be correct on upload.

- https://github.com/owncloud/ocis/issues/1248#issuecomment-1595902593

### Expected behavior

Old local file should have original date on the server after upload.

### Client version number

```
ownCloud 4.1.0.11250 279a6f
Libraries Qt 5.15.8, OpenSSL 1.1.1t 7 Feb 2023
Using virtual files plugin: suffix
osx-22.5.0
```

### Server information

oCIS 3.0

### Additional context

Desktop client only sends `Upload-Metadata: filename`:

```
23-06-19 15:03:47:322 [ info sync.httplogger ]:	"a429b677-89f8-4d3c-a955-605de0331f63: 
Request: POST https://ocis.ocis-traefik.daily.owncloud.works/dav/spaces/0f1c87f7-92e5-4549-8197-c7ece53e2e87$4c510ada-c86b-4815-8820-42cdf82c3d51 
Header: { X-OC-Mtime: 1685710240, 
Content-Type: application/offset+octet-stream, 
Content-Length: 1040524, 
Upload-Offset: 0, 
Tus-Resumable: 1.0.0, 
Upload-Metadata: filename L1Bob3Rvcy0wMDEuemlw,checksum U0hBMSBiYjg4ZTY1ZTRkMzNlZTVhMWNjY2QxMjViNzY2OGY1ZWM3NDlkNjNj, 
Upload-Length: 1040524, 
Authorization: Bearer [redacted], 
User-Agent: Mozilla/5.0 (Macintosh) mirall/5.0.0.11270-daily20230619 (testpilotcloud, 
macos-22.5.0 ClientArchitecture: arm64 OsArchitecture: arm64), 
Accept: */*, 
X-Request-ID: a429b677-89f8-4d3c-a955-605de0331f63, 
[...]

PR #10958, opened 2023-06-22
Include mtime in tus Upload-Metadata

TheOneRing:
Fixes: #10950

### DIFF
--- a/src/libsync/propagateuploadtus.cpp
+++ b/src/libsync/propagateuploadtus.cpp
@@ -78,15 +78,19 @@ UploadDevice *PropagateUploadFileTUS::prepareDevice(const quint64 &chunkSize)
 SimpleNetworkJob *PropagateUploadFileTUS::makeCreationWithUploadJob(QNetworkRequest *request, UploadDevice *device)
 {
     Q_ASSERT(propagator()->account()->capabilities().tusSupport().extensions.contains(QStringLiteral("creation-with-upload")));
-    // in difference to the old protocol the algrithm and the value are space seperated
+
     const auto checksumHeader = ChecksumHeader::parseChecksumHeader(_transmissionChecksumHeader);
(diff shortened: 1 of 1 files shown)


fmoc (approved):

src/libsync/propagateuploadtus.cpp
```diff
@@ -78,15 +78,19 @@ UploadDevice *PropagateUploadFileTUS::prepareDevice(const quint64 &chunkSize)
...
-    // in difference to the old protocol the algrithm and the value are space seperated
+
     const auto checksumHeader = ChecksumHeader::parseChecksumHeader(_transmissionChecksumHeader);
-    const QByteArray checkSum = QByteArray(CheckSums::toQString(checksumHeader.type()).toUtf8() + ' ' + checksumHeader.checksum()).toBase64();
-    const QByteArray base64Path = propagator()->fullRemotePath(_item->_file).toUtf8().toBase64();
-    qCDebug(lcPropagateUploadTUS) << "FullPath:" << propagator()->fullRemotePath(_item->_file) << "Base64:" << base64Path;
-    request->setRawHeader(QByteArrayLiteral("Upload-Metadata"), "filename " + base64Path + ",checksum " + checkSum);
+
+    QByteArrayList encodedMetaData;
+    auto addMetaData = [&encodedMetaData](const QByteArray &key, const QByteArray &value) { encodedMetaData << key + ' ' + value.toBase64(); };
```
erikjv, line 85: um, doesn't this result in `filename foo.txtchecksum sha1`? So does addMetaDatta need to add line separators or something? Or does the `<<` operator do that?

TheOneRing: Its a list QByteArrayList which later gets joined
`key base64Val`

erikjv: Doh you're right. My bad.

erikjv (approved):

Issue #10432, opened 2023-01-26
[QA][gui-tests] Implement `on the server` steps in python (squish)


saw-jan:
### Describe the QA issue

Currently, we are using `owncloud-test-middleware` to configure the server during the tests. In order to reduce the test dependencies we need to create those `on the server` steps in the local test directory i.e. in the gui test directory of this repo.
 
Steps to implement:
- [x] [Given steps](https://github.com/owncloud/client/issues/10432#issuecomment-1404861686)
- [x] [Then steps](https://github.com/owncloud/client/issues/10432#issuecomment-1404862169)
- [x] remove middleware code and middleware service from CI

saw-jan:
Then steps:
- [ ] `Then as "<user>" file "<fileName>" should not exist on the server`
- [ ] `Then as "<user>" folder "<folderName>" should not exist on the server`
- [ ] `Then as "<user>" file "<fileName>" should exist on the server`
- [ ] `Then as "<user>" folder "<folderName>" should exist on the server`
- [ ] `Then as "<user>" the file "<fileName>" on the server should have the content "<content>"`
- [ ] `Then as user "<user>" the file "<fileName>" should have a public link on the server`
- [ ] `Then as user "<user>" the folder "<folderName>" should have a public link on the server`
- [ ] `Then as user "<user>" the file "<fileName>" should not have any public link on the server`
- [ ] `Then as user "<user>" the folder "<folderName>" should not have any public link on the server`
- [ ] `Then the public should be able to download the file "<fileName>" without password from the last created public link by "<user>" on the server`
- [ ] `Then the public should be able to download the folder "<folderName>" without password from the last created public link by "<user>" on the server`
- [ ] `Then the public should not be able to download the file "<fileName>" from the last created public link by "<user>" on the server`
- [ ] `Then the public should not be able to download the folder "<folderName>" from the last created public link by "<user>" on the server`
- [ ] `Then the public should be able to download the file "<fileName>" with password "<password>" from the last created public link by "<user>" on the server`
- [ ] `Then the public should be able to download the folder "<folderName>" with password "<password>" from the last created public link by "<user>" on the server`
- [ ] `Then the public should not be able to download the file "<fileName>" with password "<password>" from the last created public link by "<user>" on the server`
- [ ] `Then the public should not be able to download the folder "<folderName>" with password "<password>" from the last created public link by "<user>" on the server`
- [ ] `Then the last public link share response of user "<user>" should include the following fields on the server`
```feature
[...]

PR #10959, opened 2023-06-22
[test-only] [full-ci] Verify fields of share response

Salipa-Gurung:
## Description
Implemented `on the server` steps in the local test directory.
These `Then` step implementation asserts whether the given resource(file/folder) is present on the server or not.

Renamed step:

```diff
-Then user "<user>" on the server should have a share with these details:
-| <field_name> | <value> | 

+Then user "<user>" in the server should have a share with these details:
+| <field_name> | <value> | 
```
Part of https://github.com/owncloud/client/issues/10432

### DIFF
--- a/test/gui/shared/scripts/helpers/UserHelper.py
+++ b/test/gui/shared/scripts/helpers/UserHelper.py
@@ -6,12 +6,14 @@ from helpers.ConfigHelper import get_config
 createdUsers = {}
 
 
-def basic_auth_header(user=None):
-    # default admin auth
-    token = b64encode(b"admin:admin").decode()
-    if user:
+def basic_auth_header(user=None, password=None):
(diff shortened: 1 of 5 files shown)


TheOneRing:
Just as a side node, the whole owncloud 10 sharing dialog is considered deprecated and we should avoid development of new tests for it.
CC: @michaelstingl

Salipa-Gurung:
> Just as a side node, the whole owncloud 10 sharing dialog is considered deprecated and we should avoid development of new tests for it.

So, in future, the sharing menu will redirect the user to the web browser like in oCIS?

michaelstingl:
> So, in future, the sharing menu will redirect the user to the web browser like in oCIS?

Maybe in the future, but no ETA yet.

PR #10971, opened 2023-06-27
[do-not-merge] build with qt6

saw-jan:


saw-jan:
ctest can pass with the image that has qt6

PR #10974, opened 2023-06-27
Drop support for Qt 5

fmoc:
We want to support building only for Qt 6.2 and newer from now on.

Note: currently the client doesn't build with Qt 5.15 any more anyway, so we either merge this PR now or another one fixing compatibility until we can merge this one.


TheOneRing:
See https://github.com/owncloud/client/pull/10975

HanaGemela:
@fmoc please provide steps to test or just move to Done if there is nothing to test

Issue #10978, opened 2023-06-27
Move capability, avatar and approvider fetching out of connection validator


TheOneRing:
Currently the capabilities might not be updated for a long time after the client was started. The actual date depends on various variables.

Until this is done we don't update neither the avatar nor the appprovider.

We need to move the code to its own class, and perform at least one check once we come online.

saw-jan:
@TheOneRing need steps to confirm this

Issue #10622, opened 2023-03-22
Research: Clarify ownership for Account objects and NetworkJobs


TheOneRing:
Don't pass a shared pointer to the network jobs but ensure the network jobs are parented by the account.

PR #10981, opened 2023-06-28
Fetch the capabilities and other infos once connected

TheOneRing:
Fixes: #10978


src/gui/fetchserversettings.cpp
```diff
@@ -0,0 +1,125 @@
...
+void FetchServerSettingsJob::runAsyncUpdates()
+{
+    // those jobs are:
+    // - never auth jobs
+    // - might get queued
+    // - have the default timeout
+    // - must not be parented by this object
+
+    // ideally we would parent them to the account, but as things are messed up by the shared pointer stuff we can't at the moment
+    // so we just set them free
```
erikjv, line 93: Maybe create an issue for this, so we don't forget about it?

erikjv (approved):

src/gui/fetchserversettings.cpp
TheOneRing: https://github.com/owncloud/client/issues/10622

Issue #8360, opened 2021-01-14
"Remove all dialog": Keep files will make the dialog pop up again and again


TheOneRing:
After clicking on keep files, a  new dialog pops up which asks to keep all files or remove all.

Might have been a special scenario, local db was deleted and all files on the remote where deleted.

gabi18:
Couldn't reproduce the issue that the 'Remove all' dialog popups again and again.
- deleted all folder and files on server,  clicked on 'Keep files' -> the popup didn't reappear
- deleted all folder and files locally, same as above

gabi18:
When testing to delete all both on server and locally the popup also appears once. 
Clicking on 'Keep files' cannot restore the files but they are still visible in the client.

[20210216_1611_owncloud.log.1.gz](https://github.com/owncloud/client/files/5989764/20210216_1611_owncloud.log.1.gz)

![all_deleted](https://user-images.githubusercontent.com/75983861/108085659-dca1ae00-7075-11eb-8c8a-0571762626d9.png)

gabi18:
I could reproduce the issue on Windows 10 with VFS enabled.

Steps done:
1. Enable VFS support in client
2. Delete all folder/files on server
3. The 'Keep files / Remove files' popup appears
4. Click 'Keep files' 
   Result: virtual folder/files are kept locally - no attempt to resync (see #8442)
5. Quit owncloud-client and restart
6. 'Keep files / Remove files' popup appears
   Result: client tries to resync after clicking 'Keep files' but doesn't succeed and the popup appears again and again

gabi18:
> I could reproduce the issue on Windows 10 with VFS enabled.
> 
> Steps done:
> 
>     1. Enable VFS support in client
> 
>     2. Delete all folder/files on server
> 
>     3. The 'Keep files / Remove files' popup appears
> 
>     4. Click 'Keep files'
>        Result: virtual folder/files are kept locally - no attempt to resync (see [[QA] No re-sync from client to server after 'Keep files'  #8442](https://github.com/owncloud/client/issues/8442))

NOT relevant here (for Windows VFS On), the issue above was reported for Linux

> 
>     5. Quit owncloud-client and restart

No need to restart on Windows, the popup appears again and again after clicking 'Keep files'. The files which should be kept are virtual - no chance to keep them.

> 
>     6. 'Keep files / Remove files' popup appears
>        Result: client tries to resync after clicking 'Keep files' but cannot succeed (because of VFS) and the popup appears again and again

[...]

PR #10983, opened 2023-06-28
If all files are removed on either sync side abort the sync

TheOneRing:
Instead of pausing the current sync we now abort the sync, ask the user and start a second sync

Fixes: #8360

### DIFF
--- a/src/cmd/cmd.cpp
+++ b/src/cmd/cmd.cpp
@@ -71,10 +71,12 @@ struct CmdOptions
 struct SyncCTX
 {
     explicit SyncCTX(const CmdOptions &cmdOptions)
-        : options { cmdOptions }
+        : options{cmdOptions}
+        , promptRemoveAllFiles(cmdOptions.interactive)
     {
     }
     CmdOptions options;
+    bool promptRemoveAllFiles;
     AccountPtr account;
     QString user;
 };
@@ -141,6 +143,7 @@ void sync(const SyncCTX &ctx)
         ctx.account, ctx.options.target_url, ctx.options.source_dir, ctx.options.remoteFolder, db);
     engine->setSyncOptions(opt);
     engine->setParent(db);
+    engine->setPromtRemoveAllFiles(ctx.options.interactive);
 
     QObject::connect(engine, &SyncEngine::finished, engine, [engine, ctx, restartCount = std::make_shared<int>(0)](bool result) {
         if (!result) {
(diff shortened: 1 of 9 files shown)


src/cmd/cmd.cpp
```diff
@@ -71,10 +71,12 @@ struct CmdOptions
 struct SyncCTX
 {
     explicit SyncCTX(const CmdOptions &cmdOptions)
-        : options { cmdOptions }
+        : options{cmdOptions}
+        , promptRemoveALlFiles(cmdOptions.interactive)
```
fmoc, line 75: Typo.

src/gui/folder.cpp
```diff
@@ -1027,6 +1027,12 @@ void Folder::startSync()
     }
 
     _engine->setIgnoreHiddenFiles(_definition.ignoreHiddenFiles);
+    if (_allowRemoveAllOnce) {
```
fmoc, line 1030: What does "once" mean in this context?

src/gui/folder.cpp
```diff
@@ -1332,22 +1338,18 @@ bool Folder::virtualFilesEnabled() const
     return _definition.virtualFilesMode != Vfs::Off && !isVfsOnOffSwitchPending();
 }
 
-void Folder::slotAboutToRemoveAllFiles(SyncFileItem::Direction dir, const std::function<void(bool)> &abort)
+void Folder::slotAboutToRemoveAllFiles(SyncFileItem::Direction direction)
```
fmoc, line 1341: :+1:

fmoc (changes requested):

src/gui/folder.cpp
TheOneRing: Only for the next sync

fmoc: So if this is set, the user was prompted already?

src/gui/folder.cpp
```diff
@@ -1332,22 +1338,18 @@ bool Folder::virtualFilesEnabled() const
...
+    const QString msg = direction == SyncFileItem::Down
+        ? tr("All files in the sync folder '%1' folder were deleted on the server.\n"
+             "These deletes will be synchronized to your local sync folder, making such files "
+             "unavailable unless you have a right to restore. \n"
+             "If you decide to keep the files, they will be re-synced with the server if you have rights to do so.\n"
+             "If you decide to delete the files, they will be unavailable to you, unless you are the owner.")
+        : tr("All the files in your local sync folder '%1' were deleted. These deletes will be "
+             "synchronized with your server, making such files unavailable unless restored.\n"
+             "Are you sure you want to sync those actions with the server?\n"
+             "If this was an accident and you decide to keep your files, they will be re-synced from the server.");
```
dragotin, line 1352: :see_no_evil:

src/gui/folder.cpp
TheOneRing: yes

src/gui/folder.cpp
TheOneRing: There is an old issue that goes on and on about what to write here. Or do you men the use of the ? operator

dragotin: Yes, ? operator and = and == in one line. I dislike, because that takes longer to brain-parse than 
```
QString msg;
if (direction == SyncFileItem::Down) {
  msg = tr("...foo...");
} else {
  msg = tr("...bar...");
}
```
Yes, that means that `msg` is not `const`, but I think that is acceptable ;-)

fmoc: We usually solve this by making the code a lambda that returns the value and is called immediately. The result can be `const` again. Any compiler should optimize that out quickly, i.e., in a release build, there should be no call overhead.

dragotin: Yes, does not make the code more easily to read for "normal people", but is a clever solution of course.

fmoc (approved):

PR #11000, opened 2023-07-04
Remove removed accounts from activitywidget.cpp

TheOneRing:
I debugged a bit more and it turned out that the activitywidget.cpp class was keeping a reference to the shared pointer.
This commit only cleans the internal state which was previously guarded in the sender.

Testing: Please test removal of an account followed with interactions with the activities view.

### DIFF
--- a/src/gui/accountstate.h
+++ b/src/gui/accountstate.h
@@ -199,11 +199,6 @@ private:
 };
 }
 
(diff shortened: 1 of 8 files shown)


TheOneRing:
To do:remove

inline size_t qHash(const OCC::AccountStatePtr &acs, size_t seed)

fmoc (approved):

Issue #10996, opened 2023-07-04
Client does not check for presence of system tray at startup


Deaddy:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the bug

The client does not check whether or not a system tray is present.

On startups beyond the first startup (where one is prompted for configuration) nothing visible happens, i.e. one only sees
```
$ Downloads/ownCloud-4.1.0.11250-x86_64.AppImage
QStandardPaths: XDG_RUNTIME_DIR not set, defaulting to '/tmp/runtime-deaddy'
m_crashReporterChar:  /tmp/.mount_ownCloirZSjT/usr/bin/owncloud_crash_reporter
```
but that's it.

If the client is already configured and still authorized, actually synchronization seems to work.

Further invocations of the command only say that the GUI application already is running:
```
Downloads/ownCloud-4.1.0.11250-x86_64.AppImage
QStandardPaths: XDG_RUNTIME_DIR not set, defaulting to '/tmp/runtime-deaddy'
gui.application: Already running, exiting...
```

### Expected behavior

I expect to see the client.

In older versions invocation of the owncloud command even when the client was already running would also foreground the application, what was nice.

One simply has to check `QSystemTrayIcon::isSystemTrayAvailable()` and only hide/minimze the client if this evaluates to true.

#### Workaround:
Start client with `-s` parameter.

### Steps to reproduce the issue

1. not have a traybar
2. start an already configured client

### Screenshots

nothing is visible, that is the point. so no screenshot :-)

### Logs

See above.

### Client version number

ownCloud-4.1.0.11250-x86_64.AppImage 279a6f
Qt 5.15.8

### Desktop environment (Linux only)

anything that is not a desktop environment (e.g. xmonad, lab computers who only have rudimentary x, servers) or where no tray bar is visible

### Server information

Independent of server versions (multiple oc10.12.x and ocis-3.x versions) tested as target

### Additional context

Qt detection of system tray seems to work as expected, as this [stolen](https://stackoverflow.com/questions/72687141/how-to-check-system-tray-availability-programmatically) code easily demonstrates:

is_tray_available.cpp:
```
#include <QApplication>
#include <QSystemTrayIcon>

int main(int argc, char** argv)
{
[...]

PR #11002, opened 2023-07-04
Update the description of the commandline switch `-s`

TheOneRing:
Fixes: #10996

```
Usage: C:\CraftRoot\download\git\owncloud\owncloud-client\cmake-build-debug\bin\owncloud.exe [options] [<vfs file>]
ownCloud version 5.0.0-git
File synchronization desktop utility.


For more information, see https://www.owncloud.com

Options:
  -?, -h, --help         Displays help on commandline options.
  --help-all             Displays help including Qt specific options.
  -v, --version          Displays version information.
  -s, --show             Start with the main window visible, or if it is already running, bring it to the front. By default, the client launches in the background.
  -q, --quit             Quit the running instance.
  --logfile <filename>   Write log to file (use - to write to stdout).
  --logdir <name>        Write each sync log output in a new file in folder.
  --logflush             Flush the log file after every write.
  --logdebug             Output debug-level messages in the log.
  --language <language>  Override UI language.
  --list-languages       Lists available translations, see --language.
  --confdir <dirname>    Use the given configuration folder.
  --debug                Enable debug mode.

Arguments:
  vfs file               Virtual file system file to be opened (optional).
```

### DIFF
--- a/src/gui/main.cpp
+++ b/src/gui/main.cpp
@@ -127,7 +127,7 @@ CommandLineOptions parseOptions(const QStringList &arguments)
     parser.addOption(showSettingsLegacyOption);
 
     auto showOption =
-        addOption({{QStringLiteral("s"), QStringLiteral("show")}, QStringLiteral("Show the main window. By default the client is started in the background.")});
+        addOption({{QStringLiteral("s"), QStringLiteral("show")}, QStringLiteral("Start with the main window visible, or if it is already running, bring it to the front. By default, the client launches in the background.")});
     auto quitInstanceOption = addOption({{QStringLiteral("q"), QStringLiteral("quit")}, QStringLiteral("Quit the running instance.")});
     auto logFileOption = addOption({QStringLiteral("logfile"), QStringLiteral("Write log to file (use - to write to stdout)."), QStringLiteral("filename")});
     auto logDirOption = addOption({QStringLiteral("logdir"), QStringLiteral("Write each sync log output in a new file in folder."), QStringLiteral("name")});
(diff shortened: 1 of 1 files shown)


src/gui/main.cpp
```diff
@@ -122,7 +122,12 @@ CommandLineOptions parseOptions(const QStringList &arguments)
         return option;
     };
 
-    auto showSettingsOption = addOption({{QStringLiteral("s"), QStringLiteral("showsettings")}, QStringLiteral("Show the settings dialog while starting.")});
+    auto showSettingsLegacyOption = QCommandLineOption{{QStringLiteral("showsettings")}, QStringLiteral("Hidden legacy option")};
+    showSettingsLegacyOption.setFlags(QCommandLineOption::HiddenFromHelp);
+    parser.addOption(showSettingsLegacyOption);
+
+    auto showOption =
+        addOption({{QStringLiteral("s"), QStringLiteral("show")}, QStringLiteral("Show the main window. By default the client is started in the background.")});
```
fmoc, line 130: The second could be misleading to users who do not know the flag. The flag can be used to bring the application to front if running, or to launch it with the settings window visible after startup. We should either not detail this (as it used to be) or make sure it's complete.

fmoc (changes requested):

src/gui/main.cpp
TheOneRing: Please propose a better description.

fmoc:
```suggestion
        addOption({{QStringLiteral("s"), QStringLiteral("show")}, QStringLiteral("Start with the main window visible, or if it is already running, bring it to the front. By default, the client launches in the background.")});
```

fmoc (approved):

HanaGemela:
@TheOneRing anything to test here?

HanaGemela:
Passed in ownCloud 5.0.0.11523-alpha.1

Issue #10751, opened 2023-04-21
Display Welcome page and spinner after an account was added


TheOneRing:
When a new account is added and that account is subscribed to many spaces the client will require a couple of seconds to create all sync folders.

During this time the client appears unresponsive.

Display a welcome page and a spinner, until we are done.

TheOneRing:
"Setting up account *spinner*"

fmoc:
Please do not play any audio from the widget.

HanaGemela:
There is an additional welcome page in ownCloud 5.0.0.11486-daily20230718, mac

<img width="742" alt="Screenshot 2023-07-18 at 10 29 33" src="https://github.com/owncloud/client/assets/49001702/a082f351-7565-4c01-b66e-9d97aa0dcdd9">

PR #11007, opened 2023-07-05
Implement progress spinner after wizard

TheOneRing:
image

Fixes: #10751

### DIFF
--- a/src/gui/accountsettings.ui
+++ b/src/gui/accountsettings.ui
@@ -50,22 +50,80 @@
     </widget>
    </item>
    <item>
-    <widget class="QTreeView" name="_folderList">
-     <property name="sizePolicy">
-      <sizepolicy hsizetype="MinimumExpanding" vsizetype="Expanding">
-       <horstretch>0</horstretch>
-       <verstretch>5</verstretch>
-      </sizepolicy>
+    <widget class="QStackedWidget" name="stackedWidget">
+     <property name="currentIndex">
+      <number>1</number>
      </property>
-     <property name="contextMenuPolicy">
-      <enum>Qt::CustomContextMenu</enum>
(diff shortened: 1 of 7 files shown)


TheOneRing:
The design is questionable and the spinner makes it more obvious that the ui might freeze..

src/gui/accountsettings.ui
```diff
@@ -50,22 +50,80 @@
...
+           <width>20</width>
+           <height>40</height>
+          </size>
+         </property>
+        </spacer>
+       </item>
+       <item>
+        <widget class="QLabel" name="label">
+         <property name="text">
+          <string>Getting ready</string>
```
fmoc, line 75: "Preparing account..." might be a more elegant wording. "Readying account" would work, too. We should avoid the use of "Get".

src/gui/owncloudgui.cpp
```diff
@@ -897,19 +898,21 @@ void ownCloudGui::runNewAccountWizard()
 
                                     folderMan->setSyncEnabled(true);
                                     folderMan->scheduleAllFolders();
+                                    accountStatePtr->setSettingUp(false);
```
fmoc, line 901: Just a question in general, would it be more elegant to use a setter like this or rather emit a signal on the account state pointer? I saw we need to track the state anyway in a local variable.

fmoc (changes requested):

src/gui/owncloudgui.cpp
TheOneRing: Emitting signals for other objects feels "wrong" additionally the setter has the benefit that we can query the state and not rely only on signals.

fmoc (approved):

Issue #10998, opened 2023-07-04
Crash reporter missing in testpilotcloud 5.0.0.11375-daily20230704


michaelstingl:
```
% /Applications/testpilotcloud.app/Contents/MacOS/testpilotcloud --debug
zsh: abort      /Applications/testpilotcloud.app/Contents/MacOS/testpilotcloud --debug
```

Then crash with `Debug actions > Crash now - qFatal`

```
% /Applications/testpilotcloud.app/Contents/MacOS/testpilotcloud --version
testpilotcloud testpilotcloud 5.0.0.11375-daily20230704
511b96
Libraries Qt 6.4.3, OpenSSL 1.1.1t  7 Feb 2023
Using virtual files plugin: suffix
macos-22.5.0
```

(M1 MacBook Air)

michaelstingl:
A new file was created:

```
% ls -l | grep testpilotcloud
-rw-r--r--   1 michaelstingl  staff    3339 Jul  4 11:28 testpilotcloud-crash.log
drwxr-xr-x  10 michaelstingl  staff     320 Jul  6 09:27 testpilotcloud-logdir
```

HanaGemela:
Testing blocked by https://github.com/owncloud/client/issues/11039

HanaGemela:
testpilotcloud 5.0.0.11480-daily20230717, Intel mac

```
Hanas-MBP:~ hana$ /Applications/testpilotcloud.app/Contents/MacOS/testpilotcloud --debug
Abort trap: 6
```

Apple report created instead 
<img width="1000" alt="Screenshot 2023-07-17 at 19 52 15" src="https://github.com/owncloud/client/assets/49001702/0f4e9f9b-fdc0-4fec-9bd1-9302f6438d2a">

TheOneRing:
Internal ci issue has been fixed

HanaGemela:
Fixed in testpilotcloud 5.0.0.11501-daily20230719

Issue #8991, opened 2021-09-09
Crash reporter for the cmd app


TheOneRing:
A small command line crash reporter supporting a non interactive mode for reporting

PR #11011, opened 2023-07-06
Fix crashreporter

TheOneRing:
This change also enables the crash reporter in the cmd client

Fixes: #10998
Issue: #8991

### DIFF
--- a/changelog/unreleased/8991
+++ b/changelog/unreleased/8991
@@ -1,6 +1,7 @@
-Enhancement: We enabled the crash reporter in the cmd app
+Enhancement: Enable crash reporter in commandline client
 
-The crash reporter is still a gui app so it will only work when ui is available but it will allows
-users to report crashes of the cmd much more easily.
+While the crash reporter still requires a graphical user interface to work, this
+change nevertheless simplifies reporting crashes from the commandline client
+a lot.
 
 https://github.com/owncloud/client/issues/8991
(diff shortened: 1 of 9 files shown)


TheOneRing:
Should I add a changelog entry?

src/crashreporter/main.cpp
```diff
@@ -24,10 +24,6 @@
 
 int main(int argc, char *argv[])
 {
-    QCoreApplication::setAttribute(Qt::AA_UseHighDpiPixmaps, true);
```
fmoc, line 27: Is this related to the change? Does this really fix the crash? If so, I'd like to know how (to avoid this in the future).

fmoc:
> Should I add a changelog entry?

If it's a fix, it probably needs one.

TheOneRing:
> > Should I add a changelog entry?
> 
> If it's a fix, it probably needs one.

Its a fix for a regression, but it also adds the (ui) crash reporter to the cmd client.

src/crashreporter/main.cpp
TheOneRing: The flags are deprecated.
I will create a follow up pr which will apply our compiler settings to the crash reporter.

fmoc:
Then I guess it needs two, because the latter is a feature addition and it's a user facing component.

TheOneRing:
It was a 5.0 regression so no need to create a changelog entry for that. 3b09bbc689aee55a24e744e6614c916036d5f310

fmoc (approved):

PR #11015, opened 2023-07-07
[CI] update secret keys

saw-jan:
Some of the secret keys have changed. Updated with the new ones.


.drone.star
```diff
@@ -36,6 +36,17 @@ PYTHON = "python"
 THEGEEKLAB_DRONE_GITHUB_COMMENT = "thegeeklab/drone-github-comment:1"
 TOOLHIPPIE_CALENS = "toolhippie/calens:latest"
 
+secrets = {
+    "SQUISH_LICENSEKEY": "squish_license_server",
+    "GITHUIB_USERNAME": "github_username",
```
fmoc, line 41: Typo.

.drone.star
```diff
@@ -36,6 +36,17 @@ PYTHON = "python"
 THEGEEKLAB_DRONE_GITHUB_COMMENT = "thegeeklab/drone-github-comment:1"
 TOOLHIPPIE_CALENS = "toolhippie/calens:latest"
 
+secrets = {
+    "SQUISH_LICENSEKEY": "squish_license_server",
```
fmoc, line 40: I'm not sure we should expose this secret in PRs either... I think we could use the `push` trigger, too, to have it run on our own PRs.

.drone.star
```diff
@@ -36,6 +36,17 @@ PYTHON = "python"
 THEGEEKLAB_DRONE_GITHUB_COMMENT = "thegeeklab/drone-github-comment:1"
 TOOLHIPPIE_CALENS = "toolhippie/calens:latest"
 
+secrets = {
```
fmoc, line 39: What is this indirection good for? To have an overview of the used secrets? There should be a comment that explains why this is needed.

.drone.star
saw-jan: Yeah, that would be great. I have no idea how that can be done.
CC @xoxys

xoxys: Well... not that simple because then you will have two jobs running for every or the one for the PR `/ref/pull/**` itself and `/ref/head/**`... If you disable `/ref/pull/**` pull requests e.g. from external contributors via fork will not work but in that case you can just disable forks in drone.

fmoc: > disable forks in drone

That sounds totally acceptable to me. We should enable that for most secrets.

.drone.star
fmoc: Well, if a secret changes, you delete it in the backend and create a new one of the same name, don't you?

xoxys: No, see current changes...

.drone.star
xoxys: Feel free to do so. As a repo admin, you can change this in the Drone UI for your repos on your own.

saw-jan: I can do `disable forks` in drone if we want it

saw-jan: Better we disable forks. It is possible to cleverly print the secret value
https://github.com/owncloud/enterprise/issues/5832#issuecomment-1628221384

.drone.star
```diff
@@ -603,25 +605,15 @@ def setGuiTestReportDir():
...
-
 def uploadGuiTestLogs(ctx, server_type = "oc10"):
     trigger = {
         "status": [
             "failure",
         ],
+        "event": [
+            "cron",
+            "tag",
+        ],
```
saw-jan, line 616: we may want reports on PR failures to know exactly what failed. So instead of gh-comment, adding ci step to log reports url might be useful.

.drone.star
fmoc: @xoxys I cannot log in to https://drone.owncloud.com. Who do I need to ask for access?

.drone.star
```diff
@@ -387,7 +390,7 @@ def notification():
 
     for channel, params in notify_channels.items():
         settings = {
-            "webhook": from_secret("private_rocketchat"),
+            "webhook": from_secret("ROCKETCHAT_WEBHOOK"),
```
fmoc, line 393: Does anybody still use or depend on Rocket.Chat notifications? If not, we might just get rid of this altogether.

fmoc (commented):
Thanks, @saw-jan, looks pretty good to me. Let's fix the secret visibility, then we should be able to merge your changes.

.drone.star
saw-jan: > Does anybody still use or depend on Rocket.Chat notifications? If not, we might just get rid of this altogether.

Nightlies are usually reported to chat. Also, when tag/release is done to give the confirmation that everything went well.

fmoc (approved):
Looks good. I disabled forks in Drone's settings, which should prevent people from extracting secrets quite effectively.

fmoc:
Thanks @saw-jan!

PR #11019, opened 2023-07-07
Enable sanatizers on for mac debug builds

TheOneRing:


.github/workflows/main.yml
```diff
@@ -96,7 +101,10 @@ jobs:
 
     - name: Build
       run: |
-        if ("${{ matrix.target }}" -eq "windows-msvc2022_64-cl") {
+        if ("${{ matrix.target }}" -eq "macos-64-clang-debug" ) {
+            & "${env:GITHUB_WORKSPACE}/.github/workflows/.craft.ps1" -c --set args="-DECM_ENABLE_SANITIZERS=`"address;leak;undefined`"" owncloud/owncloud-client
```
fmoc, line 105: Is this backtick escaping _really_ how PowerShell does it?

.github/workflows/main.yml
```diff
@@ -28,21 +28,26 @@ jobs:
...
         matrix:
             include:
-            - name: Windows 64-bit cl
-              target: windows-cl-msvc2022-x86_64
-              os: windows-latest
-              fetch-depth: 0
-              container:
-            - name: macOS 64-bit clang
-              target: macos-64-clang
+#            - name: Windows 64-bit cl
```
fmoc, line 31: This is just for testing, I assume?

fmoc (changes requested):
You may want to fix the second commit message to allow for grepping for the term "sanitizer" in the history.

.github/workflows/main.yml
TheOneRing: 🙄

fmoc (approved):

test/testoauth.cpp
```diff
@@ -122,7 +122,7 @@ class OAuthTestCase : public QObject
     QScopedPointer<OAuth> oauth;
 
     virtual void test() {
-        fakeAm = new FakeAM({});
+        fakeAm = new FakeAM({}, nullptr);
```
fmoc, line 125: Why do you need to specify this explicitly? Can we make `nullptr` the default argument?

test/testutils/testutils.cpp
```diff
@@ -106,5 +105,18 @@ namespace TestUtils {
...
+
+
+    AccountManagerRaii::AccountManagerRaii(const AccountStatePtr &accountStatePtr)
+        : _accountStatePrt(accountStatePtr)
+    {
+    }
+
+    AccountManagerRaii::~AccountManagerRaii()
+    {
+        if (_accountStatePrt) {
```
fmoc, line 117: Typo: `_accountStatePtr`.

Seems like a smart pointer would do the job just fine if you specify the deleter accordingly.

fmoc (changes requested):

test/testoauth.cpp
TheOneRing: I prefer to force the user

Issue #10619, opened 2023-03-21
Always send a language header in all requests


michaelstingl:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Feature description

Latest oCIS notifications API (`/ocs/v2.php/apps/notifications/api/v1/notifications?format=json`) responds with translated content. Other endpoints will learn this too.

### Proposed solution

Desktop app should always send a language header in all requests which follows the Accept-Language spec:
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language

### Additional context

Related:
- Web: https://github.com/owncloud/web/issues/8612
- iOS: https://github.com/owncloud/ios-app/issues/1193
- Android: https://github.com/owncloud/android/issues/3980

michaelstingl:
> I would only enable that on a per endpoint basis, I don't know where we parse random strings and thus "only speak English"

Better send same to all endpoints.

fmoc:
Note: @TheOneRing thinks there might be at least one location where we actually parse an error message from the server (somewhere in the common error handling). We may need to either replace this with a more reliable solution (if possible) or force an English locale in `Accept-Language` for these requests explicitly.

TheOneRing:
https://github.com/owncloud/client/blob/e4ae0539664dbf1e9d3c522a7ec4bc41aebd4cb3/src/libsync/owncloudpropagator_p.h#L78-L87

HanaGemela:
Passed ownCloud 5.0.0.11523-alpha.1

PR #11020, opened 2023-07-07
Always send a language header in all requests

erikjv:
Fixes: #10619

### DIFF
--- a/src/libsync/accessmanager.cpp
+++ b/src/libsync/accessmanager.cpp
@@ -62,6 +62,9 @@ QNetworkReply *AccessManager::createRequest(QNetworkAccessManager::Operation op,
     // Some firewalls reject requests that have a "User-Agent" but no "Accept" header
     newRequest.setRawHeader(QByteArrayLiteral("Accept"), QByteArrayLiteral("*/*"));
 
+    // Set the language, so messages from the server are localised correctly.
+    newRequest.setRawHeader("Accept-Language", QLocale().name().toUtf8());
+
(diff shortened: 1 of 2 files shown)


src/libsync/accessmanager.cpp
```diff
@@ -62,6 +62,16 @@ QNetworkReply *AccessManager::createRequest(QNetworkAccessManager::Operation op,
     // Some firewalls reject requests that have a "User-Agent" but no "Accept" header
     newRequest.setRawHeader(QByteArrayLiteral("Accept"), QByteArrayLiteral("*/*"));
 
+    // Set the language, so messages from the server are localised correctly.
+    QLocale locale;
+    QByteArray acceptedLanguage;
+    if (locale.language() == QLocale::AnyLanguage) {
```
TheOneRing, line 68: Can the clients locale really be "AnyLanguage"?

fmoc: I assume since this is the `0` value of the enum, it's the fallback. However, I don't see a point in handling this case either, since I'd consider *not* having a valid locale set globally an error. This would only make sense if the language was a property on the request.

erikjv: ok, removed it.

TheOneRing (changes requested):
Changelog please

fmoc (approved):

fmoc (changes requested):
Indeed, changelog entry missing.

PR #10881, opened 2023-05-26
Remove support to use non standard plugin dir

TheOneRing:

PR #11023, opened 2023-07-10
Fix persisting filter settings in activity tab

erikjv:
There were two issues: first, the StatusSet was not initialised, thus containing random values. Second, no difference was made between an empty filter list, and an absent filter list (the latter would cause the filter to be reset). Both are fixed now.

Fixes: #10881

### DIFF
--- a/src/gui/issueswidget.cpp
+++ b/src/gui/issueswidget.cpp
@@ -145,19 +145,23 @@ private:
 
     void restoreFilter()
     {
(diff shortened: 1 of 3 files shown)


TheOneRing (approved):

Salipa-Gurung:
Tested in https://github.com/owncloud/client/issues/10884#issuecomment-1639851501

Issue #10596, opened 2023-03-15
Wizard: Implement selective sync for spaces


HanaGemela:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Feature description

Selective sync should be available when adding a new space on mac

### Proposed solution

User is offered a list of folders in space to sync

michaelstingl:
Yes, could be similar to "add folder" process from oC10 when adding a new Space

![CleanShot 2023-03-16 at 10 06 24@2x](https://user-images.githubusercontent.com/214010/225568423-19f67262-eba4-4466-936c-69761cbb4829.png) | ![CleanShot 2023-03-16 at 10 06 56@2x](https://user-images.githubusercontent.com/214010/225568445-6f048893-9c88-4670-bfa2-de4b6ca12333.png) | ![CleanShot 2023-03-16 at 10 07 04@2x](https://user-images.githubusercontent.com/214010/225568458-f1cd7945-30a2-47bb-bdf8-ef4f5b1b35cb.png) | ![CleanShot 2023-03-16 at 10 07 26@2x](https://user-images.githubusercontent.com/214010/225568470-77be60a9-07b5-47a1-9774-4b3e3cc93acb.png)
-- | -- | -- | ---
(1) | (2) | (3) Not planned for Spaces | (4) would be nice to have for Spaces 👍

HanaGemela:
Selective sync is available in ownCloud 5.0.0.11523-alpha.1, mac

HanaGemela:
This message about "Unchecked folders..." takes up now a lot of space. Which I like as it makes is more obvious that there is some action required. But it wasn't this way so I'm just making sure it was changed on purpose - @fmoc @TheOneRing 

<img width="960" alt="Screenshot 2023-07-24 at 10 00 07" src="https://github.com/owncloud/client/assets/49001702/dba04d0b-404c-466e-b396-56df0182f8fd">

TheOneRing:
> This message about "Unchecked folders..." takes up now a lot of space. Which I like as it makes is more obvious that there is some action required. But it wasn't this way so I'm just making sure it was changed on purpose - @fmoc @TheOneRing
[...]

PR #11044, opened 2023-07-20
Allow selective sync of spaces in folder wizard

fmoc:
Closes https://github.com/owncloud/client/issues/10596.

### DIFF
--- a/src/gui/selectivesyncwidget.cpp
+++ b/src/gui/selectivesyncwidget.cpp
@@ -101,8 +101,7 @@ QSize SelectiveSyncWidget::sizeHint() const
 
 void SelectiveSyncWidget::refreshFolders()
 {
-    Q_ASSERT(!_davUrl.isEmpty());
-    auto *job = new PropfindJob(_account, _davUrl, _folderPath, PropfindJob::Depth::One, this);
+    auto *job = new PropfindJob(_account, getDavUrl(), _folderPath, PropfindJob::Depth::One, this);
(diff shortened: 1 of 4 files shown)


src/gui/selectivesyncwidget.cpp
```diff
@@ -263,8 +264,8 @@ void SelectiveSyncWidget::slotItemExpanded(QTreeWidgetItem *item)
     QString dir = item->data(0, Qt::UserRole).toString();
     if (dir.isEmpty())
         return;
-    // TODO: legacy
-    PropfindJob *job = new PropfindJob(_account, _account->davUrl(), _folderPath + dir, PropfindJob::Depth::One, this);
+    Q_ASSERT(!_davUrl.isEmpty());
```
TheOneRing, line 267: How about a getter with that check?

fmoc: Would save a tiny bit of code duplication, but we cannot enforce the use of that getter over the variable anyway...

TheOneRing: There are private getters :D
And in general I'd say we should always use getters instead of raw vars.

fmoc: Well, if you prefer it...

TheOneRing: Qt codestyle does not use get as getter prefix

TheOneRing:
Can you check whether the ignore list computes the correct patterns?
image
Hidden files should actually not appear afaik.

TheOneRing (approved):

Issue #9117, opened 2021-10-05
Investigate removal of unused `Multifolders` settings


TheOneRing:
`Multifolder` was introduced in https://github.com/owncloud/client/commit/61b4da944c6d45332de3ebadbd6a80985719109a#diff-4b086fafd9fc04284b2316ff9b3a2cdc787fbb8ae8d0b296570ef02d067ec311R612

The whole structure of the settings should be redesigned.
Possible overlap: https://github.com/owncloud/client/issues/9128

TODO:
- Investigate why we still create new folders with the `_saveBackwardsCompatible` mode
- Find out if there are scenarios where we have  `Multifolder` settings
- If possible remove `Multifolder` from the already complex settings code

fmoc:
Investigation time boxed to 3 hours. Should you exceed this limit, we need to re-evaluate the problem in the next sprint.

erikjv:
The best "explanation" is in a comment that I removed in #10982 :
```
        // The folder is saved to backwards-compatible "Folders"
        // section only if it has the migrate flag set (i.e. was in
        // there before) or if the folder is the only one for the
        // given target path.
        // This ensures that older clients will not read a configuration
        // where two folders for different accounts point at the same
        // local folders.
```
So I think we can remove the saving, but keep on loading it for the time being.

erikjv:
After chatting with @TheOneRing , this comment refers to a feature that we do not support anymore: having 2 accounts syncing to the same folder, which thereby keeps 2 servers in sync. We don't support that, so any references to `Multifolders` can be removed.

PR #11053, opened 2023-07-24
Remove saving of legacy `Multifolders` settings entry

erikjv:
The feature that this settings group relates to is not supported for quite some time.

Fixes: #9117


TheOneRing:
Was introduced in https://github.com/owncloud/client/commit/61b4da944c6d45332de3ebadbd6a80985719109a
I've never ever seen a config file with multifolder keys.
I guess we migrated them the hard way when we complained when multiple db's where located in the same folder.

TheOneRing (approved):

PR #11065, opened 2023-07-27
Fix account check marks in filter menu

erikjv:
Fixes finding from:
- https://github.com/owncloud/client/issues/10884#issuecomment-1649818735

Also fixes interaction with the caller side of the method populating the menu.

### DIFF
--- a/src/gui/models/models.cpp
+++ b/src/gui/models/models.cpp
@@ -110,9 +110,6 @@ std::function<void()> OCC::Models::addFilterMenuItems(QMenu *menu, const QString
     for (const auto &c : candidates) {
         addAction(c, c);
     }
-    QTimer::singleShot(0, menu, [menu] {
-        menu->popup(QCursor::pos());
-    });
 
     auto resetFunction = [noFilter]() {
         noFilter->setChecked(true);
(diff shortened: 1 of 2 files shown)


fmoc:
Please provde a meaningful title and a description. If there are UI changes, please add screenshots.

fmoc (changes requested):
See above.

erikjv:
The separate commits have a useful description, but I can put them in the PR.

src/gui/models/models.cpp
```diff
@@ -99,9 +99,7 @@ std::function<void()> OCC::Models::addFilterMenuItems(QMenu *menu, const QString
             model->setFilterFixedStringSignalled(filter);
         });
         action->setCheckable(true);
-        if (currentFilter == filter) {
-            action->setChecked(true);
-        }
+        action->setChecked(currentFilter == QRegularExpression::escape(filter));
```
fmoc, line 102: The logic change is not explained by the commit message. You could have compared the value to the escaped expression in the `if`, too.

fmoc (changes requested):

src/gui/models/models.cpp
TheOneRing: Most probably Qt6 porting fallout.
The regex filter api changed for Qt6.

erikjv: It's shorter and does the same thing.

fmoc: > does the same thing

Well, not quite. We have never explicitly set `false` before. It apparently works fine in this case, but I wouldn't assume this as the default in general.

Anyway, good enough for me.

fmoc (approved):

PR #11066, opened 2023-07-27
Fix KDSingleApplication related startup bug

fmoc:
In order to notify users of the traditional Linux packages (which are usually updated by the package managers as they are shipped from repositories and therefore lack integrated updater), the updater regularly called itself with `--version` to recognize whether its own version changed (i.e., the package was updated in the background while the client was running).

Due to a change in the commandline parsing, this approach no longer works as `--version` is evaluated only after KDSingleApplication was initialized, causing the client to terminate after a few seconds and therefore rendering it unusable.

The new approach focuses on the application binary's `mtime`. Whenever it changed compared to the timestamp we initially saved when the client started, we assume the binary has changed, i.e., the package was updated in the background.

This approach does not attempt to discover and prevent downgrades. Any change causes a notification to the user, prompting them to restart.

### DIFF
--- a/src/common/utility.cpp
+++ b/src/common/utility.cpp
@@ -362,33 +362,6 @@ void Utility::crash()
     *a = 1;
 }
 
-// read the output of the owncloud --version command from the owncloud
-// version that is on disk. This works for most versions of the client,
-// because clients that do not yet know the --version flag return the
-// version in the first line of the help output :-)
-//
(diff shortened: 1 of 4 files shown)


TheOneRing (approved):

fmoc:
Let's add some logging to this, shall we?

TheOneRing (approved):

PR #11086, opened 2023-08-04
wip

TheOneRing:


src/gui/accountstate.cpp
```diff
@@ -166,6 +166,7 @@ std::unique_ptr<AccountState> AccountState::loadFromSettings(AccountPtr account,
         accountState->setState(SignedOut);
     }
     accountState->_supportsSpaces = settings.value(supportsSpacesC(), false).toBool();
+    // accountState->account()->setCredentials(new HttpCredentialsGui(accountState.get(), l));
```
erikjv, line 169: Why is this commented-out line here?

erikjv (changes requested):
I'm not very familiar with the authentication code. If this should be reviewed soonish, I'd appreciate a call with a walk-through of the changes.

Issue #10351, opened 2022-12-15
Login dialog loop when adding an HTTP basic auth account


fmoc:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the bug

When adding a new account, logging out is not possible right after the initial sync has finished. When hitting the logout button, the client will sign the user out, but immediately requests the user to log in again. When trying to log out in the dialog, the dialog closes but is immediately reopened, creating an infinite loop.

When logged in successfully a second time, signing out works normally as expected.

This bug only shows when using a server that uses HTTP basic auth, e.g., https://demo.owncloud.org (OAuth2/OIDC are not affected), and only on Windows. I checked various Windows 10 and 11 VMs. Ubuntu doesn't seem to be affected.

### Expected behavior

The client should just let a user log out normally using the button in the account settings.

Same goes for the button in the "login required" dialog.

### Steps to reproduce the issue

1. Use the wizard to add an account on any server utilizing HTTP basic authentication (e.g., https://demo.owncloud.org) on some Windows 10/11
2. Let the sync finish (it is currently unknown whether this is needed, though)
3. Sign out using logout button in the account settings dialog
4. You will see the "login required" dialog in an infinite loop

### Screenshots

![screenshot_2022-12-15_14-40-41](https://user-images.githubusercontent.com/80399010/207874382-5fb47e27-9a50-42a2-8208-6a799f2d6ad4.png)

![screenshot_2022-12-15_14-41-12](https://user-images.githubusercontent.com/80399010/207874396-80ecec19-0784-4f96-b3bd-0b623eeb639d.png)

![screenshot_2022-12-15_14-42-44](https://user-images.githubusercontent.com/80399010/207874420-1bd8928c-4134-469a-90db-db7d0a006ff4.png)

### Logs

First time trying to log out:

```
22-12-15 12:18:26:506 [ info gui.account.state ]:	AccountState state change:  "Verbunden" -> "Abgemeldet"
22-12-15 12:18:26:506 [ debug gui.activity ]	[ OCC::ActivitySettings::slotRefresh ]:	Do not check as last check is only secs ago:  3
[...]

PR #11090, opened 2023-08-04
Fix the login dialog appearing for logged out accounts

TheOneRing:
I don't think this was the original looping issue but a new one introduced with the connecting state in 4.0

Fixes: #10351

### DIFF
--- a/changelog/unreleased/10351
+++ b/changelog/unreleased/10351
@@ -1,3 +1,3 @@
-Bugfix: Authenticaiton dialog appears again and again
+Bugfix: Authentication dialog no longer appears again and again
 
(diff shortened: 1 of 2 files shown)


changelog/unreleased/10351
```diff
@@ -0,0 +1,3 @@
+Bugfix: Authenticaiton dialog appears again and again
```
fmoc, line 1:
```suggestion
Bugfix: Authentication dialog no longer appears again and again
```

fmoc (changes requested):

fmoc (approved):

PR #11092, opened 2023-08-07
Add "reset to default" button to local sync dir

### LINKED ISSUE #11078: Add "reset to default" UI element for local sync dir to last setup wizard page
### Pre-submission Checks

### Feature description

Currently, once the user changes the local sync dir, the value will be stored and shown every time the page is displayed. There is no way for the user to reset the value at the moment, they need to start over completely.

### Proposed solution

A "reset button" next to the line edit was suggested so far. Perhaps Qt provides us with some nicer option.

_No response_

### Additional context

_No response_

fmoc:
Fixes #11078.

@TheOneRing suggested to extend this to reset the entire advanced configuration section. @michaelstingl what do you think? In my opinion, this is good enough, as we do not store the preferred synchronization method anyway when the page is left (i.e., someone clicks on "back").

screenshot_2023-08-07_11-47-16


michaelstingl (approved):
Nice 🎉

erikjv (approved):

PR #11106, opened 2023-08-09
Folders will shedule themselves after vfs changed

TheOneRing:
The current code would schedule the folder before its ready and thus trigger an assert.


fmoc (approved):

Salipa-Gurung:
How can it be tested ?

PR #11109, opened 2023-08-11
Fix isChildPathOf QDir::cleanPath removes trailing slashes

TheOneRing:


src/common/filesystembase.cpp
```diff
@@ -621,7 +621,9 @@ bool FileSystem::isChildPathOf(QStringView child, QStringView parent)
     }
 
     // Slow path (`QDir::cleanPath` does lots of string operations):
-    return QString::compare(QDir::cleanPath(parent.toString()), QDir::cleanPath(child.toString()), sensitivity) == 0;
+    // cleanPath removes trailing slashes, add one to parent to be sure we handle a child path
+    // root/foo/bar is not  a child of /root/fo, therefor the trailing slash is important
+    return QDir::cleanPath(child.toString()).startsWith(QDir::cleanPath(parent.toString()) + QLatin1Char('/'), sensitivity);
```
fmoc, line 626: Looks like adding a slash to both of them works: for `/a/b/c/` for child and path as an input, I get the correct result from this snippet:

```cpp
    const QString parent("/a/b/c/");
    const QString child("/a/b/c/");
    qDebug() << QDir::cleanPath(child).startsWith(QDir::cleanPath(parent), Qt::CaseSensitivity::CaseSensitive);

**Review by fmoc — CHANGES_REQUESTED** [review]

File: src/common/filesystembase.cpp (hunk 1 above)
Comment (fmoc) [reply src/common/filesystembase.cpp:626]: Needs an exception for files: `!parent.isFile() && [...]`. Files can never have children.

**Review by fmoc — APPROVED** [review]

**Comment by TheOneRing** [comment]
Failing test case https://github.com/owncloud/client/pull/11110

File: src/common/filesystembase.cpp [hunk 2]
```diff
@@ -613,15 +617,26 @@ bool FileSystem::isChildPathOf(QStringView child, QStringView parent)
...
             // else: do the `cleanPath` version below
         }
     }
 
     // Slow path (`QDir::cleanPath` does lots of string operations):
-    return QString::compare(QDir::cleanPath(parent.toString()), QDir::cleanPath(child.toString()), sensitivity) == 0;
+    const QString cleanParent = QDir::cleanPath(parent.toString());
+    const QString cleanChild = QDir::cleanPath(child.toString());
+    // cleanPath removes trailing slashes, add one to parent to be sure we handle a child path
+    // root/foo/bar is not  a child of /root/fo, therefor the trailing slash is important
```
erikjv, line 631: nitpick: `/root/foo`

erikjv (approved):

PR #11113, opened 2023-08-11
SyncEngine: Don't expose ExcludedFiles

TheOneRing:


erikjv (approved):

TheOneRing:
The second commit is unrelated but might have fix a failure in the squish test

Issue #10964, opened 2023-06-23
Prevent use of older clients with already-upgraded databases


fmoc:
Follow-up from https://github.com/owncloud/client/issues/10473. Needs to be worked on once the former one is merged.

https://github.com/owncloud/client/blob/cde9c61e5aa5c2e1661ea5a6d225b2e1752d487b/src/common/syncjournaldb.cpp#L523-L523

fmoc:
A new version check is required after the config file check mentioned in #10473. We may be able to use the same dialog if we implement this check in between.

erikjv:
To do:

1. add a DB version check in `FolderMan::setupFoldersHelper`, and make it fail-able. 
2. move `FolderMan::instance()->setupFolders();` out of the `Application` constructor, because it can fail (together with showing the wizard if no accounts exist)

For 1. above: do this right after the call to `SyncJournalDb::maybeMigrateDb`? Maybe create a temporary instance of `SyncJournalDb` just to do a version check?

erikjv:
@TheOneRing ^

TheOneRing:
> To do:
> 
>     1. add a DB version check in `FolderMan::setupFoldersHelper`, and make it fail-able.
> 
>     2. move `FolderMan::instance()->setupFolders();` out of the `Application` constructor, because it can fail (together with showing the wizard if no accounts exist)
> 
> 
> For 1. above: do this right after the call to `SyncJournalDb::maybeMigrateDb`? Maybe create a temporary instance of 

Sounds good, `SyncJournalDb::versionCheck` (We are again not migrating).

>`SyncJournalDb` just to do a version check?

Salipa-Gurung:
How can I confirm that it prevents use of older clients with already-upgraded databases?

erikjv:
Go to a sync root, then on the command line:

```
sqlite3 .sync_journal.db
sqlite> select major from version;
5
sqlite> update version set major=6 where major=5;
sqlite> select major from version;
6
```

Salipa-Gurung:
### Desktop Client Version
ownCloud 5.0.0.11804-daily20230818 [65017f](https://github.com/owncloud/client/commit/65017f978ed45f3f2ff0e639c367bb63d6e7e333)
ubuntu-6.2.0-26-generic

#### With upgraded database user is prevented from using older client version :heavy_check_mark: 

[...]

PR #11114, opened 2023-08-11
Prevent use of older clients with already-upgraded databases

erikjv:
Fixes: #10964

### DIFF
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -192,6 +192,27 @@ bool SyncJournalDb::maybeMigrateDb(const QString &localPath, const QString &abso
     return true;
 }
 
+bool SyncJournalDb::dbIsTooNewForClient(const QString &dbFilePath)
+{
+    SqlDatabase db;
+    if (!db.openReadOnly(dbFilePath)) {
+        // No DB, we're not downgrading, so we're fine
+        return false;
+    }
+
+    SqlQuery versionQuery("SELECT major, minor, patch FROM version;", db);
+    if (!versionQuery.next().hasData) {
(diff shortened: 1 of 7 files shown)


src/common/syncjournaldb.cpp
```diff
@@ -192,6 +192,27 @@ bool SyncJournalDb::maybeMigrateDb(const QString &localPath, const QString &abso
     return true;
 }
 
+bool SyncJournalDb::checkVersion(const QString &dbFilePath)
```
TheOneRing, line 195: Why is naming things so hard, I suggested checkVersion but I'm not happy 🙈 
Maybe checkDowngrade? In which case we would need to invert the condition.

TheOneRing (changes requested):

src/gui/main.cpp
```diff
@@ -338,6 +344,17 @@ int main(int argc, char **argv)
         platform->setApplication(&app);
 
         auto ocApp = new OCC::Application(platform.get(), options.debugMode, &app);
+
+        if (!FolderMan::instance()->setupFolders().has_value()) {
```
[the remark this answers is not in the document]
erikjv: it does, see the comment I added

src/common/syncjournaldb.cpp
erikjv: Done, but is the version check correct? Meaning, do we only want to check the major version?

TheOneRing: As minor version changes are generally mean that there where no breaking changes I'd hope comparing the major version is sufficient.

TheOneRing (approved):

Issue #10960, opened 2023-06-23
addStatusFilter handles enum values in switch/case to enable compiler warnings for unhandled values


TheOneRing:
https://github.com/owncloud/client/blob/7668e0aef5f110412215f0fd7fc5f40ef26b9f33/src/gui/issueswidget.cpp#L383-L387

erikjv:
@TheOneRing can you elaborate a bit please?

TheOneRing:
At the time I opened the issue we discovered that that we forgot to handle a new state.
If the code would be refactored to iterate over all items and handle them in a switch statement, we would get warnings (now errors) when there is an unhandled case.

PR #11115, opened 2023-08-11
Make SyncFileItem::Status more robust to changes

erikjv:
The GUI uses this to display and filter status messages. To make sure we have all displayable status items listed, they are now in `SyncFileItem` together with a static assert to check we covered all items. This should make adding/removing status items less error-prone.

Fixes: #10960

### DIFF
--- a/src/libsync/syncfileitem.h
+++ b/src/libsync/syncfileitem.h
@@ -51,9 +51,6 @@ public:
     Q_ENUM(Direction)
 
     enum Status : uint8_t { // stored in 4 bits
-        // IMPORTANT: if something changes in this enum, please check `IssuesWidget::addStatusFilter`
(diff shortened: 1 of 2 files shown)


src/libsync/syncfileitem.h
```diff
@@ -113,6 +110,27 @@ class OWNCLOUDSYNC_EXPORT SyncFileItem
...
+        Restoration,
+        BlacklistedError,
+        Excluded,
+        Message,
+        FilenameReserved,
+    };
+
+    // We don't include NoStatus or Success in here
+    static_assert(
+        ErrorStatusItems.size() + OtherDisplayableStatusItems.size() == StatusCount - 2, "ErrorStatusItems or OtherDisplayableStatusItems is incomplete");
```
TheOneRing, line 132: ❤️

PR #11116, opened 2023-08-11
Reduce the amount of compiler warnings

TheOneRing:


src/gui/folderman.cpp
```diff
@@ -103,13 +103,12 @@ FolderMan::FolderMan(QObject *parent)
...
         this, &FolderMan::slotRemoveFoldersForAccount);
 
-    connect(_lockWatcher.data(), &LockWatcher::fileUnlocked,
-        this, [this](const QString &path, FileSystem::LockMode mode) {
-            if (Folder *f = folderForPath(path)) {
-                // Treat this equivalently to the file being reported by the file watcher
-                f->slotWatchedPathsChanged({path}, Folder::ChangeReason::UnLock);
-            }
-        });
+    connect(_lockWatcher.data(), &LockWatcher::fileUnlocked, this, [this](const QString &path, FileSystem::LockMode) {
```
fmoc, line 106: See above.

src/gui/folder.cpp
```diff
@@ -672,7 +672,7 @@ void Folder::slotWatchedPathsChanged(const QSet<QString> &paths, ChangeReason re
                 Q_ASSERT([&] {
                     Q_ASSERT(record.isValid());
                     // we don't intend to burn to many cpu cycles so limit this check on small files
-                    if (!record.isVirtualFile() && record._fileSize < 1_mb) {
+                    if (!record.isVirtualFile() && record._fileSize < static_cast<qint64>(1_mb)) {
```
fmoc, line 675: Wouldn't it make sense to cast the `_fileSize` property or make it the right type right away? (Just a small nitpick, this is fine as-is, too).

shell_integration/windows/OCContextMenu/OCContextMenu.cpp
```diff
@@ -219,7 +219,7 @@ IFACEMETHODIMP OCContextMenu::InvokeCommand(LPCMINVOKECOMMANDINFO pici)
         return E_FAIL;
     }
 
-    OCClientInterface::SendRequest(command, m_selectedFiles);
+    std::ignore = OCClientInterface::SendRequest(command, m_selectedFiles);
```
fmoc, line 222: Didn't know this one, nice! Way more straightforward than a void cast.

src/common/filesystembase.cpp
```diff
@@ -161,16 +161,14 @@ bool FileSystem::rename(const QString &originFileName,
     bool success = false;
     QString error;
 #ifdef Q_OS_WIN
-    const QString orig = longWinPath(originFileName);
+    const QString originalFileNameLong = longWinPath(originFileName);
     const QString dest = longWinPath(destinationFileName);
```
fmoc, line 165: You may want to implement a consistent pattern (the previous variable is called `...Long` whereas this one isn't although you call that function on both anyway).

fmoc (changes requested):
I've got a few comments. I reviewed commit by commit.

(You may want to fix the commit message of 2d47068.)

PR #11117, opened 2023-08-14
Change local collision to follow the ownCloud convention

erikjv:
If the default folder name exists, a different one was offered in the form of "ownCloud2". This is now changed to follow the ownCloud convention, and "ownCloud (2)" is now offered.

### DIFF
--- a/src/gui/folderman.cpp
+++ b/src/gui/folderman.cpp
@@ -786,8 +786,8 @@ QString FolderMan::checkPathValidityForNewFolder(const QString &path) const
 
 QString FolderMan::findGoodPathForNewSyncFolder(const QString &basePath, const QString &newFolder) const
 {
-    // reserve 3 characters to allow appending of a number 0-100
-    const QString normalisedPath = FileSystem::createPortableFileName(basePath, FileSystem::pathEscape(newFolder), 3);
+    // reserve extra characters to allow appending of a number
+    const QString normalisedPath = FileSystem::createPortableFileName(basePath, FileSystem::pathEscape(newFolder), std::string_view(" (100)").size());
 
     // If the parent folder is a sync folder or contained in one, we can't
     // possibly find a valid sync folder inside it.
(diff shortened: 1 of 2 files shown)


src/gui/folderman.cpp
```diff
@@ -786,8 +786,8 @@ QString FolderMan::checkPathValidityForNewFolder(const QString &path) const
 
 QString FolderMan::findGoodPathForNewSyncFolder(const QString &basePath, const QString &newFolder) const
 {
-    // reserve 3 characters to allow appending of a number 0-100
-    const QString normalisedPath = FileSystem::createPortableFileName(basePath, FileSystem::pathEscape(newFolder), 3);
+    // reserve 3 characters to allow appending of a number 0-100, plus 2 for " (" and 1 for ")"
```
TheOneRing, line 789: we could use `std::stringview(" (100)").size()` to have a compile time placeholder?

src/gui/folderman.cpp
```diff
@@ -801,12 +801,12 @@ QString FolderMan::findGoodPathForNewSyncFolder(const QString &basePath, const Q
...
     {
         QString folder = normalisedPath;
-        for (int attempt = 2; attempt < 100; ++attempt) {
+        for (int attempt = 2; attempt <= 100; ++attempt) {
             if (!QFileInfo::exists(folder)
                 && FolderMan::instance()->checkPathValidityForNewFolder(folder).isEmpty()) {
                 return canonicalPath(folder);
             }
-            folder = normalisedPath + QString::number(attempt);
+            folder = normalisedPath + QStringLiteral(" (") + QString::number(attempt) + QStringLiteral(")");
```
TheOneRing, line 809:
```suggestion
            folder = normalisedPath + QStringLiteral(" (%1)").arg(QString::number(attempt));
```
?

TheOneRing:
Looks like we have unit tests for this :D

src/gui/folderman.cpp
erikjv: done

src/gui/folderman.cpp
erikjv: done

TheOneRing (approved):

saw-jan:
Tested with `ownCloud 5.0.0.11734-daily20230816 9e8820`

Screenshot from 2023-08-16 18-39-28

PR #11120, opened 2023-08-15
Fix warnings in tests

erikjv:


TheOneRing:
I was faster https://github.com/owncloud/client/pull/11116 (but it kind of exploded in its expansiveness )

Issue #10919, opened 2023-06-13
Default sync root directory created every time accounts are loaded


fmoc:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the bug

See https://central.owncloud.org/t/default-folder-created-at-each-start/44101.

The default sync root directory is created over and over again if set.

### Expected behavior

It makes no sense to create the directory every time. Users may not even use it at all but use a totally different base directory to synchronize their files.

Possible solutions:
- create sync root directory only once after the wizard finishes (if synchronization is to be configured with the folder wizard)
- just don't create it unless explicitly required

I am in favor with the latter.

### Client version number

4.0.0.10896 on Ubuntu 22.04.

HanaGemela:
@fmoc could you specify the steps to recreate?
I've tried with 4.0.0 and Ubuntu 22.04
1. Add a new account
2. Log out
3. Log back in
4. Check the file explorer -> Just one owncloud folder

fmoc:
1. Add new account with a non-default sync root (you need to change the value on the final wizard page to some random path
2. Make sure the default location was created anyway (the one that you saw on the last wizard page before you changed it)
3. Close client
4. Remove directory in default location
5. Start client again (this loads the accounts again)
6. Check if the default location was created again

fmoc:
Use case 2:

1. Add new account with a non-default sync root path (i.e., change the path on the last setup wizard page)
2. Add a second sync root (with some different path)
3. Remove first sync root
4. Close client
5. Remove directory in the first location you set in the account setup wizard
6. Start client again (this loads the accounts again)
7. Check if a directory in the first location was created again

HanaGemela:
Passed in owncloud-client_5.0.0.11978-daily20230830

PR #11128, opened 2023-08-17
Do not create default sync root when loading accounts

erikjv:
Creating the default sync root is only needed at initial setup. If a user chooses to point all folders for Spaces outside that default sync root, and deletes the default sync root folder, it should not be re-created.

Fixes: #10919

### DIFF
--- a/changelog/unreleased/10919
+++ b/changelog/unreleased/10919
@@ -1,0 +1,9 @@
+Bugfix: Do not create default sync root when loading accounts
+
+Creating the default sync root is only needed at initial setup. If a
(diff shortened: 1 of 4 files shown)


TheOneRing (approved):
Looks reasonable, thx.

TheOneRing (commented):
Please provide a changelog entry

TheOneRing:
I opened an issue for the crash in the gui test.
https://github.com/owncloud/client/issues/11131

TheOneRing (changes requested):
changelog entry please

TheOneRing (approved):

Issue #11135, opened 2023-08-21
Client crashes when database version check fails


erikjv:
The check is done after the application constructor is run. In this constructor, a couple of signals get fired, which in turn queues showing the GUI window (e.g. an account is logged out, and needs new authentication). When the version check fails, it brings up a (modal) dialog, and the event loop is spun, which also brings up the GUI window. Somewhere halfway this, the quit button is clicked, tearing down the application with a half-way set up window. This causes a crash somewhere deep in a child-widget-deletion process.

The fix is to move even more out of the application constructor, so the full check can be done before any other GUI code is run. This comes down to setting up the account manager and the folder manager in the main function.

This should fix the crash mentioned in #11114 .

TheOneRing:
I think you might be correct but as the crash is in the destructor of the ui, it might also help if you call qApp->exec() in https://github.com/owncloud/client/blob/65017f978ed45f3f2ff0e639c367bb63d6e7e333/src/gui/main.cpp#L354

so that the delayed close is executed:
https://github.com/owncloud/client/blob/65017f978ed45f3f2ff0e639c367bb63d6e7e333/src/gui/main.cpp#L210-L214

Salipa-Gurung:
### Desktop client version: `ownCloud 5.0.0.11907-daily20230825 d82ccf`
Tested with both oc10 and ocis server
- older client version cannot be used with upgraded database :heavy_check_mark: 
- client does not crash :heavy_check_mark: 

ubuntu-6.2.0-26-generic | windows-10.0.19045 |
---|---
![Screenshot from 2023-08-25 12-10-01](https://github.com/owncloud/client/assets/53458341/2d50a7ff-b41f-44fc-979b-23de0b50e3ce) | ![Screenshot from 2023-08-25 12-00-01](https://github.com/owncloud/client/assets/53458341/7ede85a1-216d-406a-9c63-2f9af1fce5ff)

PR #11137, opened 2023-08-22
Fix crash when database version check fails

erikjv:
The check is done after the application constructor is run. In this constructor, a couple of signals get fired, which in turn queues showing the GUI window (e.g. an account is logged out, and needs new authentication). When the version check fails, it brings up a (modal) dialog, and the event loop is spun, which also brings up the GUI window. Somewhere halfway this, the quit button is clicked, tearing down the application with a half-way set up window. This causes a crash somewhere deep in a child-widget-deletion process.

The fix is to move even more out of the application constructor, so the full check can be done before any other GUI code is run. This comes down to setting up the account manager and the folder manager in the main function.

Fixes: #11135

### DIFF
--- a/src/gui/main.cpp
+++ b/src/gui/main.cpp
@@ -319,7 +319,6 @@ int main(int argc, char **argv)
     const auto platform = Platform::create();
 
     // Create the (Q)Application instance:
-
     QApplication app(argc, argv);
     // TODO: Can't set this without breaking current config paths
     //    setOrganizationName(QLatin1String(APPLICATION_VENDOR));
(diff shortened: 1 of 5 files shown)


TheOneRing (approved):
Looks reasonable

TheOneRing (changes requested):
Looking at the gui tests it looks questionable

TheOneRing:
See the stacktracelog in https://drone.owncloud.com/owncloud/client/16381/5/11

src/gui/main.cpp
```diff
@@ -330,91 +329,110 @@ int main(int argc, char **argv)
...
+        ocApp->gui()->slotShowSettings();
     }
+    if (!options.fileToOpen.isEmpty()) {
+        QTimer::singleShot(0, ocApp, [ocApp, fileToOpen = options.fileToOpen] { ocApp->openVirtualFile(fileToOpen); });
+    }
+
+    platform->startServices();
+
+#ifdef WITH_AUTO_UPDATER
+    // if handleStartup returns true, main()
```
TheOneRing, line 428: Hmm handling the updater AFTER we loaded everything/displayed a wizard looks wrong

TheOneRing (approved):

PR #11138, opened 2023-08-22
Use getter setter for SyncItem::instruction to simplify debugging

TheOneRing:


test/testsyncconflict.cpp
```diff
@@ -679,6 +679,7 @@ private slots:
         fakeFolder.localModifier().mkdir(QStringLiteral("C/c2"));
         fakeFolder.localModifier().insert(QStringLiteral("C/c2/bar"));
 
+
```
TheOneRing, line 682:
```suggestion
```

src/libsync/progressdispatcher.h
```diff
@@ -26,6 +26,7 @@
 #include "syncfileitem.h"
 
 namespace OCC {
+
```
TheOneRing, line 29:
```suggestion
```

fmoc (changes requested):

fmoc (approved):

PR #11154, opened 2023-08-29
Work/untranslate string

erikjv:


fmoc (approved):

fmoc:
Could be even more "technical" in my opinion, but this is fine, too.

src/gui/spaces/spacesmodel.cpp
```diff
@@ -63,7 +63,7 @@ QVariant SpacesModel::headerData(int section, Qt::Orientation orientation, int r
             case Columns::Enabled:
                 return tr("Enabled");
             case Columns::SpaceId:
-                return tr("SpaceId");
+                return QStringLiteral("Space Identifier"); // Hidden column, no need for a translatable string.
```
TheOneRing, line 66: Hmm one day due to some decision we might make the string user facing.
I think making "Space Id" or "Space ID" a QStringLiteral is fine but with "Identifier" I'm not sure...

erikjv: [Wikipedia](https://en.wikipedia.org/w/index.php?title=Identifier) indicates that both are okay. Not sure what to do.

fmoc: Just keep it `SpaceId` so that once it gets user facing, some people will be irritated and make a change.

(Not sure whether you can't just `Q_ASSERT` this instead of returning some string...)

TheOneRing (approved):

PR #11155, opened 2023-08-29
Work/merge identical strings

erikjv:


TheOneRing (changes requested):
Please fix target branch/base branch.
Also the code does not compile.

src/gui/folderman.cpp
```diff
@@ -734,15 +734,11 @@ QString FolderMan::checkPathValidityRecursive(const QString &path)
         return FolderMan::tr("No valid folder selected!");
     }
 
-#ifdef Q_OS_WIN
-    Utility::NtfsPermissionLookupRAII ntfs_perm;
```
TheOneRing, line 738: Why was the raii removed? It has effects on QFileInfo

erikjv: fixed

TheOneRing (approved):

PR #11157, opened 2023-08-30
Remove basic support for libcloudproviders

TheOneRing:


changelog/unreleased/11157
```diff
@@ -0,0 +1,7 @@
+Change: Remove basic support of libcloudproviders
```
fmoc, line 1:
```suggestion
Change: Remove libcloudproviders integration
```

I don't think we need to estimate the degree of integration...

fmoc (changes requested):

changelog/unreleased/11157
```diff
@@ -0,0 +1,7 @@
+Change: Remove libcloudproviders integration
+
+Since its implementation in 2019 it support wasn't well maintained by us.
+Additionally as far as we can tell it was not picked up by any major Linux distribution.
```
fmoc, line 4:
```suggestion
Since its implementation in 2019 we did not maintain the implementation very much anyway.
Additionally, as far as we can tell, libcloudproviders was not picked up as a feature by any major Linux desktop environment (not even GNOME, who maintain libcloudproviders at the moment).
```

fmoc (changes requested):

fmoc (approved):

Issue #11141, opened 2023-08-23
Move auto-update check before new account wizard


erikjv:
The auto updater is checked in the `main()` function after the UI is initialised. If an update is available, the application will quit. So it would be better to do this before any UI is set up. But we need to check if the updater depends on other UI to be there.

See https://github.com/owncloud/client/pull/11137#discussion_r1303036960

erikjv:
This is a bit more complicated than described because of command-line options: if a user requested opening the settings dialog, or to open a (virtual) file, then upgrading (and quitting) will not be appreciated. So, the steps are:

1. Version checks
2. if opening the settings dialog is requested, do that, and no update check.
3. if not opening the dialog, but opening a file is requested, do that, and no update check
4. if non of the above is the case, do the update check

Then lastly, if no accounts are configured (and no update is found), open the new-account-wizard.

HanaGemela:
Steps to test:
1. Update using auto updater

HanaGemela:
Update using online auto updater works from 4.2. to ownCloud 5.0.0.12103-daily20230912

PR #11174, opened 2023-09-04
Change when checking for updates

erikjv:
We now handle user-requested actions from the command-line, and not check for updates, because a user would not appreciate it if the application quits because an update is found.

Also move opening the new-account wizard to after the update check, so that when no account is configured, and an update is available, only one dialog will be shown.

Fixes: #11141


TheOneRing (approved):

TheOneRing:
Hmm it appears  to break some ui tests, the wizard does not show up.

src/gui/main.cpp
```diff
@@ -417,24 +412,34 @@ int main(int argc, char **argv)
...
-
-#ifdef WITH_AUTO_UPDATER
-    // if handleStartup returns true, main()
-    // needs to terminate here, e.g. because
-    // the updater is triggered
-    Updater *updater = Updater::instance();
-    if (updater && updater->handleStartup()) {
-        return 1;
+    // Display the wizard if we don't have an account yet, and no other UI is showing.
+    if (!options.show && AccountManager::instance()->accounts().isEmpty()) {
```
TheOneRing, line 440: why: `!options.show` ?
This appears to break the gui tests.
options.show will only bring the application to the front, which will be done by the wizard anyhow.

TheOneRing (changes requested):

src/gui/main.cpp
erikjv: Ok, removed that part. Let's see how the tests do now.

TheOneRing:
Hmm all tests failed

TheOneRing:
I updated the branch lets see

PR #11194, opened 2023-09-07
Correctly select the first account when multiAcc is off

TheOneRing:
Steps to reproduce:
- single account branding
- once an account is added, the activity page is selected instead of the account.

With this change always the first account will be selected by default.

### DIFF
--- a/src/gui/settingsdialog.cpp
+++ b/src/gui/settingsdialog.cpp
@@ -343,9 +343,12 @@ void SettingsDialog::slotSwitchPage(QAction *action)
 
 void SettingsDialog::showFirstPage()
 {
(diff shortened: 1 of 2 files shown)


erikjv (approved):

HanaGemela:
Can't reproduce with QA theme on Ubuntu 22.04

HanaGemela:
Steps to reproduce:

1. single account branding
2. one account is added
3. quit the application
4. start the application 

Actual result: the activity page is selected instead of the account.

HanaGemela:
Fixed in qa 5.0.0.12128-daily20230914, Ubuntu 22.04

PR #11198, opened 2023-09-08
Add qpa info to version info

TheOneRing:
# Old

```
ownCloud 5.0.0-git [c8c2d9](https://github.com/owncloud/client/commit/c8c2d927da5bb3ce88ce9eee620e5b659f13667d)
Libraries Qt 6.4.3, OpenSSL 3.1.2 1 Aug 2023
Using virtual files plugin: wincfapi
windows-10.0.22621
```

# New

```
ownCloud 5.0.0-git [c8c2d9](https://github.com/owncloud/client/commit/c8c2d927da5bb3ce88ce9eee620e5b659f13667d)
Libraries Qt 6.4.3, OpenSSL 3.1.2 1 Aug 2023
Using virtual files plugin: wincfapi
OS: windows-10.0.22621
QPA: windows
```
On linux this will tell us whether wayland or xcb is used.

### DIFF
--- a/src/libsync/theme.cpp
+++ b/src/libsync/theme.cpp
@@ -374,6 +374,10 @@ QString Theme::aboutVersions(Theme::VersionFormat format) const
             gitUrl = gitSHA1(format) + br;
         }
     }
+    QStringList sysInfo = {QStringLiteral("OS: %1-%2").arg(QSysInfo::productType(), QSysInfo::kernelVersion())};
+    if (auto guiApp = qobject_cast<QGuiApplication *>(qApp)) {
+        sysInfo << QStringLiteral("QPA: %1").arg(guiApp->platformName());
+    }
 
     return QCoreApplication::translate("ownCloudTheme::aboutVersions()",
         "%1 %2%7"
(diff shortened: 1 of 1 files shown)


src/libsync/theme.cpp
```diff
@@ -374,6 +374,10 @@ QString Theme::aboutVersions(Theme::VersionFormat format) const
             gitUrl = gitSHA1(format) + br;
         }
     }
+    QStringList sysInfo = {QStringLiteral("OS: %1-%2").arg(QSysInfo::productType(), QSysInfo::kernelVersion())};
+    if (auto guiApp = qobject_cast<QGuiApplication *>(qApp)) {
```
fmoc, line 378: Maybe we should `OC_EXPECT` this cast. I can't think of a situation in which this would fail regularly. 

If this is called by the CLI, too, then we should add a comment that mentions this so people don't have to speculate.

fmoc (changes requested):

src/libsync/theme.cpp
fmoc:
```suggestion
    // may be called by both GUI and CLI, but we can display QPA only for the former
    if (auto guiApp = qobject_cast<QGuiApplication *>(qApp)) {
```

fmoc (approved):

Issue #11132, opened 2023-08-18
[QA] 5.0 client doesn't suggest restarting after upgrade on Ubuntu 


HanaGemela:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the QA issue

ownCloud 5.0.0.11804-daily20230818 client doesn't get restarted after upgrade on Ubuntu 22.04

### Steps to reproduce the issue

1. Install 4.2 client
2. Upgrade to 5.0 client

### Screenshots

No screenshot

### Expected behavior

New client is running

### Actual behavior

Client needs manual restart to show new 5.0 version

fmoc:
As far as I know, the client does not restart automatically, but it should suggest to do so when it notices an upgrade when the updater is run. My guess is that this detection is not sufficiently prepared for the path changes in the new packages. There is nothing we can do about this, as this affects the 4.x code. 

CC @TheOneRing

fmoc:
This might be a regression from another recent change and is likely not due to a change in packaging.

TheOneRing:
The problem is that we changed the client install location, the old client tries to call its old binary location with not result.

I vote for won't fix.

TheOneRing:
The only side effect is that we don't restart immediately, use the user restarts the client/computer everything is fine.

fmoc:
Should be fixed now that #11201 has been merged.

fmoc:
You can't test this with 4.2.0. This was implemented in a code change in 6.x so you need to update from 6.x to another 6.x. You can use our playground update server to trigger the update:

```sh
export OCC_UPDATE_URL=https://force-update.client-desktop-playground.owncloud.works

Salipa-Gurung:
I tried to upgrade client version ownCloud 6.0.0.13002-daily20240101 to  ownCloud-6.0.0.13016-daily20240103.

![Screenshot from 2024-01-03 17-06-46](https://github.com/owncloud/client/assets/53458341/0532f7bc-2937-4d1a-9910-ab2506f62e57)

![Screenshot from 2024-01-03 16-45-17](https://github.com/owncloud/client/assets/53458341/238d853b-5309-4220-b625-01d8de2d2d97)

[...]

PR #11201, opened 2023-09-12
Ask user in dialog to restart app after update finished

fmoc:
Related to #11132.


TheOneRing (changes requested):
We are in string freeze please target master

fmoc:
Good point, thanks. The change is planned for 6.x+ anyway.

src/gui/updater/updatefinisheddialog.ui
```diff
@@ -0,0 +1,71 @@
...
+      </size>
+     </property>
+     <property name="maximumSize">
+      <size>
+       <width>350</width>
+       <height>16777215</height>
+      </size>
+     </property>
+     <property name="text">
+      <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Update finished successfully.&lt;/p&gt;&lt;p&gt;Do you want to restart the client now?&lt;/p&gt;&lt;p&gt;Note: if you dismiss this dialog, you need to restart the client manually in order for the update to take effect.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
erikjv, line 50: I'm not sure about the wording. I'll check a couple of applications to see how the wording is done there.

fmoc: I also am not too happy with this wording. Suggestions welcome.

michaelstingl: @tbsbdr @EParzefall for help? ⬆️

tbsbdr: Proposal: I'd only talk about the happy path (=restart) and keep it short:


---


### Update installed successfully
Restart now to finish installing the update.
`Restart now` `Restart later`


---

EParzefall: My suggestion: Restart now for changes to take effect (otherwise a manual restart is necessary later).

erikjv: These kind of messages are always in passive voice, meaning, the "you" is always left out. (This is in contrast to e.g. a FAQ, where the user is directly addressed.) I'd go for the second suggestion, because it makes it clear that the user needs to take an action if the restart is postponed.

erikjv (changes requested):

src/gui/application.cpp
```diff
@@ -123,10 +123,9 @@ Application::Application(Platform *platform, const QString &displayLanguage, boo
 
 #ifdef WITH_AUTO_UPDATER
     // Update checks
-    UpdaterScheduler *updaterScheduler = new UpdaterScheduler(this);
```
TheOneRing, line 126: why the change in ownership?

src/gui/updater/ocupdater.cpp
```diff
@@ -45,7 +46,30 @@ UpdaterScheduler::UpdaterScheduler(QObject *parent)
...
             this, &UpdaterScheduler::updaterAnnouncement);
 
-        connect(updater, &OCUpdater::requestRestart, this, &UpdaterScheduler::requestRestart);
+        connect(updater, &OCUpdater::requestRestart, this, [this]() {
+            // prevent dialog from being displayed twice (rather unlikely, but it won't hurt)
+            if (_updateFinishedDialog == nullptr) {
+                _updateFinishedDialog = new UpdateFinishedDialog(_parentWidget);
+                _updateFinishedDialog->setAttribute(Qt::WA_DeleteOnClose);
+                _updateFinishedDialog->show();
+                ownCloudGui::raiseDialog(_updateFinishedDialog);
```
TheOneRing, line 55: Dialogs passed to raiseDialog need to be parented by settingsDialog so passing it as QWidget is shadowing this.

src/gui/updater/ocupdater.cpp
```diff
@@ -45,7 +46,30 @@ UpdaterScheduler::UpdaterScheduler(QObject *parent)
...
+
+                connect(_updateFinishedDialog, &UpdateFinishedDialog::accepted, this, []() {
+                    if (Utility::isLinux()) {
```
TheOneRing, line 58: oc_ensure?

src/gui/updater/updatefinisheddialog.ui
TheOneRing: ^

TheOneRing (changes requested):

src/gui/application.cpp
fmoc: I just rebased the PR...

src/gui/updater/updatefinisheddialog.ui
fmoc: screenshot_2023-11-08_14-57-23

TheOneRing: We need to clarify a that we are talking about an application restart, not a system one.
The second one is overly complex.

src/gui/updater/ocupdater.cpp
fmoc: Ok. But I wonder why `raiseDialog` doesn't enforce the use with `QDialog`s then (and `QMainWindow`). We should assert this or provide two signatures.

src/gui/updater/updatefinisheddialog.ui
erikjv: The first one is nice and short. Maybe phrase the later button as "Restart later manually"?

fmoc: The buttons' text is very long (and it won't get better when translated), this is why I prefer to use "yes/no". I don't see the complexity either. Let's try, though.

fmoc: screenshot_2023-11-08_15-29-54

src/gui/application.cpp
fmoc: Because we need access to settings dialog to correctly parent the update finished dialog. It shouldn't matter who the owner is.

src/gui/updater/updatefinisheddialog.ui
TheOneRing: how about `application restart required` also for the header?

fmoc: screenshot_2023-11-08_15-44-52

src/gui/application.cpp
TheOneRing: Its a QObject so the parenting decides on the lifetime, I'd suggest to add an additional parameter.

src/gui/updater/ocupdater.cpp
```diff
@@ -45,7 +46,27 @@ UpdaterScheduler::UpdaterScheduler(QObject *parent)
...
+                _updateFinishedDialog = new UpdateFinishedDialog(_parentWidget);
+                _updateFinishedDialog->setAttribute(Qt::WA_DeleteOnClose);
+                _updateFinishedDialog->show();
+                ownCloudGui::raiseDialog(_updateFinishedDialog);
+
+                connect(_updateFinishedDialog, &UpdateFinishedDialog::accepted, this, []() {
+                    if (OC_ENSURE(Utility::isLinux())) {
+                        // restart:
+                        qCInfo(lcUpdater) << "Restarting application NOW, PID" << qApp->applicationPid() << "is ending.";
+                        qApp->quit();
```
TheOneRing, line 61: QTimer::singleShot(0?

fmoc: For the record: this code existed before.

src/gui/updater/ocupdater.cpp
TheOneRing: https://github.com/owncloud/client/blob/769fa6f9fc34d8084e87fb0669ec413cec0af667/src/gui/owncloudgui.cpp#L985-L985

fmoc: I wonder, how was this assert rendered unusable while passing the variable as a `QWidget`? I mean, the code seemed to work and the assert did not trigger.

Issue #11203, opened 2023-09-13
Client crash in AbstractNetworkJob at `abstractnetworkjob.cpp:296` (during GUI tests run)


saw-jan:
### Describe the bug

```c
Core was generated by `/drone/src/build/bin/owncloud -s --logfile - --logdebug --logflush --confdir /t'.
Program terminated with signal SIGILL, Illegal instruction.
#0  OCC::AbstractNetworkJob::~AbstractNetworkJob (this=0x7f498006bb90) at /drone/src/src/libsync/abstractnetworkjob.cpp:296
296	{
[Current thread is 1 (Thread 0x7f49c9082f00 (LWP 4232))]
#0  OCC::AbstractNetworkJob::~AbstractNetworkJob() (this=0x7f498006bb90) at /drone/src/src/libsync/abstractnetworkjob.cpp:296
#1  0x00007f49ccde1003 in QObjectPrivate::deleteChildren() () at /lib64/libQt6Core.so.6
#2  0x00007f49ccde5428 in QObject::~QObject() () at /lib64/libQt6Core.so.6
```

### Logs
<details>
<summary>Full stacktrace</summary>

```c
[New LWP 4232]
[New LWP 4236]
[New LWP 4422]
[New LWP 4416]
[New LWP 4235]
[New LWP 4247]
[New LWP 4425]
[New LWP 4417]
[New LWP 4424]
[New LWP 4599]
[New LWP 4234]
[New LWP 4423]
[New LWP 4420]
[New LWP 4598]
[New LWP 4427]
[New LWP 4605]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
Core was generated by `/drone/src/build/bin/owncloud -s --logfile - --logdebug --logflush --confdir /t'.
Program terminated with signal SIGILL, Illegal instruction.
#0  OCC::AbstractNetworkJob::~AbstractNetworkJob (this=0x7f498006bb90) at /drone/src/src/libsync/abstractnetworkjob.cpp:296
296	{
[Current thread is 1 (Thread 0x7f49c9082f00 (LWP 4232))]
#0  OCC::AbstractNetworkJob::~AbstractNetworkJob() (this=0x7f498006bb90) at /drone/src/src/libsync/abstractnetworkjob.cpp:296
#1  0x00007f49ccde1003 in QObjectPrivate::deleteChildren() () at /lib64/libQt6Core.so.6
[...]

PR #11211, opened 2023-09-15
Fix crash in network job destructor

TheOneRing:
Fixes: #11203

### DIFF
--- a/src/libsync/graphapi/spacesmanager.cpp
+++ b/src/libsync/graphapi/spacesmanager.cpp
@@ -55,7 +55,8 @@ void SpacesManager::refresh()
     if (!_account->credentials()->ready()) {
         return;
     }
-    auto drivesJob = new Drives(_account->sharedFromThis(), this);
+    // leak the job until we fixed the onwership https://github.com/owncloud/client/issues/11203
+    auto drivesJob = new Drives(_account->sharedFromThis(), nullptr);
     drivesJob->setTimeout(refreshTimeoutC);
     connect(drivesJob, &Drives::finishedSignal, this, [drivesJob, this] {
         drivesJob->deleteLater();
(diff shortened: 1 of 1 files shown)


fmoc (approved):

src/libsync/graphapi/spacesmanager.cpp
```diff
@@ -55,7 +55,8 @@ void SpacesManager::refresh()
     if (!_account->credentials()->ready()) {
         return;
     }
-    auto drivesJob = new Drives(_account->sharedFromThis(), this);
+    // leak the job until we fixed the onwership https://github.com/owncloud/client/issues/11203
```
fmoc, line 58:
```suggestion
    // TODO: leak the job until we fixed the onwership https://github.com/owncloud/client/issues/11203
```

fmoc (changes requested):

fmoc (approved):

fmoc:
Looks like the Nautilus plugin is causing problems again:
```
Initializing ownCloud-client-nautilus extension
Using python version sys.version_info(major=3, minor=10, micro=12, releaselevel='final', serial=0)
Could not connect to unix socket /tmp/runtime-headless/ownCloud/socket. [Errno 111] Connection refused
13:39:17:482 Info: Connection to 127.0.0.1:54086 closed.
```

Let's merge this PR nevertheless and see if this is just a temporary issue. Otherwise, @saw-jan can try to fix this next week.

saw-jan:
> Could not connect to unix socket /tmp/runtime-headless/ownCloud/socket. [Errno 111] Connection refused

I believe this is expected as the error is thrown due to the initialization of `SocketConnection` before the desktop-client is up in https://github.com/owncloud/client-desktop-shell-integration-nautilus/blob/ebde52785c72b3688d157bb791ef0f37389aa431/src/syncstate.py#L176
Initialization happens whenever the file is loaded. so we get `Could not connect to unix socket` error. And as per Squish, I think, it resets and reloads everything per test suite (i.e. per `tst_<suite-name>` folder). So, I would expect to get atleast 11 such error

PR #11213, opened 2023-09-19
[full-ci] [tests-only] add test for syncing of a file with special characters

Salipa-Gurung:
### Description
This pr adds test for creation and deletion of file with special characters in it's name.

### Related Issue:
https://github.com/owncloud/QA/issues/789


saw-jan (approved):
:+1: :sparkles:

PR #11216, opened 2023-09-20
Create config file backup for any version change

erikjv:
We now do a backup of the config file for any version change of the client, not only when the major version changes.

We do still allow downgrades as long as the major version does not change.

Fixes: #10789

### DIFF
--- a/src/gui/main.cpp
+++ b/src/gui/main.cpp
@@ -229,11 +229,12 @@ bool checkClientVersion()
     auto configVersion = QVersionNumber::fromString(configFile.clientVersionWithBuildNumberString());
     auto clientVersion = OCC::Version::versionWithBuildNumber();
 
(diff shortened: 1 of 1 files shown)


src/gui/main.cpp
```diff
@@ -229,11 +229,12 @@ bool checkClientVersion()
     auto configVersion = QVersionNumber::fromString(configFile.clientVersionWithBuildNumberString());
     auto clientVersion = OCC::Version::versionWithBuildNumber();
 
-    if (configVersion.majorVersion() == clientVersion.majorVersion()) {
+    if (configVersion == clientVersion) {
         // no migration needed
```
TheOneRing, line 233:
```suggestion
        // no config backup needed
```

erikjv: done

TheOneRing (approved):

Issue #11191, opened 2023-09-07
[macOS] Sonoma beta WARNING: Secure coding is not enabled for restorable state!


michaelstingl:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the bug

When I open the 5.0.0-preview builds from macOS Sonoma Terminal it prints this warning:

```
% /Applications/owncloud.app/Contents/MacOS/owncloud          
2023-09-07 12:37:39.625 owncloud[3205:176246] WARNING: Secure coding is not enabled for restorable state! Enable secure coding by implementing NSApplicationDelegate.applicationSupportsSecureRestorableState: and returning YES.
```

(Everything seems to work normal)

### Expected behavior

I shouldn't see warnings like this

### Steps to reproduce the issue

1. open `Terminal.app`
2. execute `/Applications/owncloud.app/Contents/MacOS/owncloud`

### Client version number

```
% /Applications/owncloud.app/Contents/MacOS/owncloud --version
ownCloud ownCloud 5.0.0.12034-daily20230905
https://github.com/owncloud/client/commit/f53f7c8c984d0d4ea8c09bcce1189fc302fb2ccc
Libraries Qt 6.4.3, OpenSSL 3.1.2 1 Aug 2023
Using virtual files plugin: suffix
macos-23.0.0
```

michaelstingl:
I can confirm it's fixed in the `ownCloud-5.0.0.12187-rc.2-arm64.pkg` ✅

oladapo-joseph:
how can I correct this error, running into the same error, 
using python tkinter on M1 MacOS Sonoma

michaelstingl:
Probably tkinter needs to fix in their code. Here’s the example from ownCloud:
- https://github.com/owncloud/client/commit/4041ff7ebe17b25a242b1d1ff410743f97133bb1

PR #11219, opened 2023-09-21
Signal AppKit that we are fine with SecureCoding

erikjv:
Since macOS 12 there is a feature to do object de-serialization more secure (secure coding). Starting with with macOS 14, AppKit starts to emit a warning (on the command-line) when an application does not support this feature.

We, nor Qt, use object (de)serialization with `NSCoder`, so we can safely signal back to AppKit that we are fine with secure coding.

This surpresses the AppKit warning:

Secure coding is not enabled for restorable state! Enable secure coding by implementing NSApplicationDelegate.applicationSupportsSecureRestorableState: and returning YES.

Fixes: #11191


erikjv:
For testing: when starting the client from the command-line, the warning about supporting secure coding should not show up.

TheOneRing (approved):

Issue #11131, opened 2023-08-18
Oauth crash in gui test


TheOneRing:
```
#0  0x00007f5fee708844 in __pthread_kill_implementation () from /lib64/libc.so.6
[Current thread is 1 (Thread 0x7f5feb0dca40 (LWP 8686))]
#0  0x00007f5fee708844 in __pthread_kill_implementation () at /lib64/libc.so.6
#1  0x00007f5fee6b7abe in raise () at /lib64/libc.so.6
#2  0x00007f5fee6a090f in abort () at /lib64/libc.so.6
#3  0x00007f5feec840eb in qgetenv(char const*) [clone .cold] () at /lib64/libQt6Core.so.6
#4  0x00007f5feec84b7f in  () at /lib64/libQt6Core.so.6
#5  0x00007f5feec84083 in qt_assert_x(char const*, char const*, char const*, int) () at /lib64/libQt6Core.so.6
#6  0x00007f5ff0d11e8c in OCC::OAuth::authorisationLink() const (this=0x223f030) at /drone/src/src/libsync/creds/oauth.cpp:458
        code_challenge = {d = {d = 0x2225a60, ptr = 0x7ffd3cf44168 "Њ\r\002", size = 140725626094248}, static _empty = 0 '\000'}
        query = {d = <incomplete type>}
        url = {d = 0x20c5820}
#7  0x00000000005dc6a3 in OCC::HttpCredentialsGui::authorisationLink() const (this=0x13b92d0) at /drone/src/src/gui/creds/httpcredentialsgui.cpp:152
```
[stacktrace(2).log](https://github.com/owncloud/client/files/12378252/stacktrace.2.log)
https://drone.owncloud.com/owncloud/client/16368

fmoc:
Reproduction time boxed to 1h.

saw-jan:
Frequency:
https://drone.owncloud.com/owncloud/client/16551
https://drone.owncloud.com/owncloud/client/16591/5/11

TheOneRing:
The problem is both action buttons are enabled while there might not be a url yet...

saw-jan:
Can only be verified via gui tests in CI. Also, haven't seen it recently

PR #11220, opened 2023-09-21
Only enable oauth buttons once we got an auth url

TheOneRing:
Fixes: #11131

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -759,25 +759,15 @@ void AccountSettings::slotAccountStateChanged()
 
             auto *contentWidget = qobject_cast<OAuthLoginWidget *>(_askForOAuthLoginDialog->contentWidget());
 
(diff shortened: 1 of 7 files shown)


src/gui/accountsettings.cpp
```diff
@@ -759,25 +759,15 @@ void AccountSettings::slotAccountStateChanged()
...
-
-            contentWidget->setEnabled(false);
-            connect(cred, &HttpCredentialsGui::authorisationLinkChanged, contentWidget, [contentWidget]() {
-                contentWidget->setEnabled(true);
+            connect(contentWidget, &OAuthLoginWidget::copyUrlToClipboardButtonClicked, _askForOAuthLoginDialog, [](const QUrl &url) {
+                // TODO: use authorisationLinkAsync
+                qApp->clipboard()->setText(url.toString());
             });
 
-            connect(
```
TheOneRing, line 777: No idea what that connect was supposed to do....

erikjv (approved):

Issue #11230, opened 2023-09-27
[QA] Add space button gets renamed to Add folder when logged out 


HanaGemela:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the QA issue

ownCloud 5.0.0.12187-rc.2
mac

### Steps to reproduce the issue

1. Add an ocis account
2. Log in -> check the button at bottom left, it's called Add space
3. Log out -> it's called Add folder

### Screenshots

<img width="802" alt="Screenshot 2023-09-27 at 08 04 37" src="https://github.com/owncloud/client/assets/49001702/3de22d0b-90b1-49fc-b011-33883defe613">

### Expected behavior

The button doesn’t change it's name. It just gets greyed out

### Actual behavior

Button changes its name when account is logged out

erikjv:
"Blocked": wait for the 5.0 release, so this can be merged for 5.1

TheOneRing:
> @TheOneRing What should be changed? And for which branch?

See pr. I'd say 5.1.

fmoc:
Resolved in PR #11256.

HanaGemela:
Fixed in ownCloud 5.1.0.12421-daily20231018

PR #11256, opened 2023-10-02
Fix "Add Space" button text in disconnected state

erikjv:
When the account state was disconnected from an oCIS server, the button would show "Add Folder" instead of "Add Space". This also fixes the tooltip of the button.

Fixes: #11230

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -828,9 +828,10 @@ void AccountSettings::slotAccountStateChanged()
         _toggleSignInOutAction->setText(tr("Log out"));
     }
 
-    ui->addButton->setEnabled(state == AccountState::Connected);
     if (state == AccountState::Connected) {
         ui->_folderList->setItemsExpandable(true);
+        ui->addButton->setEnabled(true);
+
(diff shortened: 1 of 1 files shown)


src/gui/accountsettings.cpp
```diff
@@ -840,8 +841,15 @@ void AccountSettings::slotAccountStateChanged()
         }
     } else {
         ui->_folderList->setItemsExpandable(false);
-        ui->addButton->setText(tr("Add Folder"));
-        ui->addButton->setToolTip(tr("You need to be connected to add a folder."));
+        ui->addButton->setEnabled(false);
+
+        if (_accountState->supportsSpaces()) {
+            ui->addButton->setText(tr("Add Space"));
```
TheOneRing, line 847: Any reason we have to change the button text at all?

TheOneRing (changes requested):
^

src/gui/accountsettings.cpp
erikjv: Default is folder, esp. when there is no account yet. So if you add an ocis account, it needs to be changed.

fmoc (approved):

PR #11259, opened 2023-10-04
Work/fix warnings and remove unused includes

erikjv:


src/gui/folder.cpp
```diff
@@ -311,6 +313,8 @@ void Folder::prepareFolder(const QString &path)
         const auto error = GetLastError();
         qCWarning(lcFolder) << "SetFileAttributesW failed on" << longDesktopIniPath << Utility::formatWinError(error);
     }
+#else
+    Q_UNUSED(path)
```
TheOneRing, line 317: for the future [[maybe_unused]] can reduce the number of needed ifdefs

TheOneRing (approved):

Issue #11217, opened 2023-09-21
[QA][gui-tests] Squish tests can sometime fail to assert account connection


saw-jan:
### Describe the QA issue

Sometimes GUI test can fail to assert that the account is logged in. This might be due to not waiting for connection status to be changed while account is still connecting

Example ci build: https://drone.owncloud.com/owncloud/client/16618/3/15

### Screenshots

![Screenshot from 2023-09-21 09-33-50](https://github.com/owncloud/client/assets/52366632/f5645ffa-9e84-4417-96f5-502cc1893fcd)

### Expected behavior

wait until connected status is visible

### Actual behavior

immediately check the status without waiting for status change

saw-jan:
Fixed in PR https://github.com/owncloud/client/pull/11268

PR #11268, opened 2023-10-09
[tests-only] [full-ci] Fix test for assertion of account connection by waiting for element if not found 

Salipa-Gurung:
### Description
This pr modifies the step for asserting the connection of account to the server by waiting for 5000ms(default) if the element is not found.
`
Then user "Alice" should be connect to the client-UI
`
### Related issue
Fixes https://github.com/owncloud/client/issues/11217

### DIFF
--- a/test/gui/shared/scripts/pageObjects/AccountSetting.py
+++ b/test/gui/shared/scripts/pageObjects/AccountSetting.py
@@ -111,6 +111,21 @@ class AccountSetting:
             )
 
     @staticmethod
+    def waitUntilAccountIsConnected(displayname, server, timeout=5000):
+        result = squish.waitFor(
+            lambda: AccountSetting.isUserSignedIn(displayname, server),
+            timeout,
+        )
+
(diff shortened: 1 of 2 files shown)


test/gui/shared/scripts/pageObjects/AccountSetting.py
```diff
@@ -110,6 +110,21 @@ def waitUntilConnectionIsConfigured(timeout=5000):
...
+    @staticmethod
+    def waitUntilAccountIsConnected(displayname, server, timeout=5000):
+        result = squish.waitFor(
+            lambda: AccountSetting.isUserSignedIn(displayname, server),
+            timeout,
+        )
+
+        if not result:
+            raise Exception(
+                "Timeout waiting for connection to be configured for "
```
saw-jan, line 122:
```suggestion
                "Timeout waiting for the account to be connected for "
```

saw-jan (approved):

Issue #11245, opened 2023-09-28
The `--language` option does not work


erikjv:
The value is read from the command-line options, but it is never used.

The question is if we still need it. It was there before there was the possibility to choose a different language in the UI. So on top of that, do we need the command-line option to override both the UI-configured language and the system language?

See also issue #11142 .

erikjv:
The value should have been passed to the `Application` class when it is constructed, and stored in the `_userEnforcedLanguage` member field (which is currently never set).

fmoc:
I use it as part of my development cycle in order to start the client with different languages. I think this should be kept.

HanaGemela:
Removed in ownCloud 6.0.0.12402-daily20231016 ✅

fmoc:
We decided on removing it as there are other means of changing the language for testing.

PR #11270, opened 2023-10-09
Remove unused command line option `--language`

erikjv:
This option was used before the UI language could be chosen in the settings.

Fixes: #11245

Todo: changelog entry


fmoc:
Personally, I'd like to keep this parameter, see https://github.com/owncloud/client/issues/11245#issuecomment-1754881023.

PR #11288, opened 2023-10-12
Fix possible crash when starting shell integration

erikjv:
The shell integration was started while the folder manager was being instantiated. It could happen that callbacks came in when this was not yet completed, resulting in a crash when calling back into the (incomplete) folder manager.

Now shell integration is started after all other startup work has been completed.

Fixes: #11280


TheOneRing (approved):

TheOneRing (dismissed):
Please add a changelog entry

fmoc (approved):

fmoc (approved):

PR #11300, opened 2023-10-18
Work/load translations early

### LINKED ISSUE #11142: Setup translations before `Application` instantiation
The version checks have translatable text, and are shown before the translations are set up. Move this to somewhere early in the `main()` function.

### LINKED ISSUE #11246: Deprecate legacy `--confdir` command-line  argument
This command-line argument causes problems with the language setup, see issue #11142 .

erikjv:
Remove the `--confdir` option and load translations before any messages are shown.

Fixes #11246. Fixes #11142.

Waits on #11316.

### DIFF
--- a/src/gui/main.cpp
+++ b/src/gui/main.cpp
@@ -36,9 +36,11 @@
 
 #include <QApplication>
 #include <QCommandLineParser>
+#include <QLibraryInfo>
 #include <QMessageBox>
 #include <QProcess>
 #include <QTimer>
+#include <QTranslator>
 #ifdef Q_OS_WIN
 #include <qt_windows.h>
 #endif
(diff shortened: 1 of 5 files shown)


erikjv:
This still needs changelog entries

erikjv:
> This still needs changelog entries

Done.

src/gui/main.cpp
```diff
@@ -114,25 +116,26 @@ CommandLineOptions parseOptions(const QStringList &arguments)
         return option;
     };
 
-    auto showSettingsLegacyOption = QCommandLineOption{{QStringLiteral("showsettings")}, QStringLiteral("Hidden legacy option")};
+    auto showSettingsLegacyOption = QCommandLineOption{{QStringLiteral("showsettings")}, QApplication::translate("CommandLine", "Hidden legacy option")};
```
TheOneRing, line 119: As it is hidden I don't think we need to translate it

TheOneRing:
@saw-jan I just realised the gui tests are relying on `--confdir` could you elaborate why?

saw-jan:
> @saw-jan I just realised the gui tests are relying on `--confdir` could you elaborate why?

yeah, we are using that option to use different config dir other than default one so that we don't mess up configs while running tests locally.

Fabian suggested if we can use `XDG_CONFIG_HOME` env during tests. so, we are going to try that one

TheOneRing:
> > @saw-jan I just realised the gui tests are relying on `--confdir` could you elaborate why?
> 
> yeah, we are using that option to use different config dir other than default one so that we don't mess up configs while running tests locally.
> 
> Fabian suggested if we can use `XDG_CONFIG_HOME` env during tests. so, we are going to try that one

thx can you prepare a pr? (please don't commit to this one)

saw-jan:
>thx can you prepare a pr? (please don't commit to this one)
 yeah, on the way

saw-jan:
>yeah, on the way

PR is ready to be merged: https://github.com/owncloud/client/pull/11316

src/gui/main.cpp
erikjv: done

TheOneRing (approved):

PR #11315, opened 2023-10-26
[tests-only] [full-ci] removing skip tag for open log dialog test scenario 

Salipa-Gurung:
Removing `@skip` tag for open log dialog test scenario.


saw-jan (approved):
:+1:

PR #11316, opened 2023-10-26
[gui-tests][full-ci] use `XDG_CONFIG_HOME` with custom location for AUT during squish tests

saw-jan:
`--confdir` option is deprecated and about to be removed in https://github.com/owncloud/client/pull/11300.
So, squish tests makes use of `XDG_CONFIG_HOME` env to set custom config location (i.e. `/tmp/owncloudtest/`) for owncloud during the test execution

NOTE: env `XDG_CONFIG_HOME` is only available for the AUT context so not available for the squish and test script

~NOTE: we have to come up with something for running tests locally with Squish.~
~**WARN: NOT RECOMMENDED to run gui-tests on local machine (doing so will remove existing client configs)**~

### DIFF
--- a/test/gui/shared/scripts/helpers/ConfigHelper.py
+++ b/test/gui/shared/scripts/helpers/ConfigHelper.py
@@ -8,11 +8,11 @@ def read_env_file():
     envs = {}
     script_path = os.path.dirname(os.path.realpath(__file__))
     env_path = os.path.abspath(os.path.join(script_path, '..', '..', '..', 'envs.txt'))
-    with open(env_path) as f:
+    with open(env_path, "rt", encoding="UTF-8") as f:
         for line in f:
             if line.startswith('#'):
                 continue
-            key, value = line.split('=')
+            key, value = line.split('=', 1)
             envs[key] = value.strip()
     return envs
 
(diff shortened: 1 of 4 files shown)


TheOneRing (dismissed):

test/gui/shared/scripts/helpers/ConfigHelper.py
```diff
@@ -3,6 +3,27 @@
 from tempfile import gettempdir
 from configparser import ConfigParser
 
+
+def read_env_file():
+    envs = {}
+    script_path = os.path.dirname(os.path.realpath(__file__))
+    env_path = os.path.abspath(os.path.join(script_path, '..', '..', '..', 'envs.txt'))
+    with open(env_path) as f:
```
TheOneRing, line 11:
```suggestion
    with open(env_path, "rt", encoding = "UTF-8") as f:
```
?

test/gui/shared/scripts/helpers/ConfigHelper.py
```diff
@@ -3,6 +3,27 @@
...
+        for line in f:
+            if line.startswith('#'):
+                continue
+            key, value = line.split('=')
```
TheOneRing, line 15:
```suggestion
            key, value = line.split('=', 1)
```

TheOneRing:
Have you thought about a different file format that is less picky with escaping etc

saw-jan:
> Have you thought about a different file format that is less picky with escaping etc

AFAIK, `ENVVARS=` in `suite.conf` file requires a plain text file with KEY=VALUE pair per line. I don't know if json or yml file works

fmoc (approved):

PR #11320, opened 2023-10-27
Move files to trash instead of deleting them

erikjv:


src/gui/folder.cpp
```diff
@@ -1007,6 +1007,12 @@ void Folder::setDirtyNetworkLimits()
     _engine->setNetworkLimits(uploadLimit, downloadLimit);
 }
 
+void Folder::reloadSyncOptions()
+{
+    // Q: abort the sync and restart afterwards?
```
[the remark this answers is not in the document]
erikjv: We decided that we will not abort the sync.

src/gui/generalsettings.cpp
```diff
@@ -80,6 +80,11 @@ GeneralSettings::GeneralSettings(QWidget *parent)
 
     _ui->crashreporterCheckBox->setVisible(Theme::instance()->withCrashReporter());
 
+    connect(_ui->moveToTrashCheckBox, &QCheckBox::toggled, this, [](bool checked) {
+        ConfigFile().setMoveToTrash(checked);
+        FolderMan::instance()->reloadSyncOptions();
+    });
```
TheOneRing, line 86: Hmm no strong opinion but wouldn't a "options changed" signal be Qt'er?

erikjv: This is about the same as done in `NetworkSettings::saveBWLimitSettings()` where a call is made to `FolderMan::instance()->setDirtyNetworkLimits();`

TheOneRing: setDirtyNetworkLimits is 10 years old :D 

I'd prefer a `Q_EMIT syncOptionsChanged()` signal.
and a 
```
connect(foo, &GeneralSettings::syncOptionsChanged, _engine, []{ setSyncOptions(loadSyncOptions());});
````
or something like that?

TheOneRing (dismissed):

TheOneRing (approved):

PR #11334, opened 2023-11-03
Fix file deletion of vfs switch

TheOneRing:


changelog/unreleased/11331
```diff
@@ -0,0 +1,7 @@
+Bugfix: Don't delete virtual files when disabling VFS
+
+We fixed a regression that was introduced in 5.0.
+When a user disabled virtual files on Windows all files currently dehydrated where deleted.
+On the server those files where still available in the trash bin.
+
+https://github.com/owncloud/client/issues/11331
```
fmoc, line 7:
```suggestion
Bugfix: Prevent permanent deletion of files when disabling VFS

When a user disabled the virtual files feature on Windows, all currently dehydrated files were deleted accidentally.
On the server those files where still available in the trash bin.
This regression was introduced in the previous release 5.0.0.

https://github.com/owncloud/client/issues/11331
```

fmoc (changes requested):

src/common/vfs.cpp
```diff
@@ -127,18 +145,37 @@ Vfs::AvailabilityResult Vfs::availabilityInDb(const QString &folderPath)
...
+            if (isDehydratedPlaceholder(absolutePath)) {
+                qCDebug(lcVfs) << "Removing local dehydrated placeholder" << relativePath;
+                FileSystem::remove(absolutePath);
+            }
+        }
+    });
+
+    _setupParams->journal->forceRemoteDiscoveryNextSync();
+
+    // Postcondition: No ItemTypeVirtualFile / ItemTypeVirtualFileDownload left in the db.
```
fmoc, line 177: Do you plan to add some kind of assert?

src/common/vfs.h
```diff
@@ -225,6 +225,14 @@ class OCSYNC_EXPORT Vfs : public QObject
      */
     [[nodiscard]] virtual AvailabilityResult availability(const QString &folderPath) = 0;
 
+    /** Removes all virtual file db entries and dehydrated local placeholders.
+     *
+     * Particularly useful when switching off vfs mode or switching to a
+     * different kind of vfs.
+     *
```
fmoc, line 232: Formatting.

```suggestion
```

src/gui/folder.cpp
```diff
@@ -769,6 +769,10 @@ void Folder::setVirtualFilesEnabled(bool enabled)
     }
 
     if (newMode != _definition.virtualFilesMode) {
+        // TODO: Must wait for current sync to finish!
```
fmoc, line 772: Is this still a TODO? I think we can safely say now that it's up to the caller to prevent this.

fmoc (changes requested):

src/gui/folder.cpp
TheOneRing: Still a TODO, removing // TODO would not solve anything.
For the next major release I'll provide a solution.

src/common/vfs.cpp
TheOneRing: Its not my code

fmoc (approved):

PR #11345, opened 2023-11-08
Abort current sync if switching vfs mode

TheOneRing:


fmoc (approved):

TheOneRing:
For testing:
- Limit bandwidth to ensure the sync will take some time
- Add a file
- While the sync is ongoing, disable/enable vfs

The client will terminate the current sync, switch vfs and start a new sync.

HanaGemela:
I don't think it works in ownCloud-6.0.0.12668-daily20231122

The seconds sync starts where the first (before the switch) stopped. This also means when VFS is turned on, the sync starts to download all the files and then when during this VFS is turned off, it keeps downloading everything. It is a new sync but starts where the previous one stopped. 
When I now pause this sync and then resume, it doesn't download anything as VFS is on.

Was it meant to work like this?

HanaGemela:
Fixed in 6.0.0 alpha 1

Issue #11204, opened 2023-09-13
Prevent: Same default sync folder for multiple accounts


TheOneRing:
Steps to reproduce:
- Add an ocis account
- Add a second ocis account, use advanced settings and use the default sync root form the first account

Both accounts use the same syncroot and differentiate the directory names _inside_ that syncroot:
- `root/Personal`
- `root/Personal (1)`

Expectations:
The client prevents the user from using the same syncroot directory, instead offering a different syncroot:
- `root/Personal`
- `root (1)/Personal`

TheOneRing:
We could use xattr and ntfs streams to "claim" a folder, if neither is supported for the folder we ignore it.
That way we should always be able to prevent multiple accounts from using the same default sync root.

Salipa-Gurung:
Tested on : 
ownCloud 6.0.0.13062-daily20240108 b46134
Libraries Qt 6.6.0, OpenSSL 3.1.4 24 Oct 2023
Using virtual files plugin: wincfapi
OS: windows-10.0.19045 (build arch: x86_64, CPU arch: x86_64)
QPA: windows

Server:
Infinite Scale 4.0.5 Community and ownCloud 10.13.4 (stable)

Adding account for brian after admin's account was added | Using default syncroot from the first account
-----|-----
| ![Screenshot from 2024-01-08 12-05-05](https://github.com/owncloud/client/assets/53458341/8f199a9e-bb99-4a3e-8d1b-75846bbd3217) | ![Screenshot from 2024-01-08 12-07-35](https://github.com/owncloud/client/assets/53458341/7d5c2cea-e3e0-46cd-994c-112119442dad) |
Client provides different syncroot :heavy_check_mark: | Client prevents user from using same syncroot directory :heavy_check_mark: 

`Finish` button remains clickable when error message is being displayed. :grey_question: 

cc: @erikjv

PR #11364, opened 2023-11-13
Spaces: Prevent same default sync folder for multiple accounts

erikjv:
Fixes: https://github.com/owncloud/client/issues/11204


src/libsync/filesystem.cpp
```diff
@@ -198,5 +203,96 @@ bool FileSystem::getInode(const QString &filename, quint64 *inode)
...
+    {
+        return ::removexattr(path, name);
+    }
+#endif // Q_OS_LINUX
+
+} // anonymous namespace
+
+static constexpr unsigned MaxValueSize = 1023; // This is without a terminating NUL character
+
+std::optional<QByteArray> FileSystem::Tags::get(const QString &path, const QString &key)
```
TheOneRing, line 229: Get and set should return the same type (string/bytearray)

erikjv:
To do: do the check for the spaces tag while checking folder paths, so recursively.

src/gui/guiutility.cpp
```diff
@@ -85,3 +86,29 @@ QString Utility::vfsFreeSpaceActionText()
 {
     return QCoreApplication::translate("utility", "Free up local space");
 }
+
+static const QString dirTag = QStringLiteral("com.owncloud.spaces.app");
```
TheOneRing, line 90: please move to anonymous namespace and make it a function

src/gui/guiutility.h
```diff
@@ -51,6 +51,10 @@ namespace Utility {
 
     QString socketApiSocketPath();
 
+    void setDirectoryTag(const QString &path);
```
TheOneRing, line 54: Please add a description.
Could you think of a more descriptive function name?

src/gui/guiutility.cpp
```diff
@@ -85,3 +93,32 @@ QString Utility::vfsFreeSpaceActionText()
...
     return QCoreApplication::translate("utility", "Free up local space");
 }
+
+void Utility::setDirectoryTag(const QString &path)
+{
+    auto result = FileSystem::Tags::set(path, dirTag(), Theme::instance()->orgDomainName().toUtf8());
+    if (!result) {
+        qCWarning(lcGuiUtility) << QStringLiteral("Failed to set tag on '%1': %2").arg(path, result.error())
+#ifdef Q_OS_WIN
+                                << QStringLiteral("(filesystem %1)").arg(FileSystem::fileSystemForPath(path))
```
TheOneRing, line 103: Wouldn't that make more sense to check in FileSystem::Tags::set(? sry for not clarifying earlier

src/libsync/filesystem.cpp
```diff
@@ -198,5 +207,100 @@ bool FileSystem::getInode(const QString &filename, quint64 *inode)
...
+        return file.readAll();
+    }
+#endif // Q_OS_MAC || Q_OS_LINUX
+
+    return {};
+}
+
+OCC::Result<void, QString> FileSystem::Tags::set(const QString &path, const QString &key, const QByteArray &value)
+{
+    OC_ASSERT(value.size() < MaxValueSize)
```
TheOneRing, line 257: Should we error on implicit overrides? Or would that only make sense for setDirectoryTag?

TheOneRing (changes requested):

src/gui/guiutility.h
erikjv: what about `markDirectoryAsSyncRoot(..)`?

src/gui/guiutility.cpp
erikjv: I kinda like doing the handling of the error outside of the "low level" function.

src/gui/guiutility.h
TheOneRing: Better :)

TheOneRing:
Please also add a changelog entry

src/gui/guiutility.h
erikjv: done

src/libsync/filesystem.cpp
erikjv: Neither? This is when one sync app is removed and another one is installed. Then the new application can take over the old sync root. Or is this a bad idea?

TheOneRing: The assert would only occur in dev setups.
For the above assert a Q_ASSERT should be sufficient.

I'd say a normal scenario would be that a sync root is properly removed, so we also removed the tag?
Maybe not an assert, but a general warning as the folder was not properly cleaned?

TheOneRing:
Changelog is still missing

src/libsync/filesystem.cpp
erikjv: Done

TheOneRing (approved):

TheOneRing:
Please fix the changelog:
> file changelog/unreleased/11204: entry type "Spaces" is invalid, valid types: map[Bugfix:2 Change:3 Enhancement:4 Security:1]

TheOneRing (approved):

Salipa-Gurung:
Tested https://github.com/owncloud/client/issues/11204#issuecomment-1880467290

PR #11365, opened 2023-11-14
Deprecate suffix VFS plugin

TheOneRing:


src/gui/newwizard/pages/accountconfiguredwizardpage.cpp
```diff
@@ -26,28 +26,18 @@ AccountConfiguredWizardPage::AccountConfiguredWizardPage(
     _ui->localDirectoryLineEdit->setText(QDir::toNativeSeparators(userChosenSyncTargetDir));
     _ui->syncEverythingRadioButton->setChecked(true);
 
-    // could also make it invisible, but then the UX is different for different installations
-    // this may be overwritten by a branding option (see below)
-    _ui->useVfsRadioButton->setEnabled(vfsIsAvailable);
+    _ui->useVfsRadioButton->setVisible(vfsIsAvailable);
```
fmoc, line 29: There is no apparent reason for this UX change. Please explain. We discussed this in depth when we introduced the new wizard and decided against hiding UX elements randomly on different platforms.

src/gui/newwizard/pages/accountconfiguredwizardpage.cpp
```diff
@@ -26,28 +26,18 @@ AccountConfiguredWizardPage::AccountConfiguredWizardPage(
...
-        }
-    }
 
     // just adjusting the visibility should be sufficient for these branding options
     if (Theme::instance()->wizardSkipAdvancedPage()) {
         _ui->advancedConfigGroupBox->setVisible(false);
     }
 
-    if (!Theme::instance()->showVirtualFilesOption()) {
+    if (!Theme::instance()->forceVirtualFilesOption()) {
```
fmoc, line 38: Again, I don't see a reason for this change.

fmoc (changes requested):

src/gui/newwizard/pages/accountconfiguredwizardpage.cpp
TheOneRing: Here we have a logic issue, I intended to hide if if vfs is enforced.
We really don't need to display it in that case.
(showVirtualFilesOption is removed)

TheOneRing: And I discovered that the whole section is hidden if vfs is forced

src/gui/newwizard/pages/accountconfiguredwizardpage.cpp
TheOneRing: Now that we don't have experimental features we basically have https://github.com/owncloud/client/pull/11365/files#diff-d101d06821ce246c78094f95aa54c84df2527a1187475bebf32b7f6cbef29f55L38

src/gui/folderstatusmodel.cpp
```diff
@@ -273,11 +273,15 @@ QVariant FolderStatusModel::data(const QModelIndex &index, int role) const
                 // the error message might contain new lines, the delegate only expect multiple single line values
                 errors.append(legacyError.error().split(QLatin1Char('\n')));
             }
+            if (f->isReady() && f->virtualFilesEnabled() && f->vfs().mode() == Vfs::Mode::WithSuffix) {
+                errors.append({
+                    tr("The suffix VFS plugin is deprecated and will be removed in the 7.0 release."),
+                    tr("Please use the context menu and select \"Disable virtual file support\" to  ensure future access to your synced files."),
+                    tr("You are going to lose access to your sync folder if you do not do so!"),
```
TheOneRing, line 280: @fmoc I had to remove the richtext as we don't render it at the moment.

changelog/unreleased/11325
```diff
@@ -0,0 +1,5 @@
+Change: Deprecate suffix VFS
+
+We decided to deprecate the suffix VFS implementations and to prepare it for removal in 7.0.
```
fmoc, line 3:
```suggestion
We decided to deprecate the suffix VFS implementations and thus prepare its removal in 7.0.
```

src/gui/application.cpp
```diff
@@ -95,15 +95,6 @@ Application::Application(Platform *platform, const QString &displayLanguage, boo
         AbstractNetworkJob::httpTimeout = cfg.timeout();
     }
 
-    // Check vfs plugins
-    if (Theme::instance()->showVirtualFilesOption() && VfsPluginManager::instance().bestAvailableVfsMode() == Vfs::Off) {
```
fmoc, line 99: I think we should log what kind of plugin is loaded. Does that happen elsewhere?

src/gui/folderstatusmodel.cpp
```diff
@@ -273,11 +273,15 @@ QVariant FolderStatusModel::data(const QModelIndex &index, int role) const
                 // the error message might contain new lines, the delegate only expect multiple single line values
                 errors.append(legacyError.error().split(QLatin1Char('\n')));
             }
+            if (f->isReady() && f->virtualFilesEnabled() && f->vfs().mode() == Vfs::Mode::WithSuffix) {
+                errors.append({
```
fmoc, line 277: Not that I have a better solution, but we may want to make this more visible to users. Maybe we should suggest them to convert to "classic" syncing on startup (i.e., when the accounts and sync folders are loaded, ideally with a single dialog that includes a list of all these directories and a button to directly apply the change.)

src/gui/folderwizard/folderwizardselectivesync.cpp
```diff
@@ -43,16 +43,14 @@ FolderWizardSelectiveSync::FolderWizardSelectiveSync(FolderWizardPrivate *parent
     _selectiveSync = new SelectiveSyncWidget(folderWizardPrivate()->accountState()->account(), this);
     layout->addWidget(_selectiveSync);
 
-    const auto vfsMode = VfsPluginManager::instance().bestAvailableVfsMode();
-    if (Theme::instance()->forceVirtualFilesOption() && vfsMode == Vfs::WindowsCfApi) {
-        // using an else if to make the condition not even more complex...
-    } else if (Theme::instance()->showVirtualFilesOption() && vfsMode != Vfs::Off && (vfsMode == Vfs::WindowsCfApi || Theme::instance()->enableExperimentalFeatures())) {
-        _virtualFilesCheckBox = new QCheckBox(tr("Use virtual files instead of downloading content immediately%1").arg(vfsMode == Vfs::WindowsCfApi ? QString() : tr(" (experimental)")));
+    if (!Theme::instance()->forceVirtualFilesOption() && Theme::instance()->showVirtualFilesOption()
+        && VfsPluginManager::instance().bestAvailableVfsMode() == Vfs::WindowsCfApi) {
```
fmoc, line 47: In the long term, we probably shouldn't hardcode the Windows implementation here, but just have `bestAvailableVfsMode` return `Vfs::Off` even if we ship such plugins.

I guess this needs some more refactoring in 7.x.

src/gui/newwizard/pages/accountconfiguredwizardpage.cpp
fmoc: Yeah, but before, this was only hidden as it was an experimental feature. Now it is just an un_available_ feature. Thus, the button should just be disabled, not hidden.

src/libsync/theme.cpp
```diff
@@ -577,19 +577,14 @@ QString Theme::versionSwitchOutput() const
 
 bool Theme::showVirtualFilesOption() const
 {
-    return enableExperimentalFeatures();
+    return true;
```
fmoc, line 580: Supported for historical reasons and prevent refactoring? Or are we going to provide this option in the future?

fmoc (changes requested):

src/gui/application.cpp
TheOneRing: line 79

src/gui/folderstatusmodel.cpp
TheOneRing: I guess a user won't want to switch all their 40+ sync folders all at once?

src/gui/newwizard/pages/accountconfiguredwizardpage.cpp
TheOneRing: The button was only visible vanilla builds as experimental features where disabled everywhere else.
So we have the exact same behaviour as before.

src/libsync/theme.cpp
TheOneRing: Supported in branding

src/gui/folderstatusmodel.cpp
fmoc: Also, who wants to switch 40+ folders manually?

fmoc (approved):
Please see my suggestion. We might want to create an issue for it.

src/gui/folderstatusmodel.cpp
TheOneRing: I think a huge red box per sync is more than sufficient.

dsonck92:
I've just noticed this change with the upgrade to 6.0.0, is there any equivalent planned for the future? I've been using the suffix based VFS driver for a long time since it works more elegantly than the conditional syncing of folders.

I have a lot of files that I only need partly available but aren't put into folders. A similar setup like Windows would be very welcome, but the suffix based variant worked decently enough so it's a bit of a surprise it got removed. I see that it was already  done in 2023, but only now in v6 is it prominently mentioned. I think it would have been nice to know where I could have seen this coming in 2023.

Issue #10757, opened 2023-04-25
Automatically upgrade rosetta2 users from `x86_64` to `arm64` builds


fmoc:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Feature description

We should automatically upgrade rosetta2 users on M1/M2 to the native `arm64` builds we now provide.

### Proposed solution

See https://developer.apple.com/forums/thread/653009. Shouldn't be a problem to add that to our updater.

### Alternative solutions you considered

Tell users to switch to `arm64` manually if they want to benefit from the native build's advantages (like generally higher efficiency, less battery consumption, ...).

fmoc:
First half implemented in https://github.com/owncloud/client/pull/11376, please review. The second half will follow in a PR to the updater server.

fmoc:
The update server already behaves like we need it to. It just uses `currentArch` to determine which binary to ship (if available, otherwise falls back to `buildArch`). As we send both parameters anyway, it's working as expected in combination with #11376.

fmoc:
Test instructions: test upgrade against playground server.

Note: blocked by another issue.

fmoc:
See also https://github.com/owncloud/client/issues/11388.

HanaGemela:
Automated updates don't work at the moment

PR #11376, opened 2023-11-16
Fix current CPU architecture detection with rosetta2

fmoc:
This PR uses a syscall to properly detect whether the application is running inside of macOS's rosetta2 emulator. It replaces uses of `QSysInfo::currentCpuArchitecture()` with a custom in our code.

It further extends the about dialog to show the detected build and CPU architectures.

This is a preparation for #10757.

Please note that the formatting changes are forced by clang-format. I'm not a fan of them.

### DIFF
--- a/src/common/utility_mac.mm
+++ b/src/common/utility_mac.mm
@@ -32,6 +32,8 @@
 #import <Foundation/NSFileManager.h>
 #import <Foundation/NSUserDefaults.h>
 
+#include <sys/sysctl.h>
+
 namespace OCC {
 
 void Utility::setupFavLink(const QString &folder)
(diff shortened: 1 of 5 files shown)


src/common/utility_mac.mm
```diff
@@ -255,4 +257,19 @@
 }
 #endif
 
+QString Utility::currentCpuArch()
+{
+    int ret = 0;
+    size_t size = sizeof(ret);
+
+    // ret will be 1 if the process is translated (most likely with rosetta2)
+    if (sysctlbyname("sysctl.proc_translated", &ret, &size, nullptr, 0) != -1) {
```
TheOneRing, line 266: No idea on how expansive this call is but as we add it to the user agent, could it make sense to cache the result locally in a static?
I don't expect this value to change at runtime :D

TheOneRing:
If you really want to you can enforce new lines with //

src/common/utility_mac.mm
fmoc: Well, syscalls always cause a context switch, so yes, we should cache this.

TheOneRing (approved):

PR #11386, opened 2023-11-21
Update release_template.md

HanaGemela:


.github/release_template.md
```diff
@@ -72,6 +72,7 @@ __TODO__:
 ### Final Steps
 
 * [ ] QA: [Smoke test](https://confluence.owncloud.com/display/OG/Manual+Tests#ManualTests-SmokeTest)
+* [ ] QA: Check the linux download pages whether the URLs are correct (e.g. contain stable, not testing)
```
fmoc, line 75: We need to do this after "publish the release on GitHub", which includes "moving" the release from testing/ to stable/. How about moving this to "Infrastructure check"?

fmoc:
Background: https://gitea.owncloud.services/client/release/commit/93e9380afe5c0b85a58bb5803ca653f1adcec47b was needed for the 5.0.0 release which lacked Linux packages. We (I) forgot about it in the meantime, causing this problem. I reverted it now to fix the problem in the long run and manually updated the download pages on the mirrors.

fmoc (changes requested):
See review above.

fmoc (approved):

PR #11396, opened 2023-11-23
Emit finished signal if a sync was aborted

TheOneRing:


src/libsync/syncengine.cpp
```diff
@@ -857,18 +857,19 @@ bool SyncEngine::shouldDiscoverLocally(const QString &path) const
...
     if (_propagator) {
+        aborting = true;
         // If we're already in the propagation phase, aborting that is sufficient
         _propagator->abort();
     } else if (_discoveryPhase) {
+        aborting = true;
         // Delete the discovery and all child jobs after ensuring
         // it can't finish and start the propagator
         disconnect(_discoveryPhase.get(), nullptr, this, nullptr);
-        _discoveryPhase.release()->deleteLater();
```
TheOneRing, line 870: the delete happens in finish

test/testchunkingng.cpp
```diff
@@ -344,18 +344,25 @@ private slots:
...
+                // while the response is delayed, the move is performed in the custructor, so immideatly
                 return new DelayedReply<FakeChunkMoveReply>(responseDelay, fakeFolder.uploadState(), fakeFolder.remoteModifier(), op, request, &parent);
             }
             return nullptr;
         });
 
         // Test 1: NEW file aborted
         fakeFolder.localModifier().insert(QStringLiteral("A/a0"), size);
+        QVERIFY(!fakeFolder.applyLocalModificationsAndSync());
+        QCOMPARE(fakeFolder.currentLocalState(), fakeFolder.currentRemoteState());
```
fmoc, line 356: Why do we expect local and remote state to be the same in this scenario? Shouldn't they differ now and be synced in the upcoming run?

fmoc (changes requested):

test/testchunkingng.cpp
TheOneRing: I added more documentation

test/testchunkingng.cpp
```diff
@@ -344,18 +346,31 @@ private slots:
         fakeFolder.setServerOverride([&](QNetworkAccessManager::Operation op, const QNetworkRequest &request, QIODevice *) -> QNetworkReply * {
             if (request.attribute(QNetworkRequest::CustomVerbAttribute).toByteArray() == "MOVE") {
                 QTimer::singleShot(50ms, &parent, [&]() { fakeFolder.syncEngine().abort(); });
+                // while the response is delayed, the move is performed in the custructor, so immideatly
```
fmoc, line 349:
```suggestion
                // while the response is delayed, the move is performed in the constructor, thus it happens immediately
```

test/testchunkingng.cpp
```diff
@@ -330,8 +330,10 @@ private slots:
         QCOMPARE(fakeFolder.currentLocalState(), fakeFolder.currentRemoteState());
     }
 
-    // Check what happens when we abort during the final MOVE and the
-    // the final MOVE is short enough for the abort-delay to help
+    // Check what happens when we abort during the final MOVE.
+    // The move succeeds on the server but as we abort it we won't relaize in time.
+    // This means that the current sync fails, as it is aborted, but the MOVE was still performed on the server.
+    // Resulting in the local state beeing the same as the remote state.
```
fmoc, line 336:
```suggestion
    // Check what happens when we abort during the final MOVE.
    // The move succeeds on the server but as we abort it we won't realize in time.
    // This means that the current sync fails, as it is aborted, but the MOVE was still performed on the server.
    // Resulting in the local state being the same as the remote state.
```

fmoc (changes requested):

fmoc (approved):

erikjv (approved):

PR #11400, opened 2023-11-28
Enable testing of switching off VFS for all VFSses

erikjv:
The test `TestSyncVirtualFiles::testWipeVirtualSuffixFiles` was only run with suffix VFS, but is also applicable for other VFSses. This change renames it to `testSwitchOffVFS` and runs it on all available VFS plugins.

Fixes: https://github.com/owncloud/client/issues/11342

### DIFF
--- a/test/testsyncvirtualfiles.cpp
+++ b/test/testsyncvirtualfiles.cpp
@@ -52,9 +52,9 @@ void markForDehydration(FakeFolder &folder, const QString &path)
     journal.schedulePathForRemoteDiscovery(record._path);
 }
 
-QSharedPointer<Vfs> setupVfs(FakeFolder &folder)
+QSharedPointer<Vfs> setupVfs(FakeFolder &folder, Vfs::Mode vfsMode = Vfs::WithSuffix)
 {
-    auto suffixVfs = QSharedPointer<Vfs>(VfsPluginManager::instance().createVfsFromPlugin(Vfs::WithSuffix).release());
+    auto suffixVfs = QSharedPointer<Vfs>(VfsPluginManager::instance().createVfsFromPlugin(vfsMode).release());
     folder.switchToVfs(suffixVfs);
 
     // Using this directly doesn't recursively unpin everything and instead leaves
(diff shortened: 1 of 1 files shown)


erikjv:
Note: this does not pass the tests when winvfs is available!

test/testsyncvirtualfiles.cpp
```diff
@@ -882,27 +906,34 @@ private slots:
...
+        QVERIFY(fakeFolder.currentLocalState().find(fileName(QStringLiteral("f1"))));
+        QVERIFY(fakeFolder.currentLocalState().find(fileName(QStringLiteral("A/a1"))));
+        QVERIFY(fakeFolder.currentLocalState().find(fileName(QStringLiteral("A/a3"))));
+        QVERIFY(fakeFolder.currentLocalState().find(fileName(QStringLiteral("A/B/b1"))));
 
         // Make local changes to a3
-        fakeFolder.localModifier().remove(QStringLiteral("A/a3") + Theme::instance()->appDotVirtualFileSuffix());
-        fakeFolder.localModifier().insert(QStringLiteral("A/a3") + Theme::instance()->appDotVirtualFileSuffix(), 100_b);
+        fakeFolder.localModifier().remove(fileName(QStringLiteral("A/a3")));
+        fakeFolder.localModifier().insert(fileName(QStringLiteral("A/a3")), 100_b); // TODO: should this really have the suffix-vfs file suffix??
```
erikjv, line 916: @TheOneRing What do you think?

TheOneRing:
ci looks good: https://github.com/owncloud/client-desktop-vfs-win/actions/runs/7020419509/job/19100078148

test/testsyncvirtualfiles.cpp
TheOneRing: Can you check what happened here previously?
With suffix files that should result in an invalid placeholder that was no longer synced.
So after this change the local and remote state should differ, we should have a "A/a3.owncloud" locally but no corresponding file on the remote.

erikjv:
The test also passes with WinVFS: https://github.com/owncloud/client-desktop-vfs-win/actions/runs/7118184195

TheOneRing (approved):

TheOneRing (approved):

Issue #11243, opened 2023-09-28
The error message in "Add sync folder" dialog is unreadable with dark theme


erikjv:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the bug

Click the "Add sync folder" button for an account, and enter an existing folder. The error is (probably) in white text on a white background:

<img width="912" alt="Screenshot 2023-09-28 at 13 10 43" src="https://github.com/owncloud/client/assets/151956/17568852-11d9-4232-96cf-ae0e83ca268d">

The label in `folderwizardsourcepage.ui` uses a custom palette, which is probably the cause.

### Expected behavior

The error message should at least be readable.

### Client version number

The 5 branch in git, sha1: `bcab27aefda27a33af7bcb70e229ecff99b95363`

TheOneRing:
We should remove all questionable uses of QPalette

erikjv:
There is now also a dark theme on Windows, so check that too.

erikjv:
The dark theme on Windows shows the same problem. So either can be used for testing.

PR #11401, opened 2023-11-28
Remove styling from the warning label in the "Add Folder" dialog

erikjv:
The styling caused unreadable warning/error messages when the OS theme was set to dark: the text color would be white, but the background was styled to be white too.

The message is still noticable: normally there is no text in the label, so an error or warning message showing up will be noticed.

Fixes: https://github.com/owncloud/client/issues/11243


TheOneRing (approved):
Was that the only occasion where we modified the palette?

erikjv:
> Was that the only occasion where we modified the palette?

I haven't checked this yet

Issue #9833, opened 2022-06-30
[QA] Run automated GUI tests in Windows


Talank:
### Describe the QA issue

Currently we are running all the automated tests in ubuntu, but with some changes, we should be able to run the tests in Windows as well (at least run it locally).

### Steps to reproduce the issue

1. Download ownCloud client (already built) and install it in windows.
2. Set up ownCloud core (if you are using virtual machine, then you can use the core that you have set up in your host, you might need to configure the host file of the virtual machine and trusted_domains domains on core config)
3. Download squish for windows (for example squish-6.7.2-qt512x-win64-msvc141.exe) and install it
4. Install python 3.8 on windows
5. Download [PythonChanger.py](https://kb.froglogic.com/squish/howto/using-external-python-interpreter-squish-6-6/PythonChanger.py) script and save it inside your squish installation directory
6. run `python PythonChanger.py` from squish installation directory
7. Run [owncloud-test-middleware](https://github.com/owncloud/owncloud-test-middleware) (just like core, you can run it in host machine, if you are using virtual machine for windows. Alternatively, you can run middleware in drone). You might need node installed if you want to run middleware locally.
8. Install required python libraries
    - pip install requests
    - pip install pycairo
    - pip install PyGObject

**NOTE the `PyGObject` library is not available in windows. You can install this library using mingw as shown [here](https://pygobject.readthedocs.io/en/latest/getting_started.html). However, the problem so far is that this tool can not be used to install python 3.8 so far. And we need python 3.8 for squish.**

## Open tasks
- [x] Tests adjustment https://github.com/owncloud/client/pull/11406
- [x] Run tests with oc10 https://github.com/owncloud/client/pull/11406
- [x] Run tests with ocis https://github.com/owncloud/client/pull/11436
~- [ ] Unskip tests (if any)~ This will be tracked in another ticket https://github.com/owncloud/client/issues/11449
- [x] Docs on running squish tests in Windows https://github.com/owncloud/docs-client-desktop/pull/554

Tests skipped for Windows:
- `Sharing link with expiration date` (difference in date format)
- `Syncing a folder having space at the end` (white space is trimmed)
- `Filenames that are rejected by the server are reported` (invalid file name for windows)
- `Sync long nested folder` (windows doesn't allow long path name)
- `Invalid system names are synced in linux` (linux only)

Feature disabled for Windows:
- Screenshots
- Stacktracing

Talank:
[...]

PR #11406, opened 2023-11-30
[full-ci][gui-tests] Automated GUI tests in windows

saw-jan:
Part of https://github.com/owncloud/client/issues/9833

From the initial work on squish tests for Windows, it looks like the current tests can be run in Windows as well with some tweaking.
Almost all the tests use the sync pattern matching to determine whether the sync has been successful or not. In Linux, we utilize socket connection and Windows has named pipe for communication.

Tests skipped for Windows:
- `Sharing link with expiration date` (difference in date format)
- `Syncing a folder having space at the end` (white space is trimmed)
- `Filenames that are rejected by the server are reported` (invalid file name for windows)
- `Sync long nested folder` (windows doesn't allow long path name)
- `Invalid system names are synced in linux` (linux only)

Feature disabled for Windows:
- Screenshots
- Stacktracing

Tasks:
- [x] adjust tests for Windows
- [x] run tests with oc10 server
- [ ] run tests with ocis server (problem with cleanup)


test/gui/shared/scripts/helpers/WinPipeHelper.py
```diff
@@ -0,0 +1,115 @@
...
+                _, message_mem = win32file.ReadFile(self._pipe, DEFAULT_BUFLEN, self._overlapped)
+                if message_mem:
+                    m_bytes = bytes(message_mem)
+                    if b'\n' in m_bytes:
+                        for m in m_bytes.split(b'\n'):
+                            print(type(m))
+                            try:
+                                # append decodable bytes
+                                m.decode('utf-8')
+                                messages += m + b'\n'
```
saw-jan, line 90: filter out the unwanted messages

TheOneRing:
> @TheOneRing what does this error mean?
> 
> Screenshot from 2023-11-30 16-19-08
> 
> [client.log](https://github.com/owncloud/client/files/13511685/client.log)

It usually means that a space that we synced was removed/disabled on the server.

test/gui/shared/scripts/helpers/ConfigHelper.py
```diff
@@ -49,14 +84,12 @@ def get_config_from_env_file(env):
...
-        get_config_from_env_file("XDG_CONFIG_HOME"), "ownCloud"
-    ),
-    'guiTestReportDir': os.path.abspath('../reports/'),
+    'clientRootSyncPath':  getClientRootPath(),
+    'tempFolderPath':  path.join(getClientRootPath(), 'temp'),
+    'clientConfigDir': getConfigHome(),
+    'guiTestReportDir': path.abspath(path.join('..', 'reports')),
     'ocis': False,
-    'custom_lib': os.path.abspath('../shared/scripts/custom_lib'),
+    'custom_lib': path.abspath(path.join('..', 'shared', 'scripts', 'custom_lib')),
```
TheOneRing, line 92: a more modern approach for paths is 
```
from pathlib import Path
foo = Path("..") / "shared/scrips/custom_lib"
or
foo = Path("../shared/scrips/custom_lib")
```
But actually 90% of windows accepts forwards slashes in paths.

test/gui/shared/scripts/helpers/SetupClientHelper.py
```diff
@@ -107,29 +108,29 @@ def getPollingInterval():
...
-    0/dav_user={davUserName}
-    0/display-name={displayUserName}
-    0/http_CredentialVersion=1
-    0/http_oauth={oauth}
-    0/http_user={davUserName}
-    0/url={local_server}
-    0/user={displayUserFirstName}
-    0/version=1
+    0{path_sep}Folders{path_sep}1{path_sep}davUrl={url}
+    0{path_sep}Folders{path_sep}1{path_sep}ignoreHiddenFiles=true
```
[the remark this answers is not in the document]
saw-jan: config in windows has `\` so I was trying to make similar to it

test/gui/shared/scripts/helpers/WinPipeHelper.py
```diff
@@ -0,0 +1,133 @@
+import os
```
saw-jan, line 1: @TheOneRing Could you review this PR? Specially this file

test/gui/shared/scripts/helpers/SyncHelper.py
```diff
@@ -63,7 +66,23 @@
         # when syncing an account that has some files/folders
         [SYNC_STATUS['SYNC'], SYNC_STATUS['OK']],
     ],
-    'synced': [SYNC_STATUS['SYNC'], SYNC_STATUS['OK']],
+    'synced': [
+        [
+            SYNC_STATUS['SYNC'],
```
TheOneRing, line 71: This looks a tiny bit weird?

test/gui/shared/scripts/helpers/WinPipeHelper.py
```diff
@@ -0,0 +1,133 @@
...
+import time
+import win32pipe, win32file, winerror, pywintypes, win32event
+
+TIMEOUT = 100
+DEFAULT_BUFLEN = 4096
+CLIENT_MESSAGES = [
+    'STATUS',
```
TheOneRing, line 8: They are the same on all platforms maybe we can share them?

test/gui/shared/scripts/helpers/WinPipeHelper.py
TheOneRing: Looks good.

test/gui/shared/scripts/helpers/WinPipeHelper.py
saw-jan: yeah, but as of now I don't see other usages. we don't need to filter out socket messages in linux and also the socket connection helper file is from https://raw.github.com/owncloud/client-desktop-shell-integration-nautilus/master/src/syncstate.py, so I won't dare to edit it :sweat_smile:

saw-jan:
> > * `Sharing link with expiration date` (difference in date format)
> 
> How do they differ?
> 
> > * `Syncing a folder having space at the end` (white space is trimmed)
> > * `Filenames that are rejected by the server are reported` (invalid file name for windows)
> > * `Sync long nested folder` (windows doesn't allow long path name)
> 
> Hmmm https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry#enable-long-paths-in-windows-10-version-1607-and-later or https://stackoverflow.com/questions/1365797/python-long-filename-support-broken-in-windows
> 
> > * `Invalid system names are synced in linux` (linux only)

Thanks for the useful links :smiley:

I would like to follow up on these tests after this PR along with the test cleanup while testing against the ocis server.

TheOneRing (approved):

Issue #7592, opened 2019-11-14
Update available shows version number instead of version string


HanaGemela:
Client: 2.6.0
macOS 10.15

Steps to recreate:
1) Make sure there is new unstable update available in the online checker
2) Switch to beta channel in the client

Actual behaviour: 

![image](https://user-images.githubusercontent.com/49001702/68853297-efdb0680-06d9-11ea-97c6-db2508e3af7c.png)

Available version string is shown so the user doesn't know if it is stable 2.6.0 or beta or RC... 

Expected result: Message shows version string so it is obvious what is going to be installed

HanaGemela:
Issue where the version has been added https://github.com/owncloud/client/issues/7288

HanaGemela:
@ogoffart the issue is that it is not obvious if it is final or RC or alpha.... That's more interesting than 12665

michaelstingl:
Needs to be fixed in updater script first.

TheOneRing:
See https://github.com/owncloud/client-updater-server/issues/98

fmoc:
Fixed in the update server, see the issue linked above.

HanaGemela:
<img width="617" alt="Screenshot 2024-01-08 at 15 59 22" src="https://github.com/owncloud/client/assets/49001702/57185439-5263-40d7-98af-73a011804e4b">
So the info about testing version is there, just the format is different

TheOneRing:
We now display a string instead of just a version, in case of the fake updater we scream fake update.
In realty we would display version + alpha or something like that.

PR #11414, opened 2023-12-01
Improve version displayed in Sparkle Updater

fmoc:
Since https://github.com/owncloud/client/issues/7592 is implemented, the first version number in the update dialog is a human readable one, as shown in the screenshot below. However, the current version is formatted in the same way the future version used to be.

screenshot_2023-12-01_15-52-38

Since 2021, we must have changed its rendering in the `Info.plist` in the bundle, causing this very technical number that further does not include alpha/beta/rc suffixes to be shown. We should change that.

Please note that while the former version number format is available to all old versions as it is sent by the update server, the latter is included in the installed bundle. This change will only affect 6.x+.


fmoc:
To test, launch a daily (once this is merged) using the force-update demo update server and check if the local version (the right one) is formatted properly (e.g., `6.0.0-git (build XXXXX)` instead of `6.0.0.XXXXX`).

src/gui/MacOSXBundleInfo.plist
```diff
@@ -23,7 +23,7 @@
         <key>CFBundleVersion</key>
         <string>${MIRALL_VERSION_FULL}</string>
         <key>CFBundleShortVersionString</key>
-        <string>${MIRALL_VERSION_STRING}</string>
+        <string>${MIRALL_BUNDLE_SHORT_VERSION_STRING}</string>
```
TheOneRing, line 26: https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleshortversionstring

TheOneRing (changes requested):
According to the mac docs this is wrong

PR #11417, opened 2023-12-05
Add extensive logging to CPU arch detection

fmoc:
Related to #10757 which does not work reliably.


src/common/utility_mac.mm
```diff
@@ -263,14 +263,31 @@
...
+
+        if (error == ENOENT) {
+            qCDebug(lcUtility) << "not running on Apple Silicon";
+        } else {
+            qCDebug(lcUtility) << "unknown error trying to detect whether proc is translated";
         }
 
-        return QSysInfo::currentCpuArchitecture();
+        const auto qtArch = QSysInfo::currentCpuArchitecture();
+        qCDebug(lcUtility) << "CPU arch reported by Qt:" << qtArch;
```
TheOneRing, line 290: return qtArch

fmoc:
Please note that this logging most likely occurs before logging is initialized properly and therefore does not show up in the logfile. This is due to all the caching of strings in the client which are generated statically once during startup as far as I can see. We can only work around this if we give up on caching of some aspects or make sure logging is initialized beforehand. Some lazy loading getter approach might work, but this will add some complexity.

PR #11437, opened 2023-12-13
[gui-tets][full-ci] Fix sync pattern checking for a single resource

saw-jan:
Fix sync pattern matching of a single resource. This will reduce the test execution time.


grgprarup (approved):
:+1:

Issue #11276, opened 2023-10-11
To support AD FS make redirect_uri port range configurable


butonic:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Feature description

I would like to authenticate users with an on-premise AD FS 2022 using OpenID Connect. While the documentation for AzureID has documented [`redirect_url` restrictions](https://learn.microsoft.com/en-us/azure/active-directory/develop/reply-url), especially for [`localhost`](https://learn.microsoft.com/en-us/azure/active-directory/develop/reply-url#localhost-exceptions) registering any of these withg AD FS does not work:
```
http://127.0.0.1/callback/
http://127.0.0.1/callback
http://127.0.0.1:48395/callback/
http://localhost/callback
http://localhost:55555/callback
https://localhost/callback
http://127.0.0.1:*
```

Only `redirect_uri` entries that include a port will let the client authenticate. All other cases lead to an error like this:
```
    Activity ID: a8f9f18b-dd70-47bf-ba00-0040000000ee
    Error details: MSIS9224: Received invalid OAuth authorization request. The received 'redirect_uri' parameter is not a valid registered redirect URI for the client identifier: '01d03691-b75e-4828-b93d-a33ac2ed6631'. Received redirect_uri: 'http://127.0.0.1:62216'.
    Node name: 4210c302-325f-4e8d-8d99-3aa76743da3e
    Error time: Tue, 10 Oct 2023 23:16:39 GMT
    Cookie: enabled
    User agent string: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:109.0) Gecko/20100101 Firefox/118.0
```

### Proposed solution

Intead of letting the client randomly pick a port allow configuring a port range. The client then randomly picks one of those ports until he finds a free port. Admins can brand clients to only use a limited range of ports and register them as valid `redirect_uri` entries in AD FS.

### Alternative solutions you considered

We could hardcode a single port, but that will lead to corner cases where the port is in use by another application.

[...]

PR #11439, opened 2023-12-14
Add branding option for oauth local port

TheOneRing:
Fixes: #11276


src/libsync/creds/oauth.cpp
```diff
@@ -249,9 +249,18 @@ void OAuth::startAuthentication()
     qCDebug(lcOauth) << "starting authentication";
 
     // Listen on the socket to get a port which will be used in the redirect_uri
-    if (!_server.listen(QHostAddress::LocalHost)) {
+    auto ports = Theme::instance()->oauthPorts();
+    while (!ports.isEmpty()) {
```
TheOneRing, line 253: use a for loop

erikjv (approved):

Issue #11346, opened 2023-11-08
Retest and automate complicated Windows VFS tests


HanaGemela:
Repeat the manual tests with proper file structure

Tests with large complicated file structures need to be prioritised for automation.

https://owncloud.testrail.io/index.php?/tests/view/5026

Actions needed

- [x] retest all the test cases manually
- [x] automate if possible
- [ ] Revisit https://owncloud.testrail.io/index.php?/tests/view/5019

### Test cases (from https://github.com/owncloud/QA/issues/789#issuecomment-1706180402)

Files operations 

| Priority | ID  | Test Case | Steps to reproduce | Expected Result | 
| --------- | --- | --------- | ------------------ | --------------- | 
| 4 :heavy_check_mark: | 1  | Rename a file | 1. Rename a file | File has been renamed on the server as well | 
| 4 :heavy_check_mark: | 1 | Delete a file | 1. Delete a file  | File has been deleted on the server as well 
| 4 :heavy_check_mark: | 1  | Edit content of a file | 1. Edit content of a file  | File has been downloaded and edited content has been synced on the server
| 4 :heavy_check_mark: | 1 | Create a new file | 1. Create a new file  | New file has been added and synced but is not virtual https://github.com/owncloud/client/pull/11466
| 4 :heavy_check_mark: | 1 | Copy and paste a file | Copy a virtual file and Paste it at the same location   | File browser asks or creates a copy https://github.com/owncloud/client/pull/11530
| 4 :heavy_check_mark: | 1 | Copy and paste a file | :heavy_check_mark: 1. Copy virtual file and Paste it outside of owncloud sync folder https://github.com/owncloud/client/pull/11466 <br> :heavy_check_mark: 2. Copy virtual file and Paste it inside a folder  https://github.com/owncloud/client/pull/11506 | Placeholder file gets downloaded and it has been copied
| 4 :heavy_check_mark: | 1 | Move a virtual file | Move a virtual file  | Virtual as well as the physical file on the server have been moved https://github.com/owncloud/client/pull/11506

VFS and Selective sync are mutually exclusive

| Priority | ID  | Test Case | Steps to reproduce | Expected Result |
| --------- | --- | --------- | ------------------ | --------------- |
| 4 :heavy_check_mark: | 1 | VFS and Selective sync | 1. Verify when VFS is on, user cannot select folders for selective sync | When VFS is on, user cannot select folders for selective sync https://github.com/owncloud/client/pull/11452
[...]

Issue #11331, opened 2023-11-03
[5.1][win] Disabling VFS deletes the dehydrated files


saw-jan:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the bug

Disabling VFS deletes all the dehydrated files (files with cloud icon) but files that are already downloaded still do not get deleted. Files are also deleted from the server

### Expected behavior

Download the cloud files

### Steps to reproduce the issue

1. Upload some files and a folder with some files in it
2. Install client 5.1 (https://download.owncloud.com/desktop/ownCloud/testing/5.1.0.12543/win/)
3. Add account - VFS enabled
4. Open some files from file-manager
5. Disabled VFS in client (:x: dehydrated files get deleted)

### Screenshots

![Screenshot from 2023-11-03 14-19-20](https://github.com/owncloud/client/assets/52366632/13297187-9773-4d13-a4f2-adc9631c9c6a) | ![Screenshot from 2023-11-03 14-19-32](https://github.com/owncloud/client/assets/52366632/83829f1b-f4c0-4dbf-8acf-5c24ccee7ce9)
--|--

### Logs
[ownCloud.log](https://github.com/owncloud/client/files/13248298/ownCloud.log)

### Client version number

ownCloud 5.1.0.12543 [652b67](https://github.com/owncloud/client/commit/652b67d1e5915b6a736b5f5858e38589060ac929)
Libraries Qt 6.4.3, OpenSSL 3.1.2 1 Aug 2023
Using virtual files plugin: wincfapi
OS: windows-10.0.22621
QPA: windows

### Installation path (Windows only)

WIndows 11

HanaGemela:
Passed in 5.1.1 rc1 win

PR #11451, opened 2023-12-27
[gui-tests][full-ci] Assert the presence of dehydrated file after VFS disabled

Salipa-Gurung:
This PR extends the scenario for `Disable/Enable VFS` in windows to verify that the dehydrated file are present after disabling VFS. 

Related Issue: https://github.com/owncloud/client/issues/11346

Added test coverage for https://github.com/owncloud/client/issues/11331

### DIFF
--- a/test/gui/shared/scripts/helpers/FilesHelper.py
+++ b/test/gui/shared/scripts/helpers/FilesHelper.py
@@ -1,5 +1,7 @@
 import os
 import re
+import ctypes
+from helpers.ConfigHelper import isWindows
 
 
 def buildConflictedRegex(filename):
@@ -78,3 +80,14 @@ def get_size_in_bytes(size):
                 return size_num * (multiplier**3)
 
     raise Exception("Invalid size: " + size)
+
+
+def get_file_size_on_disk(resource_path):
+    file_size_high = ctypes.c_ulonglong(0)
+    if isWindows():
+        return ctypes.windll.kernel32.GetCompressedFileSizeW(
+            ctypes.c_wchar_p(resource_path), ctypes.pointer(file_size_high)
+        )
+    raise Exception(
+        "'get_file_size_on_disk' function is only supported for Windows OS."
+    )
--- a/test/gui/shared/steps/vfs_context.py
+++ b/test/gui/shared/steps/vfs_context.py
@@ -1,0 +1,22 @@
+from helpers.FilesHelper import get_file_size_on_disk
+
+
+@Then('the placeholder of file "|any|" should exist on the file system')
+def step(context, file_name):
+    resource_path = getResourcePath(file_name)
+    size_on_disk = get_file_size_on_disk(resource_path)
+    test.compare(
+        size_on_disk, 0, f"Size of the placeholder on the disk is: '{size_on_disk}'"
+    )
+
+
+@Then('the file "|any|" should be downloaded')
+def step(context, file_name):
+    resource_path = getResourcePath(file_name)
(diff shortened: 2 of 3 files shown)


test/gui/tst_vfs/test.feature
```diff
@@ -8,8 +8,21 @@ Feature: Enable/disable virtual file support
...
         Given user "Alice" has been created on the server with default attributes and without skeleton files
+        And user "Alice" has created folder "folder1" in the server
+        And user "Alice" has uploaded file with content "ownCloud" to "folder1/lorem.txt" in the server
+        And user "Alice" has uploaded file with content "Lorem Ipsum is simply dummy text of the printing and typesetting industry." to "testFile.txt" in the server
         And user "Alice" has set up a client with default settings
+        Then the file "testFile.txt" should exist on the file system
+        And the file "folder1/lorem.txt" should exist on the file system
         When the user disables virtual file support
         Then the "Enable virtual file support..." button should be available
+        And the file "folder1/lorem.txt" should exist on the file system with the following content
```
saw-jan, line 19: is it possible to add the bytes assertion before checking content? IMO, that way we are asserting that the file is downloaded (with vfs disabled)

test/gui/shared/scripts/helpers/FilesHelper.py
```diff
@@ -78,3 +80,11 @@ def get_size_in_bytes(size):
...
 
     raise Exception("Invalid size: " + size)
+
+
+def get_file_size_on_disk(resource_path):
+    file_size_high = ctypes.c_ulonglong(0)
+    if isWindows():
+        return ctypes.windll.kernel32.GetCompressedFileSizeW(
+            ctypes.c_wchar_p(resource_path), ctypes.pointer(file_size_high)
+        )
```
saw-jan, line 90: else raise some exception

test/gui/shared/steps/vfs_context.py
```diff
@@ -0,0 +1,22 @@
...
+
+@Then('the file "|any|" should be downloaded')
+def step(context, file_name):
+    resource_path = getResourcePath(file_name)
+    size_on_disk = get_file_size_on_disk(resource_path)
+    file_size = os.stat(resource_path).st_size
+    test.compare(
+        size_on_disk,
+        file_size,
+        f"Size of original file: '{file_size}' is not equal to it's size on disk '{size_on_disk}'",
```
saw-jan, line 21:
```suggestion
        f"Original file size '{file_size}' is not equal to its size on disk '{size_on_disk}'",
```

test/gui/tst_vfs/test.feature
```diff
@@ -8,8 +8,23 @@ Feature: Enable/disable virtual file support
...
+        And the file "testFile.txt" should exist on the file system with the following content
+            """
+            ownCloud
+            """
+        And the file "folder1/lorem.txt" should exist on the file system with the following content
+            """
+            some contents
+            """
+        And the file "testFile.txt" should be downloaded
+        And the file "folder1/lorem.txt" should be downloaded
```
saw-jan, line 28: download check must be before reading content (because content read action downloads the file)
```suggestion
        And the file "testFile.txt" should be downloaded
        And the file "folder1/lorem.txt" should be downloaded
        And the file "testFile.txt" should exist on the file system with the following content
            """
            ownCloud
            """
        And the file "folder1/lorem.txt" should exist on the file system with the following content
            """
            some contents
            """
```

saw-jan (approved):
:sparkles:

PR #11463, opened 2024-01-05
Fix sizeliteral naming

TheOneRing:


src/common/filesystembase.h
```diff
@@ -187,24 +187,42 @@ namespace FileSystem {
...
 
-        constexpr unsigned long long operator"" _kb(unsigned long long sz)
+        constexpr unsigned long long operator"" _KiB(unsigned long long sz)
+        {
+            return operator"" _B(sz) * BinaryBase;
+        }
+
+        constexpr unsigned long long operator"" _MiB(unsigned long long sz)
+        {
+            return operator"" _KiB(sz) * BinaryBase;
```
fmoc, line 205: `kilo` prefix is usually lowercase (as the only exception in the >1 prefixes), so this should be `kiB`. Same goes for `kB`.

fmoc (changes requested):

src/common/filesystembase.h
TheOneRing: True for kB but not for KiB https://en.wikipedia.org/wiki/Byte#Multiple-byte_units

fmoc (approved):

Issue #11412, opened 2023-11-30
Another unreadable error message


erikjv:
After fixing issue #11243 , we found three more places with styled error messages. Two are in legacy dialogs (and accidentally are actually readable), but the third is a problem. This is in the `folderwizardtargetpage.ui`:

<img width="503" alt="Screenshot 2023-11-30 at 18 12 09" src="https://github.com/owncloud/client/assets/151956/f01f529a-ac9e-4fc7-acbd-bf93f2a9e3e4">

It is the selected label which reads "Placeholder text", a white text on a yellow background.

One way to fix this is to do the same as in the log browser UI:

<img width="588" alt="Screenshot 2023-11-30 at 18 09 13" src="https://github.com/owncloud/client/assets/151956/4cb509cb-3a17-4127-851a-0a60d5916b29">

The question is if we want the red text or un-styled text, and if the latter is the case, if we want to change the log browser UI too.

TheOneRing:
![image](https://github.com/owncloud/client/assets/200626/4ca9c5a4-293b-4ffa-924c-540dcaa81db1)

saw-jan:
Tested on `ownCloud 6.0.0.13582-daily20240327 [4258b6]`

![Screenshot from 2024-03-28 15-40-51](https://github.com/owncloud/client/assets/52366632/99ae7c2e-a2fe-4cee-9988-e23f14593d76) | ![Screenshot from 2024-03-28 15-42-16](https://github.com/owncloud/client/assets/52366632/14d2c7eb-ca50-4b2a-948a-211032648d85)
--|--
![Screenshot from 2024-03-28 15-47-13](https://github.com/owncloud/client/assets/52366632/41242606-967e-4fcb-954f-96aa2c72c31f)

PR #11470, opened 2024-01-12
Fix badly styled warning/error messages

erikjv:
Some error messages were styled in a way that is not compatible with dark mode. These have now been replaced by unstyled labels, but the warning icon is shown to the left of the message.

Fixes: #11412

### DIFF
--- a/src/gui/folderwizard/folderwizardremotepath.cpp
+++ b/src/gui/folderwizard/folderwizardremotepath.cpp
@@ -40,7 +40,8 @@ FolderWizardRemotePath::FolderWizardRemotePath(FolderWizardPrivate *parent)
 
 {
     _ui->setupUi(this);
-    _ui->warnFrame->hide();
+    _ui->warningIcon->setPixmap(Resources::getCoreIcon(QStringLiteral("warning")).pixmap(_ui->warningIcon->size()));
+    _ui->warningFrame->hide(); // This is `show()`n in `FolderWizardRemotePath::showWarn`
 
     _ui->folderTreeWidget->setSortingEnabled(true);
     _ui->folderTreeWidget->sortByColumn(0, Qt::AscendingOrder);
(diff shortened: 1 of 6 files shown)


src/gui/folderwizard/folderwizardremotepath.cpp
```diff
@@ -40,7 +40,8 @@ FolderWizardRemotePath::FolderWizardRemotePath(FolderWizardPrivate *parent)
 
 {
     _ui->setupUi(this);
-    _ui->warnFrame->hide();
+    _ui->warningIcon->setPixmap(Resources::getCoreIcon(QStringLiteral("warning")).pixmap(_ui->warningIcon->size()));
+    _ui->warningFrame->hide();
```
TheOneRing, line 44: Where is show() called?

src/gui/folderwizard/folderwizardremotepath.cpp
```diff
@@ -361,10 +362,9 @@ void FolderWizardRemotePath::initializePage()
...
 {
     if (msg.isEmpty()) {
-        _ui->warnFrame->hide();
-
+        _ui->warningFrame->hide();
     } else {
-        _ui->warnFrame->show();
-        _ui->warnLabel->setText(msg);
+        _ui->warningFrame->show();
+        _ui->warningLabel->setText(QStringLiteral("<html><body><b>%1</b></body></html>").arg(msg));
```
TheOneRing, line 368: Qt rhichtext might be a bit shorter than full html :)

TheOneRing (changes requested):
Please also add before and after screenshots.

src/gui/folderwizard/folderwizardremotepath.cpp
erikjv: line 367, see your comment below

src/gui/folderwizard/folderwizardremotepath.cpp
erikjv: done

saw-jan:
Tested: https://github.com/owncloud/client/issues/11412#issuecomment-2024822852

Issue #11471, opened 2024-01-15
[Metered connection] Hide feature on unsupported platforms


michaelstingl:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Feature description

"Metered connection" feature isn't supported on macOS, since it should be hidden, not disabled. There's nothing the user can do on macOS to resolve this problem.

### Proposed solution

Please hide:

![CleanShot 2024-01-15 at 15 50 58](https://github.com/owncloud/client/assets/214010/8b50a9e1-9b49-404c-8353-512b7331b7c7)

### Additional context

Tested with:

```
ownCloud 6.0.0.13116-daily20240115 [0d34c9](https://github.com/owncloud/client/commit/0d34c9844f69cba89b687ee093b2838bf1114647)
Libraries Qt 6.6.0, OpenSSL 3.1.4 24 Oct 2023
Using virtual files plugin: suffix
OS: macos-23.2.0 (build arch: arm64, CPU arch: arm64)
QPA: cocoa
```

michaelstingl:
Related:
- https://github.com/owncloud/client/pull/11305
- https://github.com/owncloud/client/issues/4808

Should be explained in the docs:
- https://github.com/owncloud/docs-client-desktop/issues/535

saw-jan:
Tested in `Ubuntu 22.04` with `ownCloud 6.0.0.13155-daily20240119`
- Option is hidden for unsupported platform :heavy_check_mark: 
![Screenshot from 2024-01-19 13-03-03](https://github.com/owncloud/client/assets/52366632/6ccb9638-8165-4d84-b5fe-dfc5d6a9881d)

saw-jan:
@TheOneRing if we can test it in macOS then we can move it to Done.

PR #11473, opened 2024-01-15
Hide pause-sync-when-metered ceckbox when not supported

erikjv:
On platforms where metered connection detection is not available, hide the checkbox instead of disabling it.

Fixes: #11471


TheOneRing (approved):

saw-jan:
Tested. see https://github.com/owncloud/client/issues/11471#issuecomment-1899890646

Issue #11472, opened 2024-01-15
Properly encode / sanitize URL's (Example: authorization_endpoint)


michaelstingl:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the bug

ownCloud Desktop sync client generates invalid URL's with Space characters:

```
https://keycloak.ocis-keycloak.latest.owncloud.works/realms/oCIS/protocol/openid-connect/auth?response_type=code
&client_id=c604798c-7977-46a5-8beb-e3df13e65413
&redirect_uri=http://127.0.0.1:54213
&code_challenge=5G3yQGITsH_CPkQDIXL7irH5rEN_BydI3s0YOCr7fHs
&code_challenge_method=S256
&scope=openid offline_access email profile
&prompt=select_account consent
&state=G_jOhNJru34iqmG1ULYA04Gd7GYY7ubu22z-gkmPyzg%3D
```

### Expected behavior

URL's should be properly encoded / sanitized.

### Screenshots

![CleanShot 2024-01-15 at 16 06 07](https://github.com/owncloud/client/assets/214010/4f65d02b-7e99-4e4a-a8bf-7097a375104f)

### Client version number

```
ownCloud 6.0.0.13116-daily20240115 [0d34c9](https://github.com/owncloud/client/commit/0d34c9844f69cba89b687ee093b2838bf1114647)
Libraries Qt 6.6.0, OpenSSL 3.1.4 24 Oct 2023
Using virtual files plugin: suffix
OS: macos-23.2.0 (build arch: arm64, CPU arch: arm64)
QPA: cocoa
```

### Server information

https://ocis.ocis-keycloak.latest.owncloud.works/

```
[Log]  ownCloud Web UI 8.0.0-rc.1  (index.html-MZztZoFM.mjs, line 1)
[Log]  Infinite Scale 5.1.0-prealpha+21e4b7d29 Community  (index.html-MZztZoFM.mjs, line 1)
```

### Additional context

Doesn't work in curl:

```
[...]

PR #11479, opened 2024-01-17
Properly encode scope and prompt items in OAuth URL

erikjv:
The scope and the prompt items of an OAuth query can be branded and can contain characters that are not valid without encoding them. This change makes sure that those get encoded properly.

Fixes: #11472


TheOneRing:
Changelog pls

src/libsync/creds/oauth.cpp
```diff
@@ -466,14 +466,12 @@ QUrl OAuth::authorisationLink() const
...
-        { QStringLiteral("redirect_uri"), QStringLiteral("%1:%2").arg(_redirectUrl, QString::number(_server.serverPort())) },
-        { QStringLiteral("code_challenge"), QString::fromLatin1(code_challenge) },
-        { QStringLiteral("code_challenge_method"), QStringLiteral("S256") },
-        { QStringLiteral("scope"), Theme::instance()->openIdConnectScopes() },
-        { QStringLiteral("prompt"), Theme::instance()->openIdConnectPrompt() },
-        { QStringLiteral("state"), QString::fromUtf8(_state) } };
+    QUrlQuery query{{QStringLiteral("response_type"), QStringLiteral("code")}, {QStringLiteral("client_id"), _clientId},
+        {QStringLiteral("redirect_uri"), QStringLiteral("%1:%2").arg(_redirectUrl, QString::number(_server.serverPort()))},
+        {QStringLiteral("code_challenge"), QString::fromLatin1(code_challenge)}, {QStringLiteral("code_challenge_method"), QStringLiteral("S256")},
+        {QStringLiteral("scope"), QString::fromUtf8(QUrl::toPercentEncoding(Theme::instance()->openIdConnectScopes()))},
```
fmoc, line 472: Can't we URL encode the entire result? Or does that cause any trouble?

TheOneRing: https://doc.qt.io/qt-6/qurl.html#toEncoded

fmoc: That was my idea as well. I thought we could just build the URL as-is and return `url.toEncoded()`, basically. Encoding selected parameters doesn't ensure that the end result will work, as more components might require encoding. Also, I could imagine that already encoded parameters might be re-encoded (i.e., broken) by `toEncoded()` later on.

erikjv: A couple lines below is:
```
        const QString davUser = QString::fromUtf8(QUrl::toPercentEncoding(_davUser)); // Issue #7762;
        // open id connect
        query.addQueryItem(QStringLiteral("login_hint"), davUser);
```
So I copied that. Also, we need to return a `QUrl`, and the encoding gives a `QByteArray`. Also, `QDesktopServices::openUrl` takes a URL, so my guess is that the parameters should be properly encoded, and that the url is passed without extra encoding to the system webbrowser.

fmoc: Quoting `addQueryItem`'s docs, "the key and value strings are expected to be in percent-encoded form", so we should pipe _all_ values through `toPercentEncoding` anyway. Nevertheless, we should use `QUrl::toEncoded` in the end, as the parameters might be decoded otherwise.

I think we need unit tests for this....

erikjv: done

saw-jan:
Tested: https://github.com/owncloud/client/issues/11472#issuecomment-1923632959

Issue #11480, opened 2024-01-18
[6] Re-login with invalid password doesn't show the login dialog again


saw-jan:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the bug

Client stuck in "Updating credentials ..." state while trying to login with invalid password

### Expected behavior

When trying to login with invalid password, client is expected to show the login dialog again

### Steps to reproduce the issue

1. Add an oC10 account to the client
2. Logout
3. Try to login with INVALID password
4. Client stuck in "updating cred..."

### Screenshots

![Screenshot from 2024-01-18 11-46-03](https://github.com/owncloud/client/assets/52366632/ee965d9b-9ac0-4f2e-9db0-905f0d92cbcc) | ![Screenshot from 2024-01-18 11-46-15](https://github.com/owncloud/client/assets/52366632/d10eac0f-bbd6-4565-9064-e9176523a027)
--|--
![Screenshot from 2024-01-18 11-46-21](https://github.com/owncloud/client/assets/52366632/256a6321-a6c1-41cc-a52a-fbf5bfdc481a)

### Logs

[ownCloud.log](https://github.com/owncloud/client/files/13972709/debug-ownCloud.log)

### Client version number

ownCloud 6.0.0-git [24a6f4](https://github.com/owncloud/client/commit/24a6f4bc2fbbfc10bdce61baef52283083bc1762)
Libraries Qt 6.4.3, OpenSSL 3.0.2 15 Mar 2022
Using virtual files plugin: suffix
OS: ubuntu-6.2.0-39-generic (build arch: x86_64, CPU arch: x86_64)
QPA: xcb

### Server information

ownCloud 10.13.4 (stable)

TheOneRing:
Supposedly fixed by https://github.com/owncloud/client/commit/f66985b27f1e466fa35d0f746b8e438fd59bea5e

saw-jan:
Tested with automated tests: https://drone.owncloud.com/owncloud/client/17706

PR #11484, opened 2024-01-18
Partly revert 11478 to fix regression #11480

TheOneRing:


fmoc (approved):

saw-jan:
Tested with `ownCloud 6.0.0.13155-daily20240119`

Login dialog re-appears if wrong password is entered

Issue #11373, opened 2023-11-16
[5] Client crash in OCC::Folder::slotWatchedPathsChanged at `src/folder.cpp:687` (during gui test)


saw-jan:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the bug

CI build: https://drone.owncloud.com/owncloud/client/17039

```c
Core was generated by `/drone/src/build/bin/owncloud -s --logfile - --logdebug --logflush'.
Program terminated with signal SIGABRT, Aborted.
#0  0x00007f27fed56844 in __pthread_kill_implementation () from /lib64/libc.so.6
[Current thread is 1 (Thread 0x7f27fb673f00 (LWP 1950))]
#0  0x00007f27fed56844 in __pthread_kill_implementation () at /lib64/libc.so.6
#1  0x00007f27fed05abe in raise () at /lib64/libc.so.6
#2  0x00007f27fecee90f in abort () at /lib64/libc.so.6
#3  0x00007f27ff2d20eb in qgetenv(char const*) [clone .cold] () at /lib64/libQt6Core.so.6
#4  0x00007f27ff2d2b7f in  () at /lib64/libQt6Core.so.6
#5  0x00007f27ff2d2083 in qt_assert_x(char const*, char const*, char const*, int) () at /lib64/libQt6Core.so.6
#6  0x00000000004e6391 in OCC::Folder::slotWatchedPathsChanged(QSet<QString> const&, OCC::Folder::ChangeReason)::$_0::operator()() const::{lambda(OCC::CheckSums::Algorithm, QByteArray const&)#1}::operator()(OCC::CheckSums::Algorithm, QByteArray const&) const (this=0x250ee10, checksum=...) at /drone/src/src/gui/folder.cpp:687
```

### Screenshots

![Screenshot from 2023-11-16 10-46-22](https://github.com/owncloud/client/assets/52366632/04269534-fb44-40b4-9a68-da07247405d3)

### Logs

[stacktrace.log](https://github.com/owncloud/client/files/13372469/stacktrace.2.log)
[client.log](https://github.com/owncloud/client/files/13372482/serverlog.2.log)

### Client version number

ownCloud 6.0.0-git afd26a

### Server information

10.13.2 - channel: stable

[...]

PR #11485, opened 2024-01-18
Fix: #11373

TheOneRing:
Remove assert that proofed that we the assumptions were correct. Under certain conditions, however, the file can be modified again before we computed the checksum causing a false positive.


fmoc:
I'm not sure I understand the description correct. Do you mean there were false-negative evaluation of this assert due to changes in the background? Maybe we should just log these cases but not remove the test entirely?

TheOneRing:
The code was only active in dev setups. So what's the point of keeping it?

TheOneRing:
Se my explanation in the issue.

fmoc:
I see.

fmoc (approved):

PR #11491, opened 2024-01-19
Rewrite Folder Status in qml

TheOneRing:
image
image


src/gui/folderstatusmodel.cpp
```diff
@@ -269,131 +238,112 @@ QVariant FolderStatusModel::data(const QModelIndex &index, int role) const
...
+    }
+    case Columns::DisplayName: {
+        if (auto *space = getSpace()) {
+            return space->displayName();
+        }
+        return f->displayName();
+    }
+    case Columns::FolderImageUrl:
+        if (f->accountState()->supportsSpaces()) {
+            // TODO: the url hast random parts to enforce a reload
```
TheOneRing, line 288: Using the random number is a hack.
In theory, we could directly use a http url.
However, we would need to provide a new networkaccess manager as the qml context runs in a different thread.
Our current nam, used for authentication and logging, is currently not designed in a way that really supports multiple instances.
https://github.com/owncloud/client/blob/1eeef49749f78eb3ec4621e4869166b720b12c29/src/libsync/creds/httpcredentials.cpp#L134-L134

erikjv (approved):

PR #11494, opened 2024-01-23
Set minimal OS version

erikjv:


src/gui/MacOSXBundleInfo.plist
```diff
@@ -34,6 +34,8 @@
         <string>F9ZGRXJE7XbQb2Kt36hwaBO4rYjkXYHiNS5hd+MkyKY=</string>
         <key>SUAllowsAutomaticUpdates</key>
         <string>NO</string>
+        <key>LSMinimumSystemVersion</key>
+        <string>${MACOSX_DEPLOYMENT_TARGET}</string>
```
fmoc, line 38: Any reason not to use `CMAKE_OSX_DEPLOYMENT_TARGET` directly?

erikjv: No, but I want to first look at the build artifacts and see if this works

src/gui/CMakeLists.txt
```diff
@@ -184,8 +184,10 @@ else()
       PROPERTIES
       MACOSX_PACKAGE_LOCATION Resources
       )
-
-  set_target_properties(owncloud PROPERTIES MACOSX_BUNDLE_INFO_PLIST ${CMAKE_CURRENT_SOURCE_DIR}/MacOSXBundleInfo.plist)
+    if (CMAKE_OSX_DEPLOYMENT_TARGET)
+        set(MACOSX_DEPLOYMENT_TARGET ${CMAKE_OSX_DEPLOYMENT_TARGET})
+    endif()
```
TheOneRing, line 189: I think the property is no longer used?

TheOneRing (changes requested):

PR #11510, opened 2024-02-07
Clean up Windows updater code

TheOneRing:


src/gui/updater/ocupdater.cpp
```diff
@@ -558,19 +507,47 @@ bool NSISUpdater::handleStartup()
                 showUpdateErrorDialog(settings.value(updateTargetVersionStringC).toString());
                 return false;
             }
-        } else {
-            qCInfo(lcUpdater) << "Triggering an update";
```
TheOneRing, line 562: If I'm correct the condition in OCUpdater::performUpdate() was the same so performUpdate was never executed?

erikjv (approved):

PR #11511, opened 2024-02-07
Always log the available QetworkInformation backends

erikjv:
This will also log the available backends when a warning occurs that no backend got loaded.


TheOneRing (changes requested):
I think we have the same code in a second location. Please check.

https://github.com/owncloud/client/blob/859cd599f6a0e10569d0e4be7dea18079aacf09d/src/gui/application.cpp#L127-L127

src/gui/main.cpp
```diff
@@ -452,12 +452,12 @@ int main(int argc, char **argv)
     setupLogging(options);
 
 #if QT_VERSION >= QT_VERSION_CHECK(6, 3, 0)
-    qCDebug(lcMain) << QNetworkInformation::availableBackends().join(QStringLiteral(", "));
     if (!QNetworkInformation::loadDefaultBackend()) {
-        qCWarning(lcMain) << "Failed to load QNetworkInformation";
+        qCWarning(lcMain) << "Failed to load QNetworkInformation. Available backends:" << QNetworkInformation::availableBackends().join(QStringLiteral(", "));
```
TheOneRing, line 456: We should also implement the fallback from https://github.com/qt/qtbase/commit/ce4531a4901504f62f8c76c3542d3b1aad2b384a#diff-db55ac8df386b2d509573b97fbdc986e20426e4b9bcf0fbc7e03a866406b1631R685 ?
In my tests this will load the glib plugin.

TheOneRing: The default network manager relies on the availability of a dbus service. Which obviously is not available in my wsl setup.

src/gui/main.cpp
```diff
@@ -248,6 +248,29 @@ void setupLogging(const CommandLineOptions &options)
     qCInfo(lcMain) << "Arguments:" << qApp->arguments();
 }
 
+void loadQNetworkInformationBackend()
+{
+#if QT_VERSION >= QT_VERSION_CHECK(6, 3, 0)
+    if (!QNetworkInformation::loadDefaultBackend()) {
+        qCWarning(lcMain) << "Failed to load default backend of QNetworkInformation.";
+        if (!QNetworkInformation::loadBackendByFeatures(QNetworkInformation::Feature::Metered)) {
+            qCWarning(lcMain) << "Failed to load backend of QNetworkInformation by metered feature.";
```
TheOneRing, line 257: If that fails too, I guess we should fall back to at least load a backend that support connectivity changes?

TheOneRing (changes requested):
^

TheOneRing (approved):

TheOneRing (approved):

PR #11530, opened 2024-02-22
[tests-only] [full-ci] Add test coverage to copy and paste virtual file in same location

Salipa-Gurung:
Added step to copy and paste a virtual file in the same location by creating its copy with ` - Copy` suffix in file name.

| Copy a virtual file and Paste it at the same location | File browser asks or creates a copy |
---|---

Related Issue: https://github.com/owncloud/client/issues/11346

### DIFF
--- a/test/gui/shared/steps/file_context.py
+++ b/test/gui/shared/steps/file_context.py
@@ -96,6 +96,14 @@ def extractZip(zip_file_path, destination_dir):
         zipFile.extractall(destination_dir)
 
 
+def addCopySuffix(resource_path, resource_type):
+    if resource_type == "file":
+        source_dir = resource_path.rsplit('.', 1)
+        return source_dir[0] + " - Copy." + source_dir[-1]
+    else:
+        return resource_path + " - Copy"
+
+
 @When(
     'user "|any|" creates a file "|any|" with the following content inside the sync folder'
 )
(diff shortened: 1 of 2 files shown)


test/gui/shared/steps/file_context.py
```diff
@@ -123,12 +128,16 @@ def step(context, username, filename, filesize):
...
     if resource_type == 'folder':
-        return shutil.copytree(source_dir, destination_dir)
+        return shutil.copytree(source_dir_path, destination_dir_path)
     else:
-        return shutil.copy2(source_dir, destination_dir)
+        try:
+            return shutil.copy2(source_dir_path, destination_dir_path)
+        except shutil.SameFileError:
+            copy_of_source_dir = addCopySuffix(source_dir_path)
+            return shutil.copy2(source_dir_path, copy_of_source_dir)
```
saw-jan, line 140: I have another thought now. how about this flow:
- check if `source` == `destination`
- if same
  -  append destination with necessary suffix
  - do the copy
- if not
  - do normal copy

can be done for both file and folder. Then we don't have to use try-except

test/gui/tst_vfs/test.feature
```diff
@@ -38,19 +38,25 @@ Feature: Enable/disable virtual file support
         Given user "Alice" has been created on the server with default attributes and without skeleton files
         And user "Alice" has uploaded file with content "sample file" to "sampleFile.txt" in the server
         And user "Alice" has uploaded file with content "lorem file" to "lorem.txt" in the server
+        And user "Alice" has uploaded file with content "test file" to "testFile.txt" in the server
```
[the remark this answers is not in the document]
Salipa-Gurung: This test was supposed to be on a virtual file so, a new test file was added because the previous two files were downloaded during the copy-paste operation.

test/gui/shared/steps/file_context.py
```diff
@@ -123,12 +131,14 @@ def step(context, username, filename, filesize):
 
 @When(r'the user copies the (file|folder) "([^"]*)" to "([^"]*)"', regexp=True)
 def step(context, resource_type, source_dir, destination_dir):
-    source_dir = getResourcePath(source_dir)
-    destination_dir = getResourcePath(destination_dir)
+    source_dir_path = getResourcePath(source_dir)
+    destination_dir_path = getResourcePath(destination_dir)
```
saw-jan, line 135:
```suggestion
    source = getResourcePath(source_dir)
    destination = getResourcePath(destination_dir)
```

test/gui/shared/steps/file_context.py
```diff
@@ -123,12 +131,14 @@ def step(context, username, filename, filesize):
...
+    if source_dir_path == destination_dir_path or destination_dir == '/':
```
saw-jan, line 136:
```suggestion
    if source_dir_path == destination_dir_path and destination_dir != '/':
```

saw-jan (approved):
:+1:

Issue #11516, opened 2024-02-13
Bug: getDirectorySyncRootMarking prevents adding spaces to sync folder


TheOneRing:
The code recursively checks its parents and accidentally prevents us from adding a new space using the "Add space" button.

https://github.com/owncloud/client/blob/2e5c31640a490540264d6d3536fe95b942da140e/src/gui/folderman.cpp#L723-L723

saw-jan:
Tested on `ownCloud 6.0.0.13582-daily20240327 [4258b6]`

New Space can be added
![Screenshot from 2024-03-28 16-16-02](https://github.com/owncloud/client/assets/52366632/b9d5fb7a-00a2-4efc-a05f-78557b2d4bc9)

PR #11532, opened 2024-02-23
Fix check for creation of a spaces folder in a spaces root

erikjv:
Fixes: #11516

### DIFF
--- a/src/gui/folderman.cpp
+++ b/src/gui/folderman.cpp
@@ -720,11 +720,29 @@ static QString canonicalPath(const QString &path)
     return selFile.canonicalFilePath();
 }
 
-static QString checkPathForSyncRootMarkingRecursive(const QString &path)
+static QString checkPathForSyncRootMarkingRecursive(const QString &path, FolderMan::NewFolderType folderType, const QUuid &accountUuid)
 {
-    auto existingTag = Utility::getDirectorySyncRootMarking(path);
-    if (!existingTag.isEmpty()) {
-        return FolderMan::tr("Folder '%1' is already in use by application %2!").arg(path, existingTag);
+    std::pair<QString, QUuid> existingTags = Utility::getDirectorySyncRootMarkings(path);
+    if (!existingTags.first.isEmpty()) {
+        if (existingTags.first != Theme::instance()->orgDomainName()) {
+            // another application uses this as spaces root folder
+            return FolderMan::tr("Folder '%1' is already in use by application %2!").arg(path, existingTags.first);
+        }
+
+        // Looks good, it's our app, let's check the account tag:
+        switch (folderType) {
+        case FolderMan::NewFolderType::SpacesFolder:
+            if (existingTags.second == accountUuid) {
+                // Nice, that's what we like, the sync root for our account in our app. No error.
+                return {};
+            }
+            [[fallthrough]];
+        case FolderMan::NewFolderType::OC10SyncRoot:
+            [[fallthrough]];
+        case FolderMan::NewFolderType::SpacesSyncRoot:
+            // It's our application but we don't want to create a spaces folder, so it must be another space root
+            return FolderMan::tr("Folder '%1' is already in use by another account.").arg(path);
+        }
     }
 
     QString parent = QFileInfo(path).path();
@@ -732,10 +750,10 @@ static QString checkPathForSyncRootMarkingRecursive(const QString &path)
         return {};
     }
 
-    return checkPathForSyncRootMarkingRecursive(parent);
+    return checkPathForSyncRootMarkingRecursive(parent, folderType, accountUuid);
 }
 
-QString FolderMan::checkPathValidityRecursive(const QString &path)
+QString FolderMan::checkPathValidityRecursive(const QString &path, FolderMan::NewFolderType folderType, const QUuid &accountUuid)
 {
     if (path.isEmpty()) {
         return FolderMan::tr("No valid folder selected!");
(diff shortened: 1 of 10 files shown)


src/gui/folderman.cpp
```diff
@@ -720,22 +720,41 @@ static QString canonicalPath(const QString &path)
...
+                // this is handled below: another application uses this as spces root folder
+            }
+        }
+
+        if (existingTags.first == Theme::instance()->orgDomainName()) {
+            // it's our application, so must be another space root
+            return FolderMan::tr("Folder '%1' is already in use by another account.").arg(path);
+        }
+        // another application uses this as spaces root folder
+        return FolderMan::tr("Folder '%1' is already in use by application %2!").arg(path, existingTags.first);
```
TheOneRing, line 746: I think it could make things easier to parse if we handle this scenario first? It would make the two checks for domain superfluous.

src/gui/folderman.cpp
```diff
@@ -825,8 +858,9 @@ QString FolderMan::findGoodPathForNewSyncFolder(const QString &basePath, const Q
     {
         QString folder = normalisedPath;
         for (int attempt = 2; attempt <= 100; ++attempt) {
+            // HvR: is this correct?
```
TheOneRing, line 861: What do you mean?

test/testfolderman.cpp
```diff
@@ -130,32 +132,32 @@ private slots:
...
         { // check for rejection of a directory with `._sync_*.db`
             QVERIFY(dir2.mkpath(QStringLiteral("db-check2")));
-            QVERIFY(folderman->checkPathValidityForNewFolder(dirPath + QStringLiteral("/db-check2")).isNull());
+            QVERIFY(folderman->checkPathValidityForNewFolder(dirPath + QStringLiteral("/db-check2"), type, uuid).isNull());
             QFile f(dirPath + QStringLiteral("/db-check2/._sync_something.db"));
             QVERIFY(f.open(QFile::Truncate | QFile::WriteOnly));
             f.close();
             QVERIFY(QFileInfo::exists(dirPath + QStringLiteral("/db-check2/._sync_something.db")));
-            QVERIFY(!folderman->checkPathValidityForNewFolder(dirPath + QStringLiteral("/db-check2")).isNull());
+            QVERIFY(!folderman->checkPathValidityForNewFolder(dirPath + QStringLiteral("/db-check2"), type, uuid).isNull());
```
TheOneRing, line 160: Please add a test case for the scenario we are trying to solve here.

src/gui/folderman.cpp
```diff
@@ -720,22 +720,41 @@ static QString canonicalPath(const QString &path)
...
 }
 
-static QString checkPathForSyncRootMarkingRecursive(const QString &path)
-{
-    auto existingTag = Utility::getDirectorySyncRootMarking(path);
-    if (!existingTag.isEmpty()) {
-        return FolderMan::tr("Folder '%1' is already in use by application %2!").arg(path, existingTag);
+static QString checkPathForSyncRootMarkingRecursive(const QString &path, FolderMan::NewFolderType folderType, const QUuid &accountUuid)
+{
+    std::pair<QString, QUuid> existingTags = Utility::getDirectorySyncRootMarkings(path);
```
TheOneRing, line 725: Why do we need to differentiate for FolderMan::NewFolderType?

TheOneRing (changes requested):

src/gui/folderman.cpp
erikjv: done

src/gui/folderman.cpp
erikjv: Is `FolderMan::findGoodPathForNewSyncFolder` only used for OC10 sync roots or Spaces sync roots? So never for Spaces folders?

src/gui/folderman.cpp
erikjv: We need to know if it's a Spaces folder we're checking for: that one is allowed (preferred!) to end up in a folder that's tagged as a sync root for Spaces.

src/gui/folderman.cpp
TheOneRing: Its used in both cases.

erikjv: done

test/testfolderman.cpp
erikjv: done

src/gui/folderman.cpp
erikjv: done

src/gui/folderman.cpp
```diff
@@ -720,22 +720,42 @@ static QString canonicalPath(const QString &path)
...
+            } else {
+                return FolderMan::tr("Folder '%1' is already in use by another account.").arg(path);
+            }
+
+        case FolderMan::NewFolderType::OC10SyncRoot:
+            [[fallthrough]];
+        case FolderMan::NewFolderType::SpacesSyncRoot:
+            // It's our application but we don't want to create a spaces folder, so it must be another space root
+            return FolderMan::tr("Folder '%1' is already in use by another account.").arg(path);
+        }
```
TheOneRing, line 747: Don't all 3 cases behave the same if the `id != account id`?

With ownCloud 10, the folder would also contain a .syncdb therefor it would be detected as claimed.
WIth ocis, we can only have one sync root, so we can't end up in here a second time.

Anyhow I guess we can combine the error message?

src/gui/folderman.cpp
```diff
@@ -825,8 +859,8 @@ QString FolderMan::findGoodPathForNewSyncFolder(const QString &basePath, const Q
     {
         QString folder = normalisedPath;
         for (int attempt = 2; attempt <= 100; ++attempt) {
-            if (!QFileInfo::exists(folder)
-                && FolderMan::instance()->checkPathValidityForNewFolder(folder).isEmpty()) {
+            // HvR: is this correct?
```
TheOneRing, line 862: I think that question was resolved?

src/gui/folderwizard/folderwizardlocalpath.cpp
```diff
@@ -63,7 +63,10 @@ QString FolderWizardLocalPath::localPath() const
 
 bool FolderWizardLocalPath::isComplete() const
 {
-    QString errorStr = FolderMan::instance()->checkPathValidityForNewFolder(localPath());
+    // HvR: is this correct?
```
TheOneRing, line 66: looks ok

test/testfolderman.cpp
```diff
@@ -183,28 +185,73 @@ private slots:
...
+        QCOMPARE(folderman->findGoodPathForNewSyncFolder(dirPath, QStringLiteral("../../../Bo/b"), folderType), QString(dirPath + QStringLiteral("/___Bo_b")));
 
         // normalise the name
-        QCOMPARE(
-            folderman->findGoodPathForNewSyncFolder(dirPath, QStringLiteral("            Bo:*<>!b          ")), QString(dirPath + QStringLiteral("/Bo____!b")));
+        QCOMPARE(folderman->findGoodPathForNewSyncFolder(dirPath, QStringLiteral("            Bo:*<>!b          "), folderType),
+            QString(dirPath + QStringLiteral("/Bo____!b")));
+    }
+
+    void testSpacesSyncRootAndFolderCreation()
```
TheOneRing, line 215: 😍

TheOneRing (changes requested):

src/gui/folderman.cpp
erikjv: done

src/gui/folderman.cpp
erikjv: yes

src/gui/folderwizard/folderwizardlocalpath.cpp
erikjv: ✅

TheOneRing (approved):
looks good

Issue #11531, opened 2024-02-23
Select 'No' button by default in the unknown certificate dialog, and show the dialog above the main window


tmoerschell:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Feature description

Two small UI/UX improvements, reducing the number of keystrokes you need to dismiss certificate errors.

Currently, when connecting to a public network with a sing-in portal, the OwnCloud client will show a certificate error, since the requests are not answered by the correct server, and thus not with the correct certificate. 

Now, the button selected by default is 'Yes'. This is the less likely answer, as one should not go about trusting every certificate he comes across. 

Furthermore, at least when the client is hidden and the error dialog pops up, the main window is shown as well and covers the pop-up. The main window is unresponsive since the dialog is modal. As a result, one first needs to figure out that there is a dialog open in the background before switching to it and closing it. 

### Proposed solution

1. Make the button selected by default be the refusal button. 
2. Make the dialog transient, i.e. make sure it is always shown on top of the main window.

### Alternative solutions you considered

No real alternative to 1., but a further improvement to encounter this situation less often would be to detect the answer as a sign-on portal and pause syncing while the server can't be reached. 

For 2., if there is a way to only show the dialog, without the main window, this would be good as well. Or, instead of directly opening a dialog, there could be a notification, which is more easily dismissed, when the client is hidden. 

### Additional context

Issue encountered on KDE Plasma. 
Source of the dialog: https://github.com/owncloud/client/blob/2db0920436d666e3d2bbbc786e6692516327bbc4/src/gui/tlserrordialog.ui

I have little experience with QT. Someone else will probably be much quicker than myself to implement this. If no one wants to do this within a week or something, I will try to do it myself.

TheOneRing:
We will take care that the No button is the default action.
For the window behavior that probably an issue with the window manager.

Could you please report your version?
Settings->About->Versions and paste it here?
Should look something like:
```
testpilotcloud 6.0.0.13196-daily20240125 ab204c
Libraries Qt 6.6.1, OpenSSL 3.2.0 23 Nov 2023
Using virtual files plugin: wincfapi
OS: windows-10.0.22631 (build arch: x86_64, CPU arch: x86_64)
QPA: windows
```

[...]

PR #11540, opened 2024-03-11
Make 'No' button default in Unknown Certificate dialog

erikjv:
Fixes: #11531

### DIFF
--- a/src/gui/tlserrordialog.cpp
+++ b/src/gui/tlserrordialog.cpp
@@ -40,6 +40,11 @@ TlsErrorDialog::TlsErrorDialog(const QList<QSslError> &sslErrors, const QString 
 
     connect(_ui->buttonBox, &QDialogButtonBox::accepted, this, &TlsErrorDialog::accept);
     connect(_ui->buttonBox, &QDialogButtonBox::rejected, this, &TlsErrorDialog::reject);
+
+    // Try to set the No-button (which will also emit the reject signal when cliced) as the default
+    if (QPushButton *noButton = _ui->buttonBox->button(QDialogButtonBox::No)) {
+        noButton->setDefault(true);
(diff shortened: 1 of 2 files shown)


TheOneRing:
The ui tests have multiple crashes in the tls dialog, please have a look.

TheOneRing (changes requested):
^

src/gui/tlserrordialog.cpp
```diff
@@ -40,6 +40,12 @@ TlsErrorDialog::TlsErrorDialog(const QList<QSslError> &sslErrors, const QString
 
     connect(_ui->buttonBox, &QDialogButtonBox::accepted, this, &TlsErrorDialog::accept);
     connect(_ui->buttonBox, &QDialogButtonBox::rejected, this, &TlsErrorDialog::reject);
+
+    // Try to set the No-button (the reject button) as the default
+    if (QPushButton *noButton = _ui->buttonBox->button(QDialogButtonBox::No)) {
+        Q_ASSERT(_ui->buttonBox->buttonRole(noButton) == QDialogButtonBox::RejectRole || _ui->buttonBox->buttonRole(noButton) == QDialogButtonBox::NoRole);
```
TheOneRing, line 46: Hmm according to https://doc.qt.io/qt-6/qdialogbuttonbox.html#ButtonRole-enum it is supposed to always be `NoRole`?

erikjv: But how can clicking the "No" button then reject the dialog?

erikjv: I removed the assert, it's superfluous to check the `NoRole` on the "No" button.

TheOneRing (approved):

Issue #11486, opened 2024-01-18
Include update information in about dialog (Move from settings to about).


rpausch-owncloud:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Feature description

Currently only the current version is displayed in the about dialog. User has no information if this is the latest or not and if they have to update or not. In order to get this information user has to scroll down in the client GUI where the information is displayed as well as the update button if applicable. For dummy users like me this is very uncommon and therfore they won't find that update button.

### Proposed solution

In additon to the version number display if this is the latest. If not display an update button or update automatically if automatic updates are enabled. See examples from Firefox and Chrome.
![297787203-667ee75e-2148-402d-95c3-bb247a432cb9](https://github.com/owncloud/client/assets/144467728/cdc79cb7-d00b-43ab-8bff-da4c2d404cfa)

![297787505-666bc42f-8b5b-4da7-9eef-7af459dc2c12](https://github.com/owncloud/client/assets/144467728/6dbdd58d-735f-41a1-8d81-1ffc347534e6)

![297787746-5293bbf0-1b2e-4aa4-bf9e-290b0c3d4dd9](https://github.com/owncloud/client/assets/144467728/8846fd88-8a84-453d-a80d-75c18877a540)

TheOneRing:
We display that information on all platforms but mac.
![image](https://github.com/owncloud/client/assets/200626/b86ff749-87e3-4ec2-b292-fdeb73e2c10d)

On Mac, we don't display it at the moment as it's handled by a third party software.

rpausch-owncloud:
I do not want it here but in the about dialig
![grafik](https://github.com/owncloud/client/assets/144467728/c6a789a7-ca55-431d-aa71-d6f65537f600)

, since as I wrote, this information is usually hidden because you have to scroll down in the UI in order to see it. Users are used to se this information in the about screen, e.g. in Firefox, Chrome, Thunderbird, Rocket.chat
![grafik](https://github.com/owncloud/client/assets/144467728/68d8e363-2c19-40f0-b7b1-89065541af4a)

saw-jan:
Tested with `ownCloud 6.0.0.14185-daily20240710 [a947c5]`

[...]

PR #11541, opened 2024-03-11
Add update information to the about dialog

erikjv:
Fixes: #11486

### DIFF
--- a/src/gui/aboutdialog.cpp
+++ b/src/gui/aboutdialog.cpp
@@ -14,8 +14,14 @@
 #include "aboutdialog.h"
 #include "ui_aboutdialog.h"
 
-#include "theme.h"
-#include "guiutility.h"
+#include "common/restartmanager.h"
+#include "gui/guiutility.h"
+#include "libsync/configfile.h"
+#include "libsync/theme.h"
+
+#ifdef WITH_AUTO_UPDATER
+#include "updater/ocupdater.h"
+#endif
 
 namespace OCC {
 
@@ -30,6 +36,8 @@ AboutDialog::AboutDialog(QWidget *parent)
 
     connect(ui->versionInfo, &QTextBrowser::anchorClicked, this, &AboutDialog::openBrowserFromUrl);
     connect(ui->aboutText, &QLabel::linkActivated, this, &AboutDialog::openBrowser);
+
+    setupUpdaterWidget();
 }
 
 AboutDialog::~AboutDialog()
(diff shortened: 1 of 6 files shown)


erikjv:
I'm not sure if we should move everything from the settings dialog to the about dialog. I think we should move the update part in the settings up higher, because it's probably more used than bandwidth limits. 

For the about dialog: the text is customisable, so I added a label about the update status. We should probably also add a button or link text to redirect to the settings when there is an update available.

TheOneRing:
Hmm the channel selection is only available in testpilotcloud so maybe we can move it to the about dialog?

erikjv:
> Hmm the channel selection is only available in testpilotcloud so maybe we can move it to the about dialog?

We can. I think it would give theming/design problems if we put it in the "about" tab, so maybe make a new (second) tab called "Updates" and put the contents of the QGroupBox in there? Also, I introduced that label at the bottom to draw more attention to updates, should that label stay? (See screenshot below.)

Screenshot 2024-03-18 at 14 03 41

TheOneRing:
I think only the text can be branded, so it should be possible to add the info into the main widget.

erikjv:
In the latest commit I moved all of the update stuff from the settings dialog to the about dialog. @TheOneRing How can I test this?

erikjv:
Screenshot 2024-04-05 at 16 51 05

I'm not sure if I can get the version number out. I think it'd be useful to show that too.

erikjv:
@TheOneRing  The `OCUpdater` could provide the version. Should I add that in a separate PR?

src/gui/aboutdialog.cpp
```diff
@@ -14,8 +14,14 @@
 #include "aboutdialog.h"
 #include "ui_aboutdialog.h"
 
-#include "theme.h"
+#include "common/restartmanager.h"
+#include "configfile.h"
```
TheOneRing, line 18: could you use the full paths please?

src/gui/aboutdialog.h
```diff
@@ -34,6 +34,10 @@ class AboutDialog : public QDialog
 private:
     void openBrowser(const QString &s);
     void openBrowserFromUrl(const QUrl &s);
+    void setupUpdaterWidget();
+
+private slots:
```
TheOneRing, line 39:
```suggestion
private Q_SLOTS:
```

TheOneRing (changes requested):

src/gui/aboutdialog.cpp
erikjv: done

src/gui/aboutdialog.h
erikjv: done

erikjv:
> I'm not sure if I can get the version number out. I think it'd be useful to show that too.

So that was during testing. Meaning it passed the testing.

PR #11545, opened 2024-03-15
Update RemoveAllFiles dialog

TheOneRing:
Goal of this change is to prepare the removal of addModalLegacyDialog

Please also have a look on https://github.com/owncloud/client/pull/11564 where a qml is used instead

image

Old:
image


TheOneRing:
@tbsbdr for the design I tried to use the available space better.

tbsbdr:
I'd remove the dialog at all.
if all files got deleted from the server, the default should be that all files get also deleted from the client. the server is the source of truth; files can be restored from the server trashbin if this was done unintentionally.

Issue #11533, opened 2024-02-23
Detect captive portal and pause sync


TheOneRing:
See https://github.com/owncloud/client/issues/11531
https://doc.qt.io/qt-6/qnetworkinformation.html#isBehindCaptivePortal-prop

erikjv:
Where should we do this exactly? In the account state? Or when the SSL cert. checks fail? Also, should we listen for this and somehow show a notification on the desktop that we stopped/resumed syncing?

PR #11551, opened 2024-03-18
Pause sync when behind a captive portal

erikjv:
Fixes: #11533

### DIFF
--- a/src/gui/guiutility.cpp
+++ b/src/gui/guiutility.cpp
@@ -104,6 +104,15 @@ bool Utility::internetConnectionIsMetered()
     return false;
 }
 
+bool OCC::Utility::internetThroughCaptivePortal()
+{
+    if (auto *qNetInfo = QNetworkInformation::instance()) {
+        return qNetInfo->isBehindCaptivePortal();
+    }
+
+    return false;
+}
+
 void Utility::markDirectoryAsSyncRoot(const QString &path)
 {
     Q_ASSERT(getDirectorySyncRootMarking(path).isEmpty());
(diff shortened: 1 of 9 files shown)


src/gui/guiutility.cpp
```diff
@@ -104,6 +104,15 @@ bool Utility::internetConnectionIsMetered()
     return false;
 }
 
+bool OCC::Utility::internetThroughCaptivePortal()
```
TheOneRing, line 107: I'm wondering whether the null pointer check justifies the overhead of moving the code to utils.
In the end it hides the use of QNetworkInformation and can make reading the code in the future more complicated (also the utils class is already quire crowded)
This was also done for `internetConnectionIsMetered`.

src/gui/networksettings.cpp
```diff
@@ -195,6 +196,18 @@ void NetworkSettings::loadMeteredSettings()
...
+void NetworkSettings::loadCaptivePortalSettings()
+{
+    if (QNetworkInformation *qNetInfo = QNetworkInformation::instance()) {
+        if (qNetInfo->supports(QNetworkInformation::Feature::CaptivePortal)) {
+            _ui->pauseSyncWhenBehindCaptivePortalCheckBox->setChecked(ConfigFile().pauseSyncWhenBehindCaptivePortal());
+            return;
+        }
+    }
+
+    _ui->pauseSyncWhenBehindCaptivePortalCheckBox->setVisible(false);
```
TheOneRing, line 208: We need to hide the tls dialog if enabled, I'd also say we need to pause, as util the captive portal is accepted there is no way to sync.

erikjv:
This is the wrong solution

PR #11563, opened 2024-03-26
[gui-tests] fix GUI tests for Windows

saw-jan:


test/gui/shared/steps/account_context.py
```diff
@@ -20,7 +20,9 @@ def step(context, accountType):
     AccountConnectionWizard.addAccount(account_details)
     space = ""
     if get_config("ocis"):
-        space = "Personal"
+        # listen for Shares space to finish sync
+        # Personal space doesn't include all sync messages
+        space = "Shares"
```
TheOneRing, line 25: The shares space is very special, could we instead create a space and check with that one?

saw-jan: I will try to find out what happens with Personal spaces both in linux and windows, and try to use it

TheOneRing: As you mentioned in the comment in the code the personal space is also special :D

saw-jan: hmm,
here, we are just trying to wait until the sync is done. We used Personal space for that. Would you suggest other way around? But creating/using new space would be out of the test scope there.

saw-jan: >hmm could you then elaborate on Personal space doesn't include all sync messages

I saw the problem once in Windows and thought there might be some differences but that actually was not the case. Could be a timing issue which seems to be solved now. Means the previous code was/is working fine

Salipa-Gurung (approved):

PR #11599, opened 2024-04-10
Show version when an update is available

erikjv:


src/gui/updater/updater.cpp
```diff
@@ -45,7 +45,7 @@ Updater *Updater::instance()
 
 QUrl Updater::updateUrl()
 {
-    QUrl updateBaseUrl(QString::fromLocal8Bit(qgetenv("OCC_UPDATE_URL")));
```
TheOneRing, line 48: Nope

TheOneRing (changes requested):
^

src/gui/updater/updater.cpp
erikjv: removed

Issue #11597, opened 2024-04-10
[QA] update to Qt6.5 for GUI tests


saw-jan:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the QA issue

The minimum required Qt version for the client is `6.5`. see https://github.com/owncloud/client/pull/11575

Open tasks:
- [x] update `qt6` and [extra-cmake-modules 6](https://download.kde.org/stable/frameworks/6.0/extra-cmake-modules-6.0.0.tar.xz) in [owncloudci/client](https://github.com/owncloud-ci/client) image (https://github.com/owncloud-ci/client/pull/29)
- [x] update `squish IDE` in [owncloudci/squish](https://github.com/owncloud-ci/squish) image (https://github.com/owncloud-ci/squish/pull/57)
- [x] update images in drone config (https://github.com/owncloud/client/pull/11606)

PR #11606, opened 2024-04-15
[full-ci][gui-tests] update docker images

saw-jan:
Updated docker images:
`owncloudci/client:fedora-38-amd64` -> `owncloudci/client:fedora-39-amd64`
`owncloudci/squish:fedora-38-7.1.1-qt64x-linux64` -> `owncloudci/squish:fedora-39-7.2.1-qt66x-linux64`

Fixes https://github.com/owncloud/client/issues/11597

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -24,8 +24,8 @@ OC_UBUNTU = "owncloud/ubuntu:20.04"
 # Eventually, we have to use image built on ubuntu
 # Todo: update or remove the following images
 # https://github.com/owncloud/client/issues/10070
-OC_CI_CLIENT_FEDORA = "owncloudci/client:fedora-38-amd64"
-OC_CI_SQUISH = "owncloudci/squish:fedora-38-7.1.1-qt64x-linux64"
+OC_CI_CLIENT_FEDORA = "sawjan/client:39-qt6.6"
+OC_CI_SQUISH = "sawjan/squish:craft"
 
 PLUGINS_GIT_ACTION = "plugins/git-action:1"
 PLUGINS_S3 = "plugins/s3"
@@ -188,7 +188,7 @@ def gui_test_pipeline(ctx):
             squish_parameters += " --tags %s" % params["tags"]
 
         steps = skipIfUnchanged(ctx, "gui-tests") + \
-                build_client(OC_CI_CLIENT_FEDORA, False)
+                build_client(OC_CI_SQUISH, False)
 
         services = testMiddlewareService(server)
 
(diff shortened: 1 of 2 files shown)


.drone.star
```diff
@@ -235,6 +235,10 @@ def build_client(image = OC_CI_CLIENT, ctest = True):
     cmake_options = '-G"%s" -DCMAKE_C_COMPILER="%s" -DCMAKE_CXX_COMPILER="%s" -DCMAKE_BUILD_TYPE="%s"'
```
TheOneRing, line 235: Can you use /usr/clang /usr/clang++ (else it uses the clang provided by craft)

saw-jan: Now I only used `cmake -G"Ninja" -DBUILD_TESTING=OFF -S ..` while building client using craft installed libraries

saw-jan:
Most of the tests passed. remaining failures are due to object error
https://drone.owncloud.com/owncloud/client/18154/5/11

TheOneRing:
> Most of the tests passed. remaining failures are due to object error https://drone.owncloud.com/owncloud/client/18154/5/11

We now use qml for the spaces browser here too, if a object name or property is missing in the code please tell me.

saw-jan:
> > Most of the tests passed. remaining failures are due to object error https://drone.owncloud.com/owncloud/client/18154/5/11
> 
> We now use qml for the spaces browser here too, if a object name or property is missing in the code please tell me.

okey, the failure was due to the change in the selector type

TheOneRing:
Please update https://github.com/owncloud/client/blob/5b5b25557840bc273fffa38613d8f7937b484091/.drone.star#L72-L76
to use /usr/bin/clang else it uses the clang from the craft environment

.drone.star
```diff
@@ -232,8 +232,11 @@ def gui_test_pipeline(ctx):
     return pipelines
 
 def build_client(image = OC_CI_CLIENT, ctest = True):
-    cmake_options = '-G"%s" -DCMAKE_C_COMPILER="%s" -DCMAKE_CXX_COMPILER="%s" -DCMAKE_BUILD_TYPE="%s"'
-    cmake_options = cmake_options % (build_config["generator"], build_config["c_compiler"], build_config["cxx_compiler"], build_config["build_type"])
+    cmake_options = '-G"%s"' % build_config["generator"]
+
+    if image != OC_CI_SQUISH:
```
TheOneRing, line 237: Why the change here?

saw-jan: For [ctest pipeline](https://drone.owncloud.com/owncloud/client/18160/4/2), I am using `owncloudci/client:fedora-39-amd64` which doesn't have to craft environment. so maintaining the previous build configs if not using `OC_CI_SQUISH` which has craft environment

saw-jan:
> Please update
> 
> https://github.com/owncloud/client/blob/5b5b25557840bc273fffa38613d8f7937b484091/.drone.star#L72-L76
> 
> 
> to use /usr/bin/clang else it uses the clang from the craft environment

these configs are only used when not using `OC_CI_SQUISH` (image with craft environment). So should be fine, right?

TheOneRing:
> > Please update
> > https://github.com/owncloud/client/blob/5b5b25557840bc273fffa38613d8f7937b484091/.drone.star#L72-L76
> > 
> > to use /usr/bin/clang else it uses the clang from the craft environment
> 
> these configs are only used when not using `OC_CI_SQUISH` (image with craft environment). So should be fine, right?

I don't really see a reason to no longer provide them, also passing the build type is always a good idea and you removed that along with the compiler.

TheOneRing:
(Also can you clean up the commit history, or should I squash the pr?)

Salipa-Gurung (dismissed):

saw-jan:
> (Also can you clean up the commit history, or should I squash the pr?)

we usually do `Squish and merge`

saw-jan:
>I don't really see a reason to no longer provide them, also passing the build type is always a good idea and you removed that along with the compiler.

okey, I will add the build_type on every build

TheOneRing (approved):

Issue #11593, opened 2024-04-09
Adopt Update dialogs to the new modal widget experience


TheOneRing:
The update dialogs are still simple dialogs that are now displayed on the full window size.
We need to evaluate how we can update their appearance for a dialog free approach.
 
https://github.com/owncloud/client/blob/8cbcb2ccf71b18410a73169be834ffc64c3c344b/src/gui/updater/ocupdater.cpp#L417-L417
https://github.com/owncloud/client/blob/8cbcb2ccf71b18410a73169be834ffc64c3c344b/src/gui/updater/ocupdater.cpp#L56-L56

They are currently displayed as a tiny dialog in the top left corner.

erikjv:
Current state:

<img width="884" alt="Screenshot 2024-04-23 at 17 39 02" src="https://github.com/owncloud/client/assets/151956/1a3f2938-8e0d-4743-a324-1a043d0b04af">

erikjv:
How can I get the message box to show up?

saw-jan:
Tested with `ownCloud 6.0.0.14181-daily20240709 [13f49d]`

![Screenshot from 2024-07-10 13-06-50](https://github.com/owncloud/client/assets/52366632/d31b9221-4002-4553-afb8-5982dd7edcb0) | ![Screenshot from 2024-07-10 13-07-02](https://github.com/owncloud/client/assets/52366632/ccf552c0-bffc-436f-a18b-05a89f90e780)
--|--

Is this the expected UI for 'Update available' dialog?
@TheOneRing @erikjv

TheOneRing:
Hmm thats the restart dialog, by then the update is already installed.

saw-jan:
I mean this dialog.
![Screenshot from 2024-07-10 13-06-50](https://github.com/owncloud/client/assets/52366632/a37a42b2-0ffd-4691-a60a-9a8a15303686)

TheOneRing:
Oh I thought that was an old screenshot, thx we should update the dialog thx.

PR #11617, opened 2024-04-25
Adopt Update dialogs to the new modal widget experience

erikjv:
Fixes: #11593

### DIFF
--- a/src/gui/updater/ocupdater.cpp
+++ b/src/gui/updater/ocupdater.cpp
@@ -22,6 +22,7 @@
 
 #include "settingsdialog.h"
 #include "updatedownloadeddialog.h"
+#include "updater/newversionavailabledialog.h"
 #include "updater/ocupdater.h"
 #include "updater/updater_private.h"
 
@@ -52,10 +53,10 @@ UpdaterScheduler::UpdaterScheduler(Application *app, QObject *parent)
             // prevent dialog from being displayed twice (rather unlikely, but it won't hurt)
             if (_updateDownloadedDialog == nullptr) {
                 _updateDownloadedDialog = new UpdateDownloadedDialog(app->gui()->settingsDialog(), updater->statusString());
-                _updateDownloadedDialog->setAttribute(Qt::WA_DeleteOnClose);
                 ocApp()->gui()->settingsDialog()->addModalWidget(_updateDownloadedDialog);
 
                 connect(_updateDownloadedDialog, &UpdateDownloadedDialog::accepted, this, []() { RestartManager::requestRestart(); });
+                connect(_updateDownloadedDialog, &UpdateDownloadedDialog::finished, this, [this]() { delete _updateDownloadedDialog.data(); });
             }
         });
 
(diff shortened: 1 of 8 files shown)


erikjv:
New message box:

Screenshot 2024-05-03 at 14 57 17

erikjv:
Update downloaded dialog:

Screenshot 2024-05-03 at 15 06 54

TheOneRing:
Hmm could we use the same icon, the app icon, for both?
Also I guess both should be centered and not aligned to top?

src/gui/updater/ocupdater.cpp
```diff
@@ -376,50 +377,17 @@ void WindowsUpdater::versionInfoArrived(const UpdateInfo &info)
...
-    QLabel *ico = new QLabel;
-    ico->setFixedSize(iconSize, iconSize);
-    ico->setPixmap(infoIcon.pixmap(iconSize));
-    QLabel *lbl = new QLabel;
     QString txt = tr("<p>A new version of the %1 Client is available.</p>"
                      "<p><b>%2</b> is available for download. The installed version is %3.</p>")
                       .arg(Utility::escape(Theme::instance()->appNameGUI()),
                           Utility::escape(info.versionString()), Utility::escape(Version::versionWithBuildNumber().toString()));
+    auto *widget = new NoUrlWidget(ocApp()->gui()->settingsDialog(), txt);
+    widget->setAttribute(Qt::WA_DeleteOnClose);
```
TheOneRing, line 385: Its not a dialog anymore, we should handle its deletion.

src/gui/updater/nourlwidget.h
```diff
@@ -0,0 +1,47 @@
...
+}
+
+namespace OCC {
+
+class NoUrlWidget : public QWidget
+{
+    Q_OBJECT
+
+public:
+    explicit NoUrlWidget(QWidget *parent, const QString &statusMessage);
```
TheOneRing, line 31: What is the meaning of the class name?

src/gui/updater/updatedownloadeddialog.cpp
```diff
@@ -44,9 +44,14 @@ UpdateDownloadedDialog::UpdateDownloadedDialog(QWidget *parent, const QString &s
     yesButton->setDefault(true);
 }
 
-UpdateDownloadedDialog::~UpdateDownloadedDialog()
+void UpdateDownloadedDialog::accept()
 {
-    delete _ui;
+    Q_EMIT accepted();
+    close();
```
TheOneRing, line 50: We should rather delete the class in the caller than rely on delete on close.

TheOneRing (changes requested):
Looks great, but please take a look on my comments.

src/gui/updater/nourlwidget.h
erikjv: It comes from the method that previously showed the dialog: `WindowsUpdater::showNoUrlDialog`. It's called from `WindowsUpdater::versionInfoArrived`, when the info has an empty URL. 🤷‍♂️

erikjv: Rename to `NewVersionAvailableDialog`, same for the calling method.

src/gui/updater/updatedownloadeddialog.cpp
erikjv: Do memory management on the creating side, but check the signals.

TheOneRing (approved):

PR #11622, opened 2024-05-02
Rewrite toolbar

TheOneRing:


src/gui/CMakeLists.txt
```diff
@@ -126,6 +126,8 @@ ecm_add_qml_module(owncloudGui
         VERSION 1.0
         NAMESPACE OCC
         QML_FILES
+            qml/AccountBar.qml
```
TheOneRing, line 129: Should I start a new subfolder?

src/gui/settingsdialog.cpp
```diff
@@ -74,207 +57,58 @@ auto minimumSizeHint(const QWidget *w)
...
-    connect(_addAccountAction, &QAction::triggered, this, [] {
-        // don't directly connect here, ocApp might not be defined yet
-        ocApp()->gui()->runNewAccountWizard();
-    });
-    _ui->toolBar->addAction(_addAccountAction);
+    // People perceive this as a Window, so also make Ctrl+W work
+    addAction(tr("Hide"), Qt::CTRL | Qt::Key_W, this, &SettingsDialog::hide);
+
+    _ui->quickWidget->rootContext()->setContextProperty(QStringLiteral("settingsDialog"), this);
+    // TODO: fix sizing
```
TheOneRing, line 80: Any better idea?

src/gui/CMakeLists.txt
erikjv: No, don't think that's necessary yet.

src/gui/settingsdialog.cpp
erikjv: No, not yet

PR #11640, opened 2024-05-10
Fix accessibility issues in settings tab

erikjv:
- buddies
- accessibility names
- tab order
- fix label text so screen readers produce something useful

Fixes: #11636


erikjv:
Screenshot 2024-05-15 at 17 19 53

TheOneRing (approved):

Issue #11635, opened 2024-05-08
The arrows "<" and ">" should be hidden by screen readers. This ensures


TheOneRing:

erikjv:
I think we should remove the arrows completely

TheOneRing:
Can we have them be managed by the qpa?

erikjv:
Apparently we're not using a `QWizard` here, so I changed our button logic to be the same as what `QWizard` does.

PR #11641, opened 2024-05-13
Improve accessibility for the setup wizard

erikjv:
Fixes: #11635


TheOneRing (approved):

TheOneRing (changes requested):
During our last call we discussed a different approach, please investigate.

TheOneRing:
> > During our last call we discussed a different approach, please investigate.
> 
> I cannot find a way to set some kind of "wizard-next-button" accessibility role. Any thoughts?

Hmm yes https://github.com/qt/qtbase/blob/0beaa63d4adf6a30c9f48919c1d9f4908de738b4/src/widgets/dialogs/qwizard.cpp#L623-L628
But I guess we should stick to the upstream text?

erikjv:
> But I guess we should stick to the upstream text?

Done

src/gui/newwizard/setupwizardwindow.ui
```diff
@@ -101,7 +101,7 @@
         <item>
          <widget class="QPushButton" name="errorMessageDismissButton">
           <property name="text">
-           <string>Dismiss</string>
+           <string>&amp;Dismiss</string>
```
TheOneRing, line 104: Apparently Mac doesn't like accessibility, see `Continue, `Back` and `Finish` ....
So I'm a bit lost here... 🙈

TheOneRing (approved):

src/gui/newwizard/setupwizardwindow.ui
erikjv: Mac ignores those, they only show up on Windows and Linux.

Issue #11637, opened 2024-05-08
"Local sync protocol" tab a11y issues


TheOneRing:
The heading of the "Local synchronization log" table is not recognized as such and is only output as normal text. In addition the table itself is output without an accessible name.

erikjv:
What does "Local sync protocol" actually mean? Same for the tab title. Is this the counterpart to "Server Activity"? If so, maybe we can remove that label and name the tab "Local Activity"?

TheOneRing:
So the activity tab provides a server side feed of what happens on the server (oc10).
This feed also includes events for unsynced folders or events that happened in the past.
The sync protocol only contains what the client did, upload/download/move....

saw-jan:
Tested with `ownCloud 6.0.0.13899-daily20240522 [7ba266]`

![Screenshot from 2024-05-29 09-36-35](https://github.com/owncloud/client/assets/52366632/833b101d-cc6c-41c3-89a4-4dd72c371b1b) | ![Screenshot from 2024-05-29 09-35-17](https://github.com/owncloud/client/assets/52366632/00112860-10dc-4afb-a4e5-abb2159abea8)
--|--
old | new

PR #11644, opened 2024-05-15
Rename "Local sync protocol" tab in activity view to "Local Activity"

erikjv:
This makes the purpose more clear to users, and removes the need for the label above the table. In case of oc10, it also makes the difference clear between this tab and the "Server Activity" tab.

The table itself now also has an a11y name.

Fixes: #11637

### DIFF
--- a/src/gui/protocolwidget.ui
+++ b/src/gui/protocolwidget.ui
@@ -10,22 +10,9 @@
     <height>515</height>
    </rect>
   </property>
-  <property name="windowTitle">
-   <string>Form</string>
-  </property>
   <layout class="QVBoxLayout" name="verticalLayout">
    <item>
     <layout class="QHBoxLayout" name="horizontalLayout">
-     <item>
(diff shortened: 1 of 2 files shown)


src/gui/protocolwidget.ui
```diff
@@ -53,6 +40,9 @@
      <property name="contextMenuPolicy">
       <enum>Qt::CustomContextMenu</enum>
      </property>
+     <property name="accessibleName">
```
TheOneRing, line 43: Hmm the tab is added in code but the description in the .ui?

erikjv: See next line:
```
<string>Local activity table</string>
```
It's the name for the table widget.

TheOneRing (approved):

saw-jan:
:heavy_check_mark: Tested. see https://github.com/owncloud/client/issues/11637#issuecomment-2136464033

PR #11647, opened 2024-05-22
OAuth: Only use one auth methode

TheOneRing:
Historically the client used both basic auth and the post body to provide the client id and secret. Some idp's however are picky and only accept one.


butonic (dismissed):
first basic then post ❤️

erikjv (approved):

PR #11664, opened 2024-05-27
Fix possible crash if the client encountered an error during folder s…

TheOneRing:
…etup


src/gui/socketapi/socketapi.cpp
```diff
@@ -952,8 +952,9 @@ bool OCC::SocketApi::FileData::isSyncFolder() const
 
 SyncFileStatus SocketApi::FileData::syncFileStatus() const
 {
-    if (!folder)
+    if (!folder || folder->canSync()) {
```
erikjv, line 955: Shouldn't that be `!folder->canSync()` ? Otherwise I don't understand what's happening.

TheOneRing: 🙈  thx

erikjv (approved):

PR #11669, opened 2024-05-31
Remove event handler from setup wizard

erikjv:
Wizards on all platforms do not react to enter keys when buttons do not have focus.


TheOneRing (changes requested):
Please also clean up the install eventfilter calls.

erikjv:
> Please also clean up the install eventfilter calls.

Done.

TheOneRing (approved):

PR #11675, opened 2024-06-04
[gui-tests][full-ci] define python dependencies in a requirements file

saw-jan:
I have added a `requirements.txt` file which contains the list of python modules that required for the GUI test code. These modules can then be installed using pip command.

### Motivation
- to keep record of python modules that we require
- add/remove modules easily in the future


test/gui/Makefile
```diff
@@ -15,6 +15,21 @@
...
+.PHONY: install
+install: pnpm-install pnpm-install-chromium pip-install
+
+.PHONY: pnpm-install
+pnpm-install:
+	cd webUI && pnpm install
+
+.PHONY: pnpm-install-chromium
+pnpm-install-chromium:
+	pnpm dlx playwright install chromium
```
saw-jan, line 27:
```suggestion
	pnpm exec playwright install chromium
```

saw-jan: exec doesnt work
```
pnpm exec playwright install chromium
undefined
 ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL  Cannot read properties of undefined (reading '/drone/src/test/gui')
make: Leaving directory '/drone/src/test/gui'
```

PrajwolAmatya (approved):

Salipa-Gurung (approved):

Issue #11653, opened 2024-05-22
[QA] Automate GUI regression tests


Salipa-Gurung:
### Synchronization

- [x] **Sync of multiple items** - Covered in scenario: `Scenario: Syncing folders each having 500 files`
  
   | ID | Test cases | Steps to reproduce | Expected Result | Assignee |
    --|--|--|--|--
   | 1 | Sync of multiple items | Sync 1000 files from the client to the server | Files are synced to the server | |
   | 2 | Sync of multiple items | Sync 1000 files from the server to the client | Files are synced to the client | |

- [x] **Force sync now** - Covered in scenario: `Scenario: Try to sync files having space at the end (Windows only)`
  
    | ID | Test cases | Steps to reproduce | Expected Result |  Assignee |
    --|--|--|--|--
     | 1 | Force sync now | 1. Make some changes to your files/folders in ownCloud <br>2. Go to your Account tab<br>3. Open three dots menu<br>4.Select 'Force sync now' | Sync has been trigered immediately |  |

- [x] **Pause/Resume sync** - Covered in scenario: `Scenario: Syncing a file from the server and creating a conflict`

    | ID | Test cases | Steps to reproduce | Expected Result |  Assignee |
    --|--|--|--|--
    | 1 | Pause sync | 1. Connect the client to the server<br>2. Let the file download be about >100MB<br>3. Pause the sync | Sync has been paused | |
    | 2 | Resume sync | Resume the sync | Sync has been resumed from where the download was paused | |

- [x] **Account: Add Folder**

    | ID | Test cases | Steps to reproduce | Expected Result |  Assignee |
    --|--|--|--|--
    | 1 | Add folder | 1. Open ownCloud desktop client<br>2. Go to account tab<br>3. Click Add Folder | 1. `Pick a local folder on your computer to sync` window has been opened<br>2. Cancel, Back and Next buttons present, Back button greyed out<br>3. New folder has been prepopulated but can be changed | @PrajwolAmatya <br>Step 1 and Step 3 has been covered in Step `When the user selects manual sync folder option in advanced section` <br> For Step 2 check for Cancel and Back buttons are skipped |
[...]

PR #11677, opened 2024-06-05
[tests-only] [full-ci] Keep files after deletion of all files

Salipa-Gurung:
Added test coverage for checking that the files are recreated during sync when `Keep files` is selected in `Remove all files` dialog after deleting all the files.

Related issue: https://github.com/owncloud/client/issues/11653

### DIFF
--- a/test/gui/tst_deletFilesFolders/test.feature
+++ b/test/gui/tst_deletFilesFolders/test.feature
@@ -13,6 +13,7 @@ Feature: deleting files and folders
         Given user "Alice" has uploaded file with content "ownCloud test text file 0" to "<fileName>" in the server
         And user "Alice" has set up a client with default settings
         When the user deletes the file "<fileName>"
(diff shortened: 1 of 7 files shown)


test/gui/tst_deletFilesFolders/test.feature
```diff
@@ -47,3 +49,12 @@ Feature: deleting files and folders
         And as "Alice" folder "test-folder1" should not exist in the server
         And as "Alice" file "textfile2.txt" should exist in the server
         And as "Alice" folder "test-folder2" should exist in the server
+        When the user deletes the file "textfile2.txt"
+        And the user deletes the folder "test-folder2"
+        And the user confirms to keep the files
```
saw-jan, line 54: hmm, master branch doesn't have this now

saw-jan:
Let's not invest on testing feature that are removed in next version

PR #11682, opened 2024-06-06
Add a11y role to the nav bar

erikjv:
This causes it to be a group on macOS for which the accessible name will be used.


erikjv:
@TheOneRing Can you test this on Windows?

erikjv:
(This doesn't fix the checkbox role of some buttons)

Issue #11678, opened 2024-06-05
It appears that the image shown in the screenshot is a decorative element, but it is recognized by screen readers as an image without alternative text.


erikjv:

erikjv:
@TheOneRing Shall I add a descriptive text like "Notification icon"?

PR #11689, opened 2024-06-07
Set a11y name on notification icon

erikjv:
Fixes: #11678

### DIFF
--- a/src/gui/notificationwidget.cpp
+++ b/src/gui/notificationwidget.cpp
@@ -142,7 +142,7 @@ void NotificationWidget::changeEvent(QEvent *e)
     case QEvent::StyleChange:
     case QEvent::PaletteChange:
     case QEvent::ThemeChange:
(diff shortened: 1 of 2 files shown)


src/gui/notificationwidget.cpp
```diff
@@ -35,6 +35,7 @@ NotificationWidget::NotificationWidget(QWidget *parent)
     _ui->setupUi(this);
     _progressIndi = new QProgressIndicator(this);
     _ui->horizontalLayout->addWidget(_progressIndi);
+    _ui->_notificationIcon->setAccessibleName(QStringLiteral("Notification icon"));
```
TheOneRing, line 38:
```suggestion
    _ui->_notificationIcon->setAccessibleName(tr("Notification icon"));
```

TheOneRing (changes requested):
^

PR #11699, opened 2024-06-14
Ci: Update checkout action

TheOneRing:


TheOneRing:
An example for a run with invalid formatted qml code can be found here https://github.com/owncloud/client/actions/runs/9515560855/job/26230040677

erikjv (approved):

PR #11701, opened 2024-06-14
WIP

erikjv:


erikjv:
I just realised that maybe I should set the focus from the log button on the whole NetworkSettings widget, and let that delegate it further.

erikjv:
Done in another PR

PR #11715, opened 2024-06-21
Fix focusing the first/last item in a qtquickwidget

TheOneRing:


erikjv:
Mac: ok
Windows: I cannot see where the focus goes after the "Add Space" button.

Also: I noticed that when there is an account-modal dialog (i.e. I was logged out, and the login "dialog" was showing), it wasn't very clear that I couldn't interact with the toolbar. Maybe that should be grayed out a bit? (Same with the high-contrast theme.)

erikjv (approved):

TheOneRing:
Thanks for testing!
Yes you where right, https://github.com/owncloud/client/compare/882f9ae8282d1829de5de237bdcdd32f0d253e49..88677493cde27452133dde6bae7a8a7d45efb5d2#diff-6165ec3f89e4d38e966727da062874b3e0a26ccd90405a72cc2984aee10fe30bR30 should have fixed that issue.

erikjv (dismissed):

PR #11719, opened 2024-06-25
Center content of the LoginRequiredDialog

erikjv:


erikjv:
Note: it is slightly off-center (to the top), because there is an invisible error label with a retry button. Changing that to be hidden would have the other UI elements "jump" when it is shown, so not a good idea.

TheOneRing (approved):

PR #11723, opened 2024-06-28
Work/a11y column sorting

erikjv:


erikjv:
Filtering is still there. I think it's better to remove it (there is a button for that!), and move the items in the sort sub-menu into the context menu itself. But with this version we can try and see how this implementation "feels".

erikjv:
With filter submenu:

Screenshot 2024-07-01 at 18 24 38

erikjv:
Without filter submenu, and putting the sort actions directly into the context menu:

Screenshot 2024-07-01 at 18 36 29

src/gui/protocolwidget.cpp
```diff
@@ -104,71 +104,111 @@ QMenu *ProtocolWidget::showFilterMenu(QWidget *parent, Models::SignalledQSortFil
...
+        auto showInWebBrowserAction = menu->addAction(CommonStrings::showInWebBrowser());
+        showInWebBrowserAction->setEnabled(false);
+        fetchPrivateLinkUrl(data.folder()->accountState()->account(), data.folder()->webDavUrl(), data.folder()->remotePathTrailingSlash() + data.path(),
+            parent, [showInWebBrowserAction, parent, pos = menu->actions().size(), menu = QPointer<QMenu>(menu)](const QUrl &url) {
+                // as fetchPrivateLinkUrl is async we need to check the menu still exists
+                if (menu) {
+                    auto replacement = new QAction(CommonStrings::showInWebBrowser());
+                    connect(replacement, &QAction::triggered, [url, parent] { Utility::openBrowser(url, parent); });
+                    menu->insertAction(showInWebBrowserAction, replacement);
+                    menu->removeAction(showInWebBrowserAction);
```
TheOneRing, line 145: I think we should be able to update showInWebBrowserAction ?

TheOneRing:
Looks good, but we should keep in mind that we can also trigger the menu by keyboard with the menu button on Windows.
So in that case we should not display it at the cursor position.

src/gui/protocolwidget.cpp
erikjv: done

erikjv:
> Looks good, but we should keep in mind that we can also trigger the menu by keyboard with the menu button on Windows.
> So in that case we should not display it at the cursor position.

Done, we now take the position of the `QTreeWidget::customContextMenuRequested` signal

TheOneRing (approved):

PR #11724, opened 2024-07-01
Pass enabled state of qml icons

TheOneRing:


src/resources/qmlresources.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+#include "resources/resources.h"
+
+using namespace OCC;
+QUrl Resources::QMLResources::resourcePath(const QString &theme, const QString &icon, bool enabled)
+{
+    auto getBool = [](bool b) { return b ? QStringLiteral("true") : QStringLiteral("false"); };
+    const auto data = QStringLiteral("{\"theme\": \"%1\", \"enabled\": \"%2\", \"icon\": \"%3\", \"systemtheme\": \"%4\"}")
+                          .arg(theme, getBool(enabled), icon, getBool(Resources::isUsingDarkTheme()));
+    return QUrl(QStringLiteral("image://ownCloud/") + QString::fromUtf8(data.toUtf8().toBase64()));
```
TheOneRing, line 25: Not a fan of all those string conversions.

src/resources/resources.cpp
```diff
@@ -182,29 +184,26 @@ CoreImageProvider::CoreImageProvider()
...
+        icon = themeIcon(iconName);
     }
+
     // the sourceSize of the Image must be provided
     Q_ASSERT(requestedSize.isValid());
     const QSize actualSize = requestedSize.isValid() ? requestedSize : icon.availableSizes().constFirst();
     if (size) {
         *size = actualSize;
     }
-    return icon.pixmap(actualSize);
```
TheOneRing, line 209: Apparently there is nothing comparable supported by qt out of the box, making me think that there might be an alternative approach available for qt?

src/resources/qmlresources.cpp
erikjv: Hmmm, can we somehow use a custom value type?

TheOneRing: As far as I understand image providers, all they get is a url...

erikjv (approved):

src/resources/resources.cpp
erikjv: apparently not

src/resources/qmlresources.cpp
erikjv: yes, dunno what I was thinking

PR #11737, opened 2024-07-09
[gui-tests][full-ci]update step to only check if the local folder is present in the remote destination wizard

PrajwolAmatya:
### Description
This PR updates the step
```feature
Then the folder "test-folder" should be present and selected in the remote destination wizard
```
to
```feature
Then the folder "test-folder" should be present in the remote destination wizard
```
Checking whether the folder is selected or not seemed to be flaky because of the timing issue for folder to be selected, so only checking the presence of the folder in the remote destination wizard.


saw-jan (approved):
:+1:

Issue #11566, opened 2024-03-28
[6][Win] Cannot enable VFS after disabling it


saw-jan:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the bug

VFS cannot be enabled after disabling it

![Screenshot from 2024-03-28 12-55-01](https://github.com/owncloud/client/assets/52366632/ad5752fd-11d5-465b-bae2-7dc1ce44bdac)

![Screenshot from 2024-03-28 12-55-14](https://github.com/owncloud/client/assets/52366632/a4682f4f-13b1-483b-88b7-f8d731700b9b)

### Expected behavior

Able to enable the VFS

### Logs

[ownCloud.log](https://github.com/owncloud/client/files/14785314/ownCloud.log)

### Client version number

ownCloud 6.0.0.13582-daily20240327 [4258b6](https://github.com/owncloud/client/commit/4258b6c0adf207fc4a1de96ccd9e13d5836e526e)
Libraries Qt 6.6.2, OpenSSL 3.2.1 30 Jan 2024
Using virtual files plugin: wincfapi
OS: windows-10.0.22621 (build arch: x86_64, CPU arch: x86_64)
QPA: windows

ishabaral:
Tested with: `ownCloud 6.0.0.14906-daily20241029 [b056cf]`

![Screenshot from 2024-10-29 16-14-38](https://github.com/user-attachments/assets/9268b9ae-f402-4349-bd1b-3bebd50a64bb)

![Screenshot from 2024-10-29 16-14-22](https://github.com/user-attachments/assets/62ba7679-db37-4843-816f-4bbdb6c4d9ae)

PR #11738, opened 2024-07-09
Show "Enable VFS" option after disableing VFS

erikjv:
Fixes: #11566


erikjv:
In the scenario above, we should also show the enable option

TheOneRing:
image
image

There is no action displayed at all

TheOneRing (changes requested):
^

TheOneRing (approved):

Issue #11739, opened 2024-07-10
Update AppImageUpdateAvailableDialog to not use a modal dialg


TheOneRing:
Update https://github.com/owncloud/client/blob/72bdc82bd1abdfd2593834ae4af5575efb09d330/src/gui/updater/appimageupdater.cpp#L164-L164

erikjv:
How can I test this?

PR #11741, opened 2024-07-11
Work/appimage updater dialog

erikjv:
Fixes: #11739


src/gui/updater/appimageupdater.cpp
```diff
@@ -161,14 +161,15 @@ void AppImageUpdater::versionInfoArrived(const UpdateInfo &info)
         return;
     }
 
-    auto dialog = new AppImageUpdateAvailableDialog(currentVersion, newVersion, ocApp()->gui()->settingsDialog());
+    auto widget = new AppImageUpdateAvailableWidget(currentVersion, newVersion, ocApp()->gui()->settingsDialog());
+    widget->setAttribute(Qt::WA_DeleteOnClose);
```
TheOneRing, line 165: I guess the close event will never be emitted as this is a plain widget.
I guess we should explicitly handle the delete in connects, `accepted`...?

erikjv: That's done in the widget: the signal handler for both buttons calls a `close()` after the signal emission.

src/gui/updater/appimageupdater.cpp
```diff
@@ -161,14 +161,24 @@ void AppImageUpdater::versionInfoArrived(const UpdateInfo &info)
...
 
-    auto dialog = new AppImageUpdateAvailableDialog(currentVersion, newVersion, ocApp()->gui()->settingsDialog());
+    QPointer<AppImageUpdateAvailableWidget> widget(new AppImageUpdateAvailableWidget(currentVersion, newVersion, ocApp()->gui()->settingsDialog()));
+    widget->setAttribute(Qt::WA_DeleteOnClose);
 
-    connect(dialog, &AppImageUpdateAvailableDialog::skipUpdateButtonClicked, this, [newVersion]() {
+    connect(widget, &AppImageUpdateAvailableWidget::skipUpdateButtonClicked, this, [newVersion, widget]() {
         qCInfo(lcUpdater) << "Update" << newVersion << "skipped by user";
         setPreviouslySkippedVersion(newVersion);
+        if (widget) {
```
TheOneRing, line 170: As the connect is on the widget it must still exist at this time.
Please also use deleteLater.

src/gui/updater/appimageupdater.cpp
TheOneRing: Please fix

src/gui/updater/appimageupdater.cpp
```diff
@@ -182,10 +192,14 @@ void AppImageUpdater::versionInfoArrived(const UpdateInfo &info)
 
         setDownloadState(Downloading);
         appImageUpdaterShim->startUpdateInBackground();
+
+        if (widget) {
```
TheOneRing, line 196: see above

TheOneRing (changes requested):

TheOneRing (approved):

PR #11761, opened 2024-07-29
[gui-tests][full-ci] extend account login scenario

saw-jan:

PR #11767, opened 2024-07-31
[gui-tests][full-ci]add test to check for open connection sync button to be disabled

PrajwolAmatya:
### Description
This PR checks for the button to open sync connection wizard to be disabled when all the available spaces are already synced.

Test for: https://github.com/owncloud/client/pull/11760

Skipped the scenario: `Scenario: login, logout and restart with oauth2 authentication`
`@skip` tag will be removed after the fix in PR https://github.com/owncloud/client/pull/11761

### DIFF
--- a/test/gui/shared/scripts/pageObjects/SyncConnectionWizard.py
+++ b/test/gui/shared/scripts/pageObjects/SyncConnectionWizard.py
@@ -375,3 +375,9 @@ class SyncConnectionWizard:
     @staticmethod
     def get_warn_label():
         return str(squish.waitForObjectExists(SyncConnectionWizard.WARN_LABEL).text)
+
+    @staticmethod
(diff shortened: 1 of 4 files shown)


test/gui/shared/scripts/pageObjects/SyncConnectionWizard.py
```diff
@@ -375,3 +375,9 @@ def get_local_sync_path():
     @staticmethod
     def get_warn_label():
         return str(squish.waitForObjectExists(SyncConnectionWizard.WARN_LABEL).text)
+
+    @staticmethod
+    def is_button_enabled():
```
saw-jan, line 380:
```suggestion
    def is_add_sync_folder_button_enabled():
```

saw-jan (approved):

PR #11773, opened 2024-07-31
Fix label displaying the synced/unsynced spaces count

erikjv:


src/gui/accountsettings.cpp
```diff
@@ -77,6 +77,13 @@ AccountSettings::AccountSettings(const AccountStatePtr &accountState, QWidget *p
 
     connect(FolderMan::instance(), &FolderMan::folderListChanged, _model, &FolderStatusModel::resetFolders);
     if (accountsState()->supportsSpaces()) {
+        for (auto folder : FolderMan::instance()->folders()) {
```
TheOneRing, line 80: Hmm if we are connected I think the lambda should have been executed.
So do we need the extra round of counting here?

erikjv: You're right... If only the lambda were correct ;-)

TheOneRing (approved):

saw-jan:
Different bug: https://github.com/owncloud/client/issues/11889

Issue #11673, opened 2024-06-04
client sometime crashes in OCC::AccountSettings::addModalLegacyDialog at `src/gui/accountsettings.cpp:559`


saw-jan:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the bug

Desktop client sometimes can crash in `OCC::AccountSettings::addModalLegacyDialog` at `/drone/src/src/gui/accountsettings.cpp:559`

Crashed in:
- https://drone.owncloud.com/owncloud/client/18407
- https://drone.owncloud.com/owncloud/client/18408

```c
###########################################
Scenario: simple sharing of a file by public link with default expiration date
-------------------------------------------
Executable: /drone/src/build/bin/owncloud
Timestamp: 2024-06-04 05:42:03
Process ID: 4715
Signal Number: 6
-------------------------------------------
<<<<< STACKTRACE START >>>>>
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
Core was generated by `/drone/src/build/bin/owncloud -s --logfile - --logdebug --logflush'.
Program terminated with signal SIGABRT, Aborted.
#0  0x00007f82d0f42834 in __pthread_kill_implementation () from /lib64/libc.so.6
[Current thread is 1 (Thread 0x7f82cd32e940 (LWP 4715))]
#0  0x00007f82d0f42834 in __pthread_kill_implementation () from /lib64/libc.so.6
No symbol table info available.
#1  0x00007f82d0ef08ee in raise () from /lib64/libc.so.6
No symbol table info available.
#2  0x00007f82d0ed898f in abort () from /lib64/libc.so.6
No symbol table info available.
#3  0x00007f82d14aea4a in qAbort() () from /ownbuild/master/linux-gcc-x86_64-squish/lib/libQt6Core.so.6
No symbol table info available.
#4  0x00007f82d14e3a05 in ?? () from /ownbuild/master/linux-gcc-x86_64-squish/lib/libQt6Core.so.6
No symbol table info available.
#5  0x00007f82d14af6d8 in QMessageLogger::fatal(char const*, ...) const () from /ownbuild/master/linux-gcc-x86_64-squish/lib/libQt6Core.so.6
[...]

PR #11793, opened 2024-08-09
Add "debug" logging to help with issue #11673

erikjv:
Helps with debugging: #11673


erikjv:
I've looked at this multiple times, but I'm not sure what the cause is. I also cannot reproduce it. So if it happens in the wild, we now have a bit more information on what might be causing it.

TheOneRing (approved):

erikjv:
For testing: please try to reproduce #11673, and see what the logs say right before the assert fails.

PR #11801, opened 2024-08-14
Add support for public clients

TheOneRing:


LuKePicci:
Hi guys, I'm sorry to revamp this old issue but I don't know a better place to ask or save this note, Google pops this out so I think it is somewhat useful to leave a ref.
I recently stumbled upon OC clients setup for SSO and I have got surprised noticing they were built/developed and maintained for so long time as if they were Oauth2 **confidential clients** even if they clearly are **public clients** I was indeed searching for a DEV note/readme explaining why such clients were not fixed to comply with proper public clients implementation while I found this old issue, so I have got two questions for you:
1) Why this PR _adds support for public clients_ rather then entirely fix this by _dropping confidential client_ implementation?
2) Why this fix/change was closed and never reappeared in other discussions?
3) Most Oauth2/OIDC capable IdP's handle public and confidential clients differently and some may also automatically determine whether it is public or confidential depending on the fact that a client credential (secret/cert/whatever) has been issued. This is backed by a statement in the standard about and IdP MUST NOT issue any credential for a client which is not expected to be authenticated. So, which was the initial OIDC reference IdP implementation against which OC clients were developed? I can't understand how it could allow an app to have a public client but still have a valid secret defined.

This whole thing is indeed curious to investigate as it turns out that the Single Page Application of the web client (which also is a public client) was correctly implemented (no secret is passed to token endpoint) while wrongly documented here (clearly includes a secret to the same token endpoint request): https://owncloud.dev/apis/http/authorization/

PR #11810, opened 2024-08-19
[a11y] Move server URL from status label to Account button

erikjv:
The URL in the status label of the account settings page was not always accessible by keyboard. The label will now only show the connection status, and the "Manage Account" button has a new entry for "Open in Web Browser".

### DIFF
--- a/changelog/unreleased/11810
+++ b/changelog/unreleased/11810
@@ -1,0 +1,11 @@
+Bugfix: Make open account in browser accessible for keyboard navigation
+
+On the account settings page, the status label contained the URL of the
+server. This URL was not accessible with keyboard navigation or when a
+screen-reader was used. Now there is an "Open in Web Browser" action in
+the pop-up menu of the "Manage Account" button (which is next to the
+status label), and the URL has been removed from the status label.
+
+https://github.com/owncloud/client/issues/11772
+https://github.com/owncloud/client/issues/11800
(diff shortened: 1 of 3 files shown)


src/gui/accountsettings.cpp
```diff
@@ -471,42 +472,40 @@ void AccountSettings::slotAccountStateChanged()
 
     switch (state) {
     case AccountState::PausedDueToMetered:
-        showConnectionLabel(tr("Sync to %1 is paused due to metered internet connection.").arg(server));
```
TheOneRing, line 474: Please remove the `server` declaration as its now unused.
We can probably also remove the safeUrl?

TheOneRing (changes requested):
^
Looks great, but I think we can clean it up, please correct me if I'm wrong.

src/gui/accountsettings.cpp
erikjv: done

TheOneRing (approved):

changelog/unreleased/11810
```diff
@@ -0,0 +1,11 @@
+Bugfix: Make accessing the account in browser accessible for keyboard navigation
```
TheOneRing, line 1:
```suggestion
Bugfix: Make open account in browser accessible for keyboard navigation
```
?

TheOneRing (approved):

PR #11843, opened 2024-08-28
Properly use opacity instead of enabled icons

TheOneRing:
This also fixes the display of disabled avatars

image


erikjv (approved):
Nice!

Issue #11782, opened 2024-08-02
All elements of the page are not output according to their visual arrangement.


TheOneRing:
Please validate.

erikjv:
I don't know how to get Narrator or VoiceOver to read the whole dialog.

That said, we (partially) fixed the tab-order, and I think that was the cause. We still have tab-order issues: the network part is only focussed after everything else on the parenting page is cycled through. This means: .... -> Log Settings button -> About button -> No Proxy radio-button.

erikjv:
@TheOneRing setting the widget inside the Network Settings group box gives it tab-focus, but nothing inside it is focussed. The first checkbox is only tab-focussed after the About button.

TheOneRing:
> @TheOneRing setting the widget inside the Network Settings group box gives it tab-focus, but nothing inside it is focussed. The first checkbox is only tab-focussed after the About button.

You mean you gave it StrongFocus?

saw-jan:
Tested on `ownCloud 6.0.0.14688-daily20240918 [967f29]`

Tab order: `General Settings` -> `Advanced` -> `Network` -> `Network: Proxy settings` -> `Network: Download Bandwidth` -> `Network: Upload Bandwidth` -> `About`

PR #11846, opened 2024-08-29
Fix tab order in settings page

erikjv:
The network settings are now tab-selected before the about button.

Fixes: #11782


src/gui/networksettings.cpp
```diff
@@ -44,6 +44,8 @@ NetworkSettings::NetworkSettings(QWidget *parent)
 {
     _ui->setupUi(this);
 
+    setFocusProxy(_ui->pauseSyncWhenMeteredCheckbox);
```
TheOneRing, line 47: ❤️  so that's how its done.

TheOneRing (approved):

Issue #11811, opened 2024-08-19
Show an icon next to the status label in the Account Settings page


erikjv:
- Round check mark for connected or good
- Plug icon for disconnected or logged out

erikjv:
resources/core/states/sync.svg etc
Resources::getCoreIcon(QStringLiteral("states/sync"))

PR #11855, opened 2024-09-02
Show icon next to the status label in Account Settings page

erikjv:
Fixes: #11811

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -347,7 +347,7 @@ void AccountSettings::slotDisableVfsCurrentFolder(Folder *folder)
     msgBox->open();
 }
 
-void AccountSettings::showConnectionLabel(const QString &message, QStringList errors)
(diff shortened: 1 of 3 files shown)


src/gui/accountsettings.cpp
```diff
@@ -360,7 +360,29 @@ void AccountSettings::showConnectionLabel(const QString &message, QStringList er
         ui->connectionStatusLabel->setToolTip(QString());
     }
     ui->accountStatus->setVisible(!message.isEmpty());
```
TheOneRing, line 362: As we might have an overlap with https://github.com/owncloud/client/blob/master/src/gui/folderstatusmodel.cpp#L56-L56 is there a chance we can refactor the existing code?

erikjv: Yes, but it's also used for the systray icon, and as data item in the folder table model, so it's more "delicate". I think it would be better to do that in a separate PR.

Issue #9532, opened 2022-03-21
Enhance test run screenshot that will give more insight on AUT crash


Talank:
that will create an screenshot AFTER the issue
so e.g. if the app crashes this will be the screen after the crash, meaning the app would not even be visible
Do we need (additionally?) the screenshot just before the problematic action? But that might be hard to achive

_Originally posted by @individual-it in https://github.com/owncloud/client/pull/9518#discussion_r830777418_

saw-jan:
I am trying to record test video in this PR https://github.com/owncloud/client/pull/9961

saw-jan:
Blocked until https://github.com/owncloud-ci/squish/issues/34

saw-jan:
Leaving to other whoever is interested in taking this.
Some work has been done in https://github.com/owncloud/client/pull/9961
Also, see if we can make use of squish APIs: https://doc.qt.io/squish/squish-api.html#test-startvideocapture-function

PR #11857, opened 2024-09-04
[gui-tests][full-ci] configurable capture video for GUI tests

saw-jan:
This PR adds an option to capture the screen during test execution. By default the option is disabled and it can be enabled using `SCREEN_RECORD_ON_FAILURE` either env or `config.ini`.

If the option is enabled, the screen recording will be available if the test scenario fails. If the test scenario passes, there will be screen recording deleted.

This helps in debugging the failing GUI tests.

Fixes https://github.com/owncloud/client/pull/9532


test/gui/shared/scripts/bdd_hooks.py
```diff
@@ -162,13 +190,18 @@ def hook(context):
...
         directory = os.path.join(get_config("guiTestReportDir"), "screenshots")
         if not os.path.exists(directory):
             os.makedirs(directory)
 
         squish.saveDesktopScreenshot(os.path.join(directory, filename))
 
+    # check video report
+    if get_config("screenRecordOnFailure"):
+        filename = get_screenrecord_name(context.title)
+        save_screenrecord(filename)
```
saw-jan, line 203:
```suggestion
        try:
                save_screenrecord(filename)
        except:
                pass
```

test/gui/drone/log_reports.sh
```diff
@@ -38,6 +38,20 @@ check_log "stacktrace.log" "Stacktrace"
 if [[ -d "${REPORT_DIR}/screenshots" ]]; then
     echo -e "Screenshots:"
     for i in "${REPORT_DIR}"/screenshots/*.png; do
-        echo -e "\t - ${LOG_URL_PATH}/screenshots/$(basename "$i")"
+        filename=$(basename "$i")
+        if [ "$filename" != "*.png" ]; then
+            echo -e "\t - ${LOG_URL_PATH}/screenshots/$(basename "$i")"
```
PrajwolAmatya, line 43:
```suggestion
            echo -e "\t - ${LOG_URL_PATH}/screenshots/$filename"
```

PrajwolAmatya (approved):

PR #11859, opened 2024-09-04
Prepare AccountModalWidget for qml

TheOneRing:


src/gui/qmlutils.cpp
```diff
@@ -14,16 +14,22 @@
...
 #include <QQuickItem>
 #include <QQuickWidget>
 
-void OCC::QmlUtils::initQuickWidget(OCQuickWidget *widget, const QUrl &src, QObject *ocContext)
+void OCC::QmlUtils::initQuickWidget(OCQuickWidget *widget, const QUrl &src, QObject *ocContext, QWidget *parent)
 {
+    if (!parent) {
+        parent = qobject_cast<QWidget *>(ocContext);
+        Q_ASSERT_X(parent, Q_FUNC_INFO, "If invoked without an explicit parent widget, ocContext is required to be a QWidget");
+    }
```
TheOneRing, line 30: If the context object is not a widget, we could transfer the ownership to qml.
What do you think? How would we then document that....

src/gui/qmlutils.h
```diff
@@ -40,6 +41,8 @@ class OCQuickWidget : public QQuickWidget
     Q_OBJECT
 public:
     using QQuickWidget::QQuickWidget;
+    void setOCContext(const QUrl &src, QWidget *parentWidget, QObject *ocContext, QJSEngine::ObjectOwnership ownership);
```
TheOneRing, line 44: Please tell me what you think, I'm horrible in naming things and not 100% happy with the signature.

erikjv (approved):

Issue #11805, opened 2024-08-16
QML: TypeError: Cannot read property 'accountState' of null


TheOneRing:
During the destruction of widgets containing a QtQuick widget, the context is reset before the qml element is destroyed.
This results in a lot of warnings.

```
24-08-16 09:24:25:558 [ warning default ]:	qrc:/qt/qml/org/ownCloud/gui/qml/AccountBar.qml:128: TypeError: Cannot read property 'currentPage' of null
24-08-16 09:24:25:559 [ warning default ]:	qrc:/qt/qml/org/ownCloud/gui/qml/AccountBar.qml:114: TypeError: Cannot read property 'currentPage' of null
24-08-16 09:24:25:559 [ warning default ]:	qrc:/qt/qml/org/ownCloud/gui/qml/AccountBar.qml:57: TypeError: Cannot read property 'currentAccount' of null
24-08-16 09:24:25:570 [ warning default ]:	qrc:/qt/qml/org/ownCloud/gui/qml/FolderDelegate.qml:257: TypeError: Cannot read property 'syncedSpaces' of null
24-08-16 09:24:25:570 [ warning default ]:	qrc:/qt/qml/org/ownCloud/gui/qml/FolderDelegate.qml:258: TypeError: Cannot read property 'accountState' of null
24-08-16 09:24:25:571 [ warning default ]:	qrc:/qt/qml/org/ownCloud/gui/qml/FolderDelegate.qml:232: TypeError: Cannot read property 'accountState' of null
24-08-16 09:24:25:571 [ warning default ]:	qrc:/qt/qml/org/ownCloud/gui/qml/FolderDelegate.qml:239: TypeError: Cannot read property 'accountState' of null
24-08-16 09:24:25:571 [ warning default ]:	qrc:/qt/qml/org/ownCloud/gui/qml/FolderDelegate.qml:242: TypeError: Cannot read property 'accountState' of null
24-08-16 09:24:25:571 [ warning default ]:	qrc:/qt/qml/org/ownCloud/gui/qml/FolderDelegate.qml:64: TypeError: Cannot read property 'model' of null

erikjv:
Setting the source for the QtQuick widget to an empty URL fixes this: all previous QML items are destroyed, so no bindings get evaluated.

PR #11885, opened 2024-09-16
Fix QML warnings on application exit

erikjv:
The `ocContext` in referenced from the QML side is deleted before the QML items (and their binding), resulting in access to a `null` context:

```
TypeError: Cannot read property 'currentPage' of null
```

Fixes: #11805


TheOneRing:
Is there a global solution we can find https://github.com/owncloud/client/blob/master/src/gui/qmlutils.h#L38C7-L38C21 ?

TheOneRing (changes requested):

PR #11895, opened 2024-09-23
fix: avatars will be shown only based on 'profile_picture capability

DeepDiver1975:


src/libsync/capabilities.cpp
```diff
@@ -253,7 +253,7 @@ bool Capabilities::avatarsAvailable() const
 {
     auto userCaps = _fileSharingCapabilities.value(QStringLiteral("user")).toMap();
     // true by default for older servers, because turning off profile pictures was introduced in later versions
```
TheOneRing, line 255:
```suggestion
```

Issue #11889, opened 2024-09-18
synced/unsynced spaces count is `0 out of 0` when adding account, restarting client gives correct number


saw-jan:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the bug

When adding the new account, the spaces count is `0 out of 0`. But after restarting the client, the count is correct.

### Expected behavior

![Screenshot from 2024-09-18 18-25-20](https://github.com/user-attachments/assets/39c45fc3-d67a-4580-803c-a1efd68283ab)
![Screenshot from 2024-09-18 18-25-20](https://github.com/user-attachments/assets/2ba08050-6c6c-44e5-a536-0fe943ed70d0)

### Client version number

ownCloud 6.0.0.14688-daily20240918 [967f29](https://github.com/owncloud/client/commit/967f29b37aaff3d7c89226322230af58f1570dcc)

PR #11897, opened 2024-09-23
Fix count of synced spaces

erikjv:
When a new account is added, the settings page for the account is created with the account in the disconnected state. At that point, it is not yet known if the server supports spaces, so there is no space manager yet, and the update signal is not connected. Moving the signal connect to `slotAccountStateChanged` for the `Connected` state fixes the issue.

Fixes: #11889

### DIFF
--- a/src/gui/accountsettings.cpp
+++ b/src/gui/accountsettings.cpp
@@ -488,6 +488,10 @@ void AccountSettings::slotAccountStateChanged()
             icon = StatusIcon::Warning;
         }
         showConnectionLabel(tr("Connected"), icon, errors);
+        if (accountsState()->supportsSpaces()) {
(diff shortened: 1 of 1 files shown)


src/gui/accountsettings.cpp
```diff
@@ -488,6 +488,10 @@ void AccountSettings::slotAccountStateChanged()
             icon = StatusIcon::Warning;
         }
         showConnectionLabel(tr("Connected"), icon, errors);
+        if (accountsState()->supportsSpaces()) {
+            connect(accountsState()->account()->spacesManager(), &GraphApi::SpacesManager::updated, this, &AccountSettings::slotSpacesUpdated);
```
TheOneRing, line 492: Won't that result in a double connect? https://github.com/owncloud/client/blob/460e3229af19072550423e44bde1dc0a06dffcab/src/gui/accountsettings.cpp#L80-L80

TheOneRing (approved):

PR #11899, opened 2024-09-25
fix: transifex push/pull

DeepDiver1975:


.github/workflows/push-transifex.yml
```diff
@@ -0,0 +1,19 @@
+name: "Push translation source and German translation to Transifex"
```
TheOneRing, line 1: English only

.github/workflows/push-transifex.yml
```diff
@@ -0,0 +1,19 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: l10n-push-source
+        uses: transifex/cli-action@v2
+        with:
+          token: ${{ secrets.TX_TOKEN }}
+          args: push -s --skip
```
TheOneRing, line 19: The generation step is missing

TheOneRing (changes requested):

.github/workflows/push-transifex.yml
TheOneRing: You might also want to remove .github/workflows/run-in-docker.sh

PR #11900, opened 2024-09-25
fix: transifex push/pull

DeepDiver1975:


.github/workflows/translate.yml
```diff
@@ -38,36 +39,20 @@ jobs:
...
+        uses: transifex/cli-action@v2
+        with:
+          token: ${{ secrets.TX_TOKEN }}
+          args: push -s --skip
 
       - name: l10n-pull
-        run: bash .github/workflows/run-in-docker.sh owncloudci/transifex:latest "cd translations && tx pull --force --skip --all"
-
-      - name: l10n-clean
-        run: git checkout translations/.tx/config
```
TheOneRing, line 50: The origin of that was:
```
# tx might apply changes to .tx/config, we don't want that
.PHONY: l10n-clean
l10n-clean:
	git checkout .tx/config
```
but I guess we no longer need that

.github/workflows/translate.yml
```diff
@@ -38,36 +39,20 @@ jobs:
...
-        # we need to use a different docker image for those two, this appears to be a bit tricky with github actions out of the box
-        run:  bash .github/workflows/run-in-docker.sh owncloudci/transifex:latest "cd translations && tx push -s --skip"
+      - name: l10n-push-source
+        uses: transifex/cli-action@v2
+        with:
+          token: ${{ secrets.TX_TOKEN }}
+          args: push -s --skip
 
       - name: l10n-pull
-        run: bash .github/workflows/run-in-docker.sh owncloudci/transifex:latest "cd translations && tx pull --force --skip --all"
```
TheOneRing, line 47: Is .github/workflows/run-in-docker.sh still used?

TheOneRing (approved):

PR #11908, opened 2024-09-26
Fix crash when multiple share dialogs were shown

erikjv:
Fixes: #11673

### DIFF
--- a/src/gui/owncloudgui.cpp
+++ b/src/gui/owncloudgui.cpp
@@ -1038,6 +1038,7 @@ void ownCloudGui::slotShowShareDialog(const QString &sharePath, const QString &l
             Utility::openBrowser(queryUrl, nullptr);
         });
     } else {
+        // oC10 code path
         const auto accountState = folder->accountState();
 
         SyncJournalFileRecord fileRecord;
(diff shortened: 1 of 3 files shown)


src/gui/owncloudgui.cpp
```diff
@@ -1063,35 +1063,21 @@ void ownCloudGui::slotShowShareDialog(const QString &sharePath, const QString &l
             maxSharingPermissions = SharePermission(0);
         }
 
-
-        ShareDialog *w = nullptr;
-        if (_shareDialogs.contains(localPath) && _shareDialogs[localPath]) {
-            qCInfo(lcApplication) << "Raising share dialog" << sharePath << localPath;
-            w = _shareDialogs[localPath];
+        if (_shareDialog && _shareDialog->localPath() == localPath) {
```
TheOneRing, line 1066: I guess we should return in that case, not sure what happens when addModalLegacyDialog is called multiple times with the same widget.

src/gui/owncloudgui.cpp
```diff
@@ -1063,35 +1063,21 @@ void ownCloudGui::slotShowShareDialog(const QString &sharePath, const QString &l
...
         } else {
-            qCInfo(lcApplication) << "Opening share dialog" << sharePath << localPath << maxSharingPermissions;
-            w = new ShareDialog(accountState, folder->webDavUrl(), sharePath, localPath, maxSharingPermissions, startPage, settingsDialog());
-            w->setAttribute(Qt::WA_DeleteOnClose, true);
-
-            _shareDialogs[localPath] = w;
-            connect(w, &QObject::destroyed, this, &ownCloudGui::slotRemoveDestroyedShareDialogs);
+            qCInfo(lcApplication) << "Opening new share dialog" << sharePath << localPath << maxSharingPermissions;
+            if (_shareDialog) {
+                _shareDialog->close();
```
TheOneRing, line 1071: It should also work if we stack them, so should add another one?

erikjv: No, just 1 sharing dialog for the latest. If a user requested one before but never used it to copy the link, then 🤷, apparently it wasn't important.

erikjv:
Note: document that this is oC10 only functionality.

TheOneRing (approved):

Issue #11905, opened 2024-09-25
Proposal: better singed out icon


tbsbdr:
Proposal to have a better to understand singed out icon

# Status Quo
![image](https://github.com/user-attachments/assets/fdc2f41a-2be4-42c9-88a4-2fa8a2ed0410)

# 1. Proposal
![Picture 1-white](https://github.com/user-attachments/assets/42e0dbd6-59fb-44ac-8442-74f16feea9e4)

# 2. Proposal

![Picture 2 white](https://github.com/user-attachments/assets/835b287c-8eed-4b91-9c10-b5e8d6337637)

# 3. Proposal
![Picture 3 white](https://github.com/user-attachments/assets/6906317a-fdea-475c-9d4d-263e41a5b8df)

@TheOneRing @erikjv @dragotin are the proposals improvements and if yes, which would you favour?

TheOneRing:
1. I think that small it looks like "exploding death stars prohibited"
2. maybe
3. hmm wrong message?

tbsbdr:
ok, lets go with 2
https://fontawesome.com/icons/plug-circle-xmark?f=classic&s=solid

PR #11910, opened 2024-09-26
Update offline icon

TheOneRing:
Fixes: #11905

Old:
image
New:
image


TheOneRing:
@tbsbdr the icon appears to be "576x512".... which might cause issues....

tbsbdr:
what should it be? 1:1?

TheOneRing:
> what should it be? 1:1?

Yes. 512x512

tbsbdr:
so do we always need to customize icons taken from font awesome? they all seem to vary their aspect ratio :(
how did we deal eg. with the plug icon in the past? - same problem here
screenshot_003178

TheOneRing:
hmm, I think this might no longer be an issue and had been a legacy condition from the time we used different sized pngs....

erikjv:
I checked (on macOS) and the icon is proportionally scaled (meaning: it doesn't look squashed/stretched in one direction)

Issue #11913, opened 2024-09-27
Alignment for basic auth


TheOneRing:
![image](https://github.com/user-attachments/assets/b60371bb-8d1e-4f98-9e9e-e592e185e754)

saw-jan:
Tested with: `ownCloud 6.0.0.14906-daily20241029 [b056cf]`

![Screenshot from 2024-10-29 14-02-57](https://github.com/user-attachments/assets/d8f2f5f3-64ea-40ea-bbb6-20dc4fab7f88)

PR #11921, opened 2024-10-02
Fix textfield layout on basic auth credentials page

erikjv:
Fixes: #11913

### DIFF
--- a/src/gui/qml/credentials/BasicAuthCredentials.qml
+++ b/src/gui/qml/credentials/BasicAuthCredentials.qml
@@ -30,67 +30,74 @@ Credentials {
         text: credentials.isReadOnlyName ? qsTr("Please enter your password to log in.") : qsTr("Please enter %1 and password to log in.").arg(credentials.userNameLabel)
     }
 
-    Label {
-        text: credentials.userNameLabel
-    }
-    TextField {
-        id: userNameField
-        placeholderText: qsTr("Enter %1").arg(credentials.userNameLabel)
-        horizontalAlignment: TextField.AlignHCenter
-        text: credentials.userName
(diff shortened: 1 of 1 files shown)


src/gui/qml/credentials/BasicAuthCredentials.qml
```diff
@@ -46,6 +46,9 @@ Credentials {
         Keys.onBacktabPressed: {
             widget.parentFocusWidget.focusPrevious();
         }
+
```
TheOneRing, line 49: We are using a layout and must not mix anchors with layouts.

TheOneRing (changes requested):
^

src/gui/qml/credentials/BasicAuthCredentials.qml
erikjv: done

src/gui/qml/credentials/BasicAuthCredentials.qml
```diff
@@ -30,43 +30,51 @@ Credentials {
...
-        Keys.onBacktabPressed: {
-            widget.parentFocusWidget.focusPrevious();
-        }
-    }
+        ColumnLayout {
+            Label {
+                text: credentials.userNameLabel
+            }
+            TextField {
+                id: userNameField
```
TheOneRing, line 43: I think now that we have this internal Panel, with this new layout, you should set fillWidth for the contained elements.
Also I'm pretty sure you should add the buttons, else weird stuff will happen.

erikjv: done

TheOneRing (approved):

Issue #11915, opened 2024-09-27
button alignment oauth screen


DeepDiver1975:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the bug

buttons are miss aligned

### Expected behavior

shall be alligned

### Screenshots

![image](https://github.com/user-attachments/assets/b8d07c28-e8eb-44f0-b441-56ef2fdced8d)

### Client version number

master

erikjv:
I cannot reproduce this on my macOS build. This is probably on Linux, so I'll try that next.

erikjv:
I tried to reproduce this on Neon Linux (Wayland), and for me it looks fine:

<img width="830" alt="Screenshot 2024-10-02 at 16 27 53" src="https://github.com/user-attachments/assets/699a2685-62db-48f2-9c0e-982ed498f838">

erikjv:
What Qt are you using? Or, is this a craft build, or a native Linux build?

DeepDiver1975:
> I tried to reproduce this on Neon Linux (Wayland), and for me it looks fine:

git the 'open browser' button and see the text change ...

saw-jan:
Tested with: `ownCloud 6.0.0.14906-daily20241029 [b056cf]`
1. Open new account wizard
2. Click `open web browser`

![Screenshot from 2024-10-29 13-57-51](https://github.com/user-attachments/assets/4583dedc-98f0-4fb0-bc5f-f2d98b899469)

PR #11923, opened 2024-10-08
Fix button alignment in oauth screen

erikjv:
When an item is in a layout, the width property is ignored, and the implicitWidth is used. So the copy-to-clipboard button's preferredWidth should be bound the implicitWidth property of the open-browser button.

Fixes: #11915

### DIFF
--- a/src/gui/qml/credentials/OAuthCredentials.qml
+++ b/src/gui/qml/credentials/OAuthCredentials.qml
@@ -59,7 +59,7 @@ Credentials {
 
             Button {
                 id: copyToClipboardButton
-                Layout.preferredWidth: openBrowserButton.width
+                Layout.preferredWidth: openBrowserButton.implicitWidth // Note: the width property is ignored in a Layout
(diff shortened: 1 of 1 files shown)


DeepDiver1975:
Kind of works ..... please make sure every element stays in proper shape. THX

Screenshot from 2024-10-09 01-48-33
Screenshot from 2024-10-09 01-48-39

DeepDiver1975 (changes requested):
alignment with sign out button is missing

erikjv:
@TheOneRing As with #11924 : how can I set that on the logout button?

TheOneRing:
> @TheOneRing As with #11924 : how can I set that on the logout button?

reauth

TheOneRing:
And https://github.com/owncloud/client/blob/49ae25f2bc560d21958619f96ddac9c0b706cf3e/src/gui/qml/credentials/Credentials.qml#L38-L59

PR #11924, opened 2024-10-09
Draft: Fix tab order in basic auth login screen

erikjv:
Fixes: #11912


erikjv:
1. not handling the event seems to "drop" it
2. trying to pass it to the parent doesn't seem to work
3. manually handling tab focus fails, because I cannot seem to get hold of the item/button for the logout component

Maybe we should always show the logout button, and use the visibility to show/hide it? Or have a dummy item at the bottom which forwards the focus to the widget when it gets tab focus?

src/gui/qml/credentials/BasicAuthCredentials.qml
```diff
@@ -61,11 +61,16 @@ Credentials {
...
-            // there is no lougout button
-            if (!credentials.isRefresh) {
+            if (loginButton.visible && loginButton.enabled) {
+                loginButton.forceActiveFocus(Qt.TabFocusReason);
+            } else if (credentials.isRefresh) {
+                // with logout button
+                logOutButton.item.logOutButton.forceActiveFocus(Qt.TabFocusReason);
+            } else {
+                // there is no lougout button
                 widget.parentFocusWidget.focusNext();
```
TheOneRing, line 71: shouldn't parentFocusWidget be able to take care?

src/gui/qml/credentials/BasicAuthCredentials.qml
```diff
@@ -61,11 +61,16 @@ Credentials {
...
-                event.accepted = true;
             }
+            event.accepted = true;
```
TheOneRing, line 73: https://doc.qt.io/qt-6/qml-qtquick-keys.html#example-usage

TheOneRing (changes requested):

PR #11930, opened 2024-10-15
Fix tab order in basic auth login screen

erikjv:
Fixes: #11912

### DIFF
--- a/src/gui/qml/credentials/BasicAuthCredentials.qml
+++ b/src/gui/qml/credentials/BasicAuthCredentials.qml
@@ -65,6 +65,8 @@ Credentials {
             if (!credentials.isRefresh) {
                 widget.parentFocusWidget.focusNext();
                 event.accepted = true;
+            } else {
+                event.accepted = false;
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/gui/qml/credentials/BasicAuthCredentials.qml
```diff
@@ -77,7 +79,7 @@ Credentials {
     Button {
         id: loginButton
         Layout.alignment: Qt.AlignHCenter
-        Layout.preferredWidth: passwordField.width
+        Layout.preferredWidth: userNameField.implicitWidth
```
TheOneRing, line 82: Wrong pr?

erikjv:
Title has been updated to reflect that these are 3 commits fixing multiple issues.

src/gui/qml/credentials/BasicAuthCredentials.qml
TheOneRing: As we also have 3 issues for that those also need linking

TheOneRing (dismissed):
Lgtm

DeepDiver1975 (changes requested):
one pr to fix one issue please

erikjv:
It's now split up

TheOneRing (approved):

PR #11927, opened 2024-10-14
[test-only][full-ci] Remove middleware steps

ishabaral:
This PR removes the middleware steps to reduce the test dependencies.

Part of https://github.com/owncloud/client/issues/10432

PR #11941, opened 2024-10-18
[test-only] Fix CI

ishabaral:
This PR fixes #11927, where all tests are not running in CI

### DIFF
--- a/test/gui/shared/scripts/helpers/SyncHelper.py
+++ b/test/gui/shared/scripts/helpers/SyncHelper.py
@@ -2,7 +2,7 @@ import os
 import re
 import sys
 import test
-import urllib
+import urllib.request
 import squish
 
 from helpers.ConfigHelper import get_config, is_linux, is_windows


test/gui/shared/scripts/helpers/SyncHelper.py
```diff
@@ -2,7 +2,7 @@
 import re
 import sys
 import test
-import urllib
+import urllib.request
```
saw-jan, line 5: We might also have to find out why the test pipeline passed despite fatal errors.

saw-jan (approved):
For now, lets merge the PR and run the tests

saw-jan:
I found out that when we run squish test with `--retry 1` and if there's some fatal error during loading some modules then the exit code will become `0` instead of `1` or other
Same fata error fails the ci if not run with `--retry 1`: https://drone.owncloud.com/owncloud/client/19730

Issue #11945, opened 2024-10-21
[QA] Nightly fails due to the OC10 pipeline getting stuck


ishabaral:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the QA issue

CI gets stuck on OC10 pipeline

Builds: [https://drone.owncloud.com/owncloud/client/19727/2/13](https://drone.owncloud.com/owncloud/client/19727/2/13)

```
LookupError: Object '{container={container={container={name='stack' type='QStackedWidget' visible='1' window={name='Settings' type='OCC::SettingsDialog' visible='1'}} name='groupBox' type='QGroupBox' visible='1'} type='OCC::QmlUtils::OCQuickWidget' unnamed='1' visible='1'} id='passwordField' type='TextField' visible='true'}' not found. Could not match properties:\n    type for object name: '{container={container={container={name='stack' type='QStackedWidget' visible='1' window=:settings_OCC_SettingsDialog} name='groupBox' type='QGroupBox' visible='1'} type='OCC::QmlUtils::OCQuickWidget' unnamed='1' visible='1'} id='passwordField' type='TextField' visible='true'}'
```

### Screenshots

![Screenshot from 2024-10-21 15-04-44](https://github.com/user-attachments/assets/1e867861-2747-4d38-8b70-a275a495b542)

### Logs

[clientlog.log](https://github.com/user-attachments/files/17458532/clientlog.log)

saw-jan:
Tests are failing because there is login with browser windows instead of username password.
>CI gets stuck on OC10 pipeline

CI got stuck because the tests were still running due to a long lookup timeout

![Both_original_and_copied_folders_can_be_synced](https://github.com/user-attachments/assets/34dd379d-0a85-49df-acb8-6cba78856b66)

saw-jan:
>Tests are failing because there is login with browser windows instead of username password.

oauth2 app state was not reverted to its initial state after the oauth test scenario.

PR #11946, opened 2024-10-21
[test-only][full-ci] restore oc10 apps to initial state

saw-jan:
Fixes https://github.com/owncloud/client/issues/11945

### DIFF
--- a/test/gui/shared/scripts/pageObjects/EnterPassword.py
+++ b/test/gui/shared/scripts/pageObjects/EnterPassword.py
@@ -71,7 +71,7 @@ class EnterPassword:
 
     def enter_password(self, password):
         squish.waitForObjectExists(
-            self.PASSWORD_BOX, get_config("maxSyncTimeout") * 1000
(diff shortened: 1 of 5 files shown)


test/gui/shared/scripts/pageObjects/EnterPassword.py
```diff
@@ -71,7 +71,7 @@ def get_username(self):
 
     def enter_password(self, password):
         squish.waitForObjectExists(
-            self.PASSWORD_BOX, get_config("maxSyncTimeout") * 1000
+            self.PASSWORD_BOX, get_config("minSyncTimeout") * 1000
```
saw-jan, line 74: this might be the cause for drone timeout. fucntion `waitForObjectExists` fails to get the password object and waits till timeout i.e. 60 seconds. so if that happens in most of the scenarios then the time is more than 1 hour.

saw-jan: now timeout set to only 5 seconds

ishabaral (approved):

PR #11947, opened 2024-10-21
[gui-tests] fail on fatal errors

saw-jan:


saw-jan:
will be fixed in https://github.com/owncloud/client/pull/11953

Issue #11928, opened 2024-10-14
MacOS: button icon size broken on auth dialog


DeepDiver1975:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the bug

icons too big

### Expected behavior

well ...not so big ;-)

### Screenshots

<img width="1129" alt="Screenshot 2024-10-14 at 13 51 12" src="https://github.com/user-attachments/assets/3ed4e900-561c-4c60-8d6a-532b7fd32ee0">

### Client version number

kiteworks-6.0.0.14799-daily20241008-arm64.pkg

ishabaral:
Tested on : ` ownCloud-6.0.0.14906-daily20241029 [b056cf]`

![Screenshot from 2024-10-29 15-18-42](https://github.com/user-attachments/assets/1d80efac-cb36-4928-bb18-1c92bf19f9a7)

PR #11948, opened 2024-10-21
Fix button sizes on high-DPI screens

erikjv:
The DPI rate of pixmaps seems to be ignored by QML, resulting in double-sized icons when shown on a high-DPI screen. This limits the height of the buttons on the oauth login page, so the icon is not excessively large.

Fixes: #11928


erikjv:
Screenshot 2024-10-21 at 14 12 36

TheOneRing (commented):
what about basic auth?

erikjv:
That doesnt have icons so I left it alone

TheOneRing (approved):

PR #11954, opened 2024-10-23
fix: pin tx version and ensure the step is failing if there is an iss…

DeepDiver1975:
…ue with tx


TheOneRing:
Could  you explain why you removed `--skip`

DeepDiver1975:
> Could you explain why you removed `--skip`

[fix: ensure the step is failing if there is an issue with tx](https://github.com/owncloud/client/pull/11954/commits/4c295fb78bf678d6d31ac620ea7ce7bd9a85d5c8)

TheOneRing:
Please explain your reasoning for example you could quote the doc:
```
--skip: Normally, if an upload fails, the client will abort. This may not
be desirable if most uploads are expected to succeed. For example, the reason
of the failed upload may be a syntax error in one of the language files. If
you set the --skip flag and an upload fails, then the client will simply
print a warning and move on to the next language file.
```

"Don't continue on error...."

TheOneRing:
Btw skip was used since forever:
https://github.com/owncloud/client/commit/3356a09d83c5d5639f16c2b2be39e28c753b3df9

DeepDiver1975:
see https://github.com/transifex/cli/blob/e7feea35c13e2d4cd8d72a9097748776320b0f6c/internal/txlib/push.go#L305-L308

If --skip is used the tx command does not abort and as a result the action step succeeds with errors.
removing skip will abort the command and the step will error and we see when stuff is broken

TheOneRing (approved):

Issue #11951, opened 2024-10-22
Qt 6.8: Native tooltips for Navigation bar


TheOneRing:
Since Qt6.8 it supports native popups, however for now it appears that they are only active for menus by default.
However, it can be enabled by setting:
`ToolTip.toolTip.popupType: Popup.Native `

This would fix the issue where the toolbar prevents interaction with the element.
![image](https://github.com/user-attachments/assets/b61e48ec-724b-4f6e-8a4b-f642c644569a)

erikjv:
Note for testing: this is only enabled for Qt 6.8 or later.

PR #11961, opened 2024-10-24
Use native tooltips for Navigation bar

erikjv:
This fixes the issue where a (non-native) tooltip could be shown where the mouse cursor is positioned, which would block any clicks to go through to the button underneath.

Fixes: #11951

### DIFF
--- a/src/gui/qml/AccountBar.qml
+++ b/src/gui/qml/AccountBar.qml
@@ -26,6 +26,12 @@ Pane {
 
     Accessible.name: qsTr("Navigation bar")
 
+    Component.onCompleted: {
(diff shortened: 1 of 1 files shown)


erikjv:
About tooltips showing outside of the "monitor" screen: I cannot reproduce this on macOS 13.5. I'll soon upgrade to the Latest and Greatest version and re-test it.

TheOneRing:
Code looks good, but it would be great if we could do it without bumping the requirements to latest.

erikjv:
On macOS 15.latest the tooltip is not shown "outside" the screen/monitor:

Screenshot 2024-10-29 at 14 31 42

TheOneRing (approved):

PR #11968, opened 2024-10-31
Add desktop branding option to add url suffix during setup

erikjv:


DeepDiver1975:
As per discussion earlier today we will simply replace the url path.Fits the need we have. Enough for now.

@erikjv as discussed: please rebase and let's see if the tests are green then. THX

erikjv:
It is already rebased

DeepDiver1975 (approved):

PR #11971, opened 2024-11-07
Fix AppImageUpdater build

erikjv:


DeepDiver1975:
Can we add a build step to the GitHub workflows to build the app image so that we prevent similar issues in the future?

TheOneRing:
BTW this is wrong

DeepDiver1975:
> BTW this is wrong

brilliant - this is why I want this to be compiled on ci ....

DeepDiver1975 (changes requested):
as per comment from @TheOneRing - wrong ....

erikjv:
> as per comment from @TheOneRing - wrong ....

I removed the method declaration, as I wrote in the new commit message:

The method `handleStartup` was renamed to `validateUpdate`, and is now
handled in the `Updater` base class. The `AppImageUpdater` does not need
to override it.

erikjv:
(I fixed the commit message in the last push)

erikjv:
> Can we add a build step to the GitHub workflows to build the app image so that we prevent similar issues in the future?

I think so, but I have no experience with gitub workflows, so I'll ask Hannah on Monday.

DeepDiver1975:
cherry-picked ci change from https://github.com/owncloud/client/pull/11975 to see it truly compiling it now ...

DeepDiver1975:
compiler issue gone - but the package step is now failing .... Monday :wave:

DeepDiver1975 (approved):

PR #11977, opened 2024-11-11
test: isTextBody

kw-tmueller:


dj4oC (commented):
Moves the free function `isTextBody` (HTTP content-type check used to decide whether to log a request/response body) into the `HttpLogger` namespace as an exported function, and adds a unit test for it.

### Findings
Security: clean — logic byte-identical, no ReDoS risk in the regex, no new external exposure (internal sync-library symbol). Stability: clean — pure refactor; confirmed via repo search the original had internal linkage so there's no ODR/symbol-collision risk from the new export; the one call site was updated consistently; does not touch discovery/propagator/conflict-resolution code, so it does not require human escalation despite living under `src/libsync/`. Performance: clean — same cached `static const QRegularExpression`, no new allocation or call-frequency change. Coverage: gap — the added test only exercises the `application/...json` alternative (2 of the regex's 3 branches, both positive); it doesn't cover the `text/.*?` branch, `application/xml`, `x-www-form-urlencoded`, or any negative case (e.g. `image/png`), so a regression that made the function always return `true` would not be caught. Since this PR's whole purpose is to add test coverage, recommend rounding it out with one case per branch plus a negative case before merge.

### Cross-platform check
Clean — no platform-specific code touched.

### Verdict
Commenting

---

PR #11982, opened 2024-11-14
Fix crash when checking the sync file status

erikjv:
Sometimes a `Folder` is created for a local path that we cannot sync (e.g. it's not writable). This folder will not have a sync engine. If the engine of this `Folder` is accessed, this will lead to a crash.

The fix is to check for the existance of a sync engine in the `Folder::canSync()` method.

Fixes: #11981

### DIFF
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -408,7 +408,7 @@ bool Folder::syncPaused() const
 
 bool Folder::canSync() const
 {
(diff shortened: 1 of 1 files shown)


erikjv:
IMO a `Folder` object without a sync engine is incomplete. And it cannot get a sync engine later on. So I think this folder object should never have been created. But refactoring that is for another PR.

DeepDiver1975 (approved):

PR #11984, opened 2024-11-18
OAuth: Prevent logout when refreshing token

erikjv:
Prevent the user being logged out when the network disappears during OAuth token refresh.


DeepDiver1975 (approved):
Seems to fix the issue -> merge

PR #11991, opened 2024-11-25
Fix possible crash when a sync error early in startup

erikjv:
During startup, sync was enabled before the UI was completely set up. This could lead to a crash when a sync error occurred, which would try to show it in the UI.

Fixes: #11988


DeepDiver1975:
Looks reasonable - THX

PR #11993, opened 2024-11-25
fix: http header keys are case in-sensitive

DeepDiver1975:
qt6.8 variant of #11992


DeepDiver1975:
we are not ready for qt 6.8 due to squish ...

PR #11996, opened 2024-11-26
WIP: Normalize file names before using them as keys

erikjv:


DeepDiver1975:
I am pretty confident that this is the wrong approach -> close

PR #12002, opened 2024-12-04
WIP: Normalize file names before using them as keys

erikjv:


src/gui/folderman.cpp
```diff
@@ -545,6 +545,9 @@ Folder *FolderMan::addFolderInternal(
 Folder *FolderMan::folderForPath(const QString &path, QString *relativePath)
 {
     QString absolutePath = QDir::cleanPath(path) + QLatin1Char('/');
+    if (!Utility::isLinux()) {
```
DeepDiver1975, line 548: Out of curiosity: why not use an ifdef? These are all ifs which never change at runtime.

PR #12005, opened 2024-12-06
OAuth: Prevent logout when refreshing token

DeepDiver1975:
Prevent the user being logged out when the network disappears during OAuth token refresh.


erikjv (approved):

DeepDiver1975:
backport of #11984

PR #12008, opened 2024-12-09
fix: remove unnecessary includes

DeepDiver1975:


DeepDiver1975:
What does that mean?
`/__w/client/client/src/gui/aboutdialog.cpp:80:28: error: invalid use of incomplete type ‘class OCC::ConfigFile’`

erikjv (approved):

PR #12025, opened 2024-12-31
build: use default cache path for squish craft target instead of separate squish path

saw-jan:
By using the following as a cache path for `linux-gcc-x86_64-squish` craft target, we can separately define which Qt version to use for this target and we also omit the necessity of creating separate squish cache path 
```ini
Packager/CacheVersion = ${Variables:ownCloudVersion}/Qt_${linux-gcc-x86_64-squish-BlueprintSettings:libs/qt6.version}_${Variables:CachePatchLvl}
```

This targets the [linux](https://download.owncloud.com/desktop/craft/cache/6.0/Qt_6.7.3_0/linux/) cache folder of defined Qt version for libraries and we no longer require [squish](https://download.owncloud.com/desktop/craft/cache/6.0/Qt_6.7.3_0/squish/) cache folder


saw-jan:
@DeepDiver1975 Could you check if this is valid and doesn't affect anything else?
I test with craft and the the libraries are downloaded from the correct cache path

saw-jan:
this is no longer required.
See https://github.com/owncloud/client/pull/12268

Issue #11883, opened 2024-09-16
[QA] GUI tests sometime fails due to the system keyring dialog


ishabaral:
### Pre-submission Checks

- [X] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [X] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the QA issue

Builds: https://drone.owncloud.com/owncloud/client/19378/6/11

### Screenshots

![Screenshot from 2024-09-16 14-46-22](https://github.com/user-attachments/assets/0ecd00d8-7adc-4cbc-9c75-a7f594f394cc)

![Screenshot from 2024-09-16 14-41-47](https://github.com/user-attachments/assets/e27fca7d-747c-4513-8334-52b9de862a23)

### Logs
 
[clientlog.log](https://github.com/user-attachments/files/17011373/clientlog.log)

saw-jan:
looks like it can happen with oc10 pipeline as well
- https://drone.owncloud.com/owncloud/client/19434/5/16

The problem is with the test environment. So the server type may not be a problem

saw-jan:
Now this appears with oc10 server pipeline :man_facepalming: 
https://drone.owncloud.com/owncloud/client/19563/2/16

saw-jan:
Seen recently: https://drone.owncloud.com/owncloud/client/19871

PrajwolAmatya:
Recent fails:
- https://drone.owncloud.com/owncloud/client/19922
- https://drone.owncloud.com/owncloud/client/19920
- https://drone.owncloud.com/owncloud/client/19918

saw-jan:
yeah, the issue can appear sometime in the CI but not that frequent. So, let's move it to backlog for now.

saw-jan:
This is still occurring these days.

saw-jan:
keyring can be locked during the test execution
https://drone.owncloud.com/owncloud/client/20165/4/13

Trying to unlock during the test also didn't work (implemented in https://github.com/owncloud/client/pull/11866)
Maybe it takes some time to unlock?

saw-jan:
Found out that the `gnome-keyring-daemon -r --unlock` command gets terminated sometimes in the middle of the test.

saw-jan:
Please, reopen the issue if it occurs again

PR #12028, opened 2025-01-07
[gui-tests][full-ci] check keyring before starting client

saw-jan:
Added keyring check before starting client. The check is done with multiple retries until a timeout is reached.

Fixes https://github.com/owncloud/client/issues/11883

THe problem with the mentioned issue was that sometimes the keyring service gets killed or disconnected in the middle of the test execution.


ishabaral (approved):
LGTM

PR #12030, opened 2025-01-08
test: Utility::UnixTimeToLargeIntegerFiletime

DeepDiver1975:
test in preparation for #12027

### DIFF
--- a/test/testutility.cpp
+++ b/test/testutility.cpp
@@ -12,6 +12,9 @@
 
 #include "common/filesystembase.h"
 #include "common/utility.h"
+#ifdef Q_OS_WIN
+#include "common/utility_win.h"
+#endif
 
 using namespace std::chrono_literals;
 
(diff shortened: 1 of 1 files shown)


CookiePLMonster (commented):
Just a recommendation - just in case, it might be worth expanding the test to also test for a `uint32` overflow that'd theoretically happen on 7 February 2106 at 06:28:15 UTC. An additional test for 28 February 2106 sounds optimal to me.

EDIT:
Also, the expected value you're comparing against in this test seems kind of low - shouldn't it be much higher, considering it's in 100ns units from the year 1600? Seems like the correct result should be `137930652000000000LL`:
https://www.wolframalpha.com/input?i=%282148591600+*+10000000%29+%2B+116444736000000000

Is the current value a placeholder?

DeepDiver1975:
> Is the current value a placeholder?

Just a placeholder - I am on Linux and cannot run the test locally. Wanted to see what the LARGE_INTEGER holds :see_no_evil: 

I need a windows machine .... :vomiting_face:

CookiePLMonster:
> Just a placeholder - I am on Linux and cannot run the test locally. Wanted to see what the LARGE_INTEGER holds 🙈

You can then use the result from WolframAlpha as I linked above.

PR #12033, opened 2025-01-15
Add serverPath column to db metadata table

erikjv:
This is a first step to fix issues where file names are normalized differently on the server vs. the client.

### DIFF
--- a/src/common/syncjournaldb.cpp
+++ b/src/common/syncjournaldb.cpp
@@ -543,8 +543,9 @@ bool SyncJournalDb::checkConnect()
         return sqlFail(QStringLiteral("prepare _deleteUploadInfoQuery"), *deleteUploadInfoQuery);
     }
 
-    QByteArray sql("SELECT lastTryEtag, lastTryModtime, retrycount, errorstring, lastTryTime, ignoreDuration, renameTarget, errorCategory, requestId "
-                   "FROM blacklist WHERE path=?1");
+    QByteArray sql(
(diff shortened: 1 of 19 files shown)


src/libsync/syncfileitem.cpp
```diff
@@ -72,6 +72,7 @@ SyncFileItemPtr SyncFileItem::fromSyncJournalFileRecord(const SyncJournalFileRec
     item->_remotePerm = rec._remotePerm;
     item->_serverHasIgnoredFiles = rec._serverHasIgnoredFiles;
     item->_checksumHeader = rec._checksumHeader;
+    item->_remoteName = QString::fromUtf8(rec._serverPath);
```
DeepDiver1975, line 75: Shouldn't we have the same name `remoteName` for DB as well? Just to be consistent ...

src/common/syncjournaldb.cpp
```diff
@@ -806,7 +807,17 @@ bool SyncJournalDb::updateErrorBlacklistTableStructure()
             sqlFail(QStringLiteral("updateBlacklistTableStructure: Add requestId"), query);
             re = false;
         }
-        commitInternal(QStringLiteral("update database structure: add errorCategory col"));
+        commitInternal(QStringLiteral("update database structure: add requestId col"));
+    }
+
+    if (columns.indexOf("remoteName") == -1) {
+        SqlQuery query(_db);
+        query.prepare("ALTER TABLE blacklist ADD COLUMN remoteName VARCHAR(4096);");
```
DeepDiver1975, line 815: Isn't there a central mechanism taking care of db migrations? This feels so wrong

PR #12039, opened 2025-01-23
fix: bypass Qt's QFile::encodeName() in csync

DeepDiver1975:
## ToDo
- [x] extract commits as individual reasonable PRs
- [x] fix failing windows test @erikjv 
- [x] implement test case for nfd/nfc encoded mkpath
- [x] implement test case for nfd/nfc encoded rename

### DIFF
--- a/test/testlocaldiscovery.cpp
+++ b/test/testlocaldiscovery.cpp
@@ -238,6 +238,83 @@ private Q_SLOTS:
         QVERIFY(!fakeFolder.currentRemoteState().find(QStringLiteral("C/.foo")));
         QVERIFY(!fakeFolder.currentRemoteState().find(QStringLiteral("C/bar")));
     }
+
+    void testNameNormalization_data()
+    {
+        QTest::addColumn<QString>("correct");
+        QTest::addColumn<QString>("incorrect");
+
+        const unsigned char a_umlaut_composed_bytes[] = {0xc3, 0xa4, 0x00};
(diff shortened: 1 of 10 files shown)


erikjv:
I moved "Clarify propagator job descriptions" out as a seperate PR: #12050 . We can merge the `fromDisk` fix into the commit where it is implemented. But other than that, I think we should keep this as 1 PR.

test/testlocaldiscovery.cpp
```diff
@@ -238,6 +238,73 @@ private Q_SLOTS:
...
+        const QString a_umlaut_composed = QString::fromUtf8(reinterpret_cast<const char *>(a_umlaut_composed_bytes));
+        const QString a_umlaut_decomposed = a_umlaut_composed.normalized(QString::NormalizationForm_D);
+
+        QTest::newRow("a_umlaut decomposed") << a_umlaut_decomposed << a_umlaut_composed;
+        QTest::newRow("a_umlaut composed") << a_umlaut_composed << a_umlaut_decomposed;
+    }
+
+    // Test that when a file/directory name on the remote is encoded in NFC, the local name is encoded
+    // in the same way, and that a subsequent sync does not change anything. And the same for NFD.
+    void testNameEncoding()
```
DeepDiver1975, line 257: @erikjv Please document what exactly this test is about. Also thinking about that the function name `testNameEncoding()` is not giving that much of an explanation. THX

erikjv: done

PR #12059, opened 2025-02-11
Add more logging on http responses - specifically in error scenarios

erikjv:


src/libsync/httplogger.cpp
```diff
@@ -146,19 +146,24 @@ void HttpLogger::logRequest(QNetworkReply *reply, QNetworkAccessManager::Operati
...
         if (!ctx->send) {
             ctx->send = true;
             ctx->timer.reset();
-        } else {
+        } else if (ctx->lastUrl != reply->url()) {
             // this is a redirect
-            if (ctx->lastUrl != reply->url()) {
-                ctx->addRedirect(reply->url());
-            } else {
-                Q_UNREACHABLE();
```
DeepDiver1975, line 160: good to see this one go - I was hitting this once ... having a client to terminate because of some (not met) condition in a logging method is just wrong.

DeepDiver1975 (changes requested):
As noted in chat: specifically on download failures it would be great to log the number of received bytes.

DeepDiver1975 (approved):

erikjv:
> As noted in chat: specifically on download failures it would be great to log the number of received bytes.

This is done, when the local (file) device is still available.

modSpike (approved):

PR #12062, opened 2025-02-13
fix: multi space systems are now properly detected and handled

DeepDiver1975:


modSpike (commented):
short version: : IMO the getName on the drive should return the correct name. We should not need to futz around with this, especially using the spaces manager. Deviations should be clearly docced else the reviewer actually has to debug the thing just to try to figure out why it's implemented in a strange way. Many people won't bother (rightfully so - just human nature).

the detailed feedback:

IMO the client should never have to figure out where to get the name from when the drive has a "name" property/getter. When I debug this on ocis I see that the "special case" drives already have reasonable names - eg the sharesIdC case is already named "Shares" (we should be able to remove that special case) and the "personal" drive is actually named for the user (in my test case Katherine Johnson). With your proposed change the Katherine space is now called "personal" which I think is a downgrade? 

Honest question: is this really the desired outcome? Are we really supposed to apply a generic name when a real personal name is already available? 

If the real personal name is preferred we can get rid of both of the special ocis related checks and this function just needs to call getName - as it should already do in the most reasonable impl

regarding the check for hasMultiplePersonalSpaces just to invoke getName: "having" to use the account to make a decision about where the name should come from is a powerful code smell as it introduces a dependency that should never be necessary for something so basic. IMO it should be clearly explained/justified in a comment any time a dependency like this is "required".  

I still don't understand the benefit of this check, as if the two "special cases" are not met (they should both be effectively skipped if not ocis, right?) it's going to call getName anyway. I don't know the history of that part of the impl but my question about "why" the default call to getName is not enough is still not clarified. If it's about skipping checks I can appreciate that, but let's look at why those checks are there in the first place and try to refine that instead of just throwing in a dep on the account.

modSpike (commented):
maybe the idea is you never want to hit the "personalC" check because on kiteworks all spaces are "personal"? 

if that is the logic, and if we realllly want to replace the real personal name on ocis with generic "Personal", ok - I fully get it - but in this case the space should have a bool prop named hasManyPersonalSpaces which is set just once in the Space ctr, and queried in this function. problem solved in every way?

DeepDiver1975:
## OCIS
- personal space holds the user's name as name of the personal space
- but we want 'Personal' to be displayed

## KW
- personal spaces holds the name of the folder
- and this is the name we want to display

Is this the way I would design this with all the knowledge at hand now? For sure not.
But this is a living thing and changes over time and needs adjustments.

Instead of using the capabilities endpoint we can think about using a branding option to change this at build time.
Reasonable as more kw specific changes will come ....

erikjv (approved):

PR #12068, opened 2025-02-17
ci: bring back linux build

DeepDiver1975:


DeepDiver1975:
build succeeds now - but AppImage is broken. Used linuxdeploy is too old ...

```
./owncloud-client-HEAD-master-linux-gcc-x86_64.AppImage -s
gui.main: Using "en_GB" translation "en_GB"
gui.main: Could not load qtkeychain translations
/tmp/.mount_owncloaiiiPC/AppRun.wrapped: symbol lookup error: /lib/x86_64-linux-gnu/libEGL_mesa.so.0: undefined symbol: wl_display_create_queue_with_name

PR #12070, opened 2025-02-18
Fix possible crash if the client encountered an error during folder s…

erikjv:
…etup

In that case the engne would return a nullptr.

(cherry picked from commit 1d689e1a70cee3672bfb19eab858b16bb51ca1da)


DeepDiver1975:
CI needs fixing ..... oh lord .....

PR #12089, opened 2025-03-04
ci: fix clang tidy

DeepDiver1975:
image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -159,8 +159,8 @@ jobs:
     - name: Clang tidy
       if: github.event_name != 'pull_request'
       run: |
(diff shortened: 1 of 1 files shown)


modSpike (commented):
I thought the solution was to remove tidy from non-mac builds but it appears tidy is still in play, with some change to - what is that - ensuring the llvm libs are loaded via pip?

a description on the PR would help here

DeepDiver1975:
> I thought the solution was to remove tidy from non-mac builds but it appears tidy is still in play,

after a second look I detected the error: pip was not installed and this is now added via the craft command

DeepDiver1975:
pfff - clang tidy is not executed on pull request .... why?

DeepDiver1975:
> pfff - clang tidy is not executed on pull request .... why?

because it takes ages to finish ....

Issue #12086, opened 2025-02-28
Add .psec to the exclude sync extensions


mmattel:
### Pre-submission Checks

- [x] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [x] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Feature description

Add a new extension to the default exclude sync list.

### Proposed solution

Add the `.psec` extension to the default extension exclude list.
This will prevent files with such an extension from being synced.

### Alternative solutions you considered

Adapt Infinite Scale that files with such an extension will not be processable by the API.

### Additional context

Files with the `psec` extension are necessary for password protected folders.

1. See the description in the PPF's [readme](https://github.com/owncloud/web/tree/master/packages/web-app-password-protected-folders) for details how PPF works and for impacts and possible solutions **inside the webUI**.
1. Because of the way it got implemented, there are some things that can be potential usability issues with likely major bad user experiences.
    1. First there is the deletion issue (PPF Location) which can be mitigated by following the suggested solution.
    2. Second there is the moving/renaming issue which also can be mitigated following the PPF Location suggestion.

    **But** those mitigations only work as long the users stays in the webUI. Atm where sync apps are involved, these mitigations can and - I strongly believe - will be an issue that are a major cause for bad user experience. The referenced readme tells the story and you only need to imagine some PPF usecases plus creative users playing around, read it carefully!

There are two possible solutions, which can also be done in parallel:
- A short term fix that should consistently be added into **all sync apps** by disallowing any sync of files having the psec extension. This would reduce the issue to the webUI in a very short timeframe.
- A longterm and consistent fix in the ocis backend that prevents psync files passing the API except the webUI.

Notes:
* ocis 7.1 will be released in the first week of March (next week), any change will not make it in.
* The next ocis release that will be set productive is min +6m to go
* The next Desktop/Android app release is in March
* An update for iOS can be made quickly

@DeepDiver1975 as dicussed
@kobergj @LukasHirt

PR #12093, opened 2025-03-04
Update the sync exclude list with two extensions relevant for Infinite Scale

mmattel:
Fixes: #12086 (Add .psec to the exclude sync extensions)

There are two extensions that need to be added to the sync exclude list, relevant for Infinite Scale:

* `.psec` which is an extension used for password protected folders.
See the referenced issue for more details
* `.space` this is a directory that is used locally when syncing Spaces.
If you do not exclude it, you get entries for not synced items

Note that `*` and `]` are intentionally based on the definition of excluded items, see the GUI for details.

@kobergj @LukasHirt 

Note that it is a bit courious finding an "exclude" item named `My Saved Places.`, imho this could be removed as no use, but possibly another story...

### DIFF
--- a/sync-exclude.lst
+++ b/sync-exclude.lst
@@ -38,4 +38,8 @@
 *.unison
 .nfs*
 
+# Relevant for Infinite Scale
+].space
+*.psec
+
 My Saved Places.


DeepDiver1975 (changes requested):
Blocking this for now until I fully understood the circumstances - THX

DeepDiver1975 (approved):

PR #12100, opened 2025-03-12
More test framework fixes

erikjv:


src/libsync/httplogger.cpp
```diff
@@ -180,6 +180,7 @@ void HttpLogger::logRequest(QNetworkReply *reply, QNetworkAccessManager::Operati
     QObject::connect(
         reply, &QNetworkReply::finished, reply,
         [reply, ctx = std::move(ctx), logSend] {
+            Q_ASSERT(reply->isFinished()); // Make sure the test framework set the state to finished before emitting the signal
```
DeepDiver1975, line 183: This is fishy to me. This assert should really stay in the test code and not be part of the production code base.

modSpike: I think part of the issue is that in production we would be validating that a real network reply was received, and we want to ensure it's finished before logging. Asserting is probably not the best way to do that, but we can't test this response in the tests, per se, because we have a fake reply in the first place.

at least I don't think this assert was intended to test the tests...but who knows.

test/testutils/syncenginetestutils.cpp
```diff
@@ -778,14 +771,15 @@ void FakeChunkMoveReply::respond()
...
-    Q_EMIT finished();
+    checkedFinished();
 }
 
 void FakeChunkMoveReply::respondPreconditionFailed()
 {
     setAttribute(QNetworkRequest::HttpStatusCodeAttribute, 412);
     setError(InternalServerError, QStringLiteral("Precondition Failed"));
     Q_EMIT metaDataChanged();
+    setFinished(true);
```
modSpike, line 782: shouldn't this be replaced with checkedFinished, too?

src/libsync/httplogger.cpp
erikjv: removed

test/testutils/syncenginetestutils.cpp
erikjv: yes, done

PR #12101, opened 2025-03-13
Make use of "Space" vs. "Folder" strings in the UI themable

erikjv:


erikjv:
Open question is if the method should be named differently

src/gui/qml/FolderDelegate.qml
```diff
@@ -231,7 +231,7 @@ Pane {
 
             Button {
                 id: addSyncButton
-                text: accountSettings.accountState.supportsSpaces ? qsTr("Add Space") : qsTr("Add Folder")
+                text: accountSettings.accountState.supportsSpaces && Theme.spacesAreCalledSpaces ? qsTr("Add Space") : qsTr("Add Folder")
```
DeepDiver1975, line 234: What about calling it the same way as the account state? `Theme.supportsSpaces`

DeepDiver1975 (changes requested):

mmattel:
Hooking myself in just to do the DE translations asap post merge and syncing

DeepDiver1975:
> Hooking myself in just to do the DE translations asap post merge and syncing

no translation change in here - no worries

modSpike:
I think it should be named "spacesAreCalledFolders" since the alternate makes one wonder "what else would they be called?"

also is there already a dep on the theme in this gui or is that new?

DeepDiver1975 (dismissed):

modSpike (approved):

src/gui/qml/FolderDelegate.qml
```diff
@@ -256,7 +256,7 @@ Pane {
                 Layout.fillWidth: true
             }
             Label {
-                text: qsTr("You are synchronizing %1 out of %2 spaces").arg(accountSettings.syncedSpaces).arg(accountSettings.syncedSpaces + accountSettings.unsyncedSpaces)
+                text: (Theme.spacesAreCalledFolders ? qsTr("You are synchronizing %1 out of %2 folder(s)") : qsTr("You are synchronizing %1 out of %2 space(s)")).arg(accountSettings.syncedSpaces).arg(accountSettings.syncedSpaces + accountSettings.unsyncedSpaces)
```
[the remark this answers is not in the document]
modSpike: this has been updated for this case

src/gui/qml/FolderDelegate.qml
erikjv: fixed

PR #12108, opened 2025-03-17
Fix translatable source strings

mmattel:
References: #12105 (Work/numeral translations en us)

While looking in the refernced PR and also open strings for the client in TX, I identified two issues:

1. Avoid the use of personal pronouns like `you`. This makes text harder to translate
2. `Space` (or Spaces) is a proper noun and therefore written with a capital S

This PR fixes these two issues.


DeepDiver1975:
We will work on this in a later point in time. But not now. Thank you for your contribution.

PR #12111, opened 2025-03-19
Work/fix settings page removal

erikjv:


erikjv:
Fixed in a "call" with @modSpike

modSpike (commented):
Lisa todo: add nullptr check here in the associated PR

modSpike (approved):

PR #12118, opened 2025-03-24
fix crash when Url has a non-empty query on adding account

modSpike:
see https://github.com/owncloud/enterprise/issues/7160 for description of crash.

this fixes the crash in the most basic way but I think Erik and/or Thomas want to look at it to see what is going on on the server side.


src/gui/newwizard/states/serverurlsetupwizardstate.cpp
```diff
@@ -131,6 +131,12 @@ void ServerUrlSetupWizardState::evaluatePage()
                     }
 
                     const auto resolvedUrl = resolveJob->result().toUrl();
+                    if (resolvedUrl.hasQuery()) {
+                        QString errorMsg = tr("The requested URL failed with query value: ");
```
erikjv, line 135: `QString errorMsg = tr("message: %1").arg(resolvedUrl.query());`

erikjv (changes requested):

erikjv (approved):

Issue #12098, opened 2025-03-12
Assert in httplogger


DeepDiver1975:
```
canceled (QNetworkReply::OperationCanceledError, 2), request-id: 50b10c3f-d805-45fd-b678-df3489e0f2c8
25-03-12 09:40:49:770 [ fatal default ]:	ASSERT: "dynamic_cast<QBuffer *>(device)" in file /home/deepdiver/Development/kiteworks/client2/src/libsync/httplogger.cpp, line 114

modSpike:
steps to reproduce?

PR #12120, opened 2025-03-25
Remove invalid assert in the HTTP logger

erikjv:
The assumption that the content "device" is a `QBuffer` when it is closed, is wrong: it can very well be a `QNetworkReply`, plus that shouldn't matter just for logging.

Fixes: #12098


erikjv:
As an example: on line 69, a case is done to `QNetworkReply` (which in turn is not a `QBuffer`).

DeepDiver1975 (approved):

PR #12122, opened 2025-04-01
Fix Space/Folder naming

erikjv:
There is no need for complex HTML here.

### DIFF
--- a/src/gui/folderwizard/spacespage.cpp
+++ b/src/gui/folderwizard/spacespage.cpp
@@ -14,6 +14,8 @@
 #include "spacespage.h"
 #include "ui_spacespage.h"
 
+#include "theme.h"
+
 using namespace OCC;
 
 SpacesPage::SpacesPage(AccountPtr acc, QWidget *parent)
@@ -23,6 +25,8 @@ SpacesPage::SpacesPage(AccountPtr acc, QWidget *parent)
     ui->setupUi(this);
 
     ui->widget->setAccount(acc);
+    ui->label->setText(
+        Theme::instance()->spacesAreCalledFolders() ? tr("Select a folder to sync it to your computer.") : tr("Select a Space to sync it to your computer."));
 
     connect(ui->widget, &Spaces::SpacesBrowser::currentSpaceChanged, this, &QWizardPage::completeChanged);
 }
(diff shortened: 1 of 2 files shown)


modSpike:
as discussed in the meeting, the real issue is that this still uses the word "Spaces" in the gui, and for kiteworks it should be "Folders" (using that new function in the Theme::spacesAreFolders" or whatever it was)

also maybe of interest: I think the translation for that text relied on it being html but not sure. If this needs new translation for plan text vs html...

src/gui/folderwizard/spacespage.cpp
```diff
@@ -23,6 +25,8 @@ SpacesPage::SpacesPage(AccountPtr acc, QWidget *parent)
     ui->setupUi(this);
 
     ui->widget->setAccount(acc);
+    ui->label->setText(
+        Theme::instance()->spacesAreCalledFolders() ? tr("Select a folder to sync it to your computer.") : tr("Select a Space to sync it to your computer."));
```
modSpike, line 29: probably best to make it "Select a Folder to...." to match cap scheme on the eqivalent spaces string. I don't know why Spaces is capped but we should be consistent at least

modSpike (changes requested):

src/gui/folderwizard/spacespage.cpp
modSpike: No, "folder" is on purpose. The reason for a capital S in Space is because it's a bit of branding/blah. - that is from Erik who will put it here next time

modSpike (approved):

src/gui/folderwizard/spacespage.ui
```diff
@@ -14,7 +14,10 @@
    <item>
     <widget class="QLabel" name="label">
      <property name="text">
-      <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Select a Space to sync it to your computer.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
+      <string notr="true">placeholder text</string>
```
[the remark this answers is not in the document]
erikjv: It's clearer when looking at the UI in Designer what's going on here. It's not translated anyway, so no problem.

DeepDiver1975 (approved):

PR #12123, opened 2025-04-01
fix: handle unexpected server response when resolving server url

DeepDiver1975:


modSpike:
this is a proposed better fix for this issue: https://github.com/owncloud/enterprise/issues/7160

to properly test this fix we need to remove the client permissions on a test kiteworks server to arrive at the same conditions that caused the original crash.

PR #12125, opened 2025-04-03
Deprecate bandwidth limits

erikjv:
Implement deprecation of bandwidth limits:
- when none are set, hide the UI
- when a limit is set, show the UI, including a warning icon and a text label warning about the deprecation

### DIFF
--- a/src/gui/networksettings.cpp
+++ b/src/gui/networksettings.cpp
@@ -101,6 +101,14 @@ NetworkSettings::NetworkSettings(QWidget *parent)
     checkAccountLocalhost();
 
     connect(_ui->pauseSyncWhenMeteredCheckbox, &QAbstractButton::clicked, this, &NetworkSettings::saveMeteredSettings);
+
+    ConfigFile config;
+    if (config.useUploadLimit() == 0 && config.useDownloadLimit() == 0) {
+        _ui->warningIcon->setVisible(false);
+        _ui->warningLabel->setVisible(false);
+        _ui->uploadBox->setVisible(false);
+        _ui->downloadBox->setVisible(false);
+    }
 }
 
 NetworkSettings::~NetworkSettings()
--- a/src/gui/networksettings.ui
+++ b/src/gui/networksettings.ui
@@ -227,6 +227,53 @@
     </widget>
    </item>
    <item>
+    <layout class="QHBoxLayout" name="horizontalLayout_7">
+     <item>
+      <widget class="QLabel" name="warningIcon">
+       <property name="sizePolicy">
+        <sizepolicy hsizetype="Fixed" vsizetype="Fixed">
+         <horstretch>0</horstretch>
+         <verstretch>0</verstretch>
+        </sizepolicy>
+       </property>
+       <property name="minimumSize">
+        <size>
+         <width>32</width>
+         <height>32</height>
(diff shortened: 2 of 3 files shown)


erikjv:
@mmattel Probably needs documentation changes.

DeepDiver1975:
@erikjv GUI tests don't like this ...
@saw-jan maybe you can help out here? THX

modSpike:
I propose that any PRs that change look or behavior of gui include screen shots, ideally of before AND after, but definitely for the after. 

so in this case a screen shot of what it looks like in the first scenario and in the second scenario. This will assist review and qa (when they need to make updates), certainly

modSpike (commented):
please add screen shots to the PR description that show the gui in each case described here (related to values in config)

erikjv:
When limits were set before upgrading:

Screenshot 2025-04-03 at 15 45 02

When the limits are removed, or none where there to begin with, that part of the UI is not there:

Screenshot 2025-04-03 at 18 54 27

Keys checked: `BWLimit/useUploadLimit` and `BWLimit/useDownloadLimit`: a value of zero (or absent) means no limit.

erikjv:
> I propose that any PRs that change look or behavior of gui include screen shots, ideally of before AND after, but definitely for the after.

This was already the agreement. My fault for being too "fast".

modSpike (approved):

src/gui/networksettings.cpp
```diff
@@ -101,6 +101,14 @@ NetworkSettings::NetworkSettings(QWidget *parent)
     checkAccountLocalhost();
 
     connect(_ui->pauseSyncWhenMeteredCheckbox, &QAbstractButton::clicked, this, &NetworkSettings::saveMeteredSettings);
+
+    ConfigFile config;
+    if (config.useUploadLimit() == 0 && config.useDownloadLimit() == 0) {
```
modSpike, line 106: ah wait! in the gui it looks like it might be possible to set ONE of these but not the other - if so should this really be an && or would an || be more appropriate?

saw-jan:
> @erikjv GUI tests don't like this ...
> @saw-jan maybe you can help out here? THX

Yeah, the failure is expected since there is change in the settings UI.
This is the failing test step (checks for the last two settings - `Download Bandwidth` and `Upload Bandwidth`):
https://github.com/owncloud/client/blob/0fc7ddc2e0ab89e9a2dfd2ac571d75b36130acda/test/gui/tst_checkAlltabs/test.feature#L30-L33

Either they are gone or the selector has been changed.
@ishabaral Could you take care of it? you need to either delete the checks for last two settings or adjust the selector if they are present.

DeepDiver1975:
> you need to either delete the checks for last two settings

This is just fine - no need to test this feature anymore - we will drop it. Thank you!

saw-jan:
Drone-ci failed to clone even after I updated the branch.
@DeepDiver1975 @erikjv Could you give it a look?

DeepDiver1975:
rebased - let's see ...

saw-jan:
> rebased - let's see ...

looks like a drone-ci issue. similar for ocis ci https://drone.owncloud.com/owncloud/ocis/44917

DeepDiver1975 (approved):

Issue #12136, opened 2025-04-10
[QA] refactor tests to use real config files


modSpike:
### Pre-submission Checks

- [x] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [x] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the QA issue

as discussed today regarding test failure for PR https://github.com/owncloud/client/pull/12081 using a "fake" or otherwise incomplete config file to bootstrap tests is not ideal so we would like to brainstorm a new approach.

### Steps to reproduce the issue

test failure on https://github.com/owncloud/client/pull/12081 is directly related to the missing journalPath setting for the folder in the "fake"/bootstrapping config file. A config generated by the app always includes a value for this setting (as well as all the others)

Important note: I have added a safety check back into the code to handle this situation so the test should now pass (fingers crossed), but we need to remove that as soon as possible because this safety check should never be hit with a normal, complete config file.

### Screenshots

![Image](https://github.com/user-attachments/assets/402d58d8-dafb-48a2-bc76-cf9533ee09d6)

modSpike:
important dev note: when this issue is complete, search code for refs to this issue id and fix the code!

saw-jan:
The agreed solution is to use the full account setup step in all the test scenarios:
https://github.com/owncloud/client/blob/3d1a2d5c8e02051fc13454586cd24f1241ae223b/test/gui/tst_addAccount/test.feature#L23-L27

and remove the use of these steps and related codes or refactor them to use the account setup dialog instead of generated config file:
https://github.com/owncloud/client/blob/3d1a2d5c8e02051fc13454586cd24f1241ae223b/test/gui/tst_activity/test.feature#L11-L13
https://github.com/owncloud/client/blob/3d1a2d5c8e02051fc13454586cd24f1241ae223b/test/gui/tst_activity/test.feature#L25

modSpike:
From our side, Erik had the thought that you could do a snapshot of the env to match the config after running the account setup gui(s). I also think that is a good approach - you could do the snapshot after first test (setting up account and syncs) succeeds then use that snapshot to set up the "clean" state (when required) for each subsequent test in the series. 

[...]

PR #12137, opened 2025-04-11
[test-only][full-ci] Refactor tests to use normal account setup flow

ishabaral:
This PR uses the full account setup in all the test scenarios and removes the use of incomplete config file when setting up account.

Fixes https://github.com/owncloud/client/issues/12136

### DIFF
--- a/test/gui/shared/steps/account_context.py
+++ b/test/gui/shared/steps/account_context.py
@@ -5,17 +5,14 @@ from pageObjects.Toolbar import Toolbar
 from pageObjects.AccountSetting import AccountSetting
 
 from helpers.SetupClientHelper import (
-    setup_client,
     start_client,
     substitute_inline_codes,
     get_client_details,
-    generate_account_config,
     get_resource_path,
 )
 from helpers.UserHelper import get_displayname_for_user, get_password_for_user
 from helpers.SyncHelper import (
     wait_for_initial_sync_to_complete,
-    listen_sync_status_for_item,
 )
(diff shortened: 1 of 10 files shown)


test/gui/Makefile
```diff
@@ -1,3 +1,6 @@
+# Automatically detect the latest Squish version directory
+SQUISH_DIR := $(shell ls -d $$HOME/squish-for-qt-* 2>/dev/null | sort -V | tail -1)
```
saw-jan, line 2: It might not work for all setups. squish installation path can be different from this.
```suggestion
SQUISH_DIR := $(shell ls -d $$HOME/squish-for-qt-* 2>/dev/null | sort -V | tail -1)
```

test/gui/shared/steps/account_context.py
```diff
@@ -97,22 +87,29 @@ def step(context):
...
     AccountConnectionWizard.add_account(account_details)
     # wait for files to sync
     wait_for_initial_sync_to_complete(get_resource_path('/', account_details['user']))
 
 
 @Given('the user has entered the following account information:')
 def step(context):
-    account_details = get_client_details(context)
+    account_details = get_client_details(context.table)
+    print(account_details)
```
PrajwolAmatya, line 112:
```suggestion
```

PrajwolAmatya (dismissed):

PrajwolAmatya (approved):

modSpike:
Wonderful! Thank you so much

PR #12143, opened 2025-04-16
Remove lingering migration impls (especially SpaceMigration) 

modSpike:
we need to remove all outdated migration impls. 

So far the only one left that I know of is the space migration routine. Remove that class and the associated tests.

also sneak in some extra cleanup: move verbose refactoring todo descriptions to confluence, add some important doc info for Utility::setupFavLink


DeepDiver1975 (dismissed):
:dancers:

modSpike:
@erikjv can you please have an extra special look at the result I ended up with in AccountState.cpp

it's weird to my eye. But the whole class is weird so...

DeepDiver1975:
from my understanding the whole block can go

```
    if (Q_UNLIKELY(Theme::instance()->enableCernBranding())) {
        if (status == ConnectionValidator::Connected) {
            setState(Connected);
            return;
            }
    }

modSpike:
yep, Thomas is right - the setState gets called eventually down in the switch so that whole block is now gone.

erikjv (approved):

Issue #13, opened 2012-10-23
Add configured folders to explorer/finder


danimo:
### Windows

To show a folder in  the favorites of Explorer. there is `Links` the users profile (to be obtained with `SHGetSpecialFolderPath(0, path, CSIDL_PROFILE, FALSE)`); We need to put a lnk file there.
### Mac OS

Finder has "Places". [Stack Overflow](http://stackoverflow.com/questions/10751199/how-to-get-finder-sidebar-favorites-content-cocoa) has details on how to implement it with Carbon API.
### Nautilus (GNOME/Unity)

Amending [.gtk-bookmarks](http://darklaunch.com/2010/10/31/where-are-nautilus-bookmarks-stored-nautilus-bookmark-location) seems to do it.
### KDE

[KDE Userbase](http://userbase.kde.org/Dolphin/File_Management#Three_kinds_of_bookmark) explains the use of XBEL bookmarks for Dolphin and other KDE apps.
## Undo

In case the folder is removed from sync, users might be afraid that data is lost., so better ask if the shortcut should be removed again.
## Implementation status:
- [x] Windows Explorer
- [x] OS X Finder
- [x] Nautilus
- [ ] KDE

PR #12146, opened 2025-04-17
[DC-30] fix: memory leak and general cleanup of AccountSettings

modSpike:
see item #13 in the refactoring todo list

short story is: we almost always create a new ConnectionValidator when checkConnectivity is run. Fun part is that in addition to being called explicitly here and there, it also auto-runs in a polling pattern. In some cases the pointer is cleaned up but what I see when I break on the "new" for that ptr is that it's only null the very first time I hit it (eg when creating a new account)


modSpike:
added mystery:

we do have a check earlier in the code that should just return/skip creation of a new validator if the current validator is not null.

for completely unknown reasons the debugger only very rarely ends up returning via this trap, so it very frequently goes on to create a new validator *even though* it appears to be a non-null pointer already. Yes, indeed Erik saw this too - I'm not crazy - but the behavior is completely inexplicable.

New tack will be to replace the already low value use of qpointer with just a normal raw pointer so we don't have to worry about some subtle weirdness related to that. 

and I will also start refactoring this function to JUST DO ONE THING instead of this, that and the other thing (including not only creating a new validator, but also adding the connections with plenty of lambdas inside lambdas)

fingers crossed that once it's simplified it behaves in a more reasonable way.

src/gui/accountstate.cpp
```diff
@@ -54,117 +53,46 @@ namespace OCC {
...
     }
 
-    _updateUrlDialog = UpdateUrlDialog::fromAccount(_account, newUrl, ocApp()->gui()->settingsDialog());
+    // todo: fix this so we can create it on stack or heap. this is gross.
+    UpdateUrlDialog *updateUrlDialog = UpdateUrlDialog::fromAccount(_account, newUrl, ocApp()->gui()->settingsDialog());
 
-    connect(_updateUrlDialog, &UpdateUrlDialog::accepted, this, [=]() {
+    ownCloudGui::raise();
+    // I see no reason to use open here - exec is also modal but it's blocking until the user makes a choice. thus we
+    // don't need to create a slot or any other nonsense -> just get the job done!
```
erikjv, line 69: Doesn't this block running jobs for other accounts?

src/gui/accountstate.cpp
```diff
@@ -441,11 +368,111 @@ void AccountState::checkConnectivity(bool blockJobs)
...
+    // only the status is set to `Connecting`, but also makes the UI show that syncing is paused.
+    // todo: #11, #12
+    // todo: decide if the value to checkConnectivity should depend on value of isCaptive
+    // checkConnectivity does different things depending on whether behindCaptivePortal is true, including blocking the queue guard :/
+    checkConnectivity();
+}
+
 void AccountState::slotConnectionValidatorResult(ConnectionValidator::Status status, const QStringList &errors)
 {
+    // the ConnectionValidator used to delete itself which was really not healthy.
```
erikjv, line 472: I don't think we need this comment.

src/gui/accountstate.cpp
modSpike: I was not able to test that as there is no reasonable way to trigger this. Based on Thomas's discussion with uppers, the whole handling is going away so is this really something you want me to revert to the old ick?

src/gui/accountstate.cpp
modSpike: I disagree - to the extent that until this is fully refactored, having some interim explanations is useful to avoid spelunking (which might also not fully reveal intent)

modSpike:
hello @saw-jan and @ishabaral I would like to ask for help with the test failure. since my last "real" commit (which passed) it has failed twice, once on a test related to syncing a file with special characters in the name, and once on a file with a very long name (but in that case the test with special chars passed). I'm really confused about what is actually happening there. I also see what look like some connection failures but not sure if those are expected or indicate some other problem.

any info is appreciated.

saw-jan:
> hello @saw-jan and @ishabaral I would like to ask for help with the test failure. since my last "real" commit (which passed) it has failed twice, once on a test related to syncing a file with special characters in the name, and once on a file with a very long name (but in that case the test with special chars passed). I'm really confused about what is actually happening there. I also see what look like some connection failures but not sure if those are expected or indicate some other problem.
> 
> any info is appreciated.

it was a flaky failure. I have restarted the build and it's green now.
we will look at how to harden those tests (might be flaky in sync check). THank you!

src/gui/accountstate.cpp
```diff
@@ -336,30 +297,25 @@ void AccountState::tagLastSuccessfullETagRequest(const QDateTime &tp)
 
 void AccountState::checkConnectivity(bool blockJobs)
 {
+    //>>>>> These are all pre-checks that may effectively cancel the check
```
DeepDiver1975, line 300: we should not use `<<<<<` or `>>>>>>` ... this is used for declaring conflicts in git - this triggers me ;-)

src/gui/owncloudgui.cpp
```diff
@@ -798,10 +798,13 @@ void ownCloudGui::runNewAccountWizard()
                     _settingsDialog->setCurrentAccount(accountStatePtr->account().data());
 
                     // ensure we are connected and fetch the capabilities
+                    // refactoring todo: WHAT is this doing here? we have a "legit" account state pointer. it has a checkConnection
+                    // that can do this job
                     auto validator = new ConnectionValidator(accountStatePtr->account(), accountStatePtr->account().data());
 
+                    // Refactoring todo: ARE YOU FREAKING KIDDING ME?!
```
DeepDiver1975, line 805: remove comments like that please - this is as unprofessional - THX

erikjv (approved):

PR #12155, opened 2025-05-02
[DC-28] fix: managed account button popup menu on Gnome

erikjv:


DeepDiver1975 (changes requested):
Still not working on Gnome 48 on Debian trixie + wayland

DeepDiver1975:
turned out to be an issue with qt 6.8.0

PR #12160, opened 2025-05-13
[tests-only][full-ci] fix remove account connection for windows

PrajwolAmatya:
### Desctiption
The accounts tab in the desktop client does not contain the display name. In windows during clean up of users, account connection need to be removed, and the selector used the display name which resulted in failure to remove the users. This PR fixes this problem in windows.

### DIFF
--- a/test/gui/shared/scripts/bdd_hooks.py
+++ b/test/gui/shared/scripts/bdd_hooks.py
@@ -189,9 +189,10 @@ def teardown_client():
         # so to work around that, remove the account connection
         close_dialogs()
         close_widgets()
-        accounts, _ = Toolbar.get_accounts()
-        for account in accounts:
-            Toolbar.open_account(account["displayname"])
+        accounts, selectors = Toolbar.get_accounts()
+        for display_name in selectors.values():
+            _, account_objects = Toolbar.get_accounts()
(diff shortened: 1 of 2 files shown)


test/gui/shared/scripts/bdd_hooks.py
```diff
@@ -189,9 +189,9 @@ def teardown_client():
         # so to work around that, remove the account connection
         close_dialogs()
         close_widgets()
-        accounts, _ = Toolbar.get_accounts()
-        for account in accounts:
-            Toolbar.open_account(account["displayname"])
+        accounts, selectors = Toolbar.get_accounts()
```
saw-jan, line 192:
```suggestion
        _, selectors = Toolbar.get_accounts()
```

test/gui/shared/scripts/bdd_hooks.py
```diff
@@ -189,9 +189,10 @@ def teardown_client():
...
+        for display_name in selectors.values():
+            _, selectors_obj = Toolbar.get_accounts()
```
saw-jan, line 194:
```suggestion
            _, account_objects = Toolbar.get_accounts()
```

saw-jan (approved):

Issue #12165, opened 2025-05-26
[QA] Use Playwright with Python to write E2E tests


PrajwolAmatya:
### Describe the QA issue

Rewrite the existing E2E test related to user login using Python. This will reduce the dependency of using Javascript for the E2E test as the GUI tests is written in Python. This will also make our test stack to be in single language.

https://playwright.dev/python/docs/intro

- [ ] Rewrite existing E2E test using Playwright with Python https://github.com/owncloud/client/pull/12170

PR #12170, opened 2025-05-28
[tests-only][full-ci] write login related e2e tests in python

PrajwolAmatya:
### Desctiption
This PR update the login related E2E test, rewritten in python language. This makes all the test code to be in same language and doesn't require the node dependencies just for the login.

closes: https://github.com/owncloud/client/issues/12165


saw-jan (approved):
:sparkles:

saw-jan (approved):

PR #12176, opened 2025-06-03
[DC-54] feat: drop oc10 specific chunked upload operations

DeepDiver1975:
Supporting OCIS and Kiteworks PDN we are dropping chunked upload capabilities from the desktop client.


src/libsync/owncloudpropagator.cpp
```diff
@@ -344,26 +344,21 @@ PropagateItemJob *OwncloudPropagator::createJob(const SyncFileItemPtr &item)
...
+    case CSYNC_INSTRUCTION_SYNC: {
         if (item->_direction != SyncFileItem::Up) {
             auto job = new PropagateDownloadFile(this, item);
             job->setDeleteExistingFolder(deleteExisting);
             return job;
+        }
+        PropagateUploadFileCommon *job = nullptr;
+        if (account()->capabilities().tusSupport().isValid()) {
+            job = new PropagateUploadFileTUS(this, item);
         } else {
```
modSpike, line 356: I think this is easier to read with the "else" left in place

src/libsync/propagateupload.h
```diff
@@ -245,37 +243,15 @@ private Q_SLOTS:
...
-
-    qint64 chunkSize() const {
-        // Old chunking does not use dynamic chunking algorithm, and does not adjust the chunk size respectively,
-        // thus this value should be used as the one classifying item to be chunked
-        return propagator()->syncOptions()._initialChunkSize;
-    }
-
 public:
-    PropagateUploadFileV1(OwncloudPropagator *propagator, const SyncFileItemPtr &item)
+    PropagateUploadFile(OwncloudPropagator *propagator, const SyncFileItemPtr &item)
```
modSpike, line 254: why is this still munged in with other classes? what is the org here - you have the decl mixed with others in one header and the impl in its own cpp?

src/libsync/propagateuploadv1.cpp
```diff
@@ -137,49 +85,18 @@ void PropagateUploadFileV1::startNextChunk()
     }
 
     // job takes ownership of device via a QScopedPointer. Job deletes itself when finishing
```
modSpike, line 87: that is suspicious. Please validate that the device is not a QObject, and if it is that it does not have a parent. If it is a QObject, the better solution is to just set the parent of the device to the job so it is automatically cleaned up when the job destructs. It is never straightforward to QObject memory management with smart pointers, and imo we should avoid it outside of exceptional cases.

src/libsync/propagateuploadv1.cpp
```diff
@@ -238,40 +154,11 @@ void PropagateUploadFileV1::slotPutFinished()
...
-        pi._errorCount = 0; // successful chunk upload resets
-        pi._contentChecksum = _item->_checksumHeader;
-        pi._size = _item->_size;
-        propagator()->_journal->setUploadInfo(_item->_file, pi);
-        propagator()->_journal->commit(QStringLiteral("Upload info"));
-        startNextChunk();
+        done(SyncFileItem::NormalError, tr("The server did not acknowledge the last chunk. (No e-tag was present)"));
         return;
     }
     // the following code only happens after all chunks were uploaded.
```
modSpike, line 164: looks like a dead comment

modSpike: oh and that error message on line 161 is now wrong?

src/libsync/configfile.h
```diff
@@ -68,8 +67,6 @@ class OWNCLOUDSYNC_EXPORT ConfigFile
 
     QString defaultConnection() const;
 
-    bool passwordStorageAllowed(const QString &connection = QString());
```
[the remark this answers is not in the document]
DeepDiver1975: nothing - but it is an unused function - no implementation anywhere -> kill :skull:

src/libsync/propagateuploadv1.cpp
```diff
@@ -32,7 +32,7 @@
 
 namespace OCC {
 
-void PropagateUploadFileV1::doStartUpload()
+void PropagateUploadFile::doStartUpload()
```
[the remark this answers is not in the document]
DeepDiver1975: agreed

src/libsync/propagateupload.h
modSpike: this still needs to be addressed: each class should go in its own header + cpp. As far as I can tell the cpp is there, but this class decl. still needs to be split out into it's own header.

src/libsync/propagateuploadv1.cpp
modSpike: how was this resolved?

src/libsync/propagateupload.h
modSpike: the comments are out of order but to your response: yes you should fix this in this PR

modSpike (dismissed):

modSpike (approved):
even better now!

PR #12179, opened 2025-06-05
feat: use conan and ownCloud artifactory

DeepDiver1975:


.conan/profiles/macos-latest-armv8.conanprofile
```diff
@@ -0,0 +1,9 @@
+[settings]
+arch=armv8
```
erikjv, line 2: you need another mac config with x86_64

PR #12194, opened 2025-06-19
[tests-only][full-ci] fix tests failing in windows

PrajwolAmatya:
### Description
Fixes the tests that are failing in windows.


test/gui/shared/steps/account_context.py
```diff
@@ -311,6 +313,8 @@ def step(context, displayname):
 )
 def step(context, sync_path, wizard):
     sync_path = substitute_inline_codes(sync_path)
+    if is_windows():
+        sync_path = sync_path.replace('/', '\\')
```
saw-jan, line 317: can we do this inside `substitute_inline_codes` function?

saw-jan (approved):

PR #12197, opened 2025-06-23
[tests-only][full-ci]remove account removal related code from hook

PrajwolAmatya:
# Description
Removed account removal related code from after hook as the clean up works fine without these codes as well.


PrajwolAmatya:
This changes does not remove users from the windows registry. Closing this PR for now.

PR #12213, opened 2025-07-10
[DC-79] fix: webfinger user info urls could not be retrieved from some ocis configurations

modSpike:
see https://kiteworks.atlassian.net/browse/DC-79


erikjv (approved):

DeepDiver1975 (approved):
Tested with demo.owncloud.omc and infinite.owncloud.com

PR #12214, opened 2025-07-11
[DC-83] Fix: remove any existing ResourceCache directories for accounts

erikjv:
Some versions of the client failed to remove the temporary cache
directory on exit, and after a crash the directory would not be removed
either. This can result in multiple GB of cached space images. So clean
this up before creating a new cache.


modSpike (changes requested):
please add a link to the corresponding jira issue!

PR #12236, opened 2025-07-28
Corrected error in translation

drlellinger:


DeepDiver1975:
Thank you!
Master branch is not yet up for translation.
We will first release branch 6.
Finally: translations are maintained on transifex.

PR #12252, opened 2025-08-19
[DC-104] fix: new account should always be selected automatically

modSpike:
see https://kiteworks.atlassian.net/browse/DC-104


DeepDiver1975 (approved):
Squish tests are sucessful now with this change. :+1: 

https://drone.owncloud.com/owncloud/client/21621

PR #12254, opened 2025-08-20
[DC-81] Fix leak of `AccountState`

erikjv:


erikjv:
After merging, we need to cherry-pick this to the 6 branch

PR #12255, opened 2025-08-20
Remove `AccountStatePtr`

erikjv:
This typedef was a left-over from long ago. If a class stores the `AccountState` as a member field, it is responsible for wrapping it in a `QPointer`.


modSpike (approved):
good start.

PR #12280, opened 2025-09-01
[DC-125] Fix some accessibility issues in new account wizard

erikjv:
Fixed:
- URL page: name and description for the server address field
- OAuth page: override URL label name, so screen readers will not read the full URL, also add a descriptive text to use the copy-url button
- Advanced settings page: add a name to the directory text-edit field


src/gui/newaccountwizard/oauthpagecontroller.cpp
```diff
@@ -71,7 +71,8 @@ void OAuthPageController::buildPage()
     _urlField = new QLabel(_page);
     _urlField->setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Fixed);
     _urlField->setEnabled(false);
-    _urlField->setAccessibleDescription(tr("Login URL"));
+    _urlField->setAccessibleName(tr("Login URL")); // Prevent screen readers from reading the full URL!!!
```
modSpike, line 74: sorry my bad: we are exclusively using "Sign in", not "Login" wrt that activity :/

modSpike (changes requested):
see the comment: we have to use "sign in", not "login"

src/gui/newaccountwizard/oauthpagecontroller.cpp
erikjv: done

PR #12290, opened 2025-09-03
[DC-111] Actively delete version keys

erikjv:
These keys were not written out anymore, but would still stay in the configuration file if no configuration change was made. Now they are removed when detected.

### DIFF
--- a/src/gui/accountmanager.cpp
+++ b/src/gui/accountmanager.cpp
@@ -235,7 +235,7 @@ const QList<AccountState *> AccountManager::accounts() const
 
 AccountPtr AccountManager::loadAccountHelper(QSettings &settings)
 {
-    if (!settings.value("version").isNull()) {
+    if (settings.contains("version")) {
         // Migration from pre-7.0:
         settings.remove("version");
     }
--- a/src/gui/accountmanager.cpp
+++ b/src/gui/accountmanager.cpp
@@ -97,7 +97,7 @@ bool AccountManager::restore()
         return false;
     }
 
-    if (!settings->contains("version")) {
+    if (settings->contains("version")) {
         // Migration from pre-7.0:
         settings->remove("version");
     }
--- a/src/gui/accountmanager.cpp
+++ b/src/gui/accountmanager.cpp
@@ -97,6 +97,11 @@ bool AccountManager::restore()
         return false;
     }
 
+    if (!settings->value("version").isNull()) {
+        // Migration from pre-7.0:
+        settings->remove("version");
+    }
+
     // If there are no accounts, check the old format.
     const auto &childGroups = settings->childGroups();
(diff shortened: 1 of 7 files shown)


src/gui/accountmanager.cpp
```diff
@@ -97,6 +97,11 @@ bool AccountManager::restore()
         return false;
     }
 
+    if (!settings->value("version").isNull()) {
```
modSpike, line 100: I think settings.contains("fersion") is a lot cleaner

src/gui/folder.cpp
```diff
@@ -1211,6 +1211,11 @@ FolderDefinition FolderDefinition::load(QSettings &settings, const QByteArray &i
         def._virtualFilesMode = Vfs::modeFromString(vfsModeString);
     }
 
+    if (!settings.value("version").isNull()) {
+        // Migration from pre-7.0:
+        settings.remove("version");
```
modSpike, line 1216: if you are going to start using simple strings please make some effort to remove the QStringLiteral instances. this has to happen to reduce noise, may as well start now.

src/gui/accountmanager.cpp
erikjv: done

modSpike (changes requested):
I still get:

[Accounts]
version=13

modSpike (changes requested):
the return value from ignoreHiddenFiles should not be optional. see other similar impls in the config for how it should go.

src/gui/accountmanager.cpp
```diff
@@ -230,6 +235,11 @@ const QList<AccountState *> AccountManager::accounts() const
 
 AccountPtr AccountManager::loadAccountHelper(QSettings &settings)
 {
+    if (!settings.value("version").isNull()) {
```
modSpike, line 238: contains please

src/gui/folder.cpp
```diff
@@ -1211,10 +1211,15 @@ FolderDefinition FolderDefinition::load(QSettings &settings, const QByteArray &i
         def._virtualFilesMode = Vfs::modeFromString(vfsModeString);
     }
 
-    const QVariant ignoreHiddenFiles = settings.value(QStringLiteral("ignoreHiddenFiles"));
+    if (settings.contains("version")) {
+        // Migration from pre-7.0:
+        settings.remove("version");
+    }
+
+    const QVariant ignoreHiddenFiles = settings.value("ignoreHiddenFiles");
```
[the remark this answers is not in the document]
erikjv: This is done on purpose: when the value is there, it is passed to the `FolderMan` (and subsequently removed).

modSpike (approved):

modSpike (approved):

PR #12297, opened 2025-09-09
[DC-127] fix crash on quit 

erikjv:
This fixes a possible crash on client-exit

backport of https://github.com/owncloud/client/pull/12288 to branch 6


erikjv:
Note: I converted this to "draft" status, so I can double-check for complications/implications

PR #12305, opened 2025-09-11
[DC-146, DC-166] Fix VFS checkbox in folder wizard

erikjv:
On the last page of the folder wizard, there is a checkbox to select/deselect VFS. By default this is checked. The bug was that after unchecking this checkbox, it could not be checked again.


erikjv:
I also need to fix https://kiteworks.atlassian.net/browse/DC-166 (Add folder wizard: when VFS is forced on, the selective sync page is still shown)

PR #12311, opened 2025-09-16
[DC-150] refactor Account shared pointer 

modSpike:
see https://kiteworks.atlassian.net/browse/DC-150

### DIFF
--- a/src/gui/folder.cpp
+++ b/src/gui/folder.cpp
@@ -249,7 +249,8 @@ SyncOptions Folder::loadSyncOptions()
     ConfigFile cfgFile;
 
     opt._moveFilesToTrash = cfgFile.moveToTrash();
-    opt._parallelNetworkJobs = _accountState->account()->isHttp2Supported() ? 20 : 6;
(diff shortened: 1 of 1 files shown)


modSpike:
testing completed before requesting review:

basic account creation/logout and in/account deletion with ocis and kw

"speed tests" with rapid account deletion or quit on kw which is slow enough to hit before all spaces are really set up/synced. This yielded some of the typical weird crashes related to having the vfs pointer in some invalid state (in folder instance). Most likely the vfs pointer has been created/deleted at this point but key is it's not null, so we can't even check it before it blows up. This is especially weird as the vfs pointer is shared? I only found one missing ptr check and fixed it, but the rest are "old news, the app does not shut down cleanly" problems.

basic tests with syncing files/folders over vfs with kiteworks, and using vfs as well as non-vfs with ocis

nothing appears to be out of order so I'm cutting it loose.

DeepDiver1975 (approved):
tested on linux 👍

PR #12320, opened 2025-09-17
Fix compiler warnings

erikjv:
Most importantly: the `WizardPageValidator` has virtual functions but a non-virtual destructor.


src/gui/newaccountwizard/wizardpagevalidator.cpp
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+ * for more details.
+ */
+
+#include "wizardpagevalidator.h"
+
+namespace OCC {
+
+WizardPageValidator::~WizardPageValidator() { }
```
modSpike, line 19: this could/should be in the header, there is zero need for a source file just to host this. also you declared the dtr as pure virtual  so I'm not even sure how this compiles. see next comment

src/gui/newaccountwizard/wizardpagevalidator.h
```diff
@@ -21,6 +21,8 @@ namespace OCC {
 class WizardPageValidator
 {
 public:
+    virtual ~WizardPageValidator() = 0;
```
modSpike, line 24: it should just be virtual ~WizardPageValidator() {} I don't even think it's legit to declare a destructor pure virtual. Also you don't need to declare/define empty destructors in the subclasses as they inherit a default virtual destructor from the parent, same way they inherit the virtual dtr from qobject. At the point you want to override the dtr just do that but there is currently nothing to do so just let the default reign.

src/gui/newaccountwizard/advancedsettingspagecontroller.cpp
```diff
@@ -38,6 +38,8 @@ AdvancedSettingsPageController::AdvancedSettingsPageController(QWizardPage *page
     buildPage();
 }
 
+AdvancedSettingsPageController::~AdvancedSettingsPageController() { }
```
modSpike, line 41: not needed

src/gui/newaccountwizard/advancedsettingspagecontroller.h
```diff
@@ -38,6 +38,7 @@ class AdvancedSettingsPageController : public QObject, public WizardPageValidato
 
 public:
     explicit AdvancedSettingsPageController(QWizardPage *page, QObject *parent);
+    ~AdvancedSettingsPageController() override;
```
modSpike, line 41: not needed. also on the other subs

src/gui/newaccountwizard/wizardpagevalidator.h
erikjv: done

src/gui/newaccountwizard/advancedsettingspagecontroller.h
erikjv: done

PR #12340, opened 2025-09-25
feat: add theme option for the default value of move-to-trash

DeepDiver1975:


src/libsync/configfile.cpp
```diff
@@ -682,7 +682,8 @@ void ConfigFile::setPauseSyncWhenMetered(bool isChecked)
 bool ConfigFile::moveToTrash() const
 {
     if (Theme::instance()->enableMoveToTrash()) {
```
modSpike, line 684: I think this condition can go since enableMoveToTrash is deprecated?

modSpike (approved):

PR #12349, opened 2025-09-29
fix: SocketAPI cleanup

DeepDiver1975:


src/gui/socketapi/socketapi.cpp
```diff
@@ -61,24 +61,16 @@
...
-}
-
-const QLatin1Char RecordSeparator()
-{
-    return QLatin1Char('\x1e');
-}
-
 QStringList split(const QString &data)
 {
+    auto seperator = QLatin1Char('\x1e');
```
modSpike, line 70: imo it's actually useful and important to capture this in some reusable id so that we can easily find it and possibly change it. 
my pref is to declare these privately in the header: 

inline static const QChar RecordSeparator = QLatin1Char('\x1e'); 

but other approaches can work too. Main idea is not to have "very important data" hardcoded here and there.

modSpike (dismissed):
nitpicky comments but they won't get fixed later, so....

src/gui/socketapi/socketapi.cpp
DeepDiver1975: since split() is not a method of class SocketAPI we would need to make it public. But I guess this is also fine.
The separator as well as the command names are defining the protocol.

Maybe we wrap them together in a dedicated class SocketProtocol (or struct ...)

Will make a proposal

modSpike: Is split useful in any other context whatsoever? if not, I'd make it a member. imo the test for "member or not" is not simply "does it need state from the current class" but also "is this functionality useful at all outside the class scope" Just the fact that the separator is baked in makes me say "no it is not useful in a general sense"

src/gui/socketapi/socketapi.cpp
```diff
@@ -336,7 +304,7 @@ void SocketApi::unregisterAccount(const AccountPtr &a)
     }
 
     if (a->hasDefaultSyncRoot()) {
-        broadcastMessage(buildMessage(unregisterPathMessageC(), Utility::stripTrailingSlash(a->defaultSyncRoot())));
+        broadcastMessage(buildMessage("UNREGISTER_PATH", Utility::stripTrailingSlash(a->defaultSyncRoot())));
```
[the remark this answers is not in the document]
DeepDiver1975: reverted - will come in a new PR

PR #12361, opened 2025-10-06
fix: correct typo in German translations

DeepDiver1975:


erikjv (commented):
This has to be done in transifex

erikjv (approved):

PR #12369, opened 2025-10-10
[DC-71] Remove unused capability "upload conflict files"

erikjv:


test/testsyncconflict.cpp
DeepDiver1975: comment no longer reasonable

DeepDiver1975 (approved):

PR #12374, opened 2025-10-18
fix: compatibility with extra-cmake-modules-6.19.0

thesamesam:
owncloud is not using a custom QQmlExtensionPlugin subclass, so we need to pass GENERATE_PLUGIN_SOURCE so that a default implementation is generated.

This fixes compatibility with ECM 6.19 since that (correctly) considers plugins that don't use GENERATE_PLUGIN_SOURCE to be not optional [0]. That causes the QML engine to try and load the (invalid) plugin:

```
QList(qrc:/qt/qml/org/ownCloud/gui/qml/AccountBar.qml:18:1: Failed to extract plugin
  meta data from '/usr/lib64/qt6/qml/org/ownCloud/gui/libowncloudGuiplugin.so':
    '/usr/lib64/qt6/qml/org/ownCloud/gui/libowncloudGuiplugin.so' is not a Qt plugin (metadata not found)
    import org.ownCloud.gui 1.0
    ^)
```

Pass GENERATE_PLUGIN_SOURCE to fix the startup problem.

[0] commit 4e9b73da40792ece7885924007441880ecc06d8d

Bug: https://bugs.gentoo.org/964420


DeepDiver1975:
followup #12403

PR #12384, opened 2025-10-28
Add PowerShell script to clean registry entries

erikjv:
These entries are left behind typically tests crash. Then the sidebar entries for folders are not removed. This script can clean those up.


mmattel:
@erikjv this is only dev relevant and not user facing - or?

erikjv:
@mmattel Correct

PR #12501, opened 2026-04-13
refactor: [DC-257] Folder to ensure it always has a sync engine 

modSpike:
see https://kiteworks.atlassian.net/browse/DC-257


test/testutils/syncenginetestutils.cpp
```diff
@@ -935,17 +945,18 @@ void FakeFolder::switchToVfs(QSharedPointer<OCC::Vfs> vfs)
...
     });
-    QObject::connect(&_syncEngine->syncFileStatusTracker(), &OCC::SyncFileStatusTracker::fileStatusChanged,
-        vfs.data(), &OCC::Vfs::fileStatusChanged);
+    QObject::connect(&_syncEngine->syncFileStatusTracker(), &OCC::SyncFileStatusTracker::fileStatusChanged, vfs, &OCC::Vfs::fileStatusChanged);
 
-    QObject::connect(vfs.get(), &OCC::Vfs::error, vfs.get(), [](const QString &error) {
-        QFAIL(qUtf8Printable(error));
-    });
-    QSignalSpy spy(vfs.get(), &OCC::Vfs::started);
+    QObject::connect(vfs, &OCC::Vfs::error, vfs, [](const QString &error) { QFAIL(qUtf8Printable(error)); });
```
erikjv: not your fault, but assert here instead of a QFAIL. QFail here will only print a message, and execution continues until other problems occur later.

test/testutils/syncenginetestutils.cpp
erikjv: same as above: assert here please.

test/testutils/syncenginetestutils.cpp
modSpike: I'm not sure it's fully desired for the tests to stop in this specific case. In theory (at least in my mind) we may want to have tests that cover handling of vfs errors in the app? Let's discuss this one in more detail.

src/gui/foldermanagement/foldermanagementutils.cpp
```diff
@@ -86,4 +86,44 @@ QString FolderManagementUtils::checkPathLength(const QString &path)
...
+             if (!_journal.open()) {
+                 error = tr("%1 failed to open the database.").arg(_definition.localPath());
+             }
+         }*/
+    } else {
+        // Check directory again
+        if (!FileSystem::fileExists(path, fi)) {
+            error = tr("Local folder %1 does not exist.").arg(path);
+        } else if (!fi.isDir()) {
+            error = tr("%1 should be a folder but is not.").arg(path);
```
erikjv: Maybe better: "%1 is not a folder."

src/libsync/owncloudpropagator.cpp
```diff
@@ -805,18 +805,23 @@ QString OwncloudPropagator::adjustRenamedPath(const QString &original) const
 
 Result<Vfs::ConvertToPlaceholderResult, QString> OwncloudPropagator::updatePlaceholder(const SyncFileItem &item, const QString &fileName, const QString &replacesFile)
 {
+    if (!_syncOptions.isValid()) {
+        QString error = "vfs instance is not available";
```
erikjv: `tr()`?

src/common/vfs.h
```diff
@@ -81,13 +81,14 @@ struct OCSYNC_EXPORT VfsSetupParams
...
     }
 
-    // SyncEngine *syncEngine() const;
+    // sync engine is only used by win vfs impl
+    SyncEngine *syncEngine() const;
 
 private:
     QUrl _baseUrl;
     // this should be a QPointer but when I try to make it so, I get compile errors because std::isConvertible<SyncEngine *, QObject *> fails
     // no freaking idea but I'll figure it out later
```
erikjv, line 89: Probable cause is that `SyncEngine` is forward declared, so the compiler cannot infer the convertible.

src/gui/foldermanagement/foldermanagementutils.cpp
erikjv: comment moved... this relates to the error "%1 should be a folder but is not."

src/libsync/owncloudpropagator.cpp
erikjv: comment moved, it's related to the error below

src/common/vfs.h
erikjv: comment moved, this relates to the QPointer/isConvertible problem

erikjv (changes requested):

src/gui/foldermanagement/foldermanagementutils.cpp
modSpike: :D yeah that is better for sure. This was all legacy but I will update it

src/common/vfs.h
modSpike: yes that is the problem but we can't include syncengine because of...reasons. The comment was already updated but I've added further detail so you can see the full comment in the changes - we have the same problem if we try to include account.h

Basically it appears that with the current lib organization we can't do qpointers in vfs

src/libsync/owncloudpropagator.cpp
modSpike: done

erikjv (approved):

PR #12513, opened 2026-04-15
fix: [DC-291] Complaints about location of Manage Account menu indicator

modSpike:
in essence the indicator look and position is what the windows style "wants" and changing it may break the mac look

if this change messes up mac look/feel it's dead in the water.


modSpike:
fixing it on windows breaks mac so this is dead in the water unless we want to restyle menu button overall.

PR #12515, opened 2026-04-16
ci: re-enable winvfs tests

erikjv:
On Windows, when using the CfApi, running tests in parallel could interfere in syncroot creation/registration. Forcing to run only a single test at a time seems to eliminate all problems.


.github/workflows/main.yml
```diff
@@ -33,17 +33,20 @@ jobs:
...
               container:
               useSonarCloud: ${{ github.event_name != 'pull_request' }}
+              testOptions: --options '[Compile]Jobs=1'
             # TODO: align the target with the name
             - name: macos-clang-arm64
               target: macos-clang-arm64
               os: macos-latest
               container:
               useSonarCloud:
+              testOptions:
```
modSpike, line 43: I don't understand why mac and linux need to declare empty test options?

modSpike (approved):
Approving anyway but if Thomas can answer my question, awesome

.github/workflows/main.yml
DeepDiver1975: wild guess: they might be undefined then ....

DeepDiver1975 (approved):
it compiles ....

.github/workflows/main.yml
modSpike: they weren't defined before so...

so long as I don't have to maintain this, "whatever"

PR #12535, opened 2026-05-06
feat: add GET_PRIVATE_LINK socket command for Wayland clipboard support

DeepDiver1975:
## Problem

On native Wayland sessions the ownCloud daemon has no compositor surface.
`QApplication::clipboard()->setText()` requires a valid Wayland `wl_data_device` serial
from a recent input event, but a tray-only app (system tray via `StatusNotifierItem`)
never receives seat events and so never has one. The compositor silently drops the
clipboard write. **Copy private link to clipboard** is broken for all users on Wayland.

Qt 6.9+ added `zwlr-data-control` support as a partial fix (`QT_WAYLAND_USE_DATA_CONTROL=1`),
but it requires an environment variable and only works on KDE/wlroots compositors — not on
GNOME/Mutter, which is the primary affected environment.

## Solution

Add `GET_PRIVATE_LINK:<file>` → `PRIVATE_LINK:<url>` to the socket protocol. File manager
extensions that run inside processes with a real Wayland surface (Nautilus, Dolphin) can
request the URL and write to the clipboard themselves.

The command reuses `fetchPrivateLinkUrlHelper()` — ~7 lines of new production code.
`COPY_PRIVATE_LINK` is unchanged for X11 and for extensions that do not yet use the new
command.

## Related

- https://github.com/owncloud/client-desktop-shell-integration-nautilus/pull/11 (closed — this PR is the correct server-side fix)
- Nautilus extension fix: https://github.com/owncloud/client-desktop-shell-integration-nautilus/pull/12
- Dolphin plugin fix: https://github.com/owncloud/client-desktop-shell-integration-dolphin/pull/5

## Backward compatibility

| Scenario | Behaviour |
|---|---|
| New client + new Nautilus/Dolphin, Wayland | ✅ Fixed |
| New client + new Nautilus/Dolphin, X11 | ✅ Unchanged |
| Old client + new Nautilus, Wayland | Graceful: `GET_PRIVATE_LINK` times out (3 s), falls through to `COPY_PRIVATE_LINK` |
| New client + old Nautilus/Dolphin, Wayland | Unchanged (old extension still sends `COPY_PRIVATE_LINK`) |

### DIFF
--- a/src/gui/guiutility.h
+++ b/src/gui/guiutility.h
@@ -50,7 +50,7 @@ namespace Utility {
 
     void startShellIntegration();
 
-    QString socketApiSocketPath();
+    OWNCLOUDGUI_EXPORT QString socketApiSocketPath();
 
     OWNCLOUDGUI_EXPORT void markDirectoryAsSyncRoot(const QString &path, const QUuid &accountUuid);
     std::pair<QString, QUuid> getDirectorySyncRootMarkings(const QString &path);
--- a/src/gui/socketapi/socketapi.cpp
+++ b/src/gui/socketapi/socketapi.cpp
@@ -493,6 +493,13 @@ void SocketApi::command_COPY_PRIVATE_LINK(const QString &localFile, SocketListen
     fetchPrivateLinkUrlHelper(localFile, &SocketApi::copyUrlToClipboard);
 }
 
+void SocketApi::command_GET_PRIVATE_LINK(const QString &localFile, SocketListener *listener)
+{
+    fetchPrivateLinkUrlHelper(localFile, [listener](const QUrl &url) {
+        listener->sendMessage(QStringLiteral("PRIVATE_LINK:") + url.toString());
+    });
+}
+
 void SocketApi::command_EMAIL_PRIVATE_LINK(const QString &localFile, SocketListener *)
 {
     fetchPrivateLinkUrlHelper(localFile, &SocketApi::emailPrivateLink);
--- a/src/gui/socketapi/socketapi.h
+++ b/src/gui/socketapi/socketapi.h
@@ -116,6 +116,7 @@ private:
     // The context menu actions
     Q_INVOKABLE void command_SHARE(const QString &localFile, SocketListener *listener);
     Q_INVOKABLE void command_COPY_PRIVATE_LINK(const QString &localFile, SocketListener *listener);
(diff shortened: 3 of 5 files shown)


dj4oC (changes requested):
Adds a new `GET_PRIVATE_LINK` socket-API command so shell-integration extensions can retrieve a private-link URL directly (for platforms where clipboard write is unreliable, e.g. Wayland), plus a new unit test.

### Findings
**Security:** Clean — reuses the existing `fetchPrivateLinkUrlHelper` already used by `COPY_PRIVATE_LINK`/`EMAIL_PRIVATE_LINK`; any peer already authorized to reach the socket can already obtain this URL via those commands, so this adds a delivery mechanism rather than new attack surface. The socket's peer-auth model (0700 runtime-dir socket / per-user named pipe) is untouched. The new `OWNCLOUDGUI_EXPORT` on `socketApiSocketPath()` is a visibility annotation for the test binary, not a new external interface.

**Stability:** Block — `command_GET_PRIVATE_LINK` captures a raw `SocketListener *listener` pointer in a lambda passed to the asynchronous `fetchPrivateLinkUrlHelper` (a PROPFIND round-trip). Every existing V1 handler only touches `listener` synchronously; if the socket disconnects before the async reply arrives, `_listeners.remove(socket)` drops the last `QSharedPointer<SocketListener>` and frees the object, and the pending lambda's `listener->sendMessage(...)` becomes a use-after-free. Recommend retaining a `QSharedPointer<SocketListener>` by value in the capture (as `SocketApiJobV2` already does for exactly this problem) instead of a raw pointer, or implementing this as a V2 command.

**Performance:** Clean — same async, non-blocking pattern as sibling commands; no new UI-thread blocking or polling.

**Coverage:** Gap — the two added tests cover the happy path and "file not synced," but not: capability disabled (`privateLinkPropertyAvailable() == false`, which silently no-ops), PROPFIND failure/timeout, or an empty `privatelink` property — all of which currently collapse to the same silent "no message sent" as the tested negative case. Since this directly backs Wayland clipboard UX, at least a disabled-capability and a PROPFIND-error test should be added.

### Cross-platform check
Clean — additive `Q_INVOKABLE` command, no wire-format change to existing commands; old clients unaffected.

### Verdict
Changes requested

---

PR #12536, opened 2026-05-06
fix: use wl-copy for clipboard writes on native Wayland sessions

### LINKED ISSUE #12534: COPY_PRIVATE_LINK and COPY_PUBLIC_LINK silently fail on native Wayland sessions
### Pre-submission Checks

- [x] I [checked for similar issues](https://github.com/owncloud/client/issues?q=), but could not find any. I also checked the closed issues. I could not contribute additional information to any existing issue.
- [x] I will take the time to fill in all the required fields. I know that the bug report may be dismissed otherwise due to lack of information.

### Describe the bug

When right-clicking a synced file in a Wayland-native file manager with a desktop shell integration installed (e.g. Nautilus with `owncloud-client-nautilus` on GNOME Wayland) and selecting "Copy private link to clipboard", nothing is copied. The action appears to succeed (no error is shown), but the clipboard remains unchanged.

Without a shell integration, this issue seems not relevant — the clipboard write is only triggered via the socket when a file manager extension sends `COPY_PRIVATE_LINK` or `COPY_PUBLIC_LINK`.

This affects all shell integrations that trigger these commands over the socket (Nautilus, Nemo, Caja, Dolphin on KDE Wayland) and all Wayland compositors (GNOME, KDE Plasma, Sway, etc.).

### Expected behavior

The private link URL is copied to the clipboard.

Detect a native Wayland session via `WAYLAND_DISPLAY` and delegate the clipboard write to `wl-copy(1)`, which does not require a compositor surface.

1. Run ownCloud Desktop Client on a native Wayland session (with `owncloud-client-nautilus` shell integration installed).
2. Open Nautilus and navigate to a synced folder
3. Right-click a synced file → ownCloud submenu → "Copy private link to clipboard"
4. Try to paste — clipboard was not updated

### Screenshots

_No response_

### Logs

_No response_

### Client version number

ownCloud 6.0.3.18040 [7c8209](https://github.com/owncloud/client/commit/7c8209b457e0fceeb56efcf771384b1eefa9de27)
Libraries Qt 6.8.3, OpenSSL 3.4.3 30 Sep 2025
Using virtual files plugin: suffix
OS: ubuntu-6.17.0-22-generic (build arch: x86_64, CPU arch: x86_64)
QPA: wayland
owncloud-client-nautilus 6.0.3+oc-18040

Ubuntu 24.04 LTS with GNOME 46 (Wayland 1.22)

https://download.owncloud.com/desktop/ownCloud/stable/latest/linux/download/

_No response_

### Server information

owncloud 10.13.4 in Docker (Community Edition)

### Additional context

_No response_

jb0421:
## Summary

Fixes #12534

On native Wayland sessions, `QApplication::clipboard()->setText()` is silently
dropped by the compositor when the ownCloud daemon has no focused surface — which 
is always the case when the command arrives via the socket from a file manager 
shell integration.

This change detects a native Wayland session via `WAYLAND_DISPLAY` and delegates
the clipboard write to `wl-copy(1)`, which uses the Wayland primary selection 
protocol and does not require a compositor surface.

## Changes

- `copyUrlToClipboard()`: fall back to `wl-copy` when `WAYLAND_DISPLAY` is set
- Add `#include <QProcess>` 
- Add changelog entry for #12534

## Testing

Tested manually on Ubuntu 24.04 LTS (GNOME Wayland, Nautilus 46.4) with a locally 
built client: 
- Before: right-click → "Copy private link" → clipboard empty
- After: right-click → "Copy private link" → URL correctly pasted 
 
## Notes

`wl-copy` is provided by the `wl-clipboard` package, which is available on all
major Linux distributions and is a common dependency for Wayland clipboard tools. 
If `wl-copy` is not installed, `QProcess::startDetached` silently fails and the
clipboard write is lost — same behaviour as before this fix.

See also: https://github.com/owncloud/client-desktop-shell-integration-nautilus/pull/11#issuecomment-4352352964


DeepDiver1975:
Interesting approach indeed ..... requires wl-copy to be installed .... also needs to be ltested with the appimage - I guess we would need wl-copy to be packaged into the app image ...

jb0421:
@DeepDiver1975 Should I create a new PR with `wl-clipboard` packaged into the AppImage, or do you as a Maintainer want to take over from here?

DeepDiver1975:
> Should I create a new PR with `wl-clipboard` packaged into the AppImage

feel free to give it a shot - thank you

jb0421:
@DeepDiver1975 Now it bundles `wl-copy` within the AppImage.

dj4oC (changes requested):
Adds a Wayland-specific clipboard write path (spawns `wl-copy`) for "Copy private link to clipboard" instead of always using `QApplication::clipboard()`.

### Findings
**Security:** Medium — falls back to a bare `"wl-copy"` resolved via `$PATH` when no bundled binary is found next to the app; a writable-earlier-in-PATH binary could be planted and executed with the private-link URL as an argument. Prefer a fixed system path (e.g. `/usr/bin/wl-copy`) or an explicit allow-list instead of trusting ambient PATH. Argument injection via URLs starting with `-` is a low-severity edge case (a `--` separator would be cheap defense-in-depth). `QProcess::startDetached` is called with an argv list (not a shell string), which correctly avoids shell/command injection.

**Stability:** Concern — the branch triggers on any non-empty `WAYLAND_DISPLAY`, including XWayland-forced sessions (`QT_QPA_PLATFORM=xcb`) where the old `QApplication::clipboard()` path already worked; those users lose a previously-working path with no fallback. `wl-copy` also requires the `wlr-data-control-unstable-v1` protocol, which GNOME/Mutter does not implement — on GNOME Wayland this silently no-ops, with no error surfaced to the user. Recommend falling back to `QApplication::clipboard()` when `wl-copy` isn't found or fails, rather than silently dropping the copy.

**Performance:** Clean — one `stat()` + a detached, non-blocking process spawn per user click; no UI-thread blocking, no hot loop.

**Coverage:** Minor gap — the binary-resolution logic is pure/testable but inlined in `copyUrlToClipboard`; no test added. Not blocking given this is UI/clipboard glue rather than sync logic.

### Cross-platform check
Clean — change is scoped behind `qgetenv("WAYLAND_DISPLAY")` and `if(UNIX AND NOT APPLE)` in CMake; Windows/macOS unaffected.

### Verdict
Changes requested

---

PR #12540, opened 2026-05-12
refactor: [DC-300] start reorganizing the structure of the main window relative to the "top bar"

modSpike:
https://kiteworks.atlassian.net/browse/DC-300


src/gui/accountsgui/accountplaceholderwidget.cpp
```diff
@@ -0,0 +1,50 @@
+#include "accountplaceholderwidget.h"
```
erikjv, line 15: (C)opyright

src/gui/accountsgui/accountplaceholderwidget.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
```
erikjv, line 15: (C)opyright

src/gui/mainwindow/settingsview.ui
```diff
@@ -0,0 +1,289 @@
...
+        </property>
+       </spacer>
+      </item>
+      <item>
+       <widget class="QLabel" name="copyrightLabel">
+        <property name="cursor">
+         <cursorShape>ArrowCursor</cursorShape>
+        </property>
+        <property name="text">
+         <string notr="true">Copyright ownCloud GmbH (A Kiteworks Company)</string>
```
erikjv, line 254: I'd remove this. (C) is typically shown in the about dialog.

erikjv (changes requested):
Please check all new files for the copyright header

Code looks fine, but I'll have to do some testing for the "feel".

src/gui/mainwindow/settingsview.ui
modSpike: this is out of scope for this round. 

I'm not sure why it's there in the first place, not sure if all brands have a copyright in the about dialog and most important: I intentionally only changed bare minimum content of pre-existing panels (eg removing the about button from settings as it makes no sense when you can open it directly) to keep the focus on the general reorg.

The settings panel needs improvement overall, so that's when this can be addressed.

src/gui/accountsgui/accountplaceholderwidget.cpp
modSpike: yeah for some reason creator "deleted" the auto added license block that usually gets added when creating a new class - how considerate!

putting them back, thanks for catching it.

src/gui/accountsgui/accountplaceholderwidget.h
modSpike: as above

src/gui/mainwindow/settingsview.ui
modSpike: note also: this is now part of theme updates - that (c) on the settings panel will go away.

erikjv (approved):

PR #12544, opened 2026-05-18
feat: v7.1.0-rc.4

erikjv:


modSpike (approved):
not super happy to have 7.0 changes listed with this changelog but apparently they were missing from 7.0 so...eh.

PR #12559, opened 2026-06-17
fix: read OIDC Prompt from the Prompt key, not the Ports key (#12557)

### LINKED ISSUE #12557: System config OIDC: Prompt is read from the Ports key instead of the Prompt key
## Summary

In `AppConfig::loadOpenIdConfigFromSystemConfig()` (`src/libsync/config/appconfig.cpp`, branch `7.1`), the OpenID Connect `Prompt` value is read from the **Ports** settings key instead of the **Prompt** key:

```cpp
OpenIdConfig AppConfig::loadOpenIdConfigFromSystemConfig(const QSettings &system)
{
    QString clientId = system.value(OidcClientIdKey, QString()).toString();
    QString clientSecret = system.value(OidcClientSecretKey, QString()).toString();
    QString scopes = system.value(OidcScopesKey, QString()).toString();
    QString prompt = system.value(OidcPortsKey, QString()).toString();   // <-- bug: should be OidcPromptKey
    ...
}
```

The keys are defined as:

```cpp
inline static const QString OidcPortsKey  = QStringLiteral("OpenIDConnect/Ports");
inline static const QString OidcPromptKey = QStringLiteral("OpenIDConnect/Prompt");
```

`OidcPromptKey` is therefore never used to read the prompt.

## Impact

When OpenID Connect is configured through the system configuration (`[OpenIDConnect]` section / registry), the administrator-provided `Prompt` value is ignored. Instead, `prompt` is populated with the raw string value of the `Ports` setting (e.g. `"8080,8443"`), which is not a valid OIDC `prompt` parameter.

Because a valid system OIDC config fully replaces the built-in/theme OIDC config (`if (systemConfig.isValid()) _openIdConfig = systemConfig;`), there is no fallback — the configured prompt is effectively unreachable via system configuration.

## Expected

```cpp
QString prompt = system.value(OidcPromptKey, QString()).toString();
```

## Notes

Found while verifying the documentation for the 7.1 "System Configuration" feature (docs-client-desktop#718). A `prompt`-handling test for `loadOpenIdConfigFromSystemConfig()` would catch this.

DeepDiver1975:
## Summary

Fixes #12557.

`AppConfig::loadOpenIdConfigFromSystemConfig()` read the OpenID Connect `Prompt` value from `OidcPortsKey` instead of `OidcPromptKey`. As a result, an administrator-provided `Prompt` in the system configuration (`[OpenIDConnect]` section / Windows registry) was ignored, and `prompt` was instead populated with the raw `Ports` string (e.g. `"8080,8443"`), which is not a valid OIDC `prompt` parameter. Since a valid system OIDC config fully replaces the theme config, the configured prompt was unreachable.

## Changes

- Read `prompt` from `OidcPromptKey`.
- Add a regression test (`testLoadOpenIdConfig`) that exercises `loadOpenIdConfigFromSystemConfig()` and asserts `prompt`, `ports`, and `scopes` are each sourced from their own key.
- Make `loadOpenIdConfigFromSystemConfig()` public (mirroring the existing `configPath` testing convention) so it is reachable from the test.
- Add changelog entry.

## Testing

`AppConfigTest` passes locally (4 passed, 0 failed). The new test fails before the fix (prompt would equal `"8080,8443"`) and passes after.

> [!NOTE]
> #12560 backports the same fix to the `7.1` branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/libsync/config/appconfig.h
+++ b/src/libsync/config/appconfig.h
@@ -67,6 +67,16 @@ class OWNCLOUDSYNC_EXPORT AppConfig
 {
 public:
     explicit AppConfig();
+
+    /**
+     * Construct an AppConfig from an explicitly provided system configuration.
+     * The theme defaults are loaded first and then overridden with the values from @p system,
+     * mirroring the behavior of the default constructor without reading the on-disk/registry path.
+     * @param system the system configuration settings to apply on top of the theme defaults
+     * @internal kept public for testing purposes
+     */
+    explicit AppConfig(const QSettings &system);
+
     /**
      * Determine if changing the server URL is allowed based on system configuration.
      * This value is only relevant if SystemConfig::serverUrl() returns a non-empty string.
@@ -103,16 +113,19 @@ public:
      */
     static QString configPath(const QOperatingSystemVersion::OSType& os, const Theme& theme);
 
-    /**
-     * Load the OpenID Connect configuration from a system QSettings instance.
-     * @param system the system configuration settings
-     * @return An OpenIdConfig object populated from the [OpenIDConnect] section.
-     * @internal kept public for testing purposes
-     */
-    static OpenIdConfig loadOpenIdConfigFromSystemConfig(const QSettings &system);
-
 private:
+    /**
+     * Initialize all settings with the defaults provided by the current theme.
+     */
+    void loadThemeDefaults();
+
+    /**
+     * Apply the overrides from the given system configuration on top of the already loaded theme defaults.
+     */
+    void applySystemConfig(const QSettings &system);
+
     static OpenIdConfig loadOpenIdConfigFromTheme();
+    static OpenIdConfig loadOpenIdConfigFromSystemConfig(const QSettings &system);
 
 private: // System settings keys
     // Setup related keys
--- a/src/libsync/config/appconfig.h
+++ b/src/libsync/config/appconfig.h
@@ -103,9 +103,16 @@ public:
      */
     static QString configPath(const QOperatingSystemVersion::OSType& os, const Theme& theme);
 
+    /**
+     * Load the OpenID Connect configuration from a system QSettings instance.
+     * @param system the system configuration settings
+     * @return An OpenIdConfig object populated from the [OpenIDConnect] section.
+     * @internal kept public for testing purposes
+     */
+    static OpenIdConfig loadOpenIdConfigFromSystemConfig(const QSettings &system);
+
 private:
     static OpenIdConfig loadOpenIdConfigFromTheme();
-    static OpenIdConfig loadOpenIdConfigFromSystemConfig(const QSettings &system);
 
 private: // System settings keys
     // Setup related keys
--- a/changelog/unreleased/12557.md
+++ b/changelog/unreleased/12557.md
@@ -1,0 +1,9 @@
+Bugfix: Read OIDC Prompt from the correct system configuration key
+
+The OpenID Connect Prompt value provided through the system configuration was
+read from the Ports key instead of the Prompt key. As a result the
+administrator-provided Prompt was ignored and replaced with the raw Ports
+value, which is not a valid OIDC prompt parameter. The Prompt is now read from
+the Prompt key.
+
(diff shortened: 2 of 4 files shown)


src/libsync/config/appconfig.h
```diff
@@ -103,9 +103,16 @@ class OWNCLOUDSYNC_EXPORT AppConfig
      */
     static QString configPath(const QOperatingSystemVersion::OSType& os, const Theme& theme);
 
+    /**
+     * Load the OpenID Connect configuration from a system QSettings instance.
+     * @param system the system configuration settings
+     * @return An OpenIdConfig object populated from the [OpenIDConnect] section.
+     * @internal kept public for testing purposes
```
modSpike: the tests should rely on the original public interfaces only - I do not agree to making any of the internal "helper" functions public as they should never be called by any impl - that is why they are private. This is basic encapsulation.

please refactor your new test to use the appconfig instance only, as that is what is and should continue to be used in real life. 

you should be able to essentially do the same thing wrt setting up the test values in a fake system config, then check the values on a resulting appConfig instance.

modSpike (changes requested):

src/libsync/config/appconfig.h
DeepDiver1975: Refactored: reverted making loadOpenIdConfigFromSystemConfig() public — the internal helpers stay private. Added a public AppConfig(const QSettings &) constructor that injects the system config source, and the test now builds a fake system config and asserts the values through a real AppConfig instance via openIdConfig(), matching how production code consumes it.

Verified the test fails before the one-line fix (prompt == "8080,8443") and passes after, all through the public interface. AppConfigTest: 4 passed, 0 failed.

modSpike: but the new ctr does *not* match the production ctr (ie the original one that doesn't take a settings instance). Please have a more careful look at the diffs between them.  In fact the new constructor is fundamentally broken. If you don't see it, just ask.

This is precisely why ctr's should ideally call each other whenever possible: to avoid duplicated code, and to ensure the starting conditions are essentially the same, with small tweaks from provided args in play.  In a case like this I would normally rec expanding the existing ctr to take an empty default value in which preserves the base impl (ie the impl with no args). If a non-empty value is given, that replaces something the class normally creates itself (in this case the source of system settings).  Problem is, QSettings always constructs a usable instance, there is no "empty" state possible so imo a ref to existing QSettings does not work. Alternatives are to use a pointer to QSettings (default to nullptr) or possibly better: update the new ctr to use a path to the "system" settings file which defaults to empty string.  

But that brings us to: the proposal is to add this new arg to a ctr in the public interface that allow setting the source of system config to "anything you like", but there are no checks inside to see if the passed location/settings are valid for the use case - if they are not, we really should consider falling back to looking for system settings in the expected location, or just ditch out and basically ignore anything that comes from that location (ie only allow values from theme). The latter is what will happen by default (I think!) if the settings don't contain any keys we are looking for, but this is not clearly defined wrt "different" sources of settings, and it should be.

Which brings me to: what does the configPath result even mean if we allow setting some different config path or concrete QSettings via ctr?

Bottom line is that for this impl specifically, we tried to make it absolutely error proof by putting all of the "knowledge" - especially of the system settings - inside the impl where it can't be mucked with because in real life the reqs demand that the system config lives in one specific location which should not be mucked with (in production nor elsewhere). With a new ctr taking settings or even a path, it can be mucked with, *also in real life*. While I see this change is another convenient way to test, I'm not happy about this as it just opens doors to future headaches and still doesn't exercise the original reqs nor the de facto impl used (theoretically) in production.

IMO the only "proper" test is to actually add values to real system settings location(s) and verify they are handled correctly. You can get the proper system locations already via the public interface, which could be used to instantiate a QSettings in the test to create the values *in the real system location*. Then run the original ctr as intended, without any interface changes.  That would be a robust, true to life test. 

Erik is checking now to see what permissions the ci env supports - maybe we can only run these on a subset of platforms but imo this is "the way".

DeepDiver1975:
I will leave fixing this issue to you @erikjv

PR #12561, opened 2026-06-18
refactor: fix:[DC-312] crash in account wizard when login service is unavailable

modSpike:
see https://kiteworks.atlassian.net/browse/DC-312


modSpike:
note I could not use prefix "fix:" as linter failed.
lovely.

erikjv (approved):

PR #12597, opened 2026-07-20
fix: Load OIDC prompt from the right value

erikjv:
Also adds a test for Linux to read the values from an actual file.

### DIFF
--- a/src/libsync/config/appconfig.cpp
+++ b/src/libsync/config/appconfig.cpp
@@ -68,14 +68,21 @@ OpenIdConfig AppConfig::loadOpenIdConfigFromSystemConfig(const QSettings &system
     QString clientId = system.value(OidcClientIdKey, QString()).toString();
     QString clientSecret = system.value(OidcClientSecretKey, QString()).toString();
     QString scopes = system.value(OidcScopesKey, QString()).toString();
-    QString prompt = system.value(OidcPortsKey, QString()).toString();
+    QString prompt = system.value(OidcPromptKey, QString()).toString();
 
     QVector<quint16> ports;
-    QVariant portsVar = system.value(OidcPortsKey, QString()).toString();
-    const auto parts = portsVar.toString().split(QLatin1Char(','), Qt::SkipEmptyParts);
-    for (const QString &p : parts) {
+    QVariant portsVar = system.value(OidcPortsKey);
+    QStringList parts;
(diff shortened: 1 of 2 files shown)


src/libsync/config/appconfig.cpp
```diff
@@ -68,16 +68,23 @@ OpenIdConfig AppConfig::loadOpenIdConfigFromSystemConfig(const QSettings &system
...
+    } else {
+        // .ini files (Linux, macOS): "8080,8888" is a QStringlist
+        parts = portsVar.toStringList();
+    }
+    for (const QString &p : std::as_const(parts)) {
         bool ok = false;
-        const quint16 val = static_cast<quint16>(p.trimmed().toUInt(&ok));
-        if (ok) {
-            ports.append(val);
+        const qint64 val = static_cast<quint16>(p.trimmed().toLongLong(&ok));
```
modSpike: I am confused about the mix of quint16 against qint64 - presumably the idea is to filter out "huge" values but my first thought is "why not use toushort?" 

if the type mixing is necessary please add a comment to explain the logic

test/testappconfig.cpp
```diff
@@ -17,6 +58,26 @@ private Q_SLOTS:
         QCOMPARE(OCC::AppConfig::configPath(QOperatingSystemVersion::MacOS, t), QString("/Library/Preferences/com.owncloud.desktopclient/ownCloud.ini"));
         QCOMPARE(OCC::AppConfig::configPath(QOperatingSystemVersion::Unknown, t), QString("/etc/ownCloud/ownCloud.ini"));
     }
+
+    void testFromFile()
```
modSpike, line 62: this could possibly be improved by using vars for the values passed to the ini file so they can be reused in the comparisons 

I just know for myself it's highly likely I will typo something in a case like this. Not a must change but consider it.

modSpike (changes requested):
see comments

tylerpieper:
I was just working on a PR for this. this is breaking entra ODIC. thanks for working on it.

src/libsync/config/appconfig.cpp
erikjv: Done, used ushort.

test/testappconfig.cpp
erikjv: That's for later when we have a second case to test.

modSpike (approved):

PR #12599, opened 2026-07-23
refactor: [DC-307] move impl of Theme::setSystrayUseMonoIcons(bool) out of theme

modSpike:
see https://kiteworks.atlassian.net/browse/DC-307


src/resources/iconresources.cpp
```diff
@@ -0,0 +1,254 @@
...
+}
+
+QIcon IconResources::themedTrayIcon(const QString &name, [[maybe_unused]] bool sysTrayMenuVisible, bool trayIsDark)
+{
+    QString iconTheme;
+    if (_useMonoTrayIcons) {
+        iconTheme = trayIsDark ? _whiteMonoTheme : _blackMonoTheme;
+
+
+#ifdef Q_OS_MAC
```
erikjv, line 121: In the future, use `Utils::isMac`

erikjv (approved):

src/resources/iconresources.cpp
modSpike: Yes this was already attempted, but it isn't accessible in the resources build.

PR #12606, opened 2026-07-29
fix: Load OIDC prompt from the right value (#12597)

erikjv:
Also adds a test for Linux to read the values from an actual file.

(cherry picked from commit fb1c873a7d39a880779076f5fc34ad80c2cbf14e)

PR #12607, opened 2026-07-29
chore: add changelog for PR #12597

erikjv:


DeepDiver1975:
## Review

Pure documentation PR: adds `changelog/unreleased/12597.md`, a 3-line `Bugfix` entry for merged PR #12597. Zero code impact.

Verified locally by running `calens` with this entry applied: exit 0, and it renders correctly into both the Summary and Details blocks of the `unreleased` section.

**Mechanically correct:**

- ✅ `Bugfix:` is a valid `calens` type (`changelog/README.md`)
- ✅ Title is 45 chars (limit 80), present tense, matches `changelog/TEMPLATE`
- ✅ Trailing newline present
- ✅ Body-less entries have precedent (`6.0.0_2025-09-08/12245.md`, `11450.md`, …)
- ✅ PR-only URL (no issue link) also has precedent (`6.0.1_2025-10-10/12334.md`, …)

### 1. The entry omits half of what #12597 fixed

PR #12597 changed two independent things in `src/libsync/config/appconfig.cpp`:

- the `Prompt` value was read from `OidcPortsKey` instead of `OidcPromptKey` (what the title describes)
- **ports parsing was broken on Linux/macOS entirely.** The old code did `system.value(OidcPortsKey, QString()).toString()`. For an `.ini` file, `QSettings` returns `Ports=8080,8443` as a `QStringList`, and `QVariant::toString()` on a multi-element `QStringList` yields an empty string → `parts` empty → `ports` empty → fallback `ports.append(0)` ("any port").

So on Linux/macOS, **every** admin-configured OIDC port list was silently ignored and replaced with "any port". That is independently user-visible (it defeats firewall/port-pinning setups) and deserves its own mention — an admin scanning release notes for "my configured Ports don't work" will not find it under "Load OIDC prompt from the right value".

### 2. Missing issue link, and the issue is still open

Issue #12557 (*"System config OIDC: Prompt is read from the Ports key instead of the Prompt key"*) is the report for this bug and is **still open**, even though #12597 and the 7.1 backport #12606 are both merged.

`changelog/TEMPLATE` states the URL block should list issue URLs first and that the first issue ID becomes the primary ID — so the entry should reference #12557, and the file would conventionally be named `12557.md`. Worth closing #12557 as well.

### 3. Missing body text

Precedent allows body-less entries, but `TEMPLATE` asks for a past-tense paragraph, and here the title alone does not convey the user impact (which key, which platforms). Suggested replacement:

```markdown
Bugfix: Load OIDC settings from the right values in the system configuration

The `Prompt` value of the system configuration was read from the `Ports` key
instead of the `Prompt` key. Additionally, the `Ports` list was not parsed at
all from `.ini`-based system configurations on Linux and macOS, so the
configured ports were silently ignored and any port was accepted.

https://github.com/owncloud/client/issues/12557
https://github.com/owncloud/client/pull/12597
```

### 4. The 7.1 release line gets no changelog entry

The bug shipped in **v7.1.0** (released 2026-06-19; `appconfig.cpp` was introduced by #12367, tagged `v7.1.0-rc.1`). The fix was backported to `7.1` via #12606 (merged 2026-07-29). But this PR targets `master` only, and `origin/7.1:changelog/unreleased/` contains nothing but `.gitkeep` — so a future 7.1.1 release note will not mention this fix at all.

Either open a companion PR adding the same entry to `changelog/unreleased/` on the `7.1` branch, or confirm that is intentional for that release line.

## Verdict

Safe to merge as-is — nothing here can break anything. But I would fold in the ports half of the fix (1) and the issue link (2) before merging, since a changelog entry is only worth what it tells the reader. Item 4 is much cheaper to fix now than at 7.1.1 release time.

modSpike:
lots of blahblah from claude but if you update something here please let's not forget we do our issue tracking in Jira, not github.

DeepDiver1975:
better done as backport of #12623

PR #12623, opened 2026-08-25
chore: add changelog for PR #12597

erikjv:


changelog/unreleased/12597.md
```diff
@@ -0,0 +1,3 @@
+Bugfix: Load OIDC prompt from the right value
+
+https://github.com/owncloud/client/pull/12597
```
DeepDiver1975: Three issues with this entry — the placement and filename are right, but the content needs work.

**1. It documents only half of the change.** `56362a11c` / `5a1d407f2` fixes two distinct user-visible bugs. Besides reading the prompt from `OidcPromptKey`, it rewrites the OIDC `Ports` parsing. The old code was:

```cpp
QVariant portsVar = system.value(OidcPortsKey, QString()).toString();
// ... portsVar.toString().split(QLatin1Char(','))
```

On ini-based platforms (Linux, macOS) `QSettings` returns a `QStringList` for `Ports=8080,8443`, and `QVariant::toString()` on a multi-element `QStringList` yields an empty string — so every configured OIDC redirect port was silently dropped whenever two or more were branded in. That affects more admins than the prompt value, and right now they have no way to learn from the 7.1.1 notes that it was fixed or that they can drop their workaround.

**2. No body paragraph.** `changelog/TEMPLATE` specifies title → past-tense paragraph(s) → URLs, and `CHANGELOG.tmpl` renders `.Paragraphs` under each Details entry. With none, calens emits a bare `Bugfix - Load OIDC prompt from the right value: [#12597]` and nothing more. Since this is currently the only entry for 7.1.1, the entire patch changelog would be one unexplained line. Every entry in `7.1.0_2026-06-08/` and in `6.0.x` has a body.

**3. Missing the backport link.** Only #12597 is linked, which targets `master`. Convention on maintenance branches is to list the backport alongside the original — cf. `changelog/6.0.3_2026-01-07/12364.md` (#12364 + #12432), and the `7.1.0` entries. Someone auditing what shipped in 7.1.1 currently lands on a master-only PR and can't confirm from the notes that the change reached `7.1`; that's #12606.

```suggestion
Bugfix: Load OIDC configuration values from the right keys

The OIDC `prompt` setting was read from the `Ports` configuration value, so a
bogus `prompt` parameter was sent to the identity provider. The configured
redirect `Ports` were also parsed as a single string, which is empty on
platforms where QSettings returns a string list for comma-separated values
(ini-based configurations) — every branded port was silently dropped when more
than one was configured. Both values are now read from their own key and with
the right type.

https://github.com/owncloud/client/pull/12597
https://github.com/owncloud/client/pull/12606
```

DeepDiver1975 (changes requested):
Docs-only cherry-pick, and the mechanics check out: the fix it documents is genuinely on `7.1` (`5a1d407f2`), `changelog/unreleased/` was empty apart from `.gitkeep` (7.1.0 is archived under `changelog/7.1.0_2026-06-08/`), the filename matches the calens primary ID, and the title is within the 80-char limit.

The content of the entry needs changes though — see the inline comment. In short:

1. The referenced commit also fixes OIDC `Ports` parsing on ini-based platforms (multi-port configs were silently dropped entirely), which arguably affects more admins than the prompt value and isn't mentioned at all.
2. There's no body paragraph, so calens renders a single unexplained line — and as the only entry for 7.1.1 so far, that would be the whole patch changelog.
3. The `7.1` backport PR (#12606) isn't linked, only the `master` PR, contrary to the convention on maintenance branches.

An applicable suggestion covering all three is in the inline comment.

changelog/unreleased/12597.md
erikjv: This is a cherry-picked commit. If changes are needed, the original one should be changed, not this one. It is also a bit late to do this right before v7.1.1 (final) is tagged.

DeepDiver1975: yes - changes apply to all changelog item and you need to adjust the one in master as well.

but it never made it to master either ....

erikjv: done. Will cherry-pick this to master when it has landed in the 7.1 branch

changelog/unreleased/12597.md
```diff
@@ -0,0 +1,13 @@
+Bugfix: Load OIDC configuration values from the right keys
+
+The OIDC `prompt` setting was read from the `Ports` configuration value,
+so a bogus `prompt` parameter was sent to the identity provider. The
+configured redirect `Ports` were also parsed as a single string, which
+is empty on platforms where QSettings returns a string list for
+comma-separated values (ini-based configurations) — every branded port
```
DeepDiver1975: Small factual nit: `branded` isn't the setting that was affected. Branded ports come from `Theme::oauthPorts()` via `loadOpenIdConfigFromTheme()`, which this code path never parsed — the ports that got dropped are the ones from the system config (`[OpenIDConnect] Ports=` in `/etc/ownCloud/ownCloud.ini`). As written, the admins who actually hit this won't recognise themselves.

```suggestion
comma-separated values (ini-based configurations) — every system-configured port
```

Rest of the entry looks fine to me.

DeepDiver1975 (approved):

PR #12624, opened 2026-08-26
chore: add 7.1.0 changelog

erikjv:


DeepDiver1975 (dismissed):
Thanks for pulling the 7.1.0 changelog together. There's a structural problem that needs fixing before this can go in, plus some duplicated notes.

## The core issue: `CHANGELOG.md` is not reproducible from the fragments in this PR

`CHANGELOG.md` here is **byte-identical** to `git show v7.1.0:CHANGELOG.md`:

```
$ diff <(git show v7.1.0:CHANGELOG.md) <(git show origin/pull/12624/head:CHANGELOG.md) && echo IDENTICAL
IDENTICAL
```

…but 5 of the 20 fragments were reworded relative to what `v7.1.0` actually shipped, and each of those edits violates a calens validation rule. So the generated file and its own sources disagree, and the post-merge `changelog / Generate Changelog` job in `.github/workflows/changelog.yml` will fail rather than reproduce this file.

I verified this by running the project's own generator (`docker run --rm -v $(pwd):$(pwd) -w $(pwd) toolhippie/calens:latest`, per `changelog/README.md`) against the PR head. **Reverting these 5 fragments to their `v7.1.0` content makes calens regenerate the committed `CHANGELOG.md` byte-for-byte.**

### 1–3. Invalid entry type `Feature:` (blocking)

`changelog/7.1.0_2026-06-08/12233.md`, `12367.md`, `12399.md` were changed from `Enhancement:` to `Feature:`:

```
entry type "Feature" is invalid, valid types: map[Bugfix:2 Change:3 Enhancement:4 Security:1]
```

`changelog/README.md` lists only `Bugfix, Change, Enhancement, Security`, and `grep -rl '^Feature:' changelog/` matches **none** of the 543 pre-existing fragments. All three shipped as `Enhancement:` at `v7.1.0`, and the `CHANGELOG.md` in this very PR renders them as "Enhancement" — so the fragments contradict the generated file.

Note calens aborts at the *first* invalid file, so fixing these one at a time costs three CI round-trips. Please fix all three together.

### 4–5. Titles over the 80-character limit (blocking)

```
title is too long (max 80 characters)
```

| file | current | len | shipped at v7.1.0 | len |
|---|---|---|---|---|
| `12487.md` | `Bugfix: Fix re-authentication prompt when access token expires but refresh token is still valid` | 95 | `Bugfix: Fix re-authentication prompt` | 36 |
| `12336.md` | `Bugfix: Fix sync of files with NFD/NFC filename normalization differences on macOS` | 82 | `Bugfix: Fix file sync with NFD/NFC filename normalization differences on macOS` | 78 |

In both cases the short form is what this PR's own `CHANGELOG.md` already contains.

## Duplicated release notes

Three entries re-announce changes that were already published in earlier releases, under the same PR numbers:

| new fragment | already released as | shipped title |
|---|---|---|
| `12093.md` | `changelog/6.0.0_2025-09-08/12093.md` | `Change: Update the sync exclude list for Infinite Scale` |
| `12245.md` | `changelog/6.0.0_2025-09-08/12245.md` | `Change: Bump minimum macOS version to 13.0` |
| `12305.md` | `changelog/6.0.1_2025-10-10/12305.md` | `Bugfix: Fix VFS checkbox in folder wizard` |

After this PR, `CHANGELOG.md` advertises each of these under two different releases. Concretely:

- **`12245.md`** — a user on macOS 12 reads the 7.1.0 notes and concludes 7.1.0 dropped their OS, when 6.0.0 did.
- **`12093.md`** — duplicate *and* less informative than the original, which named the extensions (`.space`, `.psec`).
- **`12305.md`** — the #12305 half already shipped in 6.0.1; only the #12144 work (selective-sync chooser disabled when VFS is forced, page skipped) is genuinely new in 7.1.0. Suggest scoping the entry to that.

## Contradiction within the release

`changelog/7.1.0_2026-06-08/12270.md` says Qt 6.8 "raises the minimum macOS deployment target to **12**", while `12245.md` in the same folder says macOS **13** is the minimum. Both render in the same 7.1.0 section, so readers get two different macOS floors with nothing to disambiguate them. The "12" figure is stale given the 13.0 bump shipped in 6.0.0 — suggest dropping that sentence or correcting it to 13.

## Checked and fine

- Date and compare link are correct — tag `v7.1.0` is dated 2026-06-08, matching the folder name and `compare/v6.0.3...v7.1.0`. The TOC anchor matches the template's generation rule.
- All 49 referenced PR numbers are genuine ancestors of `v7.1.0` — no misattributions.
- `changelog/unreleased/.gitkeep` survives; `12369.md` and `12374.md` were correctly moved and both did ship in 7.1.0. Nothing left behind or dropped.
- Filename-vs-first-link mismatches (`12233.md` linking #12189 first, `12305.md` linking #12144 first) are **not** a problem — calens ignores filenames and past releases do this routinely.
- The three failing build jobs look like infrastructure, not this PR: *"The job was not acquired by Runner of type hosted even after multiple attempts."*

---

Suggested fix: revert the 5 edited fragments to their `v7.1.0` content, drop or re-scope the 3 duplicates, and fix the macOS floor in `12270.md`. If the duplicates are dropped, `CHANGELOG.md` will need regenerating (and will then legitimately differ from the shipped one — that's expected and fine).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

erikjv:
Review done by @DeepDiver1975 has been fixed in subsequent commit.

modSpike (approved):

PR #12630, opened 2026-08-31
refactor: [DC-309] update docs for Theme

modSpike:
see https://kiteworks.atlassian.net/browse/DC-309

### DIFF
--- a/src/libsync/theme.h
+++ b/src/libsync/theme.h
@@ -17,9 +17,8 @@
 #include <QObject>
 
 
-#include <QFileInfo>
-#include <QObject>
-#include <QPalette>
+// #include <QFileInfo>
+#include <QColor>
 
 #include "owncloudlib.h"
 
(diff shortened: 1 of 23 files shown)


src/libsync/theme.h
```diff
@@ -173,97 +213,182 @@ class OWNCLOUDSYNC_EXPORT Theme : public QObject
...
     QUrl updateCheckUrl() const;
 
-
     /**
-     * If this returns true, the user cannot configure the proxy in the network settings.
-     * The proxy settings will be disabled in the configuration dialog.
-     * Default returns false.
+     * @brief forceSystemNetworkProxy
+     *
+     * If this returns true, the user cannot configure the network proxy in the application Settings.
```
erikjv: add: "and the system proxy is always used"

to make it clear that "no proxy" is NOT chosen. (the settings page has 3 options: "no proxy", "user defined", "system")

src/libsync/theme.h
```diff
@@ -173,97 +213,182 @@ class OWNCLOUDSYNC_EXPORT Theme : public QObject
...
+     * @return definitions for URL actions
+     */
     virtual QVector<std::tuple<QString, QString, QUrl>> urlActions() const;
 
     /**
-     * Set the default value for move to trash option
-     * Default: false
+     * @brief moveToTrashDefaultValue sets a default value for move-to-trash option in application Settings
+     *
+     * @default false
```
erikjv: we should probably switch this to `true`

src/libsync/theme.h
```diff
@@ -173,97 +213,182 @@ class OWNCLOUDSYNC_EXPORT Theme : public QObject
...
 
     /**
-     * @brief Allow the system configuration to override theme values.
+     * @brief Allow a system configuration to override theme values related to the OIDC parameters.
+     *
      * @default false
+     *
+     * @return whether a system configuration can be used to define the OIDC values.
      */
     virtual bool allowSystemConfigOverrides() const;
```
erikjv: move this to right below the OIDC methods

src/libsync/theme.h
erikjv: this comment moved, it was about the text below the brief line

erikjv (changes requested):

erikjv (approved):

